PR #14, opened 2017-07-03
Enable testing under Python 3.4

leouieda:


leouieda:
Doesn't work because of some missing dependencies. Not worth fixing.

Issue #62, opened 2017-11-28
Implement argument in Figure.savefig to display figure


leouieda:
Add an extra argument (`show=False`) to `Figure.savefig` to open the figure in an external viewer (use `launch_external_viewer`). This operation is non-blocking so the script will keep going. This is a way for scripts to display the figure without relying on `Figure.show`, which would not work because the temp directory with the figure is cleaned when the `Figure` object is deleted. It makes more sense because scripts will want a figure file to be saved, unlike in notebooks where you want to see them inline most of the time.

PR #72, opened 2017-11-30
add show=false to Figure.savefig()

seisman:
First attempt to implement issue #62.

Free free to correct me if anything goes wrong.

### DIFF
--- a/doc/first-steps.ipynb
+++ b/doc/first-steps.ipynb
@@ -97,7 +97,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
+   },
    "outputs": [],
    "source": [
     "fig.show()"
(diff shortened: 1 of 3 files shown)


doc/first-steps.ipynb
```diff
@@ -205,6 +207,24 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "fig.savefig('first-steps-central-america.png', show=True)"
```
leouieda: This might be a bit problematic as a code cell. The notebook is executed every time we build the documentation and it would be bad for a window to open whenever that happens. 

Could you please include this as a Markdown code block inside the previous cell? Like I did for `fig.show(external=True)`.

seisman:
@leouieda I have made the changes and rebased it on the latest master.

leouieda:
Merged. Thanks!

Issue #92, opened 2017-12-04
Wrapper for 'gmt which'


leouieda:
The wrapper should be a function in `gmt/modules.py`.  See how other wrappers are built in `gmt/base_plotting.py`.

This module will be useful to download the new GMT tutorial and topography data using the `@filename` syntax. For example, running for the first time `gmt which @tut_quakes.ngdc` will download the data to the GMT hidden folder and provide the file name to stdout. We will use this to make a higher level `gmt.datasets` package that loads these data as numpy arrays and xarray grids.

The wrapper will need to capture stdout and return it. See discussion on #43

PR #106, opened 2017-12-30
wrapper for gmt info

### LINKED ISSUE #43: Wrapper for gmt info
Create a wrapper function for `gmt info`. Data input should be a file name `data='some_file.txt'`. Doesn't really make much sense to read in numpy arrays for this because most of it can be done using numpy functions.

See how other modules are wrapped, for example in `gmt/base_plotting.py`.

seisman:
Fix #43. Still work in progress. Any suggestions?


leouieda:
@seisman mostly looks good to me. I have some minor comments that I'll leave inline.

gmt/__init__.py
```diff
@@ -17,6 +17,7 @@
 # Import modules to make the high-level GMT Python API
 from .session_management import begin as _begin, end as _end
 from .figure import Figure
+from .modules import *
```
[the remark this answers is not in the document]
leouieda: Please don't `import *`. It's better to explicitly import the functions that should be public.

gmt/modules.py
```diff
@@ -0,0 +1,31 @@
...
+def info(data, **kwargs):
+    """
+    Get information about data tables.
+
+    {gmt_module_docs}
+
+    Parameters
+    ----------
+    data: str
+        A data file name.
```
leouieda: Let's keep it only for data files for now. We can think about adding pandas.Dataframe later but I don't know if it will be helpful. 

In this case, `fname` is a better name for the argument. I'll probably refactor `plot` to use `fname` for files as well.

gmt/modules.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    assert isinstance(data, str), 'Only accepts file names.'
+
+    with NamedTemporaryFile(prefix="gmt-python-", suffix='.txt',
+                            mode='r') as tmpfile:
+        arg_str = ' '.join([data, build_arg_string(kwargs),
+                           "->" + tmpfile.name])
+        with LibGMT() as lib:
+            lib.call_module('gmtinfo', arg_str)
```
leouieda: I don't think you need the "gmt" in the name.

seisman:
Maybe I need to add more tests?

leouieda:
>  Maybe I need to add more tests?

@seisman yes please! I try to have as many configurations as possible tested. The one you have now looks good for the basic functionality.

Would you mind adding some more information to the docstring? You can copy some of the basic descriptions from the GMT docs. Not all options need to be included, just the most common ones. We can think of some aliases for them as well.

Do you think we should print the output instead of returning it? There could be a flag to trigger a return instead of printing. Either way, if the returned string is just a list of numbers, it would be better to return an array/list of the number instead.

leouieda:
@seisman I'm updating the branch and merging this as is. I'm working on #92 and I'll make the changes mentioned above. 

Thanks for putting in all the work!

Issue #88, opened 2017-12-04
Utility functions for creating temporary files


leouieda:
A lot of the clib tests need dummy file names for dumping output and then reading it out. Some are using `NamedTemporaryFile` but some set the file name by hand. They all need to create a *closed* file so that GMT can write to it. It also needs to be deleted once the function is done with the file. Ideally, it should be cleaned even if an exception occurs. Using plain `NamedTemporaryFile` won't work because it keeps the file open for writing in Python.

Create a `GMTTempFile` context manager class in `gmt/utils.py` that creates a temp file with `NamedTemporaryFile` and closes it immediately. It should include the prefix `gmt-python-` and suffix that is user specified (optional). When exiting, delete the file using `os.remove`. The class should include a method `read` to open and return the entire contents of the file. It should also replace the tabs that GMT uses with spaces to make it more Python friendly (with an options to keep them).

leouieda:
A second PR should replace this class in all cases where a temporary file is being used for testing.

PR #107, opened 2017-12-30
Use "with" when generating unique figure names

seisman:
`with` statement is more pythonic.


seisman:
Wait for #88.

leouieda:
@seisman thanks for the refactor!

PR #110, opened 2018-01-02
Raise an informative exception if IPython is not available

### LINKED ISSUE #108: Unhelpful exception raised when IPython is not installed
gmt/tests/test_figure.py::test_figure_show <- /soft/anaconda2/envs/python3.6/lib/python3.6/site-packages/gmt_python-0+unknown-py3.6.egg/gmt/tests/test_figure.py FAILED [ 80%]
E TypeError: 'NoneType' object is not callable

/soft/anaconda2/envs/python3.6/lib/python3.6/site-packages/gmt_python-0+unknown-py3.6.egg/gmt/figure.py:308: TypeError

It doesn't work when I am tring
fig_alias = gmt.Figure()
fig_alias.coast(R='-90/-70/0/20', J='M6i', G='gray', S='blue', B=True)
fig_alias.show()
File "<stdin>", line 1, in <module>
File "/users1j/cliu/soft/anaconda2/envs/python3.6/lib/python3.6/site-packages/gmt_python-0+unknown-py3.6.egg/gmt/figure.py", line 308, in show
img = Image(data=png, width=width)
TypeError: 'NoneType' object is not callable

seisman:
Fix #108.

When IPython is not installed, `gmt.Figure().show()` will report an error. In this case, it's better to open the figure in an external window even without the option `external=True`.

### DIFF
--- a/gmt/figure.py
+++ b/gmt/figure.py
@@ -297,7 +297,7 @@ class Figure(BasePlotting):
             Only if ``external=False``.
 
         """
-        if external or Image is None:
+        if external:
             pdf = self._preview(fmt='pdf', dpi=600, anti_alias=False,
                                 as_bytes=False)
             launch_external_viewer(pdf)
--- a/gmt/figure.py
+++ b/gmt/figure.py
@@ -297,7 +297,7 @@ class Figure(BasePlotting):
             Only if ``external=False``.
 
         """
-        if external:
+        if external or Image is None:
             pdf = self._preview(fmt='pdf', dpi=600, anti_alias=False,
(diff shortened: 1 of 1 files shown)


leouieda:
@seisman I don't know if I like this behavior. It's not doing what one would expect with `external=False`. It's better to raise an exception. But in this case, we should check if `Image is None` and then raise an informative exception like "Cannot find IPython. Make sure you have it installed or use 'external=True' to open in an external viewer.".

leouieda:
@seisman perfect, thanks!

PR #119, opened 2018-01-15
rename OSX to macOS

seisman:
`OSX` is renamed to `macOS` since 2016.


leouieda:
It's hard to keep up with the apple names :) thanks!

PR #126, opened 2018-02-13
Implement I/O of color palettes

### LINKED ISSUE #109: how to build a colorbar
I want to know how to add a colorbar on the map

leouieda:
Fixes #109

Implement the infrastructure needed to make and use GMT color palettes (CPTs).

Creating a CPT:
1. Create a virtual file for output of type GMT_IS_PALETTE
2. Call makecpt using the virtual file as output
3. Create an empty `Palette` structure
4. Copy the GMT_PALETTE from the virtual file into the Palette

Using a CPT:
1. Pass in a Palette structure generated using the above.
2. Open a virtual file for input and pass in the Palette as 'data' (cast to
   void pointer)
3. Use the virtual file in the -C/cmap option

For all this to work, need Python wrappers for the GMT_PALETTE structure. This
is mostly because the creation of the CPT is done in a different session (when
running makecpt). So we can't use the GMT allocated CPT as a void pointer like
we do for the rest of the data. We need a structure which Python controls the
memory.

### DIFF
--- a/gmt/clib/core.py
+++ b/gmt/clib/core.py
@@ -100,6 +100,7 @@ class LibGMT():  # pylint: disable=too-many-instance-attributes
         self._c_write_data = None
         self._c_open_virtualfile = None
         self._c_close_virtualfile = None
+        self._c_read_virtualfile = None
         self._bind_clib_functions(libname)
 
     @property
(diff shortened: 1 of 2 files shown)


leouieda:
Turns out that copying the CPT from GMT allocated memory to Python memory will be a lot of work. I'll leave this open for now and implement an easier workaround using temp files instead. Once that is working, I'll substitute the inner workings with the ctypes structures from this PR.

leouieda:
Closing as things have changed too much since this was opened.

PR #133, opened 2018-02-16
Remove portrait option that has been deprecated

leouieda:
The latest version of GMT deprecated option -P. Remove it from the library.


leouieda:
Waiting for the new GMT conda package to build before this can pass.

PR #138, opened 2018-03-08
check if xdg-open exists for linux OS

seisman:

`xdg-open` may be not installed by default for some Linux OS (for example, Ubuntu 16.04 LTS). It's better to check if `xdg-open` exists before using it.


leouieda:
@seisman this is great, thanks!

PR #141, opened 2018-03-10
Generalize vector_to_arrays

### LINKED ISSUE #139: Let Figure.plot() accept list as arguments
*If this is a **question or feature request**, delete the above lines and instead:*

* Describe the feature you desire in as much detail as possible:
Current Figure.plot() doesn't accept lists of x and y as arguments. Code below will report an error:

```python
import gmt

region = [150, 240, -10, 60]
x = [160, 170, 180]
y = [20, 30, 50]

fig = gmt.Figure()
# Create a 6x6 inch basemap using the data region
fig.basemap(region=region, projection='X6i', frame=True)
# Plot using triangles (i) of 0.3 cm
fig.plot(x, y, style='i0.3c', color='black')
fig.show()
```

* Are you willing to try to implement and maintain this feature? 
  We'll help you through the process.

  Yes.

seisman:
Fixes #139.

### DIFF
--- a/gmt/clib/utils.py
+++ b/gmt/clib/utils.py
@@ -48,7 +48,7 @@ def vectors_to_arrays(vectors):
     True
     >>> all(isinstance(i, np.ndarray) for i in arrays)
     True
-    >>> data = [[1, 2], [3, 4], [5, 6]]
+    >>> data = [[1, 2], (3, 4), range(5, 7)]
     >>> all(isinstance(i, np.ndarray) for i in vectors_to_arrays(data))
     True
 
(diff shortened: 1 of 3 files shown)


gmt/clib/utils.py
```diff
@@ -115,18 +120,22 @@ def _series_to_array(vector):
...
+    >>> type(_as_array(np.array([5, 6, 7])))
+    <class 'numpy.ndarray'>
+    >>> type(_as_array([3, 4, 5]))
     <class 'numpy.ndarray'>
 
     """
     if isinstance(vector, pandas.Series):
         return vector.as_matrix()
+    elif isinstance(vector, list):
+        return np.array(vector)
```
leouieda: @seisman just found out that you can use [numpy.asarray](https://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html) to do the conversion. In this case, even if the object is already an array the memory is not copied. So you can replace the isinstance check with:

```python
return np.asarray(vector)
```

This will also handle tuples, so we get an added bonus.

leouieda:
@seisman perfect! Merging.

PR #155, opened 2018-03-31
Custom exception for invalid inputs

### LINKED ISSUE #70: Create an exception for invalid input
Right now, both invalid inputs and failures (status != 0) raise `GMTCLibError`. This doesn't help with the testing because I can't tell which one was raised. It would be better to have a `GMTCLibInvalidInput` (and a `GMTInvalidInput` as well) exception to keep things separate. 

All functions/methods in `gmt/clib/` would need to modified to raise this exception instead. The tests in `gmt/tests/test_clib.py` would need to modified to check for this exception.

leouieda:
Fixes #70

Defines the exception GMTInvalidInput and uses it throughout the
library.
Include a section in CONTRIBUTING.md about checking inputs and outputs and
which exceptions to raise.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -258,9 +258,10 @@ necessary), but sometimes inputs must have a given set of properties or be from
 a predefined list of values.
 
 In these cases, an exceptions should be raised to signal the invalid
-inputs/outputs. We provide custom exceptions for this (see the API reference).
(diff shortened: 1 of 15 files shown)


leouieda:
@seisman just a heads-up about this change.

seisman:
Got it. The change seems good to me.

PR #161, opened 2018-04-14
Test if has any conflicts with matplotlib and basemap

### LINKED ISSUE #66: Test with matplotlib and basemap installed
Create a test environment in the CIs that installs matplotlib and basemap to make sure we don't get any conflicts between the libraries.

Pointed out in #32

seisman:
Fixes #66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,10 @@ matrix:
               #- PYTHON=3.6
               #- COVERAGE=true
               #- BUILD_DOCS=true
+        - os: linux
+          env:
+              - PYTHON=3.6
+              - EXTRA_PACKAGES="matplotlib basemap"
 
 before_install:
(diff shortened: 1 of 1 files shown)


leouieda:
@seisman thanks for this! I think the failure is because the install is mixing packages from the default and conda-forge channels. Matplotlib is already being installed because we use it for testing. The easiest solution is to just add `basemap` to the `requirements.txt` file instead of creating a separate build just for this.

seisman:
Didn't know that there are some incompatibility between the default channel and conda-forge channel. 
I'll close this PR and open a new one.

leouieda:
Yeah, that's one of the tricky parts about using conda and the reason why I included the conda env part in the install instructions. See https://conda-forge.org/docs/conda-forge_gotchas.html#using-multiple-channels

Issue #164, opened 2018-04-20
Segfault on MacOS during LibGMT.extract_region


chandrex:
<img width="911" alt="screen shot 2018-04-20 at 16 54 44" src="https://user-images.githubusercontent.com/37407269/39061246-94aed9da-44bb-11e8-945b-cff9c77a53d9.png">
Hi, I've been trying to install gmt-python for the past day and a half with no avail. I am on macOS 10.13.4, Python 3.6.5 (installed with Anaconda 3). I've been following all of the installation instruction, and everything seems fine except when I get to the very end, with the gmt.test().

I am in the conda environment (using source activate gmt-python), and trying this in the Python interpreter. This is what happens: 

>>> **(gmt-python) Chandras-MacBook-Pro:~ chandra$** python
Python 3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:44:09)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> >>> import gmt
>>> gmt.test()

returns

>>> ============================= test session starts ==============================
platform darwin -- Python 3.6.5, pytest-3.5.0, py-1.5.3, pluggy-0.6.0 -- /anaconda3/envs/gmt-python/bin/python
cachedir: .pytest_cache
Matplotlib: 2.2.2
Freetype: 2.8.1
rootdir: /Users/chandra, inifile:
plugins: mpl-0.9
collected 126 items
gmt/base_plotting.py::gmt.base_plotting.BasePlotting._preprocess <- ../../anaconda3/envs/gmt-python/lib/python3.6/site-packages/gmt/base_plotting.py **PASSED [  0%]**
gmt/figure.py::gmt.figure.Figure <- ../../anaconda3/envs/gmt-python/lib/python3.6/site-packages/gmt/figure.py **FAILED [  1%]**
gmt/clib/core.py::gmt.clib.core.LibGMT <- ../../anaconda3/envs/gmt-python/lib/python3.6/site-packages/gmt/clib/core.py **PASSED [  2%]**
gmt/clib/core.py::gmt.clib.core.LibGMT._check_dtype_and_dim <- ../../anaconda3/envs/gmt-python/lib/python3.6/site-packages/gmt/clib/core.py **PASSED [  3%]**
[...]

PR #180, opened 2018-05-11
Re-enable Mac testing on Travis

leouieda:
There seems to be a segfault problem (#164) that doesn't happen on
Linux. I'm re-enabling the Mac build to check if it happens on Travis as
well.


leouieda:
Figured out the problem and fixed it in #188

PR #186, opened 2018-06-14
Use Black for code formatting

leouieda:
Following PEP8 is a bit annoying and requires a lot of hand formatting.
Black automatically formats code following PEP8 with few exceptions.
Most notably the line length is 88 instead of 80, which is actually a
lot better.

Use `black` to format all our code and adjust the CI configurations to
match.

Removed the `pep8` and `lint` targets from the `Makefile` (only `check`
exists now since black is very fast).
Introduces the `format` target to run black on the entire code.
Now we don't need to check for PEP8, just run `make format` and
everything will be fine.
We run the checks on the CI to make sure black was run on the code.


leouieda:
@seisman just a heads-up on the changes in PR. It's not a big change and I hope it'll make things easier in the long run.

PR #187, opened 2018-06-16
Use the Fatiando a Terra CI scripts

leouieda:
Fatiando provides scripts for handling all of the CI tasks we need:
https://github.com/fatiando/continuous-integration
Use them instead of rewriting everything.


leouieda:
Done in #188 instead

PR #204, opened 2018-06-30
Support Python 3.7

### LINKED ISSUE #190: Add Python 3.7
This is a reminder to add Python 3.7 to testing which is planned to be [released on June 27, 2018](https://www.python.org/dev/peps/pep-0537/#id6). 

- [x] `.travis.yml`
- [x] `CLASSIFIERS` in `setup.py`

seisman:
Fixes #190.

Add tests for Python 3.7 while keeping Python 3.6 as the default version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,10 @@ matrix:
               - BUILD_DOCS=true
               - DEPLOY_DOCS=true
               - DEPLOY_PYPI=true
+        - os: linux
+          env:
+              - PYTHON=3.7
+              - BUILD_DOCS=true
         - os: osx
           env:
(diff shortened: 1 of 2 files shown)


leouieda:
Thanks @seisman. It looks like it will still be a while before this will work. conda-forge has very few packages for 3.7 still and when we install `requirements.txt` python gets downgraded to 3.6.

leouieda:
We can leave this open until there is more support from other packages. I don't think it will be long.

seisman:
@leouieda I think this PR is ready to be merged.

leouieda:
:+1:

PR #210, opened 2018-07-08
Refactor gmt.clib 

leouieda:
Renamed `LibGMT` to `Session` because that is what it 
actually is (the GMT API session). 
Rename `clib.core` to `clib.session` and break up `clib.utils`
into `clib.loading` (functions for loading libgmt) and 
`clib.conversion` (functions that handle the ctypes 
conversions from numpy, etc).
The `create` and `destroy` methods now set the session
pointer (`Session.session_pointer`), which simplifies
the `__enter__` and `__exit__` methods.
`Session.get_constant` is now implemented as 
`Session.__getitem__` so getting a constant is now
much shorter: `self["GMT_IS_MATRIX"]`.
The constant values are stored as module variables instead
of in the class.

Re-enables intersphinx.


leouieda:
I'm still not entirely happy with the virtual file mechanics. It would make a lot more sense for the virtual file to be its own class but that doesn't match with the way GMT does things. For one, it would need the current `clib.Session` object as input, which would look silly in Python.

PR #212, opened 2018-07-12
Added basic contour functionality

mjziebarth:
Hi everyone,
started using gmt-python yesterday and really like the project. Since I'll be writing some solutions for my own use, I figured I may contribute them right away. So, I am leaning a bit on the pull-request-style of development and don't expect this to be ready to merge straight away.

## Changes proposed in this pull request
- Add function to call gmt contour from BasePlotting.

### Note
The code is copied from the plot function and is very basic so far. Tested only locally for crashes in Figure so far.

Looking forward to your input!

PR #216, opened 2018-07-24
Don't install black on TravisCI automatically

leouieda:
Black is Python 3.6 only so installing it through the requirements.txt
file was causing a silent upgrade of Python.

Will help with the problems in #212


leouieda:
Build is failing again. It's something to do with `libboost` being from defaults. I have no idea why it would be. Maybe we should install all the dependencies of GMT separately to enforce using conda-forge.

leouieda:
Got implemented in #224

Issue #219, opened 2018-07-26
Convenience classes and additional keywords


mjziebarth:
## Add more convenience classes or keyword parameters to facilitate ease of use

First off: Thanks for the work so far!

Coming from the scipy ecosystem and being an avid matplotlib user, I figured I'd leave some feedback from my experience starting with gmt-python. Since I'll likely be using it during the next couple of years, I'm willing to implement the ideas I have, but they may be quite big so I figured it would be best to discuss them beforehand.

I really like gmt as a tool and the achievable results, but the learning curve is quite steep. Especially since the documentation is not really concise or easy to read and there does not seem to be a big question-answering community around, at least in my limited experience. So I feel like one of the big advantages of gmt-python besides the native interaction with python would be to use the quite common python-intuitivity.

That being said, here's some ideas I had (some obviously inspired by my experiences with matplotlib):

 - Create convenience classes that wrap the interaction with some parts of the gmt workflow. E.g. a Grid-class that can be created from numpy arrays and passed to the grdimage method, handling the xyz2grid calls internally (possibly with caching and whatnot). Or a Colormap class that handles the .cpt creation automatically (like the matplotlib colormaps). Or projection classes that have all the required configuration inside their __init__s

 - Add further self-describing keywords (like https://github.com/GenericMappingTools/gmt-python/issues/48) that are automatically converted to the gmt argument string inside e.g. Figure.plot.

In the end, this mostly boils down to creating more logic inside the methods and wrapping the gmt syntax into the structure supplied by the python language. I have some conceptional code which I use myself in [this](https://github.com/mjziebarth/gmt-python-extensions) repository, just to give you a rough idea.

What do you think about the idea? In case the two points seem generally interesting, I'd be willing to discuss in detail what could be possible implementations and develop them.

**Are you willing to help implement and maintain this feature?**

Yes

weiji14:
Hi @mjziebarth :wave:, I know it's been almost 2 years since this was posted, and you'd probably moved on with other things in life, but just to give you (and others) a bit of an update on what we have/haven't achieved in terms of your ideas:

> Create convenience classes that wrap the interaction with some parts of the gmt workflow. E.g. a Grid-class that can be created from numpy arrays and passed to the grdimage method, handling the xyz2grid calls internally (possibly with caching and whatnot).

[...]

PR #225, opened 2018-08-19
grdcontour Implementation

akshmakov:
Implemented `Figure.grdcontour` that uses `grdcontour` gmt module

Plots grids to contours, with its own set of distinct inputs from `pscontour`

Test and Initial Documentation is complete

**TODO**

- [x] Expand Docstring for function.
- [x] Review chosen shorthand arguments for consistency with other modules
- [x] Refine tests
- [x] Regenerate baseline images.


akshmakov:
@leouieda Build fails, due to test image mismatches, I suspect the density of features is playing a role, since 2 fail and 2 do not, and the ones that fail are denser contours. I will make a todo to refine the tests, but I suspect regenerating the images on the same machine (like what needed to be done for #212) will help. Not sure how to handle it

leouieda:
@akshmakov I don't know why, but in #212 the image generated on a Mac fails the linux tests but the image I generated on my linux machine passes all tests. Let me know when you're done with the code and I'll regenerate the images on my machine.

gmt/base_plotting.py
```diff
@@ -126,7 +126,48 @@ def coast(self, **kwargs):
...
 
     @fmt_docstring
-    @use_alias(R="region", J="projection", B="frame", I="shading", C="cmap")
+    @use_alias(
+        R="region",
+        J="projection",
+        A="annotation_interval",
+        C="contour_interval",
+        B="frame",
+        G="label_placement",
```
leouieda: Choosing these names is really hard. I would rename them to single words:
* `A="annotation"`
* `C="interval"` (contour is already in the method name so `fig.grdcontour(contour_interval=..)` seems a bit repetitive)
* `G="label"`

I'm not sure about `limit`. This type of things is often called a `range` in other parts of GMT and some Python functions. But `range` is a builtin function so we should use `ranges` instead. I'm not a huge fan of that. Maybe we can leave `limit` for now until we can think of something better. This is one of those options that should actually be multiple options.

leouieda: Please, feel free to disagree :)  I'm not set on these names and am willing to change

akshmakov: You can tell my C bias as this was my attempt at **less verbosity**. 

One thought I had is that `-C` can be aliased as `cmap` as per `countour` since it accepts a `CPT` file  input as well, it just also accepts a number for contour intervals so is not compatible the other way.  @mjziebarth   maybe has some thoughts as well?

akshmakov: @leouieda One thing that happens is that something like `contour_interval`  increases the verbosity  when using the functions directly, but makes it clearer to construct keyword dictionaries. Application designers are more likely to use this pattern but it is still useful for managing plot options in scripts.

e.g.

    OPTS={ 
           projection:"W0/6i",
           coast:1, 
           borders:1,
           annotation_interval:1000,
           contour_interval:100,
     }
     fig.coast(**OPTS)
     fig.grdcontour(grd, **OPTS)
     

vs 

    OPTS={ 
           projection:"W0/6i",
           annotation:1000,
           interval:100,
     }
     # ...


More verbose names are easier to work with, and additionally reusing the keywords for incompatible  parameters  can prevent user from using a single dictionary to store common keyword arguments, but  on the other hand this is already true in some cases and is certainly true for many GMT single letter arguments...

In any case I will take your advice for the names,  and instead put my thoughts in #219 because I think this is a bigger discussion than `grdcontour`

mjziebarth: From someone new to GMT: The more verbose names may make the difference between intuitively being able to write the code and having to consult the man pages.
But I agree with @akshmakov that this is probably best for general discussion e.g. in #219

leouieda: Yeah, I agree. All of this will have to be updated at some point. For now, we need some time to use these functions and see where they fall short. No point in trying to predict all possible use cases.

akshmakov: Updated with the following simplified keywords

```python
A="annotation",
B="frame",
C="interval",
G="label_placement",
J="projection",
L="limit",
Q="cut",
R="region",
S="resample",
U="logo",
W="pen",
```

akshmakov:
@leouieda Finished expanded docs and keyword expansion. Removed WIP Status, ready to merge pending updated baseline plots

Are you on the island right now? Stay safe! :cloud_with_rain: 

P.s. Might baseline issue be due to default fonts? I am generating on linux machines as well and they fail on travis due to baseline mismatch, but pass on my machine

akshmakov:
image

leouieda:
@akshmakov thanks again for the contribution! I re-generated the baselines and fixed some code formatting issues. Once CI gods are happy, I'll merge this in!

leouieda:
Thanks @akshmakov! Please feel free to add yourself to the `AUTHORS.md` file (through another PR).

akshmakov:
@leouieda will do on another commit, hopefully for another issue, thank you for the baseline images, I'm happy to contribute!

Is there a style guide that I can review? Would `black` flag these instances?  I only recently installed and learned to use it, so my future PR's shouldn't be annoying 

P.S. I have some suspicion that baseline mismatches may be related to default system fonts.

PR #245, opened 2018-11-25
Wrap surface

### LINKED ISSUE #243: Wrapper for surface

Hi there, I'd like to help implement the [`surface`](https://gmt.soest.hawaii.edu/doc/latest/surface.html) function which "Grids table data using adjustable tension continuous curvature splines". There's a lot of plotting related functionality in gmt-python now, but there's more to GMT than that! Basically, this feature request aims to bring some data processing functionality into gmt-python. 

The [`surface`](https://gmt.soest.hawaii.edu/doc/latest/surface.html) function is actually just one small part of the 'Gridding of Data Tables' family of functions in GMT (see [module quickref](https://gmt.soest.hawaii.edu/doc/latest/quick_ref.html)), but I'd like to start with `surface` because 1) It appears to be a common gridding function (there's a [tutorial](https://gmt.soest.hawaii.edu/doc/latest/GMT_Tutorial.html#gridding-with-splines-in-tension) for it) and 2) I'm using it.

#### Things to keep in mind:

- Some future proofing necessary. The code will be eventually be shared with other grid functions like [`nearneighbor`](https://gmt.soest.hawaii.edu/doc/latest/nearneighbor.html), [`sphinterpolate`](https://gmt.soest.hawaii.edu/doc/latest/sphinterpolate.html), etc.
- Agree on some aliases first with #79 (Easier to read aliases in docstrings) (see also #48). Minimum required argument for `surface` is `-G` output file name, `-l` grid spacing and `-R` region of interest.
- Add a sample dataset, `@tut_ship.xyz` used in the ['Gridding of arbitrarily spaced data' tutorial](https://gmt.soest.hawaii.edu/doc/latest/GMT_Tutorial.html#gridding-of-arbitrarily-spaced-data).

#### Idea for structure:

Store the `surface` function in a `Gridding` class under `gridding.py`. Similar to how e.g. `psconvert` is in the Figure class under [`figure.py`](https://github.com/GenericMappingTools/gmt-python/blob/master/gmt/figure.py). This means we can store future gridding functions like `nearneighbor` in the same place.

```python
# File at gmt/gridding.py
"""
Define the Gridding class that handles all gridding.
"""

class Gridding:
    def surface(x=None, y=None, z=None, data=None, **kwargs):
        raise NotImplementedError("GMT gridding surface")
```

This is my initial stab at a test case file. Loosely based on the following bash code used in the tutorial.

```bash
    gmt blockmedian -R245/255/20/30 -I5m -V @tut_ship.xyz > ship_5m.xyz
    gmt surface ship_5m.xyz -R245/255/20/30 -I5m -Gship.nc -V
```

```python
# File at gmt/tests/test_surface.py
"""
Test Gridding.surface
"""

from .. import Gridding
from ..datasets import load_tut_ship

def test_surface():
    ship_data = load_tut_ship()
    outputfile = Gridding.surface(
        x=ship_data.x,
        y=ship_data.y,
        z=ship_data.z,
        G="ship.nc",
        I="5m",
        R="245/255/20/30",
    )
    return outputfile
```

I've actually got some boilerplate code ready on my fork. Just wanting to get the green light and check if I've missed out on any important details.

**Are you willing to help implement and maintain this feature?** Yes!

weiji14:

Wrapping the [surface](https://gmt.soest.hawaii.edu/doc/latest/surface.html) gridding function. A new load_tut_ship() dataset function was also added as an example for the tests.

Question: Do I put the API docs for surface under [Data Processing](https://www.gmtpython.xyz/latest/api/index.html#data-processing) (along with gmt.info and gmt.grdinfo)? Or have it under another section?

Fixes #243

**Reminders**


gmt/gridding.py
```diff
@@ -0,0 +1,77 @@
...
+                file_context = lib.virtualfile_from_vectors(x, y, z)
+            else:
+                raise GMTInvalidInput("Unrecognized data type: {}".format(type(data)))
+            with file_context as infile:
+                if "G" not in kwargs.keys():
+                    kwargs.update({"G": tmpfile.name})
+                outfile = kwargs["G"]
+                arg_str = " ".join([infile, build_arg_string(kwargs)])
+                lib.call_module(module="surface", args=arg_str)
+        result = xr.open_dataset(outfile)
```
leouieda:
```suggestion
        result = xr.open_dataset(outfile).load()
        result.close()
```

xarray sometimes doesn't really load the data into memory, it creates a memory map to the file on disk. Since this is a temporary file, we have to make sure it's loaded into memory and the file is closed.

gmt/tests/test_surface.py
```diff
@@ -0,0 +1,96 @@
...
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
+TEMP_GRID = os.path.join(TEST_DATA_DIR, "tmp_grid.nc")
+
+
+def test_surface_input_file():
+    """
+    Run surface by passing in a filename
+    """
+    fname = which("@tut_ship.xyz", download="c")
+    output = surface(data=fname, I="5m", R="245/255/20/30")
```
leouieda: Please use the full argument names (`region` instead of `R`, etc) in the code. We don't want to encourage people to use the single letter arguments.

gmt/datasets/tutorial.py
```diff
@@ -38,6 +38,27 @@ def load_japan_quakes():
     return data
 
 
+def load_tut_ship():
```
leouieda: Could you rename this function to `load_sample_bathymetry`? I really don't like the name `tut_ship`.

gmt/datasets/tutorial.py
```diff
@@ -38,6 +38,27 @@ def load_japan_quakes():
...
+    first time you invoke this function. Afterwards, it will load the data from
+    the cache. So you'll need an internet connection the first time around.
+
+    Returns
+    -------
+    data :  pandas.Dataframe
+        The data table. Columns are x, y, and z.
+    """
+    fname = which("@tut_ship.xyz", download="c")
+    data = pd.read_csv(fname, sep="\t", header=None, names=["x", "y", "z"])
```
leouieda: I always prefer using longer explicit names to avoid confusion. In this case, avoid using `x, y, z`. The coordinates are actually longitude, latitude, and bathymetry.

```suggestion
    data = pd.read_csv(fname, sep="\t", header=None, names=["longitude", "latitude", "bathymetry"])
```

gmt/tests/test_surface.py
leouieda: I know we're doing this in other places but I'm trying to change that.

gmt/tests/test_surface.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    Run surface with the -Goutputfile.nc parameter.
+    """
+    ship_data = load_tut_ship()
+    data = ship_data.values  # convert pandas.DataFrame to numpy.ndarray
+    try:
+        output = surface(data=data, I="5m", R="245/255/20/30", G=TEMP_GRID)
+        assert os.path.exists(TEMP_GRID)
+        grid = xr.open_dataset(TEMP_GRID)
+        assert output == grid  # check that original output is same as that from file
```
leouieda: This comparison might not actually work. I'm also not sure I like the fact that you get an output grid when specifying an output file. Maybe we should stick with one or the other. I would suggest not loading the grid if `G` is provided and returning nothing in that case.

leouieda:
@weiji14 thanks for submitting this PR! I have a few comments that need to addressed first.

I wasn't sure if we should merge this yet because I'm making some big changes in the C lib wrapper and how we build the wrapper modules. But I'll merge this in as a test case for some of the changes I'm planning. 

Let me know if you need any help.

> Do I put the API docs for surface under Data Processing (along with gmt.info and gmt.grdinfo)? 

Yep, you can put it under the "tabular data" section.

gmt/datasets/tutorial.py
weiji14: Sounds good, done!

gmt/datasets/tutorial.py
weiji14: Ok, I've made the change and noted in the commit message that the sample dataset actually has longitude in the 0-360 format.

gmt/tests/test_surface.py
weiji14: Yeah I do prefer the explicit style. I've made the change, but also left in a test for the short form (-I and -R).

gmt/gridding.py
weiji14: Let me try testing `xr.open_dataset` on a big dataset (>10GB) I have first. I feel like this should be wrapped around a python `with` statement to make sure the file does close even if it's half loaded or something.

gmt/tests/test_surface.py
weiji14: Yes it's not a particularly well thought out test to be honest. The `-G`outputfile.nc parameter is actually a required argument for [`surface`](https://gmt.soest.hawaii.edu/doc/latest/surface.html), but it's technically optional in this current implementation. I've looked at the other gridding modules and they're not consistent. For example, `-G` is required for [`nearneighbor`](https://gmt.soest.hawaii.edu/doc/latest/nearneighbor.html) but optional for [`triangulate`](https://gmt.soest.hawaii.edu/doc/latest/triangulate.html) and [`greenspline`](https://gmt.soest.hawaii.edu/doc/latest/greenspline.html) (they write to stdout).

I could write a lot of if/then statements to check if the `-G`outputfile.nc parameter is used and return either None or an xarray.Dataset, but I'm not sure if that will be intuitive for other users. A user setting a `-G` parameter will expect `gmt.surface` to output to a file, but would they expect it to return None?

gmt/gridding.py
weiji14: Ok, I've forced the data to be loaded/persisted, using:

```python
    with xr.open_dataset(outfile) as dataset:
        result = dataset.load()
```

Ideally there would be a big dataset we can add to our test suite to make sure this works, but it might slow things down too much so I'll leave it at that.

weiji14:
@leouieda thanks for taking the time to review this! I've made some commits to address the issues you've highlighted, and have now added `surface` to the docs.

You've definitely spotted the problematic part in this comment https://github.com/GenericMappingTools/gmt-python/pull/245#pullrequestreview-179556577. I do think there needs to be a bit more discussion there on how we handle the output. I.e. do we want the gridding function to create a file only by default, or have it output an xarray.Dataset, or have some gmt-python specific flag (e.g. return_grid = True/False) that allows the user to decide.

Another question is, what's a suitable alias for -G? 'outputfile' seems a bit long, 'grdfile' or 'out_grdfile' seems quite technical/specific. My preference is for 'outfile', but I'm open to suggestions.

weiji14:
Ok, it's been almost a month already so I'm just going ahead with the changes suggested by @leouieda and me:

- The `-G` parameter now has an alias called `outfile`
- Setting this `outfile (G)` parameter means the grid won't be returned in python (i.e. `None` is returned)

Also, I do have a jupyter notebook called `process_grid_data.ipynb` showcasing this new functionality and that of `gmt.grdcontour` (See #225) but I'll put that in a separate pull request once this one is merged in!

weiji14:
In terms of working on filenames, I think a good example would be the pandas library, and in fact, I did look at their [I/O](http://pandas.pydata.org/pandas-docs/stable/io.html) implementation when wrapping `surface`. But really, this should be in a separate topic. Anyways, thanks for merging this in, I'll definitely contribute some more in the future!

leouieda:
> I think a good example would be the pandas library

Yep. And they have no function that computes things and saves them directly to a file. You always do `pd.read_csv` then compute something and `DataFrame.to_csv` to save to a file. So it feels weird doing `gmt.suface(data.longitude, data.latitude, data.bathymetry, outlfile="grid.nc")`. 

But yes, we'll discuss this more in another issue :)

> I'll definitely contribute some more in the future!

:rocket: :tada:

PR #251, opened 2018-11-29
Drop Py35, update guides, Makefile, and CI config

leouieda:
Copied over some updated guides and CI configuration from Fatiando a
Terra. Includes a maintenance guide for review PRs and making releases.
Some Makefile rules have changed to make things simpler.

Also decided to drop Python 3.5 so we don't have to test against it or
be limited by things that are 3.6+ only. Should be OK since most people
are on 2.7 or 3.6 and Anaconda now ships with 3.7 by default.

Simplified the README as well because PyPI will fail to render the `raw`
rst directive.

Fixes some lint errors because of an update in pylint.


seisman:
@leouieda Do you think it helps to give each job a name to make the travis build matrix more readable? See https://github.com/travis-ci/travis-ci/issues/5898 for discussions and https://docs.travis-ci.com/user/customizing-the-build/#naming-jobs-within-matrices.

leouieda:
@seisman that does look a lot nicer: https://travis-ci.org/GenericMappingTools/gmt-python/builds/461100703?utm_source=github_status&utm_medium=notification

Merging this in.

PR #257, opened 2018-12-24
Remove gmt from requirements.txt

seisman:

Strictly speaking, gmt-python doesn't dependent on conda-forge's gmt, thus I propose to remove the gmt dependency from requirements.txt. 

Also, such change makes it easy to test gmt-python with latest gmt master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,8 @@ env:
         # The file with the listed requirements to be installed by conda
         - CONDA_REQUIREMENTS=requirements.txt
         - CONDA_REQUIREMENTS_DEV=requirements-dev.txt
-        # Get GMT6 from the development channel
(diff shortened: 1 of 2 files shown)


leouieda:
@seisman I've been thinking about this a lot lately. Our current setup is very fragile. It's actually kind of broken with the hang ups in the conda packages (I still haven't been able to solve conda-forge/gmt-feedstock#57). 

I'll try to get this working on Travis with the conda packages. We'll just have to instruct people to compile GMT themselves. I'll include a build script for GMT that fetches the latest master and add it to the Makefile for easier updating and deploying. 

I'm a bit hesitant because GMT master is not very stable but it might be better than what we currently have.

leouieda:
:+1: thanks!

PR #261, opened 2019-01-16
Build GMT master instead of using conda-forge

seisman:
The conda-forge development package creates all sorts of problems for us with 
mixtures of different channels, etc. It's a hassle to setup and we end up falling behind
new features added to GMT 6 (which is what we support).

Replace the conda-forge package with GMT built from the master branch on the CI
services. Update the install instructions to tell users to build GMT from source.


.travis.yml
```diff
@@ -13,6 +13,7 @@ branches:
         - master
         # Regex to build tagged commits with version numbers
         - /\d+\.\d+(\.\d+)?(\S*)?$/
+        - /test-*/
```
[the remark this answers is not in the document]
leouieda: Why is it there? Travis should build this branch when the PR is opened so we don't really need it.

ci/travis-build-gmt.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# To return a failure if any commands inside fail
+set -e
+
+export INSTALLDIR="$HOME/gmt-install-dir";
+export COASTLINEDIR="$HOME/gmt-install-dir/coast";
+
+git clone --depth=1 https://github.com/GenericMappingTools/gmt "${HOME}/gmt-master"
+cd "${HOME}/gmt-master"
+bash ci/travis-setup.sh
```
leouieda: @seisman this might be a problem because it uses the Ubuntu packages to compile GMT. So it won't work on Mac. Ideally, we should try to use the conda installed packages but this has causes a lot of trouble for me in the past.

seisman: @leouieda Currently, the master branch is only built on Linux with Python 3.6 installed. If you really want to test the GMT master branch on Mac, we should use homebrew.

seisman:
@leouieda Three of the six build jobs fail. The GMT master branch fails because of some incompatible changes of psbasemap in the latest GMT. We can open a new PR to fix these tests failures.

ci/travis-build-gmt.sh
leouieda: OK, do you have any idea how that works? If we can get that setup, I would be willing to ditch the conda-force based jobs in favor of this. It's a lot easier to control and would actually help test GMT itself.

leouieda: I'll try to get this setup in a separate PR

seisman: @leouieda See this PR https://github.com/GenericMappingTools/gmt/pull/146. 

The only problem is it's too slow, (3 minutes for Linux builds, 13 minutes for Mac builds). 
That's why this PR is not merged into GMT master branch.

If you think it's acceptable, I can add it to gmt-python.

leouieda:
Perhaps we should outsource this to a `continuous-integration` repo on this org so we can share the same scripts as base GMT?

seisman:
Since we always need to clone the main GMT repo to build the latest GMT, it makes more sense to put these scripts in the main GMT repo.

leouieda:
> Since we always need to clone the main GMT repo to build the latest GMT, it makes more sense to put these scripts in the main GMT repo.

Right, I forgot about this. That makes sense.

helpers/build-gmt-master.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Download and build the GMT master branch. Used by TravisCI.
+
+set -e
+
+mkdir -p "$INSTALLDIR"
+mkdir -p "$COASTLINEDIR"
+git clone --depth=1 --branch=master https://github.com/GenericMappingTools/gmt.git tmp
```
seisman: Why not clone it to gmt-master directly? Any side-effects?

leouieda: It's because of the cache. I'm having Travis cache `gmt-master/build` so we don't recompile things that didn't change. But git won't checkout into an existing directory.

seisman: Got it.

leouieda:
@seisman I'm wondering if we should stop relying on the conda-forge development builds and tell people to build GMT from source. This is package is still under development and doesn't have to be trivial to use at the moment. When GMT 6 is release, we can start relying on the conda-forge package to make installing easier. What do you think?

We can add some instructions to the Install page to make this easier.

leouieda:
The only thing is that we are gonna get a lot more issues because of this :( But I think it's a necessary evil.

leouieda:
OK, I'll update the install instructions and then we should be good to go. I think I'll try to maintain the conda-forge packages still but only as a last resort or for the binder demo.

leouieda:
@GenericMappingTools/python-contributors please be advised that we'll need to build GMT from source from now on. Don't hesitate to get in touch if you need help getting that to work. We'll work on including some instructions for this. A good starting point is this script https://github.com/GenericMappingTools/gmt/blob/master/ci/build-gmt.sh

leouieda:
Merging.

PR #268, opened 2019-01-22
Use sphinx-gallery to manage examples and tutorials

leouieda:

Enables the usage of sphinx-gallery with PyGMT. For now, requires features that aren't yet released so we need to install from the Github master branch of sphinx-gallery. Implements an image scrapper for `pygmt.Figure` that captures them and saves to the correct file names. Need to add a global figure registry (`SHOWED_FIGURES` in `pygmt/figure.py`) so that we can request all figure objects that have called their `show` method. Replace the current jupyter notebooks by gallery notebook style `.py` files, which are easier to maintain on version control. 

TODO:

- [x] Make a notebook example to check that it works
- [x] Add instructions for adding examples in CONTRIBUTING and ask people to submit them


doc/conf.py
```diff
@@ -17,11 +22,9 @@
     "sphinx.ext.intersphinx",
     "numpydoc",
     "nbsphinx",
```
seisman: @leouieda I tried it locally. It's really nice!!!
Don't forget to remove nbsphinx, since we no longer use notebooks after this PR is merged.

seisman: Forget it. I just saw your comments in #269.

leouieda: Thanks! I quite like it too.

I'm leaving it because I'll want to have one notebook in there as a "Using PyGMT in Jupyter" tutorial (#269).

PR #274, opened 2019-01-23
Use uuid.uuid4 to generate unique names

### LINKED ISSUE #273: Use the uuid package to generate figure names

GMT figure names need to be unique and are generated by the `unique_name` function in `pygmt/helpers/tempfile.py`. Currently, it uses `tmpfile` to generate a unique name but this isn't ideal and kind of a hack. A better way would be to use the builtin [`uuid.uuid4`](https://docs.python.org/3.6/library/uuid.html#uuid.uuid4) function.

seisman:

Use `uuid.uuid4().hex` to generate unique names.

Fixes #273 

**Reminders**


pygmt/helpers/tempfile.py
```diff
@@ -17,12 +18,10 @@ def unique_name():
...
     -------
     name : str
-        A unique name generated by ``tempfile.NamedTemporaryFile``
+        A unique name generated by ``uuid.uuid4``
 
     """
-    # Use the tempfile module to generate a unique name.
-    with NamedTemporaryFile(prefix="gmt-python-") as tmpfile:
-        return os.path.split(tmpfile.name)[-1]
+    return "gmt-python-" + uuid.uuid4().hex
```
seisman: I'm not sure how to deal with the prefix `gmt-python`. Possible ways:
1. No prefix at all
2. Change `gmt-python-` to `pygmt-`
3. Add arguments `prefix` and `suffix` to `unique_name`.

pygmt/helpers/tempfile.py
```diff
@@ -17,12 +18,10 @@ def unique_name():
     Returns
     -------
     name : str
-        A unique name generated by ``tempfile.NamedTemporaryFile``
+        A unique name generated by ``uuid.uuid4``
```
leouieda:
```suggestion
        A unique name generated by :func:`uuid.uuid4`
```

leouieda: We can use the intersphinx mappings to link directly to the python docs.

pygmt/helpers/tempfile.py
leouieda: I think we can remove that prefix. Not sure why I added it but it really doesn't make a difference.

leouieda:
The failure is because of #276. Will rerun the checks after that is fixed.

PR #281, opened 2019-01-25
Remove unused images

seisman:
Two unused images are removed.


leouieda:
@seisman thanks for noticing. I'm looking into using those still so better leave them in to avoid cluttering the history.

PR #287, opened 2019-01-31
Update travis to make latest GMT build script working

seisman:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,10 @@ addons:
         packages:
             - cmake
             - build-essential
+            - ninja-build
             - libcurl4-gnutls-dev
             - libnetcdf-dev
-            - libgdal1-dev
+            - libgdal-dev
             - libfftw3-dev
             - libpcre3-dev
             - liblapack-dev
@@ -38,7 +39,7 @@ addons:
     homebrew:
         packages:
             - cmake
-            - curl
+            - ninja
             - netcdf
             - gdal
             - fftw


.travis.yml
```diff
@@ -27,6 +27,7 @@ addons:
         packages:
             - cmake
             - build-essential
+            - ninja
```
weiji14: I think you might want to use [ninja-build](https://packages.ubuntu.com/trusty/ninja-build) instead of [ninja](https://packages.ubuntu.com/trusty/ninja) here. At least according to https://github.com/GenericMappingTools/gmt/pull/353/files#diff-354f30a63fb0907d4ad57269548329e3R102

seisman:
> Note that [libgdal1-dev](https://packages.ubuntu.com/trusty/libgdal1-dev) is a transitional package. For people using Ubuntu versions bionic and above, it might be better to use [libgdal-dev](libgdal-dev) instead (and also make the change upstream into GMT)?

Yeah, you're right. libgdal1-dev is a transitional dummy package that depends on libgdal-dev. We should install libgdal-dev instead.

.travis.yml
weiji14: Oops, should have been clearer. Use `ninja-build` just for linux, and keep using `ninja` for homebrew.

seisman: OK. I've made the changes. I never used ninja before.

weiji14:
So it's giving the [error](
https://travis-ci.org/GenericMappingTools/pygmt/jobs/486733383#L720) `The command "if [ "$TEST" == "true" ]; then bash helpers/build-gmt-master.sh; fi" failed and exited with 1 during .`.

I think we need to set the environment variable `TEST` to false, See https://github.com/GenericMappingTools/gmt/pull/353/files#diff-354f30a63fb0907d4ad57269548329e3R11. I know we set it to False on [this line](https://github.com/GenericMappingTools/pygmt/blob/79c879e426103caf4090318df55aaeffc651527e/helpers/build-gmt-master.sh#L14) but it doesn't seem to propagate all the way through?

.travis.yml
weiji14: Neither have I, not sure what it does :rofl:

seisman:
@weiji14 You're viewing an older build. The [latest build](https://travis-ci.org/GenericMappingTools/pygmt/builds/489850191) passes on Linux and is waiting for the macOS.

weiji14:
Oh yes, I'm getting confused with the pygmt and upstream gmt scripts... Let's see if the macOS builds run ok.

seisman:
The macOS builds (https://travis-ci.org/GenericMappingTools/pygmt/jobs/489850197) and (https://travis-ci.org/GenericMappingTools/pygmt/jobs/489850198) fail. Seems there're something wrong with miniconda.

seisman:
@weiji14 @leouieda The latest builds passed.

seisman:
@weiji14 I just slept for one night and woke up to find that it was fixed. 😄

PR #289, opened 2019-02-28
Add a new method to shift plot origins

### LINKED ISSUE #267: Add a new method to Figure to shift the plot

Unlike tools like matplotlib, GMT plots things to a canvas/page using physical coordinates. You can control where the plot is placed on this canvas with the `X` and `Y` arguments. Traditionally, if you want to shift the origin to another location, you'd call `gmt plot -T X=... Y=...`. The `-T` served tell `plot` to not actually plot anything. This is a bit unnatural and we should probably not allow the `-T` option. Instead, we should have a method like `Figure.shift_origin` that does this by calling plot. See #253  

seisman:

Add `Figure.shift_origin()` to shift plot origins.

Fixes #267.

**Reminders**

- [X] Add tests for new features or tests that would have caught the bug that you're fixing.
- [X] Add new public functions/methods/classes to `doc/api/index.rst`.
- [X] Write detailed docstrings for all functions/methods.


seisman:
@leouieda Can you help review it?

pygmt/figure.py
```diff
@@ -284,6 +284,27 @@ def show(self, dpi=300, width=500, method="static"):
             img = Image(data=png, width=width)
         return img
 
+    def shift_origin(self, xshift=None, yshift=None):  # pylint: disable=no-self-use
```
leouieda:
```suggestion
    def shift_origin(self, xshift=None, yshift=None): 
```

@seisman you still need to call `self._preprocess()` to make sure we're plotting to the right figure. This is not a great design, granted. But the only other I can think of for calling the `figure` module automatically is through a metaclass, which I'm not eager to do.

pygmt/figure.py
```diff
@@ -284,6 +284,27 @@ def show(self, dpi=300, width=500, method="static"):
...
+        """
+        Shift origins in x and/or y directions.
```
leouieda: Some more information here would be great.

pygmt/figure.py
```diff
@@ -284,6 +284,38 @@ def show(self, dpi=300, width=500, method="static"):
...
+        This method shifts plot origin relative to the current origin by (*xshift*,*yshift*)
+        and optionally append the length unit (**c**, **i**, or **p**).
+
+        Prepend **a** to shift the origin back to the original position
+        after plotting, prepend **c** to center the plot on the center of the
+        paper (optionally add shift), prepend **f** to shift the origin relative
+        to the fixed lower left corner of the page, or prepend **r** [Default] to
+        move the origin relative to its current location.
+
+        Detailed usage at http://gmt.soest.hawaii.edu/doc/latest/GMT_Docs.html#plot-positioning-and-layout-the-x-y-options
```
seisman: What's the best way to reference the official GMT documentation?

seisman:
> Could you also add a gallery plot that uses this method? Maybe plotting an inset or something.

I have no idea what to add in the gallery. We should use the `inset` module for inset plots and `subplot` for multi-panel plots (both are not implemented yet). I think `Figure.shift_origin()` should only be used for multi-panel plots with some irregular layout.

pygmt/figure.py
leouieda: I still don't know of a good way to do this yet. I'm open to suggestions. It would be best to have this as a variable somewhere and input into the docstring somehow.

leouieda:
@seisman you're right about the inset. The usage in the tutorial is good as that is the intended use of the method. Thanks!

pygmt/figure.py
seisman: @leouieda Can we rename the `gmt_module_docs` decorator to something like `gmt_docs`, which supports following conversions:
```
{gmt_docs basemap} => http://gmt.soest.hawaii.edu/doc/latest/basemap.html
{gmt_docs gmt xy-full} => http://gmt.soest.hawaii.edu/doc/latest/gmt.html#xy-full
```

leouieda: I would actually like to get rid of most of those decorators in favor of something simpler to maintain. But I don't think there is a good way to format the docstring without using a decorator. We could certainly add that functionality. But we should be using the https://www.generic-mapping-tools.org/gmt/latest/ domain now.

seisman: What about putting the raw links in the docstrings, without using a decorator or any substitutions? If someday the GMT documentation site changes, it's still easy to update all the links.

PR #293, opened 2019-03-13
Add pygmt.config() to change gmt defaults locally and globally

### LINKED ISSUE #288: Support changing gmt defaults

Changing gmt defaults in GMT CLI:
1. `gmt set PAR1=value1 PAR2=value2`
2. `gmt plot ... --PAR1=value1 --PAR2=value2`

In pygmt, these two should be:

1. `gmt.set(PAR1=value1, PAR2=value2)`
2. `gmt.plot(projection="xxx", region="xxx", PAR1=value1, PAR2=value2)`

Another way is using dict, e.g. `gmt.conf['PAR1']=value1`, which seems more pythonic. But we need call `gmt set` for each assignment. I'm not sure if it's technically possible.

seisman:

Support changing gmt defaults globally and locally.

Usage:
```
pygmt.config(PARAMETER=value)  # Global

# Locally set a value that will revert back to the default
with pygmt.config(PROJ_ELLIPSOID=value):
    ...

pygmt.whatever(...)
```

Fixes #288.

**Reminders**


leouieda:
@seisman Looking at the linter output, I hadn't remembered that `set` is a builtin class in Python. Redefining it can have consequences for the other functions in the module and elsewhere. We might have to change this name to avoid unintended behavior. I'm not sure what to do because we also don't want to completely change names used in GMT. Here are a few options:

1. `Set` (since it's a class, but it looks strange)
2. `configure`
3. `set_parameter`

I'm sure what the best option would be. I'm inclined toward `pygmt.configure`. Either way, we should tell pylint to stop complaining about lower case class names.

seisman:
@leouieda What about `pygmt.conf()` or `pygmt.config()`? These two are also commonly used and are much shorter than `pygmt.configure()`.

leouieda:
Sure, `pygmt.config` is explicit enough :+1:

pygmt/tests/test_config.py
```diff
@@ -0,0 +1,34 @@
...
+            X="15c",
+        )
+
+    fig.basemap(
+        region="0/10/0/10",
+        projection="X10c/10c",
+        frame=["af", '+t"Blue Annotation"'],
+        X="15c",
+    )
+    return fig
```
weiji14: The changed global config setting doesn't end after the function has finished running (and causes other tests to have blue font...). Is there a way to reset the `FONT_ANNOT_PRIMARY` back to the default setting at the end?

pygmt/modules.py
```diff
@@ -142,3 +142,45 @@ def which(fname, **kwargs):
     if not path:
         raise FileNotFoundError("File '{}' not found.".format(fname))
     return path
+
+
+class config:
```
weiji14: Pylint doesn't like lowercase names for classes...

pygmt/tests/test_config.py
seisman: I added `config(FONT_ANNOT_PRIMARY="black")` at the end of the test to revert the changes.

tjcrone:
I've been testing pygmt.config() out locally and it appears to work. Does anyone know the status of this merge and what is left to do here? I would be happy to make a PR if there are specific issues that need to be addressed.

Without this merge, I was using gmt.conf for various settings, which does work, but as far as I can tell there is no way to get pygmt to reload gmt.conf without restarting the kernel and importing pygmt. Is there a way to get pygmt to reload the gmt.conf file without restarting the kernel? Thanks!

weiji14:
Hi @tjcrone, functionality wise, this PR is basically done. We probably just need a minor fix that renames the class `config` to `Config` to satisfy the linter (and also change the tests to use capitalized `Config`). It would be nice to use lower case like in [chainer here](https://github.com/chainer/chainer/blob/6cdf0c436305135db221c17689192e8da5bac509/chainer/configuration.py#L158) but I guess we could always wrap around that main class in a separate PR.

Other than that, it would be nice to have some sort of gallery example or tutorial to highlight the use of `pygmt.Config`. Would you like to contribute an example since you have a bit more of a grasp on this?

tjcrone:
@weiji14, I would be happy to add an example, which might go best in the [user guide](https://www.pygmt.org/dev/tutorials/frames.html). It would be nice to see the docs rendered after this is merged so I can see what is needed. If this is merged will the docs render?

weiji14:
@tjcrone, you can preview the docs by clicking on the "View deployment" button in any of our PRs (see https://github.com/GenericMappingTools/pygmt/pull/344#issuecomment-541274587). Instructions for making a tutorial are at https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#tutorials, and you follow the instructions at https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#documentation to build the documentation locally before you push up to Github. I'll merge this in first so that `pip install https://github.com/GenericMappingTools/pygmt/archive/master.zip` will include `pygmt.config` and you can create a new branch to work on the tutorial.

PR #296, opened 2019-03-20
Pin version of CI scripts to 1.0.0

leouieda:
Don't clone from master to avoid breaking the builds when the scripts
update.

**Reminders**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,7 +118,7 @@ before_install:
       fi
     - cd "$TRAVIS_BUILD_DIR"
     # Get the Fatiando CI scripts
-    - git clone https://github.com/fatiando/continuous-integration.git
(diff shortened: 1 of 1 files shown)


leouieda:
@seisman could you configure appveyor to only build branches with a `.appveyor.yml`? I don't have access to the admin panel since you set it up.

seisman:
@leouieda I've removed pygmt from appveyor, since currently we don't have an .appveyor.yml file.

leouieda:
Thanks, @seisman. I'm trying out Azure Pipelines for Windows. Seems to be better and faster.

seisman:
@leouieda I'm also trying to build gmt for windows using Azure Pipelines. It takes only 10 minutes for one build, much faster than appveyor.

leouieda:
:+1: yep. It's kind of complicated and I hate their sintax but it's a lot faster.

PR #301, opened 2019-04-09
pandas.read_table is deprecated, use pandas.read_csv instead

seisman:
`read_table` is deprecated since pandas 0.24. Use `read_csv` instead.

Fixes the FutureWarning:
```
FutureWarning: read_table is deprecated, use read_csv instead.
  data = pd.read_table(fname, header=1, sep=r"\s+")
```


leouieda:
:+1: thanks!

PR #312, opened 2019-06-20
Enable Mac and Windows CI on Azure Pipelines

leouieda:

Delegate the style checks and Mac builds to Azure instead of Travis (for speed).
Deployment still happens on Travis.

**Reminders**

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,214 @@
+# Configuration for Azure Pipelines
+########################################################################################
+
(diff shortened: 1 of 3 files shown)


leouieda:
The library is not being found on Windows because the DLL from conda is `gmt.lib` and not `gmt_w64.lib`. We should implement a series of trials for loading the library before giving up. I'll leave that for another PR and merge in with Windows failing now since we don't officially support it yet.

PR #285, opened 2019-01-25
WIP: Let get_clib_path find the path to the gmt shared library on Windows

seisman:
**Description of proposed changes**
The name of the shared library is `gmt_w64.dll` for 64-bit Windows, and `gmt_w32.dll` for 32-bit Windows, and the library is located in `C:\programs\gmt6\bin` by default. This PR changes `get_clib_path` so that it can return the correct library name for Windows.

I've tried to install pygmt on Windows and run the test. The GMT6 binary is the one provided by Joaquim. 

The tests give 19 failures. Most of the failures are due to different image size, which may be caused by slightly different ghostscript (the version I use is 9.22). Other failures seems related to the slightly different behaviors of Python packages for different system, and permission issues. See the attached log for details. [pytest.log](https://github.com/GenericMappingTools/pygmt/files/2797717/pytest.log) 

Although currently we don't have CI tests on Windows, I think we can at least merge this PR into master, so that Windows users can have a try and help us find bugs on Windows.

Fixes #46 

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

Issue #211, opened 2018-07-12
ctypes and cffi?


paugier:
It is just a small and fast question. I see in the documentation: "Interface with the GMT C API directly using ctypes (no system calls)."

I understood that now, cffi was just better than ctypes for new projects (see for example https://www.quora.com/How-can-I-decide-between-using-ctypes-and-CFFI-in-Python).

I was wondering if there is a reason why gmt-python doesn't use [cffi](https://cffi.readthedocs.io) instead of ctypes?

leouieda:
Hi @paugier, thanks for the link! I hadn't seen that post.

Basically, I didn't really know about CFFI until recently. I've been having a good experience with ctypes and [numpy arrays have builtin support for it](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.ctypes.html). Right now, I don't know if it's worth the trouble of switching if ctypes is working. But I'm open to suggestions if there is anything specific that CFFI can address. 

Thanks for taking the time to give feedback! I really appreciate it.

seisman:
Can we close this 2.5-year-old issue? I don't think we will switch to cffi in the next few years.

weiji14:
Yep, close it as wontfix. But feel free to reopen if someone wants to work on it :slightly_smiling_face:

seisman:
Just tried cffi. The following example shows how to call the API function `GMT_Get_Enum`:
```python
>>> from cffi import FFI
>>> ffi = FFI()
>>> ffi.cdef("int GMT_Get_Enum (void *V_API, char *key);")
>>> lib = ffi.dlopen("libgmt.dylib")
>>> lib.GMT_Get_Enum(ffi.NULL, b"GMT_VF_LEN")
32
```
For comparison, the ctypes version is 
https://github.com/GenericMappingTools/pygmt/blob/50145914b5f42710f8f0ac3a7a676195d8a054c4/pygmt/clib/session.py#L263-L280

PR #313, opened 2019-06-20
Fix issues for running on Windows

### LINKED ISSUE #46: Support for Windows
You have mentioned that this has been tested in Unix and OSX. I tried intalling in Windows through cloning the repo. However when I import gmt in Jupyter, it gave me the following error.

GMTOSError: Unknown operating system: win32

Any solution for a Windows support?

leouieda:

Trying to fix the issues for running on Windows now that we have CI.

Fixes #46 

**Reminders**


pygmt/clib/loading.py
```diff
@@ -37,85 +37,46 @@ def load_libgmt(env=None):
...
-    else:
-        libpath = libname
-    return libpath
+    libnames = clib_name(os_name=sys.platform)
+    libpath = env.get("GMT_LIBRARY_PATH", "")
+    error = False
+    for libname in libnames:
+        try:
+            libgmt = ctypes.CDLL(os.path.join(libpath, libname))
+            check_libgmt(libgmt)
```
seisman: If both ctypes.CDLL and check_libgmt don't raise any errors, it means we find the library, then we should break out of the loop. Right?
```suggestion
            check_libgmt(libgmt)
            break
```

leouieda: Of course! Thanks, Dongdong. I stopped coding yesterday because I couldn't even see this basic error anymore :)

leouieda:
I managed to track down the failures to GMT API session creation causing a crash. It might be something that Windows is a bit more sensitive to. Need to investigate.

seisman:
Just provide more information:

I tried pygmt with the latest master branch of GMT. It works well. Running `pygmt.test()` results in some failed tests, same as what I saw several months ago in #285.

I also tried GMT 6.0.0rc1 provided by conda-forge. GMT itself works well on Windows, but `import pygmt` crashes.

leouieda:
Yeah, I'm still scratching my head over this one. I wondered if it's a compiler flag on the conda recipe that is messing things up.

leouieda:
I just noticed that the conda-forge build log has `Build GMT for developers   : yes` in the cmake log. @seisman do you know if this means we're making debug build by mistake?

seisman:
No. It only means installing the header files (`*.h`) for developers.

leouieda:
Yeah I figured we wouldn't be this lucky. Thanks

leouieda:
Oh good. Now with rc2 the Linux build is failing.

weiji14:
It's crashing on this line:

https://github.com/GenericMappingTools/pygmt/blob/da2cfec3f64fe02f859f1915e0d87ca1a915ec51/pygmt/clib/session.py#L348

seisman:
Yes, but as I said in previous comments, pygmt works with self-built GMT on Windows, but fails with the GMT provided by conda-forge. So it's really difficult to track down the crash.

weiji14:
@StevePem will be working on this with me today!

weiji14:
We seriously need to work on this. I've had 3 Windows users today that couldn't even start contributing properly because of the infamous "gmt_w32.dll" library could be found error. We've tried to set the environment variable using `set GMT_LIBRARY_PATH=C:\path\to\gmt.dll` but that still doesn't work, and we even dove into the code at `loading.py`, ran it line by line, but that still failed.

I saw a note at https://github.com/GenericMappingTools/pygmt/pull/312#issuecomment-503831325 that 'gmt.lib' might be used. What is the difference with that and "gmt.dll"?

pygmt/clib/loading.py
weiji14: I just realized that this change is what's been causing our Windows builds to fail... Since we have a list of dll files to check, if one of the dll fails, the variable `error` turns `True` and stays `True` even if a later dll works. Our Linux and MacOS tests work because there's only one item in the list.

leouieda:
@weiji14 yes, this needs a lot of attention. It's been really hard to debug without a working Windows computer. I have recently gotten one so I can try running things locally and track down the problem. It seems to only happen with the conda packages.

leouieda:
@StevePem all the help we can is very very welcome! I'm probably late for this but thank you for taking the time to work on this!

weiji14:
Good luck @leouieda! I've tried a whole day recently to fix this but it's definitely not a trivial matter as it requires a solid knowledge on GMT's C API. The sudden crash with no error message definitely doesn't help, and I even considered switching from `ctypes` to `cffi` as mentioned in https://github.com/GenericMappingTools/pygmt/issues/211#issue-340538174 but it's getting past the limits on my C knowledge...

leouieda:
@weiji14 :cry: yeah, this has bugged me quite a bit. I don't think it's a ctypes issue. I suspect it might be something to do with our conda-forge build. I'm going to work hard on this and some other major issues this week. Hopefully we can get the major blocking points for 0.1 done before AGU :crossed_fingers:

michaelgrund:
Hey guys,
after some time, I finally got pygmt properly working on my Windows 10 notebook using Anaconda. However, for this purpose I made use of the Linux subsystem available from the latest Win10 versions. I followed the instructions given here [https://gist.github.com/kauffmanes/5e74916617f9993bc3479f401dfec7da] to install Anaconda first. Afterwards it was quite easy to install pygmt as outlined in the installing section using the bash available from the Linux subsystem. Putting  **GMT_LIBRARY_PATH=$HOME/anaconda3/envs/pygmt/lib**  into .bashrc also works quite well without any errors and warnings. Finally, I successfully tested some of my pygmt jupyter notebooks developed on my "real" Linux machine. I know this is no perfect solution, but probably it may serve as workaround for some of the windows users.

leouieda:
Hi @michaelgrund thanks for sharing! Yeah, we might just have to stick with that for now. It's not the ideal solution since it wouldn't integrate with a user's existing setup.

seisman:
The issue was fixed in #434.

PR #314, opened 2019-06-20
Add Google Analytics to measure site visitors

leouieda:
Add the tracking script to the documentation template so we can know how
many people are visiting the docs and which pages they see the most.
The data are anonymous and we do not store cookies to track users in
other sites. This limits the amount of information we can get but is
much more ethical and less creepy.

**Reminders**


leouieda:
@GenericMappingTools/python-contributors does anyone have a good argument against this? I wasn't keen on enabling this before I figured out how to make it so we don't track users or store cookies.

weiji14:
I don't really have a solid argument against them (as long as there won't be any annoying cookie popups while I'm in Europe). I do think some transparency on how this is going to work is good though, so a few questions:

- What analytics exactly are being measured (e.g. visit counts, 'masked' IP addresses, etc?). Maybe provide a link on how the google-analytics.js script works.
- Who will be able to access the analytics? Will it be shared with the whole team or just the core maintainers, and how regular will the data be reported so that we can analyze it?

leouieda:
> as long as there won't be any annoying cookie popups while I'm in Europe

That is one of the reasons I disabled cookies. It also makes analytics less creepy because it can't follow you across different websites. 

> I do think some transparency on how this is going to work is good though

I agree. We should include a disclaimer about this in the footer of the documentation at least.

> What analytics exactly are being measured

Mostly page view counts and how long people spend on each. I think we can get some data on whether a person visited multiple pages but not their IP address or any identifying information. 

>  Maybe provide a link on how the google-analytics.js script works. 

There is a lot of information here: https://developers.google.com/analytics/devguides/collection/analyticsjs/ It's a bit hard to read, though. I got this idea from how Jupyter uses it on their website. 

> Who will be able to access the analytics?

Right now it's just me but I can share access with core developers who want it (anyone in @GenericMappingTools/python). 

> how regular will the data be reported so that we can analyze it?

I didn't think of any sort of formal reporting. This kind of metric is interesting to look at to see which pages people visit the most and which ones don't get seen at all. We can act on this info by better promoting pages which we want people to read but that don't get much love for some reason. It's also a good metric to have in case we want to apply for funding or argue that the project is used and has X numbers of users. Download counts are skewed because of CI and bots so page views of the docs gives something to compare that to.

leouieda:
@weiji14 I added a disclaimer to the footer. What do you think?

weiji14:
> I agree. We should include a disclaimer about this in the footer of the documentation at least.

The disclaimer reads ok, pretty standard I guess. Is there an easy way to preview the new HTML page or do I need to clone this branch and build it?

> > Maybe provide a link on how the google-analytics.js script works.
> 
> There is a lot of information here: https://developers.google.com/analytics/devguides/collection/analyticsjs/ It's a bit hard to read, though. I got this idea from how Jupyter uses it on their website.

Cool. The reason I asked is more from a maintenance point of view (if the javascript code needs to be updated in the future). The details of how it works is still a bit beyond me :laughing: but someone else might get something out of it.

> I didn't think of any sort of formal reporting. This kind of metric is interesting to look at to see which pages people visit the most and which ones don't get seen at all. We can act on this info by better promoting pages which we want people to read but that don't get much love for some reason. It's also a good metric to have in case we want to apply for funding or argue that the project is used and has X numbers of users. Download counts are skewed because of CI and bots so page views of the docs gives something to compare that to.

Yes I was thinking of the funding thing too! I guess we'll just have to let it collect some statistics (say for a month) and see what it says. Github repos also have a traffic page https://help.github.com/en/articles/viewing-traffic-to-a-repository which might complement the webpage's analytics.

leouieda:
@weiji14 I added the link to their documentation as comment on the template. There is no easy way to share the html. I've been looking for a way to build and share all PR docs but haven't found anything that works widely. For now, here is a screenshot:

footer-disclaimer

seisman:
Merged.

Issue #319, opened 2019-08-08
When will command “text” be added?


chdwql:
**We need to mark the text on the figure**

**Are you willing to help implement and maintain this feature?** No

weiji14:
Hi/你好 @chdwql, unfortunately this is not implemented yet, though it has been mentioned before in https://github.com/GenericMappingTools/pygmt/issues/122 back in Jan 2018.

@leouieda, you mentioned some 'virtual file machinery' changes in https://github.com/GenericMappingTools/pygmt/issues/122#issuecomment-358436386 that was needed for `pstext`/`text` to be wrapped, I'm assuming that has been implemented already?

I've got some free time this week (university servers getting migrated) and might be able to put in a pull request for this :wave:.

weiji14:
@chdwql, we've just implemented `text` in PyGMT at #321. You should be able to use [install PyGMT from Github](https://www.pygmt.org/dev/install.html#installing-pygmt) and use it now :tada:. Let us know if you need any help using it.

PR #321, opened 2019-08-20
Wrap text

### LINKED ISSUE #122: Wrapper for pstext
Hello, I have noticed there is no support for the common used command pstext. I have tried hours for supporting the command in my own machine. However, there seems to be some problems handling input like:

when I want to realize:
```
gmt pstext -J -R -F+f12p,1,black+jTL -D-1.5c/-0.15c -K -O >> $PS << EOF
-77.15 38.89 Washington
76.22 -69.22 Zhongshanzhan
EOF
```
a common choice in the python module is:
```
gmt.pstext(<some parameters>)
```
so how to input?
```
<< EOF
-77.15 38.89 Washington
76.22 -69.22 Zhongshanzhan
EOF
```
First I have tried use the virtual file you have used. However, it seems the virtual file only supports numbers  but not strings like "Washington"..
Then, since we couldn't use the virtual database supported by GMT, then the only choice might be the python own virtual file, but it seems difficult to combine it with the console of libgmt....
Then we might need create a file in some place like /tmp or some place else to create a real file? or some realization similar to the GMT database..
I think this feather is really important and what's your advise?

weiji14:

Wrapping the [`text`](https://docs.generic-mapping-tools.org/latest/text.html) mapping function! The way I've implemented is that a user can either add a single line of text easily like so:

```python
fig.text(x=1.2, y=2.4, text="This is a line of text")
```

PyGMT text test single-line plot

or multiple lines like so:

```python
fig.text(x=[1.2, 1.6], y=[0.6, 0.3], text=["This is a line of text", "This is another line of text"])
```

PyGMT text test multi-line plot

And of course, advanced users can still pass in `textfiles` instead of using Python/NumPy objects.

TODO:

- [x] Minimal working implementation (93807e58b7df1afa3ca299348faf5b7c52dc60b7)
- [x] Handle/Add "font, angle, justify" arguments/aliases (2501988b1b14de74cc240a3304c306398c37b7ad, ff74e6dc1ba4cad1db756e1591683ad9885729b0)
- [x] Enable multiple textfile inputs (7a08ebd7a5d1b185f31b84265617d1816d2d8610)
- Others?

Fixes #122, #319 

**Reminders**


leouieda:
@weiji14 this looks good! I would leave multiple text files for a later PR if needed. I've been thinking if it even makes sense to be allowing filename inputs to all these methods in the first place (but that is for another discussion).

pygmt/base_plotting.py
```diff
@@ -530,3 +535,105 @@ def image(self, imagefile, **kwargs):
...
+            if "F" not in kwargs.keys():
+                kwargs.update({"F": ""})
+            if angle is not None and isinstance(angle, (int, float)):
+                kwargs["F"] += f"+a{str(angle)}"
+            if font is not None and isinstance(font, str):
+                kwargs["F"] += f"+f{font}"
+            if justify is not None and isinstance(justify, str):
+                kwargs["F"] += f"+j{justify}"
+
+        with GMTTempFile(suffix=".txt") as tmpfile:
```
leouieda: We should look into passing text though virtual files, actually. I was avoiding this a while ago because the API had changed but that has been standardized now in GMT.

weiji14: Sorry, I don't quite get your intention here? Do you mean passing `text` (as in the command) through virtual files or 'text' (as in plain old text)? Either way I'm confused...

weiji14:
> @weiji14 this looks good! I would leave multiple text files for a later PR if needed.

This shouldn't be too hard now that #322 is merged in. Should be a matter of setting it in the function decorator (i.e. `@kwargs_to_string(textfiles="sequence_space")`) and adding more tests. Just need to find time and set some of my priorities straight right now...

> I've been thinking if it even makes sense to be allowing filename inputs to all these methods in the first place (but that is for another discussion).

Yes we should (allow filename inputs), for at least 2 reasons. 1) `pygmt`'s virtualfile mechanism won't work sometimes with big files, whereas passing in the filename itself works. 2) [Data provenance](https://en.wikipedia.org/wiki/Data_lineage) - for outputs to gridfiles, the command ran is stored as metadata in the NetCDF file and using virtualfiles means we `grdinfo` returns something like `Command: surface @GMTAPI@-000000 -G/tmp/pygmt-81pytnbc.nc ...` instead of `Command: surface @input.txt -Goutput.nc`. But again, this is outside of scope for this PR.

liamtoney:
@weiji14 what do you think about adding this to the 0.1.0 milestone? It seems pretty integral.

weiji14:
Sure, why don't you review it and see if there's any issues :smile: I'll update the branch now.

pygmt/base_plotting.py
```diff
@@ -574,3 +580,107 @@ def legend(self, spec=None, **kwargs):
...
+        Full option list at :gmt-docs:`text.html`
+
+        {aliases}
+
+        Parameters
+        ----------
+        textfiles : str or list
+            A text data file name, or a list of filenames containing 1 or more records
+            with (x, y[, font, angle, justify], text).
+        x, y : float or 1d arrays
```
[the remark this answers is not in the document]
weiji14: Yep, it's `np.atleast_1d()` that does the trick. You're right that we could update `plot` to be consistent, we can change that later.

weiji14:
Cool, thanks for reviewing @liamtoney!

PR #325, opened 2019-09-07
Add sequence_space converter in kwargs_to_string

### LINKED ISSUE #322: How to handle one or more file inputs

There are a lot of GMT processing scripts that take in **one or more** table/grid inputs, and I was wondering how this should be handled. This came up as I was trying to wrap [`text`](https://gmt.soest.hawaii.edu/doc/latest/text.html) at #321 and it also applies to [`grdtrack`](https://gmt.soest.hawaii.edu/doc/latest/grdtrack.html) at #308 and other functions (see e.g. this [search](https://duckduckgo.com/?q=one+or+more+site%3Ahttps%3A%2F%2Fgmt.soest.hawaii.edu%2Fdoc%2Flatest&ia=web)).

In gmt bash, it looks something like this:

```bash
gmt func data1.txt data2.txt -A -B -C
```

**Implementation ideas**

So the least amount of work would be to disallow multiple file inputs and force the user write a `for-loop` to process each file separately. This 'feature' would then need to be documented somewhere.

Another way I've thought about is to patch the [`kwargs_to_string` function in decorators.py](https://github.com/GenericMappingTools/pygmt/blob/f42da4382629015c568612f045f749fc2c6bd1a1/pygmt/helpers/decorators.py#L196-L286) to include a `'sequence_space'` convertor. This means a user can pass in a list of filenames (e.g. `["data1.txt", "data2.txt"]`) which will be converted into a space-separated string (i.e. `"data1.txt data2.txt"`) that can then be passed into the wrapper function when we do:

```python
arg_str = " ".join([fnames, build_arg_string(kwargs)])
lib.call_module("somefunction", arg_str)
```

Thoughts?

**Are you willing to help implement and maintain this feature?** Yes!

weiji14:

Introduce new 'sequence_space' converter for transforming a Python list or tuple into a space separated string. Useful in cases for e.g. when multiple file inputs can be passed into some module. 

For example:
```python
@kwargs_to_string(files="sequence_space")
def module(*args, **kwargs):
    ...

>>> module(files=["data1.txt", "data2.txt"]
{'files': 'data1.txt data2.txt'}
```

Fixes #322

**Reminders**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -286,12 +286,7 @@ def kwargs_to_strings(convert_bools=True, **conversions):
             for arg, fmt in conversions.items():
                 if arg in kwargs:
                     value = kwargs[arg]
-                    issequence = fmt in (
-                        "sequence",
-                        "sequence_comma",
-                        "sequence_plus",
-                        "sequence_space",
-                    )
+                    issequence = fmt in separators
                     if issequence and is_nonstr_iter(value):
                         kwargs[arg] = separators[fmt].join(
                             "{}".format(item) for item in value
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -210,6 +210,7 @@ def kwargs_to_strings(convert_bools=True, **conversions):
       string
     * 'sequence_comma': transforms a sequence into a ``','`` separated string
     * 'sequence_plus': transforms a sequence into a ``'+'`` separated string
+    * 'sequence_space': transforms a sequence into a ``' '`` separated string
 
     Parameters
     ----------
@@ -224,7 +225,7 @@ def kwargs_to_strings(convert_bools=True, **conversions):
     Examples
     --------
 
-    >>> @kwargs_to_strings(R='sequence', i='sequence_comma')
+    >>> @kwargs_to_strings(R='sequence', i='sequence_comma', files='sequence_space')
     ... def module(*args, **kwargs):
     ...     "A module that prints the arguments it received"
     ...     print('{', end='')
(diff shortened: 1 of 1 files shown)


pygmt/helpers/decorators.py
```diff
@@ -273,7 +286,12 @@ def new_module(*args, **kwargs):
             for arg, fmt in conversions.items():
                 if arg in kwargs:
                     value = kwargs[arg]
-                    issequence = fmt in ("sequence", "sequence_comma", "sequence_plus")
+                    issequence = fmt in (
+                        "sequence",
+                        "sequence_comma",
+                        "sequence_plus",
+                        "sequence_space",
+                    )
```
seisman: I think you can change these lines to
```python
issequence = fmt in valid_conversions
```

weiji14: Yeah, in fact I could probably refactor the code to just have a `separators` dictionary and access the `separators.keys()` instead of having valid_conversions.

seisman: I agree, but it may be better to open another PR after this one is merged.

weiji14:
While working on `grdview` in #330, I've also been thinking about how to handle the [`-Ggrd_r -Ggrd_g -Ggrd_b`](https://docs.generic-mapping-tools.org/latest/grdview.html#g) case. The way I've implemented it in https://github.com/GenericMappingTools/pygmt/pull/330/files#diff-d304eb8534d09d80d3ec5bd2ba1f26b8R330-R336 actually allows `drapegrid="red.nc -Ggreen.nc -Gblue.nc"` since I didn't wrap it in a dummy_context but that's more a hack than an elegant solution...

Might open another PR on this so that users can do `drapegrid=["red.nc", "green.nc", "blue.nc"]` and the script will parse it to `-Gred.nc -Ggreen.nc -Gblue.nc`.

seisman:
It should already works, as we already can have multiple -B options.

```python
import pygmt
fig = pygmt.Figure()
fig.basemap(region="0/10/0/10", projection="X10c", frame=["xaf", "yaf", "+tTitle"])
fig.show()
```

weiji14:
Wow, I never knew that worked! Must have missed that in one of the tutorials :scream:

pygmt/helpers/decorators.py
weiji14: Agreed. There's lots of room to refactor that function such as using f-strings since we're supporting python > 3.6 anyway.

PR #329, opened 2019-09-28
Wrap makecpt

### LINKED ISSUE #214: Wrapper for makecpt
## Python 3.6, GMT 6

* Operating system:
* Python installation: system (inside docker container):
* Version of GMT/Python: 6, 3.6
* How you installed GMT/Python: Docker image

Apparently the examples shown here: https://www.gmtpython.xyz/latest/design.html#the-gmt-python-api are not working. How do i generate a colorbar and a legend for my plot ? Is there a way to work directly with the gmt modules ?

weiji14:

Wrapping the [`makecpt`](https://docs.generic-mapping-tools.org/latest/makecpt.html) function! Tentatively implemented here under the `mathops.py` file because it is classfied under ["Mathematical operations on tables or grids"](https://docs.generic-mapping-tools.org/latest/modules.html#mathematical-operations-on-tables-or-grids) in GMT upstream, but I've placed the documentation for makecpt under ['Plotting'](https://www.pygmt.org/dev/api/index.html#plotting) for now.

Example code of how it works:

```python
import pygmt
fig = pygmt.Figure()
pygmt.makecpt(cmap="oleron", series="-4500/4500")
fig.grdimage(grid=pygmt.datasets.load_earth_relief(), projection="W0/6i")
fig.show()
```

Produces:

PyGMT makecpt to change grid colour to oleron

Cross-referencing Julia wrapper/implementation at https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.makecpt.

Parameters/Aliases to wrap:

- [ ] A = alpha/transparency
- [x] C = cmap (or color in GMT.jl)
- [ ] D
- [ ] E = data_levels
- [ ] F = force_rgb
- [x] G = truncate
- [x] H = output
- [x] I = reverse (or inverse in GMT.jl)
- [ ] M = overrule_bg
- [ ] N = no_bg/nobg
- [ ] Q = log
- [ ] S = auto
- [x] T = series (or range in GMT.jl)
- [ ] W = wrap/categorical
- [x] Z = continuous
...

Fixes #214 

**Reminders**


joa-quim:
Hi @weiji14, it would be nice if we had more coordination with the Julia wrapper. Specially in what regards the keyword names. For example, looking at your example I believe you named the option -T ``series``, while I called it [range](https://github.com/GenericMappingTools/GMT.jl/blob/master/src/makecpt.jl#L78). The julia wrapper lacks lots of documentation (it's a huge task) but the comment section of each module has the names given to all options. And I have done that to all modules (but the grd|gmt math ones). Those names can be still changed or added more aliases, so any discussion about them is completely open.

weiji14:
Funny you said that, I did think of using ‘range’ at one point (might have looked at your Julia implementation actually) but unfortunately that’s a reserved function in Python so had to come up with another name. Other than that though, I do agree that we should keep things consistent across implementations.

Might be good to track this upstream, perhaps at https://github.com/GenericMappingTools/gmt/pull/230? That way we don’t keep reinventing the wheel/aliases each time.

As a side note, if you need help with documentation, there’s this [hacktoberfest](https://hacktoberfest.digitalocean.com/details) thing coming up that would be terrific for encouraging more contributions.

joa-quim:
Ah, yes ``range`` is that keyword used often in loops. I had the same issue with the keyword ``end`` that is also reserved in Julia.

The long-format GMT issue is the right place but when we go ahead of it we must baptize the keywords anyway.

Thanks for the ``hacktoberfest`` link. It just got me in a period of short free time but I'll try to get in, even if not right from beginning of October.

weiji14:
I think this is ready for review. The `cmap (C)` and `series (T)` aliases should pretty much cover the most basic needs, and with `truncate (G)` and `reverse (I)` that should cover most use-cases (at least the ones I can think of). The `output (H)` functionality is quite different from upstream GMT. Here I've set it to take a file name, instead of have it produce an output to 'stdout' in modern mode.

Happy to consider any other suggestions too :smile:

pygmt/mathops.py
```diff
@@ -0,0 +1,59 @@
...
+    Creates a static color palette table (CPT).
+
+    Full option list at :gmt-docs:`makecpt.html`
+
+    Parameters
+    ----------
+    cmap (C) : str
+        Selects the master color palette table (CPT) to use in the interpolation.
+        Full list of built-in color palette tables can be found at
+        :gmt-docs:`GMT_Docs.html#built-in-color-palette-tables-cpt`.
```
seisman: This URL has moved to https://docs.generic-mapping-tools.org/latest/cookbook/cpts.html#built-in-color-palette-tables-cpt

pygmt/mathops.py
```diff
@@ -0,0 +1,59 @@
...
+    {aliases}
+    """
+    with Session() as lib:
+        if "H" not in kwargs.keys():  # if no output is set
+            arg_str = build_arg_string(kwargs)
+        elif "H" in kwargs.keys():  # if output is set
+            outfile = kwargs.pop("H")
+            if not outfile or not isinstance(outfile, str):
+                raise GMTInvalidInput("'output' should be a proper file name.")
+            arg_str = " ".join([build_arg_string(kwargs), f"-H > {outfile}"])
```
seisman: Other functions in PyGMT which output to a file use `->` instead of `>`. I don't know why but maybe you can try if `->` also works.
```suggestion
            arg_str = " ".join([build_arg_string(kwargs), f"-H -> {outfile}"])
```

weiji14: I'm pretty sure it'll work, and I did notice it in other parts of the code but never quite understood why `->` is used. Is it meant to be a cross-platform thing?

pygmt/mathops.py
weiji14: Ok, done.

pygmt/mathops.py
weiji14: Ok, I was wrong, `->` doesn't work. The [`test_output_to_cpt_file`](https://github.com/GenericMappingTools/pygmt/blob/c22d0657d39a6aa4f7459a59f55fb45679d81ac4/pygmt/tests/test_makecpt.py#L77-L83) function fails with this error message:

```python-traceback
    def test_makecpt_output_to_cpt_file():
        """
        Save the generated static color palette table to a .cpt file
        """
        with GMTTempFile(suffix=".cpt") as cptfile:
>           makecpt(output=cptfile.name)

../pygmt/tests/test_makecpt.py:82: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../pygmt/helpers/decorators.py:187: in new_module
../pygmt/helpers/decorators.py:282: in new_module
../pygmt/mathops.py:65: in makecpt
    lib.call_module(module="makecpt", args=arg_str)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pygmt.clib.session.Session object at 0x7fe9ae6ad470>, module = 'makecpt', args = ' -H -> /tmp/pygmt-dz2kwkqs.cpt'

    def call_module(self, module, args):
        """
        Call a GMT module with the given arguments.
    
        Makes a call to ``GMT_Call_Module`` from the C API using mode
        ``GMT_MODULE_CMD`` (arguments passed as a single string).
    
        Most interactions with the C API are done through this function.
    
        Parameters
        ----------
        module : str
            Module name (``'pscoast'``, ``'psbasemap'``, etc).
        args : str
            String with the command line arguments that will be passed to the
            module (for example, ``'-R0/5/0/10 -JM'``).
    
        Raises
        ------
        GMTCLibError
            If the returned status code of the function is non-zero.
    
        """
        c_call_module = self.get_libgmt_func(
            "GMT_Call_Module",
            argtypes=[ctp.c_void_p, ctp.c_char_p, ctp.c_int, ctp.c_void_p],
            restype=ctp.c_int,
        )
    
        mode = self["GMT_MODULE_CMD"]
        status = c_call_module(
            self.session_pointer, module.encode(), mode, args.encode()
        )
        if status != 0:
            raise GMTCLibError(
                "Module '{}' failed with status code {}:\n{}".format(
>                   module, status, self._error_message
                )
            )
E           pygmt.exceptions.GMTCLibError: Module 'makecpt' failed with status code 71:
E           makecpt [ERROR]: Syntax error: No input files expected unless -E or -S are used

../pygmt/clib/session.py:490: GMTCLibError
--------------------------------------------------------------------------------- Captured stderr call ---------------------------------------------------------------------------------
makecpt [ERROR]: Syntax error: No input files expected unless -E or -S are used
```

Issue #231, opened 2018-08-22
Adding feature of legend and colorbar


palgunadi1993:
I wasn't able to plot my **legend and colorbar** since it is very important issues, I think it's good to have a feature like that. I really appreciate from what you've done so far, I know it's still ongoing progress. However, it seems to have a promising result in the near future. Good Job

jbrussell:
I would like to bump this feature request. Although I'm able to quickly make beautiful maps using this amazing tool, without the ability to add a color bar or legend they cannot be used for research purposes. Very much looking forward to this crucial update!

ardiansyahfauzi:
I have been trying GMT/Python for one week and I found that this is a cool method to use GMT software in an easy way. However, in my case, legend and color bar are badly in need. I am looking forward to these features. Good job!!

liamtoney:
@weiji14 I think we can close this now.

weiji14:
Oh yes, definitely! Everyone subscribed to this thread, `legend` and `colorbar` is implemented, and we're working on some gallery examples. stay tuned!

PR #333, opened 2019-10-08
Wrap legend

### LINKED ISSUE #260: Support for legend with "handles" and "labels" like MATLAB/matplotlib
## Description of the desired feature

A wrapper for the [GMT 6 `legend` command](http://gmt.soest.hawaii.edu/doc/latest/legend.html) with support for specifying "handles" and "labels" as arguments, as in MATLAB ([legend docs](https://www.mathworks.com/help/matlab/ref/legend.html)) and matplotlib ([legend docs](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.legend.html)). The plotting commands (i.e., `gmt.Figure.plot()`, for now) are modified to return "handles" which are provided as inputs to `gmt.Figure.legend()`. This addresses part of #214 and #231.

Below is a minimal working example using my [`add-legend` fork](https://github.com/liamtoney/gmt-python/tree/add-legend):
```python
import gmt

fig = gmt.Figure()

h1 = fig.plot(x=[-5], y=[5], region=[-10, 10, -5, 10], projection='X3i/0', frame='a',
              style='a15p', pen='1.5p,purple')

h2 = fig.plot(x=[0], y=[5],
              style='t10p', color='cyan')

h3 = fig.plot(x=[5], y=[5],
              style='d5p', color='yellow', pen=True)

fig.legend([[h1, h2, h3], ['I am a star!', 'I am a triangle!', 'I am a diamond!']],
           F=True, D='g0/0+w2i+jCM')    

fig.show()
``` 
example

## Are you willing to help implement and maintain this feature?

Yes. You can view the source code for my implementation [here](https://github.com/liamtoney/gmt-python/commits/add-legend/gmt/base_plotting.py).

**The good news:**
- The legend wrapper accepts either the standard *specfile* (see [GMT legend docs](http://gmt.soest.hawaii.edu/doc/latest/legend.html#legend-codes)) or a list containing a list of handles and a list of labels. So folks can still build a complex legend by creating a *specfile*.
- The only modification to `gmt.Figure.plot()` is that it now returns `kwargs`. So existing users won't have to change their syntax for the plot command.

**The bad news:**
- This implementation makes use of `GMTTempFile` since I don't know how to do it any other way (yet).
- Right now I'm using regular expression matching (via [`re`](https://docs.python.org/3/library/re.html#module-re)) to separate symbol type and size from the `style` argument of `gmt.Figure.plot()`, which is probably really sketchy.
- This only works for symbol entries in the legend, i.e. entries in *specfile* which have the form:
`S - symbol size fill pen - text`
I feel like this covers a broad portion of typical legend use cases, though.
- I have not written any tests.

liamtoney:

A wrapper for the [GMT 6 `legend` command](http://gmt.soest.hawaii.edu/doc/latest/legend.html) with support for specifying "handles" and "labels" as arguments, as in MATLAB ([legend docs](https://www.mathworks.com/help/matlab/ref/legend.html)) and matplotlib ([legend docs](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.legend.html)). The plotting commands (i.e., `gmt.Figure.plot()`, for now) are modified to return "handles" which are provided as inputs to `gmt.Figure.legend()`. This addresses part of #214 and #231.

Fixes #260 

**Reminders**


pygmt/clib/session.py
```diff
@@ -112,7 +112,7 @@ class Session:
     """
 
     # The minimum version of GMT required
-    required_version = "6.0.0rc4"
+    required_version = "6.0.0rc2"
```
liamtoney: I probably shouldn't have committed this

weiji14: I was just looking at this too. You should be able to revert that commit easily enough though.

pygmt/base_plotting.py
```diff
@@ -530,3 +534,71 @@ def image(self, imagefile, **kwargs):
         with Session() as lib:
             arg_str = " ".join([imagefile, build_arg_string(kwargs)])
             lib.call_module("image", arg_str)
+
+    @fmt_docstring
+    @use_alias(R="region", J="projection")
```
weiji14:
```suggestion
    @use_alias(R="region", J="projection", D="position", F="box")
```
Create an alias for arguments F (box) and D (position), and update your docstring and tests to use these aliases instead.

Pinging @joa-quim. As per your comment at https://github.com/GenericMappingTools/pygmt/pull/329#issuecomment-536178749 on standardizing aliases, I noticed that GMT.jl uses `position` for [colorbar](https://github.com/GenericMappingTools/GMT.jl/blob/8cff1c46195534c87b95b8b5bfa4cba7e706bdb6/src/psscale.jl#L8) and [logo](https://github.com/GenericMappingTools/GMT.jl/blob/8cff1c46195534c87b95b8b5bfa4cba7e706bdb6/src/gmtlogo.jl#L13), but `ref_point` for [image](https://github.com/GenericMappingTools/GMT.jl/blob/8cff1c46195534c87b95b8b5bfa4cba7e706bdb6/src/psimage.jl#L12) and `refpoint` for [legend](https://github.com/GenericMappingTools/GMT.jl/blob/8cff1c46195534c87b95b8b5bfa4cba7e706bdb6/src/pslegend.jl#L18). Shall we standardize these to `position`?

seisman:
The auto-legend feature has been merged into GMT 6.0 branch in https://github.com/GenericMappingTools/gmt/pull/1749. Now we can use `-l` option in the `plot` module, which enables legends automatically. Below is a simple script showing how we can get auto-legends even without calling the legend module explicitly. 

```
gmt begin plot png
  gmt plot -R0/7/3/7 -Jx1i -B @Table_5_11.txt -Sc0.15i -Glightgreen -Wfaint -lApples
  gmt plot @Table_5_11.txt -W1.5p,gray -l"My Lines"
  gmt plot @Table_5_11.txt -St0.15i -Gorange -lOranges
gmt end show
```
This is what we will get:
image

With the auto-legend feature merged, it's almost straightforward to wrap the legend module in pygmt, and we can write a Python script like:

```
import pygmt
fig = pygmt.Figure()
fig.basemap(projection="x1i", region="0/7/3/7", frame=True)
fig.plot("@Table_5_11.txt", style="c0.15i", color="lightgreen", pen="faint", label="Apples")
fig.plot("@Table_5_11.txt", pen="1.5p,gray", label="My lines")
fig.plot("@Table_5_11.txt", style="t0.15i", color="orange", label="Oranges")
fig.show()
```

The only problem is, GMT 6.0.0 is not official released yet. It shouldn't take too long. We plan to release GMT 6.0.0 when [the GMT6 paper](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2019GC008515) is officially published.

liamtoney:
Great. This should simplify the wrapper a lot. From a dev perspective, to build and test this wrapper do I now need to clone the GMT repo and build from source? Or are there plans for another conda release?

weiji14:
We're still using 6.0.0rc4 for pygmt development, and that's what our code is tested against. I think we should wrap this up and refactor the code after 6.0.0 stable is released. The implementation doesn't have to be perfect, having something is better than nothing!

seisman:
FYI, the GMT team is planing to release 6.0.0rc5 (with the auto-legend feature included) next week, and release the final 6.0.0 on Nov. 1.

weiji14:
Right, it's not quite as easy as it seems... The auto-legend only gets triggered when we run `gmt end`, so the following won't work:

> ```
> import pygmt
> fig = pygmt.Figure()
> fig.basemap(projection="x1i", region="0/7/3/7", frame=True)
> fig.plot(data="@Table_5_11.txt", style="c0.15i", color="lightgreen", pen="faint", l="Apples")
> fig.plot(data="@Table_5_11.txt", pen="1.5p,gray", l="My_lines")
> fig.plot(data="@Table_5_11.txt", style="t0.15i", color="orange", l="Oranges")
> fig.show()
> ```

Produces:

Figure no legend

I'm assuming though that calling `legend` would trigger the legend being plotted, is that correct @seisman？ I can start on some side work to wrap a "label" alias for `text` (and other functions). Plus we need to find a way to handle blank spaces so that `label="My lines"` works instead of having `label="My_lines"`.

liamtoney:
So could the code have an empty legend call after plotting, e.g.
```python
import pygmt
fig = pygmt.Figure()
fig.basemap(projection="x1i", region="0/7/3/7", frame=True)
fig.plot(data="@Table_5_11.txt", style="c0.15i", color="lightgreen", pen="faint", l="Apples")
fig.plot(data="@Table_5_11.txt", pen="1.5p,gray", l="My_lines")
fig.plot(data="@Table_5_11.txt", style="t0.15i", color="orange", l="Oranges")
fig.legend()  # Empty legend call
fig.show()
```
where `fig.legend()` when called with no arguments somehow accesses the hidden specfile if available?

weiji14:
Hmm, I just tried using an empty `fig.legend()` on your branch (after making sure `spec=None` is allowed in the base_plotting.py's `legend` function) and it's complaining that the `-D` (a.k.a `position`) argument is needed. So again, not very automatic but at least we can get rid of the regex bits :man_shrugging:

weiji14:
Ah wait, I just noticed a new [`-T`](https://docs.generic-mapping-tools.org/latest/legend.html?#t) argument mentioned in the [modern mode auto-legend note](https://docs.generic-mapping-tools.org/latest/legend.html#auto-legends) that outputs a `specfile`. We could perhaps set that to `True` if no specfile is given (i.e just a blank legend call) and use the auto-generated specfile to auto-generate the legend, what do you think?

weiji14:
@liamtoney, here's the [git diff patch file](https://github.com/GenericMappingTools/pygmt/files/3767172/legend_patch.txt) you can refer to to get `spec=None` working (you can apply it on top of 24330cdbe1698c87b106e679d43b57e30a147a24 using `git apply legend_patch.txt`). I'm going off to bed now :zzz:, but definitely keep working on this if you have time!

liamtoney:
I noticed that if you enclose the `label` string in an additional pair of quotes, you can use spaces. Still annoying but at least there's a workaround.
```python
import pygmt

fig = pygmt.Figure()

fig.plot(x=[-5, 0, 5], y=[5, 5, 5], region=[-10, 10, -5, 10], projection='X3i/0',
         frame='a', style='a15p', color='purple', label='"I am a star!"')

fig.legend(box=True, position="g0/0+jCM")

fig.show()
```
download

liamtoney:
@weiji14, @seisman, @leouieda — what do you think about having a default argument for `position`? Say, `JTR+jTR` to by default place the legend in the top-right corner, inside. That way a user could have more fluid experience by simply calling `pygmt.legend()`.

weiji14:
I was thinking the same thing. The thing is, GMT does have a default position setting, and we would be 'reinventing the wheel' a bit here...

weiji14:
I was thinking the same thing. The thing is, GMT does have a default position setting, and we would be 'reinventing the wheel' a bit here...

pygmt/tests/test_legend.py
```diff
@@ -0,0 +1,98 @@
...
+G 0.05i
+L 9 4 R Smith et al., @%5%J. Geophys. Res., 99@%%, 2000
+G 0.1i
+P
+T Let us just try some simple text that can go on a few lines.
+T There is no easy way to predetermine how many lines will be required,
+T so we may have to adjust the box height to get the right size box.
+"""
+
+    specfile = GMTTempFile().name
```
weiji14:
```suggestion
    with GMTTempFile() as specfile:
        with open(specfile.name, "w") as file:
            file.write(specfile_contents)

        fig = Figure()
        fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
        fig.legend(specfile.name, position="JTM+jCM+w5i")
```
Would be nice to wrap it around a `with` statement, that way the temporary file gets cleaned up even if the test fails. Oh, and **don't** just commit this automatically! You'll need to change the lines manually below this (Github can't suggest changes for multiple lines yet...).

pygmt/tests/test_legend.py
```diff
@@ -0,0 +1,98 @@
...
+
+    with open(specfile, "w") as file:
+        file.write(specfile_contents)
+
+    fig = Figure()
+
+    fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
+    fig.legend(specfile, position="JTM+jCM+w5i")
+
+    return fig
```
weiji14:
```suggestion
    return fig


def test_legend_fails():
    """
    Test legend fails with invalid spec
    """
    fig = Figure()
    with pytest.raises(GMTInvalidInput):
        fig.legend(spec=["@Table_5_11.txt"])

```

Just adding another test to make sure we capture invalid inputs like a Python list in this case. Needed for full test coverage.

pygmt/tests/test_legend.py
```diff
@@ -0,0 +1,98 @@
+"""
+Tests for legend
+"""
+import pytest
+
+from .. import Figure
+from ..helpers import GMTTempFile
```
weiji14:
```suggestion
from ..exceptions import GMTInvalidInput
from ..helpers import GMTTempFile
```

Needed for the new test below.

liamtoney:
@weiji14 I've made those changes. My commit history is a little messy (sorry) — does anything need to be done about that prior to merging?

pygmt/tests/test_legend.py
```diff
@@ -0,0 +1,109 @@
...
+
+        fig = Figure()
+
+        fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
+        fig.legend(specfile.name, position="JTM+jCM+w5i")
+
+    return fig
+
+
+@pytest.mark.mpl_image_compare
```
weiji14:
```suggestion
```
Oops, you'll need to remove this decorator as there's no output image to test!

liamtoney:
Thanks for walking me through this, @weiji14!

weiji14:
Phew, good work @liamtoney. Thanks for having the stamina to see this one through! Feel free to open up another pull request to get the auto-position done (and we can debate around the implementation :laughing:). Also it'd be great to have a gallery example for `legend` on the website.

pygmt/base_plotting.py
```diff
@@ -181,6 +181,8 @@ def grdcontour(self, grid, **kwargs):
             Do not draw contours with less than `cut` number of points.
         S : string or int
             Resample smoothing factor.
+        l : str
```
seisman: grdcontour doesn't support `-l` option now. Currently only `plot` supports auto-label.

weiji14:
Oops, that's my bad :man_facepalming:. I'll fix that.

pygmt/base_plotting.py
weiji14: Resolving in https://github.com/GenericMappingTools/pygmt/pull/357.

PR #311, opened 2019-06-19
Fixes for using GMT 6.0.0rc1

claudiodsf:
Also fixes the error message on incompatible GMT version.

Fixes #310

PR #343, opened 2019-10-11
Move gmt from requirements.txt to CI scripts instead

weiji14:

Edit: Simply placing `gmt` in our Continuous Integration (CI) scripts' `CONDA_INSTALL_EXTRA` environment variable so that the `requirements.txt` file remains valid for `pip`.

~~Renaming instances of `requirements.txt` -> `conda-requirements.txt`, and `requirements-dev.txt` -> `conda-requirements-dev.txt`.~~

~~The `requirements.txt` file is a standard file used by pip to install Python packages, but we use it here for installing conda packages?! There are actually some automated bots/services that gets confused by this `requirements.txt` file, thinking it should use if to install PyPI packages but fails because it's not meant for that.~~

~~May want to raise this upstream too at https://github.com/fatiando/continuous-integration.~~

Needed for Zeit Now continuous documentation PR at #344, may help resolve https://github.com/weiji14/deepbedmap/issues/153

**Reminders**


seisman:
As far as I can see, all the packages in the requirements.txt and requirements-dev.txt can be installed via pip, except `gmt=6.0.0rc4`, right? As I said in #257, pygmt doesn't really dependent on the gmt package provided by conda-forge. I removed gmt from the requirements.txt files in #257, but it was added back in #311.

weiji14:
Yes, the first problem is that the `gmt=6.0.0rc4` line in `requirements.txt` doesn't work in the `pip install` case.

Second problem is that our CI build actually uses the `requirements.txt` file to install conda packages. Ideally we should be using an `environment.yml` (or similar) file to do that...

.travis.yml
```diff
@@ -24,8 +24,8 @@ env:
         - secure: "md4fgPt9RC/sCoN5//5PcNHLUd9gWQGewV5hFpWW88MRTjxTng1Zfs8r7SqlF2AkEEepFfyzq0BEe9c3FMAnFbec3KmqdlQen4V8xDbLrcTlvkPlTrYGbAScUvdhhqojB//hMHoTD4KvxAv9CiUwFBO4hCMmj2buWHUbV9Ksu5WCW9mF/gkt/hIuYAU6Mbwt8PiYyMgUpzMHO1vruofcWRaVnvKwmBqHB0ae86D4/drpwn4CWjlM12WUnphT2bssiyPkw24FZtCN6kPVta6bLZKBxu0bZpw2vbXuUG+Yh19Q4mp8wNYT3XSHJf8Hl5LfujF48+cLWu+6rlCkdcelyVylhWLFc3rGOONAv4G8jWW2yNSz/bLQfJnMpd81fQEu5eySmFxB7mdB0uyKpvIG1jMJQ73LlYKakKLAPdYhMFyQAHoX9gvCE3S4QR95DBMi5gM/pZubOCcMLdjPHB5JKpJHSjxbOzyVwgmsUIEgd5Bi2vZvvYQXn1plk4xpQ3PhXc+/gi33bzY89mKcfOn0HJ2pD1vLqDXRCBsMCakoLZ0JB/6bacaz4FngbsGWuQ+I1cz20lJGL/MSi9bW1G7Uoidt3GXXWDmXrWt70vIXlLIxr8XV0Mu/rPbauGgWE+ZSYEfvdM5sP+FNF7vQ5de+Fkvzg5Z3tTfR+O1W+d7+vM4="
         - TWINE_USERNAME=Leonardo.Uieda
         # The file with the listed requirements to be installed by conda
-        - CONDA_REQUIREMENTS=requirements.txt
-        - CONDA_REQUIREMENTS_DEV=requirements-dev.txt
+        - CONDA_REQUIREMENTS=conda-requirements.txt
+        - CONDA_REQUIREMENTS_DEV=conda-requirements-dev.txt
         - CONDA_INSTALL_EXTRA="codecov twine"
```
weiji14: What about adding `gmt=6.0.0rc4` here, and removing it from `requirements.txt`? I reckon we should still rename it to `conda-requirements.txt` though.

```suggestion
         - CONDA_INSTALL_EXTRA="codecov twine gmt==6.0.0rc4"
```

seisman: Yes, I think it makes more sense to put `gmt` in `CONDA_INSTALL_EXTRA`. 

As for the renaming, after we remove `gmt` from requirements.txt, I don't see any reason to rename these files. Users can use `pip -r requirements.txt` to install all pygmt dependencies.

weiji14: Ok, I've placed `gmt` in `CONDA_INSTALL_EXTRA` and renamed things back. Just waiting for tests to pass. Oh, and just ignore the 'Now' bot for now :see_no_evil:

leouieda:
This raises an interesting issue that we'll have to think about. If we publish to PyPI, even though the install would work, the user will get an error on import because GMT hasn't been installed as well. Which means that our PyPI packages are almost useless unless we bundle GMT and other binaries into a wheel (which I'm not looking forward to doing). It would be very frustrating for non-GMT users to `pip install pygmt` and not get something useful. 

I don't have a solution for this. If anyone has some thoughts, that would be great!

PR #350, opened 2019-10-20
Quickfix min/max elevation for new SRTM15+V2 earth relief grids

weiji14:
Upstream GMT has updated its earth relief grids (see https://github.com/GenericMappingTools/gmt/pull/1846) to use the new Global Bathymetry and Topography at 15 Arc Sec (SRTM15+V2) data by @btozer and others!

References:

- Tozer, B., Sandwell, D. T., Smith, W. H. F., Olson, C., Beale, J. R., & Wessel, P. (2019). Global Bathymetry and Topography at 15 Arc Sec: SRTM15+. Earth and Space Science. https://doi.org/10.1029/2019EA000658

This pull request fixes the test failures by changing the minimum/maximum elevation values to that of the new earth relief grids.

Fixes #

**Reminders**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -107,8 +107,7 @@ class Session:
     ...             ses.call_module("grdinfo", "{} -C ->{}".format(fin, fout.name))
     ...             # Read the contents of the temp file before it's deleted.
     ...             print(fout.read().strip())
-    -180 180 -90 90 -8425 5551 1 1 361 181
-
+    -180 180 -90 90 -8596 5559 1 1 361 181
     """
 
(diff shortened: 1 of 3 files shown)


weiji14:
Hmm, I would have thought there would be more tests (e.g. from grdimage) failing. Is the `@pytest.mark.mpl_image_compare` decorator not doing its job properly?!!

seisman:
The grdimage tests use the lowest resolution (60m) earth relief data. I think we can't see any difference between the old and new dataset for such low resolution.

weiji14:
Yes, but surely there would be some?? I'll merge this in first and investigate later.

PR #352, opened 2019-10-27
Properly trigger pytest matplotlib image comparison

weiji14:

All the while, we weren't actually testing the image comparisons! I've actually had a hunch a long while back, but it was the change to the SRTM15+V2 grids since #350 that made me think things weren't right. There's also a bigger degree ° symbol now on the x and y axis for some of the plots using `basemap`.

| baseline-test_grdimage | test_grdimage | test_grdimage-failed-diff |
|:---:|:---:|:---:|
| Old | New | Diff |

This Pull Request adds the `--mpl` flag and saves any different/failed results via the `--mpl-results-path=results` which will store the diff images under $TESTDIR/results. See upstream [pytest-mpl docs](https://github.com/matplotlib/pytest-mpl/blob/954073d458ab4d96796c67d6aeb994d17ee39817/README.rst#using).

I'll do this the [Test-Driven Development (TDD)](https://en.wikipedia.org/wiki/Test-driven_development#Test-driven_development_cycle) style, make the tests fail first (so we have a record), and then fix the images to make the tests pass. The failures come mostly from `grdimage` and `grdcontour`, but also `makecpt` and a couple of other random ones...

TODO:
- [x] Trigger test failures (30 actual failures in total, ~35 images to change below).
- [x] Fix `grdcontour`(5), `grdimage` (3), `logo` (2).
- [x] Fix `coast` (4), `image` (1), `makecpt` (8)
- [x] Fix `basemap` (4) and `plot` (8).

Fixes #

**Reminders**


pygmt/tests/test_basemap.py
```diff
@@ -48,9 +48,7 @@ def test_basemap_loglog():
 def test_basemap_power_axis():
     "Create a power axis basemap plot"
     fig = Figure()
-    fig.basemap(
-        R=[0, 100, 0, 5000], J="x1p0.5/-0.001", B='x1p+l"Crustal age"y500+lDepth'
-    )
+    fig.basemap(R=[0, 100, 0, 5000], J="x1p0.5/-0.001", B='x1p+l"Crustal age"')
```
weiji14: I've fixed this because the original image had "Depth" instead of "Crustal Age" as the x-axis label... See https://github.com/GenericMappingTools/pygmt/pull/352/files?short_path=edaeec0#diff-edaeec03fad61caa0eca5752b8b18d02

seisman: I think the original test wants to add labels for both X and Y axis. It should be changed to:
```suggestion
    fig.basemap(R=[0, 100, 0, 5000], J="x1p0.5/-0.001", B=['x1p+l"Crustal age"', 'y500+lDepth'])
```

weiji14: Ok, I'll change that in a minute. Will need to regenerate the plot too.

PR #358, opened 2019-11-01
Add legend gallery entry

liamtoney:

Just adding a legend example to the gallery.

**Reminders**


examples/gallery/plot/legend.py
```diff
@@ -0,0 +1,26 @@
...
+------
+
+The :meth:`pygmt.Figure.legend` method can automatically create a legend for
+symbols plotted using :meth:`pygmt.Figure.plot`.
+"""
+import pygmt
+
+fig = pygmt.Figure()
+
+fig.basemap(projection="x1i", region="0/7/3/7", frame=True)
```
weiji14: I think it's nicer to use a list here
```suggestion
fig.basemap(projection="x1i", region=[0, 7, 3, 7], frame=True)
```

examples/gallery/plot/legend.py
```diff
@@ -0,0 +1,26 @@
...
+    data="@Table_5_11.txt",
+    style="c0.15i",
+    color="lightgreen",
+    pen="faint",
+    label="Apples",
+)
+fig.plot(data="@Table_5_11.txt", pen="1.5p,gray", label='"My lines"')
+fig.plot(data="@Table_5_11.txt", style="t0.15i", color="orange", label="Oranges")
+
+fig.legend(position="JTR+jTR", box=True)
```
weiji14: It might be nice to have a little offset from the edge of the map frame, something like this:
```suggestion
fig.legend(position="JTR+jTR+o0.2c", box=True)
```

examples/gallery/plot/legend.py
```diff
@@ -0,0 +1,26 @@
+"""
+Legend
+------
+
+The :meth:`pygmt.Figure.legend` method can automatically create a legend for
+symbols plotted using :meth:`pygmt.Figure.plot`.
```
weiji14: Might want to be a little bit more descriptive? E.g. tell users that they will need to add `label="SomeText"` to the `plot` argument in order for `legend` to work!

Issue #368, opened 2019-11-04
grdimage can not plot data with decreasing latitudes


MarkWieczorek:
`grdimage` fails when attempting to plot grids that are not "global" [0, 360, -90, 90].

The xarray dataset I am trying to plot is equispaced in latitude and longitude, and normally would work with gmt. When trying to plot with pygmt however, the routine goes into some kind of infinite loop and never terminates.

The xarray dataset has the following characteristics:

```
Coordinates:
  * latitude   (latitude) float64 90.0 80.0 70.0 60.0 ... -60.0 -70.0 -80.0
  * longitude  (longitude) float64 0.0 10.0 20.0 30.0 ... 330.0 340.0 350.0
```
and I tried to plot the data using

```
fig.grdimage(grid, region=[0., 360., -90, 90], projection="W0/6i")
```

I note that the redundant data points at 360 E are not provided, and the data point at the south pole is not provided. I further note that this is the standard format for Driscoll and Healy sampled grids which are employed in `pyshtools`.

MarkWieczorek:
With more investigation, it turns out that grdimage works when reading a netcdf file, but **not** when using an xarray dataarray. In particular, the following works:

```
fig.grdimage('test.cdf', region='g', projection="W0/6i")
```

but the following doesn't
```
grid = xr.load_dataset('test.cdf')
fig.grdimage(grid.data, region='g', projection="W0/6i") # data is the name of the dataarray
```

MarkWieczorek:
I've narrowed the problem down to the following:

pygmt can not deal with grids where latitudes decrease, such as [90, 80, ....-80].

By inverting the array of latitudes, everything worked fine (but with an upside down image). Given that gmt can correctly parse this, the problem is somewhere in pygmt.

weiji14:
Hi @MarkWieczorek, thanks for taking a look at this, and opening a Pull Request too! I'll try and review your work in a bit.

weiji14:
Closed in #369.

PR #369, opened 2019-11-05
Allow for grids with negative lat/lon increments

MarkWieczorek:
pygmt assumes that both the latitudes and longitudes of an xarray Dataset increase with increasing indices of the array. However, it is common for the latitudes to be arranged with decreasing values such that the first row corresponds to 90 N and the last to 90 S. Furthermore, in some weird and very old cases, planetary longitude has been defined as increasing in the westward direction.

Here, I have modified the function `dataarray_to_matrix` to check whether the longitude and latitude increments are positive. If they are negative, the array is flipped on the corresponding dimension. 

There are four possible scenarios, and I have attempted to avoid creating an additional array. If there is a better way to do this, please let me know!

Fixes # https://github.com/GenericMappingTools/pygmt/issues/368

**Reminders**


weiji14:
Right, I've taken a hard look at this, and you've definitely done a good job at getting the functionality right! I think we can improve the readability of the code though, e.g. by using [`np.flip`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.flip.html) instead of `grid.values[:, ::-1]`. Will recommend some [suggested changes](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-a-suggested-change) shortly.

I'm not sure if you're familiar with writing tests, but could you add some for the `dataarray_to_matrix` function? Just add them above these ones:

https://github.com/GenericMappingTools/pygmt/blob/55af9b67b5f04a1cde30cba013eec31d44874d4f/pygmt/tests/test_clib.py#L767-L786?

Just give me a shout out if you need any help :smile:

pygmt/clib/conversion.py
```diff
@@ -102,7 +104,19 @@ def dataarray_to_matrix(grid):
             )
         region.extend([coord.min(), coord.max()])
         inc.append(coord_inc)
-    matrix = as_c_contiguous(grid.values[::-1])
+    if inc[1] < 0:
```
weiji14: So this uses xarray's builtin [sort](https://xarray-test.readthedocs.io/en/latest/reshaping.html#sort) mechanism that simplifies the code by heaps.
```suggestion

    if any([i < 0 for i in inc]):  # Sort grid properly if there are negative increments
        inc = [abs(i) for i in inc]
        grid = grid.sortby(variables=list(grid.dims), ascending=True)

    matrix = as_c_contiguous(grid.values[::-1])
```
Note :warning:: **do not** commit this change directly, you'll need to manually edit the file and make the change to the lines accordingly.

Also, it would be good to see some unit tests to make sure it actually works for every possible combination of flipped latitude/longitude coordinates.

MarkWieczorek: That works. I suspect that the sorting operation is slower than simply flipping the array, but compared to everything involved with the projection routines, this is certainly not a limiting operation.

As for the tests, I'd be happy to try, but I'm not sure what the point would be. It seems that the test would be to check if xarray works properly, not pygmt. What kind of test were you thinking of?

**Notes**
* It appears that this works even if the DataSet doesn't supply coordinates (the input grid is unmodified).
* The only problem I could envision is if the user defined the longitude coordinates as `[0, 90, 180, 270, 0]`. In this case the sort operation would provide incorrect results. It is hard to imagine someone doing this. If there is a test for this scenario, I would probably redefine the last coordinate somewhere else in the code. As part of such a test, you might also want to check that the two data values are the same (I think that gmt might already do this).

weiji14: If speed is the concern, it would be best to pass in a file anyway. You're right that the test is more of a check on xarray than pygmt, but we're actually not testing that `dataarray_to_matrix` properly now so it's a good time as any to add it.

weiji14:
@leouieda, do you want to take a look at this `dataarray_to_matrix` function if you have time?

pygmt/clib/conversion.py
```diff
@@ -102,6 +104,11 @@ def dataarray_to_matrix(grid):
             )
         region.extend([coord.min(), coord.max()])
         inc.append(coord_inc)
+
+    if any([i < 0 for i in inc]):  # Sort grid when there are negative increments
+        inc = [abs(i) for i in inc]
+        grid = grid.sortby(variables=list(grid.dims), ascending=True)
```
leouieda: One thing to check here is if this generates a memory copy. Though there isn't a good way around it, I suppose, since `as_c_contiguous` will have to make a copy if that's the case.

pygmt/clib/conversion.py
leouieda: > If speed is the concern, it would be best to pass in a file anyway. 

I'm not sure even GMT handles this properly. So passing in a file probably doesn't help much.

pygmt/clib/conversion.py
weiji14: I believe it does (create a memory copy). Maybe we should revert back to the if-then checks then? Not as readable but if speed is what matters then it'll be faster.

Alternatively, we *could* require users to sort their grid properly (be it via flipping the axes or using sortby) before passing into PyGMT, but it won't be a fantastic user experience.

MarkWieczorek: I think that this is probably ok as is: If the grid is properly ordered, then this code block doesn't get executed anyways.

weiji14:
Just checking in here @MarkWieczorek. Have you had a chance to write the tests yet? I'd be keen to merge this PR as soon as possible as I've been hitting the same flipped lat/lon problem a lot recently too :joy:

MarkWieczorek:
Sorry but no. I all likelihood, this will probably take a week or two (I'm busy, plus I've never written pytests before). If you do this for me, I promise instead to improve the web documentation later...

Also, I think that this issue is probably related: https://github.com/GenericMappingTools/pygmt/issues/375

I'm not sure exactly what is going on, but I think that `dataarray_to_matrix` probably needs to check if the last redundant longitude column is missing, and if so, to add it.

weiji14:
Deal, I'll write up the tests for this now. I did take a look at that other issue, but it will require a bit more investigation.

weiji14:
Thanks @MarkWieczorek! Hopefully this resolves your original issue at #368, I know it'll definitely help my workshop tutorials run more smoothly tomorrow. Definitely feel free to contribute more in the future, I can see you've got a lot of great ideas :smile:

leouieda:
Thank you for the contribution Mark! :tada:

PR #379, opened 2019-11-18
Proj classes

### LINKED ISSUE #356: Create Projection classes instead of strings

The `Figure` methods take a `projection` argument that is using the same arguments as GMT `-J` (`"M10i"` for Mercator 10 inches). But a lot of projections have many arguments and it's impossible to remember them all. A relatively simple alternative would be to implement classes (based on a single `Projection` class maybe) that take the projection arguments in the constructor (and so allows tab-completion). The key would be for these classes to define a `__str__` method so that when `str` is called on them, they produce the GMT `-J` code. For example:

```python
>>> proj = pygmt.Orthographic(central_longitude=10, central_latitude=52, width=10)
>>> print(str(proj))
G10/52/10
```

The plotting method pre-processing would only have to call `str` on `projection`, which works fine if you pass in a string as well. 

**Are you willing to help implement and maintain this feature?** Yes but I'd welcome anyone else who wants to try :)

sixy6e:
**WIP; Create Projection classes instead of strings**

This pull request is to address #356 and create projection classes instead of requiring the user to format their own strings which can be cumbersome and complicated.
The desired functionality was to enable tab completion enabling users to immediately see what projections are supported. It is also desirable for the classes to utilise the __str__ method to return a correctly formatted string supported by GMT.

An example of creating an lambert azimuthal equal area projection:
```python
>>> from pygmt import projection
>>> proj = projection.LambertAzimuthalEqualArea(lon0=30, lat0=-20, horizon=60, width="8i")
>>> proj
LambertAzimuthalEqualArea(lon0=30, lat0=-20, horizon=60, width='8i')
>>> print(proj)
A30/-20/60/8i
```

The supported projections are also defined via an [Enum](https://docs.python.org/3/library/enum.html).  These act as constants (change in one place only) and serve multiple purposes:
* defining and retrieving the projection name
* defining and retrieving the GMT code
* listing the supported projections
* easy boolean comparison of projection labels

All the classes are relying on [attrs](https://github.com/python-attrs/attrs) for simplicity in creation, whilst getting all the benefits that attrs provides such as an easier declaration of validators thus enabling parameter checks for invalid values.

Arguments need to be supplied as keywords (was simpler to define projections classes with mixtures of default values and None for parameters; the benefit is greater clarity to the user and avoiding mixups of incorrect parameter ordering).
The projection classes are also defined to be immutable, acting as a kind of config (this can be changed if desired). 

Fixes #356 

**Reminders**

- [ ] Finish including the other projections that GMT supports.

### DIFF
--- a/pygmt/projection.py
+++ b/pygmt/projection.py
@@ -1,0 +1,315 @@
+#!/usr/bin/env python
+
+"""
+Contains the projections supported by GMT, and the necessary mechanisms
+to create a projection and output a valid GMT projection string.
+
+>>> from pygmt import projection
+>>> proj = projection.LambertAzimuthalEqualArea(lon0=30, lat0=-20, horizon=60, width="8i")
+>>> proj
+LambertAzimuthalEqualArea(lon0=30, lat0=-20, horizon=60, width='8i')
+>>> print(proj)
+A30/-20/60/8i
+"""
+
+from enum import Enum
+import attr
+
+
+class Supported(Enum):
+
+    """
+    The supported projections and their GMT code.
+    """
+
+    UNDEFINED = ""
+    LAMBERT_AZIMUTH_EQUAL_AREA = "A"  # DONE
+    ALBERS_CONIC_EQUAL_AREA = "B"  # DONE
+    CASSINI_CYLINDRICAL = "C"  # DONE
+    CYLINDRICAL_STEROGRAPHIC = "JCyl_stere/"  # includes `/` according to https://docs.generic-mapping-tools.org/latest/proj_codes.html  # DONE
+    EQUIDISTANT_CONIC = "JD"  # DONE
+    AZIMUTHAL_EQUIDISTANT = "E"  # DONE
+    AZIMUTHAL_GNOMIC = "F"  # DONE
+    AZIMUTHAL_ORTHOGRAPHIC = "G"  # DONE
+    GENERAL_PERSPECTIVE = "G"  # DONE
+    HAMMER_EQUAL_AREA = "H"
+    SINUSOIDAL_EQUAL_AREA = "I"
+    MILLER_CYLINDRICAL = "J"
+    ECKERT_IV_EQUAL_AREA = "Kf"
+    ECKERT_VI_EQUAL_AREA = "Ks"
+    LAMBERT_CONIC_CONFORMAL = "L"
+    MERCATOR_CYLINDRICAL = "M"  # DONE
+    ROBINSON = "N"
+    OBLIQUE_MERCATOR_1 = "Oa"
+    OBLIQUE_MERCATOR_2 = "Ob"
+    OBLIQUE_MERCATOR_3 = "Oc"
+    POLAR = "P"
+    POLYCONIC = "Poly"
+    EQUIDISTANT_CYLINDRICAL = "Q"
+    WINKEL_TRIPEL = "R"
+    GENERAL_STEREOGRAPHIC = "S"  # DONE
+    TRANSVERSE_MERCATOR = "T"
+    UNIVERSAL_TRANSVERSE_MERCATOR = "U"
+    VAN_DER_GRINTEN = "V"
+    MOLLWEIDE = "W"
+    LINEAR = "X"
+    CYLINDRICAL_EQUAL_AREA = "Y"  # DONE
+
+
+@attr.s()
+class _Projection:
+
+    """
+    Base class for all projections.
+    """
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False, default="{_code}")
+    _code: str = attr.ib(init=False, repr=False,
+                         default=Supported.UNDEFINED.value)
+
+    def __str__(self):
+        exclude = attr.fields(self.__class__)._fmt
+        kwargs = attr.asdict(self, filter=attr.filters.exclude(exclude))
+        return self._fmt.format(**kwargs)
+
+
+@attr.s(kw_only=True)
+class _Azimuthal(_Projection):
+
+    """
+    Base class for azimuthal projections.
+    """
+
+    lon0: float = attr.ib()
+    lat0: float = attr.ib()
+    horizon: float = attr.ib(default=90)
+    width: str = attr.ib()
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False,
+                        default="{_code}{lon0}/{lat0}/{horizon}/{width}")
+    _code: str = attr.ib(init=False, repr=False,
+                         default=Supported.UNDEFINED.value)
+
+    @horizon.validator
+    def check_horizon(self, attribute, value):
+        """
+        Validate the horizon attribute.
+        """
+        if value > 180:
+            raise ValueError("horizon must be less than or equal to 180")
+
+
+@attr.s(kw_only=True)
+class _Cylindrical(_Projection):
+
+    """
+    Base class for cylindrical projections.
+    """
+
+    lon0: float = attr.ib()
+    lat0: float = attr.ib()
+    width: str = attr.ib()
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False,
+                        default="{_code}{lon0}/{lat0}/{wdith}")
+    _code: str = attr.ib(init=False, repr=False,
+                         default=Supported.UNDEFINED.value)
+
+@attr.s(kw_only=True)
+class _Conic:
+
+    """
+    Base class for conic projections.
+    """
+
+    lon0: float = attr.ib()
+    lat0: float = attr.ib()
+    lat1: float = attr.ib()
+    lat2: float = attr.ib()
+    width: float = attr.ib()
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False,
+                        default="{_code}{lon0}/{lat0}/{lat1}/{lat2}/{width}")
+
+
+@attr.s(frozen=True)
+class LambertAzimuthalEqualArea(_Azimuthal):
+
+    """
+    Definition for the lambert azimuthal equal area projection.
+    """
+
+    # private; we don't want the user to care or know about
+    _code: str = attr.ib(init=False, repr=False,
+                         default=Supported.LAMBERT_AZIMUTH_EQUAL_AREA.value)
+
+
+@attr.s(frozen=True)
+class AzimuthalEquidistant(_Azimuthal):
+
+    """
+    Definition for the azimuthal equidistant projection.
+    """
+
+    horizon: float = attr.ib(default=180, kw_only=True)
+
+    # private; we don't want the user to care or know about
+    _code: str = attr.ib(init=False, repr=False,
+                         default=Supported.AZIMUTHAL_EQUIDISTANT.value)
+
+
+@attr.s(frozen=True)
+class AzimuthalGnomic(_Azimuthal):
+
+    """
+    Definition for the azimuthal gnomic projection.
+    """
+
+    horizon: float = attr.ib(default=60, kw_only=True)
+
+    # private; we don't want the user to care or know about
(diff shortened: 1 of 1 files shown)


pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+@attr.s()
+class _Projection:
+
+    """
+    Base class for all projections.
+    """
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False, default="{_code}")
+    _code: str = attr.ib(init=False, repr=False, default=Supported.UNDEFINED.value)
```
leouieda: `_code` and `_fmt` are the main things that child classes need to specify, right? If so, they should abstract in the base class. To do that, the `_Projection` and `_Azimuthal` et al should probably be [abstract base classes](https://docs.python.org/3/library/abc.html).

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    unit: str = attr.ib(default="i")
+    center: list = attr.ib(init=False)
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(
+        init=False,
+        repr=False,
+        default="{_code}{central_longitude}/{central_latitude}/{horizon}/{width}{unit}",
+    )
+    _code: str = attr.ib(init=False, repr=False, default=Supported.UNDEFINED.value)
```
leouieda: There is probably no need to define `_code` as "undefined" here. It's what has to be implemented by the base classes, right?

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    central_latitude: float = attr.ib()
+    width: float = attr.ib()
+    unit: str = attr.ib(default="i")
+    center: list = attr.ib(init=False)
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(
+        init=False,
+        repr=False,
+        default="{_code}{central_longitude}/{central_latitude}/{wdith}{unit}",
```
leouieda:
```suggestion
        default="{_code}{central_longitude}/{central_latitude}/{width}{unit}",
```

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
+#!/usr/bin/env python
+
```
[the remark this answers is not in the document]
sixy6e: Will remove.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+"""
+Contains the projections supported by GMT, and the necessary mechanisms
+to create a projection and output a valid GMT projection string.
+
+>>> from pygmt import projection
+>>> proj = projection.LambertAzimuthalEqualArea(central_longitude=30, central_latitude=-20, horizon=60, width=8, unit="i")
+>>> proj
+LambertAzimuthalEqualArea(central_longitude=30, central_latitude=-20, horizon=60, width=8, unit='i', center=[30, -20])
+>>> print(proj)
+A30/-20/60/8i
```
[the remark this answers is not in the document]
sixy6e: Will remove the module level example. I'll put some examples together in the tutorials section soon.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+"""
+
+from enum import Enum
+import attr
+
+
+class Supported(Enum):
```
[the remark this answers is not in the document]
sixy6e: It's not really needed anymore.  I was mostly using it as a place holder for the GMT projection codes, and to keep track (for myself) what had and had not been implemented.
I'll remove it when the projection class module gets closer to completion.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    UNIVERSAL_TRANSVERSE_MERCATOR = "U"
+    VAN_DER_GRINTEN = "V"
+    MOLLWEIDE = "W"
+    LINEAR = "X"
+    CYLINDRICAL_EQUAL_AREA = "Y"  # DONE
+
+
+@attr.s()
+class _Projection:
+
```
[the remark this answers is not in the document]
sixy6e: Sorry, my bad.  I'll update all places.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    Base class for all projections.
+    """
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False, default="{_code}")
+    _code: str = attr.ib(init=False, repr=False, default=Supported.UNDEFINED.value)
+
+    def __str__(self):
+        exclude = attr.fields(self.__class__)._fmt
+        kwargs = attr.asdict(self, filter=attr.filters.exclude(exclude))
```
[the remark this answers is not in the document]
sixy6e: `_code` is required for the string formatter to work. But I thought it best to keep the attribute hidden from the user, and let the class populate by default the required GMT code.

pygmt/projection.py
sixy6e: Good pickup :+1:

pygmt/projection.py
sixy6e: As `_Projection`, `_Azimuthal`, `_Cylindrical` are considered private, I wouldn't expect the user to call them directly.  But they could and it would fail if `_code` is removed, as the `_fmt` attribute contains `{_code}`.  It is just an empty string placeholder, as such in order to avoid failure from a user calling it explicitly, both the attribute `_code` and the portion of the `str` containing `{_code}` could be removed.
Any thoughts on that approach?

pygmt/projection.py
sixy6e: Will fix it.

pygmt/projection.py
sixy6e: Actually, scratch that, I can't modify the formatter string, as it used by most of the subclasses.

sixy6e: Woops again. I see what you mean.  I'll remove it, and maybe I should get to bed ;)

pygmt/projection.py
sixy6e: I'm not sure how well an `ABC` will work with a child class being defined via `attrs`.  From the docs, `attrs` trawls the class hierarchy and collects all attributes, and then writes its' own methods to access the required attributes.

So setting an attribute on a child class on a property defined via an `ABC` fails.

eg:
```python
from abc import ABCMeta, abstractproperty
@attr.s
class _Proj:
    __metaclass__ = ABCMeta

    @abstractproperty
    def longitude(self):
        pass

    @abstractproperty
    def _code(self):
        pass

@attr.s
class MyProj(_Proj):
    longitude: float = attr.ib()
    _code: str = attr.ib(default="X", init=False, repr=False)
```

Calling:
```python
p = MyProj(145)
  File "<stdin>", line 1, in <module>
  File "<attrs generated init __main__.MyProj-2>", line 3, in __init__
AttributeError: can't set attribute
```

A way around this would be to create a different attribute and define a property that returns it, but I think it then adds additional complexity to the class definition when you need to define the attribute.

eg
```python
@attr.s
class MyProj(_Proj):
    _longitude = attr.ib()
    __code = attr.ib(default="X", init=False, repr=False)

    @property
    def longitude(self):
        return self._longitude

    @property
    def _code(self):
        return self.__code
```

Was this along the lines of what you meant when defining the base classes to be an [abstract base class](https://docs.python.org/3/library/abc.html)?
I'll admit I haven't done much with them since making the move to [attrs](https://www.attrs.org/en/stable/index.html).

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+
+    """
+    Base class for all projections.
+    """
+
+    # private; we don't want the user to care or know about
+    _fmt: str = attr.ib(init=False, repr=False, default="{_code}")
+    _code: str = attr.ib(init=False, repr=False, default=Supported.UNDEFINED.value)
+
+    def __str__(self):
```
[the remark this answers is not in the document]
sixy6e: I'll add it in.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    unit : str
+        The unit for the figure width in ``i`` for inch, ``c`` for centimetre.
+        Default is ``i``.
+    """
+
+    central_longitude: float = attr.ib()
+    central_latitude: float = attr.ib()
+    horizon: float = attr.ib(default=90)
+    width: float = attr.ib()
+    unit: str = attr.ib(default="i")
```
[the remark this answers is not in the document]
sixy6e: Haha, I'm on board with this one!  I'll make the changes.

pygmt/projection.py
```diff
@@ -0,0 +1,590 @@
...
+    central_latitude : float
+        The latitude of the projection centre.
+    width : float
+        The figure width.
+    unit : str
+        The unit for the figure width in ``i`` for inch, ``c`` for centimetre.
+        Default is ``i``.
+    """
+
+    # private; we don't want the user to care or know about
```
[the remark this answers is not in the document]
sixy6e: Fair comment.  I'll remove them.

MarkWieczorek:
Is this PR about finished? I'd like to start implementing it in pyshtools if it is ready to go...

sixy6e:
Sorry @MarkWieczorek not quite there yet.  I've been busy of late and haven't had much of chance to do more work on it.
Most of the projections are defined, with only Linear and Polar left to do (each has their quirks which complicate things).
I've started work on the unittests, and then it'll be the docs and examples.

meghanrjones:
Hi @sixy6e, I'm checking in to see whether you're interested in continuing work on this PR, prefer to pass the PR to someone else, or rather abandon the PR for a separate design? For context, I'm giving a talk about PyGMT at SciPy next month and would like to solicit feedback on potential API improvements (including projection classes). Please let me know your preferred path forward 🚀 and thanks for your work on this to date 🙌

maxrjones:
> Hi @sixy6e, I'm checking in to see whether you're interested in continuing work on this PR, prefer to pass the PR to someone else, or rather abandon the PR for a separate design? For context, I'm giving a talk about PyGMT at SciPy next month and would like to solicit feedback on potential API improvements (including projection classes). Please let me know your preferred path forward 🚀 and thanks for your work on this to date 🙌

Following up on this @sixy6e because I'm keen to move this forward over this holidays. Unless I hear differently I'll go ahead and build this further as a separate PR.

sixy6e:
Hi @maxrjones 

Thanks for the reminder ping. I've been out of action for quite a while with project work. 

I'm keen to wrap this work up. I'll do a refresh of where the work was at tomorrow. From memory there were a couple projections left but they might've been a slightly different structure. So will need some input.

sixy6e:
I'm currently adding in more tests.  For the output projection string, is there a preference for including the prefix "-J" or leaving it out?

maxrjones:
> I'm currently adding in more tests. For the output projection string, is there a preference for including the prefix "-J" or leaving it out?

Nice! My initial reaction would be to leave it out, but open to other suggestions here.

sixy6e:
Woops.  Sorry @maxrjones 

I wasn't thinking, and just auto-piloted with a rebase off "main" as my original branch was quite old without much further thought.  Please let me know how you'd like to progress.

maxrjones:
> I wasn't thinking, and just auto-piloted with a rebase off "main" as my original branch was quite old without much further thought. Please let me know how you'd like to progress.

> I've reset my branch to just before the rebase to avoid including the mass history (i probably should've done git pull --rebase).. But please let me know what you would prefer.

We use squash commits when merging any PRs into main, meaning that there are no benefits for the overall project history of using rebase rather than merge for keeping PR branches up to date with main. We also clean up the log that is included in the commit message when merging PRs. So, I recommend merging main into your proj-classes branch either locally and pushing to your fork or using the GitHub 'Update branch' button and then pulling the merge commit to your clone. While git log will show all the commits included in the merge (over 1000 in your case), the GitHub PR interface will only show the last merge commit so it's still easy to track which of the commits are associated with your changes.

> Update as to where things are at. All projections should now be implemented, along with a unittests for each projection.

Awesome! I'll take a more detailed look tomorrow - really appreciate your continued work on this!

PR #386, opened 2019-12-19
minor docs improvements

kmaterna:
First ever pull request!  I've been using pygmt and it's great- thanks for making it.  
I found a few minor points in the docs that would hopefully clarify "coast" and "-B" in my experience.

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -115,6 +115,9 @@ class BasePlotting:
             ``'river[/pen]'``
             Draw rivers. Specify the type of rivers and [optionally] append pen
             attributes.
+        L : str
+            ``'[g|j|J|n|x]refpoint'``
+            Draws a simple map scale centered on the reference point specified.            
         N : str
             ``'border[/pen]'``
             Draw political boundaries. Specify the type of boundary and
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -64,6 +64,7 @@ class BasePlotting:
         B="frame",
         D="resolution",
         I="rivers",
+        L="map_scale",
         N="borders",
         W="shorelines",
         G="land",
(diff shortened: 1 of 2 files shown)


pygmt/base_plotting.py
```diff
@@ -115,6 +115,9 @@ def coast(self, **kwargs):
             ``'river[/pen]'``
             Draw rivers. Specify the type of rivers and [optionally] append pen
             attributes.
+        L : str
```
weiji14: Do you mind adding an alias for `L` to `coast`? [GMT.jl](https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.coast) uses `map_scale` and I think we can copy that. It'll just be a one line addition.

pygmt/helpers/decorators.py
```diff
@@ -23,7 +23,7 @@
             *Required if this is the first plot command*.
             Select map projection.""",
     "B": """\
-        B : str
+        B : str or list
```
weiji14: Good spotting, cross referencing #361.

pygmt/base_plotting.py
```diff
@@ -115,6 +115,9 @@ def coast(self, **kwargs):
...
+            ``'[g|j|J|n|x]refpoint'``
+            Draws a simple map scale centered on the reference point specified.
```
weiji14:
```suggestion
            Draws a simple map scale centered on the reference point specified.
```
Just remove the extra whitespace to make the tests pass.

pygmt/base_plotting.py
kmaterna: Ok, I think that's done in the new commit. Thanks!

weiji14:
Thanks @kmaterna! Feel free to point out any other things in the documentation you'd like to improve, or new features you'd like to see in the future. Pull requests like this are always very welcome :smiley:

PR #397, opened 2020-03-11
Update GMT constant GMT_STR16 to GMT_VF_LEN for GMT API change in 6.1.0

seisman:

PyGMT fails with the latest GMT master branch with the following error message:
```
GMTCLibError: Constant 'GMT_STR16' doesn't exits in libgmt.
```

If you have the GMT master branch installed, you can reproduce the issue by
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region='0/10/0/10', projection='X10c', frame=True)
fig.plot(x=5, y=5, style='c0.2c')
```

The error is caused by the recent change of the length of virtual file names
from `GMT_STR16` to `GMT_VF_LEN` in the core GMT
(see https://github.com/GenericMappingTools/gmt/pull/2861).

Changing `GMT_STR16` to `GMT_VF_LEN` is the easiest fix.
To keep compatibility with both GMT 6.0.0 and the upcomming 6.1.0,
this PR checks GMT version and use `GMT_STR16` for 6.0.0, otherwise
use `GMT_VF_LEN`.

**Reminders**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -987,7 +987,12 @@ class Session:
             valid_modifiers=["GMT_IS_REFERENCE", "GMT_IS_DUPLICATE"],
         )
 
-        buff = ctp.create_string_buffer(self["GMT_STR16"])
+        # The core GMT changes GMT_STR16 to GMT_VF_LEN in 6.1.0
+        # See https://github.com/GenericMappingTools/gmt/pull/2861
+        if Version(self.info["version"]) < Version("6.1.0"):
+            buff = ctp.create_string_buffer(self["GMT_STR16"])
+        else:
+            buff = ctp.create_string_buffer(self["GMT_VF_LEN"])
 
         status = c_open_virtualfile(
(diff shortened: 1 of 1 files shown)


leouieda:
@seisman thanks for the fix! Sorry I've been away from the project for so long. Hoping to get some time during spring break to do some long delayed refactoring here.

It's a bit bad that we're advertising the GMT API as a stable thing in the papers and then changing it like that. I understand that it probably needed to be done but this type of thing is what makes it hard to work on PyGMT. A better thing to do would have been to recommend using the new variable to keep the old one for compatibility until GMT 7.

leouieda:
We really do need to setup CI against GMT master to find these things.

leouieda:
Actually, if any one you have any suggesting for possible tests for this that would be great. I'm not keen on disabling the "required" checks for merging since that creates a bad precedent.

seisman:
I don't think it's possible to add any tests to make the coverage checker happy, since we can't simply test two GMT versions in one CI job. 

Let's merge it first.

PR #404, opened 2020-03-26
Tidy up travis.yml file

### LINKED ISSUE #403: Migrate Travis CI from Services to Github Apps

Some travis CI test statuses are not being reported back to GIthub (see https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798), basically because we're using the deprecated 'Services' integration:

Deprecated Travis CI Service image

So we'll need to move to install the new [Github Apps integration](https://github.com/marketplace/travis-ci). Instructions can be found at https://docs.travis-ci.com/user/migrate/legacy-services-to-github-apps-migration-guide/. Also should take a look at our `.travis.yml` which has some [deprecated keys](https://docs.travis-ci.com/user/build-config-validation/#deprecated_key).

weiji14:

Just tidying up the .travis.yml to remove some keys that will be deprecated in [dpl](https://github.com/travis-ci/dpl) v2, see https://blog.travis-ci.com/2019-10-24-build-config-validation. Developers can validate the `.travis.yml` file using https://config.travis-ci.com/explore.

Fixes #403 

**Reminders**


.travis.yml
```diff
@@ -83,14 +80,12 @@ deploy:
     # Push the built HTML in doc/_build/html to the gh-pages branch
     - provider: script
       script: continuous-integration/travis/deploy-gh-pages.sh
```
leouieda: Is there an alternative for these or are they the default now? This was required before so that the files we want to deploy didn’t get delete.

weiji14: They do have a key called `cleanup` now (opposite to the `skip_cleanup`) which defaults to false. Should I put that in?

leouieda: Yeah, it might be best to set it explicitly if they decide to change it later on.

weiji14: Ok, I've added back `cleanup:false`. For future reference, see https://travis-ci.community/t/skip-cleanup-true-is-now-deprecated-how-to-go-for-gradle-publish/6641.

PR #421, opened 2020-04-17
Update PyPI install instructions and API disclaimer message

weiji14:

In anticipation of [v0.1.0](https://github.com/GenericMappingTools/pygmt/milestones/0.1.0) release, we're recommending users to just `pip install pygmt` (from [PyPI](https://pypi.org/project/pygmt)) now instead of installing from Github.

Also warning Windows users that using GMT from conda-forge won't work and suggest workarounds. Plus put a disclaimer that the API (functions and classes) can still change in future releases.

Fixes #

**Reminders**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,10 +35,12 @@ PyGMT
 Disclaimer
 ----------
 
-🚨 **This package is in the early stages of design and implementation.** 🚨
+🚨 **This package is still undergoing rapid development.** 🚨
 
-All functions/classes/interfaces are subject to change as we experiment with new design
-ideas and implement new features. **This is NOT a finished product.**
+All of the API (functions/classes/interfaces) is subject to change until we reach v1.0.0
+as per the `semantic versioning specification <https://semver.org/spec/v2.0.0.html>`__.
+There may be non-backward compatible changes as we experiment with new design ideas and
+implement new features. **This is not a finished product, use with caution**
 
 We welcome any feedback and ideas!
 Let us know by submitting
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -32,16 +32,18 @@ PyGMT requires Generic Mapping Tools (GMT) version 6 as a minimum, which is the 
 released version that can be found at
 `this website <https://www.generic-mapping-tools.org>`__.
 
-We need the very latest GMT since there are many changes being made to GMT itself in
+We need the latest GMT since there are many changes being made to GMT itself in
 response to the development of PyGMT, mainly the new
-`modern execution mode <https://gmt.soest.hawaii.edu/projects/gmt/wiki/Modernization>`__.
+`modern execution mode <https://docs.generic-mapping-tools.org/latest/cookbook/introduction.html#modern-and-classic-mode>`__.
 
-Compiled conda packages of GMT for Linux, Mac and Windows are provided through
+Compiled conda packages of GMT for Linux and Mac are provided through
 `conda-forge <https://anaconda.org/conda-forge/gmt>`__.
 Advanced users can also
 `build GMT from source <https://github.com/GenericMappingTools/gmt/blob/master/BUILDING.md>`__
-instead, which is not so recommended but we would love to get feedback from anyone who
-tries.
+instead, which is not so recommended but we would love to get feedback from anyone who tries.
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -35,10 +35,12 @@ PyGMT
...
 
-🚨 **This package is in the early stages of design and implementation.** 🚨
+🚨 **This package is still undergoing rapid development.** 🚨
 
-All functions/classes/interfaces are subject to change as we experiment with new design
-ideas and implement new features. **This is NOT a finished product.**
+All of the API (functions/classes/interfaces) is subject to change until we reach v1.0.0
+as per the `semantic versioning specification <https://semver.org/spec/v2.0.0.html>`__.
+There may be non-backward compatible changes as we experiment with new design ideas and
+implement new features. **This is not a finished product, use with caution**
```
leouieda, line 43: :+1: nice, I like this! Though most people will ignore it and complain when we break things...

doc/install.rst
```diff
@@ -32,16 +32,18 @@ PyGMT requires Generic Mapping Tools (GMT) version 6 as a minimum, which is the
 released version that can be found at
 `this website <https://www.generic-mapping-tools.org>`__.
 
-We need the very latest GMT since there are many changes being made to GMT itself in
+We need the latest GMT since there are many changes being made to GMT itself in
```
leouieda, line 35:
```suggestion
We need the latest GMT (>=6.0.0) since there are many changes being made to GMT itself in
```

doc/install.rst
```diff
@@ -32,16 +32,18 @@ PyGMT requires Generic Mapping Tools (GMT) version 6 as a minimum, which is the
...
+Compiled conda packages of GMT for Linux and Mac are provided through
 `conda-forge <https://anaconda.org/conda-forge/gmt>`__.
 Advanced users can also
 `build GMT from source <https://github.com/GenericMappingTools/gmt/blob/master/BUILDING.md>`__
-instead, which is not so recommended but we would love to get feedback from anyone who
-tries.
+instead, which is not so recommended but we would love to get feedback from anyone who tries.
+For Windows, conda GMT packages are available, but they do not currently work with PyGMT,
+so users will need to build from source or use the Windows Subsystem for Linux, see
+`here <https://github.com/GenericMappingTools/pygmt/pull/313>`__ for more details.
```
leouieda, line 46: I'm wondering if it's time we removed the GMT 6 install instructions from here. It's a workaround that was necessary but might not be any more. There shouldn't be problems installing PyGMT on the base conda environment. 

Maybe instead of removing, we could move it further down and label it "If you have trouble install GMT 6" or something like that. What do you think?

doc/install.rst
```diff
@@ -96,15 +98,13 @@ Installing PyGMT
...
 
-Alternatively, you can clone the git repository and install using ``pip``::
+Alternatively, you can install the development version from the Github repository::
 
-    git clone https://github.com/GenericMappingTools/pygmt.git
-    cd pygmt
-    pip install .
+    pip install https://github.com/GenericMappingTools/pygmt/archive/master.zip
 
 This will allow you to use the ``pygmt`` library from Python.
```
leouieda, line 109: I guess we can leave the conda-forge instructions for later since I'm not entirely confident that it will work easily. So we can:

1.  release 0.1.0 without conda instructions
2. build the  conda-forge feedstock and make sure everything works
3. release 0.1.1 with only the docs changes or some minor fixes that may be required

This shouldn't be much work since most of the release is automated anyway.

weiji14: Was thinking along the same lines too. [Murphy's Law](https://en.wikipedia.org/wiki/Murphy's_law) might strike and I think we should leave it to the 0.1.1 release.

doc/install.rst
weiji14: It's probably better to keep it near the beginning for now, just so users (especially Windows ones) are aware of potential problems up front. The conda-forge `pygmt` package might only work for Linux/MacOS initially, and we need to remember that WIndows still takes up ~90% of the world's operating system market share. It'll be more frustrating for someone to have gone through installing `gmt` and `pygmt` from conda-forge and realize that it doesn't work on WIndows because they missed the fine print at the bottom.

PR #424, opened 2020-05-07
Remind contributors to select the correct branch for PRs

seisman:

New features should be merged into master branch, and bug fixes for 0.1.0 should be merged into 0.1 branch.

We also need to discuss how to merge bugfix commits into master branch, just like what we already did in the GMT main repository (https://github.com/GenericMappingTools/gmt/issues/1032).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,6 +10,7 @@ Fixes #
 
 **Reminders**
 
+- [ ] Correct base branch selected? Choose `master` branch for new features, `0.x` branch (e.g., `0.1`) for bug fixes.
 - [ ] Run `make format` and `make check` to make sure the code follows the style guide.
 - [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
 - [ ] Add new public functions/methods/classes to `doc/api/index.rst`.


seisman:
> The `cherry-pick` technique in [GenericMappingTools/gmt#1032](https://github.com/GenericMappingTools/gmt/issues/1032) is definitely the way to go.

`cherry-pick` gives a cleaner commit history. My only concern is that it's really easy to forget to do the cherry-pick.

weiji14:
> My only concern is that it's really easy to forget to do the cherry-pick.

Ok, I've ticked the 'require linear history' setting for the 'master' branch. That should keep things clean by preventing merge commits (i.e. only cherry-picks/squash merges allowed).

Require linear history setting

seisman:
> > My only concern is that it's really easy to forget to do the cherry-pick.
> 
> Ok, I've ticked the 'require linear history' setting for the 'master' branch. That should keep things clean by preventing merge commits (i.e. only cherry-picks/squash merges allowed).
> 
> Require linear history setting

That's not exactly my point. For example, if a bug-fix PR is merged into branch 0.1, we should cherry-pick it into the master branch. However, we may forget to cherry-pick it, then the bug will be fixed in 0.1.1 release, but not in the 0.2.0 release. 

I'm not sure if we have a mechanism to make sure we never forget to cherry-pick a commit/PR.

weiji14:
Ah ok. That was why I thought about the backporting route (i.e. commit to `master` and then cherry-pick to `0.1`). That way everything will definitely go into v0.2.0.

There's actually a lot of CLI tools to do this, and I'm just looking up on Github bots that can automate this sort of workflow. See e.g. discussion at https://github.com/python/core-workflow/issues/8, and an example of a bot at https://github.com/python/miss-islington.

leouieda:
@seisman @weiji14 not sure I see a need to keep things separate at this time. We're not at a stable point and there is no promise of backward compatibility. Keeping separate branches creates more maintenance overhead for little gain right now.

I would go for merging everything into master. Then when a release comes we can think if we need to do 0.2.0 or 0.1.x depending on what was merged.

leouieda:
And we definitely *will* break compatibility. The current design is flaky and will not work in the long term. It suffers from the same limitations as GMT (difficult to maintain, hard to find documentation, cryptic syntax) and we can surely do better.

seisman:
I'm OK with merging everything to master branch. 

This PR is no longer needed.

Issue #366, opened 2019-11-04
pygmt does not display images using show()


MarkWieczorek:
gmt version = 6.0.0 (macOS, installed via brew)

pygmt does not display newly created figures using `show()`.

```
In [4]: fig = pygmt.Figure()
   ...: fig.coast(shorelines=True, region=[-90, -70, 0, 20], projection="M8i", frame=True)
   ...: fig.show()
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
Out[4]: <IPython.core.display.Image object>
```

MarkWieczorek:
I note that this does in fact work in a jupyter notebook, but not in a standard terminal session.

seisman:
In a IPython terminal, you need to call `fig.show(method="external")` to display the figure.

MarkWieczorek:
That appears to save the figure as a temporary file, and then opens it with an external application.

Would it be possible to
1. use a syntax that is compatible with matplotlib, or
2. plot the pygmt figure inside of a matplotlib figure or axis?

What I would really like to do is have a simple way of choosing to plot something using either standard matplotlib or gmt, such as:

```
fig, ax = plt.subplots(1,1)
myplot(data, ax=ax, format='gmt')
myplot(data, ax=ax, format='mpl')
```

weiji14:
Hi @MarkWieczorek, I'm not sure if there's a way to put pygmt figures inside matplotlib (point 2), but in regard to a matplolib compatible syntax (point 1), @leouieda definitely put some thought before on how to make the display mechanisms better at #269.

If it's subplots you're interested in, the modern mode in GMT 6 does have a [subplot](https://docs.generic-mapping-tools.org/latest/subplot.html) feature, but that's not wrapped yet. I actually have some local code to implement that but it'll take time to get into PyGMT properly.

Could you describe in a bit more detail what you're trying to plot? If I'm reading this correctly, are you interested in showing the figure inline inside of an IPython terminal?

[...]

PR #427, opened 2020-05-13
Experimental subplot implementation that mimics matplotlib's style

### LINKED ISSUE #20: Implement subplot command

weiji14:

Experimenting with a more matplotlib-like syntax for managing subplots (an offshoot from #412). Preview of tutorial is at https://pygmt-git-fancysubplots.gmt.now.sh/tutorials/subplots.html, which is actually translated and adapted from the great example at https://docs.gmt-china.org/latest/tutorial/subplot/ (in Chinese). Specifically, this mimics the [`fig, axs = plt.subplots()` style](https://stackoverflow.com/a/55771757/6611055) of defining subplots! An example of how it looks like currently:

```python
import pygmt

fig, axs = pygmt.subplots(nrows=1, ncols=2, figsize=("6c", "3c"))
fig.sca(ax=axs[0, 0])
fig.sca(ax=axs[0, 1])
fig.end_subplot()
fig.show()

# or

fig, axs = pygmt.subplots(nrows=1, ncols=2, figsize=("6c", "3c"))
fig.basemap(region=[0, 3, 0, 3], frame=True, ax=axs[0, 0])
fig.basemap(region=[0, 3, 0, 3], frame=True, ax=axs[0, 1])
fig.end_subplot()
fig.show()
```

produces:

Basic Subplot

The code isn't perfect yet, and I'm very open to suggestions on how to go about improving this implementation. Specifically, it would be cool if we could do `ax.plot()` like in matplotlib. Calling expert Pythonistas for tips :wave:

As it currently stands, there's a high-level `subplots` function in subplot.py (somewhat based on the matplotlib `plt.subplots` equivalent at https://github.com/matplotlib/matplotlib/blob/v3.2.1/lib/matplotlib/pyplot.py#L1044). This wraps around the new class SubPlot (inherited from the main Figure class to include subplot functionality). Note that since `set` is a Python reserved character so I've used `sca` instead to set the current active subplot, following matplotlib.

Parameters/Aliases to wrap:

- [x] F = dims/dimensions/size/sizes
- [x] A = autolabel
- [x] B = frame
- [x] C = clearance
- [ ] J = projection
- [x] M = margins
- [x] R = region
- [x] S (C/R) = layout
- [x] T= title

References:
- https://matplotlib.org/3.2.1/gallery/subplots_axes_and_figures/subplots_demo.html
- https://matplotlib.org/3.2.1/tutorials/introductory/pyplot.html
- https://stackoverflow.com/a/55771757/6611055
- https://matplotlib.org/api/_as_gen/matplotlib.pyplot.sca.html
- https://nbviewer.jupyter.org/github/SHTOOLS/SHTOOLS/blob/master/examples/notebooks/plotting-maps.ipynb#Creating-Subplots
- https://basemaptutorial.readthedocs.io/en/latest/subplots.html

TODO:
- [x] Initial implementation
- [x] Alias `ax` to the [`-c`](https://docs.generic-mapping-tools.org/latest/gmt.html#c-full) command in GMT plotting functions
- [ ] Find a way to call `ax[0, 1].plot()` directly instead of using `fig.sca(ax)`.

Addresses https://github.com/GenericMappingTools/pygmt/issues/366#issuecomment-549386681, Supersedes and Closes #412, Fixes #20.

**Reminders**


leouieda:
Hi Wei Ji, thanks for doing this! I’ve been looking through this a bit now. I like the subplots function! But I think we shouldn’t do sca as well. A major fault of matplotlib is having too many ways of doing the same thing. We can stick close to their api but need to be careful not to repeat their mistakes.

weiji14:
> But I think we shouldn’t do sca as well. A major fault of matplotlib is having too many ways of doing the same thing. We can stick close to their api but need to be careful not to repeat their mistakes.

Yes, the good ol [Zen of Python](https://www.python.org/dev/peps/pep-0020/) - "There should be one-- and preferably only one --obvious way to do it.". I definitely prefer using `ax=` to set the current subplot.

However, the `sca` function (which is basically an alias for GMT subplot's [`set`](https://docs.generic-mapping-tools.org/latest/subplot.html#synopsis-set-mode)) would be helpful say, if we want to label our subplots "abcd" or set a clearance/space between subplots. Would there be a better way to do those without resorting to using `sca`?

weiji14:
Ping @MarkWieczorek, I noticed that SHTOOLS has a tutorial on creating subplots using PyGMT's using X and Y offsets at https://nbviewer.jupyter.org/github/SHTOOLS/SHTOOLS/blob/master/examples/notebooks/plotting-maps.ipynb#Creating-Subplots. Right now this PR is attempting to have a syntax like `fig, ax = pygmt.subplots(nrows=1, ncols=2)` which could make it easier to create subplots in the future (i.e. without relying on X and Y offsets).

Would be great to get hear your thoughts on this, e.g. would you like to do `grid.plotgmt(..., ax=ax[0])` in SHTOOLS? Let us know if there's anything in particular from `matplotlib` you would like or not like to have implemented here :grin:

MarkWieczorek:
This looks good to me! I would be happy to test this with shtools, but this will have to wait until mid october....

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,224 @@
...
+
+###############################################################################
+# For example, to activate the subplot on the top right corner (index: 2) at
+# ``row=0`` and ``col=2``, so that all subsequent plotting commands happen
+# there, you can use the following command:
+
+###############################################################################
+# .. code-block:: default
+#
+#     fig.sca(ax=axs[0, 2])
```
weiji14, line 77: Note to self to remove mention of `fig.sca`, if we're going to hide this away as per https://github.com/GenericMappingTools/pygmt/pull/427#issuecomment-646983555.

weiji14 (commented):
Alright, I'm marking this as ready to review, though there are still some rough spots. The translation over from https://docs.gmt-china.org/latest/tutorial/subplot/ is complete, and I think it's worth the team taking a look. Let me know if I should cherry-pick stuff out into separate PRs to make it easier to review.

Note that there are some functionality from https://docs.generic-mapping-tools.org/latest/subplot.html that aren't covered yet:
- `-Fs` (subfigsize?) at https://docs.generic-mapping-tools.org/latest/subplot.html#f
- The matplotlib way of plotting using something like `ax[0, 1].plot()` - quite tricky to implement!!

PR #239, opened 2018-11-13
Explicitly install netCDF4

seisman:
**Description of proposed changes**

conda-forge/xarray-feedstock [drops optional dependencies](https://github.com/conda-forge/xarray-feedstock/pull/37) several days ago. Thus we need install netCDF4 explicitly.

Fixes #238

**Reminders**

- [X] Run `make format` and `make check` to make sure the code follows the style guide.
- [X] Add tests for new features or tests that would have caught the bug that you're fixing.
- [X] Add new public functions/methods/classes to `doc/api/index.rst`.
- [X] Write detailed docstrings for all functions/methods.
- [X] If adding new functionality, add an example to docstrings or tutorials.

PR #429, opened 2020-05-17
Add netCDF4 as a dependency and install netcdf4

seisman:

xarray needs the netCDF4 package for reading (perhaps also writing) netCDF files. However, netCDF4 isn't installed by default since it's an optional dependency of xarray. Although pygmt doesn't import netCDF4 package explicitely, some features won't work without the package.

See #239 for past discussions.

**Reminders**

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -55,6 +55,7 @@ PyGMT requires the following libraries:
 * `numpy <http://www.numpy.org/>`__
 * `pandas <https://pandas.pydata.org/>`__
 * `xarray <http://xarray.pydata.org/>`__
+* `netCDF4 <https://github.com/Unidata/netcdf4-python>`__
 * `packaging <https://pypi.org/project/packaging/>`__
 
 The following are optional (but recommended) dependencies:
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -55,12 +55,12 @@ PyGMT requires the following libraries:
 * `numpy <http://www.numpy.org/>`__
 * `pandas <https://pandas.pydata.org/>`__
 * `xarray <http://xarray.pydata.org/>`__
-* `netCDF4 <https://github.com/Unidata/netcdf4-python>`__
 * `packaging <https://pypi.org/project/packaging/>`__
 
 The following are optional (but recommended) dependencies:
 
 * `IPython <https://ipython.org/>`__: For embedding the figures in Jupyter notebooks.
+* `netCDF4 <https://github.com/Unidata/netcdf4-python>`__: For reading netCDF4 data via `xarray <http://xarray.pydata.org/>`__
 
 
 Installing GMT and other dependencies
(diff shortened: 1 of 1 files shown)


doc/install.rst
```diff
@@ -79,7 +80,7 @@ First, we must configure conda to get packages from the
 Now we can create a new conda environment with Python and all our dependencies installed
 (we'll call it ``pygmt`` but you can change it to whatever you want)::
 
-     conda create --name pygmt python=3.6 pip numpy pandas xarray packaging gmt=6.0.0
+     conda create --name pygmt python=3.6 pip numpy pandas xarray netcdf4 packaging gmt=6.0.0
```
weiji14, line 83: Once #430 is merged, would it be better to just instruct users to `conda install pygmt` since it'll pull in all the side dependencies as well?

weiji14:
> But I wonder if there are advanced users that might not want to install it because of GPL issues (see https://github.com/intake/intake-xarray/issues/41), though that would need to fixed on our setup.py.

Then again, GMT is LGPL licensed, but PyGMT is BSD3 licensed, so I'm not sure if it matters. I'm not a lawyer though.

doc/install.rst
seisman: Yes, perhaps we should merge https://github.com/conda-forge/pygmt-feedstock/pull/2 first, then instruct users to install pygmt with:
```
conda create --name pygmt pygmt
```

weiji14:
> Yeah. It falls back to a scipy netcdf3 implementation but the GMT files don't conform to that style so no luck.

> _Originally posted by @leouieda in https://github.com/GenericMappingTools/pygmt/pull/239#issuecomment-438473652

Just a note that `xarray.open_dataset` supports other 'engines' besides `netcdf4` and `scipy`. See http://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html?highlight=engine. For example, I've often used the [`h5netcdf`](https://github.com/shoyer/h5netcdf) engine which is BSD3 licensed. We might want to look into supporting those in the future (via some setting, or as a kwargs type input) as an alternative to non-`netcdf4` users, just not now in this PR :smile:

seisman:
@weiji14 I just moved netcdf4 to an optional dependency, and remove it from the conda install command.

weiji14:
If we're going to put this as optional, then maybe remove it from `setup.py` too?

seisman:
> If we're going to put this as optional, then maybe remove it from `setup.py` too?

I personally prefer to have all dependencies installed. 

AFAIK, `pygmt.datasets.load_earth_relief()` needs netCDF4 for working. Since this function is used in the tests and the tutorial, it's better to have it installed, especially for new Python users.

weiji14:
Yeah, better to have it as a non-optional dependency then. We could have a `pip install pygmt[minimal]` setup in the future if anyone really doesn't want `netcdf`.

seisman:
> Yeah, better to have it as a non-optional dependency then. We could have a `pip install pygmt[minimal]` setup in the future if anyone really doesn't want `netcdf`.

Done.

seisman:
Merged into PR #430. Closing.

Issue #365, opened 2019-11-04
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.


MarkWieczorek:
GMT version = 6.0.0 (macOS installed via brew)

pygmt complains about the `GMT_COMPATIBILITY` parameter. By default, this is set to 4, but pygmt expects 6. Setting `GMT_COMPATIBILITY = 6` in my root `gmt.conf` file removes this warning, but setting this parameter to 6 should rather be done locally by pygmt itself.

```
In [3]: fig = pygmt.Figure()
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
```

seisman:
Any reason you want to have a "gmt.conf" file in you home directory? The simplest way is to remove  the "root" `gmt.conf`.

MarkWieczorek:
Historically, gmt has always had a default configuration file, which as of version 6 should be located in `.gmt`. I note that gmt v6 does not itself create a default configuration file for you, but the documentation tells you how to do this, and I presume that everyone will in fact do this. This default configuration is most often used to set high level defaults, such as the paper size (which differs between the US and Europe), as well as the preferred measurement units (SI or english).

It doesn't appear to be a good idea to ask the user to delete the default gmt configuration file just to suppress pygmt warnings.

seisman:
This issue has been fixed in upstream GMT codes in https://github.com/GenericMappingTools/gmt/pull/1981.

weiji14:
> This issue has been fixed in upstream GMT codes in [GenericMappingTools/gmt#1981](https://github.com/GenericMappingTools/gmt/pull/1981).

Closing this issue as it should be resolved. If anyone else has this problem, let us know and we can reopen it if necessary.

PR #432, opened 2020-05-17
Set GMT_COMPATIBILITY to 6 when pygmt session starts

### LINKED ISSUE #428: pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
pygmt continues to provide annoying errors about GMT_COMPATIBILITY. These can arise in many circumstances (such as in jupyter notebooks), but here is the easiest one:

```
In [1]: import pygmt
In [2]: exit
pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
>  gmt --version
6.0.0
> brew info gmt
gmt: stable 6.0.0 (bottled), HEAD
Tools for manipulating and plotting geographic and Cartesian data
https://www.generic-mapping-tools.org/
/usr/local/Cellar/gmt/6.0.0_5 (1,584 files, 150.6MB) *
```
OS: macos
pygmt: v0.1 (installed via pypi)
gmt: installed by brew (v. 6.0.0_5)
python: 3.7.7

weiji14:

To prevent users getting "pygmt-session [ERROR]: GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6". Overrides [GMT_COMPATIBILITY](https://docs.generic-mapping-tools.org/6.0/gmt.conf.html#gmt-miscellaneous-parameters) setting in users `gmt.conf` file.

Fixes #428, #365

**Reminders**

### DIFF
--- a/pygmt/session_management.py
+++ b/pygmt/session_management.py
@@ -15,6 +15,7 @@ def begin():
     prefix = "pygmt-session"
     with Session() as lib:
         lib.call_module("begin", prefix)
+        # pygmt relies on GMT modern mode with GMT_COMPATIBILITY at version 6
         lib.call_module("set", "GMT_COMPATIBILITY 6")
 
 
--- a/pygmt/session_management.py
+++ b/pygmt/session_management.py
@@ -15,6 +15,7 @@ def begin():
     prefix = "pygmt-session"
     with Session() as lib:
         lib.call_module("begin", prefix)
+        lib.call_module("set", "GMT_COMPATIBILITY 6")
 
 
 def end():
--- a/pygmt/tests/test_session_management.py
+++ b/pygmt/tests/test_session_management.py
@@ -48,7 +48,5 @@ def test_gmt_compat_6_is_applied(capsys):
             lib.call_module("set", "GMT_COMPATIBILITY 6")
         end()
         begin()  # Restart the global session
-        assert os.path.exists("gmt.conf")
-        os.remove("gmt.conf")
         assert os.path.exists("pygmt-session.pdf")
         os.remove("pygmt-session.pdf")
--- a/pygmt/tests/test_session_management.py
+++ b/pygmt/tests/test_session_management.py
@@ -20,3 +20,35 @@ def test_begin_end():
     begin()  # Restart the global session
     assert os.path.exists("pygmt-session.pdf")
     os.remove("pygmt-session.pdf")
+
+
+def test_gmt_compat_6_is_applied(capsys):
+    """
+    Ensure that users with old gmt.conf files won't get pygmt-session [ERROR]:
+    GMT_COMPATIBILITY: Expects values from 6 to 6; reset to 6.
+    """
+    end()  # Kill the global session
+    try:
+        with Session() as lib:
+            # pretend that gmt.conf has GMT_COMPATIBILITY = 5
+            lib.call_module("gmtset", "GMT_COMPATIBILITY 5")
+        begin()
+        with Session() as lib:
+            lib.call_module("basemap", "-R10/70/-3/8 -JX4i/3i -Ba")
+            out, err = capsys.readouterr()  # capture stdout and stderr
+            assert out == ""
+            assert err != (
+                "pygmt-session [ERROR]: GMT_COMPATIBILITY:"
+                " Expects values from 6 to 6; reset to 6.\n"
+            )
+            assert err == ""  # double check that there are no other errors
+    finally:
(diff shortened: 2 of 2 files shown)


pygmt/session_management.py
```diff
@@ -15,6 +15,7 @@ def begin():
     prefix = "pygmt-session"
     with Session() as lib:
         lib.call_module("begin", prefix)
+        lib.call_module("set", "GMT_COMPATIBILITY 6")
 
 
 def end():
```
weiji14, line 21: Can't find a good way to reset the `GMT_COMPATIBILITY` setting back to what's in the original `gmt.conf`. Probably doesn't matter too much?

pygmt/tests/test_session_management.py
```diff
@@ -20,3 +20,35 @@ def test_begin_end():
...
+    try:
+        with Session() as lib:
+            # pretend that gmt.conf has GMT_COMPATIBILITY = 5
+            lib.call_module("gmtset", "GMT_COMPATIBILITY 5")
+        begin()
+        with Session() as lib:
+            lib.call_module("basemap", "-R10/70/-3/8 -JX4i/3i -Ba")
+            out, err = capsys.readouterr()  # capture stdout and stderr
+            assert out == ""
+            assert err != (
```
weiji14, line 40: Note: To future maintainers on what `capsys` is. It's a pytest fixture for capturing stdout/stderr. See https://docs.pytest.org/en/5.4.2/capture.html#accessing-captured-output-from-a-test-function

pygmt/session_management.py
seisman: See https://github.com/GenericMappingTools/pygmt/issues/428#issuecomment-629809995

pygmt/session_management.py
```diff
@@ -15,6 +15,7 @@ def begin():
     prefix = "pygmt-session"
     with Session() as lib:
         lib.call_module("begin", prefix)
+        lib.call_module("set", "GMT_COMPATIBILITY 6")
```
seisman, line 18: Perhaps add a comment here why pygmt needs GMT_COMPATIBILITY=6.

seisman, line 18:
```suggestion
        lib.call_module("begin", prefix)
        # pygmt replies on GMT modern mode, which requires GMT_COMPATIBILITY to 6
        lib.call_module("set", "GMT_COMPATIBILITY 6")
```

pygmt/tests/test_session_management.py
```diff
@@ -20,3 +20,35 @@ def test_begin_end():
...
+            )
+            assert err == ""  # double check that there are no other errors
+    finally:
+        with Session() as lib:
+            # revert gmt.conf back to GMT_COMPATIBILITY = 6
+            lib.call_module("set", "GMT_COMPATIBILITY 6")
+        end()
+        begin()  # Restart the global session
+        assert os.path.exists("gmt.conf")
+        os.remove("gmt.conf")
```
seisman, line 52: I don't think there is a "gmt.conf" file in the current directory. It should be in the temporary gmt session directory.

weiji14: I kinda agree, but wouldn't the assertion fail if there isn't a `gmt.conf` file? :confused: Things got really confusing for me when I was designing this test with `gmt.conf` files all over the place, one kept popping up in my `pygmt` folder with `GMT_COMPATIBILITY 5` for some reason.

weiji14:
@MarkWieczorek, do you want to test this branch (using `pip install https://github.com/GenericMappingTools/pygmt/archive/gmt_compat_6.zip`)? Just want to make sure that it fixes your issue.

MarkWieczorek:
It works!

pygmt/tests/test_session_management.py
weiji14: Ok, you're right, it's in "/home/.../pygmt/tmp-test-dir-with-unique-name/gmt.conf". I'll remove these lines once the tests on Windows pass in the other PR.

seisman: I still don't understand why the test works, but since the test pass, OK to merge.

weiji14: Yeah, it's a complicated test. but removing the line at https://github.com/GenericMappingTools/pygmt/blob/8df13da7d7163ed265c22b7de18e1df344ec3664/pygmt/session_management.py#L19 should fail this test, as encountered by Mark.

Issue #436, opened 2020-05-19
Code coverage reports are broken


seisman:
**Description of the problem**

The code coverage reports failed to work recently. See https://codecov.io/gh/GenericMappingTools/pygmt/pulls for recent reports.

The CI reports the following errors:
```
+ codecov -e PYTHON AGENT_OS
Uploading coverage to Codecov
find: -not -path './bower_components/**' -not -path './node_modules/**' -not -path './vendor/**': unknown primary or operator

      _____          _
     / ____|        | |
    | |     ___   __| | ___  ___ _____   __
    | |    / _ \ / _  |/ _ \/ __/ _ \ \ / /
    | |___| (_) | (_| |  __/ (_| (_) \ V /
     \_____\___/ \____|\___|\___\___/ \_/
                                    v2.0.20

==> Detecting CI provider
  -> Got branch from git/hg
  -> Got sha from git/hg
==> Preparing upload
==> Processing gcov (disable by -X gcov)
    Executing gcov (['find', '/Users/runner/runners/2.168.2/work/1/s', "-not -path './bower_components/**' -not -path './node_modules/**' -not -path './vendor/**'", '-type', 'f', '-name', '*.gcno', '', '-exec', 'gcov', '-pb', '', '{}', '+'])
    Error running `['find', '/Users/runner/runners/2.168.2/work/1/s', "-not -path './bower_components/**' -not -path './node_modules/**' -not -path './vendor/**'", '-type', 'f', '-name', '*.gcno', '', '-exec', 'gcov', '-pb', '', '{}', '+']`: Command '['find', '/Users/runner/runners/2.168.2/work/1/s', "-not -path './bower_components/**' -not -path './node_modules/**' -not -path './vendor/**'", '-type', 'f', '-name', '*.gcno', '', '-exec', 'gcov', '-pb', '', '{}', '+']' returned non-zero exit status 1.
==> Collecting reports
    + /Users/runner/runners/2.168.2/work/1/s/coverage.xml bytes=39890
==> Appending environment variables
    + PYTHON
    + AGENT_OS
==> Uploading
    .url https://codecov.io
    .query commit=9f52c71854c719b33b774deb6b17648f677153e8&branch=HEAD&token=<secret>&yaml=.codecov.yml&package=py2.0.20
    Pinging Codecov...
Error: HTTP 400
Provided token is not a UUID.
```

weiji14:
[...]

PR #440, opened 2020-05-20
Improve how PyGMT finds the GMT library

seisman:

On Windows, ctypes can't find the gmt library in conda's path. The reason is still unclear to me. But ctypes provides a function [find_library](https://docs.python.org/3/library/ctypes.html#ctypes.util.find_library) which search DLLs in the system search path (i.e., the variable **PATH**). 

Thus, on Windows, if **GMT_LIBRARY_PATH** is not defined, we first call `find_library` to check if gmt.dll, gmt_w64 or gmt_w32 exists in one of the PATH. `find_library` returns its full path (e.g., `C:\Miniconda\envs\testing\Library\bin\gmt.dll`) if it's found, else returns None.

The new function `clib_full_names` returns the list of full names to the gmt library. When **GMT_LIBRARY_PATH** is not defined, it returns 
- macOS: `["libgmt.dylib"]` 
- Linux: `["libgmt.so"]`
- Windows: `["gmt.dll", "gmt_w64.dll", "gmt_w32.dll", "C:\Miniconda\envs\testing\Library\bin\gmt.dll"]`.

On Windows, the full path returned by `find_library` is placed at the end of the list, to better mimic the [library search order on Windows](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order?redirectedfrom=MSDN#standard-search-order-for-desktop-applications).

With this PR merged, the variables **GMT_LIBRARY_PATH** is no longer required for Windows users.

**Reminders**

### DIFF
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -51,7 +51,7 @@ def load_libgmt():
     return libgmt
 
 
-def clib_name(os_name):
+def clib_names(os_name):
     """
     Return the name of GMT's shared library for the current OS.
 
@@ -62,20 +62,20 @@ def clib_name(os_name):
 
     Returns
     -------
-    libname : list of str
+    libnames : list of str
         List of possible names of GMT's shared library.
 
     """
     if os_name.startswith("linux"):
-        libname = ["libgmt.so"]
+        libnames = ["libgmt.so"]
     elif os_name == "darwin":
         # Darwin is macOS
-        libname = ["libgmt.dylib"]
+        libnames = ["libgmt.dylib"]
     elif os_name == "win32":
-        libname = ["gmt.dll", "gmt_w64.dll", "gmt_w32.dll"]
+        libnames = ["gmt.dll", "gmt_w64.dll", "gmt_w32.dll"]
     else:
         raise GMTOSError('Operating system "{}" not supported.'.format(sys.platform))
-    return libname
+    return libnames
 
 
 def clib_full_names(env=None):
@@ -96,7 +96,7 @@ def clib_full_names(env=None):
     """
     if env is None:
         env = os.environ
-    libnames = clib_name(os_name=sys.platform)  # e.g. libgmt.so, libgmt.dylib, gmt.dll
+    libnames = clib_names(os_name=sys.platform)  # e.g. libgmt.so, libgmt.dylib, gmt.dll
     libpath = env.get("GMT_LIBRARY_PATH", "")  # e.g. $HOME/miniconda/envs/pygmt/lib
 
     lib_fullnames = [os.path.join(libpath, libname) for libname in libnames]
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -11,7 +11,7 @@ import ctypes
 from ..exceptions import GMTOSError, GMTCLibError, GMTCLibNotFoundError
 
 
-def load_libgmt(env=None):
+def load_libgmt():
     """
     Find and load ``libgmt`` as a :py:class:`ctypes.CDLL`.
 
@@ -37,14 +37,11 @@ def load_libgmt(env=None):
         couldn't access the functions).
 
     """
-    if env is None:
-        env = os.environ
-    libnames = clib_name(os_name=sys.platform)
-    libpath = env.get("GMT_LIBRARY_PATH", "")
+    lib_fullnames = clib_full_names()
     error = True
-    for libname in libnames:
+    for libname in lib_fullnames:
         try:
-            libgmt = ctypes.CDLL(os.path.join(libpath, libname))
+            libgmt = ctypes.CDLL(libname)
             check_libgmt(libgmt)
             error = False
             break
@@ -52,7 +49,7 @@ def load_libgmt(env=None):
             error = err
     if error:
         raise GMTCLibNotFoundError(
-            "Error loading the GMT shared library '{}':".format(", ".join(libnames))
+            "Error loading the GMT shared library '{}':".format(", ".join(lib_fullnames))
         )
     return libgmt
 
@@ -82,6 +79,16 @@ def clib_name(os_name):
     else:
         raise GMTOSError('Operating system "{}" not supported.'.format(sys.platform))
     return libname
+
+
+def clib_full_names(env=None):
+    if env is None:
+        env = os.environ
+    libnames = clib_name(os_name=sys.platform)
+    libpath = env.get("GMT_LIBRARY_PATH", "")
+
+    lib_fullnames = [os.path.join(libpath, libname) for libname in libnames]
+    return lib_fullnames
 
 
 def check_libgmt(libgmt):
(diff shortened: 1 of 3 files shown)


weiji14:
Ok, just woke up and finished breakfast. Just to make sure we're on the same page, we'll be releasing all of these 'Windows' fixes into `v0.1.1`, and not `v0.2.0` yes?

seisman:
Yes, v0.1.0 is good. Perhaps easier to rename the current 0.2.0 milestone to 0.1.1.

weiji14:
Or we could rename the milestones to `0.1.x` and `0.2.x`?

seisman:
@weiji14 I think we finally have a good solution for the Windows issue. No workaround is needed anymore. Please review the PR. I bet more tests should be added, but I can't see the [coverage report](#436).

weiji14:
Awesome, give me a while, there's a lot to review... Not sure what's up with codecov but we can use `make test` to see the test coverage locally for now.

pygmt/clib/loading.py
```diff
@@ -37,22 +32,21 @@ def load_libgmt(env=None):
...
         try:
-            libgmt = ctypes.CDLL(os.path.join(libpath, libname))
+            libgmt = ctypes.CDLL(libname)
             check_libgmt(libgmt)
             error = False
             break
         except OSError as err:
             error = err
     if error:
         raise GMTCLibNotFoundError(
```
weiji14, line 46: Test coverage missing for this section (Lines 43,44,46). Will need to edit and add back the deleted `test_load_libgmt_fail`.

pygmt/clib/loading.py
```diff
@@ -84,6 +78,37 @@ def clib_name(os_name):
...
+    Returns
+    -------
+    lib_fullnames: list of str
+        List of possible full names of GMT's shared library.
+
+    """
+    if env is None:
+        env = os.environ
+    libnames = clib_name(os_name=sys.platform)
+    libpath = env.get("GMT_LIBRARY_PATH", "")
```
weiji14, line 100: I always found this section hard to follow, maybe add some examples here?
```suggestion
    libnames = clib_name(os_name=sys.platform)  # e.g. libgmt.so, libgmt.dylib, gmt.dll
    libpath = env.get("GMT_LIBRARY_PATH", "")  # e.g. $HOME/miniconda/envs/pygmt/lib
```

pygmt/clib/loading.py
```diff
@@ -84,6 +78,37 @@ def clib_name(os_name):
...
+
+    lib_fullnames = [os.path.join(libpath, libname) for libname in libnames]
+    # Search for DLLs in PATH if GMT_LIBRARY_PATH is not defined [Windows only]
+    if not libpath and sys.platform == "win32":
+        for libname in libnames:
+            libfullpath = find_library(libname)
+            if libfullpath:
+                lib_fullnames.append(libfullpath)
```
weiji14, line 108: These lines were missing on my test coverage (since I'm on Linux) so I can't check this properly. But looking at the Azure Pipelines test on Windows, it seems okay.

pygmt/clib/loading.py
seisman: The `load_libgmt` function won't fail, as I removed the parameter *env* from this function. Perhaps I should add *env* back, so we can pass a "fake" _env_?

pygmt/clib/loading.py
seisman: Is it OK to add a Windows-only test?

pygmt/clib/loading.py
weiji14: Maybe remove the `try... except` clause then since there won't be an OSError?

seisman:
> I think we can also improve the documentation section at https://github.com/GenericMappingTools/pygmt/blob/cdll-path/doc/install.rst#finding-the-gmt-shared-library to add instruction on how to set the `GMT_LIBRARY_PATH` on MacOS and Windows (instead of just Linux).

Let's put it in another PR.

pygmt/clib/loading.py
weiji14: Wait, scratch that. If you set a fake `GMT_LIBRARY_PATH` using `os.environ["GMT_LIBRARY_PATH"] = "/not/a/real/path"`, that would be the same as passing in a fake env right?

pygmt/clib/loading.py
```diff
@@ -84,6 +78,37 @@ def clib_name(os_name):
     return libname
```
weiji14, line 78: Should really rename all the `libname` to `libnames` (plural) here in the `clib_name` function since it's a list.
```suggestion
    return libnames
```

pygmt/tests/test_clib_loading.py
```diff
@@ -18,13 +18,6 @@ def test_load_libgmt():
     check_libgmt(load_libgmt())
 
 
-def test_load_libgmt_fail():
-    "Test that loading fails when given a bad library path."
-    env = {"GMT_LIBRARY_PATH": "not/a/real/path"}
-    with pytest.raises(GMTCLibNotFoundError):
-        load_libgmt(env=env)
-
-
```
weiji14, line 27: Maybe something like this (will need to `import os` on the top, and reset GMT_LIBRARY_PATH too)
```python
def test_load_libgmt_fail():
    "Test that loading fails when given a bad library path."
    os.environ["GMT_LIBRARY_PATH"] = "not/a/real/path"
    with pytest.raises(GMTCLibNotFoundError):
        load_libgmt()
```

seisman: Have to save the original value and revert it back after the test.

Fixed in 9251d7c1196a35f6ed9a73718e710fc91952d075.

pygmt/clib/loading.py
weiji14: The `find_library` function is cross platform, I wonder if we should remove the `sys.platform == 'win32'` requirement so that it's easier for Linux/MacOS users too. Comes at the risk of users picking up random `libgmt` libraries from elsewhere though.

pygmt/clib/loading.py
seisman: Done in 4bc0bd754783df22b4fc956da6588be671c53e11. I also renamed the function `clib_name` to `clib_names`.

pygmt/clib/loading.py
seisman: I'm not sure if `find_library` works on macOS. I can import pygmt correctly, which means `ctypes.CDLL("libgmt.dylib") works. However `find_library` returns nothing for me.
```python
In [1]: import pygmt

In [2]: from ctypes.util import find_library

In [3]: find_library("gmt")

In [4]: find_library("libgmt")

In [5]: find_library("libgmt.dylib")

In [6]: find_library("m")
Out[6]: '/usr/lib/libm.dylib'

In [7]: find_library("libm")
Out[7]: '/usr/lib/libm.dylib'

In [8]: find_library("libm.dylib")
Out[8]: '/usr/lib/libm.dylib'
```

weiji14: Could have sworn someone mentioned this before, but a quick scan through the old issues turn up nothing. Should be good to go then, but will keep it in the back of my mind :smile:

PR #454, opened 2020-05-24
Fix the test for GMT_COMPATIBILITY=6

### LINKED ISSUE #453: Test test_gmt_compat_6_is_applied leaves a gmt.conf file in the current directory

To reproduce the bug, run:
```
mkdir tmp-test-dir-with-unique-name
cd tmp-test-dir-with-unique-name
pytest ../pygmt/tests/test_session_management.py
```
then you'll see a "gmt.conf" file in the `tmp-test-dir-with-unique-name` directory.

The "gmt.conf" file is generated by the test "test_gmt_compat_6_is_applied":

https://github.com/GenericMappingTools/pygmt/blob/809a888755c5dedc86ea74d0ae0497295b5311a4/pygmt/tests/test_session_management.py#L30-L34

At the end of the test, Lines 46-49 revert the GMT_COMPATIBILTY to 6. However, these codes only affect the gmt.conf in the session directory. The "gmt.conf" in the current directory is still untouched, with GMT_COMPATIBILITY set to 5.

https://github.com/GenericMappingTools/pygmt/blob/809a888755c5dedc86ea74d0ae0497295b5311a4/pygmt/tests/test_session_management.py#L45-L49

We should remove the "gmt.conf" from the current directory at the end of the test.

seisman:

Clean up the global "gmt.conf" in the current directory, after ending the local session but before beginning the global session.

Fixes #453

**Reminders**


weiji14:
Wasnt't this what https://github.com/GenericMappingTools/pygmt/pull/432#discussion_r426306125 was about... Oh well, I think you've documented it better here though.

Are you sure about removing the `try...except` block? It's more to ensure the file gets deleted *when* the test fails rather than when the test passes.

seisman:
> Are you sure about removing the `try...except` block? It's more to ensure the file gets deleted _when_ the test fails rather than when the test passes.

You're right. I assumed that the test always passes.

pygmt/tests/test_session_management.py
```diff
@@ -43,10 +44,12 @@ def test_gmt_compat_6_is_applied(capsys):
             )
             assert err == ""  # double check that there are no other errors
     finally:
-        with Session() as lib:
-            # revert gmt.conf back to GMT_COMPATIBILITY = 6
-            lib.call_module("set", "GMT_COMPATIBILITY 6")
```
seisman, line 48: These lines are not needed, as they only change the gmt.conf in the session directory, which will be deleted after the `end()` statement.

pygmt/tests/test_session_management.py
```diff
@@ -43,10 +44,12 @@ def test_gmt_compat_6_is_applied(capsys):
...
         end()
-        begin()  # Restart the global session
+        # Clean up the global "gmt.conf" in the current directory
+        assert os.path.exists("gmt.conf")
+        os.remove("gmt.conf")
```
seisman, line 50: Have to remove the "gmt.conf" in the current directory so that it won't affect other tests.

PR #462, opened 2020-05-27
WIP: Add a separate CI job for testing GMT master branch

seisman:

Test GMT master branch.

Fixes #

**Reminders**


pygmt/datasets/earth_relief.py
```diff
@@ -8,7 +8,7 @@
 from ..exceptions import GMTInvalidInput
 
 
-def load_earth_relief(resolution="60m"):
+def load_earth_relief(resolution="01d"):
```
weiji14, line 11: Did the rename occur a while ago? Looking at https://docs.generic-mapping-tools.org/6.0/datasets/earth_relief.html, it seems to be already called '01d' in GMT 6.0?

seisman: It was called 60m at first, then renamed to 01d at least before Jan, 2020 (this is the last build date of the URL you just posted), but 60m is still valid for backward compatibility. 

Paul decided to drop the 60m resolution for GMT 6.1.0, since 60m is not documented.

seisman: BTW, don't take too this PR too seriously. I made some quick changes just to make sure that pygmt doesn't crash with the lastest GMT (see a recent crash and its fix https://github.com/GenericMappingTools/gmt/pull/3514).

weiji14: Yeah I saw the segfault and wondered what was happening. Would be nice to cherry-pick this '60m' -> '01d' renaming commit into a separate PR though, in case this one takes some time to get working properly.

seisman: Yes, we need to do it more carefully. It's a little tricky as we now have so many failing tests.

Issue #147, opened 2018-03-14
Return values of gmt.info


seisman:
This issue is opened to further discuss the https://github.com/GenericMappingTools/gmt-python/pull/106#issuecomment-363649758 about `gmt.info`.

> Do you think we should print the output instead of returning it? There could be a flag to trigger a return instead of printing.

IMHO, returning the output makes more sense than printing it. It's easy for users to print it via `print(gmt.info())` if they need.

> Either way, if the returned string is just a list of numbers, it would be better to return an array/list of the number instead.

As far as I know, the outputs of `gmt info` can be divided into following types when different options are used:

**Type 1**
```
$ gmt info points.txt 
points.txt: N = 20	<11.5309/61.7074>	<-2.9289/7.8648>	<0.1412/0.9338>
```
I think this type is useless in most cases. We can leave it as it is.

**Type 2**
```
$ gmt info points.txt -C
11.5309	61.7074	-2.9289	7.8648	0.1412	0.9338
```
We can convert this type to a list/array.

**Type 3**
```
$ gmt info points.txt -I0.1                                                            
-R11.5/61.8/-3/7.9
```
`-Rw/e/s/n` is useless for Python. Should be convert to a list/array.

**Type 4**
```
$ gmt info points.txt -T0.1
-T11.5/61.8/0.1
```
Useless. Should be convert to a list/array.

**Types 5**
```
$ gmt info points.txt -Ib   
11.5309	-2.9289
61.7074	-2.9289
61.7074	7.8648
11.5309	7.8648
11.5309	-2.9289
```
`gmt info -Ib` reports the bounding box polygon for the data files, which is usually used in further `fig.plot()`. So it may make more sense to return it as two arrays `x` and `y`. A much general way is to return it as a ndarray, so users can access the first column via `output[:,0]` and the ith column via `output[:,i-1]`.

If `-Af|s` option is used, the output becomes more complicated, e.g.:
```
$ gmt info points.txt points-copy.txt -C -Af
11.5309	61.7074	-2.9289	7.8648	0.1412	0.9338
11.5309	61.7074	-2.9289	7.8648	0.1412	0.9338
```
In such cases, we can return the output as a ndarray.

**Conclusion**

What we can do with the `gmt.info` outputs:
 
1. convert it to an array (type 2-4)
2. convert it to a multi-dimensional ndarray (type 2-5 and some more complicated cases).
3. leave the output as it is (type 1 and other cases not discussed in the issue).

[...]

PR #474, opened 2020-06-01
Turn all short aliases into long form

### LINKED ISSUE #473: Only document the long-form arguments

GMT command line and its C API only accept single-character arguments, e.g., **-B** and **-J**, whereas Python and PyGMT users mostly would prefer long-form arguments instead, e.g., **frame** for **-B**, **projection** for **-J**. Currently, PyGMT supports both single-character and long-form arguments, but we definitely will disallow single-character arguments in the future. See issue #262 for the reasons.

Currently, the documentation is pretty confusing. Take the documentation of `Figure.coast()` as an example. Below is what it looks like now:
image

For PyGMT users who never used GMT before, it's not clear what the aliases mean. I believe most of them are confused about the inconsistency of the single-character arguments in the coast documentation and the long-form arguments in the tutorials and gallery. 

It will be a lot of work to fully implement #262, but at least now we should stop documenting single-character arguments and document long-form arguments instead.

What we can do:

- Replace all single-character arguments with the corresponding long-form arguments, e.g., **B**->**frame**, **J**->**projection**.
- Remove the aliases from the docstring

weiji14:

To make things more readable, we'll shift our documentation to use long form arguments by default. Starting from the common aliases R, J, B, U, CPT, G and W. Will move on to all others spread across across the API Reference page.

See also upstream GMT docs at https://docs.generic-mapping-tools.org/latest/std_opts.html.

New looking documentation for this branch can be previewed at https://pygmt-git-longaliascommonopts.gmt.now.sh/api/index.html

Files to change:
- [x] helpers/decorators.py
- [ ] base_plotting.py
- [x] figure.py
- [x] gridding.py
- [x] mathops.py
- [x] modules.py

Fixes #473

**Reminders**

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -274,6 +274,7 @@ class BasePlotting:
         {R}
         {B}
         {G}
+        {U}
         {W}
         """
         kwargs = self._preprocess(**kwargs)
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -266,9 +266,9 @@ class BasePlotting:
         limit : str or list of 2 ints
             Do no draw contours below `low` or above `high`, specify as string
             ``'[low]/[high]'``  or list ``[low,high]``.
-        cut : string or int
+        cut : str or int
             Do not draw contours with less than `cut` number of points.
-        resample : string or int
+        resample : str or int
             Resample smoothing factor.
         {J}
         {R}
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -475,7 +475,7 @@ class BasePlotting:
 
         Parameters
         ----------
-        x, y : float or 1d arrays
+        x/y : float or 1d arrays
             The x and y coordinates, or arrays of x and y coordinates of the
             data points
         data : str or 2d array
@@ -579,7 +579,7 @@ class BasePlotting:
 
         Parameters
         ----------
-        x, y, z : 1d arrays
+        x/y/z : 1d arrays
             Arrays of x and y coordinates and values z of the data points.
         data : str or 2d array
             Either a data file name or a 2d numpy array with the tabular data.
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -70,6 +70,7 @@ class BasePlotting:
         W="shorelines",
         G="land",
         S="water",
+        U="timestamp",
     )
     @kwargs_to_strings(R="sequence")
     def coast(self, **kwargs):
@@ -439,6 +440,7 @@ class BasePlotting:
         i="columns",
         l="label",
         C="cmap",
+        U="timestamp",
     )
     @kwargs_to_strings(R="sequence", i="sequence_comma")
     def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
@@ -623,7 +625,7 @@ class BasePlotting:
                 lib.call_module("contour", arg_str)
 
     @fmt_docstring
-    @use_alias(R="region", J="projection", B="frame")
+    @use_alias(R="region", J="projection", B="frame", U="timestamp")
     @kwargs_to_strings(R="sequence")
     def basemap(self, **kwargs):
         """
@@ -664,7 +666,7 @@ class BasePlotting:
             lib.call_module("basemap", build_arg_string(kwargs))
 
     @fmt_docstring
-    @use_alias(R="region", J="projection")
+    @use_alias(R="region", J="projection", U="timestamp")
     @kwargs_to_strings(R="sequence")
     def logo(self, **kwargs):
         """
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -835,12 +835,12 @@ class BasePlotting:
         textfiles : str or list
             A text data file name, or a list of filenames containing 1 or more
             records with (x, y[, font, angle, justify], text).
-        x, y : float or 1d arrays
+        x/y : float or 1d arrays
             The x and y coordinates, or an array of x and y coordinates to plot
             the text
         text : str or 1d array
             The text string, or an array of strings to plot on the figure
-        angle: int/float or bool
+        angle: int, float or bool
             Set the angle measured in degrees counter-clockwise from
             horizontal. E.g. 30 sets the text at 30 degrees. If no angle is
             given then the input textfile(s) must have this as a column.
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -598,7 +598,7 @@ class BasePlotting:
             Network information
         label_placement : str
             Placement of labels
-        I : book
+        I : bool
             Color the triangles using CPT
         triangular_mesh_pen : str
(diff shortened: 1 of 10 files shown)


pygmt/helpers/decorators.py
```diff
@@ -14,30 +14,30 @@
...
+        projection : str
             *Required if this is the first plot command*.
             Select map projection.""",
     "B": """\
-        B : str or list
+        frame : str or list
             Set map boundary frame and axes attributes.""",
     "U": """\
-        U : bool or str
+        logo : bool or str
```
seisman, line 29: **-U** option is called **timestamp** in GMT, and **stamp** in GMT.jl. See https://github.com/GenericMappingTools/gmt/blob/362acc1302701ff3fe169638de5346f317e44670/src/gmt_init.c#L370. I think we should follow the GMT long-name options. 
```suggestion
        timestamp : bool or str
```

pygmt/helpers/decorators.py
```diff
@@ -55,13 +55,13 @@ def fmt_docstring(module_func):
...
-    * ``{J}``: J (projection)
-    * ``{B}``: B (frame)
-    * ``{U}``: U (insert time stamp)
-    * ``{CPT}``: CPT (the color palette table)
-    * ``{G}``: G (color)
-    * ``{W}``: W (pen)
+    * ``{R}``: region (bounding box as west, east, south, north)
+    * ``{J}``: projection (coordinate system to use)
+    * ``{B}``: frame (map frame and axes parameters)
+    * ``{U}``: logo (insert time stamp logo)
```
seisman, line 61:
```suggestion
    * ``{U}``: timestamp (insert time stamp logo)
```

pygmt/helpers/decorators.py
weiji14: Cool, good to have a standard to follow. I'll batch commit these changes.

seisman:
Perhaps we should update all the short-form arguments (if they have aliases) in this PR, not to update them in multiple PRs. I can help with it if you also agree.

weiji14:
> Perhaps we should update all the short-form arguments (if they have aliases) in this PR, not to update them in multiple PRs. I can help with it if you also agree.

Might as well. I'll work on `gridding.py`, `mathops.py` first since I wrote those. Do you want to work on `base_plotting.py`?

seisman:
> Might as well. I'll work on `gridding.py`, `mathops.py` first since I wrote those. Do you want to work on `base_plotting.py`?

OK.

weiji14:
> > Might as well. I'll work on `gridding.py`, `mathops.py` first since I wrote those. Do you want to work on `base_plotting.py`?
> 
> OK.

Alright, `gridding.py` and `mathops.py` are done. I'm moving on to `modules.py`, and `gmt info` seems to be a tricky one, as you mentioned in #147. Should I just turn the short aliases into long ones for now and ignore the fact that it produces strings like `-Rw/e/s/n` or `-Tzmin/zmax/dz` which go against the essence of this PR?

seisman:
> Should I just turn the short aliases into long ones for now and ignore the fact that it produces strings like `-Rw/e/s/n` or `-Tzmin/zmax/dz` which go against the essence of this PR

Yes, just ignore it.

pygmt/base_plotting.py
```diff
@@ -261,12 +263,12 @@ def grdcontour(self, grid, **kwargs):
...
             Do no draw contours below `low` or above `high`, specify as string
             ``'[low]/[high]'``  or list ``[low,high]``.
-        Q : string or int
+        cut : string or int
             Do not draw contours with less than `cut` number of points.
-        S : string or int
+        resample : string or int
             Resample smoothing factor.
         {J}
         {R}
```
weiji14, line 274: Need to add `{U}` for timestamp to `grdcontour` docstring here.

seisman: Ideally, we should add V to all the modules, and B, U, X, Y, and p to all plotting modules. I personally prefer to leave that to another separate PR.

weiji14: Maybe add the `{U}` for now since it's aliased already.

Agree to add `B, U, X, Y` later. Was thinking if there was a way to add it to all of them via the `_preprocess` function.

seisman: Done.

pygmt/base_plotting.py
```diff
@@ -100,34 +101,34 @@ def coast(self, **kwargs):
...
-        A : int, float, or str
+        area_thresh : int, float, or str
             ``'min_area[/min_level/max_level][+ag|i|s|S][+r|l][+ppercent]'``
             Features with an area smaller than min_area in km^2 or of
             hierarchical level that is lower than min_level or higher than
             max_level will not be plotted.
         {B}
         C : str
```
weiji14, line 110: No long alias for `C` in coast? GMT.jl uses `river_fill`, but has a system to handle lakes/rivers separately using a tuple. See https://www.generic-mapping-tools.org/GMT.jl/latest/coast/. We could do `riverlake_fill` :laughing:

seisman: I'm hesitating to add new long-name aliases in this PR. We should have more discussions in separate issues (i.e., one issue for each module), as we don't want to break backward compatibility due to some stupid names.

weiji14: Ok, I had a feeling it wasn't just randomly left out. Will treat it as an exception now then.

pygmt/base_plotting.py
```diff
@@ -586,21 +590,26 @@ def contour(self, x=None, y=None, z=None, data=None, **kwargs):
...
-            ``'[p|t]'``
+        levels : str
+            Contour file or level(s)
+        D : str
+            Dump contour coordinates
+        E : str
+            Network information
+        label_placement : str
+            Placement of labels
+        I : book
```
weiji14, line 601:
```suggestion
        I : bool
```

pygmt/base_plotting.py
```diff
@@ -634,7 +643,7 @@ def basemap(self, **kwargs):
         [optionally] gridlines. A simple map scale or directional rose may also
         be plotted.
 
-        At least one of the options *B*, *L*, or *T* must be specified.
+        At least one of the options *frame*, *L*, or *T* must be specified.
```
weiji14, line 646: I've only changed `B` to frame here. GMT.jl uses map_scale for `L`, rose for `Td` and compass for `Tm` (see https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.basemap), do we want to go with those?

seisman: Yes to me.

weiji14:
Do you want to take a look at https://pygmt-git-longaliascommonopts.gmt.now.sh/api/index.html to see if we missed any stray short aliases (other than the ones we agreed not to change)?

seisman:
> Do you want to take a look at https://pygmt-git-longaliascommonopts.gmt.now.sh/api/index.html to see if we missed any stray short aliases (other than the ones we agreed not to change)?

Made a small fix. Now it's good to me.

> P.S. Not a user facing change, but we should also update all the tests to use the long alias names at some point.

Yes, I agree.

Issue #340, opened 2019-10-10
Check Python 3.8 compatibility


seisman:
**Description of the problem**

Python 3.8 is [scheduled](https://www.python.org/dev/peps/pep-0569/) to be released on 2019-10-14. We need to check if pygmt works well with Python 3.8.

weiji14:
Not sure whether our upstream dependencies have Python 3.8 compatibility but won't hurt to try.

Referencing similar previous issue for Python 3.7 #190.

leouieda:
We can wait for conda-forge to start building packages for it. No one is going to update so soon and we probably won't have any issues since we don't use the parts of the code base that change a lot.

weiji14:
Just been cross-referencing our conda requirements at https://github.com/GenericMappingTools/pygmt/blob/63d23d4e0702edb94234f10d23d210cdf1adb54b/requirements.txt#L2-L6 with https://conda-forge.org/status/ and I think most packages have Python 3.8 compatibility now? Will need to test things out to be sure. Adding this to the v0.1.0 milestone.

Also, it would be good to take note of the Python version deprecation policies at [PEP 606 -- Python Compatibility Version](https://www.python.org/dev/peps/pep-0606/), and [NEP 29 — Recommend Python and Numpy version support as a community policy standard](https://numpy.org/neps/nep-0029-deprecation_policy.html). The latter recommends supporting at least two of the latest minor versions of Python and the inclusion of the following text somewhere in our project's development guidelines:

> This project supports:
> - All minor versions of Python released 42 months prior to the project, and at minimum the two latest minor versions.
> - All minor versions of numpy released in the 24 months prior to the project, and at minimum the last three minor versions.

which in practice means:

> On Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016)
> On Dec 26, 2021 drop support for Python 3.7 (initially released on Jun 27, 2018)
> On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019)

PR #475, opened 2020-06-10
Run Continuous Integration tests on Github Actions

weiji14:

Consolidate all our Linux/macOS/Windows unit tests into one `.github/workflows/ci_tests.yaml` file using a matrix build specification ~~(instead of having .travis.yml and .azure-pipelines.yml)~~. Will keep Travis CI because it's needed for the PyPI releases, but will disable Azure Pipelines for now.

Original attempt was made in https://github.com/weiji14/pygmt/pull/88, see some of the previous discussion/reviews there.

Fixes #

**Reminders**


weiji14:
Good work on the `wget` workaround @seisman! Tests seem to be failing on the new earth_relief grids only, should fix that first.

seisman:
> Good work on the `wget` workaround @seisman! Tests seem to be failing on the new earth_relief grids only, should fix that first.

FYI, we can't use `curl` to download files on Linux and windows, as it always timeouts. It's not clear to me why wget works and curl doesn't. GMT relies on the curl library to download remote files with a 10-s timeout, and of course GMT fails just like the curl command.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+      # Cache the ${HOME}/.gmt directory, for docs and testing
+      - name: Cache GMT directory
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: |
+            ~/.gmt/cache
+            ~/.gmt/server
+          key: cache-gmt-${{ runner.os }}-20200609
```
weiji14, line 66:
```suggestion
          key: cache-gmt-${{ github.ref }}-${{ runner.os }}-20200609
```
Sorry, should actually include the github ref here (either master, or the pull request number).

seisman: I don't get it. Does ` key: cache-gmt-${{ github.ref }}-${{ runner.os }}-20200609` mean that each branch and PRs will have different caches?

weiji14: Short answer: Yes. Long answer: That's to ensure that the PR cache doesn't 'pollute' the master cache. If you look at the next line `restore-keys: cache-gmt-refs/heads/master-`, it means that each PR can restore from the master's cache, but the master branch cannot see anything from the PRs. See also https://github.com/manubot/rootstock/pull/316/files#r386037243 where I got the idea from.

seisman: The PR cache should never be allowed to pollute the master cache. If it can, then it's a bad design of the cache mechanism. 

Anyway, it looks no harm to have `github.ref` in it.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install GMT and required dependencies
+        shell: bash -l {0}
+        run: |
+          requirements_file=full-conda-requirements.txt
+          cat requirements.txt requirements-dev.txt > $requirements_file
+          cat << EOF >> $requirements_file
+          gmt=6.0.0
+          make
+          EOF
+          conda install --yes --file $requirements_file
```
weiji14, line 50: We can speed up the dependency install using caching too, see https://github.com/goanpeca/setup-miniconda#caching. I've used it for one of my projects, the relevant commit is at https://github.com/weiji14/deepicedrain/commit/391632bc590c43409be4913325df5f023708c5f3.

seisman: Yes, I saw that. But I'm not sure how the caches are updated. For example, we may cache numpy v1.18.0 now. When new numpy versions are released, conda will download and install the latest version (e.g., numpy v1.19.0). What will happen to the caches? Do the caches contains the tarballs of numpy v1.18.0, v1.19.0 or both? Do we have to manually update the cache key?

weiji14: The cache would hash the `environment.yml/requirement.txt` files, and unless the hash changes, the cache will stay at v1.18.0 (unless we manually update the cache key). Maybe we should start thinking about a minimum supported dependency policy, as I suggested at https://github.com/GenericMappingTools/pygmt/issues/340#issuecomment-554545546.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+    - cron: '0 0 * * *'
+
+jobs:
+  test:
+    name: ${{ matrix.os }} - Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.7]
```
weiji14, line 22: Should we expand this to cover 3.6 and 3.8 as well?

seisman: Yes, but after we confirm that the current jobs all succees.

weiji14: Maybe I should revert 1b947d3fde42cbe99988cf341f26e88418399640 so that GH Actions runs on GMT 6.0, otherwise they will definitely fail.

seisman: Yes, 6.0.0 first.

weiji14: Ok, done. Should we disable the Azure Pipelines tests? Maybe just leave the "Style Checks" one so we'll know when it's working again.

seisman: Yes, sounds good.

seisman:
In bf2126f0c721f1de80b8643b3cb8e8cae2e39223, we cached the downloaded conda packages to improve the speed. However, it doesn't help at all. 

Comparing two runs [before](https://github.com/GenericMappingTools/pygmt/runs/820782421) and [after](https://github.com/GenericMappingTools/pygmt/runs/820845519) caching conda packages:

- before: setup miniconda (50 s) + install GMT and dependencies (3m54s)
- after: setup miniconda (2m37s) + install GMT and dependencies (1m48s)

I think we shouldn't cache conda packages, to make the actions more readable.

weiji14:
> I think we shouldn't cache conda packages, to make the actions more readable.

Sounds good. We can always add it back in later if we decide a cache for conda packages is needed.

seisman:
> > I think we shouldn't cache conda packages, to make the actions more readable.
> 
> Sounds good. We can always add it back in later if we decide a cache for conda packages is needed.

OK. I'll revert it after the CI jobs pass.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
seisman, line 10: Does this mean only PRs for the master branch are tested? What will happen if I have a PR for another PR?

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+      # Workaround for the timeouts of 'gmt which' on Linux and Windows
+      - name: Download remote data using wget (Linux & Windows)
+        shell: bash -l {0}
+        run: |
+          if [ "$RUNNER_OS" == "Windows" ]; then choco install wget; fi  # install wget on Windows
+          mkdir ~/.gmt ~/.gmt/cache ~/.gmt/server
+          wget --no-check-certificate https://oceania.generic-mapping-tools.org/gmt_hash_server.txt -P ~/.gmt/server/
+          for data in earth_relief_01d.grd earth_relief_30m.grd earth_relief_10m.grd; do
+            wget --no-check-certificate https://oceania.generic-mapping-tools.org/${data} -P ~/.gmt/server/
+          done
```
seisman, line 77: We need to change these lines when bumping to 6.1. In 6.1, the earth relief data are stored in `~/.gmt/server/earth/earth_relief/` directory.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+          for data in ridge.txt Table_5_11.txt tut_bathy.nc tut_quakes.ngdc tut_ship.xyz usgs_quakes_22.txt; do
+            wget --no-check-certificate https://oceania.generic-mapping-tools.org/cache/${data} -P ~/.gmt/cache/
+          done
+        if: steps.cache.outputs.cache-hit != 'true' && runner.os != 'macOS'
+
+      # Download remote files, if not already cached
+      - name: Download remote data (macOS)
+        shell: bash -l {0}
+        run: gmt which -Gu @earth_relief_01d @earth_relief_30m @earth_relief_10m @ridge.txt @Table_5_11.txt @tut_bathy.nc @tut_quakes.ngdc @tut_ship.xyz @usgs_quakes_22.txt
```
seisman, line 86: Need to change `gmt which -Ga` in GMT 6.1.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install the package
+        shell: bash -l {0}
+        run: |
+          python setup.py sdist --formats=zip
+          pip install dist/*
+
+      # Run the tests
+      - name: Test with pytest
+        shell: bash -l {0}
+        run: make test PYTEST_EXTRA="-r P"
```
seisman, line 99: Is the `PYTEST_EXTRA` necessary? I never understand what does it do.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,99 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
```
seisman, line 1: Add a badge to README.md

.github/workflows/ci_tests.yaml
weiji14: Then the CI won't run. I suppose we could change it so that it runs for any Pull Request.

.github/workflows/ci_tests.yaml
weiji14: On our travis.yml, it says https://github.com/GenericMappingTools/pygmt/blob/6b8d159454b07f3694117da530441835352f6504/.travis.yml#L65-L66

Might be best to leave it so that stdout and stderr is captured.

seisman: The variable is not used in any files of pygmt. I also can't find it in pytest documentation, so not an pytest environmental variable. Confusing.

weiji14: I can trace it back using `git blame` to #188. But yeah, can't find it in the pytest docs either, it's strange.

.github/workflows/ci_tests.yaml
weiji14: FYI, I've changed the "Required" status checks from Azure Pipeline to Github Actions for the Windows and macOS builds.

.github/workflows/ci_tests.yaml
weiji14: Ok, found it in the commandline.

```
-r chars              show extra test summary info as specified by chars: (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed, (p)assed, (P)assed with output, (a)ll except passed
                        (p/P), or (A)ll. (w)arnings are enabled by default (see --disable-warnings), 'N' can be used to reset the list. (default: 'fE').
```

seisman: I just ran `make test` and `make test PYTEST_EXTRA="-r P"`, the output are exactly the same, except the random directory names for images. But if I add `-r P` to the variable `PYTEST_ARGS` in Makefile, I can see more output like:
```
> basemap [WARNING]: Your primary y annotation interval exceeds the secondary interval.
> basemap [WARNING]: GMT expects it to be the other way around (primary annotations are closest to axis, secondary are further away)
> basemap [WARNING]: Consider correcting your command - proceeding with your selections
> basemap [WARNING]: Your primary y tick interval exceeds the secondary interval.
> basemap [WARNING]: GMT expects it to be the other way around (primary annotations are closest to axis, secondary are further away)
> basemap [WARNING]: Consider correcting your command - proceeding with your selections
> basemap [WARNING]: Your primary y annotation interval exceeds the secondary interval.
> basemap [WARNING]: GMT expects it to be the other way around (primary annotations are closest to axis, secondary are further away)
> basemap [WARNING]: Consider correcting your command - proceeding with your selections
> basemap [WARNING]: Your primary y tick interval exceeds the secondary interval.
> basemap [WARNING]: GMT expects it to be the other way around (primary annotations are closest to axis, secondary are further away)
> basemap [WARNING]: Consider correcting your command - proceeding with your selections
```

seisman: The extra outputs are useful. I think we should change the Makefile instead. 

Changing 
```
PYTEST_ARGS=--cov-config=../.coveragerc --cov-report=term-missing --cov=$(PROJECT) --doctest-modules -v --mpl --mpl-results-path=results --pyargs
```
to 
```
PYTEST_ARGS=--cov-config=../.coveragerc --cov-report=term-missing --cov=$(PROJECT) --doctest-modules -v --mpl --mpl-results-path=results --pyargs ${PYTEST_EXTRA}
```

Makefile
```diff
@@ -1,7 +1,7 @@
 # Build, package, test, and clean
 PROJECT=pygmt
 TESTDIR=tmp-test-dir-with-unique-name
-PYTEST_ARGS=--cov-config=../.coveragerc --cov-report=term-missing --cov=$(PROJECT) --doctest-modules -v --mpl --mpl-results-path=results --pyargs
+PYTEST_ARGS=--cov-config=../.coveragerc --cov-report=term-missing --cov=$(PROJECT) --doctest-modules -v --mpl --mpl-results-path=results --pyargs ${PYTEST_EXTRA}
```
weiji14, line 4: This won't work, because we didn't declare the `$PYTEST_EXTRA` environment variable. Also need to remove the `$PYTEST_EXTRA` string from the ci_tests.yaml file.

seisman: I think it works. It doesn't have to be an environmental variable. `make test` and `make test PYTEST_EXTRA="-r P"` are using different pytext arguments.

seisman: Now although the tests all pass, we actually can see more "errors" by using `-r P`:
```
______________________ test_call_module_invalid_arguments ______________________
----------------------------- Captured stderr call -----------------------------
gmtinfo [ERROR]: Error for input file: No such file (bogus-data.bla)
________________________ test_call_module_error_message ________________________
----------------------------- Captured stderr call -----------------------------
gmtinfo [ERROR]: Error for input file: No such file (bogus-data.bla)
__________________________ test_extract_region_fails ___________________________
----------------------------- Captured stderr call -----------------------------
pygmt-session [ERROR]: No hidden PS file found
____________________________ test_get_default_fails ____________________________
----------------------------- Captured stderr call -----------------------------
pygmt-session [ERROR]: Syntax error: Unrecognized keyword NOT_A_VALID_NAME
________________________ test_text_nonexistent_filename ________________________
----------------------------- Captured stderr call -----------------------------
text [ERROR]: Error for input file: No such file (notexist.txt)
_______________________________ test_which_fails _______________________________
----------------------------- Captured stderr call -----------------------------
gmtwhich [ERROR]: File 5a9f5f5cb31c4bb99525d0d82eaf0fc6 not found!

```

weiji14: Isn't it because the CI (Travis and GH Actions) still uses `make test PYTEST_EXTRA="-r P"`? If we run `make test` locally after checking out this branch, it probably won't work.

.github/workflows/ci_tests.yaml
```diff
@@ -0,0 +1,98 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
```
weiji14, line 9:
```suggestion
  pull_request
```
Do we need the `:` here?

Makefile
seisman: Yes. Look at the captured stderr above:
```
gmtinfo [ERROR]: Error for input file: No such file (bogus-data.bla)
```
This is a GMT error message, but I believe it's already corrected captured in the pygmt test.

I believe the initial idea of PYTEST_EXTRA is that, when we or other users run `make test` locally, we won't see these "error" messages, as they're a little confusing. But in CI, `make test PYTEXT_EXTRA="-r P"` gives more verbose messages, useful for debugging.

.github/workflows/ci_tests.yaml
seisman: I think yes. 

See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration.

> If you need to specify activity types or configuration for an event, you must configure each event separately. You must append a colon (:) to all events, including events without configuration.

Makefile
weiji14: Then we don't actually need to change the Makefile. You'll see that `PROJECT` is declared before using `$(PROJECT)`, but we don't declare `PYTEST_EXTRA` before using `$(PYTEST_EXTRA)`, it is effectively blank.

seisman: If you run `make test`, then `PYTEST_EXTRA` is blank; if you run `make test PYTEST_EXTRA="-r P"`, then variable PYTEST_EXTRA is defined in shell, and passed into Makefile.

seisman: `make test` runs:
```
cd tmp-test-dir-with-unique-name; pytest --cov-config=../.coveragerc --cov-report=term-missing --cov=pygmt --doctest-modules -v --mpl --mpl-results-path=results --pyargs  pygmt
```
and `make test PYTEST_EXTRA="-r P"` runs:
```
cd tmp-test-dir-with-unique-name; pytest --cov-config=../.coveragerc --cov-report=term-missing --cov=pygmt --doctest-modules -v --mpl --mpl-results-path=results --pyargs -r P pygmt
```

.github/workflows/ci_tests.yaml
weiji14: Good to know, it looks a bit funny though.

Makefile
weiji14: Right, I think I get it now. It wasn't working properly before, but now it does.

weiji14:
Thanks for all the help @seisman. The macOS tests seem to be a bit slower on Github Actions ~10min compared to before on Azure Pipelines ~7min. WIndows seems slightly faster sometimes. I'm sure we'll improve this once we get GMT 6.1 working properly, will merge this now :tada:

Issue #451, opened 2020-05-24
Baseline images need updates due to the recent updates of earth relief data


seisman:
**Description of the problem**

Upstream GMT has recently updated the global earth relief data (see https://github.com/GenericMappingTools/gmtserver-admin/pull/40 for the recipe change), which lead to some test failures in PyGMT.

We need to re-generate the baseline images using the latest earth relief data and GMT 6.0.0.

See https://github.com/GenericMappingTools/pygmt/pull/401 for a previous fix.

seisman:
While the Travis CI reports several test failures, the Azure Pipelines CIs don't. It's simply because we already [cached the earth relief in the Azure Pipelines CI](https://github.com/GenericMappingTools/pygmt/pull/438), thus Azure Pipelines CI is still using the old earth relief data. We need to update the cache key `20200519` to a new date to manually re-cache the new earth relief data.

seisman:
Re-open the issue again, since the GMT data server updated again recently.

The biggest change is that the GMT data server now provides earth relief grids in both pixel- and gridline-registrations.

In GMT 6.1.0 (not released yet), users can give a name like @earth_relief_30m_p or @earth_relief_30m_g to specify which registration they want to use. If no registration is given, e.g., @earth_relief_30m, then the pixel-registration grid is the default.

For GMT 6.0.0, it's a breaking change. The grids changes from gridline-registration to pixel-registration. 

Thus, we need to update the failing PyGMT tests.

weiji14:
We'll need to be very careful about this change (from gridline/node-registration to pixel-registration). It's been touched upon before in https://github.com/GenericMappingTools/pygmt/issues/375. Personally I'm in favour of pixel registration since that's what `xarray` assumes (I think?), but we might want to change the following lines:

https://github.com/GenericMappingTools/pygmt/blob/0267dd15c7626f5903cdb5a3e12ca545d70c884a/pygmt/clib/session.py#L567-L569

Currently it defaults to gridline-registration. Ideally we would be able to specify whether pixel- or gridline/node- is used.

seisman:
Yes, it looks a big issue. 

> Currently it defaults to gridline-registration GMT_GRID_NODE_REG.

Yes, the default registration in GMT is also gridline registration, so we will keep as it is.

> Ideally we would be able to specify whether pixel- GMT_GRID_PIXEL_REG or gridline/node- GMT_GRID_NODE_REG is used.

[...]

Issue #390, opened 2020-01-14
Cylindrical projections 'Q' plot incorrectly in pygmt when using xarrays


MarkWieczorek:
**Pygmt incorrectly plots xarrays when using cylindrical projections ('Q').** 

This is not an issue with gmt: if the dataarray is first exported to a netcdf file, pygmt works as intended when using the file. Thus, this appears to be a problem with how pygmt treats internal xarrays, and only for the case of cylindrical projections (the most basic of all the projections!). This is demonstrated in the following script.

The script generates a simple dataset, and the plots the following:

* upper row: 'Q0/0/6i' and 'Q180/0/6i' reading data from a file.
* middle row: 'W0/6i' and 'W180/6i' using internal xarray.
* bottom row:  'Q0/0/6i' and 'Q180/0/6i' using internal xarray.

As you see, the upper two rows plot the data correctly. However, for the bottom left image, the data are **not** plotted with the correct `central_longitude`. Furthermore, there is a problem with the first longitude band not being plotted.

It is possible that this is related to this issue: https://github.com/GenericMappingTools/pygmt/issues/375

```
# Creata a data array in gridline coordinates of sin(lon) * cos(lat)
interval = 10
lat = np.arange(90, -90-interval, -interval)
lon = np.arange(0, 360+interval, interval)
longrid, latgrid = np.meshgrid(lon, lat)
data = np.sin(np.deg2rad(longrid)) * np.cos(np.deg2rad(latgrid))

# create a DataArray
dataarray = xr.DataArray(data, coords=[('latitude', lat,
                                       {'units': 'degrees_north'}),
                                       ('longitude', lon,
                                       {'units': 'degrees_east'})], 
                         attrs = {'actual_range': [-1, 1]})
dataset = dataarray.to_dataset(name='dataarray')
dataset.to_netcdf('test.grd')

# create projected images using a cylindrical and mollweide projection.
fig = pygmt.Figure()
fig.grdimage(dataset.dataarray, region='g', projection='Q0/0/6i', frame='a90f30g30')
fig.grdimage(dataset.dataarray, region='g', projection='Q180/0/6i', X='7i', frame='a90f30g30')
fig.grdimage(dataset.dataarray, region='g', projection='W0/6i', frame='a90f30g30', X='-7i', Y='4i')
fig.grdimage(dataset.dataarray, region='g', projection='W180/6i', X='7i', frame='a90f30g30')
[...]

Issue #364, opened 2019-11-02
grdimage shading fails for xarray input


liamtoney:
**Description of the problem**

Passing `shading=True` in a simple `fig.grdimage()` call produces an error if the input grid is an `xarray.DataArray`. The error does not appear when I use the special "@" filename.

**Full code that generated the error**

```python
import pygmt

fig = pygmt.Figure()

dem = pygmt.datasets.load_earth_relief('10m')

fig.grdimage(dem, region=[-180, 180, -90, 90], frame='af',
             projection='Cyl_stere/6i', cmap='geo', shading=True)

fig.show(method='external')
```
This works, though:

```python
import pygmt

fig = pygmt.Figure()

fig.grdimage('@earth_relief_10m', region=[-180, 180, -90, 90], frame='af',
             projection='Cyl_stere/6i', cmap='geo', shading=True)

fig.show(method='external')
```

**Error message summary**
```
grdimage [ERROR]: Passing max <= zmin prevents automatic CPT generation!
grdimage [ERROR]: Failed to read CPT geo.
```

<details>
<summary><b>Full error message (debug verbosity)</b></summary>
<br>

```
grdimage [DEBUG]: Map distance calculation will be using great circle approximation with authalic auxiliary latitudes and authalic (R_2) radius = 6371007.1809 m, in meter.
grdimage [DEBUG]: Look for file -180/180/-90/90 in /Users/liamtoney/.gmt
grdimage [DEBUG]: Look for file -180/180/-90/90 in /Users/liamtoney/.gmt/cache
grdimage [DEBUG]: Look for file -180/180/-90/90 in /Users/liamtoney/.gmt/server
grdimage [DEBUG]: Look for file -180/180/-90/90 in /Users/liamtoney/.gmt/server/srtm1
grdimage [DEBUG]: Look for file -180/180/-90/90 in /Users/liamtoney/.gmt/server/srtm3
grdimage [DEBUG]: Got regular w/e/s/n for region (-180/180/-90/90)
grdimage [DEBUG]: Look for file geo in /Users/liamtoney/.gmt
grdimage [DEBUG]: Look for file geo in /Users/liamtoney/.gmt/cache
grdimage [DEBUG]: Look for file geo in /Users/liamtoney/.gmt/server
grdimage [DEBUG]: Look for file geo in /Users/liamtoney/.gmt/server/srtm1
grdimage [DEBUG]: Look for file geo in /Users/liamtoney/.gmt/server/srtm3
[...]

PR #327, opened 2019-09-19
Upgrade gmt from 6.0.0rc1 to 6.0.0rc4

weiji14:
**Description of proposed changes**

Testing out the new *patched* GMT 6.0.0rc4 release (see upstream https://github.com/GenericMappingTools/gmt/issues/1471) for PyGMT, a step towards readying it for the GMT6 release coming soon. Full list of changes at https://github.com/GenericMappingTools/gmt/compare/6.0.0rc1...6.0.0rc4 in addition to the patch at https://github.com/conda-forge/gmt-feedstock/pull/85.

Builds upon #311. Might also be a good time to add GMT back to our conda installation (i.e. revert some aspects of https://github.com/GenericMappingTools/pygmt/pull/261).

Noticed some tests were breaking on my local instance when developing on `GMT 6.0.0rc4`, specifically `test_session_management.py` ~~and `test_sphinx_gallery.py`. Has something to do with modern/classic mode incompatibility, possibly because of https://github.com/GenericMappingTools/gmt/issues/1142~~. Tracking and resolving it here in this pull request.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #476, opened 2020-06-10
Properly allow for either pixel or gridline registered grids

### LINKED ISSUE #375: grdimage does not correctly plot xarray DataArrays when they don't contain the redundant longitude at 360 E
`grdimage` of pygmt does not correctly plot xarray DataArrays when they do not contain the redundant column of data at 360 E. 

Using pygmt directly with a netcdf file provides the correct output, such as with a command like:
```
fig.grdimage('test.grd', region='g', projection='A0/0/6i')
```
test

However, here is the same output generated by pygmt when using an xarray in memory:
```
fig.grdimage(grid, region='g', projection='A0/0/6i')
```
longitude

weiji14:

`xarray` assumes `pixel` registration (aka data values represent centre of the pixel), whereas GMT defaults to using `gridline` registration (aka data values represent top left corner of pixel). PyGMT currently defaults to `gridline` registration, following upstream GMT, but users have no way to easily set it to `pixel` registration.

Gridline vs Pixel registration

See http://xarray.pydata.org/en/stable/plotting.html#coordinates and https://docs.generic-mapping-tools.org/latest/cookbook/options.html#grid-registration-the-r-option for context. NOAA also has a good page that talks about grid registration at https://ngdc.noaa.gov/mgg/global/gridregistration.html

Grid vs Cell registration

This is a major bug, and may cause breaking changes for some users. Users will have to be aware of what type of registration their raster grids' are using, and unfortunately, those using `xarray` may need to set the registration manually to **p** (for pixel) for correctness, unless we default to using `pixel` registration in PyGMT. Alternatively, the [`salem`](https://salem.readthedocs.io) library has an xarray accessor to represent an xarray grid as a [center_grid](https://salem.readthedocs.io/en/v0.2.3/generated/salem.Grid.center_grid.html#salem-grid-center-grid) or [corner_grid](https://salem.readthedocs.io/en/v0.2.3/generated/salem.Grid.corner_grid.html#salem.Grid.corner_grid), as used in https://github.com/weiji14/deepbedmap/pull/150.

Will be good to have some discussion on what's the best plan forward.

Fixes #375? #390? Needed for #451.

**Reminders**


weiji14:
> Specifically, the code region.extend([coord.min(), coord.max()]) can't surely be valid for both gridline and pixel registered cases. If I'm thinking properly, it is incorrect for the pixel registered case (which xarray uses) by a pixel, the region should be something like coord.min() - inc/2, coord.max() + inc/2).

Found a way to illustrate this problem: when running `grdinfo` at https://github.com/GenericMappingTools/try-gmt on the actual grd file vs the virtual earth_relief grid:

```python
import pprint
import pygmt

pprint.pprint(pygmt.grdinfo("@earth_relief_60m"), width=1000)
```
produces:
```
('/home/jovyan/.gmt/server/earth_relief_60m.grd: Title: Earth Relief at 01 arc degree\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: Command: grdfilter SRTM15+V2.1.nc -Fg111.2 -D1 -I01d -rp -Gearth/earth_relief/earth_relief_01d_p.grd=ns+s0.5+o0 --IO_NC4_DEFLATION_LEVEL=9 --IO_NC4_CHUNK_SIZE=4096 --PROJ_ELLIPSOID=Sphere\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: Remark: Obtained by Gaussian Cartesian filtering (111.2 km fullwidth) from SRTM15+V2.1.nc [Tozer et al., 2019; http://dx.doi.org/10.1029/2019EA000658]\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: Pixel node registration used [Geographic grid]\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: Grid file format: ns = GMT netCDF format (16-bit integer), CF-1.7\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: x_min: -180 x_max: 180 x_inc: 1 name: longitude n_columns: 360\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: y_min: -90 y_max: 90 y_inc: 1 name: latitude n_rows: 180\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: z_min: -8182 z_max: 5651.5 name: elevation (m)\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: scale_factor: 0.5 add_offset: 0 packed z-range: [-16364,11303]\n'
 '/home/jovyan/.gmt/server/earth_relief_60m.grd: format: netCDF-4 chunk_size: 180,180 shuffle: on deflation_level: 9\n')
```
whereas
```python
grid = pygmt.datasets.earth_relief.load_earth_relief()
pprint.pprint(pygmt.grdinfo(grid))
```
produces
```
(': Title: \n'
 ': Command: \n'
 ': Remark: \n'
 ': Gridline node registration used [Cartesian grid]\n'
 ': Unrecognized grid file format! Probably not a GMT grid\n'
 ': x_min: -179.5 x_max: 179.5 x_inc: 1 name:  n_columns: 360\n'
 ': y_min: -89.5 y_max: 89.5 y_inc: 1 name:  n_rows: 180\n'
 ': z_min: -8182 z_max: 5651.5 name: z\n'
 ': scale_factor: 1 add_offset: 0\n')
```

Notice how the x_min/x_max of the virtual file isn't at -180/180, ditto with y_min/y_max. Plus it thinks the grid is gridline registered, rather than pixel registered. Also, not sure if it's okay treating the grid as Cartesian instead of Geographic (probably a separate issue with the virtualfile mechanism). I've pushed a commit using `GMT_GRID_PIXEL_REG` as the default to see how things turn out.

seisman:
There may be a potential bug in GMT. Please follow https://github.com/GenericMappingTools/gmt/issues/3502 and https://github.com/GenericMappingTools/gmt/pull/3503 for upstream feedback.

seisman:
It's confirmed that there was a bug when passing a pixel-registered matrix to GMT C API. It was already fixed in https://github.com/GenericMappingTools/gmt/pull/3503 and merged in to master. I can confirm that the current PR works well. Thus what we need to do is: 

1. decide the default registration for array data
2. try our best to detect and guess the grid registration (especially for netcdf file loaded as xarray)
3. let users specify the registration they want to use
4. wait for the GMT 6.1 release.

seisman:
> doesn't seem okay treating the grid as Cartesian instead of Geographic (but that's probably a separate issue with the virtualfile mechanism). Not sure if it's a limitation of the [`GMT_Create_Data`](https://docs.generic-mapping-tools.org/dev/api.html#gmt-create-data) C API?

https://github.com/GenericMappingTools/pygmt/blob/86c46f075d6f68a1367bc0fad161f98ba99c14c3/pygmt/clib/session.py#L1243

Changing `GMT_CONTAINER_ONLY` to `GMT_CONTAINER_ONLY|GMT_GRID_IS_GEO`, then GMT will know it's a geographic grid.

pygmt/clib/session.py
```diff
@@ -1242,7 +1242,7 @@ def virtualfile_from_grid(self, grid, registration="GMT_GRID_PIXEL_REG"):
         gmt_grid = self.create_data(
             family,
             geometry,
-            mode="GMT_CONTAINER_ONLY",
+            mode="GMT_CONTAINER_ONLY|GMT_GRID_IS_GEO",
```
seisman, line 1245: No. I didn't mean this change. Users can pass both Cartesian and geographic grids to GMT. We need to use `GMT_CONTAINER_ONLY` or `GMT_CONTAINER_ONLY|GMT_GRID_IS_GEO` for different grid type.

weiji14: Ah sorry, I thought it 'knew' automatically what the grid type was. Should actually get some tests in for this.

> BTW, `GMT_CONTAINER_ONLY` is equivalent to `GMT_CONTAINER_ONLY|GMT_GRID_IS_CARTESIAN`.

Ok, good to know.

weiji14: Just for reference, this was added in GMT 5.4. See:
- https://docs.generic-mapping-tools.org/dev/changes.html?highlight=gmt_grid_is_geo#api-changes
- https://docs.generic-mapping-tools.org/dev/api.html?highlight=gmt_grid_is_geo#exporting-a-data-set
- https://github.com/GenericMappingTools/gmt/commit/b6d0af653c7b5f33d6af172f1ec4ff6be9d4eed3.

weiji14:
> It's confirmed that there was a bug when passing a pixel-registered matrix to GMT C API. It was already fixed in [GenericMappingTools/gmt#3503](https://github.com/GenericMappingTools/gmt/pull/3503) and merged in to master. I can confirm that the current PR works well. Thus what we need to do is:
> 
>     1. decide the default registration for array data
> 
>     2. try our best to detect and guess the grid registration (especially for netcdf file loaded as xarray)

I've added an automatic gridline/pixel registration detector in 6d4ece422fc155d61ca0d288d6207a0b0dca79ae. It simply does `grdinfo` on the xarray.DataArray's NetCDF source (specifically via [`grid.encoding["source"]`](http://xarray.pydata.org/en/v0.15.1/io.html#reading-encoded-data), and will work on the `earth_relief` grids. If that doesn't work (e.g. for grids generated purely in Python), it will fallback to gridline registration by default.

seisman:
> I've added an automatic gridline/pixel registration detector in [6d4ece4](https://github.com/GenericMappingTools/pygmt/commit/6d4ece422fc155d61ca0d288d6207a0b0dca79ae). It simply does `grdinfo` on the xarray.DataArray's NetCDF source (specifically via [`grid.encoding["source"]`](http://xarray.pydata.org/en/v0.15.1/io.html#reading-encoded-data), and will work on the `earth_relief` grids. If that doesn't work (e.g. for grids generated purely in Python), it will fallback to gridline registration by default.

I thought it may be simple to save the registration information to xarray.DataArray's `attrs` dict, i.e.,  dem.attrs['registration'] = "pixel" or "gridline". 

1. If `dem.attrs['registration']` is defined by users or by the `load_earth_relief()` function, then PyGMT should pass the registration directly to GMT
2. If it's undefined, try to automatically detect it
3. If auto-detect fails, fall backs to the default registration.

FYI, the implementation may be totally different and simplified if our next target release is for GMT 6.1.0.

1. In GMT 6.1.0, earth_relief data have both pixel and gridline registration. It means we need to change the `load_earth_relief()` function to accept argument like `registration="pixel"` or `registration="gridline"`. Thus, we may know the registration information directly from the argument (it has a default), and no need to call `grdinfo`. 

2. earth_relief data now have file names like `earth_relief_xxy_p` or `earth_relief_xxy_g`. We may get the registration information by parsing the file name.

3. In GMT 6.1.0, `grdinfo -C` returns the grid registration as the last column of the output.

weiji14:
> I thought it may be simple to save the registration information to xarray.DataArray's `attrs` dict, i.e., dem.attrs['registration'] = "pixel" or "gridline".
> 
> 1. If `dem.attrs['registration']` is defined by users or by the `load_earth_relief()` function, then PyGMT should pass the registration directly to GMT
> 
> 2. If it's undefined, try to automatically detect it
> 
> 3. If auto-detect fails, fall backs to the default registration.

So 2 and 3 are implemented already in this PR. As for 1, I agree that users should be able to define "pixel" or "gridline" as they like, but I'm not sure if using `dem.attrs["registration"]` is a good idea. To be honest, `xarray` doesn't handle CRS/Projection information well either (see https://github.com/pydata/xarray/issues/2288), even for CF valid datasets, and the last thing the geo- community wants is another standard for handling geographic information in grids as per https://xkcd.com/927/.

> 
> FYI, the implementation may be totally different and simplified if our next target release is for GMT 6.1.0.
> 
> 1. In GMT 6.1.0, earth_relief data have both pixel and gridline registration. It means we need to change the `load_earth_relief()` function to accept argument like `registration="pixel"` or `registration="gridline"`. Thus, we may know the registration information directly from the argument (it has a default), and no need to call `grdinfo`.
> 
> 2. earth_relief data now have file names like `earth_relief_xxy_p` or `earth_relief_xxy_g`. We may get the registration information by parsing the file name.
> 
> 3. In GMT 6.1.0, `grdinfo -C` returns the grid registration as the last column of the output.

We'll need to think of other non earth_relief grids too, not just the earth_relief ones. We *could* treat the earth_relief grids as a special case I suppose, if there is say, a speed benefit of not running `grdinfo` first? Good to know about `grdinfo -C`, will refactor the code after GMT 6.1.0.

seisman:
> As for 1, I agree that users should be able to define "pixel" or "gridline" as they like, but I'm not sure if using `dem.attrs["registration"]` is a good idea. 

I don't know much about the standards (TLTR). To me, `dem.attrs["registration"]` is simple, and easy to keep backward compatibility if we find a real standard in the future, unless the standard uses `registration` for other purposes.

> We'll need to think of other non earth_relief grids too, not just the earth_relief ones. 

I think all GMT remote files follow the same naming convention. See https://github.com/GenericMappingTools/gmtserver-admin/issues/44.

MarkWieczorek:
> Users will have to be aware of what type of registration their raster grids' are using, and unfortunately, those using xarray may need to set the registration manually to p (for pixel) for correctness, unless we default to using pixel registration in PyGMT. 

Whenever I use xarray, my xarray grids are gridline registered. In my opinion, we should not assume that a given xarray array is pixel registered. Xarray has a very basic plotting routine that does assume pixel registration, but its my opinion that their plotting routine needs lots of improvement: the addition of gridline registered grids is just one thing that they need to improve.

seisman:
@weiji14 I saw that you're adding two parameters `coord_sys` and `registration` to some grid-related modules. It seems to work well, but please keep in mind that, GMT can both read and write grids, and PyGMT can read/write grid and xarray.DataArray. Input and output grids may have different registrations, e.g., users may have a gridline-registered grid, and call `grdsample -rp` to produce a pixel-registered grid.

weiji14:
> @weiji14 I saw that you're adding two parameters `coord_sys` and `registration` to some grid-related modules. It seems to work well, but please keep in mind that, GMT can both read and write grids, and PyGMT can read/write grid and xarray.DataArray. Input and output grids may have different registrations, e.g., users may have a gridline-registered grid, and call `grdsample -rp` to produce a pixel-registered grid.

Thanks for reminding me, I'll change `registration` as an alias for `-r` (i.e. output registration) to keep in line with GMT.

Will need to have a rethink about how to refactor the implementation, all the if-thens are making my head spin, and they slow down the code. Was thinking of using boolean flags like `is_geographic` and `is_pixel_reg`, but the `False`s get wiped out and the `True`s get turned into empty strings `""`.

seisman:
I asked Paul about details of grid registration. Here are my questions and his answers.

**Q1:** Is the "node_offset" attribution in netCDF files equivalent to the grid registration, i.e., node_offset = 0 or no node_offset means gridline registration, and node_offset=1 means pixel registration?

**A1:** Code does this:
```
		if (header->registration == GMT_GRID_PIXEL_REG) {
			int reg = header->registration;
			gmt_M_err_trap (nc_put_att_int (ncid, NC_GLOBAL, "node_offset", NC_LONG, 1U, &reg));
		}
		else
			nc_del_att (ncid, NC_GLOBAL, "node_offset");
```
So only if it is pixel reg do we write the registration value, otherwise it is assumed it is gridline reg and we write nothing
When reading we check for it:
```
		if (!nc_get_att_int (ncid, NC_GLOBAL, "node_offset", &i)) {	/* GMT wrote the registration in the grid */
			header->registration = i;
			set_reg = false;	/* Do not update it below since we know the registration */
		}
```

**Q2:** Does GMT always write the node_offfset attributes to netCDF files?  "ncdump earth_relief_01d_p.grd" does show node_offset=1, but "ncdump earth_relief_01d_g.grd" doesn't show the node_offset attribution.

**A2:** See above.  only GMT pixel grids will have that variable.

**Q3:** Is the node_offset attribution specific to GMT? Or is it a standard attribution defined by COARDS convention?

**A3:** I think it is specific to GMT.  If we find it then we know for sure.  Otherwise, we do a lot of checking to figure out what the file really is.  Occasionally we find really shitty files that pretend to be gridline but clearly are pixel (e.g., latitudes in whole degrees going from -89.5 to 89.5). So reading a netcdf grid is complicated since anyone might have created one.

seisman:
For any netCDF grids produced by GMT (e.g., earth relief data), we can always check the `node_offset` attribution. If `node_offset=1`, then we know it's pixel registration. If it's not undefined, then it's gridline registration (the default value in GMT). 

As `node_offset` is specific to GMT, it makes more sense and also easier to detect registration if PyGMT uses gridline registration as the default. For any xarray.DataArray grids (created by users, read from a GMT netCDF file, or read from a netCDF file produced by other tools), automatically detecting grid registration is simply checking the value of the `node_offset` attrs.

pygmt/clib/conversion.py
```diff
@@ -102,7 +105,10 @@ def dataarray_to_matrix(grid):
                     dim
                 )
             )
-        region.extend([coord.min(), coord.max()])
+        if registration == "GMT_GRID_PIXEL_REG":
+            region.extend([coord.min() - coord_inc / 2, coord.max() + coord_inc / 2])
+        elif registration == "GMT_GRID_NODE_REG":
+            region.extend([coord.min(), coord.max()])
```
seisman, line 111: In GMT, `GMT_GRID_NODE_REG=0` and `GMT_GRID_PIXEL_REG=1`. Thus, the if-else statement can be simplified as code:
```
# node_offset (i.e., registration in int) is 0 or 1
region.extend([coord.min() - coord_inc / 2 * node_offset, coord.max() + coord_inc / 2 * node_offset])
```

seisman:
As mentioned above, I personally prefer to have the registration and coord_sys information in the xarray.DataArray attributes. The reasons are:

1. No need to add these two parameters to every grd* functions (registration and coord_sys are all internally handled by the virtualfile_from_grid function)
2. Input and output grids can have different registrations and coords_sys, and the codes can be simplified. 

For example, users can generate a grid by calling `surface()` function, and plot it via `grdimage()`. The `surface()` function has to return registration and coord_sys. Thus, the codes may look like:
```
grid, registration, coord_sys = surface(...)
grdimage(grid, registration, coords)
```
If registration and coord_sys are attributions of a xarray.DataArray object, the code looks like
```
grid = surface()
grdimage(grid)
```

pygmt/clib/conversion.py
weiji14: Awesome, then a boolean "registration" should be the way to go, and GMT.jl uses a boolean style `reg` argument too as at https://forum.generic-mapping-tools.org/t/deciding-on-default-grid-registration-and-grid-type-for-pygmt/583/4. Will wrap then in later.

weiji14:
> As mentioned above, I personally prefer to have the registration and coord_sys information in the xarray.DataArray attributes. The reasons are:

Ok, then I'll probably have to rename `coord_sys` to something better (I can see it will clashing with the namespace of some other library already). If we're going with a boolean type, maybe something like `is_geographic`? I.e. True (1) is `GMT_GRID_IS_GEO` while False (0) is `GMT_GRID_IS_CARTESIAN` (default).

Also, we'll need to do the following:

1. Change the attribute when [`grdsample -T`](https://docs.generic-mapping-tools.org/dev/grdsample.html#t) or [`grdedit -T`](https://docs.generic-mapping-tools.org/latest/grdedit.html#t) is called (when those functions are wrapped in the future).
2. Document this 'feature' (that an attribute needs to be set to change the registration/coord_sys) somewhere visible.

pygmt/base_plotting.py
```diff
@@ -420,7 +424,9 @@ def grdview(self, grid, **kwargs):
                     drapegrid = kwargs["G"]
                     if data_kind(drapegrid) in ("file", "grid"):
                         if data_kind(drapegrid) == "grid":
-                            drape_context = lib.virtualfile_from_grid(drapegrid)
+                            drape_context = lib.virtualfile_from_grid(
+                                drapegrid, coord_sys, in_reg
```
weiji14, line 428: @seisman, your idea of using xarray attributes would actually make sense here in `grdview`, as we would otherwise need two `in_reg`s to deal with the drapegrid and reliefgrid separately.

weiji14:
Scratch that, let's use `node_offset` (ties in with https://github.com/GenericMappingTools/pygmt/pull/476#issuecomment-649864135) . I just tried reading the earth_relief grids using [xarray.open_dataset](xarray.pydata.org/en/v0.15.1/generated/xarray.open_dataset.html#xarray.open_dataset) (instead of xarray.open_dataarray), and this is how it looks:

```
<xarray.Dataset>
Dimensions:  (lat: 180, lon: 360)
Coordinates:
  * lon      (lon) float64 -179.5 -178.5 -177.5 -176.5 ... 177.5 178.5 179.5
  * lat      (lat) float64 -89.5 -88.5 -87.5 -86.5 -85.5 ... 86.5 87.5 88.5 89.5
Data variables:
    z        (lat, lon) float32 ...
Attributes:
    Conventions:  CF-1.7
    history:      grdfilter SRTM15+V2.1.nc -Fg111.2 -D1 -I01d -rp -Gearth/ear...
    GMT_version:  6.1.0_e80cfbd_2020.05.28 [64-bit]
    node_offset:  1
    title:        Earth Relief at 01 arc degree
    description:  Obtained by Gaussian Cartesian filtering (111.2 km fullwidt...
```

So we can follow this 'convention' instead of defining our own. We'll just need to modify these lines:

https://github.com/GenericMappingTools/pygmt/blob/516e799bb9753ed45b9e122b8ebb5fd446320866/pygmt/datasets/earth_relief.py#L40-L52

pygmt/base_plotting.py
seisman: `grdmath` may also take more than one grid.

pygmt/clib/session.py
```diff
@@ -1237,17 +1240,58 @@ def virtualfile_from_grid(self, grid):
...
+            assert grid.attrs["geocoord_type"] == 1
+            _gtype = "GMT_GRID_IS_GEO"
+        except (AssertionError, KeyError):
+            _gtype = "GMT_GRID_IS_CARTESIAN"
+
+        try:
+            assert grid.attrs["node_offset"] == 1
+            _registration = "GMT_GRID_PIXEL_REG"
+        except (AssertionError, KeyError):
+            _registration = "GMT_GRID_NODE_REG"
```
weiji14, line 1253: I've renamed `coord_sys` to `geocoord_type`. Where `1=GMT_GRID_IS_GEO` and `0=GMT_GRID_IS_CARTESIAN`. As for the `node_offset` attribute, `1=GMT_GRID_PIXEL_REG` and `0=GMT_GRID_NODE_REG`.

This is fine for plotting, but a problem we'll face is that these attributes might not get propagated when passing them into non-plotting modules (i.e. data processing ones like `grdcut`). That `pygmt.open()` function is starting to look really appealing now, xref https://github.com/GenericMappingTools/pygmt/pull/492#discussion_r446431644.

weiji14: You know what, I have another idea - make a `gmt` [xarray accessor](http://xarray.pydata.org/en/v0.15.1/internals.html#extending-xarray). I'll open up an issue to describe this further, but for the purposes of this PR, the way we would access the `registration` from the xr.DataArray is via `grid.gmt.registration` instead of through `grid.attrs["registration"]`. The benefit of the [xarray dataset/dataarray accessor](http://xarray.pydata.org/en/v0.15.1/generated/xarray.register_dataset_accessor.html#xarray.register_dataset_accessor) is that it would be more stable than keeping stuff in `attrs`. It also means we won't need a `pygmt.open()` function, and the #494 PR might not be needed anymore.

weiji14: Ok, now that #500 is merged (along with the bump to GMT 6.1.0), this PR should be prime for a good refresh. Will need some time resolve the merge conflicts/clean up the code!

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,18 @@ def test_grdimage_fails():
...
+@pytest.mark.mpl_image_compare
+def test_grdimage_over_dateline(xrgrid):
+    """
+    Ensure no gaps are plotted over the 180 degree international dateline.
+    Specifically checking that `xrgrid.gmt.gtype = 1` sets `GMT_GRID_IS_GEO`,
+    and that `xrgrid.gmt.registration = 0` sets `GMT_GRID_NODE_REG`. Note that
+    there would be a gap over the dateline if a pixel registered grid is used.
+    """
+    fig = Figure()
+    assert xrgrid.gmt.registration == 0  # gridline registration
```
weiji14, line 87: This dateline test seems to work sometimes, but crashes at other times, reminds me of the black image we got a while ago. Was there a regression in GMT?

| baseline-test_grdimage_over_dateline | test_grdimage_over_dateline-failed-diff | test_grdimage_over_dateline
|:---:|:---:|:---:|
| Expected | Diff | Generated |

seisman: @weiji14 Please see if #517 works for you. 

You can cherry-pick commit b64f3713017749a80d704ef9683dd140b95b6ea7 to your local branch and see it can fix the random crashes.

weiji14: Nope, still get the all black image :/

seisman: I ran the tests 10 times, and always get the expected figure, but I'm using macOS.

seisman: Can you add `V='d'` to the grdimage function, and post the output so that I can compare if there are any differences.

weiji14: Yes, the macOS tests all pass, but if you look at the Ubuntu ones, only Python 3.8 passes (3.6 and 3.7 fails). Not sure about Windows.

weiji14: I get like 510 lines of `grdimage (gmtsupport_free_range): tried to free unallocated memory`. This is without cherry-picking #517 by the way, on Python 3.7:

<details>

```
grdimage [DEBUG]: Map distance calculation will be using great circle approximation with authalic auxiliary latitudes and authalic (R_2) radius = 6371007.1809 m, in meter.
grdimage [DEBUG]: Map distance calculation will be using great circle approximation with authalic auxiliary latitudes and authalic (R_2) radius = 6371007.1809 m, in meter.
grdimage [DEBUG]: Projected grid is non-orthogonal, nonlinear, or dpi was changed
grdimage [INFORMATION]: Read header from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 1
grdimage [DEBUG]: Grid/Image dimensions imply w/e/s/n = 0/359/-89/90, inc = 1/1, gridline registration, n_layers = 1
grdimage [DEBUG]: Geographic input grid, longitudes span less than 360
grdimage [DEBUG]: Chosen boundary condition for all edges: geographic
grdimage [DEBUG]: Geographic input grid, longitude cells span exactly 360
grdimage [DEBUG]: Object ID 1 : Registered Grid Memory Reference 5593eda267a0 as an Input resource with geometry Surface [n_objects = 2]
grdimage [DEBUG]: Successfully created a new Grid container
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [DEBUG]: Projected values in meters: -9.00996e+06 9.00996e+06 -9.00996e+06 9.00996e+06
grdimage [INFORMATION]: Map scale is 7094.46 km per cm or 1:7.09446e+08.
grdimage [DEBUG]: Running in PS mode modern
grdimage [DEBUG]: Use PS filename /home/user/.gmt/sessions/gmt_session.3726/gmt_6.ps-
grdimage [DEBUG]: Create hidden PS file /home/user/.gmt/sessions/gmt_session.3726/gmt_6.ps-
grdimage [DEBUG]: Got session name as pygmt-session and default graphics formats as pdf
grdimage [INFORMATION]: Allocate and read data from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 2
grdimage [INFORMATION]: Importing grid data from user matrix memory location
grdimage [DEBUG]: Grid is considered to have a 360-degree longitude range.
grdimage [DEBUG]: Chosen boundary condition for left and right edges: geographic
grdimage [DEBUG]: Chosen boundary condition for bottom edge: natural
grdimage [DEBUG]: Chosen boundary condition for top    edge: geographic
grdimage [INFORMATION]: 359 (of 361) inconsistent grid values at North pole.
grdimage [INFORMATION]: Set boundary condition for top    edge: geographic
grdimage [INFORMATION]: Set boundary condition for bottom edge: natural
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [INFORMATION]: project grid files
grdimage [DEBUG]: Object ID 2 : Registered Grid Memory Reference 5593ed9c3af0 as an Input resource with geometry Surface [n_objects = 3]
grdimage [DEBUG]: Successfully duplicated a Grid
grdimage [DEBUG]: gmt_project_init: IN: Inc [0/0] n_columns/n_rows [361/180] dpi = 0 offset = 0
grdimage [DEBUG]: gmt_project_init: OUT: Inc [0/0] n_columns/n_rows [361/180] dpi = 0 offset = 0
grdimage [INFORMATION]: Grid projection from size 361x180 to 361x180
grdimage [DEBUG]: Successfully added data array to previously registered Grid container
grdimage [DEBUG]: gmt_grd_project: In [0/360/-89/90] and out [0/1/0/1]
grdimage [INFORMATION]: gmt_grd_project: Output grid extrema [-27.5593/114.589] exceed extrema of input grid [-1/1] due to resampling
grdimage [INFORMATION]: gmt_grd_project: Output grid clipped to input grid extrema
grdimage [DEBUG]: GMT_Destroy_Data: freed memory for a Grid for object 1
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 1 [n_objects = 2]
grdimage [DEBUG]: gmtlib_unregister_io: Object no 1 has non-NULL resource pointer
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 6 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: No current cpt file found
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 6 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: No current cpt file found
grdimage [DEBUG]: GMT: 0. Will try to find subdir=cpt stem = turbo suffix=.cpt
grdimage [DEBUG]: GMT: 1. gmt_getsharepath trying current dir
grdimage [DEBUG]: GMT: 2. gmt_getsharepath trying USERDIR /home/user/.gmt
grdimage [DEBUG]: GMT: 3. gmt_getsharepath trying USERDIR subdir /home/user/.gmt/cpt
grdimage [DEBUG]: GMT: 4. gmt_getsharepath trying SHAREDIR subdir /home/user/miniconda3/envs/pygmt/share/gmt/cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Given full path to file /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Found readable file /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Object ID 3 : Registered CPT File /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt as an Input resource with geometry Non-Geographical [n_objects = 3]
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_palette: Passed ID = 3 and mode = 8
grdimage [INFORMATION]: Reading CPT from File /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Reading CPT from /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [INFORMATION]: Auto-stretching CPT file turbo to fit data range -1 to 1
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 6 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: Object ID 4 : Registered CPT File /home/user/.gmt/sessions/gmt_session.3726/gmt.6.cpt as an Output resource with geometry Non-Geographical [n_objects = 4]
grdimage [DEBUG]: gmtapi_begin_io: Output resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_export_palette: Passed ID = 4 and mode = 0
grdimage [INFORMATION]: Write CPT to File /home/user/.gmt/sessions/gmt_session.3726/gmt.6.cpt
grdimage [DEBUG]: Writing CPT to /home/user/.gmt/sessions/gmt_session.3726/gmt.6.cpt
grdimage [DEBUG]: GMT_End_IO: Output resource access is now disabled
grdimage [INFORMATION]: Save current CPT file to /home/user/.gmt/sessions/gmt_session.3726/gmt.6.cpt !
grdimage [INFORMATION]: Evaluate image pixel colors
grdimage [INFORMATION]: Basic z(x,y) with optional illumination and no PostScript colormasking.
grdimage [INFORMATION]: Creating 24-bit color image
grdimage [DEBUG]: Current size of half-baked PS file /home/user/.gmt/sessions/gmt_session.3726/gmt_6.ps- = 55422.
grdimage (GMT_grdimage): tried to free unallocated memory
grdimage [DEBUG]: GMT_Destroy_Data: freed memory for a Grid for object 2
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 2 [n_objects = 3]
grdimage [DEBUG]: gmtlib_unregister_io: Object no 2 has non-NULL resource pointer
grdimage [DEBUG]: gmtlib_garbage_collection: Destroying object: C=0 A=0 ID=3 W=Input F=CPT M=File S=Used P=5593edb86930 N=/home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage (gmtsupport_free_range): tried to free unallocated memory
...
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage [DEBUG]: GMTAPI_Garbage_Collection freed 1 memory objects
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 3 [n_objects = 2]
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 4 [n_objects = 1]
grdimage (gmtlib_free_tmp_arrays): tried to free unallocated memory
```

</details>

seisman: I don't see any difference between macOS and Linux.

weiji14: Hold on, that error message was from one of the successful runs. It seems to work fine in ipython but fails when I run it using `make test`. Give me a couple of minutes.

weiji14: Ok, got the failing debug info (not sure why it's not printed in the CI). The main difference seems to be an `inf` value:
```
# good
grdimage [INFORMATION]: gmt_grd_project: Output grid extrema [-27.5593/114.589] exceed extrema of input grid [-1/1] due to resampling
```

```
# bad
grdimage [INFORMATION]: gmt_grd_project: Output grid extrema [-inf/114.589] exceed extrema of input grid [-inf/1] due to resampling
```

<details>

```
grdimage [DEBUG]: Map distance calculation will be using great circle approximation with authalic auxiliary latitudes and authalic (R_2) radius = 6371007.1809 m, in meter.
grdimage [DEBUG]: Map distance calculation will be using great circle approximation with authalic auxiliary latitudes and authalic (R_2) radius = 6371007.1809 m, in meter.
grdimage [DEBUG]: Projected grid is non-orthogonal, nonlinear, or dpi was changed
grdimage [INFORMATION]: Read header from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 1
grdimage [DEBUG]: Grid/Image dimensions imply w/e/s/n = 0/359/-89/90, inc = 1/1, gridline registration, n_layers = 1
grdimage [DEBUG]: Geographic input grid, longitudes span less than 360
grdimage [DEBUG]: Chosen boundary condition for all edges: geographic
grdimage [DEBUG]: Geographic input grid, longitude cells span exactly 360
grdimage [DEBUG]: Object ID 1 : Registered Grid Memory Reference 562abb2b9ce0 as an Input resource with geometry Surface [n_objects = 2]
grdimage [DEBUG]: Successfully created a new Grid container
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [DEBUG]: Projected values in meters: -9.00996e+06 9.00996e+06 -9.00996e+06 9.00996e+06
grdimage [INFORMATION]: Map scale is 7094.46 km per cm or 1:7.09446e+08.
grdimage [DEBUG]: Running in PS mode modern
grdimage [DEBUG]: Use PS filename /home/user/.gmt/sessions/gmt_session.6871/gmt_68.ps-
grdimage [DEBUG]: Create hidden PS file /home/user/.gmt/sessions/gmt_session.6871/gmt_68.ps-
grdimage [DEBUG]: Got session name as pygmt-session and default graphics formats as pdf
grdimage [INFORMATION]: Allocate and read data from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_grid: Passed ID = 0 and mode = 2
grdimage [INFORMATION]: Importing grid data from user matrix memory location
grdimage [DEBUG]: Grid is considered to have a 360-degree longitude range.
grdimage [DEBUG]: Chosen boundary condition for left and right edges: geographic
grdimage [DEBUG]: Chosen boundary condition for bottom edge: natural
grdimage [DEBUG]: Chosen boundary condition for top    edge: geographic
grdimage [INFORMATION]: 359 (of 361) inconsistent grid values at North pole.
grdimage [INFORMATION]: Set boundary condition for top    edge: geographic
grdimage [INFORMATION]: Set boundary condition for bottom edge: natural
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [INFORMATION]: project grid files
grdimage [DEBUG]: Object ID 2 : Registered Grid Memory Reference 562abb070e10 as an Input resource with geometry Surface [n_objects = 3]
grdimage [DEBUG]: Successfully duplicated a Grid
grdimage [DEBUG]: gmt_project_init: IN: Inc [0/0] n_columns/n_rows [361/180] dpi = 0 offset = 0
grdimage [DEBUG]: gmt_project_init: OUT: Inc [0/0] n_columns/n_rows [361/180] dpi = 0 offset = 0
grdimage [INFORMATION]: Grid projection from size 361x180 to 361x180
grdimage [DEBUG]: Successfully added data array to previously registered Grid container
grdimage [DEBUG]: gmt_grd_project: In [0/360/-89/90] and out [0/1/0/1]
grdimage [INFORMATION]: gmt_grd_project: Output grid extrema [-inf/114.589] exceed extrema of input grid [-inf/1] due to resampling
grdimage [INFORMATION]: gmt_grd_project: Output grid clipped to input grid extrema
grdimage [DEBUG]: GMT_Destroy_Data: freed memory for a Grid for object 1
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 1 [n_objects = 2]
grdimage [DEBUG]: gmtlib_unregister_io: Object no 1 has non-NULL resource pointer
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 68 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: No current cpt file found
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 68 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: No current cpt file found
grdimage [DEBUG]: GMT: 0. Will try to find subdir=cpt stem = turbo suffix=.cpt
grdimage [DEBUG]: GMT: 1. gmt_getsharepath trying current dir
grdimage [DEBUG]: GMT: 2. gmt_getsharepath trying USERDIR /home/user/.gmt
grdimage [DEBUG]: GMT: 3. gmt_getsharepath trying USERDIR subdir /home/user/.gmt/cpt
grdimage [DEBUG]: GMT: 4. gmt_getsharepath trying SHAREDIR subdir /home/user/miniconda3/envs/pygmt/share/gmt/cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Given full path to file /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: gmt_get_filename: In: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt Out: /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Found readable file /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Object ID 3 : Registered CPT File /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt as an Input resource with geometry Non-Geographical [n_objects = 3]
grdimage [DEBUG]: gmtapi_begin_io: Input resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_import_palette: Passed ID = 3 and mode = 8
grdimage [INFORMATION]: Reading CPT from File /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: Reading CPT from /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [DEBUG]: GMT_End_IO: Input resource access is now disabled
grdimage [INFORMATION]: Auto-stretching CPT file turbo to fit data range -inf to 1
grdimage [DEBUG]: gmtlib_get_graphics_item: Fig: 68 Subplot: 2 Panel: () Inset: 0
grdimage [DEBUG]: Object ID 4 : Registered CPT File /home/user/.gmt/sessions/gmt_session.6871/gmt.68.cpt as an Output resource with geometry Non-Geographical [n_objects = 4]
grdimage [DEBUG]: gmtapi_begin_io: Output resource access is now enabled [container]
grdimage [DEBUG]: gmtapi_export_palette: Passed ID = 4 and mode = 0
grdimage [INFORMATION]: Write CPT to File /home/user/.gmt/sessions/gmt_session.6871/gmt.68.cpt
grdimage [DEBUG]: Writing CPT to /home/user/.gmt/sessions/gmt_session.6871/gmt.68.cpt
grdimage [DEBUG]: GMT_End_IO: Output resource access is now disabled
grdimage [INFORMATION]: Save current CPT file to /home/user/.gmt/sessions/gmt_session.6871/gmt.68.cpt !
grdimage [INFORMATION]: Evaluate image pixel colors
grdimage [INFORMATION]: Basic z(x,y) with optional illumination and no PostScript colormasking.
grdimage [INFORMATION]: Creating 24-bit color image
PSL: Colormap of 2 colors created
PSL: Image depth reduced to 1 bits
PSL: DEFLATE compressed 8280 to 668 bytes (91.9% savings at compression level 5)
grdimage [DEBUG]: Current size of half-baked PS file /home/user/.gmt/sessions/gmt_session.6871/gmt_68.ps- = 22240.
grdimage (GMT_grdimage): tried to free unallocated memory
grdimage [DEBUG]: GMT_Destroy_Data: freed memory for a Grid for object 2
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 2 [n_objects = 3]
grdimage [DEBUG]: gmtlib_unregister_io: Object no 2 has non-NULL resource pointer
grdimage [DEBUG]: gmtlib_garbage_collection: Destroying object: C=0 A=0 ID=3 W=Input F=CPT M=File S=Used P=562abf53bc60 N=/home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage (gmtsupport_free_range): tried to free unallocated memory
...
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage (gmtsupport_free_range): tried to free unallocated memory
grdimage [DEBUG]: GMTAPI_Garbage_Collection freed 1 memory objects
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 3 [n_objects = 2]
grdimage [DEBUG]: gmtlib_unregister_io: Unregistering object no 4 [n_objects = 1]
grdimage (gmtlib_free_tmp_arrays): tried to free unallocated memory
```

</details>

seisman: Yes, I saw the similar *inf* when debugging the `shading=True` failure in https://github.com/GenericMappingTools/pygmt/issues/364. That's why I thought `GMT_IN` should work.

seisman: It could be another potential GMT bug, which is usually difficult to debugging from the PyGMT side.

weiji14: FYI, this is the output using `GMT_IS_DUPLICATE`, which generates the reddish-brown image above. Using `V="i"` for brevity, note that `-1.1949e+26` is printed instead of `-inf`:

```
grdimage [INFORMATION]: Read header from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [INFORMATION]: Map scale is 7094.46 km per cm or 1:7.09446e+08.
grdimage [INFORMATION]: Allocate and read data from file @GMTAPI@-S-I-G-M-G-N-000000
grdimage [INFORMATION]: Importing grid data from user matrix memory location
grdimage [INFORMATION]: 359 (of 361) inconsistent grid values at North pole.
grdimage [INFORMATION]: Set boundary condition for top    edge: geographic
grdimage [INFORMATION]: Set boundary condition for bottom edge: natural
grdimage [INFORMATION]: project grid files
grdimage [INFORMATION]: Grid projection from size 361x180 to 361x180
grdimage [INFORMATION]: gmt_grd_project: Output grid extrema [-1.1949e+26/114.589] exceed extrema of input grid [-1.1949e+26/1] due to resampling
grdimage [INFORMATION]: gmt_grd_project: Output grid clipped to input grid extrema
grdimage [INFORMATION]: Reading CPT from File /home/user/miniconda3/envs/pygmt/share/gmt/cpt/turbo.cpt
grdimage [INFORMATION]: Auto-stretching CPT file turbo to fit data range -1.1949e+26 to 1
grdimage [INFORMATION]: Write CPT to File /home/user/.gmt/sessions/gmt_session.12777/gmt.68.cpt
grdimage [INFORMATION]: Save current CPT file to /home/user/.gmt/sessions/gmt_session.12777/gmt.68.cpt !
grdimage [INFORMATION]: Evaluate image pixel colors
grdimage [INFORMATION]: Basic z(x,y) with optional illumination and no PostScript colormasking.
grdimage [INFORMATION]: Creating 24-bit color image
PSL: Colormap of 3 colors created
PSL: Image depth reduced to 2 bits
PSL: DEFLATE compressed 16380 to 779 bytes (95.2% savings at compression level 5)
```

pygmt/clib/conversion.py
```diff
@@ -101,14 +101,19 @@ def dataarray_to_matrix(grid):
...
+        )
         inc.append(coord_inc)
 
-    if any([i < 0 for i in inc]):  # Sort grid when there are negative increments
+    if any(i < 0 for i in inc):  # Sort grid when there are negative increments
         inc = [abs(i) for i in inc]
         grid = grid.sortby(variables=list(grid.dims), ascending=True)
 
-    matrix = as_c_contiguous(grid.values[::-1])
+    matrix = as_c_contiguous(grid[::-1].values)
```
weiji14, line 116: It seems like we practically **always** create a memory copy here because the numpy view on a flipped grid `grid[::-1]` is never (or hardly ever?) c_contiguous. I don't know if there's a way around this, but if we need to switch to use `GMT_IS_DUPLICATE` for stability in #517, wouldn't that mean we create 2 memory copies?!

weiji14: I'll try using [`xarray.DataArray.copy`](xarray.pydata.org/en/v0.16.0/generated/xarray.DataArray.copy.html) at 5a685b2fcf36768fccc345c92e4fab2afc19dbe3 to force a deep copy (that should return a c_contiguous array) before it enters the `as_c_contiguous` function (which may or may not return a numpy copy). It seems to give a more consistent black image, rather than the reddish-brown one :sweat_smile:

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,18 @@ def test_grdimage_fails():
     fig = Figure()
     with pytest.raises(GMTInvalidInput):
         fig.grdimage(np.arange(20).reshape((4, 5)))
+
+
+@pytest.mark.mpl_image_compare
+def test_grdimage_over_dateline(xrgrid):
```
weiji14, line 79: I'm starting to think that this test is exposing some [flaky tests](https://docs.pytest.org/en/stable/flaky.html) in our test suite (see also https://github.com/GenericMappingTools/pygmt/pull/327#issuecomment-534318735). The test passes pretty much all the time (?) when I just run `pytest --verbose --mpl pygmt/tests/test_grdimage.py` or even `pytest --verbose --mpl pygmt/tests/test_grd*.py`, but running `make test` causes the test to fail :confused:

I probably went down the wrong rabbit hole, but I've tried pretty much all of the [pytest plugins to test for flakiness](https://docs.pytest.org/en/stable/flaky.html#plugins) including pytest-xdist and pytest-flakefinder, and they pretty much agree that it's not `test_grdimage_over_dateline` which is the issue, it's one of the tests that precede it, but I'm having trouble tracking it down since the failures are all intermittent and semi-random.

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,19 @@ def test_grdimage_fails():
     fig = Figure()
     with pytest.raises(GMTInvalidInput):
         fig.grdimage(np.arange(20).reshape((4, 5)))
+
+
+@pytest.mark.runfirst
+@pytest.mark.mpl_image_compare
+def test_grdimage_over_dateline(xrgrid):
```
weiji14, line 80: Running `test_grdimage_over_dateline` first (before the other tests) makes the tests pass on Linux (see commit 349904708048842f3c74a01b4a68cc42d62f5d18 and bf03e5b6deacd2f9e1c547762a483940d6df9770). This is a hacky workaround for the flakiness I described at https://github.com/GenericMappingTools/pygmt/pull/476/files#r453429872.

pygmt/clib/conversion.py
weiji14: Reverted 5a685b2fcf36768fccc345c92e4fab2afc19dbe3 at 60c791de17382495c181bddc5959282a7c25db7d, will stick to just `grid[::-1].values` for now (i.e. flipping the grid left right in xarray rather than in numpy).

seisman:
Now all the tests pass. Does it mean this PR works?

weiji14:
> Now all the tests pass. Does it mean this PR works?

Short answer: Yes (and it has always worked for macOS).

Long answer: See https://github.com/GenericMappingTools/pygmt/pull/476#discussion_r455416772. The tests pass when I run `test_grdimage_over_dateline` **before** the other tests. On my Linux box, it works fine if I just call `grdimage` as normal in a script. But the `make test` fails consistently for some unknown reason.

Makefile
```diff
@@ -29,7 +29,8 @@ test:
 	@echo ""
 	@cd $(TESTDIR); python -c "import $(PROJECT); $(PROJECT).show_versions()"
 	@echo ""
-	cd $(TESTDIR); pytest $(PYTEST_ARGS) $(PROJECT)
```
seisman, line 32: Please add some comments here, explain why we need two pytest.

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,19 @@ def test_grdimage_fails():
     fig = Figure()
     with pytest.raises(GMTInvalidInput):
         fig.grdimage(np.arange(20).reshape((4, 5)))
+
+
+@pytest.mark.runfirst
```
seisman, line 78: Also add a brief comment here, explain why we need "runfirst".

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,19 @@ def test_grdimage_fails():
...
+@pytest.mark.mpl_image_compare
+def test_grdimage_over_dateline(xrgrid):
+    """
+    Ensure no gaps are plotted over the 180 degree international dateline.
+    Specifically checking that `xrgrid.gmt.gtype = 1` sets `GMT_GRID_IS_GEO`,
+    and that `xrgrid.gmt.registration = 0` sets `GMT_GRID_NODE_REG`. Note that
+    there would be a gap over the dateline if a pixel registered grid is used.
```
seisman, line 85: Add a link to issue #375?

pygmt/tests/test_grdimage.py
```diff
@@ -51,3 +73,19 @@ def test_grdimage_fails():
...
+    """
+    fig = Figure()
+    assert xrgrid.gmt.registration == 0  # gridline registration
+    xrgrid.gmt.gtype = 1  # geographic coordinate system
+    fig.grdimage(grid=xrgrid, region="g", projection="A0/0/1c", V="d")
```
seisman, line 90:
```suggestion
    fig.grdimage(grid=xrgrid, region="g", projection="A0/0/1c")
```

weiji14: I'll switch it to `V="i"` for now, so we still get some extra information.

seisman:
There is a pytest warning:
```
python3.7/site-packages/_pytest/mark/structures.py:327: PytestUnknownMarkWarning: Unknown pytest.mark.runfirst - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/latest/mark.html
    PytestUnknownMarkWarning,
```

weiji14:
Just for completeness, although this PR does fix #375, it doesn't fix @MarkWieczorek's reported bug at #390. I tried running the example but couldn't get the cylindrical 'Q' projection to plot in xarray (bottom row), and the Mollweide 'W' projection doesn't work too (middle row). Using a NetCDF file works fine though (top row). This results are the same when ran on the code in this PR or the current master branch at f401f8527cfee4dc8f48d78ad93d23b678ecfbc8, so it might be an upstream GMT 6.1 issue.

| test | test |
|:--:|:--:|
|This PR| Master branch at f401f8527cfee4dc8f48d78ad93d23b678ecfbc8 |

seisman:
I once tried to debug the issue in #390, it seems the boundary conditions are slightly different for an xarray input and a netCDF input. Mostly likely a GMT bug.

weiji14:
Yeah, the resampling doesn't work properly on virtualfiles perhaps. Will merge and try to track down the problem another time.

Issue #479, opened 2020-06-17
Figure.text: Parsing of additional arguments when plotting text from file


hemmelig:
**Description of the problem**

I am finding that adding text from a file isn't working as expected (based on the GMT documentation). The additional arguments are not parsed, instead the input files are treated as if there were only 3 input columns (x, y, text).

**Full code that generated the error**

You will need an additional file:
`echo 117,5.2,Helvetica,0,CM,SABAH > example.txt`
```python
fig = pygmt.Figure()
with pygmt.config(MAP_FRAME_TYPE="plain"):
    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
fig.coast(land="black", water="skyblue", N="1/2p,white")

fig.text(textfiles="example.txt", font="12p,Helvetica-Bold,white")

fig.show()
```

**Full error message**

![image](https://user-images.githubusercontent.com/10199651/84949523-49c70300-b0e5-11ea-9beb-cb56dd76bac5.png)

This actually appeared to be a problem with GMT 6 as well. Maybe I'm just missing something? I am interpreting it from: `This is one or more files containing 1 or more records with (x, y[, font, angle, justify], text).` If so, apologies. I also tried delimiting with spaces instead of commas to no avail.

**System information**

Please paste the output of `python -c "import pygmt; print(pygmt.__version__)"`:

```
v0.1.1
```

(Note: I had to change the above command, as pygmt.show_versions() didn't exist!)

seisman:
The [text documentation](https://docs.generic-mapping-tools.org/latest/text.html) says:
> This is one or more files containing 1 or more records with (x, y[, font, angle, justify], text). The attributes in brackets can alternatively be set directly via -F.

The *font*, *angle* and *justify* are not needed unless you add more modifiers to **-F** option. 

For example, if you don't use **-F** option, then your input data should be: `x y text`, e.g.,
```
echo 117 5.2 SABAH | gmt text -R108/120/-5/8 -JM20c -Ba -pdf map
```

If you use **-F+f+a+j**, then your input file format is: `x y font angle justify text`, e.g., 
```
echo 117 5.2 Helvetica 0 CM SABAH | gmt text -R108/120/-5/8 -JM20c -Ba -F+f+a+j -pdf map
```

hemmelig:
Ok - I understand now, though I do not think it is made adequately clear in the documentation that you must add this -F+f+a+j in order for it to work. 

[...]

PR #480, opened 2020-06-18
Add tutorial for pygmt.Figure.text

### LINKED ISSUE #469: Need examples for adding texts on maps

We don't have examples for adding texts on maps using `Figure.text()`. The example should be a Python script in the `examples/tutorials` directory. See the other Python scripts in that directory for reference.

hemmelig:
Add tutorial covering the basics of text annotation with PyGMT.
Follows the same format as the existing tutorials. This tutorial
covers the currently available suite of arguments in the PyGMT binding
to the GMT 6 program 'text'.

Fixes #469


examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+# Plot names of seas
+fig.text(text=["CELEBES SEA", "JAVA SEA"], x=[119, 112], y=[3.25, -4.6])
+fig.text(text="SULU SEA", x=119.12, y=7.25, angle=-40)
+fig.text(text="SOUTH CHINA SEA", x=112, y=6, angle=40)
+fig.text(text="MAKASSAR STRAIT", x=118.4, y=-1, angle=65)
+fig.show()
+
+########################################################################################
+# ``justify`` argument
+# --------------------
```
weiji14, line 80: I really like this `justify` plot. It would be nice to put it in the gallery since there are other GMT commands that use it (e.g. legend), but let's leave it here for now.

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+#
+# Text can be further configured by passing an argument corresponding to the
+# flag names in GMT, following the same convention as described in the GMT
+# documentation. It is hoped that over time more bindings to these arguments
+# will be written into PyGMT.
+
+fig = pygmt.Figure()
+fig.basemap(region=[0, 1, 0, 1], projection="X5c", frame="WSen")
+fig.text(text="Green", x=0.5, y=0.5, G="green")
```
weiji14, line 127: We're actually trying to document only the long-form arguments (see #473), and have made a start in #474. Just that we haven't released it yet, see differences in documentation between https://www.pygmt.org/dev/api/generated/pygmt.Figure.text.html and https://www.pygmt.org/v0.1.1/api/generated/pygmt.Figure.text.html#pygmt.Figure.text. I've been thinking whether we should cut a v0.1.2 release to incorporate this PR and other documentation changes, before GMT 6.1.0 comes out around end of June.

Personally I have nothing against using short aliases like this to achieve additional functionality (since I do it too!). Better way would be to add an alias for G=color here, see #385 for an example on how to do so.

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+# It is also possible to add annotations from a file containing `x`, `y`, and
+# `text` fields. Here we give a complete example.
+
+fig = pygmt.Figure()
+with pygmt.config(MAP_FRAME_TYPE="plain"):
+    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
+fig.coast(land="black", water="skyblue")
+
+# Plot region names / sea names
+fig.text(textfiles="examples.txt", font="22p,Helvetica-Bold,white")
```
weiji14, line 69: It would be nice to be able to show what the contents of the text file looks like. There's probably a quick way to generate a text file directly with python?

hemmelig: Yeah, something like:

```python
with open('examples.txt', 'w') as f:
    print('114,0.5,BORNEO.', file=f)
```

I did wonder about that. Is it worth including something to wipe the file at the end or just happy to leave it?

examples/tutorials/text.py
hemmelig: Yes, I was planning to collect some thoughts on how best to bring the rest of the available flags, or short-form arguments, into the PyGMT fold for `text` and submit an issue. 

So for now, should I remove this last section from the tutorial?

weiji14: Or we could do a quick PR to get some extra aliases into `text`. They're basically one-line additions (per alias). Do you have a list of what you want besides `color`?

examples/tutorials/text.py
weiji14: Ah cool, I didn't know you could `print()` to a file like that! Learn something new everyday :smile:

> I did wonder about that. Is it worth including something to wipe the file at the end or just happy to leave it?

Maybe try using [`tempfile`](https://docs.python.org/3/library/tempfile.html)? See https://stackoverflow.com/questions/5344287/create-read-from-tempfile for an example. Something like:

```python
with tempfile.NamedTemporaryFile():
    fig.text(...)
```

examples/tutorials/text.py
hemmelig: It would make sense to also add an alias for W (`pen`?) if adding one for G.

weiji14: > It would make sense to also add an alias for W (`pen`?) if adding one for G.

Yes, W=pen seems like another obvious choice.

> I find having `font` as an alias for F when +f also relates to `font` a bit ambiguous. There was also the point I raised in Issue #479 - when adding text from a text file, it is unclear if it's actually possible to emulate the way you tell GMT to expect `font`, `angle`, or `justify` fields in the file (`-F+f+a+j`).

Looking at the current v0.1.1 `text` implementation https://github.com/GenericMappingTools/pygmt/blob/809a888755c5dedc86ea74d0ae0497295b5311a4/pygmt/base_plotting.py#L864-L872

I think if you want to emulate `-F+f+a+j`, you could do `fig.text(..., font=True, angle=True, justify=True), and that would then use treat the columns in the textfile as font/angle/justify respectively. Unfortunately it won't be possible to change the order of the columns (e.g. `-F+a+j+f` I think.

But yes, you're right that it's an ambiguous beast. We could either document this particular case for GMT 'power' users, or find a better way to implement the code block above, bearing in mind that PyGMT is meant to be accessible to new users (see [Project Goals](https://www.pygmt.org/latest/#project-goals)) who might not be as familiar with the GMT suite. It would be nice to somewhat emulate what [`matplotlib.pyplot.text`](https://matplotlib.org/3.2.1/api/_as_gen/matplotlib.pyplot.text.html) does for example (i.e. have a `fontdict` like argument).

examples/tutorials/text.py
seisman: I personally prefer to codes like:
```
with open('examples.txt', 'w') as f:
    f.write('114 0.5 BORNEO.')
```
since `f.write()` is more commonly used than `print("xxx", file=f)` and `tempfile`, and more readable to new Python users.

I also prefer to have the input data space-separated. Although GMT supports many separators, e.g., spaces, tabs, and commas, commas are also used as the separator of font attributes, e.g., `22p,Helvetica-Bold,white`.

examples/tutorials/text.py
seisman: angle, font, and justify should also accept list of input, so that user can plot texts with varying angles, fonts or justification. It should be addressed in a separate issue/PR.

examples/tutorials/text.py
hemmelig: Sure, in defense of `print(..., file=f)`, it calls str() on whatever is passed in as a list so you can do `print(114, 0.5, "BORNEO", file=f). However, you're right that explicit is better than implicit, so:

```python

# Create file
with open("examples.txt", "w") as f:
    f.write("114 0.5 BORNEO")

# Plot region names / sea names
fig.text(textfiles="examples.txt", font="22p,Helvetica-Bold,white")

# Cleanup
os.remove("examples.txt")
```

examples/tutorials/text.py
hemmelig: I don't think that works:

```
> justify = True
> isinstance(justify, str)
False
```

etc

weiji14: Oh right, forgot that `justify` isn't an alias like `frame`, so the boolean doesn' convert to an empty string. In that case we could use `justify=""`?

hemmelig: Yes - works for `justify` and `font`, but wouldn't work for angle. Ideally though you would want to be able to set them in the same way?

hemmelig: To illustrate:

```python
# Create file
with open("examples.txt", "w") as f:
    print(114, 0.5, "22p,Helvetica-Bold,white", "LM", "BORNEO", file=f)

# Plot region names / sea names
fig.text(textfiles="examples.txt", font="", justify="")

# Cleanup
os.remove("examples.txt")
```

This works.

weiji14: > Yes - works for `justify` and `font`, but wouldn't work for angle. Ideally though you would want to be able to set them in the same way?

True, should allow `str` type for `angle` as well. I'll start a Pull Request to add some enhancements to `text`, give me a few minutes.

weiji14: Ok, I've started PR #481 to handle this, so `angle=""` should work once it's merged in to master. Also checked that `angle=True`, `font=True` and `justify=True` works (See test case at https://github.com/GenericMappingTools/pygmt/pull/481/files#diff-c20c076528e6988bc916b4bfc47e7a8d.
).  The documentation at https://www.pygmt.org/dev/api/generated/pygmt.Figure.text does mention that `bool` types are allowed. That would translate directly to `-F+a+f+j` to GMT.

examples/tutorials/text.py
weiji14: See also https://docs.generic-mapping-tools.org/latest/cookbook/features.html#placement-of-text (maybe link to it?).

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+for i, y_justify in enumerate(["T", "M", "B"]):
+    y_pos = 3.5 - i*1.5
+    fig.plot(x=[0., 4.], y=[y_pos, y_pos], W='3p,red@85')
+    for j, x_justify in enumerate(["L", "C", "R"]):
+        justify_text = x_justify + y_justify
+        x_pos = 0.5 + j*1.5
+        fig.text(text=justify_text, x=x_pos, y=y_pos,
+                 font="28p,Helvetica-Bold,black", justify=justify_text)
+        fig.plot(x=[x_pos, x_pos], y=[0., 4.], W="3p,red@85")
+fig.show()
```
weiji14, line 102: I've added two commits in #481 that could simplify this section at 9e7a3534b2ba5f5668cdd94db1954a554308d6c0 and 187a733d72fc552284bdad3d5c2b81ad5c6d878d. Instead of passing an x/y pair, we could use `position="TR"` to plot text on the Top Right. Something like this:

```python
fig = pygmt.Figure()
fig.basemap(region=[0, 5, 0, 2.5], projection="x1c", frame="a")
for position in ("TL", "TR", "BL", "BR"):
    fig.text(position=position, text=position, offset="j0.5c+v")
fig.show()
```
produces:

Four corner text plot

examples/tutorials/text.py
hemmelig: In which case, perhaps it would be best to wait until the above has been merged, so we can give the example as:

```python
fig = pygmt.Figure()
with pygmt.config(MAP_FRAME_TYPE="plain"):
    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
fig.coast(land="black", water="skyblue")

# Create file
with open("examples.txt", "w") as f:
    f.write("114 0.5 22p,Helvetica-Bold,white 0 CM BORNEO\n")
    f.write("119 3.25 12p,Helvetica-Bold,black 0 CM CELEBES SEA\n")
    f.write("112 -4.6 12p,Helvetica-Bold,black 0 CM JAVA SEA\n")
    f.write("112 6 12p,Helvetica-Bold,black 40 CM SOUTH CHINA SEA\n")
    f.write("119.12 7.25 12p,Helvetica-Bold,black -40 CM SULU SEA\n")
    f.write("118.4 -1 12p,Helvetica-Bold,black 65 CM MAKASSAR STRAIT\n")

# Plot region names / sea names
fig.text(textfiles="examples.txt", angle=True, justify=True, font=True)

# Cleanup
os.remove("examples.txt")

fig.show()
```

weiji14: Ok, we've merged in #481! If you can merge in the new changes from master into this branch (e.g. by clicking on the 'Update Branch' button), the new features/enhancement should be available and we can simplify some of the example code. Let us know if you need any help.

seisman:
@hemmelig Do you still have time working on this PR? These are all good `text` examples and should be merged into the documentation.

seisman:
No response from @hemmelig for 7 days. I'll take over this PR.

hemmelig:
What exactly needs finishing? I have been on fieldwork and moving home, apologies.

seisman:
@hemmelig No apologies. The code style checking is still failing (you should be able to run `make format` to update the code style automatically), and there are still a few review comments above that should be addressed.

seisman:
Other tests are also failing. It seems we introduced a bug in v0.2.0.

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+# A list of all recognised fonts can be found `here<https://docs.generic-mapping-tools.org/latest/cookbook/postscript_fonts.html>`_),
+# including details of how to use non-default fonts.
+
+fig = pygmt.Figure()
+with pygmt.config(MAP_FRAME_TYPE="plain"):
+    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
+fig.coast(land="black", water="skyblue")
+
+# Customising the font style
+fig.text(text="BORNEO", x="114.", y=0.5, font="22p,Helvetica-Bold,white")
```
seisman, line 52:
```suggestion
fig.text(text="BORNEO", x=114., y=0.5, font="22p,Helvetica-Bold,white")
```

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+#
+# Text annotations can be added to a map using the :meth:`text` method of the
+# :class:`pygmt.Figure`.
+#
+# Full details of the GMT6 command `text` can be found `here<https://docs.generic-mapping-tools.org/latest/text.html>`_.
+# The Python binding to this command is documented `here<https://www.pygmt.org/latest/api/generated/pygmt.Figure.text.html>`_.
+#
+# Here we create a simple map and add an annotation using the ``text``, ``x``,
+# and ``y`` arguments to specify the annotation text and position in the
+# projection frame. ``text``, ``x``, and ``y`` accept `int`, `str`, or `float`.
```
seisman, line 23: I'm not sure if it's a good idea to accept *x* and *y* in string-type.

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+# The anchor is specified with a two letter (order independent) code, chosen
+# from:
+# * Horizontal anchor: L(eft), C(entre), R(ight)
+# * Vertical anchor: T(op), M(iddle), B(ottom)
+
+fig = pygmt.Figure()
+fig.basemap(region=[0, 4, 0, 4], projection="X10c", frame="WSen")
+for i, y_justify in enumerate(["T", "M", "B"]):
+    y_pos = 3.5 - i*1.5
+    fig.plot(x=[0., 4.], y=[y_pos, y_pos], W='3p,red@85')
```
seisman, line 95:
```suggestion
    fig.plot(x=[0., 4.], y=[y_pos, y_pos], pen='3p,red@85')
```

examples/tutorials/text.py
```diff
@@ -0,0 +1,128 @@
...
+    for j, x_justify in enumerate(["L", "C", "R"]):
+        justify_text = x_justify + y_justify
+        x_pos = 0.5 + j*1.5
+        fig.text(text=justify_text, x=x_pos, y=y_pos,
+                 font="28p,Helvetica-Bold,black", justify=justify_text)
+        fig.plot(x=[x_pos, x_pos], y=[0., 4.], W="3p,red@85")
```
seisman, line 101:
```suggestion
        fig.plot(x=[x_pos, x_pos], y=[0., 4.], pen="3p,red@85")
```

examples/tutorials/text.py
hemmelig: Should I remove that part of the line, pending possibly disallowing strings to be passed in?

seisman: Yes.

weiji14:
I'll work on finalizing this @seisman, shouldn't let a good tutorial example sit here for so long!

examples/tutorials/text.py
weiji14:
```suggestion
# projection frame. ``text``, ``x``, and ``y`` accepts either `int` or `float`.
```

examples/tutorials/text.py
```diff
@@ -0,0 +1,133 @@
...
+        x_pos = 0.5 + j * 1.5
+        fig.text(
+            text=justify_text,
+            x=x_pos,
+            y=y_pos,
+            font="28p,Helvetica-Bold,black",
+            justify=justify_text,
+        )
+        fig.plot(x=[x_pos, x_pos], y=[0.0, 4.0], pen="3p,red@85")
+fig.show()
```
weiji14, line 107:
```suggestion
# * Vertical anchor: T(op), M(iddle), B(ottom)
# * Horizontal anchor: L(eft), C(entre), R(ight)

fig = pygmt.Figure()
fig.basemap(region=[0, 3, 0, 3], projection="X10c", frame=["WSne", "af0.5g"])
for position in ("TL", "TC", "TR", "ML", "MC", "MR", "BL", "BC", "BR"):
    fig.text(
        text=position,
        position=position,
        font="28p,Helvetica-Bold,black",
        justify=position,
    )
fig.show()
```

examples/tutorials/text.py
weiji14: Ok, I've simplified the code at commit cb9f7137518de2433f2be86a1a22d0e0bb8ac72a.

examples/tutorials/text.py
weiji14: The example might be another reason to support `textfile=io.StringIO` input as with #576, just so we don't need to explicitly create a temporary .txt file. But let's do this another time :wink:

examples/tutorials/text.py
weiji14: Ok, as per your code at https://github.com/GenericMappingTools/pygmt/pull/480/files#r443139292, I've updated the code at commit 20994aa0294dec11ca376dd4c2fac485fdb04e42

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+"""
+
+import os
+import pygmt
+
+###############################################################################
+# Basic map annotation
+# --------------------
+#
+# Text annotations can be added to a map using the :meth:`text` method of the
```
seisman, line 16:
```
:meth:`text`
```
is not linked to the method documentation.
```suggestion
# Text annotations can be added to a map using the :meth:`pygmt.Figure.text` method of the
```

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+# :class:`pygmt.Figure`.
+#
+# Here we create a simple map and add an annotation using the ``text``, ``x``,
+# and ``y`` arguments to specify the annotation text and position in the
+# projection frame. ``text`` accepts 'str' types, while ``x``, and ``y``
+# accepts either 'int' or 'float' types.
```
seisman, line 22: Please mention that they can also accept lists/arrays.

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+
+fig = pygmt.Figure()
+with pygmt.config(MAP_FRAME_TYPE="plain"):
+    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
```
seisman, line 26: As an introductory tutorial, we probably should avoid using unrelated methods `pygmt.config()`, although plain frames sometimes look better than "fancy" frames. But I'm also OK to keep `pygmt.config()` here.

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+
+# Plotting text annotations using lists of elements
+fig.text(text=["CELEBES SEA", "JAVA SEA"], x=[119, 112], y=[3.25, -4.6])
+
+fig.show()
+
+###############################################################################
+# Changing font style
+# -------------------
+# The size, family/weight, and colour of an annotation can be specified using
```
seisman, line 40: Let's use US spellings.
```suggestion
# The size, family/weight, and color of an annotation can be specified using
```

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+# the ``font`` argument.
+#
+# A list of all recognised fonts can be found at
+# :gmt-docs:`cookbook/postscript_fonts.html`, including details of how to use
+# non-default fonts.
+
+fig = pygmt.Figure()
+with pygmt.config(MAP_FRAME_TYPE="plain"):
+    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
```
seisman, line 49: Same as above for `pygmt.config()`.

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+###############################################################################
+# Plotting from a text file
+# -------------------------
+#
+# It is also possible to add annotations from a file containing `x`, `y`, and
+# `text` fields. Here we give a complete example.
+
+fig = pygmt.Figure()
+with pygmt.config(MAP_FRAME_TYPE="plain"):
+    fig.basemap(region=[108, 120, -5, 8], projection="M20c", frame="a")
```
seisman, line 66: Same as above for `pygmt.config()`.

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+with open("examples.txt", "w") as f:
+    f.write("114 0.5 0 22p,Helvetica-Bold,white CM BORNEO\n")
+    f.write("119 3.25 0 12p,Helvetica-Bold,black CM CELEBES SEA\n")
+    f.write("112 -4.6 0 12p,Helvetica-Bold,black CM JAVA SEA\n")
+    f.write("112 6 40 12p,Helvetica-Bold,black CM SOUTH CHINA SEA\n")
+    f.write("119.12 7.25 -40 12p,Helvetica-Bold,black CM SULU SEA\n")
+    f.write("118.4 -1 65 12p,Helvetica-Bold,black CM MAKASSAR STRAIT\n")
+
+# Plot region names / sea names from a text file
+fig.text(textfiles="examples.txt", angle=True, font=True, justify=True)
```
seisman, line 79: Please add a few comments explaining the meanings of `angle=True, font=True, justify=True` and columns of the input file.

examples/tutorials/text.py
```diff
@@ -0,0 +1,139 @@
...
+###############################################################################
+# ``angle`` argument
+# ------------------
+# ``angle`` is an optional argument used to specify the clockwise rotation of
+# the text from the horizontal.
+
+fig = pygmt.Figure()
+fig.basemap(region=[0, 4, 0, 4], projection="X10c", frame="WSen")
+for i in range(0, 360, 30):
+    fig.text(text=f"`          {i} Degrees", x=2, y=2, justify="LM", angle=i)
```
seisman, line 119: `@.` is a shorthand for the degree symbol.
```suggestion
    fig.text(text=f"`          {i}@.", x=2, y=2, justify="LM", angle=i)
```

image

examples/tutorials/text.py
weiji14: I think it's ok to keep it here for now, this might reduce the number of 'fancy' frames out there in the world :laughing:

examples/tutorials/text.py
weiji14: Something like this? Feel free to proofread before I batch commit this.
```suggestion
# Create space-delimited file
with open("examples.txt", "w") as f:
    f.write("114 0.5 0 22p,Helvetica-Bold,white CM BORNEO\n")
    f.write("119 3.25 0 12p,Helvetica-Bold,black CM CELEBES SEA\n")
    f.write("112 -4.6 0 12p,Helvetica-Bold,black CM JAVA SEA\n")
    f.write("112 6 40 12p,Helvetica-Bold,black CM SOUTH CHINA SEA\n")
    f.write("119.12 7.25 -40 12p,Helvetica-Bold,black CM SULU SEA\n")
    f.write("118.4 -1 65 12p,Helvetica-Bold,black CM MAKASSAR STRAIT\n")

# Plot region names / sea names from a text file, where
# the longitude (x) and latitude (y) coordinates are in the first two columns.
# Setting angle/font/justiry to True will indicate that those columns are
# present in the text file too (Note: must be in that order!).
# Finally, the text to be printed will be in the last column
fig.text(textfiles="examples.txt", angle=True, font=True, justify=True)
```

seisman: Looks good.

examples/tutorials/text.py
weiji14: Personally I prefer the British spelling, but ok to go with this, if only because we use `color` as a long alias argument in many of our PyGMT modules.

seisman (approved):

weiji14:
Thanks for working on this @hemmelig!! The new documentation is now at https://www.pygmt.org/dev/tutorials/text.html :smile:

Issue #247, opened 2018-11-26
Arguments with spaces in strings


seisman:
**Description of the problem**

**Full code that generated the error**

I tried to plot a basemap with an axis labelled as "Crust Age". I tried in three different ways:
```python
import gmt

fig = gmt.Figure()
# 1st way: FAIL
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B='x20+lCrust Age')
# 2nd way: SUCCEED
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B='x20+l"Crust Age"')
# 3rd way: FAIL
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B="x20+l'Crust Age'")
fig.savefig("map.pdf")
```

**Full error message**

```
psbasemap [ERROR]: Unrecognized option -<
Traceback (most recent call last):
  File "test.sh", line 6, in <module>
    fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B="x20+l'Crust Age'")
  File "/Users/seisman/Gits/gmt/gmt-python/gmt/helpers/decorators.py", line 199, in new_module
    return module_func(*args, **kwargs)
  File "/Users/seisman/Gits/gmt/gmt-python/gmt/helpers/decorators.py", line 294, in new_module
    return module_func(*args, **kwargs)
  File "/Users/seisman/Gits/gmt/gmt-python/gmt/base_plotting.py", line 472, in basemap
    lib.call_module("basemap", build_arg_string(kwargs))
  File "/Users/seisman/Gits/gmt/gmt-python/gmt/clib/session.py", line 490, in call_module
    module, status, self._error_message
gmt.exceptions.GMTCLibError: Module 'basemap' failed with status code 71:
psbasemap [ERROR]: Unrecognized option -<
```

The 1st way is equivalent to command line

    gmt basemap -R0/100/0/50 -JX10c/5c -Bx20+lCrust Age 

The command fails because "Age" is thought to be a file. However, the error message is almost useless to help users find the error. Do you think it's technically possible to automatically add quotes around strings with spaces if users forget to do that?

The 2nd way works as expected. However, I don't understand why the 3rd way fails? The equivalent cmd version works:
```
gmt basemap -R0/100/0/50 -JX10c/5c -Bx20+l'Crust Age' -pdf map
```

**System information**

* Operating system: macOS
* Python installation (Anaconda, system, ETS): Anaconda
[...]

PR #481, opened 2020-06-19
Enhance text with extra functionality and aliases

weiji14:

Adding extra functionality to [`text`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.text), including aliases that were missing in initial implementation at #321. In support of the text tutorial at #480.  Preview documentation at https://pygmt-git-enhance-text.gmt.vercel.app/api/generated/pygmt.Figure.text.html

TODO:
- [x] Allow setting `angle=True` so that users can parse the angle from a textfile (in GMT it will be `-F+a`. (4c5261381769b42f6914dfd96893780a7c9bea31)
- [x] Alias [W](https://docs.generic-mapping-tools.org/latest/text.html#w)=pen (138617f4940f9d4192c19c6206e009b7883d27ac)
- [x] Alias [C](https://docs.generic-mapping-tools.org/latest/text.html#c)=clearance (45a9858a27b92f55c35c1169200f96b512f08997)
- [x] Alias [G](https://docs.generic-mapping-tools.org/latest/text.html#g)=color/fill (9b8e837572da8e560295abd52d810a30f4d14872)
- [x] Alias [-F+c](https://docs.generic-mapping-tools.org/latest/text.html#f)=position (9e7a3534b2ba5f5668cdd94db1954a554308d6c0)
- [x] Alias [D](https://docs.generic-mapping-tools.org/latest/text.html#d)=offset (187a733d72fc552284bdad3d5c2b81ad5c6d878d)
- [ ] etc

I'd be quite keen to enable text placement without setting an x or y by using `position` (`-F+c` in GMT) to get the x/y from the map frame's region. For example, the GMT gallery example at https://docs.generic-mapping-tools.org/latest/gallery/ex40.html does `gmt text -Dj0.1i/0.1i -F+cLT+jTL+f18p+t"T = 100 km"`to print the text at the top-left (TL) position in the map frame.

Fixes #

**Reminders**


seisman:
These lines convert a list or a tuple into comma-separated strings. They don't make sense here.
https://github.com/GenericMappingTools/pygmt/blob/0267dd15c7626f5903cdb5a3e12ca545d70c884a/pygmt/base_plotting.py#L820-L822

weiji14:
> These lines convert a list or a tuple into comma-separated strings. They don't make sense here.
> 
> https://github.com/GenericMappingTools/pygmt/blob/0267dd15c7626f5903cdb5a3e12ca545d70c884a/pygmt/base_plotting.py#L820-L822

True. I can understand the rationale for `font` (e.g. doing `font=("12p", "Helvetica", "green")`, but not sure why I used it for `angle` and `justify`. Anyway, if we're going to support passing in an array or list of angle/font/justify, this probably has to go.

pygmt/base_plotting.py
```diff
@@ -813,7 +813,7 @@ def legend(self, spec=None, position="JTR+jTR+o0.2c", box="+gwhite+p1p", **kwarg
             lib.call_module("legend", arg_str)
 
     @fmt_docstring
-    @use_alias(R="region", J="projection", B="frame")
+    @use_alias(R="region", J="projection", B="frame", W="pen")
```
weiji14, line 816: How should we alias `G`? [I initially thought `color` would be good](https://github.com/GenericMappingTools/pygmt/pull/480#discussion_r442518040), but GMT.jl uses `fill` (see https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.text). Maybe `fill` then, since color could be confused with the text color?

seisman: `fill` is good to me.

pygmt/base_plotting.py
```diff
@@ -855,25 +855,50 @@ def text(
...
+            get a straight rectangle [Default]. Append upper case 'O' to get a
+            rounded rectangle. In paragraph mode (*paragraph*) you can also
+            append lower case 'c' to get a concave rectangle or append upper
+            case 'C' to get a convex rectangle.
+        fill : str
+            Sets the shade or color used for filling the text box [Default is
+            no fill]. Alternatively, use fill='c' to plot the text and then use
+            the text dimensions (and *clearance*) to build clip paths and turn
+            clipping on. This clipping can then be turned off later with clip
+            -C. To not plot the text but activate clipping, use fill='C'
```
weiji14, line 896:
```suggestion
            no fill].
```

Probably should remove this section? Since [`clip`](https://docs.generic-mapping-tools.org/latest/clip.html) isn't implemented yet, and I'm not even sure if I understand what it does. Or we can just keep it in case someone does need to use it.

seisman: The text clipping feature won't work without the **clip** function. I think it's OK to remove this part.

weiji14: Ok done. Technically users can still use `fill='c'` as we're just not documenting it. But best not to confuse people what the `clip` thing is about until it is implemented.

pygmt/base_plotting.py
EJFielding: Fill is much better. (I have been lurking here.)

weiji14: Cool, feel free to suggest any other changes :wink:

pygmt/base_plotting.py
```diff
@@ -826,20 +834,24 @@ def text(
...
         font=None,
         justify=None,
         **kwargs,
     ):
         """
-        Plot or typeset text on maps
+        Plot or typeset text strings of variable size, font type, and
+        orientation.
 
         Used to be pstext.
```
seisman, line 849: It's OK to remove this line.
```suggestion
```

pygmt/base_plotting.py
```diff
@@ -853,42 +865,102 @@ def text(
...
+            font size. Optionally, use modifier '+t' to set the shape of the
+            textbox when using *fill* and/or *pen*. Append lower case 'o' to
+            get a straight rectangle [Default]. Append upper case 'O' to get a
+            rounded rectangle. In paragraph mode (*paragraph*) you can also
+            append lower case 'c' to get a concave rectangle or append upper
+            case 'C' to get a convex rectangle.
+        fill : str
+            Sets the shade or color used for filling the text box [Default is
+            no fill].
+            instead.
```
seisman, line 917:
```suggestion
```

pygmt/base_plotting.py
```diff
@@ -853,42 +865,102 @@ def text(
...
+                )
+            kind = "vectors"
+
         if kind == "vectors" and text is None:
-            raise GMTInvalidInput("Must provide text with x and y.")
+            raise GMTInvalidInput("Must provide text with x/y pairs or position")
         if kind == "file":
             for textfile in textfiles.split(" "):  # ensure that textfile(s) exist
                 if not os.path.exists(textfile):
                     raise GMTInvalidInput(f"Cannot find the file: {textfile}")
```
seisman, line 950: We shouldn't check if text files exist since GMT not only search input files in the current directory but also search for files in the `~/.gmt`, `~/.gmt/cache` and `~/.gmt/server` directories.

pygmt/tests/test_text.py
```diff
@@ -96,6 +97,49 @@ def test_text_nonexistent_filename():
...
 
+@pytest.mark.mpl_image_compare
+def test_text_position_four_corners(region):
+    """
+    Print text at four corners (top left/right, bottom left/right) of map.
+    """
+    fig = Figure()
+    for position in ("TL", "TR", "BL", "BR"):
+        fig.text(
+            region=region, projection="x1c", frame="a", position=position, text=position
```
seisman, line 108: Two issues here:

1. Every time `frame='a'` is used, a basemap/frame is plotted. Here, the same basemap is plotted 4 times. It's not a big issue here, but if a basemap is plotted hundreds of times, the frame will look blurry. It's better to use the `basemap()` function to plot a basemap before calling `text()`.

2. The baseline image is incorrect. Here is what I'm getting using the equivalent command-line version:
```
gmt begin map png
echo TL | gmt text -Jx1c -R0/5/0/2.5 -Ba -F+cTL
echo TR | gmt text -Jx1c -R0/5/0/2.5 -Ba -F+cTR
echo BL | gmt text -Jx1c -R0/5/0/2.5 -Ba -F+cBL
echo BR | gmt text -Jx1c -R0/5/0/2.5 -Ba -F+cBR
gmt end show
```
Note the differences in **TL** and **BL**.
image

pygmt/tests/test_text.py
```diff
@@ -96,6 +97,49 @@ def test_text_nonexistent_filename():
...
+    """
+    Print text at four corners (top left/right, bottom left/right) of map,
+    offset by 0.5 cm, with a line drawn from the original to the shifted point.
+    """
+    fig = Figure()
+    for position in ("TL", "TR", "BL", "BR"):
+        fig.text(
+            region=region,
+            projection="x1c",
+            frame="a",
```
seisman, line 135: Same here. Use a `basemap()` call instead.

pygmt/tests/test_text.py
```diff
@@ -96,6 +97,49 @@ def test_text_nonexistent_filename():
...
+    """
+    fig = Figure()
+    with pytest.raises(GMTInvalidInput):
+        fig.text(
+            region=region, projection="x1c", x=1.2, y=2.4, position="MC", text="text"
+        )
+
+
+@pytest.mark.mpl_image_compare
+def test_text_position_offset_with_line(region):
```
seisman, line 125: Again, **TL** and **BL** are slightly shifted.

seisman:
> textfiles : str or list
>      A text data file name, or a list of filenames containing 1 or more
>      records with (x, y[, font, angle, justify], text).

The current implement only allows **-F+a+f+j**, so the input format should be 
```
x, y[, angle, font, justify], text
```

pygmt/base_plotting.py
weiji14: You're right, I forgot about the `@...` files. I feel like there was a reason I added this in commit 7a08ebd7a5d1b185f31b84265617d1816d2d8610, but can't recall now.

pygmt/tests/test_text.py
weiji14: >     1. Every time `frame='a'` is used, a basemap/frame is plotted. Here, the same basemap is plotted 4 times. It's not a big issue here, but if a basemap is plotted hundreds of times, the frame will look blurry. It's better to use the `basemap()` function to plot a basemap before calling `text()`.

Good point. I would still rather have the unit tests use `text` only, so I'll add a centre-middle (CM) text to set up the plot and then loop through the four corners.

>     2. The baseline image is incorrect. Here is what I'm getting using the equivalent command-line version:
> Note the differences in **TL** and **BL**.

I did wonder why the left text were offset by a bit. I'll recreate/clean up my dev environment and try generating the images again.

weiji14:
> The current implementation only allows **-F+a+f+j**, so the input format should be
> 
> ```
> x, y[, angle, font, justify], text
> ```

Done. I seem to recall reading somewhere (was it upstream in GMT?) that the reason was because we want all the numerical values first, and then the string values after. Can't seem to track down that issue/comment though :/

pygmt/tests/test_text.py
weiji14: Ok, I've fixed point 1 (not plotting the frame 4 times), but still having trouble with point 2. I can reproduce your plot using the bash script, but somehow when using pygmt, 'TL' and 'BL' are still offset slightly. The thing is, I don't see that test failing in our CI either :confused:

pygmt/base_plotting.py
weiji14: Ok done, a missing file will now raise a GMTClibError instead of GMTInvalidInput as before:

```python
pygmt.exceptions.GMTCLibError: Module 'text' failed with status code 71:
text [ERROR]: Error for input file: No such file (notexist.txt)
```

so removing the `os.path.exists()` check should be fine.

pygmt/base_plotting.py
```diff
@@ -849,70 +858,128 @@ def text(
...
+                                "y": np.atleast_1d(y),
+                                "text": np.atleast_1d(text),
+                            }
+                        ).to_csv(
+                            tmpfile.name,
+                            sep="\t",
+                            header=False,
+                            index=False,
+                            quoting=csv.QUOTE_NONE,
+                        )
```
weiji14, line 982:
```suggestion
                        file_context = lib.virtualfile_from_vectors(
                            np.atleast_1d(x), np.atleast_1d(y), np.atleast_1d(text)
                        )
```

@leouieda mentioned before at https://github.com/GenericMappingTools/pygmt/pull/321#discussion_r334993927 to use virtualfiles, and I finally figured out what he meant. Unfortunately, `virtualfile_from_vectors` doesn't yet support `str` types. This is the error on `_check_dtype_and_dim`:

```python-traceback
        if array.dtype.name not in DTYPES:
            raise GMTInvalidInput(
>               "Unsupported numpy data type '{}'.".format(array.dtype.name)
            )
E           pygmt.exceptions.GMTInvalidInput: Unsupported numpy data type 'str864'.

```

From what I can tell, we'll need to use [`GMT_Put_Strings`](https://docs.generic-mapping-tools.org/latest/api.html?highlight=gmt_char#gmt-put-strings) is that right? We should do this refactor in a separate PR, this one is already getting a bit too long.

seisman: Yes, as I understand it, we need to use `GMT_Put_Strings` to pass trailing strings to GMT.

pygmt/tests/test_text.py
```diff
@@ -92,10 +93,48 @@ def test_text_nonexistent_filename():
...
 
 
+@pytest.mark.mpl_image_compare
+def test_text_position(region):
+    """
+    Print text at center middle and four corners (top left/right, bottom
+    left/right) of map.
+    """
+    fig = Figure()
+    fig.text(region=region, projection="x1c", frame="a", position="CM", text="C M")
```
seisman, line 107: 1. Why do you put a space between **C** and **M**?
2. It's better to also test **ML**, **MR**, **TC** and **BC**.

pygmt/tests/test_text.py
```diff
@@ -92,10 +93,48 @@ def test_text_nonexistent_filename():
...
+    for position in ("TL", "TR", "BL", "BR"):
+        fig.text(frame="a", position=position, text=position)
```
seisman, line 109: `frame="a"` is not needed.
```suggestion
        fig.text(position=position, text=position)
```

pygmt/tests/test_text.py
```diff
@@ -92,10 +93,48 @@ def test_text_nonexistent_filename():
...
+@pytest.mark.mpl_image_compare
+def test_text_position_offset_with_line(region):
+    """
+    Print text at centre middle and  four corners (top left/right, bottom
+    left/right) of map, offset by 0.5 cm, with a line drawn from the original
+    to the shifted point.
+    """
+    fig = Figure()
+    fig.text(region=region, projection="x1c", frame="a", position="CM", text="C M")
+    for position in ("TL", "TR", "BL", "BR"):
```
seisman, line 133: Same as above:

1. Why do you put a space between C and M?
2. It's better to also test ML, MR, TC and BC.

pygmt/tests/test_text.py
weiji14: >     1. Why do you put a space between **C** and **M**?

It's to make sure that we don't hit #247.

>     2. It's better to also test **ML**, **MR**, **TC** and **BC**.

Ok.

pygmt/tests/test_text.py
```diff
@@ -78,6 +78,16 @@ def test_text_input_single_filename():
     return fig
 
 
+@pytest.mark.mpl_image_compare
+def test_text_input_external_filename():
+    """
+    Run text by passing in an external filename to textfiles
```
seisman, line 84: GMT calls `@filename` remote files.
```suggestion
def test_text_input_remote_file():
    """
    Run text by passing in an remote file to textfiles
```

weiji14: Ah yes, I'll change that. Coming up with names are hard :laughing:

weiji14:
Great, thanks for taking the time to review this so thoroughly @seisman! I'll open a new issue for tackling those future improvements.

PR #484, opened 2020-06-21
Unpin pylint version and fix some lint warnings

weiji14:

Not sure why pylint was pinned to 2.2.2 in #383, but it's broken and we'll need to update to 2.3.0 or newer (see https://github.com/PyCQA/astroid/issues/649). Using 2.2.2 will give the following error:

```
 AttributeError: 'Import' object has no attribute 'infer_name_module'
```

Will also fix some lint warnings in this PR. I've refreshed the `.pylintrc` file for pylint v2.5.3 by running:

```
pylint --generate-rcfile > .pylintrc
```

and then manually kept some old PyGMT specific settings from earlier PRs (e.g. #73).
- jobs=4
- max-module-lines=2000
- max-args=10
- max-attributes=10
- good-names=i,j,k,ex,Run,_,w,e,s,n,x,y,z

Also added back 'attribute-defined-outside-init' and 'bad-continuation' to the disable list, the latter is to fit in with black (see https://github.com/psf/black/issues/48).

References:
- http://pylint.pycqa.org/en/2.5/user_guide/message-control.html

Fixes #

**Reminders**


seisman:
https://github.com/PyCQA/astroid/issues/649#issuecomment-467825639

> The reason we pin pylint is that new versions often introduce new types of errors, and we don't want our CI to break due to these new errors. For example 2.3.0 introduces a new "no-else-raise" error.

I guess @leouieda pinned pylint for the same reason.

pygmt/helpers/utils.py
```diff
@@ -170,7 +170,7 @@ def is_nonstr_iter(value):
 
     """
     try:
-        [item for item in value]  # pylint: disable=pointless-statement
```
seisman, line 173: It's possible to simplify the function using this python code
```python
from collections.abc import Iterable

return isinstance(value, Iterable) and not isinstance(value, str)
```
See https://stackoverflow.com/a/61139278

seisman:
Please merge the master branch in and see if the CI pass (still 43 failures).

pygmt/helpers/utils.py
weiji14: Ok, done!

seisman:
There is still one warning for me:
```
pylint pygmt setup.py
************* Module pygmt
pygmt/__init__.py:42:4: C0415: Import outside toplevel (clib) (import-outside-toplevel)
```
but I'm using pylint 2.4.4

.pylintrc
```diff
@@ -371,6 +536,15 @@ min-public-methods=2
 
 [IMPORTS]
 
+# List of modules that can be imported at any level, not just the top level
+# one.
+allow-any-import-level=clib.Session,
```
weiji14, line 541:
```suggestion
allow-any-import-level=clib,
```
Maybe I should change this?

seisman: It's all up to you, as I know very little about pylint.

weiji14: Lemme downgrade my pylint and see if it works.

weiji14: Then it breaks pylint 2.5.x... I've allowed for any import level for both `clib` and `clib.Session` so that pylint 2.4 and pylint 2.5 will work. Will merge this branch once the CI passes.

PR #490, opened 2020-06-24
Fix a bug when passing data to GMT in Session.open_virtual_file()

### LINKED ISSUE #406: plot crashes when plotting arrows

I'm trying to plot two vectors using the following script. It works well if pen is set to `1p,black` (i.e. a simple line), but crashes if set to `1p,black+ve0.2c`. Actually, the first **plot** works, but the second one crashes.

```python
import pygmt
import numpy as np

fig = pygmt.Figure()
fig.basemap(projection="X15c/10c", region="400/600/5/8", frame=True)
fig.plot(x=[530.0, 530.0], y=[6.0, 7.0], pen='1p,black+ve0.2c')
fig.plot(x=[570.0, 570.0], y=[6.0, 7.0], pen='1p,black+ve0.2c')
fig.savefig("map.pdf")
```

**Full error message**

```
python(64354,0x11ac3bdc0) malloc: *** error for object 0x7f8f28e50900: pointer being freed was not allocated
python(64354,0x11ac3bdc0) malloc: *** set a breakpoint in malloc_error_break to debug
[1]    64354 abort      python test.py
```

**System information**

* Operating system: macOS
* Version of GMT:  6.1.0_e6eb3b9_2020.03.30
* Version of Python: 3.7.4
* Version of this package: 0.0.1a0+75.g088b0d1
* If using conda, paste the output of `conda list` below:

<details>
<summary>output of conda list</summary>
<pre>
PASTE OUTPUT OF CONDA LIST HERE
</pre>
</details>

seisman:

External programs like PyGMT can pass dataset/momory to GMT. By default,
GMT can read, modify and free the momery, which sometimes can cause
crashes.

Issue #406 reports an example in which PyGMT crashes. The issue was reported
to the upstream (see https://github.com/GenericMappingTools/gmt/issues/3515
and https://github.com/GenericMappingTools/gmt/pull/3528). It turns out
to be an API user error (i.e., a PyGMT bug).

As per the explanation of Paul, external programs like PyGMT should
always use `GMT_IN|GMT_IS_REFERENCE` to tell GMT that the data is
read-only, so that GMT won't try to change and free the memory.

This PR makes the change from `GMT_IN` to `GMT_IN|GMT_IS_REFERENCE`
in the `Session.open_virtual_file()` function, updates a few docstrings,
and also adds the script in #406 as a test.

Fixes #406. 

GMT C API references: https://docs.generic-mapping-tools.org/latest/api.html#resource-registration

**Reminders**


pygmt/clib/session.py
```diff
@@ -922,6 +922,9 @@ def open_virtual_file(self, family, geometry, direction, data):
         direction : str
             Either ``'GMT_IN'`` or ``'GMT_OUT'`` to indicate if passing data to
             GMT or getting it out of GMT, respectively.
+            By default, GMT can modify the data you pass in. Add modifier
+            ``'GMT_IS_REFERENCE'`` to tell GMT the data are read-only, or
+            ``'GMT_IS_DUPLICATE'' to tell GMT to duplicate the data.
```
weiji14, line 927: Was thinking if we should expose `GMT_IS_DUPLICATE` as an option in PyGMT. Doesn't seem to be an obvious use case from reading https://github.com/GenericMappingTools/gmt/pull/3528, so maybe leave it for now.

pygmt/clib/session.py
```diff
@@ -1086,7 +1089,9 @@ def virtualfile_from_vectors(self, *vectors):
         for col, array in enumerate(arrays):
             self.put_vector(dataset, column=col, vector=array)
 
-        with self.open_virtual_file(family, geometry, "GMT_IN", dataset) as vfile:
+        with self.open_virtual_file(
+            family, geometry, "GMT_IN|GMT_IS_REFERENCE", dataset
```
weiji14, line 1093: It's interesting that `virtualfile_from_grid()` function already uses `GMT_IN|GMT_IS_REFERENCE` but the `_vectors` one doesn't. Anyways, just a comment.

pygmt/clib/session.py
seisman: It's not clear to me when we should or need to use `GMT_IS_DUPLICATE`, but I think we can leave it for now.

Issue #489, opened 2020-06-23
Redesign of the load_earth_relief() function


seisman:
**Description of the problem**

PyGMT provides the [`load_earth_relief()`](https://github.com/GenericMappingTools/pygmt/blob/86c46f075d6f68a1367bc0fad161f98ba99c14c3/pygmt/datasets/earth_relief.py#L11) function to load GMT's earth_relief data as a `xarray.DataArray` object, so that users can plot it or directly manipulate the data. The function won't work in GMT 6.1.0.

Currently, the `load_earth_relief()` function first calls the [`which()`](https://github.com/GenericMappingTools/pygmt/blob/86c46f075d6f68a1367bc0fad161f98ba99c14c3/pygmt/datasets/earth_relief.py#L38) function to get the full path of the data (`which` will download the data if it's not available locally), and then read the netCDF file in by `grid = xr.open_dataarray(fname)`. It works well with GMT 6.0.0 because each resolution is stored as a single grid.

In GMT 6.1.0, we will have many problems:

1. earth relief data with resolutions >= 06m are stored as a single grid, e.g., `earth_relief_06m_p.grd`. The function still works
2. earth relief data with resolutions <= 05m are stored as several smaller tiles. If you want to plot an earth relief map in a 3°x3° region using the 03s data, you don't have to download the ~6.8 GB large file anymore, i.e., the command
`gmt grdimage @earth_relief_03s -R120/123/30/33 -pdf map` will only need to download 8 tiles (~3 MB) instead of the old 6.8 GB single grid. However, `gmt which @earth_relief_03s -Ga` (`-Ga` means automatically putting the data in the appropriate directory, which is new in GMT 6.1.) will download all tiles (hundreds or thousands of tiles, ~6.8 GB in total). That's definitely what we want to avoid.

**Solutions**

AFAIK, there are no easy solutions. 

1. We can do `gmt grdinfo @earth_relief_03s -R120/123/30/33` to download the tiles (ignore the output from `grdinfo`) and use `gmt which` to return the full paths that exist locally (it will return all existing tiles, including tiles outside our target region). However, since the data are stored as tiles, we have to read several files and merge them. It's more complicated for 03s and 01s data. The original 03s and 01s data don't have data in oceans and GMT fills the ocean with the 15s data. Thus we have to resample the data and merge them.

[...]

PR #492, opened 2020-06-24
Wrap grdcut

seisman:

Wrap the `grdcut` function, which is the first step is to redesign 
the `load_earth_relief()` function (see #489).

Ideally it should take a grid file or a `xarray.DataArray` as input, and
store output as a grid file or return a `xarray.DataArray`.

Supported features:

- [X] input: grid file; output: grid file
- [X] input: grid file; output: `xarray.DataArray`
- [ ] input: `xarray.DataArray`; output: grid file
- [ ] input: `xarray.DataArray`; output: `xarray.DataArray`

Passing `xarray.DataArray` to `grdcut` is not implenmented in the
GMT 6.0.0. See https://github.com/GenericMappingTools/gmt/pull/3532 for
a feature request to the upstream GMT.

The long-form names of the grdcut function come from [GMT.jl](https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.grdcut), but they're open for better naming.

The implementation of the `grdcut()` function is very similar to the [`surface()` function](https://github.com/GenericMappingTools/pygmt/blob/516e799bb9753ed45b9e122b8ebb5fd446320866/pygmt/gridding.py#L22).

**Reminders**

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -72,6 +72,7 @@ Operations on grids:
 
     grdinfo
     grdtrack
+    grdcut
 
 GMT Defaults
 ------------
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -70,9 +70,9 @@ Operations on grids:
 .. autosummary::
     :toctree: generated
 
+    grdcut
     grdinfo
     grdtrack
-    grdcut
 
 GMT Defaults
 ------------
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -18,7 +18,7 @@ from .filtering import blockmedian
 from .gridding import surface
 from .sampling import grdtrack
 from .mathops import makecpt
-from .modules import config, info, grdinfo, which
+from .modules import config, info, grdinfo, grdcut, which
 from . import datasets
 
 
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -18,7 +18,8 @@ from .filtering import blockmedian
 from .gridding import surface
 from .sampling import grdtrack
 from .mathops import makecpt
-from .modules import config, info, grdinfo, grdcut, which
+from .modules import config, info, grdinfo, which
+from .gridops import grdcut
 from . import datasets
 
 
--- a/pygmt/gridops.py
+++ b/pygmt/gridops.py
@@ -1,0 +1,119 @@
+"""
+GMT modules for grid operations
+"""
+
+import xarray as xr
+
+
+from .clib import Session
+from .helpers import (
+    build_arg_string,
+    fmt_docstring,
+    kwargs_to_strings,
+    GMTTempFile,
+    use_alias,
+    data_kind,
+    dummy_context,
+)
+from .exceptions import GMTInvalidInput
+
+
+@fmt_docstring
+@use_alias(
+    G="outgrid",
+    R="region",
+    J="projection",
+    N="extend",
+    S="circ_subregion",
+    Z="z_subregion",
+)
+@kwargs_to_strings(R="sequence")
+def grdcut(grid, **kwargs):
+    """
+    Extract subregion from a grid.
+
+    Produce a new *outgrid* file which is a subregion of *grid*. The
+    subregion is specified with *region*; the specified range must not exceed
+    the range of *grid* (but see *extend*). If in doubt, run
+    :meth:`pygmt.grdinfo` to check range. Alternatively, define the subregion
+    indirectly via a range check on the node values or via distances from a
+    given point. Finally, you can give *projection* for oblique projections to
+    determine the corresponding rectangular *region* setting that will give a
+    grid that fully covers the oblique domain.
+
+    Full option list at :gmt-docs:`grdcut.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    grid : str
+        The name of the input grid file.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    {J}
+    {R}
+    extend : bool or int or float
+        Allow grid to be extended if new *region* exceeds existing boundaries.
+        Give a value to initialize nodes outside current region.
+    circ_subregion : str
+        ``'lon/lat/radius[unit][+n]'``.
+        Specify an origin (*lon* and *lat*) and *radius*; append a distance
+        *unit* and we determine the corresponding rectangular region so that
+        all grid nodes on or inside the circle are contained in the subset.
+        If **+n** is appended we set all nodes outside the circle to NaN.
+    z_subregion : str
+        ``'[min/max][+n|N|r]'``.
+        Determine a new rectangular region so that all nodes outside this
+        region are also outside the given z-range [-inf/+inf]. To indicate no
+        limit on *min* or *max* only, specify a hyphen (-). Normally, any NaNs
+        encountered are simply skipped and not considered in the
+        range-decision. Append **+n** to consider a NaN to be outside the given
+        z-range. This means the new subset will be NaN-free. Alternatively,
+        append **+r** to consider NaNs to be within the data range. In this
+        case we stop shrinking the boundaries once a NaN is found [Default
+        simply skips NaNs when making the range decision]. Finally, if your
+        core subset grid is surrounded by rows and/or columns that are all
+        NaNs, append **+N** to strip off such columns before (optionally)
+        considering the range of the core subset for further reduction of the
+        area.
+
+    Returns
+    -------
+    ret: xarray.DataArray or None
+        Return type depends on whether the *outgrid* parameter is set:
+
+        - xarray.DataArray if *outgrid* is not set
+        - None if *outgrid* is set (grid output will be stored in *outgrid*)
+    """
+    kind = data_kind(grid)
+
+    with GMTTempFile(suffix=".nc") as tmpfile:
+        with Session() as lib:
+            if kind == "file":
+                file_context = dummy_context(grid)
+            elif kind == "grid":
+                raise NotImplementedError(
+                    "xarray.DataArray is not supported as the input grid yet!"
+                )
+                # file_context = lib.virtualfile_from_grid(grid)
+                # See https://github.com/GenericMappingTools/gmt/pull/3532
+                # for a feature request.
+            else:
+                raise GMTInvalidInput("Unrecognized data type: {}".format(type(grid)))
+
+            with file_context as infile:
+                if "G" not in kwargs.keys():  # if outgrid is unset, output to tempfile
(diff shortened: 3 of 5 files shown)


weiji14:
This is good work @seisman! It was on my TODO list a while back, but I realize that cropping a Cartesian bounding box could be done easily enough in `xarray` using [`xr.sel`](http://xarray.pydata.org/en/latest/generated/xarray.Dataset.sel.html#xarray.Dataset.sel). I do know that [`grdcut`](https://docs.generic-mapping-tools.org/latest/grdcut.html) handles more than just rectangular Cartesian cuts though (e.g. circular cuts [`-S`](https://docs.generic-mapping-tools.org/latest/grdcut.html#s), and Geographic ones), and it would be good to support those too.

pygmt/modules.py
```diff
@@ -147,6 +151,106 @@ def which(fname, **kwargs):
...
+        - None if *outgrid* is set (grid output will be stored in *outgrid*)
+    """
+    kind = data_kind(ingrid)
+
+    with GMTTempFile(suffix=".nc") as tmpfile:
+        with Session() as lib:
+            if kind == "file":
+                file_context = dummy_context(ingrid)
+            elif kind == "grid":
+                raise GMTInvalidInput(
```
weiji14, line 230: Would it be better to raise a [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError) here instead?
```suggestion
                raise NotImplementedError(
```

pygmt/modules.py
```diff
@@ -147,6 +151,106 @@ def which(fname, **kwargs):
...
+    determine the corresponding rectangular *region* setting that will give a
+    grid that fully covers the oblique domain.
+
+    Full option list at :gmt-docs:`grdcut.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    ingrid : str
```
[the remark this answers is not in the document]
seisman: Done in b578f858eea2b4915aba4bbe387db3265b4b4751

seisman:
I will open a feature request or a bug report to remind us that grdcut doesn't support xarray.DataArray due to a missing feature from upstream GMT.

seisman:
Do you have any idea why the test `test_grdcut_dataarray_in_fail` fails?

seisman:
There is one more failure on Windows:
```
    def test_grdcut_file_in_dataarray_out():
        "grdcut an input grid file, and output as DataArray"
>       outgrid = grdcut("@earth_relief_01d", region="0/180/0/90")

C:\Miniconda\envs\testing\Lib\site-packages\pygmt\tests\test_grdcut.py:27: 
C:\Miniconda\envs\testing\Lib\site-packages\pygmt\helpers\decorators.py:194: in new_module
C:\Miniconda\envs\testing\Lib\site-packages\pygmt\helpers\decorators.py:304: in new_module
C:\Miniconda\envs\testing\Lib\site-packages\pygmt\modules.py:251: in grdcut
    return result

self = <pygmt.helpers.tempfile.GMTTempFile object at 0x000001FD000AA250>
args = (None, None, None)

    def __exit__(self, *args):
        if os.path.exists(self.name):
>           os.remove(self.name)
E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\pygmt-llrbi2c3.nc'
```

You use https://github.com/GenericMappingTools/pygmt/blob/516e799bb9753ed45b9e122b8ebb5fd446320866/pygmt/gridding.py#L90-L94
while I use
```
        if outgrid == tmpfile.name:  # if user did not set outgrid, return DataArray
            result = xr.open_dataarray(outgrid)
        else:
            result = None  # if user sets an outgrid, return None
```
Is it the reason?

weiji14:
> You use
> 
> https://github.com/GenericMappingTools/pygmt/blob/516e799bb9753ed45b9e122b8ebb5fd446320866/pygmt/gridding.py#L90-L94
> 
> while I use
> 
> ```
>         if outgrid == tmpfile.name:  # if user did not set outgrid, return DataArray
>             result = xr.open_dataarray(outgrid)
>         else:
>             result = None  # if user sets an outgrid, return None
> ```
> 
> Is it the reason?

Yes, see commit 3324235b83a08bde932061c9d3f12bfd29b8a0d6, the comment which reads:

> The xarray.Dataset may sometimes be lazily loaded, so we use `.load()` to force loading of the data. See e.g. https://github.com/pydata/xarray/blame/9f4474d657193f1c7c9aac25bb2edf94755a8593/doc/io.rst#L169-L170. This is important because we want to persist the data even after the tempfile is deleted.

seisman:
> Do you have any idea why the test `test_grdcut_dataarray_in_fail` fails?

Never mind. I just forgot to push my changes.

pygmt/modules.py
```diff
@@ -147,6 +151,107 @@ def which(fname, **kwargs):
...
+    extend : bool or int or float
+        Allow grid to be extended if new *region* exceeds existing boundaries.
+        Give a value to initialize nodes outside current region.
+    circ_subregion : str
+        ``'lon/lat/radius[unit][+n]'``.
+        Specify an origin (*lon* and *lat*) and *radius*; append a distance
+        *unit* and we determine the corresponding rectangular region so that
+        all grid nodes on or inside the circle are contained in the subset.
+        If **+n** is appended we set all nodes outside the circle to NaN.
+    z_subregion :
```
weiji14, line 199:
```suggestion
    z_subregion : str
```

pygmt/__init__.py
```diff
@@ -18,7 +18,7 @@
 from .gridding import surface
 from .sampling import grdtrack
 from .mathops import makecpt
-from .modules import config, info, grdinfo, which
+from .modules import config, info, grdinfo, grdcut, which
```
weiji14, line 21: Probably best to start a separate file instead of putting it in `modules.py`. GMT puts grdcut under ["Convert or extract subsets of data"](https://docs.generic-mapping-tools.org/latest/modules.html#convert-or-extract-subsets-of-data). Can't come up with a good name, but maybe `conversion.py`, or `extraction.py`, or `subset.py`?

doc/api/index.rst
```diff
@@ -72,6 +72,7 @@ Operations on grids:
 
     grdinfo
     grdtrack
+    grdcut
```
weiji14, line 75:
```suggestion
    grdcut
    grdinfo
    grdtrack
```
Make it ordered alphabetically

pygmt/__init__.py
seisman: As it's invisible to users, perhaps we can do the moving after we have more functions in `modules.py` and have a better idea to organize them?

doc/api/index.rst
weiji14: Oh, and just a note that you can 'batch' up suggested changes on the 'Files Changed' tab in Github. See https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes.  Saves the CI from doing too much work.

pygmt/__init__.py
weiji14: Yes, but it's harder to run tools like `git blame` on files that have moved unless people are very good with `git`. Best to put it in the right place from the start.

seisman: Fair enough. We already have a `clib/conversion.py` file, so `conversion.py` is not a good name.

weiji14: Sigh, I wished we had just put everything as a separate `*.py` file like in [GMT.jl](https://github.com/GenericMappingTools/GMT.jl/tree/master/src) instead of grouping them :facepalm:. Would `converter.py` do? Or `convertsubset.py`? I'm really bad with names :cry:

seisman: > I wished we had just put everything as a separate `*.py` file like in [GMT.jl](https://github.com/GenericMappingTools/GMT.jl/tree/master/src) instead of grouping them

Me too. It's also easier to find the functions.

seisman: It seems all modules were first designed to be in a `modules` directory (https://github.com/GenericMappingTools/design-python#package-organization), but later Leo changed his mind (https://github.com/GenericMappingTools/pygmt/commit/252d5df5f493ca2dafb3bed2a0b6909e37c40578)

seisman: What about [`gridops.py`](https://docs.generic-mapping-tools.org/latest/modules.html#grid-operations), similar to the `mathops.py` we already have. I think it makes sense to have grdcut, grdpaste and grdblend in that group. 

BTW, I don't think it's a good choice to put modules by groups. We probably need to open a new issue discussing if we want to put modules in separate *.py files.

weiji14: Yeah, let's do `gridops.py` for now. I don't think we can avoid renaming/shifting the modules in the future, but that's a topic for another issue.

pygmt/gridops.py
```diff
@@ -0,0 +1,119 @@
...
+
+            with file_context as infile:
+                if "G" not in kwargs.keys():  # if outgrid is unset, output to tempfile
+                    kwargs.update({"G": tmpfile.name})
+                outgrid = kwargs["G"]
+                arg_str = " ".join([infile, build_arg_string(kwargs)])
+                lib.call_module("grdcut", arg_str)
+
+        if outgrid == tmpfile.name:  # if user did not set outgrid, return DataArray
+            with xr.open_dataarray(outgrid) as dataarray:
```
weiji14, line 114: After starting #494, I'm starting to wonder if we should have a `pygmt.open` function to load NetCDF files **with** the `node_offset` attribute inside the xarray.DataArray, as needed if we're going to start chaining operations as you mentioned in https://github.com/GenericMappingTools/pygmt/pull/476#issuecomment-649872282.

seisman: If I understand the GMT C API correctly, GMT has its own data types (internally) for [grids](https://docs.generic-mapping-tools.org/dev/api.html#gmt-grids), [CPT](https://docs.generic-mapping-tools.org/dev/api.html#color-palette-tables-cpt) and [data tables](https://docs.generic-mapping-tools.org/dev/api.html#data-tables). It also provides API functions [`GMT_Read_Data`](https://docs.generic-mapping-tools.org/dev/api.html#gmt-read-data) and [`GMT_Write_Data`](https://docs.generic-mapping-tools.org/dev/api.html#gmt-write-data), and two special modules [`gmtread`](https://github.com/GenericMappingTools/gmt/blob/master/src/gmtread.c) and [`gmtwrite`](https://github.com/GenericMappingTools/gmt/blob/master/src/gmtwrite.c) which I presume can read and write any GMT data types. Perhaps in the future we need to wrap these GMT data types. In this case, `pygmt.read()` and `pygmt.write()` may make more sense.

PR #494, opened 2020-06-26
More informative metadata when loading earth_relief grids

weiji14:
**Description of proposed changes**

Return a richer set of attributes when running `pygmt.datasets.load_earth_relief()`, to include all the NetCDF header info metadata such as a proper title, description (including the DOI) and GMT command history used to produce the grid. In particular, the `node_offset` attribute will be useful for #476. This PR will also be useful for #489.

Before (using `xr.open_dataarray(fname)`):

```python
<xarray.DataArray 'elevation' (lat: 180, lon: 360)>
array([[ 2999. ,  2997.5,  2995.5, ...,  3003.5,  3002.5,  3000.5],
       [ 3126.5,  3127. ,  3127. , ...,  3125. ,  3125.5,  3126. ],
       [ 3051. ,  3050.5,  3051.5, ...,  3058. ,  3056. ,  3053.5],
       ...,
       [-3888.5, -3884.5, -3885.5, ..., -3897.5, -3894. , -3890.5],
       [-3494.5, -3566. , -3612.5, ..., -3213.5, -3294. , -3402.5],
       [-3898.5, -3859. , -3812. , ..., -3982.5, -3966. , -3939.5]],
      dtype=float32)
Coordinates:
  * lon      (lon) float64 -179.5 -178.5 -177.5 -176.5 ... 177.5 178.5 179.5
  * lat      (lat) float64 -89.5 -88.5 -87.5 -86.5 -85.5 ... 86.5 87.5 88.5 89.5
Attributes:
    long_name:         elevation relative to the geoid
    units:             meters
    vertical_datum:    EMG96
    horizontal_datum:  WGS84
```

After (using `xr.open_dataset(fname).to_array(name="elevation").squeeze(drop=True)`):

```python
<xarray.DataArray 'elevation' (lat: 180, lon: 360)>
array([[ 2999. ,  2997.5,  2995.5, ...,  3003.5,  3002.5,  3000.5],
       [ 3126.5,  3127. ,  3127. , ...,  3125. ,  3125.5,  3126. ],
       [ 3051. ,  3050.5,  3051.5, ...,  3058. ,  3056. ,  3053.5],
       ...,
       [-3888.5, -3884.5, -3885.5, ..., -3897.5, -3894. , -3890.5],
       [-3494.5, -3566. , -3612.5, ..., -3213.5, -3294. , -3402.5],
       [-3898.5, -3859. , -3812. , ..., -3982.5, -3966. , -3939.5]],
      dtype=float32)
Coordinates:
[...]

Issue #487, opened 2020-06-22
Deciding on default grid registration and grid type for PyGMT


weiji14:
**Description of the problem**

PyGMT currently assumes `xarray.DataArray` grids as gridline registered (`GMT_GRID_NODE_REG`), and of cartesian type (`GMT_GRID_IS_CARTESIAN`). This hardcoding has very likely resulted in the issue reported by @MarkWieczorek at #375. Note that this is a non-issue for users passing in NetCDF filenames directly (phew).

Currrently we are working on allowing for either **pixel/gridline** registered and **cartesian/geographic** grids at #476, by changing the `virtualfile_from_grid` mechanism. But we'll need to decide on a good **default** going forward.

Current (hardcoded) default:
- Gridline registered `GMT_GRID_NODE_REG`
- Cartesian type `GMT_GRID_IS_CARTESIAN`

Proposed default:
- Pixel registered `GMT_GRID_PIXEL_REG`
- Geographic type `GMT_GRID_IS_GEO`?

For the former, my rationale for changing the default to pixel registration is because `xarray` does so (see http://xarray.pydata.org/en/stable/plotting.html#coordinates), though there are libraries like [`salem`](https://salem.readthedocs.io) that allow changing an xarray grid coordinates from corner (gridline) to centre (pixel) based (see [`salem.Grid.corner_grid`](https://salem.readthedocs.io/en/v0.2.3/generated/salem.Grid.corner_grid.html#salem-grid-corner-grid)).

For the latter (cartesian/geographic), users will probably realize quite quickly that their grid isn't Geographic and set the appropriate Cartesian flag. It will probably be less surprising for  those users plotting across the dateline too (e.g. #390).

In any case, I say we **keep the current default for now** (say for a v0.1.2 release), and **introduce any breaking changes in v0.2.0** (also after GMT 6.1.0 comes out). 

**System information**

Please paste the output of `python -c "import pygmt; pygmt.show_versions()"`:

```
PyGMT information:
  version: v0.1.1+18.gd203cd5f.dirty
System information:
  python: 3.7.6 | packaged by conda-forge | (default, Jun  1 2020, 18:57:50)  [GCC 7.5.0]
  executable: /home/user/miniconda3/envs/pygmt/bin/python
  machine: Linux-5.4.0-37-generic-x86_64-with-debian-bullseye-sid
Dependency information:
  numpy: 1.18.5
  pandas: 1.0.5
  xarray: 0.15.1
[...]

PR #500, opened 2020-06-29
Initialize a GMTDataArrayAccessor

### LINKED ISSUE #499: Make a gmt xarray accessor to store metadata from grdinfo

Metadata (data about data) is important, and there are specific kinds of metadata used by GMT when it decides how to plot a grid, such as:

- Registration type: Gridline/Pixel 
- Coordinate system: Cartesian/Geographic
- Region: xmin/xmax, ymin/ymax, zmin/zmax
- Default Colormap (For GMT 6.1.0+ see https://github.com/GenericMappingTools/gmtserver-admin/pull/62)

Instead of explicitly wrapping `gmt info` or `gmt grdinfo` as in #147, why not allow the user to just **access** it via `grid.gmt.someattribute`? This would make use of [xarray accessors](http://xarray.pydata.org/en/v0.15.1/internals.html#extending-xarray) and call `grdinfo` under the hood to retrieve those metadata properties.

Pros:

- We can design it so that the attributes (e.g. z-range) update **dynamically** when we cut/transform/resample/etc the grid, unlike in https://github.com/pydata/xarray/issues/2247 where the attributes become out of date or are dropped with each operation
- Will be useful for autodetecting pixel/gridline registration in #476
- Won't need #494 so much anymore.
- Could store the region/projection information inside, and have `fig.basemap()` et al. use it automatically
- Tab-completion!!
- Avoids clashing with namespace of other libraries for things like `crs` and whatnot.

Cons:
- We would be designing yet another standard, as per https://xkcd.com/927/
- Would involve some development effort
- Might not be as discoverable as simply using `grid.attrs["someattribute"]` or `pygmt.grdinfo(grid)`, but that's just a documentation thing.

In reality, this `gmt` accessor can do more than just hold metadata. We could extend it to do `grid.gmt.plot` and more. But let's start with having it hold the metadata we want, and then build on top of it from there.

References:
- https://github.com/GenericMappingTools/pygmt/issues/147#issuecomment-450018956
- https://github.com/GenericMappingTools/pygmt/pull/476#discussion_r446874666
- https://github.com/pydata/xarray/issues/2247

Other xarray accessor examples:
- rioxarray - https://corteva.github.io/rioxarray/stable/examples/crs_management.html
- [cf-xarray](https://github.com/xarray-contrib/cf-xarray) (Get CF attributes)
- xgeo - https://geosynopsis.github.io/xgeo/docs/html/examples.html
- salem - https://salem.readthedocs.io/en/v0.2.3/xarray_acc.html

weiji14:

An [xarray accessor](http://xarray.pydata.org/en/v0.15.1/internals.html#extending-xarray) for GMT specific information! Currently holds the gridline/pixel registration and cartesian/geographic type properties. Created a new 'Metadata' section in the API docs for this, and moved info and grdinfo here.

Enables functionality like this:

```python
print(grid.gmt.registration)
1  # for pixel registration
```

```python
print(grid.gmt.gtype)
0  # for cartesian coordinates
```

Will be extremely helpful for #476.

First step in fixing #499; Supersedes/Closes #494.

**Reminders**


pygmt/modules.py
```diff
@@ -214,3 +217,85 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        self._obj = xarray_obj
+        try:
+            self._source = self._obj.encoding["source"]
+            self._info = grdinfo(self._source)
+        except KeyError:
+            default_reg_and_gtype = "Gridline node registration used [Cartesian grid]"
+            # logging.warning(
+            #    msg="Cannot find a NetCDF source of xarray grid. "
+            #    f"Will fallback to using GMT's default setting: {default_reg_and_gtype}"
+            # )
```
weiji14, line 251: Will uncomment this after we decide on #487.

pygmt/modules.py
```diff
@@ -214,3 +217,85 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        """
+        Registration type of the grid, either Gridline (0) or Pixel (1).
+        """
+        try:
+            return self._registration
+        except AttributeError:
+            if "Gridline node registration used" in self._info:
+                self._registration = 0
+            elif "Pixel node registration used" in self._info:
+                self._registration = 1
```
weiji14, line 265: Can definitely be refactored once we hard pin to `GMT > 6.1.0`, though I do like the readability here.

pygmt/modules.py
```diff
@@ -214,3 +217,85 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    >>> # See if grid uses Cartesian (0) or Geographic (1) coordinate system
+    >>> grid.gmt.gtype
+    1
+    """
+
+    def __init__(self, xarray_obj):
+        self._obj = xarray_obj
+        try:
+            self._source = self._obj.encoding["source"]
+            self._info = grdinfo(self._source)
```
weiji14, line 245: I know we're not trying to maintain GMT 6.0 compatibility anymore, and I've had a look at how to refactor the code using `grdinfo -C`. However, the code becomes very unreadable, because we're trying to index the string output from `grdinfo`. 

```suggestion
            self._info = grdinfo(self._source)
            self._registration = int(grdinfo(self._source, C="n", o=10))
            self._gtype = int(grdinfo(self._source, C="n", o=11))
```

The best I can come up with is to add the two lines here, so that `xarray.DataArray` grids **with** a NetCDF source can return the `registration` and `gtype` info quickly, but grids **without** a source take a small one-time penalty with the string parsing, e.g. `if "Gridline node registration used" in self._info: self._registration = 0`.

To be honest, it might not be any faster even, since `grdinfo` will be run 3 times. Or maybe there's another way to refactor this for GMT 6.1 that I'm missing?

seisman: I'm not sure if I understand your codes correctly. Can we just run:
```python
try:
       self._source = self._obj.encoding["source"]
       self._registration = int(grdinfo(self._source, C="n", o=10))
       self._gtype = int(grdinfo(self._source, C="n", o=11))
except KeyError:
       self._registration = 0
       self._gtype = 0
```

seisman: The two grdinfo calls can also be reduce into one single call, like:
```python
self._registration, self._gtype = map(int, grdinfo(self._source, C="n", o=10,11).split())
```

I didn't try to run the code above. It may not work, but the idea is correct.

weiji14: Yes this works, it passes all the complicated tests I wrote at least. Though it might not be very readable.

seisman: Adding some comments can make it more readable. I think we may refactor the output of `grdinfo()` later.

BTW, `o=10,11` should be `o="10,11"`

weiji14: Cool, will make a commit before I head out later. Agree that we can refactor later to add in more useful info, I think Liam wanted to improve it at https://github.com/GenericMappingTools/pygmt/issues/147#issuecomment-450018956.

weiji14: Alright done. I'll be away for the rest of the day. Feel free to merge this one in (or make changes as you see fit) **once** #507 is merged into master. I'm not sure how we're going to handle the Scipy Sprint with so many tests breaking on GMT 6.1 but I suppose we'll have to make do.

pygmt/modules.py
weiji14: Ok, refactored in 896ed590ae5ce5198c1af810e729f4159436f1a8.

pygmt/modules.py
```diff
@@ -214,3 +217,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    1
+    >>> # See if grid uses Cartesian (0) or Geographic (1) coordinate system
+    >>> grid.gmt.gtype
+    1
+    """
+
+    def __init__(self, xarray_obj):
+        self._obj = xarray_obj
+        try:
+            self._source = self._obj.encoding["source"]
```
seisman, line 244: Please add a comment here, explaining what `_obj.encoding["source"]` contains.

pygmt/modules.py
```diff
@@ -214,3 +217,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            self._registration, self._gtype = map(
```
seisman, line 245: A comment here explain the meaning of `10,11`

pygmt/modules.py
```diff
@@ -214,3 +217,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+                int, grdinfo(self._source, C="n", o="10,11").split()
+            )
+        except KeyError:
+            self._registration = 0  # Default to Gridline registration
+            self._gtype = 0  # Default to Cartesian grid type
+            # logging.warning(
```
seisman, line 251: Remove the warning?

pygmt/modules.py
```diff
@@ -214,3 +217,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    @property
+    def registration(self):
+        """
+        Registration type of the grid, either Gridline (0) or Pixel (1).
+        """
+        return self._registration
+
+    @registration.setter
+    def registration(self, value):
+        if value in (0, 1):
```
seisman, line 266: It may be more user friendly if we can also set the properties by `grid.gmt.registration = "pixel"` and `grid.gmt.gtype = "geographic"`, but I'm OK with the current boolean values, 0 and 1.

pygmt/modules.py
weiji14: Ok, I've removed it for now, but will revisit it when we decide on #487. The intention is to not surprise users by enforcing a particular default registration/gtype (especially users used to using pixel registration/Cartesian coordinate data). The warning would ideally be surpressed by setting e.g. `grid.gmt.registration = 1`, but I haven't quite worked out how to do so.

pygmt/modules.py
weiji14: Good idea. Let's leave it for now, but introduce this once we work out how to do the `__repr__` as mentioned in https://github.com/GenericMappingTools/pygmt/pull/500#pullrequestreview-446833212.

seisman:
Please change to a better title before merging.

weiji14:
Oops sorry, missed your comment.

seisman:
Never mind. You can still change the draft release information directly. The title is not descriptive to me.

weiji14:
Yep, will do that in the release notes. I think the draft one will be overwritten every time a PR is merged.

PR #506, opened 2020-07-07
Add shading to grdview gallery example

liamtoney:

Add shading to the new `grdview()` example. This requires the bugfix of GMT 6.1.0 and will **not** work in 6.0.0, so we should probably wait to merge this until v0.2.0 is released — this is more of a bookmark. But it looks awesome:

ackley

This is an improvement on #502. @weiji14 please let me know what inline comments / other improvements we should have as well.

**Reminders**


examples/gallery/grid/grdview_surface.py
```diff
@@ -45,6 +46,7 @@ def ackley(x, y):
     surftype="s",
     cmap="roma",
     perspective="135/30",
+    I="+a45",
```
seisman, line 49: @liamtoney This is a nice improvement to the gallery. However, we prefer to to use long-form arguments (e.g., `shading="+a45"`, instead of short-form arguments (e.g., `I="+a45"`). See https://github.com/GenericMappingTools/pygmt/issues/473 for the reason, and https://github.com/GenericMappingTools/pygmt/pull/474 for our recent efforts.

However, the `grdview()` function doesn't provide a long-form alias for the **-I** option. So we need to do that first. The change is simple, just add `I="shading"` to the `use_alias` decorator of the `grdview()` function. You can take the `grdimage()` function as an example.

liamtoney: Right. Would that need to be a separate PR, or can that happen in this one?

seisman: It's a tiny change. I think it's OK to do it in this PR.

examples/gallery/grid/grdview_surface.py
```diff
@@ -45,6 +46,7 @@ def ackley(x, y):
     surftype="s",
     cmap="roma",
     perspective="135/30",
```
weiji14, line 48:
```suggestion
    perspective=[135, 30],  # Azimuth SouthEast (135°), at Elevation 30°
```
Looks awesome! Just one minor suggestion here to use a list when setting the perspective, and an inline comment.

liamtoney:
@weiji14 so once these statuses pass, I may merge?

weiji14:
Yes! Just click on the 'Squash and merge' button. You might want to format the commit message a bit, just summarize what was done and leave out things like 'fix typo'.

PR #509, opened 2020-07-09
Allow load_earth_relief() to load pixel or gridline registered data

weiji14:

Somewhat backward compatible way of retrieving the `earth_relief` grids using GMT 6.0 or GMT 6.1. Tests on GMT 6.0 should still pass as usual in this PR, but there should be fewer failures on GMT 6.1 (except those due to the cpt change).

TODO:
- [x] Modify earth_relief.py to allow for getting pixel/gridline earth_relief grids (80170245ac2e781468dd52166a993ce9866faefe) **53** test failures on GMT 6.1
- [x] Change unit tests to use `load_earth_relief(pixel_reg=False)` (9a7491059906c7c37f4c352e7abfb16908119f1e) **39** test failures on GMT 6.1

Fixes (partially) #489

**Reminders**

### DIFF
--- a/pygmt/datasets/earth_relief.py
+++ b/pygmt/datasets/earth_relief.py
@@ -3,12 +3,13 @@ Functions to download the Earth relief datasets from the GMT data server.
 The grids are available in various resolutions.
 """
 import xarray as xr
+from packaging.version import Version
 
-from .. import which
+from .. import clib, which
 from ..exceptions import GMTInvalidInput
 
 
-def load_earth_relief(resolution="01d"):
+def load_earth_relief(resolution="01d", pixel_reg=None):
     """
     Load Earth relief grids (topography and bathymetry) in various resolutions.
 
@@ -28,6 +29,12 @@ def load_earth_relief(resolution="01d"):
         ``'20m'``, ``'15m'``, ``'10m'``, ``'06m'``, ``'05m'``, ``'04m'``,
         ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'`` or ``'15s'``.
 
+    pixel_reg : bool
+        Grid registration type. Either ``True`` for pixel registration or
+        ``False`` for gridline registration. Default is ``None``, which returns
+        a pixel registered grid for GMT 6.1 or newer, and a gridline registered
+        grid for GMT 6.0.
+
     Returns
     -------
     grid : xarray.DataArray
@@ -36,7 +43,18 @@ def load_earth_relief(resolution="01d"):
 
     """
     _is_valid_resolution(resolution)
-    fname = which("@earth_relief_{}".format(resolution), download="u")
+
+    if not pixel_reg:
+        registration = ""  # If None, let GMT decide on Pixel/Gridline type
+        # TODO simplify this if-then once PyGMT no longer depends on GMT 6.0
+        if pixel_reg is False:
+            with clib.Session() as lib:
+                if Version(lib.info["version"]) >= Version("6.1.0"):
+                    registration = "_g"
+    else:
+        registration = "_p"
+
+    fname = which(f"@earth_relief_{resolution}{registration}", download="u")
     grid = xr.open_dataarray(fname)
     # Add some metadata to the grid
     grid.name = "elevation"
--- a/pygmt/datasets/earth_relief.py
+++ b/pygmt/datasets/earth_relief.py
@@ -31,8 +31,9 @@ def load_earth_relief(resolution="01d", registration=None):
 
     registration : str
         Grid registration type. Either ``pixel`` for pixel registration or
-        ``gridline`` for gridline registration. Default is ``None``, which
-        returns a pixel registered grid.
+        ``gridline`` for gridline registration. Default is ``None``, where
+        a pixel-registered grid is returned unless only the
+        gridline-registered grid is available.
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


pygmt/datasets/earth_relief.py
```diff
@@ -3,12 +3,13 @@
...
 import xarray as xr
+from packaging.version import Version
 
-from .. import which
+from .. import clib, which
 from ..exceptions import GMTInvalidInput
 
 
-def load_earth_relief(resolution="01d"):
+def load_earth_relief(resolution="01d", pixel_reg=None):
```
seisman, line 12: I personally prefer an argument like `registration="pixel"` or `registration="gridline"`, just like **-r** option in most GMT modules.

EJFielding: I agree that `registration="pixel"` is much better.

weiji14: Yep, this can be arranged.

seisman:
@weiji14 Do you want to implement the registration option only, or do you also want to support `region` (by using `grdcut` in this PR?

weiji14:
Let's keep this PR small. I actually have no idea how the `grdcut` and `which` commands go together, might be better for you to handle it.

seisman:
OK. Please merge the master branch in then I can review it.

pygmt/datasets/earth_relief.py
```diff
@@ -28,6 +29,11 @@ def load_earth_relief(resolution="01d"):
         ``'20m'``, ``'15m'``, ``'10m'``, ``'06m'``, ``'05m'``, ``'04m'``,
         ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'`` or ``'15s'``.
 
+    registration : str
+        Grid registration type. Either ``pixel`` for pixel registration or
+        ``gridline`` for gridline registration. Default is ``None``, which
```
seisman, line 34: See https://docs.generic-mapping-tools.org/latest/datasets/earth-relief.html.

> If *reg* is not specified we will return the pixel-registered version unless only the gridline-registered file is available.

pygmt/datasets/earth_relief.py
```diff
@@ -36,7 +42,21 @@ def load_earth_relief(resolution="01d"):
...
+        _registration = ""  # If None, let GMT decide on Pixel/Gridline type
+        with clib.Session() as lib:
+            if registration is not None and Version(lib.info["version"]) >= Version(
+                "6.1.0"
+            ):
+                _registration = f"_{registration[0]}"
+    else:
+        raise GMTInvalidInput(
+            f"Invalid grid registration: {registration}, should be either "
+            "'pixel', 'gridline' or None. Default is None for pixel registration."
```
seisman, line 56: Same here.

pygmt/datasets/earth_relief.py
```diff
@@ -36,7 +42,21 @@ def load_earth_relief(resolution="01d"):
 
     """
     _is_valid_resolution(resolution)
-    fname = which("@earth_relief_{}".format(resolution), download="u")
+
+    if registration in ("pixel", "gridline", None):
+        _registration = ""  # If None, let GMT decide on Pixel/Gridline type
```
seisman, line 47: The variable name `_registration` is a little confusing here. What about *reg*?

pygmt/datasets/earth_relief.py
```diff
@@ -36,7 +42,21 @@ def load_earth_relief(resolution="01d"):
...
+        with clib.Session() as lib:
+            if registration is not None and Version(lib.info["version"]) >= Version(
+                "6.1.0"
+            ):
```
seisman, line 51: I think this is also good:
```suggestion
            if registration and Version(lib.info["version"]) >= Version("6.1.0"):
```

weiji14:
Oops, should have used `mkdir -p ~/.gmt` instead of `mkdir ~/.gmt` at https://github.com/GenericMappingTools/pygmt/blob/a5af2dc5ecca2a02dc8bacb24f8d55d56c330e4f/.github/workflows/ci_tests.yaml#L72

Should I just fix it here?

pygmt/datasets/earth_relief.py
```diff
@@ -36,7 +42,21 @@ def load_earth_relief(resolution="01d"):
...
+                _registration = f"_{registration[0]}"
+    else:
+        raise GMTInvalidInput(
+            f"Invalid grid registration: {registration}, should be either "
+            "'pixel', 'gridline' or None. Default is None for pixel registration."
+        )
+
+    fname = which(f"@earth_relief_{resolution}{_registration}", download="u")
```
seisman, line 59: It's better to use `download="a"` in GMT 6.1, but it's OK now.

pygmt/tests/test_datasets.py
```diff
@@ -80,9 +80,15 @@ def test_earth_relief_01d():
```
seisman, line 80: After bumping to GMT 6.1, we need to add a test with `registration="pixel"` or None.

seisman:
> Oops, should have used `mkdir -p ~/.gmt` instead of `mkdir ~/.gmt` at
> 
> https://github.com/GenericMappingTools/pygmt/blob/a5af2dc5ecca2a02dc8bacb24f8d55d56c330e4f/.github/workflows/ci_tests.yaml#L72
> 
> Should I just fix it here?

It's a little weird. For the ubuntu job, I saw
```
Cache restored from key: cache-gmt-refs/heads/master-macOS-20200609
```
why macOS?

weiji14:
> It's a little weird. For the ubuntu job, I saw
> 
> ```
> Cache restored from key: cache-gmt-refs/heads/master-macOS-20200609
> ```
> 
> why macOS?

Ah, we'll need to append the matrix-os to the restore-from key then. Maybe I should open a separate PR :facepalm:

seisman:
> > It's a little weird. For the ubuntu job, I saw
> > ```
> > Cache restored from key: cache-gmt-refs/heads/master-macOS-20200609
> > ```
> > 
> > 
> > why macOS?
> 
> Ah, we'll need to append the matrix-os to the restore-from key then. Maybe I should open a separate PR 🤦

I see. Please open another PR instead.

seisman:
Also, `mkdir -p` is better.

seisman:
BTW, the PR title can be changed to "Allow load_earth_relief() to load pixel or gridline registered data". It's more readable to me.

pygmt/datasets/earth_relief.py
```diff
@@ -36,7 +42,21 @@ def load_earth_relief(resolution="01d"):
...
     _is_valid_resolution(resolution)
-    fname = which("@earth_relief_{}".format(resolution), download="u")
+
+    if registration in ("pixel", "gridline", None):
+        _registration = ""  # If None, let GMT decide on Pixel/Gridline type
+        with clib.Session() as lib:
+            if registration is not None and Version(lib.info["version"]) >= Version(
+                "6.1.0"
+            ):
+                _registration = f"_{registration[0]}"
```
seisman, line 52: If GMT<6.1.0, we should raise an warning like "The 'registration' argument is only available for GMT>=6.1.0".

pygmt/datasets/earth_relief.py
weiji14: Done.

pygmt/datasets/earth_relief.py
weiji14: `registration="gridline"` will work on GMT 6.0, so I'll just raise `GMTInvalidInput` when GMT 6.0 users try to pass in `registration="pixel"`.

PR #512, opened 2020-07-11
Let pytest generate both HTML and XML coverage reports

seisman:
Let pytest generate coverage reports in both XML and HTML formats.
HTML version is useful for viewing reports locally, and XML version is for codecov.

If all tests pass, the XML and HTML coverage reports will be copied into
the root directory; otherwise, they will be in the tmp-test-dir-with-unique-name
directory.


seisman:
@weiji14 As you're away, I'll merge this PR without reviewing. We can fix any issues later.

PR #513, opened 2020-07-11
Eliminate Azure Pipelines CI jobs, use GitHub Actions instead

seisman:
Add steps for "building documentation" and "uploading coverage to
codecov" in GitHub Actions.

We no longer need the Azure Pipelines jobs, except the one doing style checks.

Actually, I just found that the coverage were never successfully uploaded to codecov
in the Azure Pipelines jobs, due to the incorrect codecov token.

One more thing, codecov does't require a token for public repositories.


weiji14:
> We no longer need the Azure Pipelines jobs, except the one doing style checks.

We can actually run the style checks on Github Actions too, but let's leave that for another day, we need a break from all the maintenance :laughing:.

> Actually, I just found that the coverage were never successfully uploaded to codecov
> in the Azure Pipelines jobs, due to the incorrect codecov token.

This doesn't surprise me, I noticed that the coverage reports all come from the Linux CIs on Travis. Good though that there's a Github Actions for codecov (is there nothing Github Actions can't do?)!

> One more thing, codecov does't require a token for public repositories.

:+1:

Issue #483, opened 2020-06-21
Figure.text: Allow passing in a list/array of angle/font/justify values to text


weiji14:
**Description of the desired feature**

Following up from https://github.com/GenericMappingTools/pygmt/pull/481#pullrequestreview-434464455, it would be desirable if users can:

1. Pass in a list/array of angle/font/justify values to [`text`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.text) (currently only a single angle/font/justify value is allowed, and applied to all texts)
2. Handle more flexible input formats, e.g., -F+a+j+f and -F+a+f+j (see also https://github.com/GenericMappingTools/pygmt/issues/479#issuecomment-647067718).

On the non-user facing side, it would be nice too if we can pass the `x,y,text` input via a **virtualfile** as mentioned in https://github.com/GenericMappingTools/pygmt/pull/321#discussion_r334993927, instead of a temporary tab-delimited file. This will be a more efficient method of passing data around rather than relying on the `pandas.to_csv` method. Will need to modify `virtualfile_from_vectors` to accept `str` type arrays, see https://github.com/GenericMappingTools/pygmt/pull/481#discussion_r443133421.

**Are you willing to help implement and maintain this feature?** Happy to review a PR for it :smile:

seisman:
Here are some technical notes, which may be useful (or may be wrong) if someone wants to implement this feature.

1. [`put_vector()`](https://github.com/GenericMappingTools/pygmt/blob/86c46f075d6f68a1367bc0fad161f98ba99c14c3/pygmt/clib/session.py#L721) function (wrapper of GMT API `GMT_Put_Vector()`) can only pass numerical and datetime vectors/arrays to GMT API
2. All trailing strings should be passed to GMT API by the `GMT_Put_Strings()` function (no Python wrap for it yet)
3. You can call `GMT_Put_Vectors()` multiple times, and pass one vector each time
4. You can only call `GMT_Put_Strings()` one time to pass all trailing string arrays to GMT

Thus, to implement this feature, you need to:

1. wrap the `GMT_Put_Strings()` function, i.e., `put_strings()`
2. pass the **leading** numerical arrays to GMT via the `put_vector()` function one by one
3. Combine all the trailing arrays into one string array, and pass it to GMT via the `put_strings()` function
4. Keep your fingers crossed 🤞

For example, if you want to pass the following arrays to make `text()` accept input format *x, y, angle, justify, font, text*: 
```
[...]

PR #471, opened 2020-06-01
Wrap plot3d

weiji14:
Wrapping the `plot3d` function! Original GMT `plot3d` function can be found at https://docs.generic-mapping-tools.org/latest/plot3d.html. Current implementation is mostly copy-modify-pasted from `plot`, including the tests. Also added extra aliases to the `basemap` function to make it 3D capable.

Live documentation preview of `plot3d` API is at https://pygmt-git-mapping-plot3d.gmt.vercel.app/api/generated/pygmt.Figure.plot3d.html. Gallery example is at https://pygmt-git-mapping-plot3d.gmt.vercel.app/gallery/plot/scatter3d.html

**Description of proposed changes**

This is the gallery example of a 3D plot with the [Iris flower dataset](https://en.wikipedia.org/wiki/Iris_flower_data_set), inspired by the example at https://scikit-learn.org/stable/auto_examples/cluster/plot_cluster_iris.html :smile:

3D scatter plot of Iris flower dataset made using PyGMT plot3d

Here's another example with a bit of a geographic curve:

```python
fig = Figure()
fig.plot3d(
    x=data[:, 0],
    y=data[:, 1],
    z=data[:, 2],
    zscale=5,
    perspective=[225, 30],
    region=[10, 70, -5, 10, 0, 1],
    projection="R270/4i",
    style="s1c",
    color="green",
    frame=["ag", "zag"],
)
fig.savefig("example_plot3d.png")
fig.show()
```
produces:
example_plot3d

Alternatively, we could follow GMT.jl's example at https://www.generic-mapping-tools.org/GMT.jl/latest/plot/ and just have one `plot` function, which would handle 3D plotting if the `z` dimension is provided. Thoughts?

Fixes #467

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [x] If adding new functionality, add an example to docstrings or tutorials.

Issue #262, opened 2019-01-16
Disallow single character arguments


leouieda:
**Description of the desired feature**

A major usability problem for GMT are the complex and cryptic single character arguments and modifiers. This is why we implemented the entire alias system in GMT/Python (which is making its way back into GMT 6). Our docs all encourage using the long form names and this is not strange in the Python ecosystem. Since we're actively discouraging the use of `B` and the like, why are we maintaining the capability of doing so? Hardcore GMT users who value brevity are not our target audience. And if you're using Jupyter or any modern text editor, tab completion eliminates the hassle of longer names (which we currently can't use because of the alias system). 

I propose getting rid of the single letter arguments all together. 

Doing this will allow us to:

* Define functions arguments normally in the Python functions instead of relying on `**kwargs`. 
* Get rid of the `use_alias` decorator and replace it with a translator function that's called inside the function body. This functions would be a lot easier than the decorator to maintain (which is an ugly function inside a function inside a function).
* Use tab completion for the long argument names
* Simplify the testing a lot but not needing to test the aliases

The one downside I see that we will have to fully implement all arguments in order for them to be used (like `X` and `Y`, etc). But we were gonna have to do this eventually any way.

I'd love to get some feedback on this.

leouieda:
Pinging @GenericMappingTools/gmt-python @liamtoney @weiji14 @akshmakov @mjziebarth

seisman:
GMT is trying to implement long-form options, see https://github.com/GenericMappingTools/gmt/pull/230. 

If it's merged, we no longer need `use_alias` anymore.

leouieda:
Yep. We'll still need to parse and do some conversion on the arguments, though.

liamtoney:
What would be the Pythonic implementation of the numerous modifiers that follow option arguments, i.e. the Python version of
```
--longoption[=[<directive>:]<arg>][+<mod1[=<arg1>]][+<mod2[=<arg2>]]...
```
as given in [GenericMappingTools/gmt#230](https://github.com/GenericMappingTools/gmt/pull/230)?

The more complex cases could be tricky — and they do come up. For example, I recently wrote something like the following to produce a frame overlay with map scale:
```python
fig.basemap(frame='af', L='g{0}/{1}+w100k+c{0}/{1}+l'.format(*SCALE_LOC))
```
[...]

PR #516, opened 2020-07-11
Wrap meca

tjnewton:
Add [meca](https://docs.generic-mapping-tools.org/latest/supplements/seis/meca.html) functionality to plot focal mechanisms. Started during SciPy2020 sprint.

Adds meca function to base_plotting.py to add [focal mechanism](https://en.wikipedia.org/wiki/Focal_mechanism) plotting to pygmt. 

```python
# MWE:
import pygmt

fig = pygmt.Figure()
fig.meca(focal_mechanism=[239.384, 34.556, 12., 180, 18, -88, 0, 72, -90, 5.5, 0, 0, 0],
         region=[239, 240, 34, 35.2], convention='c2c', projection='m4c')
fig.show(method='external')
```

Live documentation preview at https://pygmt-git-fork-tjnewton-add-meca.gmt.now.sh/api/generated/pygmt.Figure.meca.html

**Reminders**


liamtoney:
@tjnewton and I will work on this more tomorrow, adding tests and whatnot.

rmodrak:
*My question for you guys, as end-users, is: In what formats would you want to provide MT/FMT info? As a filename? List of params? Dictionary? Something with Pandas?*

Hi Liam,  It sounds like you are looking to replace the classic psmeca command syntax, in which values are read from an ascii table, with something more Pythonic?   Then perhaps one (or both) of the following  conventions from obspy:

- a dictionary with keys 'm_rr', 'm_tt', 'm_pp', 'm_rt', 'm_rp'
- a list of moment tensor elements in the order [m_rr, m_tt, m_pp, m_rt, m_rp]

liamtoney:
Hi Ryan, thanks for reminding us that we should try to align with existing libraries! Tyler and I were discussing two options — the original file input as well as an array input with the same info for non-file usage — but I think your suggestion is much more Pythonic. So we can add a third option for a keyword (i.e., dict) based option.

---

@tjnewton I propose the following sketched-out method definition, similar to what we had but with a third option that involves providing a dictionary for `spec` (same as `focal_mechanism` above, shortening here for demo only) along with `lon`, `lat`, and `depth` required. This is more verbose source code but I think will make for a nicer user experience, such as:
```python
fig.meca(
    lon=239.384,
    lat=34.556,
    depth=12.0,
    spec=dict(
        strike1=180,
        dip1=18,
        rake1=-88,
        strike2=0,
        dip2=72,
        rake2=-90,
        mantissa=5.5,
        exponent=0,
    ),
    region=[239, 240, 34, 35.2],
    **other_kwargs
)
```
All names/details are drafts of course.
```python
def meca(
    self,
    lon=None,
    lat=None,
    depth=None,
    spec=None,
    convention=None,
    plot_lon=None,
    plot_lat=None,
    text=None,
    **kwargs
):

    if type(spec) is dict:  # The user opted for the nice keyword-based approach
        if (
            lat is None or lon is None or depth is None
        ):  # They need to give the location!
            raise Error('Location not fully specified.')

        # These are constants related to GMT and could be defined elsewhere (top of file?)
        AKI_PARAMS = [
            'strike',
            'dip',
            'rake',
            'magnitude'
        ]
        GCMT_PARAMS = [
            'strike1',
            'dip1',
            'rake1',
            'strike2',
            'dip2',
            'rake2',
            'mantissa',
            'exponent',
        ]

        # Aki and Richards
        if set(spec.keys()) == set(AKI_PARAMS):
            convention = 'a'
            foc_params = AKI_PARAMS

        # GCMT
        elif set(spec.keys()) == set(GCMT_PARAMS):
            convention = 'c'
            foc_params = GCMT_PARAMS

        # Other cases....
        elif ...:
            pass  # Same strategy as above

        else:
            raise Error('Focal mech params not understood')

        # Construct the vector (note that order matters here, hence the list comprehension!)
        spec = [lon, lat, depth] + [spec[key] for key in foc_params]

        # Add in plotting options, if given, otherwise add 0s as required by GMT
        for arg in plot_lon, plot_lat, text:
            if arg is None:
                spec.append(0)
            else:
                spec.append(arg)

    elif convention is None:
        raise Error('We need a convention to know how to interpret the input!')

    kind = data_kind(spec)
    if kind == 'file':
        pass  # Use file as-is (dummy file context)...
    elif kind == 'matrix':
        pass  # Convert to virtual file and then use as-is...
    else:
        raise Error('File type not understood')

    # Build arg str and call meca...
```

liamtoney:
I also wonder if we could replace the conventions that require `mantissa` and/or `exponent` with just the values themselves. It seems odd to specify them separately, perhaps a consideration for CLI GMT only that we needn't worry about in Python. I.e. for the GCMT convention we could replace those two with just `moment` to make things nicer:
```python
moment = 5000

mantissa = 10  # Hard-coded
exponent = np.log10(moment)

mantissa ** exponent  # == 4999.999999999999
```

pygmt/base_plotting.py
```diff
@@ -984,3 +984,161 @@ def text(
...
+        **kwargs,
+    ):
+        """
+        Plot focal mechanisms.
+
+        {aliases}
+
+        Parameters
+        ----------
+        spec: dict, 1D array, 2D array, or str
```
liamtoney, line 1011: We could be more specific about when this is a dictionary — e.g. "If `spec` is a dictionary, the following conventions are supported and need to be present as keys in the dictionary"

examples/gallery/plot/meca.py
```diff
@@ -13,13 +13,20 @@
...
+    projection="M6c",
+    land="grey",
+    water="lightblue",
+    shorelines=True,
+    resolution="f",
+    frame="a",
+)
 
 # store focal mechanisms parameters in a dict
 focal_mechanisms = dict(strike=330, dip=30, rake=90, magnitude=3)
```
liamtoney, line 27:
```suggestion
focal_mechanism = dict(strike=330, dip=30, rake=90, magnitude=3)
```
Singular here

examples/gallery/plot/meca.py
```diff
@@ -13,13 +13,20 @@
...
 
 # pass the focal mechanism data to meca in addition to the scale and event location
-fig.meca(focal_mechanisms, scale='1c', lon=-124.3, lat=48.1, depth=12.0)
+fig.meca(focal_mechanisms, scale="1c", lon=-124.3, lat=48.1, depth=12.0)
```
liamtoney, line 30:
```suggestion
fig.meca(focal_mechanism, scale="1c", lon=-124.3, lat=48.1, depth=12.0)
```
And here

liamtoney:
Whelp, our current implementation fails when we supply the `text` kwarg in the new test `test_meca_spec_dictionary()`, because we put everything into a numpy array and create a virtual file from that, but the array has dtype string if we add a string to it and our framework can't handle strings. Looks like we'll have to use the method implemented in `Figure.text()`, i.e. create a `GMTTempFile` ☹️ 

@weiji14 is that why we have to use the tempfile in `Fig.text()`? Seems like a parallel issue where data tables have mixed dtypes (lat, lon, and text)

weiji14:
> Whelp, our current implementation fails when we supply the `text` kwarg in the new test `test_meca_spec_dictionary()`, because we put everything into a numpy array and create a virtual file from that, but the array has dtype string if we add a string to it and our framework can't handle strings. Looks like we'll have to use the method implemented in `Figure.text()`, i.e. create a `GMTTempFile` frowning_face
> 
> @weiji14 is that why we have to use the tempfile in `Fig.text()`? Seems like a parallel issue where data tables have mixed dtypes (lat, lon, and text)

Well, the tempfile in `Fig.text()` is actually a pandas tempfile, which works but we're trying to move towards using a pure GMT method (see #483) by passing in the strings via the GMT C API. I'll see if I can wrap `GMT_Put_Strings()` now (might take a while), in the meantime, maybe disable the 'text' argument first?

pygmt/tests/test_meca.py
```diff
@@ -0,0 +1,156 @@
...
+    Check full moment tensor, especially with big isotropic components. The
+    data files and original script are provided by Carl Tape, available from
+    https://github.com/carltape/compearth, licensed under MIT license. Also see
+    https://github.com/GenericMappingTools/gmt/issues/661.
+    """
+
+    fig = Figure()
+
+    for file in (
+        "fullmt_ipts1_iref1",
```
weiji14, line 26: By the way, I don't know if this `meca` implementation works with `pandas.DataFrame`s yet, but just building on from DongDong's answer at https://forum.generic-mapping-tools.org/t/accessing-gmt-test-suite-files-using-syntax/655/3, you can actually 'download' the data files directly using [`pandas.read_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html)  like so:

```python
df = pd.read_csv("https://raw.githubusercontent.com/GenericMappingTools/gmt/master/test/seis/fullmt_ipts1_iref1", sep="\s+", header=None)
```

And maybe use [`pd.to_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html) to output to a file. Or you could wait for it to show up on the GMT cache.

seisman: We can add these files to the GMT cache, but I don't think we should.

GMT already has many tests checking if meca works well with various focal mechanisms. This PyGMT test is doing the same thing, and it's testing GMT, not PyGMT. 

IMHO, the PyGMT tests should focus on testing the Python codes, for examples, there are several different ways to pass focal mechanisms to `meca()` to test.

- pass a file
- passing an array 
- pass a single focal mechanism like `dict(strike=0, dip=90, rake=0, magnitude=5),`
- pass multiple focal mechanisms like `dict(strike=[0, 30], dip=[30, 60], rake=[30, 30], magnitude=[4, 5]`)

One more thing, some baseline images are too big because you're plotting too many beachballs (300KB + 500 KB).

liamtoney: These are all good points, @seisman. We will re-think the tests — we can likely do away with files. I think more tests like  [`test_meca_spec_dictionary()`](https://github.com/tjnewton/pygmt/blob/45b01de8c68dce8103faaa0dc3b98e2b71347ec0/pygmt/tests/test_meca.py#L80), but maybe with different mechanism types, are what we need.

> * pass multiple focal mechanisms like dict(strike=[0, 30], dip=[30, 60], rake=[30, 30], magnitude=[4, 5])

I haven't tested, but I don't think this works yet.

liamtoney:
> maybe disable the 'text' argument first?

Yeah, that might be what we have to do for now.

liamtoney:
@tjnewton as @seisman pointed out above, it would be good to add a  test that utilizes the three different input formats for `meca` — file, array, and dictionary. Another good test would be trying each of the five conventions. Wanna tackle these? I can provide help as needed.

We'll probably want to remove the bulky tests I added, which are testing GMT not PyGMT as @seisman pointed out.

tjnewton:
> Wanna tackle these?

@liamtoney I sure do. I'll push these tests sometime this week! Thanks for the feedback Liam, @seisman, @weiji14, and @rmodrak

tjnewton:
Next up, pandas.DataFrames then I'll round this out with some tests.

tjnewton:
The three test_mecas are passing on my system (macOS 10.14.6, Python 3.7.6) but not here 🤔

tjnewton:
I would still like to incorporate @liamtoney's idea of allowing the user to specify moment instead of both mantissa and exponent, but I will have to think about an elegant way to do this. For now I'm going to clean up the code to pass more style checks.

tjnewton:
I could use some feedback on the best approach to resolve the remaining linting warnings. Should I store the PARAMS variables elsewhere? 
` ``pygmt/base_plotting.py:1001:4: R0914: Too many local variables (37/15) (too-many-locals)
pygmt/base_plotting.py:1103:12: C0103: Variable name "AKI_PARAMS" doesn't conform to snake_case naming style (invalid-name)
pygmt/base_plotting.py:1104:12: C0103: Variable name "GCMT_PARAMS" doesn't conform to snake_case naming style (invalid-name)
pygmt/base_plotting.py:1114:12: C0103: Variable name "MT_PARAMS" doesn't conform to snake_case naming style (invalid-name)
pygmt/base_plotting.py:1115:12: C0103: Variable name "PARTIAL_PARAMS" doesn't conform to snake_case naming style (invalid-name)
pygmt/base_plotting.py:1116:12: C0103: Variable name "PRINCIPAL_AXIS_PARAMS" doesn't conform to snake_case naming style (invalid-name)
pygmt/base_plotting.py:1096:8: R1702: Too many nested blocks (6/5) (too-many-nested-blocks)
pygmt/base_plotting.py:1001:4: R0912: Too many branches (59/12) (too-many-branches)
pygmt/base_plotting.py:1001:4: R0915: Too many statements (128/50) (too-many-statements)
pygmt/base_plotting.py:1001:4: R0201: Method could be a function (no-self-use)```

Any thoughts @weiji14, @seisman, and @liamtoney?

pygmt/base_plotting.py
```diff
@@ -994,3 +994,358 @@ def text(
...
+                "t_azimuth",
+                "t_plunge",
+                "n_exponent",
+                "n_azimuth",
+                "n_plunge",
+                "p_exponent",
+                "p_azimuth",
+                "p_plunge",
+                "exponent",
+            ]
```
weiji14, line 1127: Could you put this into a dictionary? E.g. `{"AKI_PARAMS" : [...], "MT_PARAMS": [...], ...}`. This might help to simplify the if-then statements later.

pygmt/base_plotting.py
```diff
@@ -994,3 +994,358 @@ def text(
...
+            and `plot_lat` if provided. A small circle is plotted at the
+            initial location and a line connects the beachball to the circle.
+            Specify pen and optionally append ``+ssize`` to change the line
+            style and/or size of the circle.
+        {J}
+        {R}
+        {B}
+        """
+
+        # Check the spec and parse the data according to the specified
```
seisman, line 1094: I haven't read the codes, but please keep in mind that, another module [coupe](https://docs.generic-mapping-tools.org/latest/supplements/seis/coupe.html), which plots beachballs on cross-sections, will use exactly the same codes here, since **meca** and **coupe** share the same input file and **-S** option. 

We definitely don't want to maintain two copies of the same codes.

pygmt/base_plotting.py
tjnewton: I sure can. Great idea.

pygmt/base_plotting.py
tjnewton: Thanks for the tip. I haven't looked at coupe but I will.

weiji14: Good points. I'm not familiar with beachball plots, but there's a similar story going on with the `velo` function PR at #525 which shares a lot of the same implementation code as `plot` and even `plot3d` at #471. It might be the case that we'll end up having a 'core' function of some sort, and just change the module name (e.g. `meca` or `coupe`, `velo` or `plot` or `plot3d`).

For the sake of this PR though, I think it's fine to leave it like so for now. The most important thing is to have comprehensive unit tests, and we can refactor the code later (while keeping the tests the same) when `coupe` is wrapped.

tjnewton:
We now have a relatively comprehensive set of tests covering various data conventions and all input types. meca can accept dicts, pd.DataFrames, np.ndarrays, and files. I have taken care of all pylint style issues except those below. Are we aiming for a 10/10 pylint score (do I need to fix these)? @weiji14 @seisman @liamtoney 
```
pygmt/base_plotting.py:1001:4: R0914: Too many local variables (36/15) (too-many-locals)
pygmt/base_plotting.py:1116:8: R1702: Too many nested blocks (6/5) (too-many-nested-blocks)
pygmt/base_plotting.py:1001:4: R0912: Too many branches (53/12) (too-many-branches)
pygmt/base_plotting.py:1001:4: R0915: Too many statements (119/50) (too-many-statements)
pygmt/base_plotting.py:1001:4: R0201: Method could be a function (no-self-use)
```

weiji14:
Hi @tjnewton, just wanted to say that you've done a terrific job! I'll leave the proper review for @liamtoney, @seisman (or @sgjholt who was interested in this [on the forum](https://forum.generic-mapping-tools.org/t/pygmt-psmeca-option/692)) since they would know more about focal mechanism plots, but I'm quite happy with the `meca` tests here and think this would make a great addition to PyGMT.

In terms of the pylint score, 10/10 would be ideal of course, but the warnings are all refactor related (that's what the **R**1234 stands for), and I think we can open up a separate PR to handle that since this one is already getting very long. If you think you can fix any of them, do it now, otherwise just silence them temporarily (see https://stackoverflow.com/questions/35469253/silence-pylint-warning-about-unused-variables-for-string-interpolation).

pygmt/base_plotting.py
```diff
@@ -994,3 +994,368 @@ def text(
...
+            - ``"aki"`` — *strike, dip, rake, magnitude*
+            - ``"gcmt"`` — *strike1, dip1, rake1, strike2, dip2, rake2,
+              mantissa, exponent*
+            - ``"mt"`` — *mrr, mtt, mff, mrt, mrf, mtf, exponent*
+            - ``"partial"`` — *strike1, dip1, strike2, fault_type, magnitude*
+            - ``"principal_axis"`` — *t_exponent, t_azimuth, t_plunge,
+              n_exponent, n_azimuth, n_plunge, p_exponent, p_azimuth, p_plunge,
+              exponent*
+
+        scale: str
```
sgjholt, line 1048: One thing I noticed when I was using this code is that you cannot use the GMT options directly. So you can only plot full moment tensors (force double couple) by doing something like: 

`fig.meca(
    spec="focal_mechanisms.txt", convention='mt', scale="1", component="dc", offset="0.5p"
)`

Which is great for novice users, and is obviously more pythonic, but I feel like this would be a little frustrating for GMT vets - who still have valid reasons for wanting to use this API (for context I am not a GMT vet, but know people who are). 

Is is not possible to also allow this in addition to the keywords:

`fig.meca(
    spec="focal_mechanisms.txt", Sd='1', C="0.5p"
)` 

Or did you guys decide this was not okay in the grand scheme of things?

When I tried my hand at [this](https://github.com/sgjholt/pygmt/commit/775faa7637aa8fdc13bc372d4fa51987ca26134e)  that was the first thing I did (this version is definitely far inferior overall - but I think allowing GMT options directly is a strength).

liamtoney: Hi there! Thanks for reviewing. Yes, you should be able to pass in a file to `spec`. Or an array containing the same info.

pygmt/base_plotting.py
```diff
@@ -994,3 +994,368 @@ def text(
...
+                    # delete optional parameters from copy for convention check
+                    for item in drop_list:
+                        del spec_conv[item]
+                else:
+                    spec_conv = spec
+
+            # set convention and focal parameters based on spec convention
+            convention_assigned = False
+            for conv in param_conventions:
+                if set(spec_conv.keys()) == set(param_conventions[conv]):
```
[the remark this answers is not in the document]
liamtoney: And I think as a constant `param_conventions` is supposed to be all caps. Not sure how we handle constants though. Not a big deal.

weiji14:
> One concern I have is that the DataFrame input option seems to add a lot of complexity to the code. I'm not too familiar with how we've integrated DataFrames into our other wrappers, so I can't really review that. @weiji14 any thoughts?

One solution would be to turn dictionary arguments into a pandas Dataframe and then just handle everything via pandas. But we'll need to sort the columns in the right order, and that's extra complexity too! Best to leave it until the refactoring process in a separate PR. I'm fine with the current implementation as long as the functionality right now is up to standard,

pygmt/base_plotting.py
tjnewton: I was so excited to get away from GMT syntax that I didn't think about making that a possibility. Shouldn't be too difficult. I'll implement this over the weekend. It's neat to see your previous work on this!

pygmt/base_plotting.py
```diff
@@ -994,3 +994,353 @@ def text(
...
+
+    @fmt_docstring
+    @use_alias(R="region", J="projection", B="frame", C="offset")
+    @kwargs_to_strings(R="sequence",)
+    def meca(
+        self,
+        spec,
+        scale,
+        lon=None,
+        lat=None,
```
seisman, line 1006: Should we use `longitude` and `latitude` instead? My feeling is most Python package use long, descriptive argument names instead of short names.

pygmt/base_plotting.py
seisman: > Is is not possible to also allow this in addition to the keywords:
> 
> fig.meca( spec="focal_mechanisms.txt", Sd='1', C="0.5p" )

PyGMT is discouraging the use of short-form arguments. See https://github.com/GenericMappingTools/pygmt/issues/262 for the reasons.

pygmt/base_plotting.py
```diff
@@ -994,3 +994,353 @@ def text(
...
+
+                # Add in plotting options, if given, otherwise add 0s as
+                # required by GMT
+                for arg in plot_lon, plot_lat:
+                    if arg is None:
+                        spec.append(0)
+                    else:
+                        if "C" not in kwargs:
+                            kwargs["C"] = True
+                        spec.append(arg)
```
seisman, line 1213: I didn't read the codes carefully. Just want to let you know that, since GMT 6.1, plot_lon and plot_lat are not required.

tjnewton:
> Hi @tjnewton, just wanted to say that you've done a terrific job! I'll leave the proper review for @liamtoney, @seisman (or @sgjholt who was interested in this [on the forum](https://forum.generic-mapping-tools.org/t/pygmt-psmeca-option/692)) since they would know more about focal mechanism plots, but I'm quite happy with the `meca` tests here and think this would make a great addition to PyGMT.
> 
> In terms of the pylint score, 10/10 would be ideal of course, but the warnings are all refactor related (that's what the **R**1234 stands for), and I think we can open up a separate PR to handle that since this one is already getting very long. If you think you can fix any of them, do it now, otherwise just silence them temporarily (see https://stackoverflow.com/questions/35469253/silence-pylint-warning-about-unused-variables-for-string-interpolation).

Thanks, I appreciate that! I silenced those pylint warnings for now.

pygmt/base_plotting.py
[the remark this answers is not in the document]
tjnewton: I pushed your suggested changes. Thanks for the input! pylint gets angry at this as an all caps constant so I changed back to lowercase for now.

pygmt/base_plotting.py
```diff
@@ -994,3 +994,360 @@ def text(
...
+                    if (
+                        isinstance(lon, list)
+                        or isinstance(lat, list)
+                        or isinstance(depth, list)
+                    ):
+                        if (len(lon) != len(lat)) or (len(lon) != len(depth)):
+                            raise GMTError(
+                                "Unequal number of focal mechanism "
+                                "locations supplied."
+                            )
```
seisman, line 1243: I have a strong feeling that these codes can be greatly simplified into one or two lines by converting everything to numpy array first.

My another concern is that, here you check if lon, lat, and depth are list, but users may also pass 1d numpy arrays to lon, lat, and depth.

pygmt/base_plotting.py
tjnewton: Thanks for pointing this out. Pushing this change now.

pygmt/base_plotting.py
tjnewton: The purpose of this block is to make sure an array will not be built from lists of unequal lengths. I could do the same thing by making a numpy array, mapping it to the correct size, then searching for missing values but in my opinion that is less readable code. If you have a better idea I'm happy to implement it.

Longitude, latitude, and depth can not be 1d numpy arrays. We specify int float and list types in the docstring.

seisman: All other PyGMT functions can accept 1d numpy arrays. For example, the *x*, *y* and *sizes* arguments of the `plot` function can be int, float, list or numpy arrays. `meca` should also support it.

tjnewton: done

pygmt/base_plotting.py
tjnewton: Got it

tjnewton:
> Hi @tjnewton, this is looking like it covers pretty much all the bases at this point. Could you merge in the master branch? We can see what the test results do after that, but I'm ready to approve this. This PR is getting mighty long!
> 
> I do think some refactoring may be required, but as @weiji14 suggests that could happen later.

Sounds good. I have addressed all immediate concerns here and I'm happy to address refactoring in a new PR.

pygmt/tests/data/focal_mechanisms.txt
```diff
@@ -0,0 +1 @@
+-127.43 40.81 12 -3.19 1.16, 3.93, -1.02, -3.93, -1.02, 23
```
seisman, line 1: This file is only used in one test. It's better to write the focal mechanism to a temporary file, plot it and then remove it.

pygmt/tests/test_meca.py
```diff
@@ -0,0 +1,282 @@
...
+    # Mixed
+    fig.meca(
+        dict(strike=10, dip=35, rake=129, magnitude=5),
+        longitude=3.4,
+        latitude=0.6,
+        depth=0,
+        scale="2.5c",
+    )
+
+    return fig
```
seisman, line 86: Why plot so many beachballs in this test?

seisman: Sorry to be too picky here. We're trying our best not to add big images into the repository, as it will significantly increase the whole repository size. 

All other tests looks good to me. This test figure is as large as ~50 KB, but most beachballs seem unnecessary to me.

seisman:
> Nice! Looks like we've got everything except for the two code coverage checks. I'm approving this, but I want to wait on merging until we get input from a core dev — not sure how critical these two checks are.

The coverage checks mean this PR needs more tests, but I'm OK with the current status of this PR (except my two comments above). We can refactor the codes and add more tests later.

pygmt/tests/test_meca.py
liamtoney: Whoops, this is another leftover from when I was just wrapping the GMT tests. Agreed.

tjnewton: Addressed in 50f582c

pygmt/tests/data/focal_mechanisms.txt
tjnewton: Addressed in b173b49

pygmt/tests/test_meca.py
```diff
@@ -0,0 +1,241 @@
...
+    Test supplying a file containing focal mechanisms and locations to the
+    `spec` argument.
+    """
+
+    fig = Figure()
+
+    focal_mechanism = [-127.43, 40.81, 12, -3.19, 1.16, 3.93, -1.02, -3.93, -1.02, 23]
+
+    # writes temp file to pass to gmt
+    with open(os.path.join(TEST_DATA_DIR, "temp.test"), mode="w") as temp_file:
```
liamtoney, line 190: I think this is a really good application for the GMTTempFile, docstring w/ usage example below:

https://github.com/GenericMappingTools/pygmt/blob/2884fd04a20ef6281c33d75de6c4521ba968b032/pygmt/helpers/tempfile.py#L27

liamtoney: I think the usage example would be:
```python
with GMTTempFile() as temp:
    with open(temp.name, mode="w") as temp_file:
        temp_file.write(...)
    some_gmt_module(temp.name, ...)
```
Then you don't need to use the `os.remove()` since we have the context manager. LMK if that has issues!

tjnewton: Oooo thanks Liam, that's clever! Testing this test is a little slow on my end because `make test` on my machine freezes on `../pygmt/tests/test_grdview.py::test_grdview_grid_dataarray` every time starting today -_-

tjnewton: Finally passed after a restart but the grdview tests are very slow now. Looks like test_meca is passing now.

weiji14: Sorry about the slow test (we'll need to streamline this somehow at some point). A workaround is to only run the meca tests locally. Use `pytest -v --mpl pygmt/tests/test_meca.py`.

seisman: The changes in 6c3af68c0364d2d163909fd648716b1bd26ca1be doesn't make use of `GMTTempFile`. The example given by @liamtoney is correct. In that example, the temporary file is written to the file defined by the variable `temp.name` and deleted when exiting GMTTempFile context. You should follow that example instead.

tjnewton: Thanks @weiji14 that's very helpful.

seisman: FYI, you can always use a command like below to run the tests in only one script:
```
pytest --mpl --mpl-results-path=results pygmt/tests/test_meca.py
```

tjnewton: @seisman can you please explain? I used GMTTempFile in what I believe is the correct syntax. Including the `as temp` that Liam mentioned in the second line will throw a pylint warning for an unused variable. The relevant block is:
```    with GMTTempFile():
        with open("temp.name", mode="w") as temp_file:
            temp_file.write(" ".join([str(x) for x in focal_mechanism]))
        # supply focal mechanisms to meca as a file
        fig.meca(
            "temp.name",
            convention="mt",
            component="full",
            region=[-128, -127, 40, 41],
            scale="2c",
            projection="M14c",
        )```

seisman: Your codes write the data to a file called `temp.name`, and the file is not deleted. 

You can think GMTTempFile as a wrapper of the Python [`tempfile.TemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile). 

A similar use case is: 
https://github.com/GenericMappingTools/pygmt/blob/4cfb7a2fdc165e8cdeeab88027b3fb09e0eaa134/pygmt/tests/test_text.py#L284-L297

seisman: > Including the `as temp` that Liam mentioned in the second line will throw a pylint warning for an unused variable.

After using `with GMTTempFile() as temp:`, you should also remove the double quotes around `"temp.name"`.

tjnewton: I see, so the temp.file object needed to be passed to meca rather than a temporary file name string. Should be fixed now.

seisman: Yes, I think this PR is good to merge. Thanks for your contribution!

liamtoney: `temp.name` is a string with the system filename.

tjnewton: thank @liamtoney I apparently need some rest

seisman:
@tjnewton Again, thanks for your contribution to PyGMT. Feel free to open another PR to add yourself to [the AUTHORS list](https://github.com/GenericMappingTools/pygmt/blob/master/AUTHORS.md) after this PR is merged.

tjnewton:
> @tjnewton Again, thanks for your contributions to PyGMT. Feel free to open another PR to add yourself to [the AUTHORS list](https://github.com/GenericMappingTools/pygmt/blob/master/AUTHORS.md) after this PR is merged.

Will do, thanks! I plan to continue contributing to the project.

weiji14:
Thanks @tjnewton, really appreciate the work you've done here! :smile:

@liamtoney, could you squash and merge in this PR since you've worked on it too? Will need to shorten the commit message to capture the gist of what was achieved in this PR.

liamtoney:
Congrats @tjnewton for your first contribution to PyGMT! Definitely let us know how we could improve this process and make it more accessible. I think a lot of seismologists will be excited about this implementation!

Issue #515, opened 2020-07-11
Global earth relief data loaded as xarray doesn't work if projection center is not 0


seisman:
**Description of the problem**

It may work in GMT 6.0.0 + PyGMT v0.1.2, or not. Currently, PyGMT master + GMT 6.1.0 gives me a wrong result. 

**Full code that generated the error**

```python
import pygmt
from pygmt.datasets import load_earth_relief

grid = load_earth_relief()
fig = pygmt.Figure()
fig.grdimage(grid, cmap="geo", region='d', projection="H10c")

fig.shift_origin(xshift="12c")
fig.grdimage(grid, cmap="geo", region='g', projection="H10c")

fig.shift_origin(xshift="-12c", yshift="6c")
fig.grdimage(grid, cmap="geo", region='d', projection="H120/10c")

# this one sometimes crashes
#fig.shift_origin(xshift="12c")
#fig.grdimage(grid, cmap="geo", region='g', projection="H120/10c")
fig.savefig("map.png")
```

**Output**

![image](https://user-images.githubusercontent.com/3974108/88252841-a5a12080-cc7d-11ea-8db7-42d3409d7509.png)

weiji14:
Not sure if #476 will fix this. I just refactored the code there but it's still giving the strange static-looking image in the gallery at https://pygmt-24935dva1.vercel.app/gallery/grid/track_sampling.html#sphx-glr-gallery-grid-track-sampling-py. This is with the pixel registered earth relief grid.

![Sampling along tracks image distorted](https://user-images.githubusercontent.com/23487320/87236343-50267180-c43c-11ea-8c28-005070575bed.png)

seisman:
@PaulWessel Please see if you have any idea why it fails. FYI, we're still using `GMT_IN|GMT_IS_REFERENCE` mode.

PaulWessel:
Remind me: The xarray is being passed into GMT as a grid via a GMT_MATRIX structure?

seisman:
> Remind me: The xarray is being passed into GMT as a grid via a GMT_MATRIX structure?

Yes, `GMT_IS_GRID|GMT_VIA_MATRIX`.

PaulWessel:
In gmtapi_import_grid:

```
case GMT_IS_REFERENCE|GMT_VIA_MATRIX:	/* The user's 2-D grid array of some sort, + info in the args [NOT YET FULLY TESTED] */
[...]

PR #464, opened 2020-05-30
Support various datetime types as input

seisman:
**Description of proposed changes**

This PR allows PyGMT accept vectors of various datetime types so that PyGMT can plot datetime axis. 

Vectors of following datetime types are supported:

- [x] [numpy.datetime64](https://numpy.org/doc/stable/reference/arrays.datetime.html)
- [x] [pandas.DateTimeIndex](https://pandas.pydata.org/docs/user_guide/timeseries.html)
- [x] Raw datetime strings in both ISO and non-ISO formats, e.g., `2010-01-01`, `1/1/2018`, `Jul 5, 2019`
- [x] Python's [built-in datetime and date](https://docs.python.org/3/library/datetime.html)

TODO:

- [x] Wait for the GMT 6.1.0 release
- [ ] Add more tests
- [ ] Add a tutorial for calendar axis
- [x] Test for xarray.DataArray

Known issues but won't be fixed in this PR:

- [ ] The `region` argument doesn't work with numpy `datetime64` objects, have to convert to a string using [`np.datetime_as_string`](https://numpy.org/devdocs/reference/generated/numpy.datetime_as_string.html) in order to set the map frame bounds.
- [ ] Support [`timedelta64`](https://pandas.pydata.org/pandas-docs/stable/user_guide/timedeltas.html), e.g. GPS time from a certain epoch. 

Fixes #242.

**Example script:**
```python
import datetime
import pygmt
import numpy as np
import pandas as pd

fig = pygmt.Figure()
fig.basemap(projection="X15c/5c", region="2010-01-01/2020-01-01/0/10", frame=True)

# numpy.datetime64 types
x = np.array(
    ["2010-06-01", "2011-06-01T12", "2012-01-01T12:34:56"], dtype="datetime64"
)
y = [1.0, 2.0, 3.0]
fig.plot(x, y, style="c0.2c", pen="1p")

# pandas.DatetimeIndex
x = pd.date_range("2013", freq="YS", periods=3)
y = [4, 5, 6]
fig.plot(x, y, style="t0.2c", pen="1p")

# xarray.DataArray
x = xr.DataArray(data=pd.date_range(start="2015-03", freq="QS", periods=3))
y = [7.5, 6, 4.5]
fig.plot(x, y, style="s0.2c", pen="1p")

# raw datetime strings
x = ["2016-02-01", "2017-03-04T00:00"]
y = [7, 8]
fig.plot(x, y, style="a0.2c", pen="1p")

# the Python built-in datetime and date
[...]

PR #520, opened 2020-07-13
Wrap GMT_Put_Strings to pass str columns into GMT C API directly

weiji14:

Used to insert 1D numpy arrays of string type from PyGMT directly into GMT via the [C API](https://docs.generic-mapping-tools.org/6.1/api.html#gmt-c-api). Saves us having to write to intermediate temporary CSV files using `pandas`. Needed for #483 and potentially #516.

I'll definitely need help with this one, this is really stretching my knowledge about C programming (read: no idea if what I'm doing makes sense).

See also:
- https://docs.generic-mapping-tools.org/6.1/api.html#gmt-put-strings
- Wrap GMT_Put_Vector and GMT_Write_Data (#74) and Wrap GMT_Put_Matrix (#75)

Fixes #

**Reminders**


pygmt/clib/session.py
```diff
@@ -791,11 +792,62 @@ def put_vector(self, dataset, column, vector):
...
+            copy. Use :func:`numpy.ascontiguousarray` to make sure your vector
+            is contiguous (it won't copy if it already is).
+
+        Parameters
+        ----------
+        dataset : :class:`ctypes.c_void_p`
+            The ctypes void pointer to a ``GMT_Dataset``. Create it with
+            :meth:`~gmt.clib.Session.create_data`.
+        column : int
+            The column number of this vector in the dataset (starting from 0).
```
seisman, line 817:
```C
int GMT_Put_Strings (void *API, unsigned int family, void *X, char **array);
```
Unlike numeric vectors (i.e., GMT_Put_Vector), you can't put strings at a specific column. In GMT dataset, there is only a pointer to hold the string array. It has to be the "last" column. That's also why have to combine all trailing arrays into a single string array and then pass to GMT.

weiji14: Ah ok, I'll take out the column argument then.

pygmt/clib/session.py
```diff
@@ -791,11 +792,62 @@ def put_vector(self, dataset, column, vector):
...
+        Raises
+        ------
+        GMTCLibError
+            If given invalid input or ``GMT_Put_Strings`` exits with status !=
+            0.
+
+        """
+        c_put_strings = self.get_libgmt_func(
+            "GMT_Put_Strings",
+            argtypes=[ctp.c_void_p, ctp.c_uint, ctp.c_void_p, ctp.c_uint],
```
seisman, line 831: It's also tricky to pass string arrays `char **` via ctypes. 

I'm not familiar with ctypes. Below are the codes I wrote to pass datetime strings to GMT. Datetime strings are just special strings that GMT can recognize. To pass strings to GMT, I need to use codes at lines 775-778.

https://github.com/GenericMappingTools/pygmt/blob/f401f8527cfee4dc8f48d78ad93d23b678ecfbc8/pygmt/clib/session.py#L773-L783

Perhaps there are other better and simpler ways that I don't know.

weiji14: And I have no C programming experience at all :laughing:. I'll try my best, but you might actually be better placed to handle this PR, just feel free to make any changes.

seisman: OK. I think I can contribute a little.

pygmt/clib/session.py
```diff
@@ -56,6 +56,7 @@
     np.uint64: "GMT_ULONG",
     np.uint32: "GMT_UINT",
     np.datetime64: "GMT_DATETIME",
+    np.str_: "GMT_TEXT",
```
weiji14, line 59: Can we figure out if `GMT_TEXT` is the right enum here? Or should it be `GMT_CHAR` or something else?

seisman: I don't think so. We even don't need this. Unlike GMT_Put_Vector, which can accept float, double or int, GMT_Put_Strings only accept strings. Thus we don't need to pass *gmt_type*.

seisman: See this link for available GMT types: https://github.com/GenericMappingTools/gmt/blob/e4aab495643fb914b1a642cc6de4ffff041e1d62/src/gmt_resources.h#L84

weiji14: You're right. looking at the docs at https://docs.generic-mapping-tools.org/6.1/api.html#gmt-put-strings, it says:

>     int GMT_Put_Strings (void *API, unsigned int family, void *X, char **array);
> 
> where family is either GMT_IS_VECTOR or GMT_IS_MATRIX, X is either a GMT_VECTOR or GMT_MATRIX, and array is the a pointer to your string array.

Does that mean we need to pass in `GMT_VECTOR`/`GMT_MATRIX` instead of *gmt_type*? I'm confused as to how this works.

seisman: We can attach the strings to a vector or a matrix. *X* can be a GMT_VECTOR or GMT_MATRIX, which is created by the `create_data` function. Thus, we have to pass *family* (`either GMT_IS_VECTOR or GMT_IS_MATRIX`) to let GMT know the type of *X*.

pygmt/tests/test_clib_put.py
```diff
@@ -0,0 +1,192 @@
...
+                "GMT_IS_VECTOR",
+                "GMT_IS_POINT",
+                "GMT_WRITE_SET",
+                wesn,
+                tmp_file.name,
+                dataset,
+            )
+            # Load the data and check that it's correct
+            news = tmp_file.loadtxt(unpack=True, dtype=np.str)
+            npt.assert_allclose(news, s)
```
weiji14, line 39: > We can attach the strings to a vector or a matrix. _X_ can be a GMT_VECTOR or GMT_MATRIX, which is created by the `create_data` function. Thus, we have to pass _family_ (`either GMT_IS_VECTOR or GMT_IS_MATRIX`) to let GMT know the type of _X_.

Right, so we can't pass in *just* strings then, but need to tie it to an existing Vector or Matrix? I've tried that (haven't committed it to Github) but am now getting segfaults, which seems promising in a way.

seisman: I think *dataset* is the Vector or Matrix.

seisman: I mean, what we want to do is tie the strings to the dataset (the dataset is either a vector or a matrix).

seisman: Reading this C code (https://github.com/GenericMappingTools/gmt/blob/master/src/testapi_vector_strings.c) may be useful for you.

weiji14: Yes that is helpful, I was reading the GMT.jl one at https://github.com/GenericMappingTools/GMT.jl/blob/master/src/pstext.jl and got really lost.

weiji14: Ah, I think I got it! I was counting the `strings` array as a column using `dim=[4, 5, 1, 0],  # columns, rows, layers, dtype`, but the columns should not be counted. I.e. an array with `x, y, angle, strings` would just use `dim=3` instead of `dim=4`. Hold on, I'll make a commit.

seisman: PyGMT passes vectors, matrix or grids to GMT. As I understand it, GMT.jl is using a different way passing data to GMT. 

GMT.jl wraps the GMT data types like GMT_DATASET, GMT_GRID, GMT_PALETTE (https://docs.generic-mapping-tools.org/latest/api.html#gmt-resources). So what GMT.jl do is attaching vectors to the GMT_DATASET in Julia, and they pass the GMT_DATASET directly to GMT. This is why we find so many GMT API bugs but everything work for GMT.jl.

weiji14: Ok, maybe not. Using `dim=2` runs without a segfault in 42af00ef8b16afd6c844d648860a6435338dd864, but the 'abcdef ' `text` is not output to the textfile, only the `x` and `y` array data is written. Maybe we do need to use `dim=3`.

weiji14: Actually I'm not sure if it did work, or if it's just that `GMT_Put_Strings` doesn't write the strings to the tmp_file. I'll need to either wrap `GMT_Get_Strings` or refactor `text` to use `put_strings` to be sure. Anyways, dinner first.

seisman: @weiji14 Here is a minimal example for experimenting with passing strings array to C. I just post it here for your reference. 

The C code (test.c):
```c
#include <stdio.h>

int myfunc (char **array)
{
	fprintf(stderr, "%s %s %s\n", array[0], array[1], array[2]);
	return 0;
}
```

To compile the C code into a shared library, run:
```bash
gcc -fpic -c test.c
gcc -shared -o test.so test.o
```
The following Python code can correctly pass strings array to the myfunc function in test.so library:
```python
import ctypes

lib = ctypes.CDLL("test.so")
print(lib)
string = ["ABC", "DEF", "GHI"]
lib.myfunc.restype = ctypes.c_int

arr = (ctypes.c_char_p * len(string))()
arr[:] = [s.encode() for s in string]
lib.myfunc(arr)
```
The Python trick comes from https://stackoverflow.com/questions/3494598/. It works but I don't understand the details, and don't know if it's the simplest way.

pygmt/tests/test_clib_put.py
```diff
@@ -0,0 +1,197 @@
+"""
```
seisman, line 1: What about separating the tests into three files:

- test_clib_put_vector.py
- test_clib_put_matrix.py
- test_clib_put_strings.py

weiji14: Yes, that can be arranged.

pygmt/clib/session.py
```diff
@@ -791,11 +791,61 @@ def put_vector(self, dataset, column, vector):
...
+        c_put_strings = self.get_libgmt_func(
+            "GMT_Put_Strings",
+            argtypes=[ctp.c_void_p, ctp.c_uint, ctp.c_void_p, ctp.c_void_p],
+            restype=ctp.c_int,
+        )
+
+        strings_pointer = strings.ctypes.data_as(ctp.c_char_p)
+        status = c_put_strings(
+            self.session_pointer, self[family], dataset, strings_pointer
+        )
```
weiji14, line 838: > The following Python code can correctly pass strings array to the myfunc function in test.so library:
> 
> ```python
> # ...
> arr = (ctypes.c_char_p * len(string))()
> arr[:] = [s.encode() for s in string]
> lib.myfunc(arr)
> ```
> 
> The Python trick comes from https://stackoverflow.com/questions/3494598/. It works but I don't understand the details, and don't know if it's the simplest way.

Yes I tried this following your datetime example, but I think (?) it's functionally equivalent to [`strings.ctypes.data_as(ctp.c_char_p)`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.ctypes.html)? Either way it runs, but the string data doesn't seem to get written to the virtualfile so the test fails. I've also tried `strings.ctypes.data_as(ctp.POINTER(ctp.c_char_p))` but it doesn't seem to make a difference. :confused:

seisman: `strings.ctypes.data_as(ctp.c_char_p)` may not work. 

What I did is adding a print statement inside the GMT_Put_Strings function, printing the string arrays passed to it. Using `strings.ctypes.data_as(ctp.c_char_p)` crashes immediately for me, but using my code can print the strings correctly, but then crash 🤦‍♂️.

weiji14: How about using:

```python
c_put_strings = self.get_libgmt_func(
    "GMT_Put_Strings",
    argtypes=[
        ctp.c_void_p,
        ctp.c_uint,
        ctp.c_void_p,
        ctp.POINTER(ctp.c_char_p),
    ],
    restype=ctp.c_int,
)
strings_pointer = (ctp.c_char_p * len(strings))()
strings_pointer[:] = np.char.encode(strings)
```

Does it crash for you?

seisman: Yes, it still crashes, but I can see that the strings are correctly passed to the GMT_Put_Strings function.

weiji14: What about `strings.ctypes.data_as(ctp.POINTER(ctp.c_char_p))`? Are the strings correctly passed in?

seisman: No, just segmentation fault

seisman: After a little debugging on the GMT C code, it looks a potential GMT bug.

seisman: See https://github.com/GenericMappingTools/gmt/pull/3718 for the bug report.

weiji14: Yep, I was just taking a look at it (thanks for tracking down the bug by the way!). Sounds like we'll need to pin to GMT > 6.1.1 for the next release?

seisman: Actually I'm not 100% sure https://github.com/GenericMappingTools/gmt/pull/3718 is related to the issue here. I just had the feeling that the string array may be freed by Python when GMT tries to read it. 

Hopefully, GMT 6.1.1 can fix issue and #515.

pygmt/clib/session.py
```diff
@@ -793,11 +793,68 @@ def put_vector(self, dataset, column, vector):
...
+            ],
+            restype=ctp.c_int,
+        )
+
+        strings_pointer = (ctp.c_char_p * len(strings))()
+        strings_pointer[:] = np.char.encode(strings)
+
+        status = c_put_strings(
+            self.session_pointer, self[family], dataset, strings_pointer
+        )
```
seisman, line 847: With the upstream PR https://github.com/GenericMappingTools/gmt/pull/3718 and some changes in the Python code (I don't push the changes yet), the `GMT_Put_Strings` wrapper works well. Please see the full discussion in that PR, especially comments https://github.com/GenericMappingTools/gmt/pull/3718#issuecomment-664084529 and https://github.com/GenericMappingTools/gmt/pull/3718#issuecomment-664089340.

For GMT_Put_Vector, GMT_Put_Matrix, GMT_Put_Strings, they don't duplicate the vector, matrix or strings by default. So if the vector, matrix, or strings are freed in Python, we should see some unexpected behaviors or crashes. The situation never happens for GMT_Put_Vector and GMT_Put_Matrix, but happens for GMT_Put_Strings.

vectors/matrix and strings are passed using different ways. Here is how we pass vectors,
```
vector_pointer = vector.ctypes.data_as(ctp.c_void_p)
```
here is how we pass strings)
```
        strings_pointer = (ctp.c_char_p * len(strings))()
        strings_pointer[:] = np.char.encode(strings)
```
As I understand it, `vector_pointer` is a pointer pointing to the original `vector` memory, but `strings_pointer` looks like a new pointer (I don't understand the codes. Maybe I'm wrong here) and is freed when exiting the function.

For strings, if we have a way to pass data just like vectors or matrix, then we probably don't need PR #3718, which duplicates the strings.

weiji14: Ideally we could just do `strings_pointer = strings.ctypes.data_as(ctp.c_char_p)`. The only difference seems to be `ctp.c_void_p` and `ctp.c_char_p`. No idea about how memory is freed/handled by Python/NumPy, I would think that they're both just a *direct* pointer to the same thing in memory (Printing the `c_char_p`/`c_void_p` pointer yields the same number (something like `123456789`) from what I recall), but I'm not too sure here either. Might have to wait for [Hurrican Douglas](https://en.wikipedia.org/wiki/Hurricane_Douglas_(2020)) to pass over Hawaii before we can test this properly.

pygmt/clib/session.py
```diff
@@ -1098,8 +1158,12 @@ def virtualfile_from_vectors(self, *vectors):
             family, geometry, mode="GMT_CONTAINER_ONLY", dim=[columns, rows, 1, 0]
         )
 
-        for col, array in enumerate(arrays):
+        # Use put_vector for first n columns with numerical type data
+        for col, array in enumerate(arrays[:columns]):
             self.put_vector(dataset, column=col, vector=array)
+        # Use put_strings for last column with string type data
+        for array in arrays[columns:]:
+            self.put_strings(dataset, family="GMT_IS_VECTOR", strings=array)
```
seisman, line 1166: These codes are incorrect. We can't call `GMT_Put_Strings` multiple times.

See the comments in https://github.com/GenericMappingTools/pygmt/issues/483#issuecomment-647251367, we have to combine all trailing string arrays into one array and call `GMT_Put_Strings` once.

weiji14: It does just call put strings once, or zero times. If you examine the code before, the slice either returns a list with one item or a list with zero items.

seisman: Do you mean these lines?

```python
        columns = len(arrays)
        if np.issubdtype(arrays[-1].dtype, np.str_):
            columns -= 1
```

It won't work for input with multiple trailing string arrays. For example, for the `text` module, the input array may be
```
x = [1, 2]
y = [3, 4]
font = ["15p,1,blue", "15p,1,red"]
text = ["labelA", "labelB"]
```
We have to combine `font` and `text` arrays into one single string array before passing it via GMT_Put_Strings.
```
newarray = ["15p,1,blue labelA", "15p,1,red labelB"]
```

seisman:
I just pushed one commit (2485437302e8224e8325e5d956d89f7c3ed6e7e3). With this commit, the `text()` module works well with the GMT PR https://github.com/GenericMappingTools/gmt/pull/3718.

Note that the test `test_put_strings` still fails, most likely because `write_data` doesn't output strings at all.

pygmt/clib/session.py
weiji14: Ok I see, but do we want to the concatenation:

1. **Before** passing it into `lib.virtualfile_from_vectors`, i.e. at the `fig.text` (and also `fig.meca`, etc) level
2. **In** the clib itself, i.e. find all the arrays that have a 'str' data type, and concatenate them inside `lib.virtualfile_from_vectors` before passing it into `put_strings`.

seisman: Option 2 is easier and more intuitive for module wrapper developers.

pygmt/clib/session.py
seisman: FYI, https://github.com/GenericMappingTools/gmt/pull/3718 is already merged into GMT master, and will be backported to 6.1 branch soon.

pygmt/clib/session.py
weiji14: Not so easy for the `clib` developer though :laughing:. I do agree with going with Option 2, but I'm not sure how the memory pointers would work if we concatenate two numpy arrays with 'str' dtypes. It might be that we'll be duplicating memory somehow.

pygmt/clib/session.py
weiji14: Great! I'll test it out when I get the time (need to prepare some stuff for an online conference next week). Still need to wait on the grid problem at #515 but that's a separate issue.

pygmt/clib/session.py
seisman: From what I now about Python and GMT, I think we may have to duplicate the memory in PyGMT, and then duplicate it again in GMT.

pygmt/clib/session.py
seisman: Just triggered the GMT master branch testing. Ideally, we should only see one failure from `test_put_strings`.

weiji14:
> I just pushed one commit ([2485437](https://github.com/GenericMappingTools/pygmt/commit/2485437302e8224e8325e5d956d89f7c3ed6e7e3)). With this commit, the `text()` module works well with the GMT PR [GenericMappingTools/gmt#3718](https://github.com/GenericMappingTools/gmt/pull/3718).
> 
> Note that the test `test_put_strings` still fails, most likely because `write_data` doesn't output strings at all.

Good to see that only the `test_put_strings` are failing on GMT master now. If `write_data` doesn't output strings as you say, do you know how we can unit test that `put_strings` works (ideally without relying on test_text_*)?

seisman:
> > I just pushed one commit ([2485437](https://github.com/GenericMappingTools/pygmt/commit/2485437302e8224e8325e5d956d89f7c3ed6e7e3)). With this commit, the `text()` module works well with the GMT PR [GenericMappingTools/gmt#3718](https://github.com/GenericMappingTools/gmt/pull/3718).
> > Note that the test `test_put_strings` still fails, most likely because `write_data` doesn't output strings at all.
> 
> Good to see that only the `test_put_strings` are failing on GMT master now. If `write_data` doesn't output strings as you say, do you know how we can unit test that `put_strings` works (ideally without relying on test_text_*)?

I have no idea. Perhaps it's a bug in GMT_Write_Data, or we just don't pass the correct arguments.

seisman:
Ping @PaulWessel for a quick question.

Currently, PyGMT has wrapped the `GMT_Write_Data` function. It can output the numeric vectors or matrix, but it can't output the text strings. So, for data table, does `GMT_Write_Data` function outputs the text strings by default? Are there any modes which controls the output of text strings?

PaulWessel:
What is the object you pass to GMT_Write_Data?  Are you passing a GMT_IS_DATASET or is this done via vectors with string columns?  I see gmtapi_write_vector is not even bothering to write strings, so if that is what you do then I need to add that bit.

seisman:
> What is the object you pass to GMT_Write_Data? Are you passing a GMT_IS_DATASET or is this done via vectors with string columns? I see gmtapi_write_vector is not even bothering to write strings, so if that is what you do then I need to add that bit.

We create a dataset with `GMT_IS_DATASET|GMT_VIA_VECTOR`, and then call `GMT_Put_Vectors` and `GMT_Put_Strings` to attach the vectors and strings to the dataset. When writing, the family is `GMT_IS_VECTOR`, geometry is `GMT_IS_POINT`, mode is `GMT_WRITE_SET`.

weiji14:
This is the create_data code:

https://github.com/GenericMappingTools/pygmt/blob/7fa1f0c7332895cf1322b0f7dedb2b6226e1594c/pygmt/tests/test_clib_put_strings.py#L15-L20

And this is the write_data code:

https://github.com/GenericMappingTools/pygmt/blob/7fa1f0c7332895cf1322b0f7dedb2b6226e1594c/pygmt/tests/test_clib_put_strings.py#L31-L38

PaulWessel:
OK, I can fix the writing.  Currently no text is considered in the output.  I will do the same for matrix even though you may not be using that.
The same issue affects reading vectors from file - no provision for trailing text.  I am not sure if you are using _GMT_Read_Data_ to read into a **GMT_VECTOR** but if you do then you will not get any trailing text.  I will look at that as well.

seisman:
> The same issue affects reading vectors from file - no provision for trailing text. I am not sure if you are using _GMT_Read_Data_ to read into a **GMT_VECTOR** but if you do then you will not get any trailing text. I will look at that as well.

Currently, PyGMT doesn't wrap `GMT_Read_Data`, but I believe it will.

pygmt/clib/session.py
```diff
@@ -1087,6 +1148,16 @@ def virtualfile_from_vectors(self, *vectors):
...
         columns = len(arrays)
+        # Find arrays that are of string dtype from column 3 onwards
+        # Assumes that first 2 columns contains coordinates like longitude
+        # latitude, or datetime string types.
+        str_cols = [
+            col + 2
+            for col, array in enumerate(arrays[2:])
+            if np.issubdtype(array.dtype, np.str_)
+        ]
+        columns -= len(str_cols)
```
weiji14, line 1159: I've hardcoded things here so that the first two columns will never be passed into `put_strings`, since we can have `datetime` strings in the first 2 columns (x and y) since #464. Problem is that this hardcoding prevents having a `datetime` column on the 3rd column, which might affect `plot3d` (at #471) if someone decides to plot a time axis on the z-axis. I'm not sure how often anyone would plot time on the z-axis, but maybe we should find a way around this hardcoding.

pygmt/clib/session.py
```diff
@@ -1098,8 +1169,28 @@ def virtualfile_from_vectors(self, *vectors):
...
+                    np.char.add(arrays[str_cols[0]], " "), arrays[str_cols[1]]
+                )
+            else:
+                raise NotImplementedError(
+                    f"Unable to handle {len(str_cols)} columns of string arrays. "
+                    "Please use only 1 or 2 instead."
+                )
+            self.put_strings(
+                dataset, family="GMT_IS_VECTOR|GMT_IS_DUPLICATE", strings=strings
+            )
```
weiji14, line 1193: Ok, I've handled the issue at https://github.com/GenericMappingTools/pygmt/pull/520#discussion_r462591754 (was trickier than I thought) and also added some tests, Up to 2  'str' type columns can be passed into `virtualfile_from_vectors` now.

I've asked a question on stackoverflow at https://stackoverflow.com/questions/63332200/concatenate-multiple-numpy-string-arrays-with-a-space-delimiter to handle `n>2` cases, may refactor this later.

weiji14: Ok, refactored at 4f1ccda5a736ff807f68be1c6462dc7eda0a30cd.

pygmt/tests/test_clib_put_strings.py
```diff
@@ -0,0 +1,67 @@
...
+with clib.Session() as _lib:
+    gmt_version = Version(_lib.info["version"])
+
+
+@pytest.mark.xfail(
+    condition=gmt_version < Version("6.1.1"),
+    reason="GMT_Put_Strings only works for GMT 6.1.1 and above",
+)
+def test_put_strings():
+    "Check that assigning a numpy array of dtype str to a dataset works"
```
weiji14, line 22: Temporarily using xfail so all the tests will be green. If it looks alright, we can merge this in first and remove these `xfail`s after bumping the minimum required GMT version to 6.1.1. Can also do it when we refactor `text` to use `put_strings` :smile:

pygmt/clib/session.py
seisman: I don't think you recent changes make sense. If I understand your codes correctly, if someone pass arrays like the order below:
```
x = [0.5],
y = [0.5],
just = ["BL"],
angle = [90],
text = ["this is a label"]
```
you changes pass *x*, *y*, *angle* as the 0, 1 and 3 columns by calling *put_vectors*, and pass "BL this is a label" by calling `put_strings`. This won't work because column 2 is missing (I don't test it, because the **text** module actually don't allow pass *angle* after *justification*, but I think maybe there are other cases we may want to pass arrays like this).

weiji14: Yes I did think of this. I was thinking of modifying `lib.virtualfile_from_vectors` to explicitly have a `num_vectors` and `str_vectors` arguments, so we know which arrays are numerical and which are strings. This would affect a lot of code in the library though.

weiji14: Specifically, the `*extra_arrays` in `plot` wouldn't work if we change to something like `def virtualfile_from_vectors(num_arrays, str_arrays)`:

https://github.com/GenericMappingTools/pygmt/blob/62580f1f8de641ee14a21ae762dc207b3ae29595/pygmt/base_plotting.py#L553-L555

seisman: When looping over arrays, we need to check the **first** np.str_ string. In the example above, it's *just*. Then we can pass all vectors before *just* as vectors, and combine the remaining arrays as one np.str_ array (although *angle* is numeric), and pass it to `put_strings`.

weiji14: > When looping over arrays, we need to check the **first** np.str_ string.

Yes, but we would still need to skip over the first 2 columns (or 3 columns in the case of `plot3d`) because datetimes can be string types. Our check for datetime isn't particularly nice since it tries to convert the whole array:

https://github.com/GenericMappingTools/pygmt/blob/62580f1f8de641ee14a21ae762dc207b3ae29595/pygmt/clib/session.py#L721-L729

> In the example above, it's _just_. Then we can pass all vectors before _just_ as vectors, and combine the remaining arrays as one np.str_ array (although _angle_ is numeric), and pass it to `put_strings`.

Ok, I can handle the 'combine remaining arrays' part, but is it okay to pass in `angle` to `put_strings` too?

seisman: > Yes, but we would still need to skip over the first 2 columns (or 3 columns in the case of `plot3d`) because datetimes can be string types. Our check for datetime isn't particularly nice since it tries to convert the whole array:

Skipping the first 2 or 3 columns should work for most cases. If I understand it correctly, we can even pass all arrays as a single string, e.g., a single array `string = ["0.5 0.5 BL 90 this is a lable"]` should also work. Again, I haven't test it.

weiji14: Ok, I'll refactor the code to handle the 'combine remaining arrays', but will continue to skip the first 2 columns for now, until someone files a bug report for `plot` or `plot3d`.

pygmt/tests/test_clib.py
```diff
@@ -399,6 +402,47 @@ def test_virtualfile_from_vectors():
...
+    size = 5
+    x = np.arange(size, dtype=np.int32)
+    y = np.arange(size, size * 2, 1, dtype=np.int32)
+    strings = np.array(["a", "b", "c", "d", "e"], dtype=np.str)
+    with clib.Session() as lib:
+        with lib.virtualfile_from_vectors(x, y, strings) as vfile:
+            with GMTTempFile() as outfile:
+                lib.call_module("gmtinfo", f"{vfile} ->{outfile.name}")
+                output = outfile.read(keep_tabs=True)
+        bounds = "\t".join([f"<{i.min():.0f}/{i.max():.0f}>" for i in (x, y)])
```
weiji14, line 420: By the way, [`gmtinfo`](https://docs.generic-mapping-tools.org/6.1/gmtinfo) doesn't actually report anything for the text columns in the virtualfile. So at most, this test only checks that calling `put_strings` doesn't crash PyGMT.

pygmt/tests/test_clib_put_strings.py
```diff
@@ -0,0 +1,67 @@
...
+    with clib.Session() as lib:
+        dataset = lib.create_data(
+            family="GMT_IS_DATASET|GMT_VIA_VECTOR",
+            geometry="GMT_IS_POINT",
+            mode="GMT_CONTAINER_ONLY",
+            dim=[2, 5, 1, 0],  # columns, rows, layers, dtype
+        )
+        x = np.array([1, 2, 3, 4, 5], dtype=np.int32)
+        y = np.array([6, 7, 8, 9, 10], dtype=np.int32)
+        strings = np.array(["a", "b", "c", "d", "e"], dtype=np.str)
```
seisman, line 32: Please change the elements of the string array to different length, to test passing strings of variable lengths. For example:
```suggestion
        strings = np.array(["a", "bc", "defg", "hijklmn", "opqrst"], dtype=np.str)
```

pygmt/tests/test_clib_put_strings.py
```diff
@@ -0,0 +1,67 @@
...
+                "GMT_IS_VECTOR",
+                "GMT_IS_POINT",
+                "GMT_WRITE_SET",
+                wesn,
+                tmp_file.name,
+                dataset,
+            )
+            # Load the data and check that it's correct
+            newx, newy, newstrings = tmp_file.loadtxt(
+                unpack=True, dtype=[("x", np.int32), ("y", np.int32), ("text", "<U1")]
```
seisman, line 52: `("text", "<U1")` also need update, but I'm not sure what's the correct dtype.

pygmt/tests/test_clib.py
```diff
@@ -399,6 +402,47 @@ def test_virtualfile_from_vectors():
...
+@pytest.mark.xfail(
+    condition=gmt_version < Version("6.1.1"),
+    reason="GMT_Put_Strings only works for GMT 6.1.1 and above",
+)
+def test_virtual_from_vectors_one_string_column():
+    "Test passing in one column with string dtype into virtual file dataset"
+    size = 5
+    x = np.arange(size, dtype=np.int32)
+    y = np.arange(size, size * 2, 1, dtype=np.int32)
+    strings = np.array(["a", "b", "c", "d", "e"], dtype=np.str)
```
seisman, line 414: Again, a string array with variable lengths.

pygmt/tests/test_clib.py
seisman: Why not use `lib.write_data`?

pygmt/tests/test_clib.py
```diff
@@ -399,6 +402,47 @@ def test_virtualfile_from_vectors():
...
+    condition=gmt_version < Version("6.1.1"),
+    reason="GMT_Put_Strings only works for GMT 6.1.1 and above",
+)
+def test_virtual_from_vectors_two_string_columns():
+    "Test passing in two columns of string dtype into virtual file dataset"
+    size = 5
+    x = np.arange(size, dtype=np.int32)
+    y = np.arange(size, size * 2, 1, dtype=np.int32)
+    strings1 = np.array(["a", "b", "c", "d", "e"], dtype=np.str)
+    strings2 = np.array(["f", "g", "h", "i", "j"], dtype=np.str)
```
seisman, line 435: Again, a string array with variable lengths.

pygmt/tests/test_clib.py
weiji14: Because it complicates the test. But yes, I could do that.

weiji14: Actually, it's hard because unlike in `test_clib_put_strings`, we don't use `dataset = lib.create_data(...)`, so there's not 'dataset' for `lib.write_data(...)` to write to.

seisman: OK, I see. I think you can call `gmt select` instead. If you don't give any selection criteria, it should output the original input data (If I remember it correctly, file header and segment header may be discarded, but it doesn't affect the use case here). However, due to a possible bug/limitation of **select** (see https://github.com/GenericMappingTools/gmt/issues/3907), you have to give at least one option. **gmt select -Vw** should work for the test.

PaulWessel:
gmt select is meant to take an option for action though.  It was never mean to just pass things through.

pygmt/tests/test_clib.py
weiji14: Cool, I knew there was a `gmt` command for this! Using `gmt select -Vw` works, and I've changed the test so it should be all good now.

seisman: @weiji14 See the feedback from https://github.com/GenericMappingTools/gmt/issues/3907. **gmt convert** is a better choice, although **gmt select** also works.

pygmt/tests/test_clib_put_strings.py
weiji14: Xfails have been removed in #577 at c0bb8be065870879128f3f2804c161746e0dc9f7.

PR #522, opened 2020-07-14
WIP: Update 6 baseline images due to GMT 6.1.0 changes

seisman:
This PR updates the baseline images of 6 failing tests and removes the `xfail` markers.

These tests fail due to slight changes between GMT 6.0 and GMT 6.1, and are not related to grid data.

**Reminders**


seisman:
OK. We can hold this PR for a while.

seisman:
> We should look into using https://matplotlib.org/3.2.2/api/testing_api.html#matplotlib.testing.decorators.check_figures_equal for the `grd*` changes, i.e. check that PyGMT is reproducing the images output by the **current** GMT version, otherwise this is a never-ending game.

I'm a little confused here. When I first read the link, I thought we can avoid unnecessary image updates by increasing the RMS threshold, but I don't see why `check_figures_equal` is preferred over `image_comparison`.

One more thing, we are using the [pytest-mpl](https://github.com/matplotlib/pytest-mpl) plugin, but the link above is for matplotlib.

weiji14:
> I'm a little confused here. When I first read the link, I thought we can avoid unnecessary image updates by increasing the RMS threshold, but I don't see why `check_figures_equal` is preferred over `image_comparison`.
> 
> One more thing, we are using the [pytest-mpl](https://github.com/matplotlib/pytest-mpl) plugin, but the link above is for matplotlib.

Sorry for the confusion. I wanted to point out how matplotlib states that

> "This [`check_figures_equal`] decorator should be preferred over `image_comparison` when possible in order to keep the size of the test suite from ballooning.".

This is a totally different way of testing, different to using `pytest-mpl`. I.e. Instead of testing against a *static* baseline image, I'm considering whether we should *just* check the differences between 1. a plot generated using `xarray.DataArray` grids and 2. `NetCDF file` grids. The risk here is that there's a bug which affects both `xarray.DataArray` and `NetCDF` grids, but currently all our problems stem from using `xarray.DataArray` grids. If we assume that `NetCDF` grids are always fine in GMT, then this would be safe to do.

seisman:
> check the differences between 1. a plot generated using `xarray.DataArray` grids and 2. `NetCDF file` grids.

It sounds a really good idea.

seisman:
> We should look into using https://matplotlib.org/3.2.2/api/testing_api.html#matplotlib.testing.decorators.check_figures_equal for the `grd*` changes, i.e. check that PyGMT is reproducing the images output by the **current** GMT version, otherwise this is a never-ending game.

Tried a little bit of the mpl decorator. It doesn't work for us, because it has some matplotlib-specific codes. But the idea is great. I think we need to implement our own way to compare two figures, like:

```python
fig1 = pygmt.Figure()
fig1.basemap(....)

fig2 = pygmt.Figure()
fig2.basemap(....)

pygmt.check_figure_equal(fig1, fig2, tol="0.01")
```

PR #527, opened 2020-07-15
Improve the cache step in GitHub Action

### LINKED ISSUE #523: The GitHub Actions caches don't work for PRs

See one of the [recent PR](https://github.com/GenericMappingTools/pygmt/pull/522/checks?check_run_id=867502859) for an example.

The caches are correctly downloaded, but the "cache" step still runs:

image

seisman:
Use one single `wget` command to download all cache files.

1. Create a list of files to download
2. Using one single wget command to download all files to cachedir
3. Move the downloaded files to the correct directories, overriding existing old files.

Fixes #523.

**Reminders**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -91,7 +91,7 @@ jobs:
           path: |
             ~/.gmt/cache
             ~/.gmt/server
-          key: cache-gmt-${{ runner.os }}-${{ github.ref }}-20200711
+          key: cache-gmt-${{ runner.os }}-${{ github.ref }}-20200715
           restore-keys: cache-gmt-${{ runner.os }}-refs/heads/master-
 
       # Workaround for the timeouts of 'gmt which' on Linux and Windows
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yaml
```diff
@@ -91,22 +91,42 @@ jobs:
...
+            echo "${GMT_DATA_SERVER}/server/earth/earth_relief/$file" >> $wget_input
           done
-          for data in ridge.txt Table_5_11.txt test.dat.nc tut_bathy.nc tut_quakes.ngdc tut_ship.xyz usgs_quakes_22.txt; do
-            wget --no-check-certificate https://oceania.generic-mapping-tools.org/cache/${data} -P ~/.gmt/cache/
+          for file in $cache; do
+            echo "${GMT_DATA_SERVER}/cache/$file" >> $wget_input
           done
+
+          mkdir cachedir
+          wget --no-check-certificate -i $wget_input -P cachedir
```
weiji14, line 117: It would be nice to just put the `wget_input.txt` file somewhere (e.g. in `.github/workflows`) instead of making one from scratch. But I see that we need the `$data` and `$cache` variables.

.github/workflows/ci_tests.yaml
```diff
@@ -91,22 +91,42 @@ jobs:
...
+          cd cachedir
+          mv gmt_hash_server.txt gmt_data_server.txt ~/.gmt/server/
+          for file in $data; do
+            mv $file ~/.gmt/server/earth/earth_relief/
+          done
+          for file in $cache; do
+            mv $file ~/.gmt/cache/
+          done
+          cd ..
+          rm -r cachedir $wget_input
```
weiji14, line 129: I know this works, but it seems like a lot of code to handle something that ought to be simple!

PR #528, opened 2020-07-15
Redesign the display mechanism by Leo

seisman:
**Description of proposed changes**

In #269, @leouieda proposed a new design of the PyGMT display mechanism. The core idea of #269 is to have a uniform `Figure.show()` function to open images using external viewer and show images in Jupyter notebook.

This is a draft implementation by @leouieda. It was done a year ago, but never finished. This is how the `Figure.show()` function in this PR behaves (the behaviors are controlled by the variable `SHOW_CONFIG`):

- By default, `Figure.show()` open images using an external viewer and don't show the image in notebooks
- Users have to call `pygmt.enable_notebook()` to disable external viewer and enable inline images in notebooks
- Setting environmental variable `PYGMT_EXTERNAL_VIEWER` to `false` can also disable external viewer, useful when building documentations.

Recently, I'm also trying a new design (see PR #529). To me, my new design is better and more reasonable. Here, I just open this PR for more comparisons and discussions.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #529, opened 2020-07-15
Improve Figure.show for displaying previews in Jupyter notebooks and external viewers

### LINKED ISSUE #269: Rethink the display mechanisms

Currently, there many ways of displaying a PyGMT figure:

* Calling `Figure.show()` in the notebook returns an IPython PNG display object for the jupyter notebook
* Putting the `Figure` object at the end of a notebook cell displays the figure as PNG
* Calling `Figure.show(external=True)` will open a preview in an external PDF viewer or the browser. The external viewer is non-blocking so this doesn't work for a script.
* Calling `Figure.savefig(..., show=True)` will save the figure and open it on an external viewer

This is not ideal. It would be better to have:

* Calling `Figure.show()` opens a preview of the figure and halts the Python process until the viewer is closed. It can also return `self` so that if it's called at the end of a notebook cell it would get displayed on the notebook as well.
* Putting the `Figure` at the end of a notebook cell displays a PNG preview.
* Having some way of tell `Figure.show()` to not open the viewer. This is required when using a notebook or when building the docs (we don't want windows opening when we build the gallery or tutorials). This can be done through a function `pygmt.enable_notebook` which would set a flag to tell `show` not to do anything and/or through an environment variable `PYGMT_DISPLAY` which can be set to `none` or other options that the user might want. Ideally, we should have both. The env variable will be useful when building the docs.

### LINKED ISSUE #366: pygmt does not display images using show()

seisman:
### Problem

The problem was first raised by @leouieda in issue #269. In summary, there are too many inconsistent ways for displaying PyGMT images:

- `Figure.show()` inserts a PNG image in a Jupyter notebook
- `Figure.show(method="external")` opens a preview in an external PDF viewer or the browser.
- Putting the Figure object at the end of a notebook cell displays the figure as PNG
- `Figure.savefig(..., show=True)` save the figure and open it on an external viewer

The main problem is, `Figure.show()` only works in Jupyter notebooks, and when running PyGMT in scripts, we have to use `Figure.show(method="external")` (see a related issue #366). In the PyGMT documentation, we use `Figure.show()` in the tutorials and gallery examples, and these examples won't work if users copy and save the codes in Python files. It definitely cause a lot of confusion, that's why we have to add some notes at the beginning of each tutorial (#806). 

### The Design in PR #528

In #269, @leouieda also proposed a new design of the PyGMT display mechanism. 

> - Calling Figure.show() opens a preview of the figure and halts the Python process until the viewer is closed. It can also return self so that if it's called at the end of a notebook cell it would get displayed on the notebook as well.
> - Putting the Figure at the end of a notebook cell displays a PNG preview.
> - Having some way of tell Figure.show() to not open the viewer. This is required when using a notebook or when building the docs (we don't want windows opening when we build the gallery or tutorials). This can be done through a function pygmt.enable_notebook which would set a flag to tell show not to do anything and/or through an environment variable PYGMT_DISPLAY which can be set to none or other options that the user might want. Ideally, we should have both. The env variable will be useful when building the docs.
> 

The core idea of the proposal in #269 is to have a uniform `Figure.show()` function to open images using external viewers or insert images in Jupyter notebooks. #528 is a draft implementation by @leouieda. It was started a year ago but never finished. 

### The New Design in this PR

This PR implements a new design for the display mechanism. The new design follows the discussions in #269, but is slightly different from #269 and #528 in details.

In the new design of this PR, PyGMT tries to detect the current running environment (is it in a normal terminal or a Jupyter notebook), and show images in a Jupyter notebook if possible. Otherwise, it opens images using external viewers. So, `Figure.show()` will work in both Jupyter notebooks and Python scripts in a consistent way.

Here are the default behaviors for some common use cases:

| Cases  | Open externally?  | Inline display?   |   Notes |
|---|---|---|---|
| Run a script by `python display.py`   | YES | NO  | note1  | 
| Run a script by `ipython display.py`. | YES | NO  |   |   
| Run codes in a Python terminal interactively  |  YES | NO   |   |   
| Run codes in an IPython terminal interactively  |  YES | NO   |   |   
| Run codes in `jupyter qtconsole` | NO | YES | |
| Run codes in a notebook, i.e., `jupyter notebook` or `jupyter lab` | NO | YES | |
| Run codes in `jupyter console` | YES | NO | note2 |

- note1: It's non-blocking. If the script calls `Figure.show()` multiple times, ideally, it should display the 1st figure, halts the Python process until the viewer is closed, and then display the 2nd figure. As I understand it, it's impossible.
- note2: jupyter console generates a PNG file, and opens it using an external viewer.

When building documentation, it's also possible to disable external display using any one of the two ways:

1. `export PYGMT_USE_EXTERNAL_DISPLAY=false`
2. Use `pygmt.set_display(method="notebook")`

### Summarized features in this PR

1. Consistent way (`Figure.show`) to preview images in both Jupyter notebooks and external viewers
2. Disable external viewers by setting `PYGMT_USE_EXTERNAL_DISPLAY` to `false` (useful for building documentation)
3. New function `pygmt.set_display(method=None, dpi=None, width=None)` to set the display method ("external" or "notebook"), and image *dpi* (default to 300) and *width* (default to 500) in Jupyter notebooks, **globally**. It affects all the subsequent figures.
4. ~~`Figure.show(method=None, dpi=None, width=None)` can set the display method ("external" or "notebook"), and image *dpi* and *width* in Jupyter notebooks, **locally**. It only affects the current figure.~~

### Testing this PR

As listed in the "TODO" section, I don't write add unit tests for the new feature yet. So we need your help to test this PR manually. Here I provide a series of steps you can follow to help test this PR.

**Testing the choice of notebook display or external viewer**
```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
fig.show()
```
1. Save the script into a Python file (e.g., `test1.py`), and run it using `python test1.py`, or `ipython test1.py`. It will open the image in an external viewer, but you may encounter the issue in #1061
2. Run the codes in a Python/IPython interpreter iteratively. It will open the image in an external viewer.
3. Run the codes in a Jupyter notebook (e.g., `ipython notebook`, `jupyter notebook` or `jupyter lab`). You should see an image inserted in the notebook.

**Testing the display in notebook**

In a Jupyter notebook, run:
```
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
fig
```
Putting the Figure instance `fig` in the last cell and run, it will also display an image, similar to `Figure.show()`

**Testing changing display method globally or locally**

Images will be inserted into the notebook by default. As mentioned above, there are two ways to change the display method:

- The **GLOBAL** way: `pygmt.set_display(method="xxx")`
- The **LOCAL** way: `Figure.show(method="xxx")`

The below script can be used to test if the two methods work as expected. You could run it in a Jupyter notebook, in a Python file, or in a Python/IPython interpreter. The comments in the script indicate the expected behavior in a Jupyter notebook:
```
import pygmt

# Figure 1 should be inserted in the notebook
fig1 = pygmt.Figure()
fig1.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
fig1.show()

pygmt.set_display(method="external")
# Figure 2 should be opened in an external viewer
fig2 = pygmt.Figure()
fig2.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure2")
fig2.show()

# Figure 3 should be opened in an external viewer
fig3 = pygmt.Figure()
fig3.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure3")
fig3.show()

pygmt.set_display(method="notebook")
# Figure 4 should be inserted in the notebook
fig4 = pygmt.Figure()
fig4.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure4")
fig4.show()

# Figure 5 should be inserted in the notebook
fig5 = pygmt.Figure()
fig5.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure5")
fig5.show()

# Figure 6 should be opened in an external viewer
fig6 = pygmt.Figure()
fig6.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure6")
fig6.show(method="external")
```

**Testing change dpi and width**

**SKIP THIS TEST! THE FEATURE IS TEMPORARILY REMOVED FROM THIS PR!**

```
import pygmt

# dpi=300
fig1 = pygmt.Figure()
fig1.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
fig1.show()
# still Figure 1, but dpi=10
fig1.show(dpi=10)

pygmt.set_display(dpi=50, width=200)
# dpi=50
fig2 = pygmt.Figure()
fig2.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure2")
fig2.show()

# dpi=50
fig3 = pygmt.Figure()
fig3.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure3")
fig3.show()

# dpi=100
fig4 = pygmt.Figure()
fig4.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure4")
fig4.show(dpi=100, width=200)

pygmt.set_display(dpi=300)
# dpi=300
fig5 = pygmt.Figure()
fig5.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure5")
fig5.show()

fig6 = pygmt.Figure()
# dpi=100
fig6.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure6")
fig6.show(dpi=100)
```

### TODO

- [ ] Add unit tests to make sure the new display mechanism works as expected
- [ ] Discuss about the default *dpi* (currently 300) and *width* (currently 500) in Jupyter notebooks
- [ ] More?

Fixes #269. Fixes #366.

**Reminders**


leouieda:
@seisman thanks for pushing this forward! It would be good to have a manual method of setting the display since there is no guarantee that the notebook detection always works (and this is by design of jupyter). How about merging the two functions you have into `pygmt.set_display("notebook")` and `pygmt.set_display("external")`?

weiji14:
cc @MarkWieczorek who made some comments in #366 (and this PR should ideally close it). Also might want to check out how `tqdm` does it at https://github.com/tqdm/tqdm#ipythonjupyter-integration, not sure if I like their `import` style but it might give you some ideas.

I'd also like to second @leouieda's comments about merging the two functions, since what we need is essentially a boolean True/False toggle (similar to the xarray accessor at #500).

weiji14 (commented):
Sorry for letting this sit for so long, but I think we should try to wrap this one up soon (the constant merge notifications are getting annoying). I think the code could be simplified with some more thought, and definitely need to add some more unit tests. Will try to help out if I can.

pygmt/figure.py
```diff
@@ -26,6 +22,32 @@
 # This is needed for the sphinx-gallery scraper in pygmt/sphinx_gallery.py
 SHOWED_FIGURES = []
 
+# Configuration options for Jupyter notebook support
+SHOW_CONFIG = {
+    "external": True,  # Open in an external viewer
+    "notebook": True,  # Notebook display
+    "dpi": 200,  # default DPI
+}
```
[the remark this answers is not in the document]
weiji14: Note that `xarray` has a `set_options` API (see http://xarray.pydata.org/en/v0.16.2/generated/xarray.set_options.html and code at https://github.com/pydata/xarray/blob/v0.16.2/xarray/core/options.py) that we could mimic.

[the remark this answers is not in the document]
seisman: > Is there a way to avoid the use of a global `SHOW_CONFIG` variable? How about setting these variables in the `__init__` of the `pygmt.Figure` class as `self.is_external` and `self.is_notebook` (or is there some reason I'm missing)?

I think moving these settings into `__init__` means that we have to re-detect the environment for each `Figure()` instance.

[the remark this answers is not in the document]
weiji14: > > Is there a way to avoid the use of a global `SHOW_CONFIG` variable? How about setting these variables in the `__init__` of the `pygmt.Figure` class as `self.is_external` and `self.is_notebook` (or is there some reason I'm missing)?
> 
> I think moving these settings into `__init__` means that we have to re-detect the environment for each `Figure()` instance.

Right, let's not put it in `__init__.py` then. But should we keep this in `figure.py`? Or do we want to consider moving some of this logic into an `options.py` file, i.e. in line with my comment at https://github.com/GenericMappingTools/pygmt/pull/529/files#r569876074 on adapting `xarray`'s `set_options` API?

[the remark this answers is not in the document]
seisman: This PR provides a `pygmt.set_display()` function. Now it can select the display method ("notebook" or "external"), but it's easy to allow it to set other configurations (e.g., *width*, *dpi*) for "notebook" display.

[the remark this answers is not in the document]
weiji14: > This PR provides a `pygmt.set_display()` function. Now it can select the display method ("notebook" or "external")

Ok with this.

> but it's easy to allow it to set other configurations (e.g., _width_, _dpi_) for "notebook" display.

Can we postpone the `dpi` and `width` configuration setting in a separate PR (i.e. keep this PR solely focused on automatic notebook or external display)? More to keep the scope of this PR small, but also because the `dpi` setting seems like a parallel issue outside of resolving #269 (i.e. needs more discussion on what a suitable default `dpi` is).

pygmt/figure.py
```diff
@@ -235,62 +262,37 @@ def savefig(
         if show:
             launch_external_viewer(fname)
 
-    def show(self, dpi=300, width=500, method="static"):
+    def show(self):
```
weiji14, line 265:
```suggestion
    def show(self, dpi=300, width=500, method=None):
```
As I just touched on in https://github.com/GenericMappingTools/pygmt/pull/529/files#r578027757 (and also https://github.com/GenericMappingTools/pygmt/pull/529/files#r560667857), maybe keep the dpi/width setting the same for now in this PR, and focus on changing the display `method` to be automated [the suggestion here is to use `display=None` (auto), but this can be overridden].

seisman: Yes, it sounds a good suggestion.

weiji14: Sorry if this is a lot of work, I do realize you've mostly just adapted Leo's implementation in #529 which changed the `dpi` setting :slightly_smiling_face:

seisman: Have to re-learn what I did half-year ago 😃

seisman:
@GenericMappingTools/python @GenericMappingTools/python-contributors @GenericMappingTools/python-maintainers @MarkWieczorek 

This PR changes the way PyGMT displays images in a "**backward-incompatible**" way. It fixes our two oldest issues #269 and #366. You can find more details by reading the long PR description.

Please help test the new design in different environments, give your comments and suggestions, and also try hard to break it and help improve it.

pygmt/figure.py
```diff
@@ -235,62 +255,79 @@ def savefig(
         if show:
             launch_external_viewer(fname)
 
-    def show(self, dpi=300, width=500, method="static"):
+    def show(self, method=None, dpi=None, width=None):
```
weiji14, line 258: Maybe I should clarify what I meant at https://github.com/GenericMappingTools/pygmt/pull/529#discussion_r578027757. I think that this PR should only change the `method` part to `method=None` (auto detect) and keep the `dpi` and `width` settings untouched.

```suggestion
    def show(self, dpi=300, width=500, method=None):
```

pygmt/figure.py
```diff
@@ -394,3 +433,33 @@ def _repr_html_(self):
         subplot,
         text,
     )
+
+
+def set_display(method=None, dpi=None, width=None):
```
weiji14, line 438:
```suggestion
def set_display(method=None):
```

Personally I don't see a use-case for setting a global `dpi` or `width` (but happy to be proven wrong). My suggestion is that we only let this `set_display` function set the method and not do anything about the dpi or width. Haven't looked too closely, but I think most of the changes were made in ca0ecb3d649bbf00b380e8848b3c82b9cd28a68c and could be partially reverted there. Not doing anything about the dpi/width would also greatly reduce the diff on this PR and make it easier to review. Discuss first though.

pygmt/figure.py
```diff
@@ -394,3 +433,33 @@ def _repr_html_(self):
...
+    dpi : int
+        Set the default DPI (dots-per-inch) used for PNG image previews that
+        are inserted into the notebook.
+    """
+    if method == "notebook":
+        SHOW_CONFIG["notebook"] = True
+        SHOW_CONFIG["external"] = False
+    elif method == "external":
+        SHOW_CONFIG["notebook"] = False
+        SHOW_CONFIG["external"] = True
```
weiji14, line 457: Rather than have two configurations, can we not just have a single boolean config object? Are there cases when we need to set both `SHOW_CONFIG["notebook"]` and `SHOW_CONFIG["external"]` to `True`, or both of them to `False`?

How about just having `SHOW_CONFIG["external"]`, defaulting to `False` on Jupyter, but `True` on regular *.py scripts.

weiji14 (commented):
Thanks @seisman, I still think we can simplify this Pull Request right down to be focused solely on handling either `display="external"` or `display="notebook"`. Setting the dpi/width seems like a nice to have and out of scope of the original issue at #269 (wasn't mentioned at all in fact). But happy to hear your thoughts on it.

pygmt/figure.py
seisman: I just reverted the changes about *dpi* and *width*.

seisman:
> I still think we can simplify this Pull Request right down to be focused solely on handling either `display="external"` or `display="notebook"`. Setting the dpi/width seems like a nice to have and out of scope of the original issue at #269 (wasn't mentioned at all in fact). 

Just made the changes to focus on the display method.

> (even better to just have a boolean `external=True/False` flag). 

PyGMT used `external=True/False` in the initial versions, and then decided to use `method="static/external"` when adding the support for NASA WorldWind Web (#132), although the WorldWind Web feature is removed later (#275).

I prefer to use `method="external"` so it's possible to add more display features in the future.

pygmt/figure.py
seisman: We can discuss the global *dpi* and *width* setting in a separate PR.

pygmt/figure.py
seisman: > Are there cases when we need to set both `SHOW_CONFIG["notebook"]` and `SHOW_CONFIG["external"]` to `True`.

Initially, I allowed setting both to `True`, and realized that it makes the codes more complicated, but I'm not sure if users need it.

> or both of them to `False`?

Yes, we need both of them to `False` when building documentation.

weiji14: Could we maybe have a single setting like `SHOW_CONFIG["display_method"]` (or another name) instead? That parameter could have either 'external', 'notebook' or 'sphinx' settings, and there won't be as many if-elif statements.

seisman: In commit 6e59fa3d2d625a407a974cd17663f8863767b59c, I changed it to `SHOW_CONFIG["method"]`, and possible values are "external", "notebook" and "none" (str type "none", but `None`).

pygmt/tests/test_figure.py
```diff
@@ -142,8 +142,7 @@ def test_figure_show():
     """
     fig = Figure()
     fig.basemap(region="10/70/-300/800", projection="X3i/5i", frame="af")
-    img = fig.show(width=800)
-    assert img.width == 800
+    fig.show()
```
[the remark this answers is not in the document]
seisman: I can't revert the change. The old `Figure.show()` returns the Image object, but the new `Figure.show()` doesn't.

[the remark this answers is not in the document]
weiji14: Ah I see. Would it be possible to have the Image object be returned for `method='notebook'` and `None` for `method='external'`? Otherwise this will be a bit of a backward incompatible break.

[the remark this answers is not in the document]
seisman: Yes, it's possible, but why do we want `Figure.show()` to return an Image object? What are the use cases?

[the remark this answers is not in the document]
weiji14: For plotting the Image in another Jupyter notebook cell? Or plotting it again and again without having to call `fig.show()` (which calls `psconvert`). The [IPython.display.Image](https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html) object also has other methods (e.g. `data` which gives raw image data) that could be used.

pygmt/figure.py
```diff
@@ -235,62 +253,79 @@ def savefig(
...
+        "notebook" or "external").
+
+        The ``method`` parameter can also override the default display method
+        for the current figure. Parameters ``dpi`` and ``width`` can be used
+        control the figures in the notebook.
+
+        The external viewer can be disabled by setting the
+        ``PYGMT_USE_EXTERNAL_DISPLAY`` environment variable to ``false``.
+        This is mainly used for running the tests and building the
+        documentation.
```
weiji14, line 274:
```suggestion
        Note: The external viewer can be disabled by setting the
        ``PYGMT_USE_EXTERNAL_DISPLAY`` environment variable to ``false``.
        This is useful for running the unit tests and building the
        documentation in consoles without a Graphical User Interface.
```

pygmt/figure.py
```diff
@@ -235,62 +253,79 @@ def savefig(
...
+        Inserts the preview in the Jupyter notebook output if available,
+        otherwise opens it in the default viewer for your operating system
+        (falls back to the default web browser).
+
+        :func:`pygmt.set_display` can select the default display method (either
+        "notebook" or "external").
+
+        The ``method`` parameter can also override the default display method
+        for the current figure. Parameters ``dpi`` and ``width`` can be used
+        control the figures in the notebook.
```
weiji14, line 269:
```suggestion
        for the current figure. Parameters ``dpi`` and ``width`` can be used
        to control the resolution and dimension of the figure in the notebook.
```

pygmt/figure.py
```diff
@@ -235,62 +253,79 @@ def savefig(
...
-            preview (default); (2) ``'external'``: PDF preview in an external
-            program.
+            How the current figure will be displayed. Options are
 
-        Returns
-        -------
-        img : IPython.display.Image
-            Only if ``method != 'external'``.
+            - ``'external'``: PDF preview in an external program [default]
+            - ``'notebook'``: PNG preview [default in Jupyter notebooks]
```
weiji14, line 291: Need to add back the 'Returns' section in the docstring.

doc/api/index.rst
```diff
@@ -58,6 +58,13 @@ Saving and displaying the figure:
     Figure.show
     Figure.psconvert
 
+Configuring the display settings:
+
+.. autosummary::
+    :toctree: generated
+
+    set_display
```
weiji14, line 66: I'm thinking whether to put `set_display` alongside `config` and have a heading like "Advanced configuration" or something like that.

weiji14 (commented):
This is coming along nicely, just some docs related comments this time.

meghanrjones:
I will try this out within the next couple days (I am even planning to set up a windows environment 😬)

seisman:
Please note that the current implemention is problematic. The following script only shows Figure 2 when running in one cell in Jupyter notebooks, because the return value of the code block is the Image object returned by `fig2.show()`:
```
import pygmt

fig1 = pygmt.Figure()
fig1.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
fig1.show()

fig2 = pygmt.Figure()
fig2.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure2")
fig2.show()
```
So I may have to revert some changes and call `IPython.display.display()` instead, but it may cause more problem if we call `IPython.display.display()` and return `IPython.display.Image()`. 

Need some further investigation.

weiji14:
> Please note that the current implemention is problematic. The following script only shows Figure 2 when running in one cell in Jupyter notebooks, because the return value of the code block is the Image object returned by `fig2.show()`:
> 
> ```
> import pygmt
> 
> fig1 = pygmt.Figure()
> fig1.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure1")
> fig1.show()
> 
> fig2 = pygmt.Figure()
> fig2.basemap(region=[0, 10, 0, 10], projection="X10c/10c", frame="WSrt+tFigure2")
> fig2.show()
> ```
> 
> So I may have to revert some changes and call `IPython.display.display()` instead, but it may cause more problem if we call `IPython.display.display()` and return `IPython.display.Image()`.
> 
> Edit: If we call `IPython.display.display()` and return `IPython.display.Image()`, and run the above script in one cell, we will see Figure 2 twice.

My knee-jerk thought is to just put those two code blocks in two different cells... Haven't used matplotlib in a while, but does the matplotlib `fig.show` method display two images when called that way?

seisman:
> Haven't used matplotlib in a while, but does the matplotlib `fig.show` method display two images when called that way?

`plt.show()` returns `None`, and the following script shows two images.
```

plt.plot([0, 1, 2], [2, 3, 4])
plt.show()

plt.plot([10, 11, 12], [12, 13, 14])
plt.show()
```

image

seisman:
> For plotting the Image in another Jupyter notebook cell? Or plotting it again and again without having to call fig.show() (which calls psconvert). The IPython.display.Image object also has other methods (e.g. data which gives raw image data) that could be used.

Letting `Figure.show()` return an image sounds useful, but IMHO it brings more troubles and behaves inconsistently in different use cases. I believe Leo did the same thing in his original branch/PR (#528) due to the same reasons.

I think we should **NOT** let `Figure.show()` return the Image object. 

If someone wants to reuse the image, they can still save the image to disk and re-read it into memory.

Or we could add new a property to the Figure.class, for example `Figure.raw_image` which save the raw image data of the last `Figure.show()` call.

seisman:
@weiji14 Any more comments on this PR?

weiji14:
> @weiji14 Any more comments on this PR?

I would prefer if someone else reviews this as well, e.g. @meghanrjones if she has time. I'm still somewhat on the fence about not returning the IPython.display.Image object as you recommend in https://github.com/GenericMappingTools/pygmt/pull/529#issuecomment-805508310, but will see what others say. Other than that, I have no major issues with the rest of the implementation in this PR.

meghanrjones:
> > @weiji14 Any more comments on this PR?
> 
> I would prefer if someone else reviews this as well, e.g. @meghanrjones if she has time. I'm still somewhat on the fence about not returning the IPython.display.Image object as you recommend in [#529 (comment)](https://github.com/GenericMappingTools/pygmt/pull/529#issuecomment-805508310), but will see what others say. Other than that, I have no major issues with the rest of the implementation in this PR.

Yes, it's hard to believe so many days have passed since I said that I would review 😕. Sorry for that - I will work on this today unless Paul requests any 6.2 tasking.

meghanrjones:
I am running into many problems with #1061, which is why I think it is necessary to merge #1062 first.

pygmt/figure.py
```diff
@@ -235,62 +253,73 @@ def savefig(
...
-        if method not in ["static", "external"]:
-            raise GMTInvalidInput("Invalid show method '{}'.".format(method))
-        if method == "external":
-            pdf = self._preview(fmt="pdf", dpi=dpi, anti_alias=False, as_bytes=False)
-            launch_external_viewer(pdf)
-            img = None
-        elif method == "static":
-            png = self._preview(
-                fmt="png", dpi=dpi, anti_alias=True, as_bytes=True, transparent=True
+        # Set the default display method
```
meghanrjones, line 296:
```suggestion
        # Set the display method
```

pygmt/figure.py
```diff
@@ -235,62 +253,73 @@ def savefig(
...
+        PYGMT_USE_EXTERNAL_DISPLAY environment variable to **false**.
+        This is useful when running unit tests and building the documentation
+        in consoles without a Graphical User Interface.
 
-        If ``method='external'``, makes PDF preview instead and opens it in the
-        default viewer for your operating system (falls back to the default web
-        browser). Note that the external viewer does not block the current
-        process, so this won't work in a script.
+        Note that the external viewer does not block the current process,
+        so this won't work if multiple figures are shown in one script.
```
meghanrjones, line 277: I'm a bit confused about this statement, since it seems to work now. Is this a relic of #1061 that needs updating?

meghanrjones (commented):
Do you think that there should be warnings if dpi or width are used with method=external or method=none?

seisman:
> Do you think that there should be warnings if dpi or width are used with method=external or method=none?

Sounds good to me. I will open a separate PR about dpi and width settings after this PR is merged.

pygmt/figure.py
seisman: > Note that the external viewer does not block the current process,

This statement is still true. It's still non-blocking. 

> so this won't work if multiple figures are shown in one script.

#1062 partially fixes #1061, so I think we can remove this sentence.

meghanrjones (commented):
This seems like a great improvement to me! It worked well in a few different environments on MacOS and Windows.

Regarding whether figure.show() should return an image, I think that could be a separate feature request and that this PR is a good baseline for any subsequent improvements before v0.4.0.

The TODO lists unit tests, is that still the plan or do you think this is good to go @seisman?

seisman:
> The TODO lists unit tests, is that still the plan or do you think this is good to go @seisman?

Actually, I have no idea how to test the new `Figure.show()` behavior automatically, because when running `make tests`, IPython is not imported so `Figure.show()` will do nothing.

meghanrjones (approved):
I think if you can add a test for invalid input to set_display that would be good. Otherwise this looks good to me. I also do not know a way to add more unit tests, but expect any breaking changes would be noticed very soon. Please wait for @weiji14's approval before merging.

weiji14 (approved):
Ok, let's get this in then! Will see if anyone else wants to comment, otherwise merge it tomorrow.

PR #530, opened 2020-07-16
Cache GMT remote data as artifacts on Github

### LINKED ISSUE #523: The GitHub Actions caches don't work for PRs

weiji14:

Use GMT on macOS to download the GMT remote files from the GMT data server, and then upload them as artifacts on Github.

Separates the 'cache' task from our Tests CI. Meaning we can update our cache independently of our Continuous Integration tests, allowing the cache to be updated every Sunday for example without needing to manually trigger it with a PR.

References:
- https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts
- https://github.com/actions/upload-artifact
- https://github.com/dawidd6/action-download-artifact
- https://github.com/actions/download-artifact
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule

Fixes #523, Closes #527

**Reminders**


.github/workflows/ci_tests.yaml
```diff
@@ -83,39 +83,14 @@ jobs:
...
-      - name: Download remote data (macOS)
-        shell: bash -l {0}
-        run: |
-          gmt which -Ga @earth_relief_10m_p @earth_relief_10m_g @earth_relief_30m_p @earth_relief_30m_g @earth_relief_01d_p @earth_relief_01d_g
-          gmt which -Ga @ridge.txt @Table_5_11.txt @test.dat.nc @tut_bathy.nc @tut_quakes.ngdc @tut_ship.xyz @usgs_quakes_22.txt
-        if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'macOS'
+          github_token: ${{secrets.GITHUB_TOKEN}}
+          workflow: cache_data.yaml
+          name: gmt-cache
+          path: ~/.gmt
```
seisman, line 93: Is the path correct? Downloading artifacts is fast, but it may not be placed in the correct directories. GMT still tries to download all files during testing.

weiji14: Hmm, it looks like it's unzipped to the right location (e.g. `~/.gmt/server/earth/earth_relief/earth_relief_01d_g.grd`), and that's what `wget` was doing before?? Maybe we should use `$HOME` or something?

seisman: I think you can add a step listing all files in the ~/.gmt directory after downloading the artifacts.

weiji14: Hmm, $HOME doesn't work either. Then I'm not sure where the action is actually unzipping the files to!

seisman: I've some tests to do, and will make changes to this branch.

weiji14: Ok, go for it :+1:

.github/workflows/ci_tests.yaml
```diff
@@ -83,39 +83,21 @@ jobs:
         shell: bash -l {0}
         run: conda list
 
-      # Cache the ${HOME}/.gmt directory, for docs and testing
-      - name: Cache GMT directory
-        uses: actions/cache@v2
-        id: cache
+      # Download cached remote files (artifacts) from Github
+      - name: Download remote data from Github
+        uses: dawidd6/action-download-artifact@v2
```
weiji14, line 88: I'll upgrade the action to v2.6.3, there's been some changes https://github.com/dawidd6/action-download-artifact/compare/v2.0.0...v2.6.3, and we can leave out setting the github_token.

.github/workflows/ci_tests.yaml
weiji14: I see that a `~` folder is created :facepalm:. How about trying `./.gmt`, since the action uses `./` as the default, see https://github.com/dawidd6/action-download-artifact/blob/2f29da2bb95f5580767170955fca0cf48714d48b/action.yml#L31. Also, we should use `ls -alh` to see hidden dot files.

seisman: I saw the `.gmt` directory in the root directory. At least we can download the artifact, and then move it to `~/`.

weiji14: Yes, that's one way of doing it.

.github/workflows/ci_tests.yaml
```diff
@@ -83,39 +83,20 @@ jobs:
...
+          path: .gmt
+
+      # Move downloaded files to ~/.gmt directory and list them
+      - name: Move and list downloaded remote files
         shell: bash -l {0}
         run: |
-          gmt which -Ga @earth_relief_10m_p @earth_relief_10m_g @earth_relief_30m_p @earth_relief_30m_g @earth_relief_01d_p @earth_relief_01d_g
-          gmt which -Ga @ridge.txt @Table_5_11.txt @test.dat.nc @tut_bathy.nc @tut_quakes.ngdc @tut_ship.xyz @usgs_quakes_22.txt
-        if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'macOS'
+          mv .gmt/* ~/.gmt
```
weiji14, line 98: Hmm, is the syntax for `mv` slightly different on macOS? It says:

```
usage: mv [-f | -i | -n] [-v] source target
       mv [-f | -i | -n] [-v] source ... directory
```

seisman: I think you need to create the directory first. You can't move files to a non-existing directory.

.github/workflows/cache_data.yaml
```diff
@@ -0,0 +1,43 @@
...
+  gmt_cache:
+    name: Cache GMT artifacts
+    runs-on: macOS-latest
+
+    steps:
+      # Setup Miniconda
+      - name: Setup Miniconda
+        uses: goanpeca/setup-miniconda@v1.6.0
+        with:
+          python-version: ${{ matrix.python-version }}
```
seisman, line 21:
```suggestion
          python-version: 3.7
```

weiji14: Actually don't need Python at all, so will remove the line.

PR #537, opened 2020-07-18
Raise an error if short- and long-form arguments coexist

### LINKED ISSUE #377: Short- and long-form kwargs can coexist silently

Python doesn't allow keyword arguments to be repeated. Our current setup _does_ allow for e.g. the following, however:
```python
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 1, 0, 1], B='+tbar', frame='+tfoo')
fig.show(method='external')
```
This example produces

ex

silently overriding the `B='+tbar'` "repeated" argument. I feel like we should raise a `SyntaxError` in this case, like what Python does for repeated kwargs. I think this would involve modifying the `use_alias()` decorator, but I'm not totally sure. This addresses the discussion in https://github.com/GenericMappingTools/pygmt/pull/359#discussion_r343539273. If #262 were implemented, this could be a non-issue.

(I'm not sure if this is a bug or a feature request; I chose bug.)

seisman:

Raise an error if both short- and long-form arguments are used. Also add a doctest for it.

See #377 for the bug report. 

Fixes #377

**Reminders**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -180,8 +180,9 @@ def use_alias(**aliases):
     R = bla J = meh
     >>> my_module(region='bla', projection='meh')
     R = bla J = meh
-    >>> my_module(region='bla', projection='meh', J="bla")
-    # doctest: +NORMALIZE_WHITESPACE
+    >>> my_module(
+    ...    region='bla', projection='meh', J="bla"
+    ... )  # doctest: +NORMALIZE_WHITESPACE
     Traceback (most recent call last):
       ...
     pygmt.exceptions.GMTInvalidInput:
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -181,9 +181,11 @@ def use_alias(**aliases):
     >>> my_module(region='bla', projection='meh')
     R = bla J = meh
     >>> my_module(region='bla', projection='meh', J="bla")
+    # doctest: +NORMALIZE_WHITESPACE
     Traceback (most recent call last):
       ...
-    pygmt.exceptions.GMTInvalidInput: Arguments in short-form (J) and long-form (projection) can't coexist
+    pygmt.exceptions.GMTInvalidInput:
+        Arguments in short-form (J) and long-form (projection) can't coexist
     """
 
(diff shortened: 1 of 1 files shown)


seisman:
@weiji14 Any idea to fix the style error?

pygmt/helpers/decorators.py
```diff
@@ -180,7 +180,10 @@ def use_alias(**aliases):
     R = bla J = meh
     >>> my_module(region='bla', projection='meh')
     R = bla J = meh
-
+    >>> my_module(region='bla', projection='meh', J="bla")
+    Traceback (most recent call last):
+      ...
+    pygmt.exceptions.GMTInvalidInput: Arguments in short-form (J) and long-form (projection) can't coexist
```
weiji14, line 186:
```suggestion
    >>> my_module(
    ...    region='bla', projection='meh', J="bla"
    ... )  # doctest: +NORMALIZE_WHITESPACE
    Traceback (most recent call last):
      ...
    pygmt.exceptions.GMTInvalidInput:
        Arguments in short-form (J) and long-form (projection) can't coexist
```

Try using [`# doctest +NORMALIZE_WHITESPACE`](https://docs.python.org/3.8/library/doctest.html#doctest.NORMALIZE_WHITESPACE). Just used it recently for the datetime doctest https://github.com/GenericMappingTools/pygmt/blob/9a4c0a0177dac03e05d219c01105f272c9911655/pygmt/clib/conversion.py#L287-L289

pygmt/helpers/decorators.py
```diff
@@ -180,8 +180,9 @@ def use_alias(**aliases):
     R = bla J = meh
     >>> my_module(region='bla', projection='meh')
     R = bla J = meh
-    >>> my_module(region='bla', projection='meh', J="bla")
-    # doctest: +NORMALIZE_WHITESPACE
+    >>> my_module(
+    ...    region='bla', projection='meh', J="bla"
+    ... )  # doctest: +NORMALIZE_WHITESPACE
```
seisman, line 185: Thanks for the change. My old commit worked for me locally, but failed in the CI. Perhaps I'm using a different pylint verison.

Issue #524, opened 2020-07-14
GMTDataArrayAccessor doesn't work for temporary files that have been sliced


seisman:
**Description of the problem**

It may be a GMTDataArrayAccessor bug, or maybe we should update `grdcut()` to work with GMTDataArrayAccessor. The problem is that the temporary netCDF file is already deleted when GMTDataArrayAccessor tries to access it.

**Full code that generated the error**

```python
import pygmt

grid = pygmt.grdcut("@earth_relief_01d", region=[30, 50, 60, 70])
print(grid)
print(grid.gmt.registration)
```

**Full error message**

```
<xarray.DataArray 'z' (lat: 10, lon: 20)>
array([[  46. ,   -2.5,    0. ,   42. ,  150. ,  220. ,  172. ,  147. ,
         153.5,  159. ,  177. ,  172. ,  165. ,  137. ,  144.5,  150. ,
         119. ,  142. ,  131.5,  149.5],
       [  18.5,   19. ,   80.5,  139. ,  120.5,   33. ,   54. ,  202.5,
         153. ,  126. ,  175. ,  165.5,   69. ,  180. ,  147. ,  147. ,
          61. ,  106. ,  112.5,  147. ],
       [ 125.5,  173. ,  185. ,  136.5,   70.5,   51.5,  158.5,  140. ,
         142. ,   97. ,  139. ,  133. ,   58.5,   53.5,  109. ,  171. ,
         158. ,  153. ,  152. ,  156. ],
       [ 173.5,  223.5,  199.5,  130.5,   95. ,  130.5,  170.5,  114. ,
          51. ,   80. ,   78.5,   30.5,   82. ,  129. ,  137. ,  126. ,
         139. ,  118. ,  138.5,  148. ],
       [ 216.5,  163. ,  133. ,  121. ,   51.5,   -7. ,  -18. ,   65. ,
          96. ,   34.5,   19.5,   58.5,   78. ,   42. ,   69.5,   96.5,
          82. ,  155. ,  132.5,  192.5],
       [ 187. ,  163.5,  123.5,   74.5,   12. ,  -50. , -177.5, -126. ,
        -106.5,  -37. ,   94. ,  101.5,  121.5,   28. ,   38. ,   64.5,
          67. ,   89. ,  118.5,  154. ],
       [ 169. ,  137. ,   63.5,  -23. , -100. ,   33. ,   85. ,  151. ,
[...]

PR #542, opened 2020-07-22
Let load_earth_relief() support all resolutions and optional subregion

### LINKED ISSUE #489: Redesign of the load_earth_relief() function

seisman:

See #489 for the reason of the function redesign. In summary, Earth relief data >05m are stored as single grids, and we can use the `which` function to get its full path and load it as DataArray. It doesn't work for earth relief data <=05m, which are split into smaller tiles. As suggested in #489, it's easier to use `grdcut @earth_relief_xxy -Rregion -Goutput.grd`. The `grdcut` function works for all resolutions, but due to the issue #524, the DataArray returned doesn't support slice operations. 

Changes in this PR:

- For grid>05m, use the `which` function. The returned DataArray still supports slice operations, so all tests pass
- For grids<=05m, use the `grdcut` function. The returned DataArray doesn't support slice operation (see #524). The bug is also documented.
- For grids<=05m, users have to pass the `region` argument to select a subregion.
- Remove the unused function `_shape_from_resolution`
- Use two lists `tiled_resolutions` and `non_tiled_resolutions`, instead of the old `_is_valid_resolution` function
- Remove `01s` and `03s` from the test, because these two resolutions are valid
- Add a few examples 
- Add tests for high resolution grids

Fixes #489.

**Reminders**

### DIFF
--- a/pygmt/datasets/earth_relief.py
+++ b/pygmt/datasets/earth_relief.py
@@ -97,7 +97,7 @@ def load_earth_relief(resolution="01d", region=None, registration=None):
         # See https://github.com/GenericMappingTools/pygmt/issues/524
         if region is None:
             raise GMTInvalidInput(
-                f"region is required for Earth relief grid with resolutions>=05m"
+                "region is required for high resolution (<='05m') Earth relief grid"
             )
         grid = grdcut(f"@earth_relief_{resolution}{reg}", region=region)
     else:
--- a/pygmt/datasets/earth_relief.py
+++ b/pygmt/datasets/earth_relief.py
@@ -1,24 +1,27 @@
 """
-Functions to download the Earth relief datasets from the GMT data server.
-The grids are available in various resolutions.
+Function to download the Earth relief datasets from the GMT data server,
+and load as DataArray. The grids are available in various resolutions.
 """
 import xarray as xr
 
-from .. import which
+from .. import grdcut, which
 from ..exceptions import GMTInvalidInput
+from ..helpers import kwargs_to_strings
 
 
-def load_earth_relief(resolution="01d", registration=None):
+@kwargs_to_strings(region="sequence")
+def load_earth_relief(resolution="01d", region="-180/180/-90/90", registration=None):
     """
     Load Earth relief grids (topography and bathymetry) in various resolutions.
 
-    The grids are downloaded to a user data directory (usually ``~/.gmt/``) the
-    first time you invoke this function. Afterwards, it will load the data from
-    the cache. So you'll need an internet connection the first time around.
+    The grids are downloaded to a user data directory
+    (usually ``~/.gmt/server/earth/earth_relief/``) the first time you invoke
+    this function. Afterwards, it will load the grid from the data directory.
+    So you'll need an internet connection the first time around.
 
     These grids can also be accessed by passing in the file name
-    ``'@earth_relief_XXm'`` or ``'@earth_relief_XXs'`` to any grid
-    plotting/processing function.
+    ``'@earth_relief_rru[_reg]'`` to any grid plotting/processing function.
+    Refer to :gmt-docs:`datasets/remote-data.html` for more details.
 
     Parameters
     ----------
@@ -26,7 +29,11 @@ def load_earth_relief(resolution="01d", registration=None):
         The grid resolution. The suffix ``d``, ``m`` and ``s`` stand for
         arc-degree, arc-minute and arc-second. It can be ``'01d'``, ``'30m'``,
         ``'20m'``, ``'15m'``, ``'10m'``, ``'06m'``, ``'05m'``, ``'04m'``,
-        ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'`` or ``'15s'``.
+        ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'``, ``'15s'``, ``'03s'``,
+        or ``'01s'``.
+
+    region : str or list
+        The subregion of the grid to load.
 
     registration : str
         Grid registration type. Either ``pixel`` for pixel registration or
@@ -40,8 +47,33 @@ def load_earth_relief(resolution="01d", registration=None):
         The Earth relief grid. Coordinates are latitude and longitude in
         degrees. Relief is in meters.
 
+    Notes
+    -----
+    The DataArray doesn's support slice operation, for Earth relief data with
+    resolutions higher than "05m", which are stored as smaller tiles.
     """
-    _is_valid_resolution(resolution)
+
+    # earth relief data stored as single grids for low resolutions
+    non_tiled_resolutions = [
+        "01d",
+        "30m",
+        "20m",
+        "15m",
+        "10m",
+        "06m",
+    ]
+    # earth relief data stored as tiles for high resolutions
+    tiled_resolutions = [
+        "05m",
+        "04m",
+        "03m",
+        "02m",
+        "01m",
+        "30s",
+        "15s",
+        "03s",
+        "01s",
+    ]
 
     if registration in ("pixel", "gridline", None):
         # If None, let GMT decide on Pixel/Gridline type
@@ -54,8 +86,19 @@ def load_earth_relief(resolution="01d", registration=None):
             "gridline-registered grid is available."
         )
 
-    fname = which(f"@earth_relief_{resolution}{reg}", download="a")
-    grid = xr.open_dataarray(fname)
+    # different ways to load tiled and non-tiled earth relief data
+    if resolution in non_tiled_resolutions:
+        fname = which(f"@earth_relief_{resolution}{reg}", download="a")
+        with xr.open_dataarray(fname) as dataarray:
+            grid = dataarray.load()
+            _ = grid.gmt  # load GMTDataArray accessor information
+    elif resolution in tiled_resolutions:
+        # Titled grid can't be sliced.
+        # See https://github.com/GenericMappingTools/pygmt/issues/524
+        grid = grdcut(f"@earth_relief_{resolution}{reg}", region=region)
+    else:
+        raise GMTInvalidInput(f"Invalid Earth relief resolution '{resolution}'")
+
     # Add some metadata to the grid
     grid.name = "elevation"
     grid.attrs["long_name"] = "elevation relative to the geoid"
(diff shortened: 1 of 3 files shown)


seisman:
@weiji14 I think the PR is ready for review, but the `grdcut` call is not tested, because it means we have to cache at least a subset of high-resolution grids.

pygmt/datasets/earth_relief.py
```diff
@@ -54,8 +99,23 @@ def load_earth_relief(resolution="01d", registration=None):
...
+    elif resolution in tiled_resolutions:
+        # Titled grid can't be sliced.
+        # See https://github.com/GenericMappingTools/pygmt/issues/524
+        if region is None:
+            raise GMTInvalidInput(
+                "region is required for high resolution (<='05m') Earth relief grid"
+            )
+        grid = grdcut(f"@earth_relief_{resolution}{reg}", region=region)
+    else:
+        raise GMTInvalidInput(f"Invalid Earth relief resolution '{resolution}'")
```
weiji14, line 117: >     2. There is a styling error which I don't understand.

So it's a `flake8` error saying: `pygmt/datasets/earth_relief.py:100:17: F541 f-string is missing placeholders`. It doesn't error for me locally, could you run `make check` though? The tiled_resolutions list will turn into a single line, and that line 100 is this one. If I'm reading https://stackoverflow.com/questions/62565979/github-action-flake8-fails-f-string-is-missing-placeholders correctly, you'll need to change it to this:
```suggestion
        raise GMTInvalidInput(f'Invalid Earth relief resolution "{resolution}"')
```

pygmt/datasets/earth_relief.py
```diff
@@ -40,8 +47,46 @@ def load_earth_relief(resolution="01d", registration=None):
...
+    --------
+
+    # load the default grid (1 arc degree resolution and pixel registration)
+    >>> grid = load_earth_relief()
+
+    # load the 30m grid with "gridline" registration
+    >>> grid = load_earth_relief("30m", registration="gridline")
+
+    # load high-resolution grid for a specific region
+    >>> grid = load_earth_relief("01m", region=[120, 140, 30, 50])
```
weiji14, line 65:
```suggestion
    >>> # load default grid (1 arc degree resolution and pixel registration)
    >>> grid = load_earth_relief()

    >>> # load the 30m grid with "gridline" registration
    >>> grid = load_earth_relief("30m", registration="gridline")

    >>> # load high-resolution grid for a specific region
    >>> grid = load_earth_relief("01m", region=[120, 140, 30, 50])
```

seisman: Yes, it works.

pygmt/datasets/earth_relief.py
```diff
@@ -1,32 +1,39 @@
...
     resolution : str
         The grid resolution. The suffix ``d``, ``m`` and ``s`` stand for
         arc-degree, arc-minute and arc-second. It can be ``'01d'``, ``'30m'``,
         ``'20m'``, ``'15m'``, ``'10m'``, ``'06m'``, ``'05m'``, ``'04m'``,
-        ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'`` or ``'15s'``.
+        ``'03m'``, ``'02m'``, ``'01m'``, ``'30s'``, ``'15s'``, ``'03s'``,
+        or ``'01s'``.
+
+    region : str or list
+        The subregion of the grid to load. Only works
```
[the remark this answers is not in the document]
seisman: Yes, fixed.

pygmt/datasets/earth_relief.py
```diff
@@ -40,8 +48,46 @@ def load_earth_relief(resolution="01d", registration=None):
...
+        "05m",
+        "04m",
+        "03m",
+        "02m",
+        "01m",
+        "30s",
+        "15s",
+        "03s",
+        "01s",
+    ]
```
weiji14, line 90: When I run `black` on these two lists, they flatten into a single line. The "Style Checks" pass though, so I find it strange.

seisman: It's automatically formatted by `black` locally, and I have black 19.10b0.

weiji14:
```suggestion
    # earth relief data stored as single grids for low resolutions
    non_tiled_resolutions = ["01d", "30m", "20m", "15m", "10m", "06m"]
    # earth relief data stored as tiles for high resolutions
    tiled_resolutions = ["05m", "04m", "03m", "02m", "01m", "30s", "15s", "03s", "01s"]
```

Well, `non_tiled_resolution` is definitely under 88 characters, `tiled_resolution` is at exactly 88 characters :confused:

seisman: Your changes also can pass the black test. 

What's your black version?

weiji14: My editor uses the system installed one which is `19.3b0`, but I have `19.10b0` installed too.

pygmt/datasets/earth_relief.py
```diff
@@ -54,8 +83,23 @@ def load_earth_relief(resolution="01d", registration=None):
...
+            grid = dataarray.load()
+            _ = grid.gmt  # load GMTDataArray accessor information
+    elif resolution in tiled_resolutions:
+        # Titled grid can't be sliced.
+        # See https://github.com/GenericMappingTools/pygmt/issues/524
+        if region is None:
+            raise GMTInvalidInput(
+                "region is required for high resolution (<='05m') Earth relief grid"
+            )
+        grid = grdcut(f"@earth_relief_{resolution}{reg}", region=region)
```
weiji14, line 99: Sorry, one more thing. What if a user wants to get a non_tiled_resolution grid **with** a **region**? The current code doesn't seem to handle this case.

seisman: The initial design was to allow users to specify a subregion for all resolutions. However, due to #524, the returned grids can't support slice operation, which makes three or more tests fail. So I have to do different ways for tiled and non-tiled resolutions, to keep all tests pass.

seisman: If a user really want it, they need to call `load_earth_relief`, then `grdcut`, or using xarray slice operation.

I'm also OK to add "region" to all resolutions, and declare that "load_earth_relief() grids don't support slice operation" as a bug.

weiji14: Right, that's tricky. But, an unsuspecting user might pass in a `region` argument for resolutions `06m` and above, and it would just be silently ignored (and a full grid is returned)! We should at least raise a warning or `NotImplementedError` for this case.

weiji14: Sorry, slow reply. A workaround might be to just set the `registration and gtype` directly, since we know it from the user input to `load_earth_relief`.

seisman: > A workaround might be to just set the `registration and gtype` directly after calling `grdcut`, since we know it from the user input to `load_earth_relief`.

Yes, it may make codes a little complicated. If `registration=None`, then the [default earth relief grid registration](https://docs.generic-mapping-tools.org/dev/datasets/remote-data.html#id1) is `pixel` except '03s' and '01s'.

seisman: If I remember it correctly, I can't just do
```
grid = grdcut("@earth_relief_01d_p", region="0/100/0/30")
grid.gmt.registration = 1
grid.gmt.gtype = 1
```
because `grid.gmt.registration = 1` will run `grdinfo` first, then override the value by `1`. Running `grdinfo` fails because the temporary grid is gone.

weiji14: Actually you can just get it from the `fname`, should either be a `_p` or `_g` at the end. `gtype` would always be Geographic so that's easy. But to be honest, we should probably fix #524 properly.

weiji14: Ah right, then this workaround won't work. So we'll definitely have to fix #524 first.

seisman: I don't see an easy fix for #524. Perhaps the best compromise is 
> raise a warning or NotImplementedError for this case.

weiji14: Ok, let's compromise on this for now. I don't like the fact that `05m`(and higher resolution) grids which are `grdcut` won't work with `grdimage` or `grdview`, but we can fix this before doing a proper release.

weiji14: Actually, if we remove/pop the tempfile from `grid.encoding["source"]` **after** running `grdcut`, we can then **set** the registration and gtype without getting an error. Do you think it's wise to apply this hack?

seisman: > that `05m`(and higher resolution) grids which are `grdcut` won't work with `grdimage` or `grdview`, but we can fix this before doing a proper release.

I don't understand it. Why do you say it doesn't work?

weiji14: > I don't understand it. Why do you say it doesn't work?

Sorry, it will work actually, I forgot for a second that we applied the workaround in #539.

seisman: > Actually, if we remove/pop the temp.nc file from `grid.encoding["source"]` **after** running `grdcut`, we can then **set** the registration and gtype without getting an error. Do you think it's wise to apply this hack?

You mean applying the hack only in `load_earth_relief()` function? Sounds a good idea.

seisman: Implemented locally, but I'm getting this error:
```python
_________________________________________ [doctest] pygmt.clib.conversion.dataarray_to_matrix _________________________________________
060     >>> matrix.flags.c_contiguous
061     True
062     >>> # Using a slice of the grid, the matrix will be copied to guarantee
063     >>> # that it's C-contiguous in memory. The increment should be unchanged.
064     >>> matrix, region, inc = dataarray_to_matrix(grid[10:41,30:101])
065     >>> matrix.flags.c_contiguous
066     True
067     >>> print(matrix.shape)
068     (31, 71)
069     >>> print(region)
Expected:
    [-150.0, -79.0, -80.0, -49.0]
Got:
    [-149.5, -79.5, -79.5, -49.5]
```

weiji14: > I don't see an easy fix for #524. Perhaps the best compromise is
> 
> > raise a warning or NotImplementedError for this case.

RIght, let's just raise a warning here then, having layers and layers of workarounds isn't good anyway.

PR #547, opened 2020-07-25
Add a test to check passing arrays of mixed dtypes to GMT

### LINKED ISSUE #255: Figure.plot fails to plot lines correctly

Figure.plot() doesn't plot simple lines correctly.

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(J="X8c/8c", R="-10/10/0/10", B=True)
fig.plot(x=[0.0, 0.0], y=[0, 10], W="1p,gray,-")
fig.plot(x=[5.0, 5.0], y=[2, 8], W="1p,gray,-")
fig.plot(x=[3.0, 8.0], y=[0, 10], W="1p,gray,-")
fig.show(method="external")
```

**Output plot**
plot-lines-fail

Only the first line is correct. The other two lines are wrong.

**System information**

* Operating system: macOS
* Version of GMT: 6.0.0_15a9160
* Version of Python: 3.7.0
* Version of this package: latest
* If using conda, paste the output of `conda list` below

seisman:

This PR contains two commits:

- Split tests test_clib_put_vector* and test_clib_put_matrix* into two separate files (aff04b019cbdfc4e392c94f7c38f30d8f9e6fb73) 
- Add a test to check passing arrays of mixed dtypes to GMT (de73738b46a0d61897a90f980f250a662fa76fac, closes #255.)

The bug was first reported in #255, but it turns out to be an upstream bug in GMT 6.0.0. The upstream bug was fixed in GMT 6.1.0. This PR adds a test to make sure that GMT won't break it again.

**Reminders**


pygmt/tests/test_clib_put_vector.py
```diff
@@ -0,0 +1,132 @@
...
+def test_put_vector_mixed_dtypes():
+    """Passing a numpy array of mixed dtypes to a dataset.
+    See https://github.com/GenericMappingTools/pygmt/issues/255
+    """
+    dtypes = "float32 float64 int32 int64 uint32 uint64".split()
+    for dtypex in dtypes:
+        for dtypey in dtypes:
+            # skip same dtypes
+            if dtypex == dtypey:
+                continue
```
weiji14, line 58: The Python standard library has an [`itertools.combination`](https://docs.python.org/3.8/library/itertools.html#itertools.combinations) function to do exactly this. You'll need to add `import itertools` at the top first, and then change the following lines to this:

```suggestion
    for dtypex, dtypey in itertools.combinations(iterable=dtypes, r=2):
```

seisman: Great to learn new things everyday!

But it's slightly different from what I want. This is an example:

```python
In [8]: for x, y in itertools.combinations([1, 2, 3], r=2):
   ...:     print(x, y)
   ...:
1 2
1 3
2 3
```
but I want 
```
1 2 
1 3
2 1
2 3
3 1
3 2
```

weiji14: Use [`itertools.permutations`](https://docs.python.org/3.8/library/itertools.html#itertools.permutations) then? I don't think the order matters here, unless there's a module I'm not aware of that's picky about this.

seisman: The idea of the test is to verify the GMT bug fixed in https://github.com/GenericMappingTools/gmt/issues/3330. As I understand it, the order matters.

`itertools.permutations` is exactly what I need. Will make the change later.

PR #546, opened 2020-07-23
Wrap x2sys_init and x2sys_cross

weiji14:
**Description of proposed changes**

Wrapping the [x2sys_cross]( https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_cross.html) function (which also requires [x2sys_init](https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_init.html)), implemented under x2sys.py. [`X2SYS` is a tool for analyzing intersecting track data](https://doi.org/10.1016/j.cageo.2009.05.009) (e.g. ships measuring ocean bathymetry, satellite altimeters measuring land elevation)!

Live documentation preview for this Pull Request is at https://pygmt-git-x2syscross.gmt.vercel.app/api/generated/pygmt.x2sys_cross.html.

Example code:

```python
import pygmt

fname: str = pygmt.which("@tut_ship.xyz", download="a")
pygmt.x2sys_init(tag="GMT", fmtfile="xyz", force=True)
df: pd.DataFrame = x2sys_cross(tracks=[fname], tag=tag, coe="i")

assert df.shape == (14294, 12)
```

This is useful for performing crossover analysis (e.g. finding the elevation error between intersecting tracks). It will be tremendously useful for some of my [ICESat-2 satellite laser altimetry work](https://github.com/weiji14/deepicedrain), and having used it in GMT 5 back in 2015 for ICESat-1, I think it's time to bring it to the Python world!

TODO:
- [x] Alias short-form arguments in `x2sys_init`
  - [x] fmtfile (D)
  - [x] suffix (E)
  - [x] force (F)
  - [x] discontinuity ([G](https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_init.html#g))
  - [x] spacing (I)
  - [x] units ([N](https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_init.html#n)) (Nd or Nt)
  - [x] region (R)
  - [x] gap ([W](https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_init.html#w)) (Wd or Wt)
  - [x] distcalc ([j](https://docs.generic-mapping-tools.org/6.1/gmt.html#distcalc-full))
- [x] Alias short-form arguments in `x2sys_cross`
  - [x] combitable (A)
  - [x] runtimes (C)
  - [x] override ([D](https://docs.generic-mapping-tools.org/6.1/supplements/x2sys/x2sys_cross.html#d))
[...]

PR #550, opened 2020-07-27
Add common alias verbose (V)

weiji14:

Select verbosity level [Default is w], which modulates the messages written to stderr. See https://docs.generic-mapping-tools.org/6.1/gmt.html#v-full. Choose among 7 levels of verbosity:

- **q** - Quiet, not even fatal error messages are produced
- **e** - Error messages only
- **w** - Warnings [Default]
- **t** - Timings (report runtimes for time-intensive algorthms);
- **i** - Informational messages (same as "verbose=True")
- **c** - Compatibility warnings
- **d** - Debugging messages

Cherry-picked from #546.

Fixes #

**Reminders**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -36,6 +36,18 @@ COMMON_OPTIONS = {
         color : str
             Select color or pattern for filling of symbols or polygons. Default
             is no fill.""",
+    "V": """\
+        verbose : str
+            Select verbosity level [Default is w], which modulates the messages
+            written to stderr. Choose among 7 levels of verbosity:
+
+            - **q** - Quiet, not even fatal error messages are produced
+            - **e** - Error messages only
+            - **w** - Warnings [Default]
+            - **t** - Timings (report runtimes for time-intensive algorthms);
+            - **i** - Informational messages (same as "verbose=True")
+            - **c** - Compatibility warnings
+            - **d** - Debugging messages""",
     "W": """\
         pen : str
             Set pen attributes for lines or the outline of symbols.""",


pygmt/helpers/decorators.py
```diff
@@ -36,6 +36,18 @@
         color : str
             Select color or pattern for filling of symbols or polygons. Default
             is no fill.""",
+    "V": """\
+        verbose : str
```
seisman, line 40: FYI, the GMT experimental long options use `verbosity` and `spherical` for **-V** and **-j** (https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_init.c#L383), while GMT.jl uses `verbose` and `spheric_dist | spherical_dist`(https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.grdmask).

weiji14: I would definitely prefer 'verbose' for **V**, it's similar to most command-line arguments.

As for 'j', I've merged the PR already, but could make a quick change. The GMT.jl documentation actually points to http://docs.generic-mapping-tools.org/latest/gmt.html#j-full (i.e. capital **J**) rather than the proper link at https://docs.generic-mapping-tools.org/latest/gmt.html#distcalc-full (notice the `distcalc` in the link).

seisman: Yes, I also prefer `verbose` for **-V**. 

For **j**, I'm OK with the names. It depends on if we want to keep consistent with other GMT projects.

seisman: FYI, I believe the GMT server is up.

PR #555, opened 2020-08-06
Add testing.check_figures_equal to avoid storing baseline images

seisman:

This PR adds a new function `check_figures_equal` to check if two `pygmt.Figure()` objects are the same, mostly inspired by comment https://github.com/GenericMappingTools/pygmt/pull/522#pullrequestreview-447735647 and the matplotlib decorator [`check_figures_equal`](https://matplotlib.org/3.2.2/api/testing_api.html#matplotlib.testing.decorators.check_figures_equal).

What the function `check_figures_equal` does is very simple:

1. It accepts two pygmt.Figure() objects (`fig_ref` and `fig_test`) as arguments
2. save them into two PNG images
3. compare these two images using the `compare_images` function from matplotlib and calculate the RMS value
4. if the two images are the same, then delete them; if the _rms_ is greater than _tol_, generate the diff image and raise an exception.

I add three new tests:

- `test_check_figures_equal` checks the case when two images are equal
- `test_check_figures_unequal` checks the case when two images are unequal. The exception is correctly caught by `pytest.raises`.
- `test_grdimage_central_longitude` is a real test to check the images generated by passing a matrix or a grid to grdimage (we always assume that passing a netCDF grid to GMT gives the correct/reference image). The test helps me find a GMT bug, which is usually very difficult to detect (see https://github.com/GenericMappingTools/gmt/issues/3844 and https://github.com/GenericMappingTools/gmt/issues/3849).

Some known issues/limitations:

1. some paths are hard-coded, but they may be improved to be more flexible
2. The test `test_check_figures_unequal()` checks if the function `check_figures_equal` correctly raises the `GMTImageComparisonFailure` exception when two images are unequal, and I use pytest.raises to catch the exception so that the test passes. ~~However, the two images are not deleted after the test.~~ They are now.

Fixes #

**Reminders**


weiji14:
@seisman, hope you don't mind if I continue some work on this.

> 1. some paths are hard-coded, but they may be improved to be more flexible

Will try to see if it's possible to mimic the [pytest-mpl](https://github.com/matplotlib/pytest-mpl) paths, or look to matplotlib to see how they do it.

> 2. The test `test_check_figures_unequal()` checks if the function `check_figures_equal` correctly raises the `GMTImageComparisonFailure` exception when two images are unequal, and I use pytest.raises to catch the exception so that the test passes. However, the two images are not deleted after the test.

It should be easy-enough to use GMTTempFile so that the images are cleaned up after the test, but we'll need to find a way to keep the images if the test fails so that they can be examined.

pygmt/helpers/testing.py
```diff
@@ -0,0 +1,37 @@
+import os
+import sys
+from pathlib import Path
+
+from matplotlib.testing.compare import compare_images
+
+from ..exceptions import GMTImageComparisonFailure, GMTInvalidInput
+
+
+def check_figures_equal(fig_ref, fig_test, fig_prefix=None, tol=0.0):
```
weiji14, line 10: Maybe we should put this into `decorators.py` instead? It's probably easier to do something like this:

```python
@check_figures_equal
def test_some_pygmt_function():
    pass
```

See also https://realpython.com/primer-on-python-decorators/#simple-decorators

seisman: Feel free to do any changes you want.

weiji14: Ok, I've converted `check_figures_equal` into a decorator at d2ad3f56ba7c2371e4b42c1de998795e4ad7d010 and put the tests as a doctest. It's a bit tricky to make a Python decorator with parameters (see https://stackoverflow.com/questions/5929107/decorators-with-parameters), so still needs a bit more work.

pygmt/tests/test_grdimage.py
```diff
@@ -93,3 +94,12 @@ def test_grdimage_over_dateline(xrgrid):
...
     return fig
+
+
+@check_figures_equal()
+def test_grdimage_central_longitude(grid, fig_ref, fig_test):
+    """
+    Test that plotting a grid centred at different longitudes/meridians work.
+    """
+    fig_ref.grdimage("@earth_relief_01d_g", projection="W120/15c", cmap="geo")
+    fig_test.grdimage(grid, projection="W120/15c", cmap="geo")
```
weiji14, line 105:
```suggestion
@pytest.mark.parametrize("meridian", [0, 33, 120, 180])
@check_figures_equal()
@pytest.mark.parametrize("proj_type", ["H", "Q", "W"])
def test_grdimage_different_central_meridians_and_projections(
    grid, proj_type, meridian, fig_ref, fig_test
):
    """
    Test that plotting a grid centred on different meridians using different
    projection systems work.
    """
    fig_ref.grdimage(
        "@earth_relief_01d_g", projection=f"{proj_type}{meridian}/15c", cmap="geo"
    )
    fig_test.grdimage(grid, projection=f"{proj_type}{meridian}/15c", cmap="geo")

```

I'll update this test in #560 after this PR is merged :smile:.

weiji14 (commented):
Ok, this is pretty much ready for review. After stress testing this with `pytest`, I think the `check_figures_equal` function will have to end up pretty much exactly like matplotlib's https://matplotlib.org/3.3.1/_modules/matplotlib/testing/decorators.html#check_figures_equal, **except** that we're using `fig=pygmt.Figure()` instead of matplotlib's `plt.subplot`. If only we can monkeypatch it!!

It might be worth raising an issue/PR on matplotlib to see if there's a way to reduce code duplication (for the sake of long term maintainability), as they've clearly spent months/years of thought and effort into this one `check_figures_equal` function. This would involve some sort of subclassing (or I don't know, pluggability?), so that we can swap in `pygmt.Figure` into `fig_ref` and `fig_test`, while getting all of the image_compare goodness.

pygmt/helpers/decorators.py
```diff
@@ -5,14 +5,16 @@
...
 etc.
 """
-import textwrap
 import functools
+import inspect
+import os
+import textwrap
 
 import numpy as np
+from matplotlib.testing.compare import compare_images
```
seisman, line 14:
```
from matplotlib.testing.compare import compare_images
``` 
As I understand it, the code means that now `matplotlib` becomes a required dependency, even for users who never run the tests, right? 

Although PyGMT already requires `matplotlib` for testings and most users usually have matplotlib installed. I still don't want to add one dependency to PyGMT. 

When I wrote the first commit (8b7861441f3a150dbc01fba4465ae577db766328), I put the codes in `pygmt/helpers/testing.py`. By doing that way, I think matplotlib is still optional, although I haven't tested it.

weiji14: Good point, I think you're right here. I also encountered issues with circular imports when moving the code to decorators.py, hence this line:

https://github.com/GenericMappingTools/pygmt/blob/04b3f4154ab35a345c3394cd393e050f306f64ae/pygmt/helpers/decorators.py#L459

Probably should move it back under `pygmt/helpers/testing.py` then. As an aside, I've opened up a feature request at https://github.com/matplotlib/pytest-mpl/issues/94, and we might be able to do all this from `pytest-mpl` in the future.

pygmt/helpers/testing.py
```diff
@@ -0,0 +1,105 @@
...
+    --------
+
+    >>> import pytest
+    >>> import shutil
+
+    >>> @check_figures_equal(result_dir="tmp_result_images")
+    ... def test_check_figures_equal(fig_ref, fig_test):
+    ...     fig_ref.basemap(projection="X5c", region=[0, 5, 0, 5], frame=True)
+    ...     fig_test.basemap(projection="X5c", region=[0, 5, 0, 5], frame="af")
+    >>> test_check_figures_equal()
```
seisman, line 44: Check that the `tmp_result_images` directory doesn't exist when the figures are equal?

pygmt/helpers/testing.py
```diff
@@ -0,0 +1,105 @@
...
+
+    >>> @check_figures_equal(result_dir="tmp_result_images")
+    ... def test_check_figures_unequal(fig_ref, fig_test):
+    ...     fig_ref.basemap(projection="X5c", region=[0, 5, 0, 5], frame=True)
+    ...     fig_test.basemap(projection="X5c", region=[0, 3, 0, 3], frame=True)
+    >>> with pytest.raises(GMTImageComparisonFailure):
+    ...     test_check_figures_unequal()
+
+    >>> shutil.rmtree(path="tmp_result_images")  # cleanup folder if tests pass
```
seisman, line 53: Check the existence of the three images ("actual", "expected", "diff") in the directory?

pygmt/helpers/testing.py
```diff
@@ -0,0 +1,105 @@
...
+import inspect
+import os
+
+from matplotlib.testing.compare import compare_images
+
+from ..exceptions import GMTImageComparisonFailure
+from ..figure import Figure
+
+
+def check_figures_equal(*, tol=0.0, result_dir="result_images"):
```
seisman, line 14: One more thing about the result_dir is that, the `check_figures_equal` decorator generates images in `result_images` directory, while `pytest.mark.mpl_image_compare` generates images in directories like `results/tmpjtnnwqt4`.

weiji14: Yes, which was partly why I opened up the issue at https://github.com/matplotlib/pytest-mpl/issues/94, to get all of that `pytest-mpl` goodness (e.g. not having a hardcoded `result_dir`). I'll *try* to make a Pull Request to `pytest-mpl` for that, so we can just use a proper `@pytest.mark.mpl_check_equal` decorator in the future (will open a new issue after this one is merged). For now though, since we don't have many tests using `check_figures_equal` yet, we can probably just leave it like so.

pygmt/helpers/testing.py
weiji14: The directory will exist, but there won't be any files in it. Could check that the folder is empty.

pygmt/helpers/testing.py
weiji14: Ok, it'll make the doctest a bit longer but I can write this in.

pygmt/helpers/testing.py
seisman: Yes, looks good to me.

CONTRIBUTING.md
```diff
@@ -310,8 +310,38 @@ Leave a comment in the PR and we'll help you out.
 
 ### Testing plots
 
-We use the [pytest-mpl](https://github.com/matplotlib/pytest-mpl) plug-in to test plot
-generating code.
+Writing an image-based test is only slightly more difficult than a simple test.
+The main consideration is that you must specify the "baseline" or reference
+image, and compare it with a "generated" or test image. This is handled using
```
weiji14, line 315: Added some notes here to CONTRIBUTING.md, adapted from https://matplotlib.org/3.3.1/devel/testing.html#writing-an-image-comparison-test.

seisman (commented):
Looks good to me, but I can't approve it because I opened this PR.

weiji14 (approved):
Ah yes, I'll approve it then :laughing:

PR #556, opened 2020-08-06
Fix a failing test due to GMT API change

seisman:

The test `test_put_matrix_grid` fails with GMT latest due to recent changes of GMT API
in https://github.com/GenericMappingTools/gmt/pull/3848.

Before this PR, for a matrix grid, calling GMT_Write_Data always writes
the matrix grid as a matrix, no matter the geometry is `GMT_IS_POINT` or
`GMT_IS_SURFACE`.

This PR declares it as a bug (or an improvement). With that PR merged,
`geometry=GMT_IS_POINT` writes the matrix as a table, `geometry=GMT_IS_SURFACE`
writes the matrix grid as a netCDF file.

This test expects a matrix, but the new API writes a netCDF grid.
That's why it fails.

This PR fixes the issue by simply changing `GMT_IS_SURFACE` to
`GMT_IS_POINT`, so that the test can pass for GMT 6.1.0, 6.1 and master
branches.

Fixes #

**Reminders**

### DIFF
--- a/pygmt/tests/test_clib_put_matrix.py
+++ b/pygmt/tests/test_clib_put_matrix.py
@@ -76,7 +76,7 @@ def test_put_matrix_grid():
             with GMTTempFile() as tmp_file:
                 lib.write_data(
                     "GMT_IS_MATRIX",
-                    "GMT_IS_SURFACE",
+                    "GMT_IS_POINT",
                     "GMT_CONTAINER_AND_DATA",
                     wesn,
                     tmp_file.name,


pygmt/tests/test_clib_put_matrix.py
```diff
@@ -76,7 +76,7 @@ def test_put_matrix_grid():
             with GMTTempFile() as tmp_file:
                 lib.write_data(
                     "GMT_IS_MATRIX",
-                    "GMT_IS_SURFACE",
+                    "GMT_IS_POINT",
```
weiji14, line 79: It doesn't seem like we use `lib.write_data` on any user-facing code, or very much in the clib. I know GMT can write an ASCII grid, but maybe we should also test with "GMT_IS_SURFACE" to ensure that:

1. `lib.write_data` can write to a NetCDF properly; and
2. We can use `xarray` to read from that NetCDF file

seisman: Sounds a good idea, but it needs 6.1.1, so better to add the test in a new pr and only merge that pr after we bump to gmt 6.1.1

weiji14: Ah ok (forgot that we're still on GMT 6.1.0). Leave it until end of the month then.

PR #562, opened 2020-08-14
Allow datetime inputs to region argument

### LINKED ISSUE #561: Support datetime types in region argument

Support passing in 'datetime' like inputs into the 'region' or '-R' argument in `plot`.

>> * The `region` argument doesn't work with numpy `datetime64` objects, have to convert to a string using [`np.datetime_as_string`](https://numpy.org/devdocs/reference/generated/numpy.datetime_as_string.html) in order to set the map frame bounds.

> The decorator `kwargs_to_strings` convert the list `[w, e, s, n]` to a string *w/e/s/n*. We need to convert datetime64 object to string internally, but it's not ideal to do the conversion in the decorator. I don't have a good solution now.

_Originally posted by @seisman in https://github.com/GenericMappingTools/pygmt/pull/464#issuecomment-636585247_

Minimal working example:

```python
import pandas as pd
import pygmt

fig = pygmt.Figure()
fig.plot(x=["2020-06-30"], y=[1], region=[pd.Timestamp("2020-01-01"), pd.Timestamp("2020-12-31"), 0, 2])
fig.show()
```

The error message is:

```python-traceback
GMTCLibError: Module 'plot' failed with status code 71:
plot [ERROR]: Option -R parsing failure. Correct syntax:
plot [ERROR]: Offending option -R2020-01-01
```

This workaround does work, but it's cumbersome:

```python
fig = pygmt.Figure()
fig.plot(
    x=[pd.Timestamp("2020-06-30")],
    y=[1],
    region=[
        pd.Timestamp("2020-01-01").isoformat(),
        pd.Timestamp("2020-12-31").isoformat(),
        0,
        2,
    ],
    style="a10c",
    frame=True,
)
fig.show()
```

produces:

Expected plot with star in middle

**Are you willing to help implement and maintain this feature?** Yes, but where should this code logic go?

weiji14:

Allow users to pass in datetime-like inputs to the 'region' argument in modules like `plot`, `basemap`, and so on. Helpful for people who usually just want to use something like `region = [data.min(), data.max(), ...]`.

Includes support for:

- [x] Python's [built-in datetime and date](https://docs.python.org/3/library/datetime.html)
- [x] [numpy.datetime64](https://numpy.org/doc/stable/reference/arrays.datetime.html)
- [x] [pandas.Timestamp](https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.html)
- [x] [xarray.DataArray](http://xarray.pydata.org/en/v0.16.0/generated/xarray.DataArray.html#xarray.DataArray) with 'datetime64' dtype

What is not supported:

- Raw datetime strings in non-ISO format, e.g.,  `1/1/2018`, `Jul 5, 2019`

Fixes #561

**Reminders**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -359,9 +359,13 @@ def kwargs_to_strings(convert_bools=True, **conversions):
                     if issequence and is_nonstr_iter(value):
                         for index, item in enumerate(value):
                             try:
+                                # check if there is a space " " when converting
+                                # a pandas.Timestamp/xr.DataArray to a string.
+                                # If so, use np.datetime_as_string instead.
                                 assert " " not in str(item)
                             except AssertionError:
-                                # convert datetime-like items to string format
+                                # convert datetime-like item to ISO 8601
+                                # string format like YYYY-MM-DDThh:mm:ss.ffffff
                                 value[index] = np.datetime_as_string(
                                     np.asarray(item, dtype=np.datetime64)
                                 )
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -8,6 +8,8 @@ etc.
 import textwrap
 import functools
 
+import numpy as np
+
 from .utils import is_nonstr_iter
 from ..exceptions import GMTInvalidInput
 
@@ -302,6 +304,23 @@ def kwargs_to_strings(convert_bools=True, **conversions):
     >>> module(123, bla=(1, 2, 3), foo=True, A=False, i=(5, 6))
     {'bla': (1, 2, 3), 'foo': '', 'i': '5,6'}
     args: 123
+    >>> import datetime
+    >>> module(
+    ...     R=[
+    ...         np.datetime64("2010-01-01T16:00:00"),
+    ...         datetime.datetime(2020, 1, 1, 12, 23, 45),
+    ...     ]
+    ... )
+    {'R': '2010-01-01T16:00:00/2020-01-01T12:23:45.000000'}
+    >>> import pandas as pd
+    >>> import xarray as xr
+    >>> module(
+    ...     R=[
+    ...         xr.DataArray(data=np.datetime64("2005-01-01T08:00:00")),
+    ...         pd.Timestamp("2015-01-01T12:00:00.123456789"),
+    ...     ]
+    ... )
+    {'R': '2005-01-01T08:00:00.000000000/2015-01-01T12:00:00.123456'}
 
     """
     valid_conversions = [
@@ -338,9 +357,15 @@ def kwargs_to_strings(convert_bools=True, **conversions):
                     value = kwargs[arg]
                     issequence = fmt in separators
                     if issequence and is_nonstr_iter(value):
-                        kwargs[arg] = separators[fmt].join(
-                            "{}".format(item) for item in value
-                        )
+                        for index, item in enumerate(value):
+                            try:
+                                assert " " not in str(item)
+                            except AssertionError:
+                                # convert datetime-like items to string format
+                                value[index] = np.datetime_as_string(
+                                    np.asarray(item, dtype=np.datetime64)
+                                )
+                        kwargs[arg] = separators[fmt].join(f"{item}" for item in value)
             # Execute the original function and return its output
             return module_func(*args, **kwargs)
 
--- a/pygmt/tests/test_plot.py
+++ b/pygmt/tests/test_plot.py
@@ -294,7 +294,16 @@ def test_plot_scalar_xy():
 def test_plot_datetime():
     """Test various datetime input data"""
     fig = Figure()
-    fig.basemap(projection="X15c/5c", region="2010-01-01/2020-01-01/0/10", frame=True)
+    fig.basemap(
+        projection="X15c/5c",
+        region=[
+            np.array("2010-01-01T00:00:00", dtype=np.datetime64),
+            pd.Timestamp("2020-01-01"),
+            0,
+            10,
+        ],
+        frame=True,
+    )
 
     # numpy.datetime64 types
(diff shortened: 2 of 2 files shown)


pygmt/helpers/decorators.py
```diff
@@ -338,9 +357,15 @@ def new_module(*args, **kwargs):
...
-                        kwargs[arg] = separators[fmt].join(
-                            "{}".format(item) for item in value
-                        )
+                        for index, item in enumerate(value):
+                            try:
+                                assert " " not in str(item)
+                            except AssertionError:
+                                # convert datetime-like items to string format
+                                value[index] = np.datetime_as_string(
+                                    np.asarray(item, dtype=np.datetime64)
```
seisman, line 366: We use `np.datetime_as_string(array_to_datetime(vector))` to convert datetime arrays to string arrays. 
https://github.com/GenericMappingTools/pygmt/blob/2ddbb0e1977af85bf6a6ac63d7cbac9da3d08f05/pygmt/clib/session.py#L778-L782
The `array_to_datetime` function uses `pd.to_datetime(array)`
https://github.com/GenericMappingTools/pygmt/blob/2ddbb0e1977af85bf6a6ac63d7cbac9da3d08f05/pygmt/clib/conversion.py#L324

Any specific reason to use `np.asarray` here?

pygmt/helpers/decorators.py
```diff
@@ -338,9 +357,15 @@ def new_module(*args, **kwargs):
                     value = kwargs[arg]
                     issequence = fmt in separators
                     if issequence and is_nonstr_iter(value):
-                        kwargs[arg] = separators[fmt].join(
-                            "{}".format(item) for item in value
-                        )
+                        for index, item in enumerate(value):
+                            try:
+                                assert " " not in str(item)
```
seisman, line 362: I'm a little confused how the `assert` statement works. Could you please explain a little bit?

pygmt/helpers/decorators.py
```diff
@@ -338,9 +357,15 @@ def new_module(*args, **kwargs):
                     value = kwargs[arg]
                     issequence = fmt in separators
                     if issequence and is_nonstr_iter(value):
-                        kwargs[arg] = separators[fmt].join(
-                            "{}".format(item) for item in value
-                        )
+                        for index, item in enumerate(value):
```
seisman, line 360: Not sure if I understand your codes correctly. It seems the codes check all arguments. For example, **perspective=[30, 50]** is converted to **-p30/50**, and the codes also check if `30` and `50` are datetime-like, right?

However, **-p** doesn't accept any datetime-like values. It looks a waste of time. Perhaps we should check whether `arg="B"` (maybe arg="region")?

pygmt/helpers/decorators.py
weiji14: Main reason is to handle `xarray.DataArray` inputs. The `array_to_datetime`/`pd.to_datetime` function is meant to handle arrays/lists, but not scalar values. I.e. this doesn't work:

```python
import pandas as pd
import xarray as xr

pd.to_datetime(xr.DataArray(data=np.datetime64("2005-01-01")))
# TypeError: len() of unsized object
```

but this does:

```python
import numpy as np
import xarray as xr

np.asarray(xr.DataArray(data=np.datetime64("2005-01-01")))
# array('2005-01-01T00:00:00.000000000', dtype='datetime64[ns]')
```

pygmt/helpers/decorators.py
weiji14: Sorry that this isn't very readable, it's following the [EAFP style](https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python#11360880), usually used when we're handling only a few exceptional cases. The original reason that we can't pass in 'datetime'-like values to the 'region' argument is because there's a space `" "` when converting a `pandas.Timestamp` to a string (instead of a `"T"` that would make it ISO compliant).

```python
import pandas as pd

str(pd.Timestamp("2015-01-01T12:00:00.123456789"))
# '2015-01-01 12:00:00.123456789'
```

This also applies to `xarray.DataArray` datetime values. So the `assert` here checks for the space `" "`, and if it finds it, it will try to convert the item into an ISO compliant datetime value (e.g. `"2015-01-01T12:00:00.123456789"`).

pygmt/helpers/decorators.py
weiji14: > Not sure if I understand your codes correctly. It seems the codes check all arguments. For example, **perspective=[30, 50]** is converted to **-p30/50**, and the codes also check if `30` and `50` are datetime-like, right?

Yes, we do check every item using `assert " " not in str(item)`, but this is quite a fast string parsing check (see also the other comment on EAFP syntax). It is *not* explictily checking for datetime-like (which *would* be a slower check) using something like `isinstance(item, (pd.Timestamp, datetime.datetime, etc))`.

> However, **-p** doesn't accept any datetime-like values. It looks a waste of time. Perhaps we should check whether `arg="B"` (maybe arg="region")?

This would be an extra check on top of the (rather fast) `assert " " not in str(item)`, and yes, we could add it I suppose. But really, there's usually not very many items to parse here, only 4 if using `[xmin, xmax, ymin, ymax]`, maybe 6 if using `[xmin, xmax, ymin, ymax, zmin, zmax]`.

Happy to consider an alternative way though if you have any ideas, I definitely think there's a better way to do this. The unit tests are all there, so feel free to refactor/change the code as long as it passes the tests!

pygmt/helpers/decorators.py
```diff
@@ -338,9 +357,15 @@ def new_module(*args, **kwargs):
...
+                            try:
```
seisman, line 361: > because there's a space " " when converting a pandas.Timestamp to a string

Please add a comment here to make it more readable.

PR #574, opened 2020-09-02
Allow pandas.DataFrame table and 1D/2D numpy array inputs into pygmt.info

### LINKED ISSUE #564: Let pygmt.info support pandas.DataFrame or 2D numpy.array inputs

The [pygmt.info](https://www.pygmt.org/v0.1.2/api/generated/pygmt.info.html) function should accept pandas.DataFrame or 2D numpy.array matrices as input. This was mentioned in passing when `pygmt.info` was originally implemented at https://github.com/GenericMappingTools/pygmt/pull/106#discussion_r160516382. Currently it can only handle filename (str) types:

https://github.com/GenericMappingTools/pygmt/blob/bedb48d5cf105c37573f8e8c5d74b97f1f741187/pygmt/modules.py#L89-L90

What I'm particularly interested in is getting a nicely rounded xmin/xmax/ymin/ymax 'region' by simply passing in a `pandas.DataFrame` (Option 3 in the related issue at #147). Something like this:

```python
import pandas as pd

df: pd.DataFrame = pygmt.datasets.load_ocean_ridge_points()
region = pygmt.info(df, spacing="0.1")  # pygmt.info("@ridge.txt", I="0.1")
print(region)
# would output '-R-180/180/-65.7/86.8\n', or better, a proper python list
```

Yes, it's possible to do `region = [df.x.min(), df.x.max(), df.y.min(), df.y.max()]`, but [`gmtinfo -I`](https://docs.generic-mapping-tools.org/6.1/gmtinfo#i) does rounding (and I believe it rounds outside of the tight region?), and it has other smart functionality too.

It shouldn't be too hard to make `pygmt.info` support non-filename types, just need to use `lib.virtualfile_from_vectors` or the like.

**Note** that the return values of `pygmt.info` should be discussed at #147. This issue only tackles the input format type. Perhaps we should handle #147 before this tackling this issue.

weiji14:

Let [pygmt.info](https://www.pygmt.org/v0.1.2/api/generated/pygmt.info.html) support pandas.DataFrame and 1D/2D numpy array inputs. Enables code like so:

```python
import pandas as pd

df: pd.DataFrame = pygmt.datasets.load_ocean_ridge_points()
region = pygmt.info(table=df, spacing="0.1")  # pygmt.info("@ridge.txt", I="0.1")
print(region)
# '-R-180/180/-65.7/86.8\n',
```

I've renamed the 'fname' argument to 'table' (which is a :boom: breaking change) since `info` supports both file name inputs and pandas.DataFrame tables now.

Fixes #564

**Reminders**


pygmt/modules.py
```diff
@@ -55,7 +55,7 @@ def grdinfo(grid, **kwargs):
 
 @fmt_docstring
 @use_alias(C="per_column", I="spacing", T="nearest_multiple")
-def info(fname, **kwargs):
+def info(table, **kwargs):
```
weiji14, line 58: There's actually an API change here (argument name change from `fname` to `table`), and maybe we should put this into v0.2.0?

seisman: It looks a little weird that the new `info()` function support pandas.DataFrame, but not 2D numpy arrays, and the return value `-R-180/180/-65.7/86.8\n` looks useless for Python.

weiji14: > It looks a little weird that the new `info()` function support pandas.DataFrame, but not 2D numpy arrays

Yes, I could make it support 2D numpy arrays too, lemme work on that.

> and the return value `-R-180/180/-65.7/86.8\n` looks useless for Python.

That's dealt with in a parallel PR at #575 (which is branched off this PR).

weiji14: >> It looks a little weird that the new info() function support pandas.DataFrame, but not 2D numpy arrays

>Yes, I could make it support 2D numpy arrays too, lemme work on that.

Just on this, does gmt `info` work on single column data as well? i.e. 1D arrays?

seisman: Yes. For a one-column data,
```
1
2
3
4
5
```
```
gmt info test.txt -I1  
-R1/5/0/0
```

weiji14: Ok, will support this too then.

pygmt/modules.py
```diff
@@ -88,14 +90,25 @@ def info(fname, **kwargs):
...
         of dz and output this as the string *-Tzmin/zmax/dz*.
     """
-    if not isinstance(fname, str):
-        raise GMTInvalidInput("'info' only accepts file names.")
+    kind = data_kind(table)
+    with Session() as lib:
+        if kind == "file":
+            file_context = dummy_context(table)
+        elif kind == "matrix":
+            _table = np.asanyarray(table)
```
weiji14, line 98: Just FYI, this line converts `pandas.DataFrame` into a numpy.ndarray, while allowing numpy.ndarray objects to just pass through. A bit easier than using `_table = table.values if hasattr(table, values) else table`, and I think this allows for Python lists to work too (though I haven't tested it).

See also:
- https://stackoverflow.com/questions/14415741/what-is-the-difference-between-numpys-array-and-asarray-functions
- https://stackoverflow.com/questions/13187778/convert-pandas-dataframe-to-numpy-array

seisman (approved):
Looks good

PR #575, opened 2020-09-02
Sensible array outputs for pygmt info

### LINKED ISSUE #147: Return values of gmt.info

weiji14:

Output the result  of [`pygmt.info`](https://www.pygmt.org/v0.1.2/api/generated/pygmt.info.html) as a numpy.ndarray when either of per_column (C), spacing (I), or nearest_multiple (T) are used. This would be more usable that a raw string that is meant for the command line world.

This implements @seisman's original suggestion at https://github.com/GenericMappingTools/pygmt/issues/147#issue-305123527, with both 1D and 2D array outputs supported!

Examples:

```python
import pygmt

table = pygmt.datasets.load_ocean_ridge_points()

info(table=table, per_column=True)  # for basemap region -R
# array([-179.9401,  179.935 ,  -65.6182,   86.8   ])

info(table=table, nearest_multiple=0.1)  # for makecpt series -T
# array([-1.8e+02,  1.8e+02,  1.0e-01])

info(table=table, spacing=0.1)  # for basemap region -R
# array([-180. ,  180. ,  -65.7,   86.8])

info(table=table, spacing="b")  # for plot -Ap
# array([[-179.9401,  -65.6182],
#        [ 179.935 ,  -65.6182],
#        [ 179.935 ,   86.8   ],
#        [-179.9401,   86.8   ],
#        [-179.9401,  -65.6182]])
```

The parsing uses a bit of regex to remove the "-R", "-I" from the raw string, and split on "/" or " " delimiters.

https://github.com/GenericMappingTools/pygmt/blob/05a3f23ee259ba76bd11368bba2f92f02de0df9f/pygmt/modules.py#L125-L130

**Note**: Merge after #574!!! :heavy_check_mark: 

Fixes #147

**Reminders**


pygmt/modules.py
```diff
@@ -108,7 +122,16 @@ def info(table, **kwargs):
...
                 lib.call_module("info", arg_str)
-            return tmpfile.read()
+            result = tmpfile.read()
+
+        if any(arg in kwargs for arg in ["C", "I", "T"]):
+            # Converts certain output types into a numpy array
+            # instead of a raw string that is less useful.
+            result = np.loadtxt(
+                re.sub(pattern="-R|-T|/", repl=" ", string=result).splitlines()
+            )
```
seisman, line 132: Is it OK to check if a string starts with `-R` or `-T`?

weiji14: Yes, with `str.startswith` (see https://stackoverflow.com/questions/20461847/str-startswith-with-a-list-of-strings-to-test-for#20462341). I try to avoid regex as much as I can, but after several ways of experimenting on how to handle 1) `-Rw/e/s/n`; 2) `-Tmin/max/inc`; 3) `xmin xmax ymin ymax` 4) `x1 y1 x2 y2` (in several rows), this seems to be the most succinct way. I can try again though.

weiji14: Actually, this works too :grin::

https://github.com/GenericMappingTools/pygmt/blob/a52127fae0400701b60119704f5fc787f2fb55a2/pygmt/modules.py#L128-L130

seisman: Great!

seisman (approved):

weiji14:
Cool, thanks for reviewing, this will be a great addition for v0.2.0! I'll temporarily disable the required Windows Py3.8 test (random unrelated failure we should investigate at some point) to merge this.

PR #576, opened 2020-09-02
RFC Allow for io.StringIO inputs to certain modules

### LINKED ISSUE #571: Allow Figure.legend to read from StringIO
I need to display a legend of symbols for the magnitude of earthquakes. Currently the legend specification must be a file on disk. The Pythonic implementation should allow **Figure.legend(spec=**_<buffer|file>_**,...)**. My preference is to write the legend specification to io.StringIO buffer and then use the buffer for Figure.legend. Instead I have to generate the legend spec in program and write to a file and then specify **spec='my_legend.txt'**.

Figure.legend has default arg **spec=None**. I have no idea how this default works, that's a separate issue: _improve documentation_.

I am willing to implement, but have no experience with such participation.

weiji14:

Make a `tempfile_from_buffer` backend function for [`io.StringIO`](https://docs.python.org/3/library/io.html#io.StringIO) inputs. Will be useful for modules where:

- an in-memory stream of 'string-like' data makes sense, e.g. [`Figure.legend`](https://www.pygmt.org/v0.1.2/api/generated/pygmt.Figure.legend.html#pygmt.Figure.legend)'s 'spec' option (see also #571)
- a 'filename' is the only way to pass in data as using GMT virtualfiles don't work, e.g. [`x2sys_cross`](https://www.pygmt.org/dev/api/generated/pygmt.x2sys_cross.html) (see also https://github.com/GenericMappingTools/pygmt/pull/546#discussion_r459366392)

This is implemented as the 'hacky' Option 1 at https://github.com/GenericMappingTools/pygmt/issues/571#issuecomment-681139099.

Happy to discuss about the actual implementation details, such as:
- whether 'tempfile_from_buffer' is a good name
- should this live inside 'tempfile.py', or is it better placed with the other 'virtualfile_from_*' functions (which is what I modeled 'tempfile_from_buffer' after).
- etc

See also:
- https://stackoverflow.com/questions/11892623/stringio-and-compatibility-with-with-statement-context-manager
- https://stackoverflow.com/questions/3253258/what-is-the-best-way-to-write-the-contents-of-a-stringio-to-a-file#3253819
- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html?highlight=stringio

Fixes #571

**Reminders**


weiji14:
@dr-glenn, do you want to try this branch using `pip install https://github.com/GenericMappingTools/pygmt/archive/stringio_input.zip` and see if it works for your StringIO legend? Documentation (beta) is at https://pygmt-git-stringioinput.gmt.vercel.app/api/generated/pygmt.Figure.legend.html#pygmt.Figure.legend.

Happy to receive any feedback before we merge this. If it's good, we might get this in for the upcoming PyGMT v0.2.0 release :grin:

seisman:
`io.StringIO` is something new to me. Some googling gives the impression that `io.StringIO` is one of the [file-like objects](https://docs.python.org/3/glossary.html#term-file-object). Perhaps we should generalize the function to support all file-like objects?

weiji14:
> `io.StringIO` is something new to me. Some googling gives the impression that `io.StringIO` is one of the [file-like objects](https://docs.python.org/3/glossary.html#term-file-object). Perhaps we should generalize the function to support all file-like objects?

I'm a pretty new user of `io.StringIO` too. I'm aware of [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO) which is a binary file-like object, but I'm not quite sure if it is commonly used and/or whether it's worth supporting. Maybe @dr-glenn knows more.

To be fair, we'll probably need to refactor this hacky implementation (writing to a tempfile and having GMT read that file) to use [`GMT_Read_Data`](https://docs.generic-mapping-tools.org/latest/api.html#import-from-a-file-stream-or-handle) or something else at some point. Not to say that this PR isn't useful, just that we can add these 'enhancements' in one step at a time.

seisman:
> To be fair, we'll probably need to refactor this hacky implementation (writing to a tempfile and having GMT read that file) to use [`GMT_Read_Data`](https://docs.generic-mapping-tools.org/latest/api.html#import-from-a-file-stream-or-handle) or something else at some point.

Yes, that's another point. Without this PR, users just need to add a few lines of codes (maybe two or three) to write strings into a temporary file, and pass the file name to legend(), text() or any other modules. But here you add several lines of codes to support it, and these codes are most likely to be removed in the future. Not sure if it's really worth the complexity.

weiji14:
> Yes, that's another point. Without this PR, users just need to add a few lines of codes (maybe two or three) to write strings into a temporary file, and pass the file name to legend(), text() or any other modules. But here you add several lines of codes to support it, and these codes are most likely to be removed in the future. Not sure if it's really worth the complexity.

Fair point, the code here is really just a convenience function to simplify maybe 4-5 lines of code:

https://github.com/GenericMappingTools/pygmt/blob/06ee891fc5d8ba6b39443b3f618abb4d420980b0/pygmt/helpers/tempfile.py#L150-L155

As for whether it's worth it, there are some [GMT Supplemental Modules](https://docs.generic-mapping-tools.org/latest/modules.html#supplemental-modules) that don't quite support virtualfiles yet (e.g. `x2sys`, as mentioned in https://github.com/GenericMappingTools/pygmt/pull/546#discussion_r459553795), and having this `tempfile_from_buffer` helper function in the backend would be useful in the interim months to pass in `pandas.DataFrame` inputs to these modules.

seisman:
This PR will be superseded by PR #3326 and #3438. Closing.

PR #580, opened 2020-09-04
Add a test to make sure shift_origin does not crash

### LINKED ISSUE #514: Figure.shift_origin() does not work
The method shift origin does not work. This simple example never terminates in ipython

```
import pygmt
fig = pygmt.Figure()
fig.shift_origin(xshift='1i', yshift='2i')
```

In a jupyter notebook, using shift origin instead kills the kernel and provides this message:

```
gmtinfo [ERROR]: No input data found!
```

In addition to this, there is a broken link in the [web documentation](https://www.pygmt.org/latest/api/generated/pygmt.Figure.shift_origin.html#pygmt.Figure.shift_origin) to a this page in [gmt](https://docs.generic-mapping-tools.org/latest/GMT_Docs.html#plot-positioning-and-layout-the-x-y-options):

pygmt: last commit 
gmt: 6.1

seisman:

Call `shift_origin()` before setting **-J** and **-R**. It crashes for GMT<=6.1.0,
but works after GMT>=6.1.1.

Closes #514.

**Reminders**

### DIFF
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -121,6 +121,9 @@ def test_figure_show():
 def test_shift_origin():
     "Test if fig.shift_origin works"
     fig = Figure()
+    # First call shift_origin without -J & -R.
+    # Test the issue https://github.com/GenericMappingTools/pygmt/issues/514
+    fig.shift_origin(xshift="2i", yshift="3i")
     fig.basemap(R="10/70/-300/300", J="X3i/5i", B="af")
     fig.shift_origin(xshift="4i")
     fig.basemap(R="10/70/-300/300", J="X3i/5i", B="af")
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_figure.py
```diff
@@ -121,6 +121,9 @@ def test_figure_show():
 def test_shift_origin():
     "Test if fig.shift_origin works"
     fig = Figure()
+    # First call shift_origin without -J & -R.
```
weiji14, line 124:
```suggestion
    # First call shift_origin without projection and region.
```

Use long alias here? Maybe change the whole test to use projection and region instead of J and R, also frame instead of B.

seisman: Good point. I also fixed the short arguments in other functions.

weiji14 (approved):

PR #577, opened 2020-09-02
Bump the required minimum GMT version to 6.1.1

weiji14:
**Description of proposed changes**

Bumps [gmt](https://github.com/GenericMappingTools/gmt) from 6.1.0 to 6.1.1.
  - [Release notes](https://github.com/GenericMappingTools/gmt/releases)
  - [Commits](https://github.com/GenericMappingTools/gmt/compare/6.1.0...6.1.1)

TODO:
- [x] Bump the required minimum GMT version to 6.1.1
- [x] Remove pytest xfails that were set to previous 6.1.0 version
- [ ] Remove compatibility codes for GMT 6.1.0 (if any)
- [ ] etc

Issues marked [upstream](https://github.com/GenericMappingTools/pygmt/labels/upstream) that are resolved in GMT 6.1.1 will be handled in separate Pull Requests.

Supersedes #507

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #560, opened 2020-08-10
Ensure plotting xarray grids with different central meridians work on some projections

weiji14:
**Description of proposed changes**

Adds tests to make sure that plotting `xarray.DataArray` grids at different central meridians work properly for different projection types. See https://github.com/GenericMappingTools/pygmt/issues/515#issuecomment-668834119 for background context. Note that this depends on GMT >= 6.1.1 and won't work yet.

Reverts the `runfirst` workaround required in #476, reverts #531 grdtrack gallery example crash on macOS due to a problematic meridian setting.

Note that this PR *may* depend on #555 :heavy_check_mark:

Addresses #390, Fixes #515

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #585, opened 2020-09-05
Revert plotting across dateline workarounds

weiji14:

Remove some workarounds that were added because plotting grids across the dateline would crash, not needed since we've bumped the minimum required GMT version to 6.1.1 at #577. These are cherrypicked commits from #560 to keep diff small. Specifically, this PR will:

- Remove the `runfirst` workaround added in #476, so there won't be a two-step `make test` in our Makefile anymore
- Remove grdtrack gallery example workaround (due to a segfault crash) added in #531.
- Unregister setup.cfg pytest markers `runfirst`, and also 'mpl_image_compare` added in #323 (since the issue has been fixed upstream at https://github.com/matplotlib/pytest-mpl/pull/83)

Fixes #

**Reminders**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,11 +29,7 @@ test:
 	@echo ""
 	@cd $(TESTDIR); python -c "import $(PROJECT); $(PROJECT).show_versions()"
 	@echo ""
-	# There are two steps to the test here because `test_grdimage_over_dateline`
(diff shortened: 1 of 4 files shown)


seisman (approved):
Looks good.

Issue #584, opened 2020-09-05
Make test suite more efficient to run in 10 minutes


weiji14:
**Description of the desired feature**

PyGMT's test suite is growing, and in the long-term, we'll need to make testing a bit more efficient for contributors. As a goal, we should aim for ~10 minutes for the whole test suite (Windows takes up ~\~30min~ 15min now (see #586)), and <1 minute for local tests on just one module.

Here's a few ways we can tackle this (note to maintainers, feel free to edit this top post):

## Local tests

Running `make test` can be slow since it runs everything, but sometimes we only want to test one function (e.g. `grdimage`). 

- [x] E.g. at https://github.com/GenericMappingTools/pygmt/pull/516#discussion_r464184518, we should document that contributors can run just one test using something like `pytest -v --mpl pygmt/tests/test_grdimage.py`.
- [ ] Alternatively, look into something like [`pytest-fastest`](https://pytest-fastest.readthedocs.io/en/latest/) to automate testing **only** the code that has changed between the current branch and master.

### Installing dependencies

The bulk of the CI time is spent installing dependencies from conda (4min on Linux, 9min on Windows). We could:

- [x] Cache dependencies. This was experimented before, but it didn't seem to help (see https://github.com/GenericMappingTools/pygmt/pull/475#issuecomment-656392670), could try again to see things have changed
- [x] Try switching to [`mamba`](https://github.com/TheSnakePit/mamba) which is a fast reimplementation of `conda` written in C++. I've tried it before, but it's still in experimental mode at setup-miniconda, see https://github.com/conda-incubator/setup-miniconda#example-6-mamba (#841)

### Speed up full test suite

Find a way to run a minimal set of tests when Pull Request is in draft, and only full test suite when Pull Request is ready to review/in review?

**Are you willing to help implement and maintain this feature?** Let's do it together :smile:

willschlitzer:
> Find a way to run a minimal set of tests when Pull Request is in draft, and only full test suite when Pull Request is ready to review/in review?

@weiji14 Would there be a way to just have the test suite run on one machine (presumably Ubuntu, since that typically is completed in the shortest period of time) in one version of Python when the PR is a draft and then run the full suite when it is ready for review?

Also, would reducing the test suite have any impact on how long it takes for the vercel preview deployment, or is that handled separately?

weiji14:
[...]

PR #586, opened 2020-09-05
Cut Windows CI build time in half to 15 min

weiji14:

Attempt at speeding up Windows builds by updating conda version. For Python 3.6, our current Windows build takes **~31min** which is too slow!! Should at least half that time to **~15** minutes to make it more in line with macOS/Linux builds. The bulk of the time (~24min) is spent just installing dependencies using conda, so that will be what this PR will focus on optimizing.

### Before

| Windows (~31min) | macOS (~11min) |
|---|---|
| Windows Python 3.6 CI in 31min | macOS Python 3.6 CI in 11min |

### After

| Windows (~15min) | macOS (~9min) |
|---|---|
| Windows Python 3.6 CI in 15min | macOS Python 3.6 CI in 9min |

TODO:

- [x] Bump setup-miniconda from 1.6.0 to [1.7.0](https://github.com/conda-incubator/setup-miniconda/releases/tag/v1.7.0), also note name change from goanpeca/setup-miniconda to [conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda)
- [x] Try using 'latest' miniconda version [4.8.3](https://github.com/conda/conda/releases/tag/4.8.3) instead of 4.6.14 on Windows
- [ ] Cache dependencies?

Towards #584 

**Reminders**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -69,6 +69,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           channels: conda-forge
+          miniconda-version: "latest"
 
       # Install GMT and other required dependencies from conda-forge
       - name: Install GMT and required dependencies
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -65,7 +65,7 @@ jobs:
 
       # Setup Miniconda
       - name: Setup Miniconda
-        uses: goanpeca/setup-miniconda@v1.6.0
+        uses: conda-incubator/setup-miniconda@v1.7.0
         with:
           python-version: ${{ matrix.python-version }}
           channels: conda-forge


.github/workflows/ci_tests.yaml
```diff
@@ -65,10 +65,11 @@ jobs:
 
       # Setup Miniconda
       - name: Setup Miniconda
-        uses: goanpeca/setup-miniconda@v1.6.0
+        uses: conda-incubator/setup-miniconda@v1.7.0
         with:
           python-version: ${{ matrix.python-version }}
           channels: conda-forge
+          miniconda-version: "latest"
```
weiji14, line 72: I think you'll like this @seisman, The Windows conda install speed up (~24min -> ~8min) comes from just adding this one line! What it does (on Windows) is to update the miniconda version from 4.6.14 to 4.8.3, and this would use a newer/faster `conda` solver, among other things.

seisman (approved):
Great!

.github/workflows/ci_tests.yaml
seisman: It also speed up macOS. Perhaps we should also update `ci_tests_dev.yaml`?

weiji14: Not really by much, I think we can do it another time since that’s not really a bottleneck.

PR #588, opened 2020-09-06
Improve docstring of data_kind() to include xarray grid

seisman:

Improve the docstrings of `data_kind()` and add one more doctest.

**Reminders**

- [X] Add tests for new features or tests that would have caught the bug that you're fixing.
- [X] Write detailed docstrings for all functions/methods.


weiji14 (approved):
This is good. Just one suggestion to change the title to something like: "Improve docstring of data_kind() to include xarray grid".

Also, this reminds me that I'll need to do the same for the 'buffer' data_kind docstring at #576.

PR #589, opened 2020-09-07
Remove expected failures on grdview tests

weiji14:

Removes the expected failures (xfail) pytest marks used as a workaround in #503 to deal with `grdview` failures.

Part 1 to fixing #451. Part 2 will be to deal with xfails on `makecpt` and other `grd*` modules.

**Reminders**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ MANIFEST
 .pytest_cache/
 .ipynb_checkpoints/
 .vscode/
+result_images/
 tmp-test-dir-with-unique-name/
 coverage.xml
 htmlcov
--- a/pygmt/tests/test_grdview.py
+++ b/pygmt/tests/test_grdview.py
@@ -25,12 +25,15 @@ def fixture_grid(region):
 
 
 @check_figures_equal()
-def test_grdview_grid_dataarray(grid, region, fig_ref, fig_test):
+def test_grdview_grid_dataarray(grid, region):
     """
     Run grdview by passing in a grid as an xarray.DataArray.
     """
+    fig_ref = Figure()
     fig_ref.grdview(grid="@earth_relief_01d_g", region=region)
+    fig_test = Figure()
     fig_test.grdview(grid=grid)
+    return fig_ref, fig_test
 
 
 @pytest.mark.mpl_image_compare
@@ -58,14 +61,17 @@ def test_grdview_wrong_kind_of_grid(grid):
 
 
 @check_figures_equal()
-def test_grdview_with_perspective(grid, region, fig_ref, fig_test):
+def test_grdview_with_perspective(grid, region):
     """
     Run grdview by passing in a grid and setting a perspective viewpoint with
     an azimuth from the SouthEast and an elevation angle 15 degrees from the
     z-plane.
     """
+    fig_ref = Figure()
     fig_ref.grdview(grid="@earth_relief_01d_g", region=region, perspective=[135, 15])
+    fig_test = Figure()
     fig_test.grdview(grid=grid, perspective=[135, 15])
+    return fig_ref, fig_test
 
 
 @pytest.mark.xfail(
--- a/pygmt/tests/test_grdview.py
+++ b/pygmt/tests/test_grdview.py
@@ -7,38 +7,41 @@ from .. import Figure, which
 from ..datasets import load_earth_relief
 from ..exceptions import GMTInvalidInput
 from ..helpers import data_kind
+from ..helpers.testing import check_figures_equal
+
+
+@pytest.fixture(scope="module", name="region")
+def fixture_region():
+    "Test region as lonmin, lonmax, latmin, latmax"
+    return (-116, -109, -47, -44)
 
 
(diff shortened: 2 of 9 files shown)


pygmt/tests/test_grdview.py
```diff
@@ -57,199 +69,204 @@ def test_grdview_wrong_kind_of_grid(grid):
...
     z-axis frame.
     """
-    fig = Figure()
-    fig.grdview(grid=grid, perspective=[225, 30], zscale=0.005, frame="zaf")
-    return fig
+    fig_ref = Figure()
+    fig_ref.grdview(grid=gridfile, perspective=[225, 30], zscale=0.005, frame="zaf")
+    fig_test = Figure()
+    fig_test.grdview(grid=grid, perspective=[225, 30], zscale=0.005, frame="zaf")
+    return fig_ref, fig_test
```
weiji14, line 197: This ` test_grdview_with_perspective_and_zaxis_frame` is a very strange one. It passes when the full test suite is ran using `make test`, but was failing locally for me everytime I ran `pytest --mpl --verbose --doctest-modules pygmt/tests/test_grdview.py`. Even weirder is that the problem seems to be on the NetCDF plot rather than the xarray one (now I've seen everything)! The difference is a doubling on the x and y-axis frame, will report this to upstream GMT in a bit:

| NetCDF (expected, but wrong) | xarray.DataArray (test, correct?!!) |
|---|---|
| test_grdview_with_perspective_and_zaxis_frame-expected| test_grdview_with_perspective_and_zaxis_frame |

Exact error is:
```
matplotlib.testing.exceptions.ImageComparisonFailure: Image sizes do not match expected size: (1534, 2664, 3) actual size (1521, 2638, 3)
```

seisman:
Just two suggestions:

1. perhaps rename `grid` to `xgrid`. `xgrid` is more clear that it's a xarray grid.
2. The codes for the reference and test images are the same, except the input is different. So we have a lot of duplicated codes. I'm wondering if we could rewrite the tests like:
```python
fig = [None, None]
for i, grid in enumerate((gridfile, xgrid)):
    fig[i] = Figure()
    fig[i].grdview(grid=gridfile, perspective=[135, 15])
return fig
# return fig[0], fig[1]

```

pygmt/tests/test_grdview.py
seisman: The xarray.DataArray result is "incorrect", because GMT takes the input xarray grid as a Cartesian grid, although `grid.gmt.gtype=1`.

weiji14: Ok, so there's several layers of 'wrong-ness' here it seems? The NetCDF grdview plot is wrong because it has a double frame (without the zebra stripes), and the xarray grid is wrong because it's using the Cartesian plain map frame? But these are 'absolute' wrongs i.e. bugs, and our new `@check_figures_equal` decorator only does a 'relative' comparison to ensure we're reproducing GMT plots in PyGMT (be it right or wrong).

What I don't understand is why the tests pass when ran using `make test` (false negative), but fails with `pytest --mpl --verbose --doctest-modules pygmt/tests/test_grdview.py`. Seems like another flaky test issue.

pygmt/tests/test_grdview.py
```diff
@@ -3,262 +3,258 @@
...
+    projection = f"T{(region[0]+region[1])/2}/{abs((region[2]+region[3])/2)}"
+    kwargs = dict(
+        projection=projection,
+        perspective=[225, 30],
+        zscale=0.005,
+        frame=["xaf", "yaf", "zaf"],
+    )
+    fig_ref.grdview(grid=gridfile, **kwargs)
+    fig_test.grdview(grid=xrgrid, **kwargs)
+    return fig_ref, fig_test
```
weiji14, line 193: Ok, added a projection (Tranverse Mercator) from the suggestions at https://github.com/GenericMappingTools/gmt/issues/4181, and this is the new reference/test plot:

test_grdview_with_perspective_and_zaxis_frame

Tests pass when using both `make test` and `pytest --verbose --mpl --doctest-modules pygmt/tests/test_grdview.py` now :tada:

.gitignore
```diff
@@ -11,6 +11,7 @@ MANIFEST
 .pytest_cache/
 .ipynb_checkpoints/
 .vscode/
+result_images/
```
seisman, line 14: One rule in the [contributing guidelines](https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#general-guidelines) is:

> Do not commit changes to files that are irrelevant to your feature or bugfix (eg: .gitignore, IDE project files, etc).

Perhaps better to open a separate PR for the changes in `.gitignore` and `Makefile`?

pygmt/tests/test_grdview.py
seisman: Yes, it puzzles me, too.

.gitignore
weiji14: Ok, will cherry pick that commit into a separate PR.

seisman (approved):

PR #591, opened 2020-09-08
Allow passing in pandas dataframes to x2sys_cross 

weiji14:

Run crossover analysis directly on `pandas.DataFrame` inputs instead of having to write to tab-separated value (TSV) files first!

Example code:

```python
import pygmt
from tempfile import TemporaryDirectory

dataframe: pd.DataFrame = pygmt.datasets.load_sample_bathymetry()
dataframe.columns = ["x", "y", "z"]  # longitude, latitude, bathymetry

os.environ["X2SYS_HOME"] = os.getcwd()

with TemporaryDirectory(prefix="X2SYS", dir=os.environ["X2SYS_HOME"]) as tmpdir:
    tag = os.path.basename(tmpdir)
    pygmt.x2sys_init(tag=tag, fmtfile="xyz", suffix="xyz", force=True)
    output: pd.DataFrame = pygmt.x2sys_cross(tracks=[dataframe], tag=tag, coe="i", verbose="i")
```

This isn't a trivial thing to implement, because:
-  `x2sys` requires those TSV files in quite a specific format (especially for the datetime columns)
-  The TSV files **cannot** be in any arbitrary directory like `/tmp`, it must be stored either in the current working directory or in specific locations listed in the TAG_paths.txt file.

Support for pandas DataFrame inputs into [`x2sys_cross`](https://www.pygmt.org/dev/api/generated/pygmt.x2sys_cross.html) was originally left out in the original implementation at #546 because we wanted to wait for https://github.com/GenericMappingTools/gmt/issues/3717. But seeing as it's not a trivial matter, this is an interim solution.

Fixes #

**Reminders**

### DIFF
--- a/pygmt/x2sys.py
+++ b/pygmt/x2sys.py
@@ -24,8 +24,23 @@ from .helpers import (
 @contextlib.contextmanager
 def tempfile_from_dftrack(track, suffix):
     """
-    Saves a pandas.DataFrame track table to a temporary tab-separated ASCII
-    text file with a suffix (e.g. 'xyz').
+    Saves pandas.DataFrame track table to a temporary tab-separated ASCII text
+    file with a unique name (to prevent clashes when running x2sys_cross),
+    adding a suffix extension to the end.
+
+    Parameters
+    ----------
+    track : pandas.DataFrame
+        A table holding track data with coordinate (x, y) or (lon, lat) values,
+        and (optionally) time (t).
+    suffix : str
+        File extension, e.g. xyz, tsv, etc.
+
+    Yields
+    ------
+    tmpfilename : str
+        A temporary tab-separated value file with a unique name holding the
+        track data. E.g. 'track-1a2b3c4.tsv'.
     """
     try:
         tmpfilename = f"track-{unique_name()[:7]}.{suffix}"
@@ -180,9 +195,10 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
 
     Parameters
     ----------
-    tracks : str or list
+    tracks : pandas.DataFrame or str or list
         A table or a list of tables with (x, y) or (lon, lat) values in the
-        first two columns. Supported formats are ASCII, native binary, or
+        first two columns. Track(s) can be provided as pandas DataFrame tables
+        or file names. Supported file formats are ASCII, native binary, or
         COARDS netCDF 1-D data. More columns may also be present.
 
         If the filenames are missing their file extension, we will append the
--- a/pygmt/x2sys.py
+++ b/pygmt/x2sys.py
@@ -27,6 +27,7 @@ def tempfile_from_dftrack(track, suffix):
     text file with a suffix (e.g. 'xyz').
     """
     with GMTTempFile(suffix=suffix) as tmpfile:
+        tmpfile.close()  # close the file stream
         track.to_csv(
             path_or_buf=tmpfile.name,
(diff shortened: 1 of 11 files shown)


pygmt/x2sys.py
```diff
@@ -263,8 +286,23 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
+                    .read_text()
+                    .strip()
+                    .split("\n")[-1]
+                )  # e.g. "-Dxyz -Etsv -I1/1"
+                try:
+                    # 1st try to match file extension after -E
+                    suffix = re.search(pattern=r"-E(\S*)", string=lastline).group(1)
+                except AttributeError:  # 'NoneType' object has no attribute 'group'
+                    # 2nd try to match file extension after -D
+                    suffix = re.search(pattern=r"-D(\S*)", string=lastline).group(1)
```
weiji14, line 302: Is there a better way to check if `-Exyz` is in the string, and if not, fallback to parsing from `Dxyz`?

seisman: What about this one?
```python
lastline = "-Dxyz -Etsv -I1/1"
#lastline = "-Dxyz -I1/1"

for item in lastline.split():
    for key in ['-E', '-D']:
        if item.startswith(key):
            suffix = item[2:]
            break
print(suffix)
```

weiji14: Hmm, this gives me some ideas. I'll play around with it, thanks!

pygmt/x2sys.py
```diff
@@ -14,10 +16,30 @@
...
+        tmpfilename = f"track-{unique_name()[:7]}.{suffix}"
+        track.to_csv(
+            path_or_buf=tmpfilename,
+            sep="\t",
+            index=False,
+            date_format="%Y-%m-%dT%H:%M:%S.%fZ",
+        )
+        yield tmpfilename
+    finally:
+        os.remove(tmpfilename)
```
weiji14, line 40: The original implementation using GMTTempFile/NamedTemporaryFile didn't work because of some permissions issues (on macOS/Windows), which is why this try-finally block is used.

weiji14 (commented):
Ok ready for review! I understand that this isn't easy to review properly, but would really appreciate getting this in for v0.2.0 tomorrow as I'll be using it for my PhD research. There's two unit tests added, one for internal crossovers (i.e. on 1 track) and one for external crossovers (2 tracks). I'll add a tutorial example for this over the weekend to explain things better.

pygmt/x2sys.py
seisman: The code quality looks good. As you're the one who develops and uses these functions, we have to trust you. 😄 

Just one suggestion, add the comment to the codes, explaining why you use `unique_name` here. 

That's the first question when I read your codes before I see your comment here.

seisman (approved):

pygmt/x2sys.py
weiji14: > The code quality looks good. As you're the one who develops and uses these functions, we have to trust you. smile

It's all Paul's work done a decade ago, I'm just wrapping it in Python so more people can use it easily :smiley: You won't believe how many 'crossover analysis' tools have been written again and again, but that's another story.

> Just one suggestion, add the comment to the codes, explaining why you use `unique_name` here.
> 
> That's the first question when I read your codes before I see your comment here.

Ok, will do.

PR #603, opened 2020-09-12
Refactor xfail tests to avoid storing baseline images

seisman:

This PR refactors the xfail/fail tests to let the tests pass and avoid storing baseline images.

The baseline images are generated by calling the same modules using plain text or grid files are input, and using single-character arguments. 

Address #451.

**Remaining issues:**

- [ ] The two xfailing tests in `test_config` are not refactored. There is no easy way to generate references images for them.

**TODOs for future PRs:**

- [ ] Tests that are passing are not refactored. These tests still use the images in the `pygmt/test/baselines` directory as baselines.
- [ ] Images of those xfailing/failing tests are no longer needed, but this PR doesn't delete them from the repository. 

**Reminders**

### DIFF
--- a/pygmt/tests/test_basemap.py
+++ b/pygmt/tests/test_basemap.py
@@ -4,6 +4,7 @@ Tests fig.basemap.
 import pytest
 
 from .. import Figure
+from ..helpers.testing import check_figures_equal
 from ..exceptions import GMTInvalidInput
 
 
@@ -54,15 +55,15 @@ def test_basemap_power_axis():
     return fig
 
 
-@pytest.mark.xfail(
-    reason="Baseline image not updated to use earth relief grid in GMT 6.1.0",
-)
-@pytest.mark.mpl_image_compare
+@check_figures_equal()
 def test_basemap_polar():
     "Create a polar basemap plot"
-    fig = Figure()
-    fig.basemap(R="0/360/0/1000", J="P6i", B="afg")
-    return fig
+    fig_ref, fig_test = Figure(), Figure()
+    # Use single-character arguments for the reference image
+    fig_ref.basemap(R="0/360/0/1000", J="P6i", B="afg")
+    fig_test.basemap(region=[0, 360, 0, 1000], projection="P6i", frame="afg")
+
+    return fig_ref, fig_test
 
 
 @pytest.mark.mpl_image_compare
--- a/pygmt/tests/test_grdcontour.py
+++ b/pygmt/tests/test_grdcontour.py
@@ -109,7 +109,7 @@ def test_grdcontour_interval_file_full_opts():
     fig_test.grdcontour(
         **comargs_test, limit=(-25000, -1), pen=["a1p,blue", "c0.5p,blue"]
     )
-    fig_test.grdcontour(**comargs_test, limit="0", pen=["a1p,black", "c0.5p,black"])
+    fig_test.grdcontour(**comargs_test, limit=0, pen=["a1p,black", "c0.5p,black"])
 
     return fig_ref, fig_test
 
--- a/pygmt/tests/test_grdcontour.py
+++ b/pygmt/tests/test_grdcontour.py
@@ -82,14 +82,23 @@ def test_grdcontour_file():
     return fig
 
 
-@pytest.mark.xfail(
-    reason="Baseline image not updated to use earth relief grid in GMT 6.1.0",
-)
-@pytest.mark.mpl_image_compare
+@check_figures_equal()
 def test_grdcontour_interval_file_full_opts():
     """ Plot based on external contour level file """
-    fig = Figure()
-    comargs = {
+    fig_ref, fig_test = Figure(), Figure()
+    # Use single-character arguments for the reference image
+    comargs_ref = {
+        "grid": "@earth_relief_10m",
+        "R": "-161.5/-154/18.5/23",
+        "C": TEST_CONTOUR_FILE,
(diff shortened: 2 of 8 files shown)


pygmt/tests/test_basemap.py
```diff
@@ -54,15 +56,17 @@ def test_basemap_power_axis():
...
     "Create a polar basemap plot"
-    fig = Figure()
-    fig.basemap(R="0/360/0/1000", J="P6i", B="afg")
-    return fig
+    fig_test = Figure()
+    fig_test.basemap(R="0/360/0/1000", J="P6i", B="afg")
+
+    fig_ref = Figure()
+    with Session() as lib:
+        lib.call_module("basemap", "-R0/360/0/1000 -JP6i -Bafg")
```
seisman, line 67: The reference image is created by directly calling the low-level API function `call_module`.

weiji14: I like the concept in general (testing against the low level GMT API), but this can get complicated real fast for some modules that output to a file (see e.g. @liamtoney's example at https://github.com/GenericMappingTools/pygmt/issues/364#issuecomment-623179721 for `grdgradient`).

Should we have a convenience function to wrap around this low-level API better? E.g. something like `pygmt.gmt(module="basemap", arg_str="-R0/360/0/1000 -JP6i -Bafg") that removes the `with` statement.

seisman: > but this can get complicated real fast for some modules that output to a file (see e.g. @liamtoney's example at [#364 (comment)](https://github.com/GenericMappingTools/pygmt/issues/364#issuecomment-623179721) for `grdgradient`).

Most tests are simple, testing only one aspect of functions. https://github.com/GenericMappingTools/pygmt/issues/364#issuecomment-623179721's example may be the most  complicated one, but we won't use virtual files for the references images. 

> Should we have a convenience function to wrap around this low-level API better? E.g. something like `pygmt.gmt(module="basemap", arg_str="-R0/360/0/1000 -JP6i -Bafg") that removes the `with` statement.

I believe @leouieda is against it (https://github.com/GenericMappingTools/pygmt/issues/122#issuecomment-358436386).

weiji14: > Most tests are simple, testing only one aspect of functions. [#364 (comment)](https://github.com/GenericMappingTools/pygmt/issues/364#issuecomment-623179721) example may be the most complicated one, but we won't use virtual files for the references images.

True, but as you mentioned, this way of testing isn't very Pythonic or new contributor friendly. We can definitely think of a better solution that doesn't involve `clib` here, just need a bit more time, and maybe have a look around some other projects like `GMT.jl` and `matplotlib` to see how they do it.

> I believe @leouieda is against it ([#122 (comment)](https://github.com/GenericMappingTools/pygmt/issues/122#issuecomment-358436386)).

Ah yes, I thought I saw this somewhere before, thanks for linking. Definitely hold on the `pygmt.gmt`/`fig.rawcommand` feature then.

weiji14, line 67: Just as a throwaway idea, what about testing that the long aliases and list-to-str conversion works? I.e. have `fig_test` use long aliases like `region`, and have `fig_ref` use the short alias like `R`.

```suggestion
    fig_test = Figure()
    fig_test.basemap(region=[0, 360, 0, 1000], projection="P6i", frame="afg")

    fig_ref = Figure()
    fig_ref.basemap(R="0/360/0/1000", J="P6i", B="afg")
```

seisman: Sounds a clever idea. I believe we are confident that arguments like `R="0/360/0/1000", J="P6i", B="afg"` will be converted to `"-R0/360/0/1000 -JP6i -Bafg"`, thus, `fig_ref.basemap(R="0/360/0/1000", J="P6i", B="afg")` is the same as `lib.call_module("basemap", "-R0/360/0/1000 -JP6i -Bafg")`.

weiji14: Cool, if you want to update this PR to fix the failing 'polar', 'logo' (as mentioned at https://github.com/GenericMappingTools/pygmt/issues/451#issuecomment-708792970) tests, and any other ones that pop up (I saw a few more upstream GMT changes already...), I'll review it later.

seisman: OK. I'll work on this PR.

weiji14 (commented):
A lot to review, but it's looking good! Just a couple of comments.

> The two xfailing tests in test_config are not refactored. There is no easy way to generate references images for them.
>
> Images of those xfailing/failing tests are no longer needed, but this PR doesn't delete them from the repository.

Yes, will need to remember to do these in the future as separate PRs.

pygmt/tests/test_grdcontour.py
```diff
@@ -82,26 +82,36 @@ def test_grdcontour_file():
...
         "interval": TEST_CONTOUR_FILE,
         "grid": "@earth_relief_10m",
         "resample": "100",
         "projection": "M6i",
         "cut": 10,
     }
+    fig_test.grdcontour(
+        **comargs_test, limit=(-25000, -1), pen=["a1p,blue", "c0.5p,blue"]
+    )
+    fig_test.grdcontour(**comargs_test, limit="0", pen=["a1p,black", "c0.5p,black"])
```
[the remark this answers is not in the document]
seisman: Yes, it should work.

weiji14 (approved):

PR #607, opened 2020-09-14
Add an example for scatter plots with auto legends

seisman:

**Preview:** https://pygmt-git-gallery-scatter.gmt.vercel.app/gallery/plot/scatter.html

Add a new example, modified from https://matplotlib.org/gallery/lines_bars_and_markers/scatter_with_legend.html.

image

To be discussed:

- [ ] Is it a good example?
- [ ] Should we have a separate directory/category for "symbols"?
- [x] We should add an alias for **-t** option (#614).

**Reminders**

### DIFF
--- a/examples/gallery/plot/scatter.py
+++ b/examples/gallery/plot/scatter.py
@@ -2,9 +2,11 @@
 Scatter plots with a legend
 ---------------------------
 
-TODO: Add more docstrings.
+To create a scatter plot with a legend one may use a loop and create one scatter
+plot per item to appear in the legend and set the label accordingly.
 
-Modified from the matplotlib example: https://matplotlib.org/gallery/lines_bars_and_markers/scatter_with_legend.html
+Modified from the matplotlib example:
+https://matplotlib.org/gallery/lines_bars_and_markers/scatter_with_legend.html
 """
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


seisman:
>  Is it a good example?

It's a good example for me. It's modified from the matplotlib example, so users can have a direct comparison between pygmt and matplotlib. 

> Should we have a separate directory/category for "symbols"?

Yes, to me. I expect we would have many examples for symbols (different symbol styles, sizes, colors et al.). It's better to group the "symbols" examples together.

examples/gallery/plot/scatter.py
```diff
@@ -0,0 +1,41 @@
...
+        style="c",
+        sizes=sizes,
+        color=color,
+        # Set the legend label,
+        # and set the circle size to be 0.25 cm (+S0.25c) in legend
+        label=f"{color}+S0.25c",
+        transparency=70,  # set transparency level for all symbols
+    )
+
+fig.legend(transparency=30)  # set transparency level for legends
```
weiji14, line 40: Ah, this is cool, I hadn't thought of using transparent legend boxes before!

weiji14 (approved):
> > Is it a good example?
> 
> It's a good example for me. It's modified from the matplotlib example, so users can have a direct comparison between pygmt and matplotlib.

Fair enough. I think our problem now is that there are too few examples, so the more we have the better! Showcasing the transparency in this example is also good!

> > Should we have a separate directory/category for "symbols"?
> 
> Yes to me. I expect we would have many examples for symbols (different symbol styles, sizes, colors et al.). It's better to group the "symbols" examples together.

Maybe we could place it under the user guide, similar to what is being done for projections at https://www.pygmt.org/v0.2.0/projections/index.html? But that can be discussed in another issue/PR.

PR #611, opened 2020-09-17
Improve Makefile to clean __pycache__ directory recursively

seisman:

Running `make clean` only deletes the __pycache__ directory in the
project root. There are still a few __pycache__ directory left in the
subdirectories.

**Reminders**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,8 @@ lint:
 clean:
 	find . -name "*.pyc" -exec rm -v {} \;
 	find . -name "*~" -exec rm -v {} \;
-	rm -rvf build dist MANIFEST *.egg-info __pycache__ .coverage .cache htmlcov coverage.xml
+	find . -type d -name  "__pycache__" -exec rm -rv {} +
+	rm -rvf build dist MANIFEST *.egg-info .coverage .cache htmlcov coverage.xml
 	rm -rvf $(TESTDIR)
 	rm -rvf baseline
 	rm -rvf result_images


seisman:
@weiji14 Just a quick off-topic question: should we use milestone `0.2.1` instead of `0.2.x`? I think `0.2.1` is easier for us and others to track the changes in a single release. If we decide to release `0.3.0` instead of `0.2.1`, it's also much easier to  changes all `0.2.1`-related PRs/issues to `0.3.0`.

weiji14 (approved):
Looks good, didn't realize there were so many nested __pycache__ directories!

One thing that bugs me is that the Makefile commands are very UNIX specific (i.e. it works on Linux/macOS, but not on Windows). But that's for a separate PR.

> @weiji14 Just a quick off-topic question: should we use milestone `0.2.1` instead of `0.2.x`? I think `0.2.1` is easier for us and others to track the changes in a single release. If we decide to release `0.3.0` instead of `0.2.1`, it's also much easier to changes all `0.2.1`-related PRs/issues to `0.3.0`.

No strong preference, happy to go with it (though I do like that milestone v0.2.x is at https://github.com/GenericMappingTools/pygmt/milestone/2 and milestone v0.3.x is at https://github.com/GenericMappingTools/pygmt/milestone/3, but that is bound to break at some point).

seisman:
> One thing that bugs me is that the Makefile commands are very UNIX specific (i.e. it works on Linux/macOS, but not on Windows). But that's for a separate PR.

I believe we will fix it when a Windows developer joins the team.



> No strong preference, happy to go with it (though I do like that milestone v0.2.x is at https://github.com/GenericMappingTools/pygmt/milestone/2 and milestone v0.3.x is at https://github.com/GenericMappingTools/pygmt/milestone/3, but that is bound to break at some point).

OK. I'll change the milestones later.

PR #614, opened 2020-09-17
Add alias -t (transparency) to all plotting functions

seisman:

Add `transparency` (alias of **-t** option) to all plotting modules.

**Reminders**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -82,6 +82,13 @@ COMMON_OPTIONS = {
             ``[g|p]``
             Force output grid to be gridline (g) or pixel (p) node registered.
             Default is gridline (g).""",
+    "t": """\
+        transparency : float
+            Set transparency level, in (0-100] percent range.
+            Default is 0, i.e., opaque.
+            Only visible when PDF or raster format output is selected.
+            Only the PNG format selection adds a transparency layer
+            in the image (for further processing). """,
 }
 
 
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -84,7 +84,7 @@ COMMON_OPTIONS = {
             Default is gridline (g).""",
     "t": """\
         transparency : float
-            Set transparency level, in (0-100] percent range.
+            Set transparency level, in [0-100] percent range.
             Default is 0, i.e., opaque.
             Only visible when PDF or raster format output is selected.
             Only the PNG format selection adds a transparency layer
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -823,7 +823,11 @@ class BasePlotting:
 
     @fmt_docstring
     @use_alias(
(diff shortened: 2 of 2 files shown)


pygmt/helpers/decorators.py
```diff
@@ -82,6 +82,13 @@
             ``[g|p]``
             Force output grid to be gridline (g) or pixel (p) node registered.
             Default is gridline (g).""",
+    "t": """\
+        transparency : float
```
seisman, line 86: `transparency` can also be an array for a few modules, e.g., `plot`, `plot3d` and `text`, so that symbols and texts can have varying transparency. It needs more work and should be addressed in separated PRs.

weiji14: Ok, but could you add at least one unit test for `transparency` on a module that won't have array-like `transparency` inputs (e.g. `image`)?

pygmt/helpers/decorators.py
```diff
@@ -82,6 +82,13 @@
...
+            Set transparency level, in (0-100] percent range.
```
weiji14, line 87:
```suggestion
            Set transparency level, in (0-100) percent range.
```

Actually a typo on upstream GMT too https://docs.generic-mapping-tools.org/6.1/gmt.html#t-full.

weiji14 (commented):
Cool, I haven't actually used `transparency` before, but looks fine. An example would be nice :wink:

pygmt/helpers/decorators.py
seisman: It means adding another baseline image to the repository. 😄

pygmt/helpers/decorators.py
seisman: 100 is a valid value. Actually, even 0 is valid.

weiji14: Ah right, [square bracket notation](https://stackoverflow.com/questions/4396290/what-does-this-square-bracket-and-parenthesis-bracket-notation-mean-first1-last). In this case, use:
```suggestion
            Set transparency level, in [0-100] percent range.
```

pygmt/helpers/decorators.py
weiji14: Maybe just a gallery example then, if not too much work :no_mouth: I suppose we could come up with a smart `check_figures_equal` test but happy to leave that to the next PR.

seisman: There is an upcoming example in https://github.com/GenericMappingTools/pygmt/pull/607, but I'm not sure if the example does too many things (symbols, transparency and legends.).

weiji14: Ah right, missed the transparency part on that example. I don't think we can have too many examples, if anything we have too little right now (a lot of catching up to do with matplotlib at https://matplotlib.org/gallery/index.html)!

weiji14 (approved):

PR #619, opened 2020-09-20
Refactor info to allow datetime inputs from xarray.Dataset and pandas.DataFrame tables

### LINKED ISSUE #597: Running info on pandas.DataFrame with time column doesn't work

Just noticed that datetime columns being passed into `pygmt.info` doesn't work. This follows on from the `pandas.DataFrame` inputs into `pygmt.info` functionality added in #574, see also #464 and #562 where the datetime machinery should be more or less implemented.

```python
import pygmt
import pandas as pd

table = pd.DataFrame(data=[1,3,2,5,4], columns=["z"])
table["time"] = pd.date_range(start="2020-01-01", periods=5)

pygmt.info(table=table)
```

Note that the equivalent `gmt` command does work on datetime inputs.

```
!gmt info temp.txt
temp.txt: N = 5	<1/5>	<2020-01-01T00:00:00/2020-01-05T00:00:00>
```

**Full error message**

```
GMTCLibError                              Traceback (most recent call last)
<ipython-input-6-a9e68c3dc07e> in <module>
----> 1 pygmt.info(table=table)

~/pygmt/pygmt/helpers/decorators.py in new_module(*args, **kwargs)
    235                 if alias in kwargs:
    236                     kwargs[arg] = kwargs.pop(alias)
--> 237             return module_func(*args, **kwargs)
    238 
    239         new_module.aliases = aliases

~/pygmt/pygmt/modules.py in info(table, **kwargs)
    116 
    117         with GMTTempFile() as tmpfile:
--> 118             with file_context as fname:
    119                 arg_str = " ".join(
    120                     [fname, build_arg_string(kwargs), "->" + tmpfile.name]

~/miniconda3/envs/pygmt/lib/python3.7/contextlib.py in __enter__(self)
    110         del self.args, self.kwds, self.func
    111         try:
--> 112             return next(self.gen)
    113         except StopIteration:
    114             raise RuntimeError("generator didn't yield") from None

~/pygmt/pygmt/clib/session.py in virtualfile_from_matrix(self, matrix)
   1268         )
   1269 
-> 1270         self.put_matrix(dataset, matrix)
   1271 
   1272         with self.open_virtual_file(

~/pygmt/pygmt/clib/session.py in put_matrix(self, dataset, matrix, pad)
    906         )
    907         if status != 0:
--> 908             raise GMTCLibError("Failed to put matrix of type {}.".format(matrix.dtype))
    909 
    910     def write_data(self, family, geometry, mode, wesn, output, data):

GMTCLibError: Failed to put matrix of type object.
```

**System information**

```
PyGMT information:
  version: v0.1.2+55.g6deb388
System information:
  python: 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:25:08)  [GCC 7.5.0]
  executable: ~/miniconda3/envs/pygmt/bin/python
  machine: Linux-4.19.0-8-amd64-x86_64-with-debian-10.5
Dependency information:
  numpy: 1.19.1
  pandas: 1.1.1
  xarray: 0.16.0
  netCDF4: 1.5.3
  packaging: 20.4
  ghostscript: 9.27
  gmt: None
GMT library information:
  binary dir: ~/miniconda3/envs/pygmt/bin
  cores: 2
  grid layout: rows
  library path: ~/miniconda3/envs/pygmt/lib/libgmt.so
  padding: 2
  plugin dir: ~/miniconda3/envs/pygmt/lib/gmt/plugins
  share dir: ~/miniconda3/envs/pygmt/share/gmt
  version: 6.1.1
```

weiji14:

In order to support datetime inputs into [`pygmt.info`](https://www.pygmt.org/v0.2.0/api/generated/pygmt.info.html), this PR changes the backend mechanism from using `lib.virtualfile_from_matrix()` (which only supports single non-datetime dtypes) to using `lib.virtualfile_from_vectors()` (which supports datetime inputs as of #464).

**Note**: While datetime inputs can be passed into `info` now, the resulting outputs (UNIX timestamp instead of ISO datetime) might not make sense due to the issue at https://github.com/GenericMappingTools/gmt/issues/4241.

Coincidentally, this refactor also enables the following:

- Properly handle pandas.DataFrame table inputs that may have mixed dtypes (e.g. int32/float32 on different columns, see also #547).
- Passing in [`xarray.Dataset`](http://xarray.pydata.org/en/v0.16.0/data-structures.html#dataset) inputs! Note: Only for `xarray.Dataset`s made up of 1D `xarray.DataArrays`s (e.g. what you might get from using [`xarray.Dataset.from_dataframe`](http://xarray.pydata.org/en/v0.16.0/generated/xarray.Dataset.from_dataframe.html)).

Addresses #597, will be fixed once https://github.com/GenericMappingTools/gmt/issues/4241 is resolved.

**Reminders**

### DIFF
--- a/pygmt/modules.py
+++ b/pygmt/modules.py
@@ -79,9 +79,9 @@ def info(table, **kwargs):
     Parameters
     ----------
     table : str or np.ndarray or pandas.DataFrame or xarray.Dataset
-        Pass in either a a file name to an ASCII data table, a 1D/2D numpy
-        array, a pandas dataframe, or an xarray dataset made up of 1D
-        xarray.DataArray data variables.
+        Pass in either a file name to an ASCII data table, a 1D/2D numpy array,
+        a pandas dataframe, or an xarray dataset made up of 1D xarray.DataArray
+        data variables.
     per_column : bool
         Report the min/max values per column in separate columns.
     spacing : str
--- a/pygmt/modules.py
+++ b/pygmt/modules.py
@@ -108,7 +108,7 @@ def info(table, **kwargs):
             file_context = dummy_context(table)
         elif kind == "matrix":
             try:
-                # pandas.DataFrame types
+                # pandas.DataFrame and xarray.Dataset types
                 arrays = [array for _, array in table.items()]
             except AttributeError:
                 # Python lists, tuples, and numpy ndarray types
--- a/pygmt/modules.py
+++ b/pygmt/modules.py
@@ -78,9 +78,10 @@ def info(table, **kwargs):
 
     Parameters
     ----------
-    table : pandas.DataFrame or np.ndarray or str
-        Either a pandas dataframe, a 1D/2D numpy.ndarray or a file name to an
-        ASCII data table.
+    table : str or np.ndarray or pandas.DataFrame or xarray.Dataset
+        Pass in either a a file name to an ASCII data table, a 1D/2D numpy
+        array, a pandas dataframe, or an xarray dataset made up of 1D
+        xarray.DataArray data variables.
     per_column : bool
         Report the min/max values per column in separate columns.
     spacing : str
--- a/pygmt/modules.py
+++ b/pygmt/modules.py
@@ -107,10 +107,13 @@ def info(table, **kwargs):
         if kind == "file":
             file_context = dummy_context(table)
         elif kind == "matrix":
-            _table = np.asanyarray(table)
-            if table.ndim == 1:  # 1D arrays need to be 2D and transposed
-                _table = np.transpose(np.atleast_2d(_table))
-            file_context = lib.virtualfile_from_matrix(_table)
+            try:
+                # pandas.DataFrame types
+                arrays = [array for _, array in table.items()]
+            except AttributeError:
+                # Python lists, tuples, and numpy ndarray types
+                arrays = np.atleast_2d(np.asanyarray(table).T)
+            file_context = lib.virtualfile_from_vectors(*arrays)
         else:
             raise GMTInvalidInput(f"Unrecognized data type: {type(table)}")
 
--- a/pygmt/tests/test_info.py
+++ b/pygmt/tests/test_info.py
@@ -8,12 +8,16 @@ import numpy.testing as npt
 import pandas as pd
 import pytest
 import xarray as xr
+from packaging.version import Version
 
-from .. import info
+from .. import clib, info
 from ..exceptions import GMTInvalidInput
 
 TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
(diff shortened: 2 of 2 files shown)


pygmt/tests/test_info.py
```diff
@@ -33,7 +33,38 @@ def test_info_dataframe():
...
+        data={
+            "z": [10, 13, 12, 15, 14],
+            "time": pd.date_range(start="2020-01-01", periods=5),
+        }
+    )
+    output = info(table=table)
+    expected_output = (
+        "<vector memory>: N = 5 <10/15> <2020-01-01T00:00:00/2020-01-05T00:00:00>\n"
+    )
+    assert output == expected_output
```
weiji14, line 53: Note that this test currently fails, giving what looks like a UNIX timestamp rather than a ISO datetime output:

```
E       AssertionError: assert '<vector memo...1578182400>\n' == '<vector memo...5T00:00:00>\n'
E         - <vector memory>: N = 5 <10/15> <2020-01-01T00:00:00/2020-01-05T00:00:00>
E         + <vector memory>: N = 5 <10/15> <1577836800/1578182400>
```

Will need to investigate, but any idea why this is happening?

seisman: GMT_Put_Vector started to accept datetime strings in PR https://github.com/GenericMappingTools/gmt/pull/3396. In this implementation, all datetime strings are converted to `double` internally.

Looks like a GMT API bug to me.

weiji14: So the inputs are read correctly here, we can pass in datetime types using the `GMT_DATETIME` enum. The output is also the correct UNIX timestamp range, but I doubt it's something users would expect.

I suppose we can `xfail` this for 6.1.1, but would prefer to have this reported and/or fixed in GMT master first (i.e. for 6.2.0). A workaround might be to use `np.datetime64(1577836800, 's')` to get `numpy.datetime64('2020-01-01T00:00:00')`, but this isn't a nice solution.

seisman: > I suppose we can `xfail` this for 6.1.1,

Yes to me.

pygmt/modules.py
```diff
@@ -78,9 +78,10 @@ def info(table, **kwargs):
 
     Parameters
     ----------
-    table : pandas.DataFrame or np.ndarray or str
-        Either a pandas dataframe, a 1D/2D numpy.ndarray or a file name to an
-        ASCII data table.
+    table : str or np.ndarray or pandas.DataFrame or xarray.Dataset
+        Pass in either a a file name to an ASCII data table, a 1D/2D numpy
```
seisman, line 82:
```suggestion
        Pass in either a file name to an ASCII data table, a 1D/2D numpy
```

seisman (approved):

Issue #615, opened 2020-09-18
Varying transparency levels for plot, plot3d and text


seisman:
**Description of the desired feature**

#614 added the `transparency` argument to all plotting functions. In most functions, `transparency` can be an int/float value in 0-100. For a few functions (plot, plot3d and text), transparency can be a list to have varying transparency for symbols and texts.

We should extend these functions to support the varying transparency feature.

**Are you willing to help implement and maintain this feature?** Yes

weiji14:
Record by record transparency has been added to `plot` at #626, and also `plot3d` at #471, so what's left is just adding it to `text`?

On a side note, we might want to add a test for `grdimage` with transparency. I'm getting this error on `PyGMT` v0.2.1, `GMT` 6.2.0_92c5676_2020.11.10 () with `ghostscript` 9.53.3:

```python
fig = pygmt.Figure()
fig.grdimage(grid="@earth_relief_01d", transparency=70)
fig.show()
```

```python-traceback
GMTCLibError: Module 'psconvert' failed with status code 79:
psconvert [ERROR]: System call [gs -q -dNOPAUSE -dBATCH -dNOSAFER -dPDFSETTINGS=/prepress -dDownsampleColorImages=false -dDownsampleGrayImages=false -dDownsampleMonoImages=false -dUseFlateCompression=true -dEmbedAllFonts=true -dSubsetFonts=true -dMonoImageFilter=/FlateEncode -dAutoFilterGrayImages=false -dGrayImageFilter=/FlateEncode -dAutoFilterColorImages=false -dColorImageFilter=/FlateEncode -dSCANCONVERTERTYPE=2 -dALLOWPSTRANSPARENCY -dMaxBitmap=2147483647 -dUseFastColor=true -dGraphicsAlphaBits=1 -dTextAlphaBits=1 -sDEVICE=pdfwrite  -g1772x886 -r300 -sOutputFile='/home/user/.gmt/sessions/gmt_session.62321/gmt_13_intermediate.pdf' '/home/user/.gmt/sessions/gmt_session.62321/psconvert_62321d.eps'] returned error 256.
                      

```

Will submit a proper bug report once I can isolate the issue.

seisman:
> Record by record transparency has been added to `plot` at #626, and also `plot3d` at #471, so what's left is just adding it to `text`?

Yes, I think I already have a local branch for the `text` transparency. Will check if it works as expected and submit later.

[...]

PR #623, opened 2020-09-20
Test if is_nonstr_iter works for numpy arrays

seisman:

Add a doctest checking if function `is_nonstr_iter` works for numpy arrays.

**Reminders**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -172,6 +172,9 @@ def is_nonstr_iter(value):
     True
     >>> is_nonstr_iter((1, 2, 3))
     True
+    >>> import numpy as np
+    >>> is_nonstr_iter(np.array([1.0, 2.0, 3.0]))
+    True
 
     """
     return isinstance(value, Iterable) and not isinstance(value, str)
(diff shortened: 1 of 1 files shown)


pygmt/helpers/utils.py
```diff
@@ -172,6 +172,9 @@ def is_nonstr_iter(value):
     True
     >>> is_nonstr_iter((1, 2, 3))
     True
+    >>> import numpy as np
+    >>> is_nonstr_iter(np.array([1.0, 2.0, 3.0]))
+    True
```
weiji14, line 177: Do you want to test with `np.array(["abc", "def", "ghi"])`? :laughing:

By the way, what's the reason for adding this test?

seisman: > Do you want to test with `np.array(["abc", "def", "ghi"])`?

I think yes. It should be `True`.

> By the way, what's the reason for adding this test?

Trying to implement the feature of varying transparency level (#615). `transparency` can be int, float or list/array.

I need to check if `transparency` is an list/array. If yes, then append it to the `extra_arrays`, which will be passed via `GMT_Put_Vector`.

weiji14: Ah I see, looks ok then.

weiji14 (approved):

weiji14:
By the way, maybe add the `skip-changelog` label since this is a pretty small change.

PR #637, opened 2020-09-29
Refactor three xfail grdcontour tests

seisman:

Refractor three grdcontour tests using the `@check_figures_equal` decorator so that they no longer xfail.

**Reminders**

### DIFF
--- a/pygmt/tests/test_grdcontour.py
+++ b/pygmt/tests/test_grdcontour.py
@@ -6,10 +6,9 @@ import os
 import numpy as np
 import pytest
 
-from .. import Figure, grdcut
+from .. import Figure
 from ..exceptions import GMTInvalidInput
 from ..datasets import load_earth_relief
-from ..helpers import GMTTempFile
 from ..helpers.testing import check_figures_equal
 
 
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_grdcontour.py
```diff
@@ -21,49 +23,53 @@ def fixture_grid():
...
-    return fig
+    fig_test.grdcontour(grid_, interval="1000", projection="M6i")
+
+    with GMTTempFile(suffix=".nc") as tmpgrid:
+        grdcut(
+            grid="@earth_relief_01d_g",
+            region=[-180, 180, -30, 30],
+            outgrid=tmpgrid.name,
+        )
+        fig_ref.grdcontour(tmpgrid.name, interval="1000", projection="M6i")
```
weiji14, line 71:
```suggestion
    kwargs = dict(interval="1000", projection="M6i")
    fig_ref.grdcontour(
        grid="@earth_relief_01d_g", region=[-180, 180, -30, 30], **kwargs
    )
    fig_test.grdcontour(grid=grid_, **kwargs)
```
`grdcut` and `GMTTempFile` won't be needed here :)

pygmt/tests/test_grdcontour.py
```diff
@@ -6,9 +6,11 @@
 import numpy as np
 import pytest
 
-from .. import Figure
+from .. import Figure, grdcut
 from ..exceptions import GMTInvalidInput
 from ..datasets import load_earth_relief
+from ..helpers import GMTTempFile
```
weiji14, line 12:
```suggestion
from .. import Figure
from ..exceptions import GMTInvalidInput
from ..datasets import load_earth_relief
```

weiji14 (commented):
Thanks for keeping on top of this (sorry that I haven't been keeping up much, got a bit busy with PhD work again)! Just one suggestion to simplify the refactor.

Also, there's another xfail for `test_grdcontour_interval_file_full_opts` that wasn't removed, is there a reason to leave it?

seisman:
> there's another xfail for `test_grdcontour_interval_file_full_opts` that wasn't removed, is there a reason to leave it?

This test uses the grid file `@earth_relief_10m`, so we can't rewrite the test using `@check_figure_equal()` decorator. To remove the xfail label, we need to update the baseline image, but we're trying to avoid it before we have a better solution.

weiji14:
> This test uses the grid file `@earth_relief_10m`, so we can't rewrite the test using `@check_figure_equal()` decorator. To remove the xfail label, we need to update the baseline image, but we're trying to avoid it before we have a better solution.

Yeah, doesn't look like a trivial task. Happy to leave it for now then.

weiji14 (approved):

PR #639, opened 2020-10-04
Disallow passing arguments like -XNone to GMT

### LINKED ISSUE #380: Offsets "X" and "Y" provide unexpected results when provided
Everything works fine when making a single plot using `grdimage` when the offsets `X` and `Y` are not provided. Here is an example:

test-no-X-Y-flags

However, if I provide the options `X=0, Y=0`, the plot is shifted to the lower left, and the left and bottom labels are cut off the plot:

test-0-0

It appears that the left and bottom of the image are perfectly aligned with the bottom of the plot, suggesting that the size of the left and bottom labels has not been taken into account correctly.

seisman:

Arguments like `-XNone` are invalid for GMT.

This PR updates the function `build_arg_string` and checks if any value is `None`.
If yes, remove the argument.

**Limitations**:

Invalid arguments like `{'X': [None, None, None]}` are still not caught and
converted to `-XNone -XNone -XNone`, but it should be uncommon and more likely
a user error.

Fixes #380.

**Reminders**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -127,7 +127,8 @@ def build_arg_string(kwargs):
     Examples
     --------
 
-    >>> print(build_arg_string(dict(R='1/2/3/4', J="X4i", P='', E=200)))
+    >>> print(build_arg_string(dict(R='1/2/3/4', J="X4i", P='',
+    ...                             E=200, X=None, Y=None)))
     -E200 -JX4i -P -R1/2/3/4
     >>> print(build_arg_string(dict(R='1/2/3/4', J="X4i",
     ...                             B=['xaf', 'yaf', 'WSen'],
@@ -140,6 +141,8 @@ def build_arg_string(kwargs):
         if is_nonstr_iter(kwargs[key]):
             for value in kwargs[key]:
                 sorted_args.append("-{}{}".format(key, value))
+        elif kwargs[key] is None:  # arguments like -XNone are invalid
+            continue
         else:
             sorted_args.append("-{}{}".format(key, kwargs[key]))
 
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -127,7 +127,11 @@ def build_arg_string(kwargs):
     Examples
     --------
 
-    >>> print(build_arg_string(dict(R="1/2/3/4", J="X4i", P="", E=200, X=None, Y=None)))
+    >>> print(
+    ...     build_arg_string(
+    ...         dict(R="1/2/3/4", J="X4i", P="", E=200, X=None, Y=None)
+    ...     )
+    ... )
     -E200 -JX4i -P -R1/2/3/4
     >>> print(
     ...     build_arg_string(


seisman:
@MarkWieczorek Please see if this PR works for you.

pygmt/helpers/utils.py
```diff
@@ -147,6 +151,8 @@ def build_arg_string(kwargs):
         if is_nonstr_iter(kwargs[key]):
             for value in kwargs[key]:
                 sorted_args.append("-{}{}".format(key, value))
+        elif kwargs[key] is None:  # arguments like -XNone are invalid
+            continue
```
weiji14, line 155: So `X=None` should return the same result as `X=False`, following what was suggested at https://github.com/GenericMappingTools/pygmt/issues/380#issuecomment-555423909.

Question is, should the `None` detection logic be placed here, or would it be better to put it in `remove_bools` somewhere here:

https://github.com/GenericMappingTools/pygmt/blob/1bf0a305dab83434e1f9cea8ead6af964af47da5/pygmt/helpers/decorators.py#L437-L443

seisman: #640 explains why we can't deal with `None` or even boolean values in the `kwargs_to_strings` decorator.

weiji14: This is very confusing :confused:, so for most PyGMT modules, we go through 1) `kwargs_to_strings` decorator, 2) some PyGMT virtualfile logic or whatever, and then 3) the `build_arg_string` which passes things to GMT C API.

In (1), the `kwargs_to_string` has a `remove_bools` setting that turns `True` into an empty string `""`, are you suggesting in #640 that we should **not** perform that `True` to `""` conversion in (1), but move it to (3) at `build_arg_string` instead? Similar in this PR, `None` has to be handled at (3) in `build_arg_string`.

seisman: Yes, as shown in #640, the boolean argument `nongmt_args` should be `True` or `False`. However, when calling the function, the decorator `kwargs_to_string` converts `nongmt_args=True` to `nongmt_args=""`. It makes sense if `nongmt_args` is an alias of GMT options (e.g., `-N`). However, when `nongmt_args` is just a Python variable, the conversion means checking `nongmt_args` is True or False won't work.

seisman: BTW, I just updated the codes in #640, hopefully, to make the issue easier to understand.

weiji14: So reading #640, does that mean that `True`/`False` inputs will eventually need to go here in `build_arg_strings` as well? It seems weird to have both `kwargs_to_strings` and `build_arg_strings` handling different pieces of Python to GMT logic.

seisman: > does that mean that `True`/`False` inputs will eventually need to go here in `build_arg_strings` as well?

Yes.

> It seems weird to have both `kwargs_to_strings` and `build_arg_strings` handling different pieces of Python to GMT logic.

Yes, as mentioned in #640, handling boolean arguments in `kwargs_to_strings` is incorrect, and we should fix #640 by moving the True/False conversion to `build_arg_strings` (but in a separate PR).

weiji14: Ok, then this PR should be ok then for now. I'll approve it but you'll need to update the branch in order to merge.

weiji14 (approved):

PR #645, opened 2020-10-09
[DO NOT MERGE THIS PR] Add slash command "/black" to format PRs

seisman:

This PR adds a special github action to enable [slash commands](https://github.com/peter-evans/slash-command-dispatch) in comments. With the feature enabled, anyone, who have written permission to a pull request, can write `/black` at the first line of a comment to trigger the "black" action, which can format the codes automatically.

I tried to write the slash command `/black` in the comments below, but it doesn't work as I expect. It seems the special event "repository dispatch" can only read configurations files in the master branch. Thus, testing the action will definitely pollute the git history of the master branch. So I forked pygmt to my account (https://github.com/seisman/pygmt) and do some tests on my own fork. 

---

The slash command works well in my fork. Here are the changes to the master branch in my own fork: https://github.com/GenericMappingTools/pygmt/compare/master...seisman:master (I will open a PR to merge these changes into master branch).

https://github.com/seisman/pygmt/pull/5 is a PR I opened with some bad changes to codes. Writting `/format` (I use `/format` instead of `/black` in my fork) in the comment triggered the action to format the codes automatically.

Here is the screenshot:
image

---

Please review the changes of https://github.com/GenericMappingTools/pygmt/compare/master...seisman:master (I made a PR in #646 for easier review).

BTW, feel free to submit PRs to my fork https://github.com/seisman/pygmt, to check if slash commands work for PRs from external contributors.

NOTE: PR #645 won't be merged.

### DIFF
--- a/.github/workflows/black-command.yml
+++ b/.github/workflows/black-command.yml
@@ -9,7 +9,7 @@ jobs:
       # Checkout the pull request branch
       - uses: actions/checkout@v2
         with:
-          token: ${{ secrets.ACTIONS_BOT_TOKEN }}
+          token: ${{ secrets.PAT }}
           repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
           ref: ${{ github.event.client_payload.pull_request.head.ref }}
 
--- a/.github/workflows/black-command.yml
+++ b/.github/workflows/black-command.yml
@@ -9,7 +9,7 @@ jobs:
       # Checkout the pull request branch
       - uses: actions/checkout@v2
         with:
-          token: ${{ secrets.PAT }}
+          token: ${{ secrets.ACTIONS_BOT_TOKEN }}
           repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
           ref: ${{ github.event.client_payload.pull_request.head.ref }}
 
@@ -38,7 +38,7 @@ jobs:
       - name: Add reaction
         uses: peter-evans/create-or-update-comment@v1
         with:
-          token: ${{ secrets.PAT }}
+          token: ${{ secrets.ACTIONS_BOT_TOKEN }}
           repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
           comment-id: ${{ github.event.client_payload.github.payload.comment.id }}
           reaction-type: hooray
--- a/.github/workflows/black-command.yml
+++ b/.github/workflows/black-command.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Add reaction
         uses: peter-evans/create-or-update-comment@v1
         with:
-          token: ${{ secrets.ACTIONS_BOT_TOKEN }}
+          token: ${{ secrets.PAT }}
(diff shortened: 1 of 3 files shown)


.github/workflows/black-command.yml
```diff
@@ -0,0 +1,44 @@
...
+          black .
+          git config --global user.name 'actions-bot'
+          git config --global user.email '58130806+actions-bot@users.noreply.github.com'
+          git commit -am "[black-command] fixes"
+          git push
+
+      - name: Add reaction
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          token: ${{ secrets.PAT }}
```
weiji14, line 41: Whose Personal Access Token (PAT) is this, and what scope does it have? I see it's stored as a secret in the PyGMT repo (though I can't see the secret itself), and know that it's needed to let the Github Action 'bot' do the work. Just asking because I realize at https://github.com/conda-forge/gmt-feedstock/pull/100#discussion_r501595820 that we might need to use a PAT there too. Was wondering if we could have a GMT organization-wide PAT for these Github Actions that require permission to make commits.

weiji14 (commented):
Thanks for doing this @seisman, it will be a really useful addition to the project! I know just how much trial and error setting up this sort of Github Action can take, but it's looking promising.

Just one comment about the PAT for now. Later on, once we settle on the implementation, it will be best to document this somewhere too.

.github/workflows/black-command.yml
seisman: > Whose Personal Access Token (PAT) is this, and what scope does it have?

It's a PAT of my own account, with the `repo` scope.  

> Was wondering if we could have a GMT organization-wide secret PAT for these Github Actions that require permission to make commits.

[It seems impossible to have an organization-wide PAT](https://stackoverflow.com/questions/31159275/how-do-i-generate-the-github-oauth-token-for-organization-accounts). The only way is someone generates a PAT and saves it as an organization-wide secret.

> Alternatively, we could look into using https://github.com/tibdex/github-app-token - i.e. make an organization GMT Github App, and use that App to generate what is basically a PAT token on the fly (rather than one of our personal ones).

Yes, it looks promising.

> I'd help out with this but haven't got admin permissions in the GMT organization.

Perhaps you can ask for admin permissions if necessary.

---

BTW, this PR doesn't work. The working one is PR https://github.com/GenericMappingTools/pygmt/pull/646.

weiji14: Ok, shall we try using tibdex/github-app-token then? The step by step instructions at https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens seem fairly straightforward, and after we make the GMT-bot Github App, we can install it here on GenericMappingTools/pygmt and at GenericMappingTools/gmt-feedstock to authenticate these bots.

And yes, I'll review things on #646 later.

seisman: It seems that the github incident is resolved, right?

PR #659, opened 2020-10-21
Fix Github to GitHub

seisman:

We should use "github" or "GitHub", not "Github".

**Reminders**


weiji14 (approved):
Looks good. Somehow I never noticed GitHub was [camel case](https://en.wikipedia.org/wiki/Camel_case) until now!

PR #661, opened 2020-10-22
Add "no_clip" to plot, text, contour and meca

seisman:
**Description of proposed changes**

This PR adds the *no_clip* argument (alias of **-N**) to plot, text, countour and meca.

The name *no_clip* was agreed when wrapping the *grdimage* method in https://github.com/GenericMappingTools/pygmt/pull/620#discussion_r491742142

Fixes #

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #627, opened 2020-09-21
Add common alias perspective (p) for plotting 3D illustrations

weiji14:
**Description of proposed changes**

Used to select a perspective view by setting the azimuth and elevation angle of the viewpoint. See https://docs.generic-mapping-tools.org/6.1/gmt.html#perspective-full.

perspective minimal docs with link to gmt manpage

Live documentation preview for this PR/branch is at https://pygmt-git-common-aliases-p.gmt.vercel.app/api/generated/pygmt.Figure.plot.html

Note that the 'perspective' name follows the upstream GMT long option name convention at https://github.com/GenericMappingTools/gmt/blob/88421f9f8fe9759692302905268f4036b287adcb/src/gmt_init.c#L408

This PR will be a part of #620 too.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #624, opened 2020-09-20
Add common aliases xshift (X) and yshift (Y)

weiji14:
**Description of proposed changes**

Used for shifting plots along the x and y dimensions. See https://docs.generic-mapping-tools.org/6.1/gmt.html#xy-full.

xshift yshift minimal docs with link to gmt manpage

Live documentation preview for this PR/branch is at https://pygmt-git-common-aliases-xy.gmt.vercel.app/api/generated/pygmt.Figure.plot.html

Note that the xshift/yshift names follow upstream GMT long option name conventions at https://github.com/GenericMappingTools/gmt/blob/88421f9f8fe9759692302905268f4036b287adcb/src/gmt_init.c#L393-L394

Details to address:
1. Should we have separate `{X}` and `{Y}` entries, or a single `{XY}` entry?
2. Do we want a) full documentation from https://docs.generic-mapping-tools.org/6.1/gmt.html#xy-full, or b) just the simplified shortened docsting found at https://docs.generic-mapping-tools.org/6.1/std-opts.html? 

See related issue at #380. This PR will be a part of #620 too.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #666, opened 2020-10-23
Complete documentation for plot

weiji14:

Considering that `plot` is one of the more widely used functions, we should document it better than what we currently have at https://www.pygmt.org/v0.2.0/api/generated/pygmt.Figure.plot.html! For reference, upstream GMT docs is at https://docs.generic-mapping-tools.org/6.1/plot.html and https://github.com/GenericMappingTools/gmt/blob/6.1/doc/rst/source/plot_common.rst_.

Live documentation preview is at https://pygmt-git-improve-docs-plot.gmt.vercel.app/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot.

TODO:

Arguments/Aliases to document (see also https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.plot-Tuple{Any}). New aliases are highlighted in **bold**:

- [x] A - **straight_line**
- [x] B - frame
- [x] C - cmap
- [x] D - **offset**
- [x] E - **error_bar**
- [x] [F](https://docs.generic-mapping-tools.org/6.1/plot.html#f) - **connection**
- [x] G - color
- [x] [I](https://docs.generic-mapping-tools.org/6.1/plot.html#i) - **intensity**
- [x] J - projection
- [x] L - **close**
- [x] N - no_clip (done at #661)
- [x] R - region
- [x] S - style
- [ ] [T](https://docs.generic-mapping-tools.org/6.1/plot.html#t) - (not aliased as discussed at https://github.com/GenericMappingTools/pygmt/pull/666#issuecomment-714887759)
- [x] U - timestamp
- [x] V - verbose (done at #550)
- [x] W - pen
- [x] X/Y - xshift, yshift (done at #624)
- [x] [Z](https://docs.generic-mapping-tools.org/6.1/plot.html#z) - **zvalue**
- [x] i - columns
- [x] l - label
- [x] p - perspective (done at #627)
- [x] t - transparency (done at #614)

Fixes #

**Reminders**

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -643,12 +643,13 @@ class BasePlotting:
         If a symbol is selected and no symbol size given, then plot will
         interpret the third column of the input data as symbol size. Symbols
         whose size is <= 0 are skipped. If no symbols are specified then the
-        symbol code (see *symbol* below) must be present as last column in the
-        input. If *symbol* is not used, a line connecting the data points will
-        be drawn instead. To explicitly close polygons, use *close*. Select a
-        fill with *color*. If *color* is set, *pen* will control whether the
-        polygon outline is drawn or not. If a symbol is selected, *color* and
-        *pen* determines the fill and outline/no outline, respectively.
+        symbol code (see **symbol** below) must be present as last column in
+        the input. If **symbol** is not used, a line connecting the data points
+        will be drawn instead. To explicitly close polygons, use **close**.
+        Select a fill with **color**. If **color** is set, **pen** will control
+        whether the polygon outline is drawn or not. If a symbol is selected,
+        **color** and **pen** determines the fill and outline/no outline,
+        respectively.
 
         Full option list at :gmt-docs:`plot.html`
 
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -606,47 +606,49 @@ class BasePlotting:
 
     @fmt_docstring
     @use_alias(
-        R="region",
-        J="projection",
+        A="straight_lines",
         B="frame",
-        S="style",
-        G="color",
-        N="no_clip",
-        W="pen",
-        i="columns",
-        l="label",
         C="cmap",
+        D="position",
+        E="error_bars",
+        G="color",
+        J="projection",
+        L="close",
+        N="no_clip",
+        R="region",
+        S="style",
         U="timestamp",
         V="verbose",
+        W="pen",
         X="xshift",
         Y="yshift",
+        i="columns",
+        l="label",
         p="perspective",
         t="transparency",
     )
     @kwargs_to_strings(R="sequence", i="sequence_comma", p="sequence")
     def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
         """
-        Plot lines, polygons, and symbols on maps.
-
-        Used to be psxy.
+        Plot lines, polygons, and symbols in 2-D.
 
         Takes a matrix, (x,y) pairs, or a file name as input and plots lines,
         polygons, or symbols at those locations on a map.
 
         Must provide either *data* or *x* and *y*.
 
-        If providing data through *x* and *y*, *color* (G) can be a 1d array
-        that will be mapped to a colormap.
+        If providing data through *x* and *y*, *color* can be a 1d array that
+        will be mapped to a colormap.
 
-        If a symbol is selected and no symbol size given, then psxy will
+        If a symbol is selected and no symbol size given, then plot will
         interpret the third column of the input data as symbol size. Symbols
         whose size is <= 0 are skipped. If no symbols are specified then the
-        symbol code (see *S* below) must be present as last column in the
-        input. If *S* is not used, a line connecting the data points will be
-        drawn instead. To explicitly close polygons, use *L*. Select a fill
-        with *G*. If *G* is set, *W* will control whether the polygon outline
-        is drawn or not. If a symbol is selected, *G* and *W* determines the
-        fill and outline/no outline, respectively.
+        symbol code (see *symbol* below) must be present as last column in the
+        input. If *symbol* is not used, a line connecting the data points will
+        be drawn instead. To explicitly close polygons, use *close*. Select a
+        fill with *color*. If *color* is set, *pen* will control whether the
+        polygon outline is drawn or not. If a symbol is selected, *color* and
+        *pen* determines the fill and outline/no outline, respectively.
 
         Full option list at :gmt-docs:`plot.html`
 
@@ -671,19 +673,33 @@ class BasePlotting:
             depending on the style options chosen.
         {J}
         {R}
-        A : bool or str
-            ``'[m|p|x|y]'``
+        straight_lines : bool or str
+            ``[m|p|x|y]``.
             By default, geographic line segments are drawn as great circle
-            arcs. To draw them as straight lines, use *A*.
+            arcs. To draw them as straight lines, use *straight_lines*.
+            Alternatively, add **m** to draw the line by first following a
+            meridian, then a parallel. Or append **p** to start following a
+            parallel, then a meridian. (This can be practical to draw a line
+            along parallels, for example). For Cartesian data, points are
+            simply connected, unless you append **x** or **y** to draw
+            stair-case curves that whose first move is along *x* or *y*,
+            respectively.
         {B}
         {CPT}
-        D : str
-            ``'dx/dy'``: Offset the plot symbol or line locations by the given
-            amounts dx/dy.
-        E : bool or str
-            ``'[x|y|X|Y][+a][+cl|f][+n][+wcap][+ppen]'``.
-            Draw symmetrical error bars.
+        position : str
+            ``dx/dy``.
+            Offset the plot symbol or line locations by the given amounts
+            *dx/dy* [Default is no offset]. If *dy* is not given it is set
+            equal to *dx*.
+        error_bars : bool or str
+            ``[x|y|X|Y][+a][+cl|f][+n][+wcap][+ppen]``.
+            Draw symmetrical error bars. Full documentation is at
+            :gmt-docs:`plot.html#e`.
         {G}
+        close : str
+          ``[+b|d|D][+xl|r|x0][+yl|r|y0][+ppen]``.
+          Force closed polygons. Full documentation is at
+          :gmt-docs:`plot.html#l`.
         no_clip : bool or str
             ``'[c|r]'``.
             Do NOT clip symbols that fall outside map border [Default plots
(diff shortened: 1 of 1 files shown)


weiji14:
Comments/Suggestions on what to do for these?

> * [ ]  [F](https://docs.generic-mapping-tools.org/6.1/plot.html#f) - **connection/connectivity**?
> * [ ]  [I](https://docs.generic-mapping-tools.org/6.1/plot.html#i) - **intens**?
> * [ ]  [T](https://docs.generic-mapping-tools.org/6.1/plot.html#t) - (do we need this?)
> * [ ]  [Z](https://docs.generic-mapping-tools.org/6.1/plot.html#z) - **level**?

seisman:
- `D - position`: why *position*? The similar option is called *offset* in [`text()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.text.html#pygmt.Figure.text).

- `E - error_bars`: It's OK to add this option here, but note that the option means more columns in the input data, so currently it only works for a plaintext file or 2D numpy arrays.

-  `F - connection/connectivity`: both look good to me.
- `I - intens`: intensity? Just like **grdimage**
- `T - (do we need this?)` No to me.
- `Z - value/level?` *value* is too general, and *level* is also a little confusing. What about *zvalue*?

weiji14:
>     * `D - position`: why _position_? The similar option is called _offset_ in [`text()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.text.html#pygmt.Figure.text).

True, I'll change it to offset, was thinking of some other module like colorbar that uses `D=position`.

>     * `E - error_bars`: It's OK to add this option here, but note that the option means more columns in the input data, so currently it only works for a plaintext file or 2D numpy arrays.

Ok.

>     * `F - connection/connectivity`: both look good to me.

Let's go with connection to keep with GMT.jl

>     * `I - intens`: intensity? Just like **grdimage**

Ok, I prefer intensity too.

>     * `T - (do we need this?)` No to me.

Alright, will ignore this one

>     * `Z - value/level?` _value_ is too general, and _level_ is also a little confusing. What about _zvalue_?

With or without an s? zvalue/zvalues? We're using `Z=trackvalues` in x2sys_cross.

seisman:
> With or without an s? zvalue/zvalues? We're using `Z=trackvalues` in x2sys_cross.

In GMT CLI, it could be a single value or an input file that contains a list of values. 

In PyGMT, I expect it to be a single value, a filename, or a list of values. So *zvalues* may make more sense. But, there are already some inconsistencies about plurals and singulars. For example:

- In [`plot()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot), we use *sizes* (plural) and *direction* (singular), although they both accept 1D numpy arrays.
- In [`text()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.text.html#pygmt.Figure.text), we use *angle*, *justify* and *font* (singluars), although they all accept a single value or an array.

seisman:
I forgot to say that I prefer to use singulars if possible, i.e., use *size* instead of *sizes* in `plot()`.

weiji14:
Do we want to change **sizes** to **size** in this PR, and possibly break backward compatibility? There will be some tutorial examples we need to fix as well.

> * `I - intens`: intensity? Just like **grdimage**

Actually for `grdimage`, we use shading=I, not intensity. But *shading* doesn't quite make sense for vector features like points and lines.

seisman:
> Do we want to change *sizes* to *size* in this PR.

Yes to me. But we need to think more about how `plot()` method will be used as it's too complicated. GMT.jl implements some new methods, like `arrow`, `bar`, `scatter`, `vlines`, by wrapping the complicated `plot` module. We may need to think about if we want to do the similar high-level wrappers.

> and possibly break backward compatibility? There will be some tutorial examples we need to fix as well.

Considering the long-term unsolved issue #256, perhaps we should try to keep backward compatibility, at least for the next few releases. Otherwise, users who use *sizes* would get an error message which is difficult to understand.

> I - intens: intensity? Just like grdimage
> Actually for grdimage, we use shading=I, not intensity. But shading doesn't quite make sense for vector features like points and lines.

Yes, *intensity* should be OK for **plot()**.

weiji14:
> > Do we want to change _sizes_ to _size_ in this PR.
> 
> Yes to me. But we need to think more about how `plot()` method will be used as it's too complicated. GMT.jl implements some new methods, like `arrow`, `bar`, `scatter`, `vlines`, by wrapping the complicated `plot` module. We may need to think about if we want to do the similar high-level wrappers.

Ah, didn't realize GMT.jl has those! The high-level API is for another discussion, I know @leouieda was keen on it, and @MarkWieczorek's https://github.com/SHTOOLS/SHTOOLS library has actually done a lot around PyGMT as well. There's also some discussion over at https://forum.generic-mapping-tools.org/t/pygmt-as-a-high-level-api-for-geovisualization/590, and your #670 feature request is an interesting step in that direction.

> > and possibly break backward compatibility? There will be some tutorial examples we need to fix as well.
> 
> Considering the long-term unsolved issue #256, perhaps we should try to keep backward compatibility, at least for the next few releases. Otherwise, users who use _sizes_ would get an error message which is difficult to understand.

Ok, will keep to using `sizes` for now, and break backward compatibility in PyGMT v0.3.0. Ideally we would have a deprecation warning set-up to be nice, but that will be more work for this PR :laughing:

> Yes, _intensity_ should be OK for **plot()**.

:+1:

pygmt/base_plotting.py
```diff
@@ -606,47 +606,53 @@ def grdview(self, grid, **kwargs):
...
+        If a symbol is selected and no symbol size given, then plot will
         interpret the third column of the input data as symbol size. Symbols
         whose size is <= 0 are skipped. If no symbols are specified then the
-        symbol code (see *S* below) must be present as last column in the
-        input. If *S* is not used, a line connecting the data points will be
-        drawn instead. To explicitly close polygons, use *L*. Select a fill
-        with *G*. If *G* is set, *W* will control whether the polygon outline
-        is drawn or not. If a symbol is selected, *G* and *W* determines the
-        fill and outline/no outline, respectively.
+        symbol code (see **symbol** below) must be present as last column in
```
seisman, line 649: It's *style*, not *symbol*.
```suggestion
        symbol code (see **symbol** below) must be present as last column in
```

weiji14: Good catch!

weiji14:
@seisman any other outstanding issues on this? I'd like to merge this in and work on the `plot3d` implementation (anticipating some tricky merge conflicts).

As agreed above, we'll figure out how to pass in an 'error' column with `error_bars (E)` and rename 'sizes' to 'size' in a separate PR next time.

pygmt/base_plotting.py
```diff
@@ -606,47 +606,52 @@ def grdview(self, grid, **kwargs):
 
     @fmt_docstring
     @use_alias(
-        R="region",
-        J="projection",
+        A="straight_lines",
```
seisman, line 609: `straight_lines` or `straight_line`?

pygmt/base_plotting.py
```diff
@@ -606,47 +606,52 @@ def grdview(self, grid, **kwargs):
...
         B="frame",
-        S="style",
+        C="cmap",
+        D="offset",
+        E="error_bars",
```
seisman, line 613: `error_bars` or `error_bar`?

pygmt/base_plotting.py
```diff
@@ -606,47 +606,52 @@ def grdview(self, grid, **kwargs):
...
-        C="cmap",
+        R="region",
+        S="style",
         U="timestamp",
         V="verbose",
+        W="pen",
         X="xshift",
         Y="yshift",
+        Z="zvalue",
+        i="columns",
```
seisman, line 628: Have we used the *columns* alias for **-i** elsewhere? It may be a bad choice, because GMT also has **-o** which controls columns of output.

In https://github.com/GenericMappingTools/gmt/pull/230, GMT uses *read-columns* for **-i** and *write-columns* for **-o**.

GMT.jl uses *incol* for **-i** and *outcol* for **-o**.

pygmt/base_plotting.py
```diff
@@ -606,47 +606,52 @@ def grdview(self, grid, **kwargs):
...
-        drawn instead. To explicitly close polygons, use *L*. Select a fill
-        with *G*. If *G* is set, *W* will control whether the polygon outline
-        is drawn or not. If a symbol is selected, *G* and *W* determines the
-        fill and outline/no outline, respectively.
+        symbol code (see **style** below) must be present as last column in the
+        input. If **style** is not used, a line connecting the data points will
+        be drawn instead. To explicitly close polygons, use **close**. Select a
+        fill with **color**. If **color** is set, **pen** will control whether
+        the polygon outline is drawn or not. If a symbol is selected, **color**
+        and **pen** determines the fill and outline/no outline, respectively.
```
seisman, line 654: Perhaps we should use _italic_, not **bold** for arguments?

pygmt/base_plotting.py
weiji14: It is used in `contour`:

https://github.com/GenericMappingTools/pygmt/blob/v0.2.0/pygmt/base_plotting.py#L576

I can take this out though, since it's not as useful in the Python world where we can select columns using indexing.

pygmt/base_plotting.py
weiji14: GMT.jl uses `straight_lines`. But maybe we should use `straight_line` here to keep it consistently singular.

pygmt/base_plotting.py
weiji14: Again, GMT.jl uses `error_bars`, but we should probably use (singular) `error_bar`.

pygmt/base_plotting.py
seisman: Yes I agree.

pygmt/base_plotting.py
weiji14: The bold makes it stand out a bit more, but yeah, perhaps use italics to make it consistent with the rest of the docs. I copied this over from https://github.com/GenericMappingTools/gmt/blob/4d9bc24d1b0c7f28b5dde20dd49716d286de4fcc/doc/rst/source/plot_common.rst_#L6-L11.

pygmt/base_plotting.py
seisman: Yes to me.

pygmt/base_plotting.py
seisman: > The bold makes it stand out a bit more

I agree, but let's keep the documentation consistent now. We may need to revisit the issue later.

pygmt/base_plotting.py
weiji14: Oops, actually i="columns" has already been aliased before in plot, just that the code has changed position. So I think we should keep it :sweat_smile:

seisman: OK, let's break backward compatibility later.

seisman (approved):

PR #668, opened 2020-10-23
Let all tests pass for GMT 6.1.1 and master

seisman:
**Description of proposed changes**

It's annoying to have failing CI tests. Every time I want to merge a PR, I have to change the branch protection rules, merge a PR, and then change the protection rules back. 

This PR contains two changes:

- The xarray shading test crashes with GMT 6.1.1, so `pytest.mark.xfail` doesn't work as we expect. 
- The two `info()` tests fail for the GMT master branch, unless the upstream issue https://github.com/GenericMappingTools/gmt/issues/4241 is fixed.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #669, opened 2020-10-24
Do not skip xarray shading tests on all platforms

seisman:

~~In PR #668, we add the `pytest.mark.skip` marker to skip the xarray shading tests on macOS. 
However, the `skip` marker doesn't have the *condition* argument, so the test is skipped on 
all platforms. In This PR, I change the `skip` marker to `skipif`.~~

xarray grid shading is known to fail with GMT 6.1.1. That's why we want to mark the test `xfail`. The `xfail` marker works well for Linux and Windows, but not for macOS, due to segmentation faults.

So ideally what we should do is:

- GMT 6.1.1 + Linux/Windows: xfail
- GMT 6.1.1 + macOS: skip
- GMT master + Linux/Windows/macOS: no marker

**Reminders**

**Notes**

### DIFF
--- a/pygmt/tests/test_grdimage.py
+++ b/pygmt/tests/test_grdimage.py
@@ -80,7 +80,7 @@ def test_grdimage_file():
 )
 @pytest.mark.xfail(
     condition=gmt_version <= Version("6.1.1") and sys.platform != "darwin",
-    reason="Upstream bug in GMT 6.1.1",
+    reason="Upstream bug in GMT 6.1.1 that causes this test to fail on Linux/WIndows",
 )
 @check_figures_equal()
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_grdimage.py
```diff
@@ -74,9 +74,13 @@ def test_grdimage_file():
     return fig
 
 
-@pytest.mark.skip(
+@pytest.mark.skipif(
+    gmt_version <= Version("6.1.1") and sys.platform == "darwin",
+    reason="Upstream bug in GMT 6.1.1",
```
weiji14, line 79:
```suggestion
    reason="Upstream bug in GMT 6.1.1 that causes segfault on macOS",
```

pygmt/tests/test_grdimage.py
```diff
@@ -74,9 +74,13 @@ def test_grdimage_file():
...
+)
+@pytest.mark.xfail(
+    condition=gmt_version <= Version("6.1.1") and sys.platform != "darwin",
     reason="Upstream bug in GMT 6.1.1",
```
weiji14, line 83:
```suggestion
    reason="Upstream bug in GMT 6.1.1 that causes this test to fail on Linux/WIndows",
```

weiji14 (changes requested):

weiji14 (approved):

PR #676, opened 2020-10-29
Complete most of documentation for makecpt

weiji14:

Finish up aliasing **most** arguments for `makecpt`, improving the documentattion to be better than what we currently have at https://www.pygmt.org/v0.2.0/api/generated/pygmt.makecpt.html! For reference, upstream GMT docs is at https://docs.generic-mapping-tools.org/6.1/makecpt.html and https://github.com/GenericMappingTools/gmt/blame/master/doc/rst/source/makecpt.rst.

Live documentation preview is at https://pygmt-git-complete-docs-makecpt.gmt.vercel.app/api/generated/pygmt.makecpt.html

TODO:

Arguments/Aliases to document (see also https://www.generic-mapping-tools.org/GMT.jl/latest/makecpt/). New aliases are highlighted in **bold**:

- [x] A - **transparency**
- [x] C - cmap
- [x] D - **background**/~~**foreground**~~?
- [ ] [E](https://docs.generic-mapping-tools.org/6.1/makecpt.html#e) - **data_levels/nlevels**?
- [x] F - **color_model**
- [x] G - truncate
- [x] H - output
- [x] I - reverse
- [x] M - **overrule_bg**?
- [x] N - **no_bg**?
- [x] Q - **log**
- [ ] [S](https://docs.generic-mapping-tools.org/6.1/makecpt.html#s) - **auto**
- [x] T - series
- [x] V - verbose
- [x] W - **categorical**
- [x] Ww - **cyclic**
- [x] Z - continuous

Fixes #

**Reminders**

**Notes**

### DIFF
--- a/pygmt/mathops.py
+++ b/pygmt/mathops.py
@@ -10,10 +10,13 @@ from .helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_ali
 @use_alias(
     A="transparency",
     C="cmap",
+    D="background",
     F="color_model",
     G="truncate",
     H="output",
     I="reverse",
+    M="overrule_bg",
+    N="no_bg",
     Q="log",
     T="series",
     V="verbose",
@@ -28,8 +31,8 @@ def makecpt(**kwargs):
     This is a module that will help you make static color palette tables
     (CPTs). In classic mode we write the CPT to standard output, while under
     modern mode we simply save the CPT as the current session CPT (but see
-    **output**). You define an equidistant set of contour intervals or pass
-    your own z-table or list, and create a new CPT based on an existing master
+    *output*). You define an equidistant set of contour intervals or pass your
+    own z-table or list, and create a new CPT based on an existing master
     (dynamic) CPT. The resulting CPT can be reversed relative to the master
     cpt, and can be made continuous or discrete. For color tables beyond the
     standard GMT offerings, visit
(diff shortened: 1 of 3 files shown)


weiji14:
> * [ ]  [E](https://docs.generic-mapping-tools.org/6.1/makecpt.html#e) - **data_levels/nlevels**?
> * [ ]  [S](https://docs.generic-mapping-tools.org/6.1/makecpt.html#s) - **auto**

These two flags ([`-E`](https://docs.generic-mapping-tools.org/6.1/makecpt.html#e) and [`-S`](https://docs.generic-mapping-tools.org/6.1/makecpt.html#s)) require a **file** input (or **pandas.DataFrame** table), but this isn't implemented in `pygmt.makecpt` yet. Probably best to leave the implementation to a separate PR?

seisman:
For `makecpt`, each option does too many things. Not sure if these are good choices:

``` 
D - background/foreground
W - categorical
```

weiji14:
> For `makecpt`, each option does too many things. Not sure if these are good choices: 
> ``` 
> D - background/foreground
> W - categorical
> ```
Yeah, it does get confusing. I couldn't think of a word that covers background/foreground (and NaN) though. Same with W which could be categorical/wrap/cyclic :/

seisman:
For **-W**, we could use `categorical` (True or False) for **-W** and `cyclic` (True or False) for **-Ww**.

weiji14:
> For **-W**, we could use `categorical` (True or False) for **-W** and `cyclic` (True or False) for **-Ww**.

Sure, I did think of that actually, but we'll need an extra check so that categorical (W) and cyclic (Ww) are not used at the same time.

pygmt/mathops.py
```diff
@@ -8,40 +8,106 @@
...
 @kwargs_to_strings(T="sequence", G="sequence")
 def makecpt(**kwargs):
     """
-    Creates a static color palette table (CPT).
+    Make GMT color palette tables.
+
+    This is a module that will help you make static color palette tables
+    (CPTs). In classic mode we write the CPT to standard output, while under
+    modern mode we simply save the CPT as the current session CPT (but see
+    *output*). You define an equidistant set of contour intervals or pass your
```
seisman, line 35: Perhaps we should not mention *classic mode* here.

seisman (approved):

Issue #566, opened 2020-08-18
Migrate Travis CI jobs to GitHub Actions


seisman:
**Description of the desired feature**

We recently migrated all the CI testings and style checks from Azure Pipelines to GitHub Actions. Currently, we're using both GitHub Actions and Travis CI.

The Travis CI is doing following jobs:

- ~~Test PyGMT with Python 3.6-3.8 on Linux~~
- [ ] Deploy documentation to gh-pages branch for master branch and tags
- [x] Publish PyGMT to PyPi for tags (#679)

The first one is already covered in the GitHub Actions, so it's OK to just remove them from Travis CI.

The 2nd and the 3rd one should be migrated to GitHub Actions.

**Are you willing to help implement and maintain this feature?** Yes, but contributions are welcome.

weiji14:
No strong objections to this, we should do all three in separate PRs. Only concern would be that this makes up totally dependent on Github, which might be problematic if GIthub servers go down (but that's a bigger issue in itself).

> The first one is already covered in the GitHub Actions, so it's OK to just remove them from Travis CI.

Yep. Or we could keep one single test around. Maybe a test for a [nightly Python build](https://docs.travis-ci.com/user/languages/python/#specifying-python-versions) for the upcoming Python 3.9?

> The 2nd and the 3rd one should be migrated to GitHub Actions.

On the 2nd one, see https://github.com/GenericMappingTools/website/pull/65 as an example on documentation build/deploy.

For the 3rd one (Publish to PyPI), we could use https://github.com/marketplace/actions/pypi-publish, and maybe test it on https://test.pypi.org (instead of the main pypi.org website) first to be sure we don't break anything.

seisman:
> > The first one is already covered in the GitHub Actions, so it's OK to just remove them from Travis CI.
> 
> Yep. Or we could keep one single test around. Maybe a test for a [nightly Python build](https://docs.travis-ci.com/user/languages/python/#specifying-python-versions) for the upcoming Python 3.9?

We rely on packages from conda-forge. If conda-forge doesn't provide numpy, pandas and xarray packages for Python 3.9, I don't think we can get the testings running.

 
> > The 2nd and the 3rd one should be migrated to GitHub Actions.
> 
> On the 2nd one, see [GenericMappingTools/website#65](https://github.com/GenericMappingTools/website/pull/65) as an example on documentation build/deploy.

[...]

Issue #671, opened 2020-10-24
Switch from versioneer to setuptools-scm


weiji14:
Ideally we would switch from versioneer to setuptools-scm (see https://github.com/pydata/xarray/issues/2853), but haven't got around to doing this yet :slightly_smiling_face:

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/642#discussion_r499294524_

PR #679, opened 2020-11-05
Migrate PyPI release to GitHub Actions

seisman:

This PR migrate the PyPI release job from Travis CI to GitHub Actions.

References:

- https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows.
- https://github.com/fatiando/boule/tree/master/.github/workflows

Address #566 

**Reminders**

**Notes**

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -39,8 +39,8 @@ jobs:
       run: |
         # Change the versioneer format to "pre" so that the commit hash isn't
         # included (PyPI doesn't allow it). Can't do this permanently because
-        # we rely the hash to indicate to the tests that this is a local
-        # verison instead of a published version.
+        # we rely on the hash to tell the tests that this is a local version
+        # instead of a published version.
         #
         # The step is only necessary for testing purpose
         sed --in-place "s/pep440/pep440-pre/g" setup.cfg
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,15 +1,19 @@
-# Workflow for publishing archives to PyPI
+# Publish archives to PyPI and TestPyPI using GitHub Actions
+
 name: Publish to PyPI
 
+# Only run for pushes to the master branch and releases.
 on:
-  # Runs for pull requests for testing purposes
-  pull_request:
+  push:
     branches:
       - master
-  # Only runs for releases
   release:
     types:
       - published
+  # Runs for pull requests should be disable unless for testing purposes
+  pull_request:
+    branches:
+      - master
 
 jobs:
   publish-pypi:
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,35 @@
+name: Publish to PyPI
+
+on: push
+
+jobs:
+  publish-pypi:
+    name: Publish to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install dependencies
+      run: pip install wheel
+
+    - name: Build source distributions and wheels
+      run: python setup.py bdist_wheel
+
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.test_pypi_password }}
+        repository_url: https://test.pypi.org/legacy/
+
+    - name: Publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.pypi_password }}
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -22,8 +22,17 @@ jobs:
     - name: Install dependencies
       run: pip install wheel
 
-    - name: Build source distributions and wheels
-      run: python setup.py bdist_wheel
+    - name: Build source and wheel distributions
+      run: |
+        # Change the versioneer format to "pre" so that the commit hash isn't
+        # included (PyPI doesn't allow it). Can't do this permanently because
+        # we rely the hash to indicate to the tests that this is a local
+        # verison instead of a published version.
+        sed --in-place "s/pep440/pep440-pre/g" setup.cfg
+        python setup.py sdist bdist_wheel
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
 
     - name: Publish to Test PyPI
       uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -10,10 +10,10 @@ on:
   release:
     types:
       - published
-  # Runs for pull requests should be disable unless for testing purposes
-  pull_request:
-    branches:
-      - master
+  # Runs for pull requests should be disabled other than for testing purposes
+  #pull_request:
+  #  branches:
+  #    - master
 
 jobs:
   publish-pypi:
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -9,7 +9,10 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v2.3.4
+      with:
+        # fecth all history so that versioneer works
+        fetch-depth: 0
 
     - name: Set up Python
       uses: actions/setup-python@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+
+    - name: Install dependencies
+      run: pip install wheel
+
+    - name: Build source distributions and wheels
+      run: python setup.py bdist_wheel
```
weiji14, line 26:
```suggestion
    - name: Build source and wheel distributions
      run: |
        # Change the versioneer format to "pre" so that the commit hash isn't
        # included (PyPI doesn't allow it). Can't do this permanently because
        # we rely the hash to indicate to the tests that this is a local
        # verison instead of a published version.
        sed --in-place "s/pep440/pep440-pre/g" setup.cfg
        python setup.py sdist bdist_wheel
        echo ""
        echo "Generated files:"
        ls -lh dist/
```

See if this works, inspired by https://github.com/fatiando/pooch/pull/196, specifically the lines at https://github.com/fatiando/pooch/blob/a9564b13a95cb4521dcf5938cea5adfd2851b385/.github/workflows/deploy.yml#L51-L61

weiji14 (commented):
I made a suggestion following @leouieda's PR at https://github.com/fatiando/pooch/pull/196. **However**, I'm not sure if I like this workaround, since we will have to switch from versioneer to setuptools_scm (#671) sooner or later.

seisman:
See the "fake" release in the test pypi repository: https://test.pypi.org/project/pygmt/0.2.0.post0.dev52/#history

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,61 @@
+# Workflow for publishing archives to PyPI
+name: Publish to PyPI
+
+on:
+  # Runs for pull requests for testing purposes
+  pull_request:
+    branches:
+      - master
```
seisman, line 8: We don't want to trigger the workflow for pushes and pull requests.
```suggestion
  # Runs for pull requests for testing purposes
  #pull_request:
  #  branches:
  #    - master
```

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: Fix up version string
+      run: |
+        # Change the versioneer format to "pre" so that the commit hash isn't
+        # included (PyPI doesn't allow it). Can't do this permanently because
+        # we rely the hash to indicate to the tests that this is a local
+        # verison instead of a published version.
+        #
+        # The step is only necessary for testing purpose
+        sed --in-place "s/pep440/pep440-pre/g" setup.cfg
```
seisman, line 42: We may want to skip this step, as it's not needed for real releases.

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,61 @@
...
+        python setup.py sdist bdist_wheel
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
+
+    - name: Publish to Test PyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.1
+      with:
+        password: ${{ secrets.test_pypi_password }}
+        repository_url: https://test.pypi.org/legacy/
```
seisman, line 55: We may skip this step, as it's unnecessary to publish to test PyPI after we merge this PR into master.

.github/workflows/publish-to-pypi.yml
weiji14: Can we ask @leouieda (who has more experience) to weigh in on this?

seisman: I believe keeping it or not makes no difference to tagged releases, as the version strings would always be "vX.Y.Z" with no suffix string.

leouieda: Yep, this step doesn't make a difference for the actual releases since the version number generated is the same. We just merged a similar workaround for setuptools_scm in [fatiando/boule#61](https://github.com/fatiando/boule/pull/61/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R70)

.github/workflows/publish-to-pypi.yml
leouieda: It's not necessary but it's nice to have this test. It avoids headaches when publishing a release and suddenly things fail because PyPI changed something.

leouieda:
I would actually recommend looking at the workflows here instead: https://github.com/fatiando/boule/tree/master/.github/workflows We're also trying out a tool called [nox](https://nox.thea.codes/en/stable/). It's pretty nice to automate some tasks with Python instead of clunky Makefiles that don't work on Windows.

.github/workflows/publish-to-pypi.yml
seisman: To have the test, I think I should keep the workaround `sed --in-place "s/pep440/pep440-pre/g" setup.cfg`, and trigger the workflow for pushes to master branch and tagged releases, right?

leouieda: 👍

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,65 @@
...
+  push:
+    branches:
+      - master
+  release:
+    types:
+      - published
+  # Runs for pull requests should be disable unless for testing purposes
+  pull_request:
+    branches:
+      - master
```
seisman, line 16:
```suggestion
  # Runs for pull requests should be disable unless for testing purposes
  #pull_request:
  #  branches:
  #    - master
```

seisman: This PR is ready for review. 

I will disable builds for pull requests after this PR is approved. After that, this workflow will be triggered for pushed to master branch (to TestPyPI only) and tagged releases (both PyPI and TestPyPI).

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: Install dependencies
+      run: python -m pip install setuptools wheel
+
+    - name: Fix up version string
+      run: |
+        # Change the versioneer format to "pre" so that the commit hash isn't
+        # included (PyPI doesn't allow it). Can't do this permanently because
+        # we rely the hash to indicate to the tests that this is a local
+        # verison instead of a published version.
```
weiji14, line 43:
```suggestion
        # we rely on the hash to tell the tests that this is a local version
        # instead of a published version.
```

weiji14 (commented):
These changes will have to be documented under MAINTENANCE.md. Could you remove mention of Travis CI PyPI deployments at https://github.com/GenericMappingTools/pygmt/blob/v0.2.0/MAINTENANCE.md#travis-ci and place it under the Github Actions section at https://github.com/GenericMappingTools/pygmt/blob/v0.2.0/MAINTENANCE.md#github-actions instead please?

MAINTENANCE.md
```diff
@@ -65,14 +65,20 @@ This is scheduled to run every Sunday at 12 noon.
 If new remote files are needed urgently, maintainers can manually uncomment
 the 'pull_request:' line in that `cache_data.yaml` file to refresh the cache.
 
+4. `publish-to-pypi.yml` (Publish archives to PyPI and TestPyPI)
+
+This workflow is ran to publish archives to PyPI and TestPyPI (for testing only).
+Archives will be pushed to TestPyPI for every commit on the *master* branch and
+tagged releases, and to PyPI for tagged releases only.
```
weiji14, line 72:
```suggestion
4. `publish-to-pypi.yml` (Publish wheels to PyPI and TestPyPI)

This workflow is ran to publish wheels to PyPI and TestPyPI (for testing only).
Archives will be pushed to TestPyPI on every commit to the *master* branch and
tagged releases, and to PyPI for tagged releases only.
```

weiji14 (commented):
Getting there! Just one minor suggestion to change 'archive' to 'wheel'.

weiji14 (approved):
Cool, thanks for doing this @seisman, and to @leouieda for your experience on the fatiando side! This will really streamline our CI going forward :smile:

P.S. remember to disable the Pull Request trigger.

.github/workflows/publish-to-pypi.yml
weiji14: The Publish to Test PyPI step will fail for the next few pushes to master (0.2.0.post0.dev52 onwards) until we release PyGMT v0.2.1 because it's trying to reupload with the same name. See e.g.
https://github.com/GenericMappingTools/pygmt/runs/1368511293?check_suite_focus=true#step:8:27

image

A little annoying, but we can safely ignore it for now :slightly_smiling_face:

seisman: Feel free to delete some releases.

weiji14: Hmm, I can't login to TestPyPI with my PyPI account, are they separate entities? Also seems like you're the only maintainer for PyGMT on TestPyPI.

seisman: Yes they are separate. I will add you and leo as maintainers later.

PR #525, opened 2020-07-14
Wrap velo

lhoupert:
Adding the velo function! Original GMT velo function can be found at https://docs.generic-mapping-tools.org/latest/supplements/geodesy/velo.html.

Live documentation preview: https://pygmt-git-fork-lhoupert-add-velo-gmt.vercel.app/api/generated/pygmt.Figure.velo.html
Gallery example preview: https://pygmt-git-fork-lhoupert-add-velo-gmt.vercel.app/gallery/seismology/velo_arrow_ellipse.html

I followed advice from @weiji14  [[see the feature request I opened here](https://github.com/GenericMappingTools/pygmt/issues/510)]. I implemented the function by mostly copy-modify-pasted from plot and edited the doc string. The options for ```velo``` are quite different from ```plot``` so most of the parameters were changed.

**Problems encountered**
I need help to understand how pygmt is handling the dataframe ```data```. The gmt ```velo``` function requires the argument ```table``` (*One or more ASCII (or binary, see -bi[ncols][type]) data table file(s) holding a number of data columns. If no tables are given then we read from standard input.*). My opinion is that in pygmt, ```Figure.velo``` should reads data values from files, numpy array or panda dataframe.

Can someone help to figure out what should be the pygmt code to run the gmt example given at the end of my request?

I converted the table from standard input into a panda dataframe:
```python
data={'Long.': {0: 0, 1: -8, 2: 0, 3: -5, 4: 5, 5: 0}, 
    ...:  'Lat.': {0: -8, 1: 5, 2: 0, 3: -5, 4: 0, 5: -5}, 
    ...:  'Evel': {0: 0, 1: 3, 2: 4, 3: 6, 4: -6, 5: 6}, 
    ...:  'Nvel': {0: 0, 1: 3, 2: 6, 3: 4, 4: 4, 5: -4}, 
    ...:  'Esig': {0: 4, 1: 0, 2: 4, 3: 6, 4: 6, 5: 6}, 
    ...:  'Nsig': {0: 6, 1: 0, 2: 6, 3: 4, 4: 4, 5: 4}, 
    ...:  'CorEN': {0: 0.5, 1: 0.5, 2: 0.5, 3: 0.5, 4: -0.5, 5: -0.5}, 
    ...:  'SITE': {0: '4x6', 1: '3x3', 2: 'NaN', 3: '6x4', 4: '-6x4', 5: '6x-4'}}  

df1 = pd.DataFrame(data)
```

[...]

PR #684, opened 2020-11-07
Allow np.object dtypes into virtualfile_from_vectors

weiji14:

Loosen the check in `virtualfile_from_vectors` to allow for any string-like dtype (np.str, np.object) by performing the check using `pd.api.types.is_string_dtype()`. The array is then converted (if needed) to a proper `np.str` dtype before giving it to `put_strings`.

**Why is this needed ?**

This is one step in enabling text input into modules like:

- `meca`, see https://github.com/GenericMappingTools/pygmt/pull/516#issuecomment-657305554, https://github.com/GenericMappingTools/pygmt/pull/516/commits/ff164e654827024e76ec892f3acc2ee0d4c47bef
- `velo`, see https://github.com/GenericMappingTools/pygmt/pull/525#discussion_r519019697

Those modules rely on `pandas.DataFrame` inputs, but a 'str' column in pandas is typically stored as an 'object' dtype (see https://stackoverflow.com/questions/21018654/strings-in-a-dataframe-but-dtype-is-object), unless users take due care to store them in the new [pandas.StringDtype](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.StringDtype.html). Either way, when we convert these pandas.Series objects to a numpy array, their dtype becomes `np.object` rather than `np.str` (hence why our code needs to handle np.object too).

After this PR is merged, we can do something like:

```python
            if kind == "matrix":
                if pd.api.types.is_numeric_dtype(data):
                    # all numeric dtypes (e.g. int, float)
                    file_context = lib.virtualfile_from_matrix(data)
                else:
                    # contains a column with object (str) dtype
                    file_context = lib.virtualfile_from_vectors(
                        *[data[column] for column in data]
                    )
```

Fixes #

**Reminders**

**Notes**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -9,6 +9,7 @@ from contextlib import contextmanager
 
 from packaging.version import Version
 import numpy as np
+import pandas as pd
 
 from ..exceptions import (
     GMTCLibError,
@@ -1160,7 +1161,7 @@ class Session:
         # Assumes that first 2 columns contains coordinates like longitude
         # latitude, or datetime string types.
         for col, array in enumerate(arrays[2:]):
-            if np.issubdtype(array.dtype, np.str_):
+            if pd.api.types.is_string_dtype(array.dtype):
                 columns = col + 2
                 break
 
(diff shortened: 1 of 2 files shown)


seisman (approved):
Great! I believe it solves a long-term headache when I use pandas with PyGMT.

PR #485, opened 2020-06-22
Github Actions CI for testing GMT master branch

weiji14:
**Description of proposed changes**

Running Continuous Integration tests on the [gmt-master](https://github.com/GenericMappingTools/gmt) branch. Based on code I had lying around on my computer from May 2020 (tested locally on https://github.com/nektos/act/), but updated with stuff from #475 et al.

This is similar to #462, but on Github Actions (instead of Azure Pipelines), and currently only on Ubuntu and macOS only (may try on Windows someday).

Fixes #345, Closes #462.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #686, opened 2020-11-07
Split meca into a standalone module from base_plotting.py

### LINKED ISSUE #685: Time to split up base_plotting.py into individual module files

The base_plotting.py file (https://github.com/GenericMappingTools/pygmt/blob/v0.2.0/pygmt/base_plotting.py) is the core of PyGMT and home to functions like `fig.plot`, `fig.grdimage` and so on. However, it has grown wayyyy too long (~2000 lines) and we're hitting pylint recommended limit (cf https://github.com/GenericMappingTools/pygmt/pull/525#issuecomment-723057170), namely `C0302: Too many lines in module`.

There will be even more modules coming too! I think we should make things modular and split each `pygmt.Figure` function into it's own standalone file like so:

### Current state

```
--pygmt
   ├── figure.py
   └── base_plotting.py (contains `plot`, `grdimage`, `meca`, etc)
```

### New state

```
--pygmt
  ├── src (new folder)
  |        ├── __init__.py
  |        ├── grdimage.py
  |        ├── meca.py
  |        └── *.py  (all other `fig.*` modules split out from base_plotting.py)
  └── figure.py
```

This is how GMT (see https://github.com/GenericMappingTools/gmt/tree/6.1/src) and GMT.jl (https://github.com/GenericMappingTools/GMT.jl/tree/master/src) organizes their source code too since there's so many GMT modules (https://docs.generic-mapping-tools.org/latest/modules.html).

Note that we originally had a `gmt/modules` folder but that was removed in 252d5df5f493ca2dafb3bed2a0b6909e37c40578, and the big `Figure` class was made because of a decision to use an object orientated style (i.e. `fig.plot()`, `fig.grdimage()`, etc) at some point in 2017. I believe there should be a way to keep that object orientated style while splitting the class into several files, so this refactor should provide the same functionality for our users going forward.

### Transition state

To ease the transition, we can do it chunk by chunk:

1. Split off the big supplementary modules (i.e. `meca`)
2. Split off the grid plotting functions (`grdcontour`, `grdimage`, `grdview`)
3. etc

### Relevant issues:

- `meca` - #516
- `velo` - #510
- `segy` - #179

weiji14:

Begin to split up `base_plotting.py` into standalone components. Starting with the largest module `meca`.

New Structure:

```
pygmt/
  ├── src/ (new folder)
  |        └── meca.py (contains `def meca()` function)
  └── base_plotting.py (contains a line `from pygmt.src.meca import meca`)

```

The following modules from `base_plotting.py` may be split up in separate PRs:

- coast
- colorbar
- grdcontour
- grdimage
- grdview
- plot
- plot3d
- contour
- basemap
- logo
- image
- legend
- text

References:
- https://stackoverflow.com/questions/9638446/is-there-any-python-equivalent-to-partial-classes
- https://stackoverflow.com/questions/37832228/python-class-method-run-when-another-method-is-invoked

Addresses #685

**Reminders**

**Notes**


willschlitzer:
@weiji14 I'm not trying to simplify the work you have done on this, but is this as "simple" as moving a single function to it's own script and adding "from pygmt.src.*script* import *script function*? As far as I can tell that is the case, but I want to check with you before doing work/submitting pull requests for the other functions.

Also, are we trying to get this completed for PyGMT v0.3? If so, do we have an expected deadline for that? I'm assuming it's before the GMT 6.2 release.

pygmt/base_plotting.py
```diff
@@ -1581,405 +1580,5 @@ def text(
...
-                file_context = lib.virtualfile_from_matrix(np.atleast_2d(spec))
-            elif kind == "file":
-                file_context = dummy_context(spec)
-            else:
-                raise GMTInvalidInput("Unrecognized data type: {}".format(type(spec)))
-            with file_context as fname:
-                arg_str = " ".join([fname, build_arg_string(kwargs)])
-                lib.call_module("meca", arg_str)
+    # GMT Supplementary modules
+    from pygmt.src.meca import meca  # pylint: disable=import-outside-toplevel
```
weiji14, line 1584: > is this as "simple" as moving a single function to it's own script and adding "from pygmt.src._script_ import _script function_? As far as I can tell, that is the case, but I want to check with you before doing work/submitting pull requests for the other functions, especially those that I'm in the process of wrapping.

Yes, this is the key line. It looks simple and works, but:

1) we would need to write a separate line for each new module (e.g. `velo`, `solar`, etc).
2) I don't really like having the pylint `disable=import-outside-toplevel` here.

To be honest, there's probably a better way to do this. Any good ideas? Or we can just settle for this for now.

pygmt/src/meca.py
```diff
@@ -0,0 +1,413 @@
...
+        only].
+    {J}
+    {R}
+    {B}
+    {V}
+    {XY}
+    {p}
+    {t}
+    """
+
```
weiji14, line 135: Note that `fig.meca` doesn't appear to use `self._preprocess` like the other `fig` modules, but probably doesn't matter according to https://github.com/GenericMappingTools/pygmt/issues/685#issuecomment-723533013?

pygmt/src/meca.py
```diff
@@ -0,0 +1,413 @@
...
+):
+    """
+    Plot focal mechanisms.
+
+    Full option list at :gmt-docs:`supplements/seis/meca.html`
+
+    Note
+    ----
+        Currently, labeling of beachballs with text strings is only
+        supported via providing a file to `spec` as input.
```
weiji14, line 53: The module has been taken out of the Python `class` which frees up 4 spaces, so should I re-wrap the lines to fit the 79 character limit here?

weiji14 (commented):
Ok, I'm opening this up for review since things kinda work, but need some critical comments on this.

> Also, are we trying to get this completed for PyGMT v0.3? If so, do we have an expected deadline for that? I'm assuming it's before the GMT 6.2 release.

Yes, best to release PyGMT v0.3.0 before GMT 6.2 (scheduled for mid-Feb, but traditionally it's always been late). It'll be good to cut a release once we have a good number of new 'features' like `inset` and such.

pygmt/base_plotting.py
seisman: Is it possible to have `from pygmt.src.meca import meca` in `pygmt/src/__init__.py`, then we only need `from pygmt.src import meca, velo, text` here.

pygmt/src/meca.py
seisman: I think yes.

pygmt/base_plotting.py
weiji14: Yep sure, I'll do that, but it would require changing two files every time we wrap a new module (instead of one right now).

seisman:
The final goal is to have one file for one module, including both plotting and non-plotting modules, right? If so, is it better to put these files in a `modules` directory (I know we already have a `modules.py` file?

weiji14:
> The final goal is to have one file for one module, including both plotting and non-plotting modules, right?

Not necessarily? We'll need to discuss this more in #685 (or a separate issue), it makes sense for the plotting stuff because there's so many.

> If so, is it better to put these files in a `modules` directory (I know we already have a `modules.py` file)?

I chose `src` because GMT and GMT.jl uses `src`, and so does matplotlib. PyGMT originally had a `modules` folder in the early days as I mentioned in https://github.com/GenericMappingTools/pygmt/issues/685#issue-738209748, but yeah, it's not good having a `modules.py` and `modules` folder. Happy to change it to another name though.

pygmt/src/meca.py
seisman: Seems OK to me.

seisman:
This PR looks good to me, but I hesitate to merge it into v0.3.0 release. Do we have enough time and human resources to finish all the modules (plotting, and maybe also non-plotting modules) before v0.3.0?

weiji14:
From a user point of view, nothing actually changes. This PR won't close #485, but it will at least enable others to start putting new functions under the src folder instead of base_plotting.py. The other functions (e.g. grdimage, grdview, etc can be moved after v0.3.0) and I should have more time next month once I hand in my thesis to review our growing backlog.

seisman (approved):

PR #689, opened 2020-11-13
Support Python 3.9

### LINKED ISSUE #644: Check Python 3.9 compatibility

Python 3.9 has been released on 2020-10-05. Should check to see if pygmt works well with Python 3.9. See also previous issue for Python 3.8 at #340.

Python 3.9 and newer release cycle

Links:
- https://blog.python.org/2020/10/python-390-is-now-available-and-you-can.html
- PEP 596 -- Python 3.9 Release Schedule at https://www.python.org/dev/peps/pep-0596/
- https://realpython.com/python39-new-features/
- PEP 602 -- Annual Release Cycle for Python at https://www.python.org/dev/peps/pep-0602/

### Drop Python 3.6 support?

Also, just bringing up the topic of whether we should follow  [NEP 29 — Recommend Python and Numpy version support as a community policy standard](https://numpy.org/neps/nep-0029-deprecation_policy.html) as noted before in https://github.com/GenericMappingTools/pygmt/issues/340#issuecomment-554545546. This would mean dropping support for Python 3.6.

In practice, we'll need to follow in the footsteps of our dependent packages:

- pandas 1.2 dropped Python 3.6 at https://github.com/pandas-dev/pandas/pull/35214
- xarray is less aggressive, they're thinking about it at https://github.com/pydata/xarray/issues/4179

Some options we can take are:

1. Agressive option - Do a PyGMT v0.2.1 release which is the **last** version to support Python 3.6, and state that PyGMT v0.3.0 requires Python 3.7 or newer. 
2. Less aggressive option - Still support Python 3.6 for PyGMT v0.3.0 (but maybe drop it from our test suite to save on CI resources).

weiji14:

Test PyGMT with Python 3.9. See also previous PR for Python 3.8 at #398.

Reminder: Set Python 3.9 Github Actions CI as a required check.

This PR also skips Python 3.6 tests for Pull Requests (but tests still run on master) to conserve on Github Actions CI resources. See https://github.community/t/how-to-conditionally-include-exclude-items-in-matrix-eg-based-on-branch/16853/6.

Fixes #644

**Reminders**

**Notes**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -44,6 +44,12 @@ jobs:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
+        # Run Python 3.6 tests on master branch only to conserve CI resources
+        isMaster:
+          - ${{ contains(github.ref, 'master') }}
+        exclude:
+          - isMaster: false  # i.e. on Pull Requests
+            python-version: 3.6
     # environmental variables used in coverage
     env:
       OS: ${{ matrix.os }}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
     # environmental variables used in coverage
     env:
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -25,7 +25,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
 
       - name: Install packages
         run: pip install black blackdoc flake8 pylint
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -76,7 +76,7 @@ First, we must configure conda to get packages from the
 Now we can create a new conda environment with Python and all our dependencies installed
 (we'll call it ``pygmt`` but you can change it to whatever you want)::
 
-     conda create --name pygmt python=3.9 pip numpy pandas xarray netcdf4 packaging gmt
+     conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging gmt
 
 Activate the environment by running::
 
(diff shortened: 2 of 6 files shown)


environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
     - conda-forge
     - defaults
 dependencies:
-    - python=3.7
+    - python=3.8
```
seisman, line 6: Why 3.8 here?

weiji14: I like 3.8 because the miniconda installer stll has 3.8 as the highest version. But we could bump it straight to 3.9 or keep it at 3.7. This is just for the devs.

doc/install.rst
```diff
@@ -76,7 +76,7 @@ First, we must configure conda to get packages from the
 Now we can create a new conda environment with Python and all our dependencies installed
 (we'll call it ``pygmt`` but you can change it to whatever you want)::
 
-     conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging gmt
+     conda create --name pygmt python=3.9 pip numpy pandas xarray netcdf4 packaging gmt
```
seisman, line 79: https://docs.anaconda.com/anaconda/packages/pkg-docs/ 

It seems that the default Python version from Anaconda is still Python 3.8. Perhaps we should still use Python 3.8 here?

environment.yml
seisman: It looks OK to me. Please see my another comment.

doc/install.rst
weiji14: Yep, sounds good.

```suggestion
     conda create --name pygmt python=3.8 pip numpy pandas xarray netcdf4 packaging gmt
```

.github/workflows/ci_tests.yaml
```diff
@@ -42,8 +42,14 @@ jobs:
...
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
+        # Run Python 3.6 tests on master branch only to conserve CI resources
+        isMaster:
+          - ${{ contains(github.ref, 'master') }}
+        exclude:
+          - isMaster: false  # i.e. on Pull Requests
+            python-version: 3.6
```
seisman, line 52: It looks good to me, but perhaps we have to change the branch protection rules?

weiji14: Yep, we'll need to change the required checks from 3.6 to 3.9 (keeping 3.7 and 3.8 intact). Will do so when ready to merge.

seisman: All CI jobs pass except the 3 that are excluded.
image

seisman (approved):

.github/workflows/ci_tests.yaml
weiji14: Ok, branch protection rules updated!

PR #695, opened 2020-11-16
Switch from versioneer to setuptools-scm

### LINKED ISSUE #671: Switch from versioneer to setuptools-scm

weiji14:

Replace versioneer added in 3683f37ad6ee5deaed0c0cdba6246bb9a0c978a2 with the more modern [setuptools-scm](https://github.com/pypa/setuptools_scm).

This PR mostly follows https://github.com/psf/black/pull/1008, and https://github.com/fatiando/boule/pull/61:
- Remove all instances of versioneer and pygmt/_version.py. ~~There is a `_version.py` file at the top-level which is git-ignored that contains a single line like ```__version__ = "0.2.2.dev3+g355d85f1"```~~
- Update our Github Actions workflows that mentioned versioneer

What this PR doesn't do:
- Place configuration under pyproject.toml instead of setup.py (see best practice recommended at https://github.com/pypa/setuptools_scm/tree/v4.1.2#pyprojecttoml-usage). Maybe do it when we remove setup.py and replace it entirely with pyproject.toml.

Fixes #671 

**Reminders**

**Notes**


setup.py
```diff
@@ -43,14 +36,21 @@
 ]
 PLATFORMS = "Any"
 INSTALL_REQUIRES = ["numpy", "pandas", "xarray", "netCDF4", "packaging"]
+# Configuration for setuptools-scm
+SETUP_REQUIRES = ["setuptools_scm"]
```
seisman, line 40: Do we need to add `setuptools_scm` to requirements-dev.txt and enviroment.yaml?

weiji14: So this is a set-up (i.e. build) requirement (used for packaging to PyPI) rather than a runtime (i.e. end-user) requirement. So I don't think we have to, but the real test will be to see if things get uploaded to TestPyPI properly after this PR is merged.

setup.py
```diff
@@ -43,14 +36,21 @@
...
+USE_SCM_VERSION = {
+    "local_scheme": "node-and-date",
+    "write_to": "_version.py",
```
weiji14, line 43: I wonder if writing to a `_version.py` file is actually needed here.

setup.py
seisman: > but the real test will be to see if things get uploaded to TestPyPI properly after this PR is merged.

Perhaps we should enable TestPyPI publishing in this PR to check if it really works as we expect.

weiji14: Yep, upload to TestPyPI works:

image
image

seisman: Great!

seisman: Wait. Should it be 0.2.1.dev9? Why 0.2.2.dev9?

weiji14: That's correct right? It should be either 0.2.1.**post**9 (https://www.python.org/dev/peps/pep-0440/#post-releases) or 0.2.2.**dev**9 (https://www.python.org/dev/peps/pep-0440/#developmental-releases).

seisman: OK. Never read the PEP 440 carefully before.

setup.py
seisman: https://github.com/fatiando/boule doesn't write the file.

setup.py
weiji14: Yeah, there's too many PEPs to read :joy: But I think we're good here.

setup.py
weiji14: Ok, I'll remove this line then (and don't add `_version.py` to .gitignore too).

seisman (approved):
Looks good.

Issue #690, opened 2020-11-14
Adopt NEP29 - drop support for Python 3.6 and NumPy 1.16


weiji14:
Originally posted at https://github.com/GenericMappingTools/pygmt/issues/644#issue-716952700, but updated slightly

Just bringing up the topic of whether we should follow [NEP 29 — Recommend Python and Numpy version support as a community policy standard](https://numpy.org/neps/nep-0029-deprecation_policy.html) as noted before in #340 (comment). This would mean dropping support for Python 3.6.

In practice, we'll need to follow in the footsteps of our dependent packages:

- pandas 1.2 dropped Python 3.6 at pandas-dev/pandas#35214
- xarray is less aggressive, they're thinking about it at pydata/xarray#4179

Some options we can take are:

- Agressive option - Let PyGMT v0.2.x be the last version series to support Python 3.6, and state that PyGMT v0.3.0 requires Python 3.7 or newer.
- Less aggressive option - Still support Python 3.6 for PyGMT v0.3.0 (but maybe drop it from our test suite entirely to save on CI resources).

---

> Does it mean that users cannot install pygmt via `pip install pygmt` or `conda install pygmt` if they're still using Python 3.6? Or they can still install it, but there is no guarantee that all PyGMT functionalities work well with Python 3.6?

It will be quite hard to support four minor versions of Python (3.6, 3.7, 3.8, 3.9), especially for the `conda` packages where there is a separate build for each one (unless we can work out how to get a 'noarch' build). We *could* still allow `pip install pygmt` Python 3.6 (with no support guarantees, i.e. no CI checks), but it will become a maintenance burden when our dependency packages (numpy, pandas, xarray) move on.

There's no rush to do all this, since Python 3.9 is just released, and there are some libraries still to catch up. Even matplotlib (who has more maintainers) is dropping Python 3.6 for their v3.4 (see https://github.com/matplotlib/matplotlib/pull/17662) so it's worth thinking about following NEP 29 for PyGMT v0.3.0.

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/issues/644#issuecomment-705376657_

**Are you willing to help implement and maintain this feature?** Yes

seisman:
As per the comment in https://github.com/GenericMappingTools/pygmt/issues/665#issuecomment-728748849, we should strictly follow semantic versioning. As we definitely will add new features in the next few weeks/months, the next release will be v0.3.0, and we can drop the Python 3.6 support now. 

[...]

PR #699, opened 2020-11-21
Drop Python 3.6 support

seisman:

- Remove Python 3.6 from our CI jobs
- Bump the minimum required Python version to 3.7 in the documentation
- Remove Python 3.6 from setup.py

Address #690.

**Reminders**

**Notes**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -44,8 +44,6 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        isMaster:
-          - ${{ contains(github.ref, 'master') }}
     # environmental variables used in coverage
     env:
       OS: ${{ matrix.os }}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -42,19 +42,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


seisman:
Do we all agree that the next release will be v0.3.0? If so I will change the projects and milestone.

.github/workflows/ci_tests.yaml
```diff
@@ -42,19 +42,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
-        # Run Python 3.6 tests on master branch only to conserve CI resources
         isMaster:
           - ${{ contains(github.ref, 'master') }}
```
weiji14, line 48: These two lines can be removed too.

```suggestion
```

doc/install.rst
```diff
@@ -16,7 +16,8 @@ Installing
 Which Python?
 -------------
 
-You'll need **Python 3.6 or greater** to run PyGMT.
+You'll need **Python 3.7 or greater** to run PyGMT. Old Python versions may
+work, but there is no guranteen that PyGMT will work as expected with old versions.
```
weiji14, line 20:
```suggestion
work, but there is no guarantee that PyGMT will work as expected with old versions.
```

weiji14 (commented):
Just two minor things, and yes :+1: for going with v0.3.0.

weiji14 (approved):

PR #700, opened 2020-11-21
Provide basic support for FreeBSD

### LINKED ISSUE #698: Basic support for FreeBSD

Although FreeBSD is not as popular as Linux/macOS/Windows, there are still some FreeBSD + GMT users (see https://forum.generic-mapping-tools.org/t/building-gmt-on-freebsd-12-1-release/967), and they're also potential PyGMT users.

As I understand it, [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) will return `freebsdX` on FreeBSD _X_, thus function [`clib_names`](https://github.com/GenericMappingTools/pygmt/blob/7e7da78c773c280d2d9c6724c79ce8b41f6532b0/pygmt/clib/loading.py#L54) will report an error:
```
Operating system "freebsdx" not supported.
```

What we can do is:

1. Confirm the return value of `sys.platform` on FreeBSD (`freebsdX` if I understand the docs correctly)
2. Confirm the suffix of shared libraries on FreeBSD (`.so` based on search results)
3. Add an if test for freebsd

For points 1 and 2, we may ask for some help in this post https://forum.generic-mapping-tools.org/t/967.

We won't have a CI job testing FreeBSD, but at least we can provide some basic support.

**Are you willing to help implement and maintain this feature?** Yes, but we need more inputs from FreeBSD users.

seisman:

This PR provides basic support for FreeBSD. 

For FreeBSD, `sys.platform` returns `freebsdX`, and the suffix of shared libraries is `.so`. 

As per the feedback in the forum https://forum.generic-mapping-tools.org/t/building-gmt-on-freebsd-12-1-release/967/20. FreeBSD users can import pygmt correctly using this PR.

Fixes #698.

**Reminders**

**Notes**

### DIFF
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -68,13 +68,14 @@ def clib_names(os_name):
     """
     if os_name.startswith("linux"):
         libnames = ["libgmt.so"]
-    elif os_name == "darwin":
-        # Darwin is macOS
+    elif os_name == "darwin":  # Darwin is macOS
         libnames = ["libgmt.dylib"]
     elif os_name == "win32":
         libnames = ["gmt.dll", "gmt_w64.dll", "gmt_w32.dll"]
+    elif os_name.startswith("freebsd"):  # FreeBSD
+        libnames = ["libgmt.so"]
     else:
-        raise GMTOSError('Operating system "{}" not supported.'.format(sys.platform))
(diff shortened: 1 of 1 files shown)


seisman:
Please see the feedback at https://forum.generic-mapping-tools.org/t/building-gmt-on-freebsd-12-1-release/967/20

weiji14 (approved):

PR #702, opened 2020-11-24
Improve how PyGMT finds the GMT library

### LINKED ISSUE #628: Find the GMT library by calling "gmt --show-library"

Currently, PyGMT first checks if the environmental variable **GMT_LIBRARY_PATH** is defined. If yes, it tries to find the GMT library in the specified path. Otherwise, it will search for the GMT library in standard library paths (LD_LIBRARY_PATH on Linux; DYLD_LIBRARY_PATH on macOS; PATH on Windows, if I understand it correctly).

As mentioned in https://github.com/GenericMappingTools/gmt/issues/4241#issuecomment-696412420, `gmt --show-library` can return the full path of the GMT library. It can help find the GMT library especially when GMT is not installed via conda. Then the search order would be:

1. GMT_LIBRARY_PATH if defined
2. Full path returned by `gmt --show-library`
3. System-wide standard library paths

seisman:
In the old codes, PyGMT first checks if the environmental variable `GMT_LIBRARY_PATH`
is defined. If yes, it tries to find the GMT library in the specified path.
Otherwise, it will search for the GMT library in standard library paths
(LD_LIBRARY_PATH on Linux; DYLD_LIBRARY_PATH on macOS; PATH on Windows).

This PR improves how PyGMT finds the library, by searching for all possible
paths. The paths from the highest priority to the lowest priority are:

1. the path defined by `GMT_LIBRARY_PATH`
2. the path returned by the command `gmt --show-library`
3. On Windows, also check the path returned by `find_library` function
4. the standard system paths

Thus, the function `clib_full_names` now yields a generator (not a list) 
of possible paths, for example:
```
["/path/defined/by/gmtlibrarypath/libgmt.so",
 "/usr/local/lib/libgmt.so",
 "libgmt.so"]
```

**Pros**:

1. More ways to find the library, so less "GMTCLibNotFoundError" errors
2. Only calls the `gmt` command if `GMT_LIBRARY_PATH` is not defined

**Cons**:

~~1. A much longer list returned by `clib_full_names`. Maybe slower loading?~~
~~2. Difficult to test.~~

**TODO**

- [x] Need to rewrite some tests in the `test_clib_loading.py` file.

Closes #628.

**Reminders**

**Notes**


pygmt/clib/loading.py
```diff
@@ -97,16 +98,37 @@ def clib_full_names(env=None):
...
+            lib_fullnames.append(os.path.join(libpath, libname))
+
+    # Search for the library returned by command "gmt --show-library"
+    try:
+        lib_fullpath = sp.check_output(
+            ["gmt", "--show-library"], encoding="utf-8"
+        ).rstrip("\n")
+        lib_fullnames.append(lib_fullpath)
+    except FileNotFoundError:  # command not found
+        pass
```
weiji14, line 120: Is there a way to avoid always calling `gmt --show-library` here? I.e. check `GMT_LIBRARY_PATH` or the `libnames` list first to see if we can load the library before resorting to a subprocess call here. Might involve quite a bit of refactoring though.

seisman: Yes, I also don't like calling `gmt --show-library` here. I'll see if there is a better way.

leouieda:
Thanks, @seisman! This is a good thing to have and I agree that it's worth the downsides. A few thoughts on the design of loading the library as a whole (might be out of scope for this PR but worth mentioning):

* I initially made the `Session` class load the library when it's used to avoid doing this work at import time. Mostly so we wouldn't have `GMTLibNotFoundError` at import.
* That clearly didn't work because we need to call `begin` at import for this whole thing to work. 
* So it might be time to rethink loading libgmt in the class instead of as a global instead to avoid searching for it every time. 
* Or we figure out a way to not call `begin` at import time.

leouieda:
And yes, these are the trickiest things to test. Using [pytest's monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html) is usually a good way to do it so you can trigger errors yourself.

pygmt/clib/loading.py
weiji14: I changed the `clib_full_names` function into a generator function with `yield` statements (see https://realpython.com/introduction-to-python-generators/#understanding-the-python-yield-statement) so it *should* only hit the `gmt --show library` part if the higher priority stuff didn't get hit. Will see how the CI tests go.

weiji14 (commented):
> **Cons**:
> 
>     1. It calls the "gmt" command in a new process
>
>     2. A much longer list returned by `clib_full_names`. Maybe slower loading?

These cons have been mitigated somewhat in commit 5f17f1e96ec9cc2d33dd41ad86beba77c7a2675f, so the `gmt` subprocess is only called if necessary (i.e. when looking in GMT_LIBRARY_PATH or the default PATH doesn't work).

>     3. Difficult to test.

Monkeypatch tests have been added in test_clib_loading.py so there should be full test coverage now. @GenericMappingTools/python-maintainers, give this a review once you have time (particularly on Windows if possible).

seisman:
@weiji14 Great work! 

I just made a few commits to this branch (dbfd7e29b4233c56262921877146fc85470a2f49, 6769dc3d5c6af3685352e6e7633901ae4dd01945 and c3887ce25e45e23cf6d50cfb7644898c3d7c581c). 

I tested the branch on macOS for many different cases:

- GMT_LIBRAYR_PATH defined
- PATH defined, GMT_LIBRARY_PATH undefined
- both PATH and GMT_LIBRARY_PATH defined
- both PATH and GMT_LIBRARY_PATH undefined, but DYLD_LIBRARY_PATH defined

They all work as I expect (e.g., no extra system calling if the library is already found via `GMT_LIBRARY_PATH`.

I think this PR is ready to merge. Do we want to merge it into v0.3.0?

weiji14:
/test-gmt-master

> I think this PR is ready to merge. Do we want to merge it into v0.3.0?

Yes please! Once the tests pass. I think there's still some missing code coverage? But we can improve on things later in another iteration.

weiji14 (approved):

PR #707, opened 2020-11-28
Add frame as a parameter on the grdview documentation

willschlitzer:
The parameter **frame** was not listed under the documentation for the grdview method. I added it to the documentation using "{B}" to standardize it with the documentation for the other methods in base_plotting.py.


seisman (approved):
Nice fix! Thanks.

PR #708, opened 2020-11-30
Fix typo in link to English Wikipedia page for Iris data set

willschlitzer:
The previous link to the Wikipedia page for the Iris data set include a forward slash at the end of the URL, which linked to a non-existent Wikipedia page that does not redirect to the correct page. I removed the forward slash and the URL now goes to the correct page.


seisman (approved):

seisman:
@willschlitzer Nice catch! Thanks!

PR #709, opened 2020-12-01
Add frame and cmap to the colorbar parameters list

willschlitzer:
The documentation for pygmt.Figure.colorbar() includes frame and cmap under the aliases list, but doesn't include them under the parameters list. I added them in this pull request.

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -179,7 +179,8 @@ class BasePlotting:
 
         Parameters
         ----------
-        {B}
+        frame : str or list
+            Set color bar boundary frame, labels, and axes attributes.
         {CPT}
         position : str
             ``[g|j|J|n|x]refpoint[+wlength[/width]][+e[b|f][length]][+h|v]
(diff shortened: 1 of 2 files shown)


pygmt/base_plotting.py
```diff
@@ -179,6 +179,9 @@ def colorbar(self, **kwargs):
 
         Parameters
         ----------
+        frame : str or list
+            Set color bar boundary frame, labels, and axes attributes.
```
seisman, line 183: Similar to your PR #707, we already have a short alias `{B}` for docstrings of `frame`.
```suggestion
        {B}
```

willschlitzer: @seisman I previously used that alias but then I changed it when I saw that it rendered as "Set map boundary frame and axes attributes." To me, that makes it sound like the frame parameter for the colorbar method has an effect on the overall map frame (like using it in basemap, grdcontour, or grdimage) instead of just on the colorbar.

seisman: Yes, you are right.

seisman (approved):

Issue #292, opened 2019-03-12
Gallery plots for each projection


leouieda:
**Description of the desired feature**

The user documentation contains a [gallery of projections available](https://www.pygmt.org/dev/projections/index.html). The plots showcase the projection so it's easy to choose which one you want (by clicking an image instead of having to google the name). **We need to fill out this gallery** with all projections listed in https://www.generic-mapping-tools.org/gmt/latest/proj_codes_GMT.html 

To make a gallery plot, place a script that generates the figure in `examples/projections` and sphinx-gallery will take care of the rest. The file must have a specific format (see the other examples). Ideally, each projection should show a case in which it's ideal (e.g. not using Mercator for high latitudes).

Any help with this is greatly appreciated! We can build this out slowly, one example at a time.

DefUs3r:
Sir,
I would like to take up this issue. I am a beginner and I have no idea of the codebase and a very slight idea of what this library does. I would like to make my first commit here. Can you help me out?

seisman:
There is a chapter "[GMT Map Projections](https://www.generic-mapping-tools.org/gmt/latest/GMT_Docs.html#gmt-map-projections
)" in the GMT cookbook. We can copy the examples.

leouieda:
Thanks, @seisman. I was wondering where that was. I want to try this out here and hopefully copy some of it over to the GMT docs later on.

leouieda:
:wave: Hi @DefUs3r, thanks for offering to help! Please have a look at our [Contributing Guide](CONTRIBUTING.md). You will need at least a bit of mapping knowledge, though, since the goal is to write example scripts.

xibalba01:
I have a background in geoinformatics and i am slightly familiar with projections so i would like to give this a try. The plots would need to be created with pyGMT right? And the commands used would be the ones from the GMT Map Projections page @seisman mentioned?

leouieda:
Hi @xibalba01 we don't mind PRs at all :slightly_smiling_face: I agree that having a small description of when the projection should be used would be very helpful!

xibalba01:
I have problems building the docs under windows:

Running `make all` or `make html` yields in:
```
(pygmt) C:\Users\mo3kLo4r\Documents\Code\OpenSource\pygmt\doc>make all
ECHO ist ausgeschaltet (OFF).
"Building API docs."
ECHO ist ausgeschaltet (OFF).
sphinx-autogen -i -t _templates -o api/generated api/*.rst
[...]

PR #728, opened 2020-12-13
Add non-geographic plots to projection gallery

willschlitzer:
I added cartesian.py to have a Cartesian example for the projection gallery for #292.

### DIFF
--- a/examples/projections/nongeo/cartesian.py
+++ b/examples/projections/nongeo/cartesian.py
@@ -1,0 +1,15 @@
+"""
+Cartesian
+=========
+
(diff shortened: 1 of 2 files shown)


willschlitzer:
I'm looking to add a polar coordinate plot example to the non-geographic gallery, but can't figure out how to plot it. I can use the P argument but am not sure how to add the other inputs, as the example code on the GMT docs uses `grdmath`, which doesn't look like an option on PyGMT at this time. I'll add polar plotting to the pull request if I can figure it out.

seisman:
> how to add the other inputs

You can simply pass *x* and *y* (i.e., angle *theta* and radius *r*) to the `plot` function.

seisman (approved):

PR #734, opened 2020-12-15
Convert line breaks from DOS to UNIX

seisman:

It seems @willschlitzer is developing PyGMT under Windows, and the new
files he created have DOS-style line breaks.

To keep the consistency in the whole project, I convert all these files
to UNIX-style line breaks using `dos2unix`. I don't know how git
and github handles line breaks, but I can't see the changes 
in the diff view.

I also removed the `x` permission of these files.

Perhaps we also need a style check for line breaks?

**Reminders**

**Notes**


willschlitzer:
Not sure on what's causing this; I'm using PyCharm Professional on Ubuntu 20.04

weiji14:
Wow, good catch, would not have spotted that. Could you add the style check in this PR too if it's not too much work?

@willschlitzer, try https://www.jetbrains.com/help/pycharm/configuring-line-endings-and-line-separators.html

weiji14:
So, chicken and egg question, do we merge this PR first, or #736 with the new style check?

Also, should we have a style check for ensuring there's no executable 'x' permissions (755)?

seisman:
> So, chicken and egg question, do we merge this PR first, or #736 with the new style check?

We can merge this PR first, and then #736.

> Also, should we have a style check for ensuring there's no executable 'x' permissions (755)?

Sounds a good idea.

weiji14 (approved):

PR #736, opened 2020-12-15
Check files for UNIX-style line breaks and 644 permission

seisman:

Running `dos2unix` to all Python scripts and check if there are any files changed,  
to make sure that we don't include any files with DOS-style line breaks.

The style checks fail due to #734. It also means that the line-break style check works as expected.

However, I can't use `/format` to automatically fix the line-break issue, as the 
slash command dispatch read the workflow in the master branch only.

Address #734.

**Reminders**

**Notes**


.github/workflows/ci_tests.yaml
```diff
@@ -31,14 +31,21 @@ jobs:
...
 
       - name: Formatting check (black and flake8)
         run: make check
 
       - name: Linting (pylint)
         run: make lint
 
+      - name: Line breaks check (dos2unix)
+        run: |
+          find . -name "*.py" -exec dos2unix {} +
```
weiji14, line 46: Presumably we want to check all files, not just Python *.py files?

.github/workflows/ci_tests.yaml
```diff
@@ -31,14 +31,21 @@ jobs:
           python-version: 3.8
 
       - name: Install packages
-        run: pip install black blackdoc flake8 pylint
+        run: |
+          pip install black blackdoc flake8 pylint
+          sudo apt-get install dos2unix
```
weiji14, line 36: Didn't know apt- get works on Windows?!

seisman: This is style check and it runs on Ubuntu only.

.github/workflows/ci_tests.yaml
seisman: Yes, but not sure what's the best/correct way to find all ASCII files.

seisman: I just updated the command to:
```
find . -type f -not -path '*/\.git/*' -exec grep -Iq . {} \; -exec dos2unix {} \;
```
The command runs `dos2unix` on all ASCII files.

.github/workflows/ci_tests.yaml
weiji14: Ah right, all good then.

.github/workflows/ci_tests.yaml
```diff
@@ -31,14 +31,21 @@ jobs:
...
 
       - name: Formatting check (black and flake8)
         run: make check
 
       - name: Linting (pylint)
         run: make lint
 
+      - name: Line breaks check (dos2unix)
+        run: |
+          find . -type f -not -path '*/\.git/*' -exec grep -Iq . {} \; -exec dos2unix {} \;
```
weiji14, line 46:
```suggestion
          find . -type f -not -path '*/\.git/*' -exec grep -Iq . {} \; -exec dos2unix --quiet {} \;
```
Do you think we should silence all the "dos2unix: converting file" messages being printed out here in ci_tests.yaml? It's really only the diff we're interested in.

weiji14:
I'll leave it up to you if you decide to implement the permissions (x, 755) check here or in a separate PR as discussed at https://github.com/GenericMappingTools/pygmt/pull/734#issuecomment-745581556.

Also, we should probably have a policy statement in CONTRIBUTING.md on the use of UNIX style line-endings and correct permissions (644) of files.

.github/workflows/ci_tests.yaml
```diff
@@ -31,14 +31,22 @@ jobs:
...
+          pip install black blackdoc flake8 pylint
+          sudo apt-get install dos2unix
 
       - name: Formatting check (black and flake8)
         run: make check
 
       - name: Linting (pylint)
         run: make lint
 
+      - name: Line breaks check (dos2unix)
```
weiji14, line 44:
```suggestion
      - name: Ensure files use UNIX line breaks and have 644 permission
```

seisman:
> Also, we should probably have a policy statement in CONTRIBUTING.md on the use of UNIX style line-endings and correct permissions (644) of files.

I just added one sentence to address the policy.

CONTRIBUTING.md
```diff
@@ -255,6 +255,9 @@ Before committing, run it to automatically format your code:
 make format
 ```
 
+We also consistently use UNIX-style line endings (`\n`) and file permission
+644 (`-rw-r--r--`) throughout the whole project.
+
```
weiji14, line 260:
```suggestion
For consistency, we also use UNIX-style line endings (`\n`) and file permission
644 (`-rw-r--r--`) throughout the whole project.
```

weiji14 (approved):
Perfect!

Issue #493, opened 2020-06-25
Add examples for plotting lines 


seisman:
**Description of the desired feature**

There are already have nice tutorials for plotting [data points of various sizes and colors](https://www.pygmt.org/dev/tutorials/plot.html), but no examples for plotting lines.

It would be good to add some examples showing 

- how to plot lines
- how to specify line attributes (pen thickness, color, and pen style)
- plot lines with various colors as determined by a CPT file

Unfortunately, PyGMT currently can't handle multi-segments data as in GMT. To plot multiple lines, we still have to call `fig.plot()` multiple times.

seisman:
https://github.com/GenericMappingTools/pygmt/pull/604#issuecomment-709696963

> Also, was just looking at the GMT.jl AGU gallery examples at https://www.generic-mapping-tools.org/GMT.jl/latest/gallery/tables/, looks like there's a lot of decorated arrows/curved line examples we can follow.
> 
> ![image](https://user-images.githubusercontent.com/23487320/96206927-77a25280-0fc6-11eb-99a5-d2e405e5a6e3.png)

willschlitzer:
I'll work on this.

willschlitzer:
I'm trying to figure out how to add a specific CPT, but I'm not getting it to plot on the actual line. I tried setting a specific CPT to `cmap`, and also tried adding `"red,green"` to add specific colors, but no luck.

weiji14:
Yeah it's complicated, took me a while to figure it out recently too. Try this for inspiration:

```python
import pygmt
import numpy as np

fig = pygmt.Figure()
fig.basemap(frame=["WSne", "af"], region=[20, 30, -10, 10])
pygmt.makecpt(cmap="batlow", series=[0, 10, 1])
x = np.arange(start=20, stop=30, step=0.2)
for zvalue in range(0, 10):
    y = zvalue * np.sin(x)
    fig.plot(x=x, y=y, cmap=True, zvalue=zvalue, pen=f"thick,+z,-")

fig.show()
```

![sine curve with different colors](https://user-images.githubusercontent.com/23487320/102444216-8623ed80-408d-11eb-91b2-4ba890b3bb8d.png)

willschlitzer:
[...]

PR #741, opened 2020-12-16
Add plotting lines tutorial

willschlitzer:
Adding a tutorial to show how to plot lines, as discussed in #493.


doc/index.rst
```diff
@@ -35,6 +35,7 @@
     tutorials/text.rst
     tutorials/contour-map.rst
     tutorials/earth-relief.rst
+    tutorials/line-plot.rst
     tutorials/configuration.rst
```
seisman, line 39: Move this section up.
```suggestion
    tutorials/line-plot.rst
    tutorials/text.rst
    tutorials/contour-map.rst
    tutorials/earth-relief.rst
    tutorials/configuration.rst
```

willschlitzer:
Would a section on plotting points belong in this tutorial, or should that be a separate tutorial?

seisman:
> Would a section on plotting points belong in this tutorial, or should that be a separate tutorial?

There is already a tutorial about plotting points (https://www.pygmt.org/dev/tutorials/plot.html).

willschlitzer:
Must have missed that; my apologies!

examples/tutorials/line-plot.py
```diff
@@ -0,0 +1,127 @@
...
+)
+fig.plot(x=[2, 4, 10], y=[3, 8, 9], pen="2p,red")
+fig.show()
+
+########################################################################################
+# Change line attributes
+# ----------------------
+#
+# The line attributes can be set using the ``pen`` parameter. ``pen`` takes a string
+# argument with the optional values *line width*,*line color*,*line style*.
```
seisman, line 66: It's not correctly rendered.
image

```suggestion
# argument with the optional values *width*,\ *color*,\ *style*. 
```

examples/tutorials/line-plot.py
```diff
@@ -0,0 +1,127 @@
...
+fig = pygmt.Figure()
+fig.plot(
+    region=[0, 10, 0, 10],
+    projection="X25c/20c",
+    frame="a",
+    x=[1, 8],
+    y=[3, 9],
+    pen="7p,green,-.-",
+)
+fig.show()
```
seisman, line 127: I'm thinking whether it's better to merge these 4 figures into a single figure, i.e., showing 4 lines with different line properties. It would save much space and also easier for people to compare the different *pen* settings.

We also have a gallery example for [different line styles](https://pygmt-qalqf8vyi.vercel.app/gallery/line/linestyles.html). I think you can add a link to that example.

doc/index.rst
```diff
@@ -32,6 +32,7 @@
     projections/index.rst
     tutorials/coastlines.rst
     tutorials/plot.rst
+    tutorials/line-plot.rst
```
seisman, line 35: I'm wondering if `plot-lines.rst` or `lines.rst` is better.

examples/tutorials/line-plot.py
willschlitzer: This was a conscious choice on my part, since the gallery already shows the different *pen* settings side by side. My thought process is that the tutorial instead takes the user through the steps of plotting a line, and shows how they can set one or more properties.

I'm happy to change it if you think its the smarter move, but I do like the way it's currently displayed.

Also, is there a good way to link to other PyGMT pages similar to using `gmt-docs`?

doc/index.rst
willschlitzer: Made the change to `plot-lines.py`.

examples/tutorials/line-plot.py
seisman: OK. let's keep it this way.

seisman: > Also, is there a good way to link to other PyGMT pages similar to using gmt-docs?

You can use to link to that file:
```
:doc:`/gallery/line/linestyles`
```

willschlitzer: Thanks. I wasn't having any luck with trying variations of `pygmt-docs`

seisman: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html 

You may read this if you want to know more about cross-references in ReST language.

PR #745, opened 2020-12-18
Add isort to sort imports alphabetically

seisman:

This PR adds [`isort`](https://pycqa.github.io/isort/) to automatically sort imports.

**Notes:**

- `profile = "black"` is added to `pyproject.toml` for compatibility between `black` and `isort`, which requires isort>=5
- `pylint` also depends on `isort`. Thus when we install pylint, `isort` is already installed.

**References:**

- https://black.readthedocs.io/en/stable/compatible_configs.html#isort
- https://pycqa.github.io/isort/docs/configuration/profiles/

**Reminders**

**Notes**

### DIFF
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -10,6 +10,8 @@
 import atexit as _atexit
 
 from pkg_resources import get_distribution
+
+# Import modules to make the high-level GMT Python API
 from pygmt import datasets
 from pygmt.figure import Figure
 from pygmt.filtering import blockmedian
@@ -18,8 +20,6 @@ from pygmt.gridops import grdcut, grdfilter
 from pygmt.mathops import makecpt
 from pygmt.modules import GMTDataArrayAccessor, config, grdinfo, info, which
 from pygmt.sampling import grdtrack
-
-# Import modules to make the high-level GMT Python API
 from pygmt.session_management import begin as _begin
 from pygmt.session_management import end as _end
 from pygmt.x2sys import x2sys_cross, x2sys_init
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -9,19 +9,20 @@
 
 import atexit as _atexit
 
+from pygmt import datasets
 from pkg_resources import get_distribution
-
-# Import modules to make the high-level GMT Python API
-from pygmt.session_management import begin as _begin, end as _end
 from pygmt.figure import Figure
 from pygmt.filtering import blockmedian
 from pygmt.gridding import surface
-from pygmt.sampling import grdtrack
-from pygmt.mathops import makecpt
-from pygmt.modules import GMTDataArrayAccessor, config, info, grdinfo, which
 from pygmt.gridops import grdcut, grdfilter
-from pygmt.x2sys import x2sys_init, x2sys_cross
-from pygmt import datasets
+from pygmt.mathops import makecpt
+from pygmt.modules import GMTDataArrayAccessor, config, grdinfo, info, which
+from pygmt.sampling import grdtrack
+
+# Import modules to make the high-level GMT Python API
+from pygmt.session_management import begin as _begin
+from pygmt.session_management import end as _end
+from pygmt.x2sys import x2sys_cross, x2sys_init
 
 # Get semantic version through setuptools-scm
 __version__ = f'v{get_distribution("pygmt").version}'  # e.g. v0.1.2.dev3+g0ab3cd78
@@ -60,10 +61,10 @@ def show_versions():
     - GMT library information
     """
 
-    import sys
-    import platform
     import importlib
+    import platform
     import subprocess
+    import sys
 
     def _get_module_version(modname):
         """Get version information of a Python module."""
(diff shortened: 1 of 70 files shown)


pyproject.toml
```diff
@@ -4,3 +4,6 @@ omit = ["*/tests/*", "*pygmt/__init__.py"]
 [tool.pytest.ini_options]
 minversion = "6.0"
 addopts = "--verbose --doctest-modules --mpl --mpl-results-path=results"
+
+[tool.isort]
+profile = "black"
```
weiji14, line 9: I've used `isort` and quite like it, but I've found different `isort` versions to apply the sort styles rather inconsistently (specifically the standard lib/firstparty/thirdparty sections, see https://pycqa.github.io/isort/#custom-sections-and-ordering). Does this "black" profile keep things simple?

pygmt/__init__.py
```diff
@@ -12,16 +12,17 @@
 from pkg_resources import get_distribution
 
 # Import modules to make the high-level GMT Python API
-from .session_management import begin as _begin, end as _end
+from . import datasets
```
weiji14, line 15: The `from .session_management import begin as _begin, end as _end` needs to come before the other stuff I think, is there a way to 'allowlist' this so it doesn't get sorted to the bottom?

seisman: It's unnecessary, but `from . import datasets` must be imported at the end. Otherwise, we will see the below errors:

```python
In [1]: import pygmt
ImportError                               Traceback (most recent call last)
<ipython-input-1-f58669fae37f> in <module>
----> 1 import pygmt

~/Gits/gmt/pygmt/pygmt/__init__.py in <module>
     14 # Import modules to make the high-level GMT Python API
     15 from .session_management import begin as _begin, end as _end
---> 16 from . import datasets
     17 from .figure import Figure
     18 from .filtering import blockmedian

~/Gits/gmt/pygmt/pygmt/datasets/__init__.py in <module>
      3 # Load sample data included with GMT (downloaded from the GMT cache server).
      4
----> 5 from .earth_relief import load_earth_relief
      6 from .tutorial import (
      7     load_japan_quakes,

~/Gits/gmt/pygmt/pygmt/datasets/earth_relief.py in <module>
      5 import xarray as xr
      6
----> 7 from .. import grdcut, which
      8 from ..exceptions import GMTInvalidInput
      9 from ..helpers import kwargs_to_strings

ImportError: cannot import name 'grdcut' from partially initialized module 'pygmt' (most likely due to a circular import) (/Users/seisman/Gits/gmt/pygmt/pygmt/__init__.py)
```

pyproject.toml
seisman: > Does this "black" profile keep things simple?

No, `profile = "black"` is equivalent to the following settings:
```
multi_line_output: 3
include_trailing_comma: True
force_grid_wrap: 0
use_parentheses: True
ensure_newline_before_comments: True
line_length: 88
```

As for section orders, the default order (https://pycqa.github.io/isort/docs/configuration/options/#sections) is:
```
('FUTURE', 'STDLIB', 'THIRDPARTY', 'FIRSTPARTY', 'LOCALFOLDER')
```
Do you mean the default values change in different `isort` versions?

pygmt/__init__.py
seisman: The circular import issue is fixed in commit 11267db38a5771a4db9937593928a91b7b19c3da.

pyproject.toml
weiji14: I think there was a point where they separated well known third-party libaries (e.g. numpy and pandas) from other third-party libraries. Maybe that has changed now with newer `isort` versions, but I'm not too sure.

Do you think we should add other configuration options? I'm looking at xarray's at https://github.com/pydata/xarray/pull/4518 and they have some extra ones besides `black`.

seisman: OK, I will look at these options later.

seisman: Most of our examples import some of the four packages: pygmt, numpy, pandas and xarray.

Currently, `isort` sorts them like 
```
import numpy as np
import pandas as pd
import xarray as xr

import pygmt
```
because numpy, pandas, and xarray are known *third-party* packages, and pygmt is *first-party* package (not sure if I understand it correctly).

I don't think I like it. We can add options:
```
known_third_party = "pygmt"
force_to_top = "pygmt
```
so that they will be sorted like:
```
import pygmt
import numpy as np
import pandas as pd
import xarray as xr
```

seisman: After reading through all the `isort` options, I added the following ones to `pyproject.toml`, similar to https://github.com/pydata/xarray/pull/4518:
```
skip_gitignore = true
force_to_top = "pygmt"
known_third_party = "pygmt"
```
the only exception is `default_section = THIRDPARTY`, as I don't understand what it is, and adding it doesn't change the codes at all.

seisman: I have to remove `force_to_top`, otherwise pylint will complain.

weiji14 (approved):

PR #750, opened 2020-12-19
Allow passing xr.DataArray as shading to grdimage

### LINKED ISSUE #618: Allow the use of a separate file for shading (grdimage -I)
The gmt command grdimage has an option `-I` that adds shading to another image by modulating the intensity.  Historically, (up until gmt v5), this command has only taken as an option another grd filename. One common use of this feature was to plot something like a gravity, the geoid, or magnetic field in color, and then apply shading using a topography file to provide geologic context.

Since gmt 5, there has been the option of using the input grd file along with grdgradient to automatically generate an intensity file that would be used with the `-I` command. I believe that pygmt only accepts the option `shading=True` which automatically generates such an intensity file using gmt defined default illumination angle parameters.

I am asking that pygmt fully implement the `grdimage -I` option. This is the **only** feature that I am currently missing in pygmt, and which is forcing me to use both gmt and pygmt separately to generate publication quality images for journal articles.

For reference, here is the relevant part of the `gmt grdimage` man page:

```
	-I Apply directional illumination. Append name of intensity grid file.
	   For a constant intensity (i.e., change the ambient light), append a value.
	   To derive intensities from <grd_z> instead, append +a<azim> [-45], +n<method> [t1], and +m<ambient> [0]
	   or use -I+d to accept the default values (see grdgradient for details).
```

seisman:

Following the comment https://github.com/GenericMappingTools/pygmt/issues/618#issuecomment-694831387 and the similar implementation in `grdview` function, this PR allows passing xr.DataArray shading to grdimage. 

Here are two examples:

The `shading.grd` file is generated using GMT command, which is not available in PyGMT yet:
```
gmt grdgradient @earth_relief_01d_g -Gshading.grd -A-45 -Nt1+a0
```

```python
import pygmt
import xarray as xr

# reading shading grid as xr.DataArray
shading = xr.open_dataarray("shading.grd")

# grid is from a file, shading is from xr.DataArray
fig = pygmt.Figure()
fig.grdimage(grid="@earth_relief_01d_g", shading=shading)
fig.show()

# both grid and shading are from xr.DataArray
fig = pygmt.Figure()
grid = pygmt.datasets.load_earth_relief(registration="gridline")
fig.grdimage(grid=grid, cmap="geo", shading=shading)
fig.show()
```

Fixes #618.

**ISSUES**

If `shading` is given as a grid file, GMT can automatically derive the gradient/intensity/shading by adding **+d** to the grid file name, e.g., `fig.grdimage(grid="@earth_relief_01d_g", shading="@earth_relief_01d_g+d")`. However, it's impossible when `shading` is given as a xr.DataArray. Perhaps we could add a new argument (e.g., `shading_parameters="+d"`) and automatically append it to the shading file name or virtual file name.

**TODO**

- [ ] Add tests

**Reminders**

**Notes**


MarkWieczorek:
I will test this later in the weekend. As for the "+d" option, here is what I have been doing in pyshtools. I have two optional parameters: `shading` and `shading_azimuth`. (There is not enough info in the `grdimage -I` documentation for me to figure out what the  `+nt1+m0` means, so I am just keeping this as a default. I am not sure if these two parameters are of interest for shading.)

`shading=True` uses the base grid to create the shading, `shading=filename` sends the grd filename to GMT as is, and `shading=xarray` creates a temporary grd file that gets sent to GMT as is. This PR should avoid writing the temporary file to disk.

Using the input value of `shading_azimuth`, I create the string ` "+a{:}+nt1+m0".format(shading_azimuth)`, which gets appended to whatever `shading` is. If `shading_azimuth` is None, then nothing gets appended to `shading`, and the file is assumed to already be a gradient file.

As you mention, as opposed to using the "+d" option, we could just instead create the gradient map in pygmt using grdgradient (when it is wrapped). I'm not sure if this would be preferable to GMT creating the gradient or not.

seisman:
> There is not enough info in the `grdimage -I` documentation for me to figure out what the `+nt1+m0` means

Yes, the **grdimage -I** documentation could be improved. **+nt1+m0** is equivalent to **-Nt1+a0** in **grdgradient** (https://docs.generic-mapping-tools.org/dev/grdgradient.html#n).

> `shading=xarray` creates a temporary grd file that gets sent to GMT as is. This PR should avoid writing the temporary file to disk.

Yes, this PR uses the xarary in memory directly, so no temporary file on disk.

pygmt/base_plotting.py
```diff
@@ -495,8 +495,15 @@ def grdimage(self, grid, **kwargs):
...
             else:
-                raise GMTInvalidInput("Unrecognized data type: {}".format(type(grid)))
-            with file_context as fname:
+                raise GMTInvalidInput(f"Unrecognized data type: {type(grid)}")
+
+            with contextlib.ExitStack() as stack:
+                # shading using an xr.DataArray
+                if "I" in kwargs and data_kind(kwargs["I"]) == "grid":
+                    shading_context = lib.virtualfile_from_grid(kwargs["I"])
+                    kwargs["I"] = stack.enter_context(shading_context)
```
weiji14, line 504: Need a unit test for `fig.grdimage(..., shading=grid, ...)` in https://github.com/GenericMappingTools/pygmt/blob/master/pygmt/tests/test_grdimage.py. Try modifying the `grdview` example (see the `drapegrid`):

https://github.com/GenericMappingTools/pygmt/blob/e5e25cf0d85dda9d2b2807258ffef88963a6f9ce/pygmt/tests/test_grdview.py#L236-L247

weiji14 (changes requested):
The implementation looks good, but the PR still needs a unit test and documentation update for the `shading` part:

https://github.com/GenericMappingTools/pygmt/blob/31c80b9a17698653cc03987a375c252ff81ccc3b/pygmt/base_plotting.py#L457-L469

Change the first line to `shading : str or xarray.DataArray`, and the second line to `Give the name of a grid file or a DataArray with intensities...`

pygmt/base_plotting.py
weiji14: I'll add a unit test for this in a bit and then we can merge it in for PyGMT v0.3.0.

pygmt/tests/test_grdimage.py
```diff
@@ -119,6 +119,20 @@ def test_grdimage_shading_xarray(grid, shading):
...
+def test_grdimage_grid_and_shading_with_xarray(grid, xrgrid):
+    """
+    Test that shading works well when xarray.DataArray is input to both the
+    ``grid`` and ``shading`` arguments.
+    """
+    fig_ref, fig_test = Figure(), Figure()
+    fig_ref.grdimage(
+        grid="@earth_relief_01d_g", region="GL", cmap="geo", shading=xrgrid
+    )
+    fig_test.grdimage(grid=grid, region="GL", cmap="geo", shading=xrgrid)
```
weiji14, line 132: Remind me again, the xarray shading only works with GMT 6.2 is it? As with issue #364? Here's the diff of Greenland (GL):

| expected | actual (wrong) | diff |
|--|--|--|
| test_grdimage_shading_with_xarray png -expected | test_grdimage_shading_with_xarray png | test_grdimage_shading_with_xarray png -failed-diff |

seisman: Tried with GMT master. It doesn't work too.

weiji14: Does this branch work now with GMT 6.2.0rc1?

weiji14: Ok, testing with GMT 6.2.0rc1, the PyGMT output seems better than before (when using GMT 6.1.1) as there isn't an obvious diagonal stripe. But expected (baseline) and actual images still look different. This is with `@check_figures_equal` by the way.

| expected | actual (wrong) | diff |
|--|--|--|
| test_grdimage_grid_and_shading_with_xarray png -expected | test_grdimage_grid_and_shading_with_xarray png | test_grdimage_grid_and_shading_with_xarray png -failed-diff |

weiji14: Ok, I figured out why the images in https://github.com/GenericMappingTools/pygmt/pull/750#discussion_r628957246 are different. For the **baseline** (left) case using `@earth_relief_01d_g`, the CPT is auto-stretched to fit the **Greenland** elevation data range -3893 to 3195.5. For the actual **test** (middle) case using xarray.DataArray, the CPT is auto-stretched to fit the **global** data range -8592.5 to 5559.

Not sure if this can be counted as a minor upstream bug, but a workaround would be to explicitly set a data range for the CPT, or plot a shaded global grid, or maybe try something else.

seisman: I think it's an upstream bug. It would be easier if you can prepare a minimal example so that @PaulWessel or @meghanrjones can reproduce and debug it.

weiji14: Ok, reported upstream at https://github.com/GenericMappingTools/gmt/issues/5294.

pygmt/tests/test_grdimage.py
```diff
@@ -117,6 +117,22 @@ def test_grdimage_shading_xarray(grid, shading):
     return fig_ref, fig_test
 
 
+@check_figures_equal()
```
weiji14, line 120: What do you think about adding an xfail here (so we can merge this PR into master and let Paul test things easier in https://github.com/GenericMappingTools/gmt/issues/5294)?
```suggestion
@pytest.mark.xfail(
    reason="Incorrect scaling of geo CPT on xarray.DataArray grdimage plot."
    "See https://github.com/GenericMappingTools/gmt/issues/5294",
)
@check_figures_equal()
```

weiji14 (commented):
The code implementation is technically correct already in `pygmt/src/grdimage.py`, just that the xarray.DataArray grdimage plot appears different to that from a NetCDF grid. What do you think about merging this PR into master so that can Paul test things easier in GenericMappingTools/gmt#5294?

pygmt/tests/test_grdimage.py
```diff
@@ -117,6 +117,22 @@ def test_grdimage_shading_xarray(grid, shading):
...
+    Test that shading works well when xarray.DataArray is input to both the
+    ``grid`` and ``shading`` arguments.
+    """
+    fig_ref, fig_test = Figure(), Figure()
+    fig_ref.grdimage(
+        grid="@earth_relief_01d_g", region="GL", cmap="geo", shading=xrgrid, verbose="i"
+    )
+    fig_ref.colorbar()
+    fig_test.grdimage(grid=grid, region="GL", cmap="geo", shading=xrgrid, verbose="i")
+    fig_test.colorbar()
```
weiji14, line 132: Btw, I tried using `makecpt` to get both plots to use the same colormap scale, but the images still differ, appears to be some slight offset in the left-right direction.

```suggestion
    makecpt(cmap="geo", series=(-3900, 3200))
    fig_ref.grdimage(
        grid="@earth_relief_01d_g", region="GL", cmap=True, shading=xrgrid, verbose="d"
    )
    fig_ref.colorbar()
    fig_test.grdimage(grid=grid, region="GL", cmap=True, shading=xrgrid, verbose="d")
    fig_test.colorbar()
```

| Expected (correct) | Generated (wrong) | Diff |
|--|--|--|
| test_grdimage_grid_and_shading_with_xarray png -expected | test_grdimage_grid_and_shading_with_xarray png | test_grdimage_grid_and_shading_with_xarray png -failed-diff |

seisman:
> The code implementation is technically correct already in `pygmt/src/grdimage.py`, just that the xarray.DataArray grdimage plot appears different to that from a NetCDF grid. What do you think about merging this PR into master (after adding an xfail to the test) so that can Paul test things easier in [GenericMappingTools/gmt#5294](https://github.com/GenericMappingTools/gmt/issues/5294)?

Sounds good to me.

pygmt/src/grdimage.py
```diff
@@ -36,15 +44,16 @@ def grdimage(self, grid, **kwargs):
...
-    image can optionally be illuminated with the file provided via the
-    ``shading`` parameter. Here, if image has no coordinates then those of the
-    intensity file will be used.
+    building a rectangular image and assigning pixels a gray-shade (or color)
+    based on the z-value and the CPT file. Optionally, illumination may be
+    added by providing a file with intensities in the (-1,+1) range or
+    instructions to derive intensities from the input data grid. Values outside
+    this range will be clipped. Such intensity files can be created from the
+    grid using `grdgradient` and, optionally, modified by `grdmath` or
+    `grdhisteq`. If GMT is built with GDAL support, ``grid`` can be an image
```
weiji14, line 53: `grdgradient` has been wrapped in #1269 so we can point to the actual function now :grin:

```suggestion
    grid using :meth:`pygmt.grdgradient` and, optionally, modified by `grdmath`
    or `grdhisteq`. If GMT is built with GDAL support, ``grid`` can be an image
```

pygmt/src/grdimage.py
```diff
@@ -116,19 +125,20 @@ def grdimage(self, grid, **kwargs):
...
+        ambient light). Alternatively, derive an intensity grid from the input
+        data grid via a call to ``grdgradient``; append **+a**\ *azimuth*,
+        **+n**\ *args*, and **+m**\ *ambient* to specify azimuth, intensity,
+        and ambient arguments for that module, or just give **+d** to select
+        the default arguments (``+a-45+nt1+m0``). If you want a more specific
+        intensity scenario then run ``grdgradient`` separately first. If we
+        should derive intensities from another file than grid, specify the file
+        with suitable modifiers [Default is no illumination]. Note: If the
+        input data is an *image* then an *intensfile* or constant *intensity*
+        must be provided.
```
weiji14, line 141:
```suggestion
        data grid via a call to :meth:`pygmt.grdgradient`; append
        **+a**\ *azimuth*, **+n**\ *args*, and **+m**\ *ambient* to specify
        azimuth, intensity, and ambient arguments for that module, or just give
        **+d** to select the default arguments (``+a-45+nt1+m0``). If you want
        a more specific intensity scenario then run  :meth:`pygmt.grdgradient`
        separately first. If we should derive intensities from another file
        than grid, specify the file with suitable modifiers [Default is no
        illumination]. Note: If the input data is an *image* then an
        *intensfile* or constant *intensity* must be provided.
```

weiji14 (approved):
Good work! I think we can merge after the tests pass so that the debugging work at https://github.com/GenericMappingTools/gmt/issues/5294 can proceed.

PR #751, opened 2020-12-19
Improve the readability of the grdview function

seisman:

When I worked on PR #750 based on the implemention of the `grdview` function,
I found that the `grdview` codes are a little difficult to understand.
The main reason is that `contextlib.ExitStack()` is new to me.
Another reason is that these two codes `fname = stack.enter_context(file_context)`
and `arg_str = " ".join([fname, build_arg_string(kwargs)])` are
separated by the long `if` block.

Fixes #

**Reminders**

**Notes**


seisman:
@weiji14 Please see if you agree with these changes.

weiji14 (approved):
Yep, looks good! Better to keep the variables `fname` closer to where it is used.

Issue #749, opened 2020-12-18
colorbar is missing documentation for "-I" and lacks a "shading" alias


MarkWieczorek:
In practice, whenever an image is plotted using shading ("grdimage -I...") it is common to use the "-I" option for the colorbar as well.  At present there is no documentation of the colorbar "-I" option, but it turns out that you can get this to work anyways. Two calling options are 
```
fig.colorbar(I='')
```
or
```
fig.colorbar(I=True)
```
For simplicity, I would probably document the second option.

Furthermore, for consistency with `grdimage` the colobar function should have an alias called `shading` so that we don't have to use the cryptic gmt short commands.

seisman:
Closed by #752.

PR #752, opened 2020-12-19
Add the shading parameter to colorbar() in base_plotting.py

willschlitzer:
As mentioned in #749, this adds the `shading` alias/parameter to `pygmt.Figure.colorbar`.

I'm not familiar with how shading will affect the color bar, other than that it makes it look different and darker, so I left a pretty vague description for it under the parameter list.

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -225,10 +225,10 @@ class BasePlotting:
             Multiply all z-values in the CPT by the provided scale. By default
             the CPT is used as is.
         shading : str or list or bool
-            Add illumination effects. Optionally, set the range of intensities from
-            -max_intens to +max_intens. If not specified, 1 is used. Alternatively,
-            append low/high intensities to specify an asymmetric range.
-            The default is no illumination.
+            Add illumination effects. Optionally, set the range of intensities
+            from -max_intens to +max_intens. If not specified, 1 is used.
+            Alternatively, append low/high intensities to specify an
+            asymmetric range. The default is no illumination.
         {V}
         {XY}
         {p}
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -227,8 +227,9 @@ class BasePlotting:
         shading : str or list or bool
             Add illumination effects. Optionally, set the range of intensities
             from -max_intens to +max_intens. If not specified, 1 is used.
-            Alternatively, append low/high intensities to specify an
-            asymmetric range. The default is no illumination.
+            Alternatively, set ``shading=[low, high]`` to specify an
+            asymmetric intensity range from *low* to *high*. 
+            The default is no illumination.
         {V}
         {XY}
         {p}
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -161,7 +161,7 @@ class BasePlotting:
         p="perspective",
         t="transparency",
     )
-    @kwargs_to_strings(R="sequence", G="sequence", p="sequence")
+    @kwargs_to_strings(R="sequence", G="sequence", p="sequence", I="sequence")
     def colorbar(self, **kwargs):
         """
         Plot a gray or color scale-bar on maps.
@@ -224,7 +224,7 @@ class BasePlotting:
         scale : float
             Multiply all z-values in the CPT by the provided scale. By default
             the CPT is used as is.
-        shading : str or bool
+        shading : str or list or bool
             Add illumination effects. Optionally, set the range of intensities from
             -max_intens to +max_intens. If not specified, 1 is used. Alternatively,
             append low/high intensities to specify an asymmetric range.
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -225,11 +225,10 @@ class BasePlotting:
             Multiply all z-values in the CPT by the provided scale. By default
             the CPT is used as is.
         shading : str or list or bool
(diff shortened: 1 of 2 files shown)


seisman:
Please see https://docs.generic-mapping-tools.org/dev/colorbar.html#i

willschlitzer:
Didn't think to check the GMT docs on this one; copied the text from the GMT docs and changed the arguments to string or boolean

seisman:
Hi @willschlitzer more work needs to be done here. 

So the possible values of the `shading` (**-I**) option are:

- bool: `shading=True` or `shading=False`
- a single value. for example `shading=0.8` means -0.8 to 0.8
- *low/high*, for example `shading="-0.3/0.8"`. 

For case 3, it would be better if we can also support `shading=[-0.3, 0.8]`, just like we can use `region="0/10/0/10"` and `region=[0, 10, 0, 10]`.

seisman:
> For case 3, it would be better if we can also support `shading=[-0.3, 0.8]`

To implement the feature, you need to add `I="sequence"` to `@kwargs_to_strings`.

willschlitzer:
Sounds good, I'll work on this tomorrow

pygmt/base_plotting.py
```diff
@@ -153,14 +153,15 @@ def coast(self, **kwargs):
...
+        I="shading",
         W="scale",
         V="verbose",
         X="xshift",
         Y="yshift",
         p="perspective",
         t="transparency",
     )
-    @kwargs_to_strings(R="sequence", G="sequence", p="sequence")
+    @kwargs_to_strings(R="sequence", G="sequence", p="sequence", I="sequence")
```
seisman, line 164: We usually put upper-case options before lower-case options.
```suggestion
    @kwargs_to_strings(R="sequence", G="sequence", I="sequence", p="sequence")
```

pygmt/base_plotting.py
```diff
@@ -223,6 +224,11 @@ def colorbar(self, **kwargs):
         scale : float
             Multiply all z-values in the CPT by the provided scale. By default
             the CPT is used as is.
+        shading : str or list or bool
+            Add illumination effects. Optionally, set the range of intensities
+            from -max_intens to +max_intens. If not specified, 1 is used.
+            Alternatively, append low/high intensities to specify an
+            asymmetric range. The default is no illumination.
```
seisman, line 231:
```suggestion
            Alternatively, set ``shading=[low, high]` to specify an
            asymmetric intensity range. The default is no illumination.
```

pygmt/tests/test_colorbar.py
```diff
@@ -154,3 +154,63 @@ def test_colorbar_scaled_z_values():
...
+
+@check_figures_equal()
+def test_colorbar_shading_boolean():
+    """
+    Create colorbar and set shading with a Boolean value
+    """
+    fig_ref, fig_test = Figure(), Figure()
+    # Use single-character arguments for the reference image
+    fig_ref.basemap(R="0/10/0/10", J="X15c", B="a")
+    fig_ref.colorbar(C="geo", I=True)
```
seisman, line 167:
```suggestion
    fig_ref.colorbar(C="geo", I="")
```

pygmt/tests/test_colorbar.py
```diff
@@ -154,3 +154,63 @@ def test_colorbar_scaled_z_values():
...
+
+@check_figures_equal()
+def test_colorbar_shading_float():
+    """
+    Create colorbar and set shading with a single float variable
+    """
+    fig_ref, fig_test = Figure(), Figure()
+    # Use single-character arguments for the reference image
+    fig_ref.basemap(R="0/10/0/10", J="X15c", B="a")
+    fig_ref.colorbar(C="geo", I="0.5")
```
seisman, line 182:
```suggestion
    fig_ref.colorbar(C="geo", I=0.5)
```

pygmt/tests/test_colorbar.py
```diff
@@ -154,3 +154,63 @@ def test_colorbar_scaled_z_values():
...
+
+    fig_test.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
+    fig_test.colorbar(cmap="geo", shading=0.5)
+    return fig_ref, fig_test
+
+
+@check_figures_equal()
+def test_colorbar_shading_string():
+    """
+    Create colorbar and set shading by passing the high/low values as a string
```
seisman, line 192:
```suggestion
    Create colorbar and set shading by passing the low/high values as a string
```

pygmt/base_plotting.py
```diff
@@ -223,6 +224,12 @@ def colorbar(self, **kwargs):
         scale : float
             Multiply all z-values in the CPT by the provided scale. By default
             the CPT is used as is.
+        shading : str or list or bool
+            Add illumination effects. Optionally, set the range of intensities
+            from -max_intens to +max_intens. If not specified, 1 is used.
+            Alternatively, set ``shading=[low, high]`` to specify an
+            asymmetric intensity range from *low* to *high*.
+            The default is no illumination.
```
seisman, line 232: @willschlitzer The tests look good, but we may still have to rephrase these sentences because it's unclear to users what *max_intens* is.

willschlitzer: I think I made a change that makes it more clear.

seisman: Looks great! Thanks.

seisman (approved):

Issue #746, opened 2020-12-18
Table for different map projections


willschlitzer:
The current projection gallery displays the different types of projections by name and a sample image. I propose that the home page for the projections also include a table listing the different projection "codes" (R, G, X, M etc.) and the associated projection name. This will allow someone to quickly identify the projection specified if they are referencing a PyGMT script but unfamiliar with the available projections (as happened to me after seeing the usage of ``projection="R-130/5i"`` in Liam's YouTube tutorial).

**Are you willing to help implement and maintain this feature?** Yes

I'm happy to do this, but am unfamiliar with how to add a table using Sphinx.

seisman:
GMT provides a table for projections (https://docs.generic-mapping-tools.org/latest/proj-codes.html), and the code is available from https://github.com/GenericMappingTools/gmt/edit/master/doc/rst/source/proj-codes.rst.

The table shows both GMT and PROJ projections codes, but perhaps we only want the GMT ones to keep it simple? Also related to https://github.com/GenericMappingTools/pygmt/issues/356.

seisman:
Closed by #753.

PR #753, opened 2020-12-19
Add table for projections

willschlitzer:
As discussed in #746, this is a table listing the projections and the associated arguments for them.

### DIFF
--- a/examples/projections/table/README.txt
+++ b/examples/projections/table/README.txt
@@ -1,8 +1,7 @@
 Projection Table
 ----------------
 
-PyGMT offers 31 map projections. These are specified using the ``projection`` argument.
-The projection codes for the projection styles are shown below.
+The below table shows the projection codes for the 31 GMT projections.
 
 .. Substitution definitions:
 .. |lon0| replace:: lon\ :sub:`0`
--- a/examples/projections/table/README.txt
+++ b/examples/projections/table/README.txt
@@ -19,7 +19,6 @@ The projection codes for the projection styles are shown below.
 
    * - PyGMT Projection Argument
      - Projection Name
-     -
    * - ``projection=`` (upper case for *width*, lower case for *scale*)
      -
    * - **A**\ |lon0|/|lat0|\ [/\ *horizon*]/\ *width*
--- a/examples/projections/table/README.txt
+++ b/examples/projections/table/README.txt
@@ -1,0 +1,26 @@
+Projection Table
+================
+
+PyGMT offers 31 map projections. These are specified using the ``projection`` argument.
+The projection codes for the projection styles are shown below.
+
+.. Substitution definitions:
+.. |lon0| replace:: lon\ :sub:`0`
+.. |lat0| replace:: lat\ :sub:`0`
+.. |lon1| replace:: lon\ :sub:`1`
+.. |lat1| replace:: lat\ :sub:`1`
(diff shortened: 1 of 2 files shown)


willschlitzer:
@seisman I'm a bit stuck. I tried combining the title format from the other README docs with the table formatting in the GMT table code. Any idea why my code is causing trouble?

seisman:
The error message is:
```
If you use an explicit folder ordering, you must specify all folders. Explicit order not found for ..\examples\projections\table
```

It seems you need to add `..\examples\projections\table` to:

https://github.com/GenericMappingTools/pygmt/blob/600d9aec47105502745b88772d2ef3f923981687/doc/conf.py#L61-L70

examples/projections/table/README.txt
```diff
@@ -0,0 +1,84 @@
+Projection Table
+----------------
+
+PyGMT offers 31 map projections. These are specified using the ``projection`` argument.
+The projection codes for the projection styles are shown below. Upper case arguments
+set *map width* and lower case arguments set *map scale*.
```
seisman, line 6: > Upper case arguments set *map width* and lower case arguments set *map scale*.

New users will be confused about what "Upper case" and "Lower case" mean. Please try to rephrase it, or just give an example.

seisman: https://pygmt-git-fork-willschlitzer-proj-table.gmt.vercel.app/projections/index.html

I just realized that we already have similar sentences at the top of this page.

> PyGMT support many map projections. Use the projection argument to specify which one you want to use in all plotting modules. The projection is specified by a one letter code along with (sometimes optional) reference longitude and latitude and the width of the map (for example, Alon0/lat0[/horizon]/width). The map height is determined based on the region and projection.

Perhaps no need to repeat it in the `projections/table/README.txt` file

willschlitzer: I removed it from the table README

examples/projections/table/README.txt
```diff
@@ -0,0 +1,83 @@
...
+     - General perspective
+   * - **H**\ [|lon0|/]\ *width*
+     - Hammer equal area
+   * - **I**\ [|lon0|/]\ *width*
+     - Sinusoidal equal area
+   * - **J**\ [|lon0|/]\ *width*
+     - Miller cylindrical
+   * - **Kf**\ [|lon0|/]\ *width*
+     - Eckert IV equal area
+   * - **-Ks**\ [|lon0|/]\ *width*
```
seisman, line 48:
```suggestion
   * - **Ks**\ [|lon0|/]\ *width*
```

examples/projections/table/README.txt
```diff
@@ -0,0 +1,83 @@
+Projection Table
+----------------
+
+PyGMT offers 31 map projections. These are specified using the ``projection`` argument.
+The projection codes for the projection styles are shown below.
```
seisman, line 5: There are similar sentences at the top of this page (https://pygmt-f6khtwuqs.vercel.app/projections/index.html).
```suggestion
The below table shows the projection codes for the 31 GMT projections.
```

seisman: @willschlitzer I'm not a native English speaker, you can rewrite the sentence if it doesn't read well.

willschlitzer: Looks good to me!

PR #754, opened 2020-12-19
Convert relative imports to absolute imports

seisman:

This PR converts all relative imports to absolute imports.

**Reminders**

**Notes**

### DIFF
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -12,16 +12,16 @@ import atexit as _atexit
 from pkg_resources import get_distribution
 
 # Import modules to make the high-level GMT Python API
-from .session_management import begin as _begin, end as _end
-from .figure import Figure
-from .filtering import blockmedian
-from .gridding import surface
-from .sampling import grdtrack
-from .mathops import makecpt
-from .modules import GMTDataArrayAccessor, config, info, grdinfo, which
-from .gridops import grdcut, grdfilter
-from .x2sys import x2sys_init, x2sys_cross
-from . import datasets
+from pygmt.session_management import begin as _begin, end as _end
+from pygmt.figure import Figure
+from pygmt.filtering import blockmedian
+from pygmt.gridding import surface
+from pygmt.sampling import grdtrack
+from pygmt.mathops import makecpt
+from pygmt.modules import GMTDataArrayAccessor, config, info, grdinfo, which
+from pygmt.gridops import grdcut, grdfilter
+from pygmt.x2sys import x2sys_init, x2sys_cross
+from pygmt import datasets
 
 # Get semantic version through setuptools-scm
 __version__ = f'v{get_distribution("pygmt").version}'  # e.g. v0.1.2.dev3+g0ab3cd78
@@ -40,7 +40,7 @@ def print_clib_info():
     Includes the GMT version, default values for parameters, the path to the
     ``libgmt`` shared library, and GMT directories.
     """
-    from .clib import Session
+    from pygmt.clib import Session
 
     lines = ["GMT library information:"]
     with Session() as ses:
--- a/pygmt/tests/test_meca.py
+++ b/pygmt/tests/test_meca.py
@@ -5,7 +5,7 @@ import os
 import pandas as pd
 import numpy as np
 import pytest
-from pygmt.helpers import GMTTempFile
+from ..helpers import GMTTempFile
 
 from .. import Figure
 
--- a/pygmt/tests/test_meca.py
+++ b/pygmt/tests/test_meca.py
@@ -5,9 +5,9 @@ import os
 import pandas as pd
 import numpy as np
 import pytest
-from ..helpers import GMTTempFile
+from pygmt.helpers import GMTTempFile
 
-from .. import Figure
+from pygmt import Figure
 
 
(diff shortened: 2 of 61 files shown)


pygmt/tests/test_meca.py
```diff
@@ -5,7 +5,7 @@
 import pandas as pd
 import numpy as np
 import pytest
-from pygmt.helpers import GMTTempFile
+from ..helpers import GMTTempFile
```
weiji14, line 8: Since we're on this. Do we want to actually want to continue using relative imports or switch to absolute imports? From https://realpython.com/absolute-vs-relative-python-imports/#pros-and-cons-of-absolute-imports:

> Absolute imports are preferred because they are quite clear and straightforward. It is easy to tell exactly where the imported resource is, just by looking at the statement. Additionally, absolute imports remain valid even if the current location of the import statement changes. In fact, PEP 8 explicitly recommends absolute imports.

> One clear advantage of relative imports is that they are quite succinct. Depending on the current location, they can turn the ridiculously long import statement you saw earlier to something as simple as this:

seisman: That's a good point. PyGMT doesn't have a deep folder structure, so it seems absolute imports are better. I'm OK with changing relative imports to absolute imports.

seisman: Please let me know if you want me to convert relative imports to absolute imports in this PR.

weiji14: Yes, I much prefer absolute imports actually. It'll also help a bit with #686. I think it's fine to do it here in this PR.

seisman: Do you know any tools which can do the conversion automatically？

weiji14: Nope. Probably just regular search and replace.

seisman: Before that, I think we should let pylint check and warn relative imports.

seisman: Tried to add either `W0403` or `relative-import` to `disable=` or `enable=` in `.pylintrc`, but none of them warn about relative imports. Am I misunderstanding how pylint works?

weiji14: I see that we have `no-absolute-import` disabled, does that do anything? Actually have no clue how to set this.

seisman: It's not related:

> no-absolute-import (W1618):
> 
> import missing `from __future__ import absolute_import` Used when an import is not accompanied by from __future__ import absolute_import (default behaviour in Python 3)

seisman:
This PR is almost done, but we have this error:
```
pygmt/__init__.py:43:4: C0415: Import outside toplevel (pygmt.clib.Session) (import-outside-toplevel)
```

@weiji14 Do you have any idea what it means?

pygmt/__init__.py
```diff
@@ -40,7 +40,7 @@ def print_clib_info():
     Includes the GMT version, default values for parameters, the path to the
     ``libgmt`` shared library, and GMT directories.
     """
-    from .clib import Session
+    from pygmt.clib import Session
```
weiji14, line 43: > This PR is almost done, but we have this error:
> 
> ```
> pygmt/__init__.py:43:4: C0415: Import outside toplevel (pygmt.clib.Session) (import-outside-toplevel)
> ```
> 
> @weiji14 Do you have any idea what it means?

Hmm, not sure why it complains here. I think this complaint means `from pygmt.clib import Session` should be placed at the top (outside of this `print_clib_info` function/module, but I thought we allowed `clib` to be imported at any level though (see https://github.com/GenericMappingTools/pygmt/pull/484/commits/4e76530efafacc95fd9d3dad497977ddbc9493ea#r446575346). Might need to disable the pylint error for this line

```suggestion
    from pygmt.clib import Session  # pylint: disable=import-outside-toplevel
```

seisman: Changing the `.pylintrc` to:
```
allow-any-import-level=pygmt.clib,
                        pygmt.clib.Session,
```
works.

Should we change the `.pylintrc` instead?

weiji14: What version of pylint are we on now? I can't quite make out what my thinking was at https://github.com/GenericMappingTools/pygmt/pull/484#discussion_r446575346...

seisman: `import-outside-toplevel` was introduced in [pylint 2.4.0](http://pylint.pycqa.org/en/latest/whatsnew/changelog.html#what-s-new-in-pylint-2-4-0), but it may have [a bug](https://github.com/PyCQA/pylint/issues/3175), which was [fixed in pylint 2.4.3](http://pylint.pycqa.org/en/latest/whatsnew/changelog.html#what-s-new-in-pylint-2-4-3).

seisman: > What version of pylint are we on now?

The CI is using pylint 2.6.0.

seisman: The following changes work well for pylint 2.6.0:
```diff
-allow-any-import-level=clib,
-                       clib.Session,
+allow-any-import-level=pygmt.clib.Session,
```

weiji14: Ah good, go with that then.

weiji14 (approved):
:rocket:

PR #758, opened 2020-12-21
WIP: Fix random failures in Windows CI 

seisman:
**Description of proposed changes**

All PyGMT tests, excluding a few marked as `xfail` or `xskip`, should pass on all platforms (Linux/macOS/Windows).

However, the tests sometimes fail on Windows. Generally, there are two different types of failures:

1. as reported in https://github.com/GenericMappingTools/pygmt/issues/717, calling the GMT library sometimes raises the error
   `OSError: exception: access violation reading 0x000001E880146FBC`. 
   It's difficult to debug, as none of us are developing PyGMT on Windows.
2. Tests that calling `grdcut` and/or `grdfilter` sometimes fail, with the error messages shown below. It turns out that the file exists but has a zero size, so it's an invalid netCDF file. However, it's still puzzling why the file size is zero.

```
WARNING: D:\a\pygmt\pygmt\examples\tutorials\3d-perspective-image.py failed to execute correctly: Traceback (most recent call last):
  File "D:\a\pygmt\pygmt\examples\tutorials\3d-perspective-image.py", line 12, in <module>
    grid = pygmt.datasets.load_earth_relief(resolution="05m", region=[-108, -103, 35, 40])
  File "C:\Miniconda3\envs\test\lib\site-packages\pygmt\helpers\decorators.py", line 411, in new_module
    return module_func(*args, **kwargs)
  File "C:\Miniconda3\envs\test\lib\site-packages\pygmt\datasets\earth_relief.py", line 103, in load_earth_relief
    grid = grdcut(f"@earth_relief_{resolution}{reg}", region=region)
  File "C:\Miniconda3\envs\test\lib\site-packages\pygmt\helpers\decorators.py", line 270, in new_module
    return module_func(*args, **kwargs)
  File "C:\Miniconda3\envs\test\lib\site-packages\pygmt\helpers\decorators.py", line 411, in new_module
    return module_func(*args, **kwargs)
  File "C:\Miniconda3\envs\test\lib\site-packages\pygmt\gridops.py", line 112, in grdcut
    with xr.open_dataarray(outgrid) as dataarray:
  File "C:\Miniconda3\envs\test\lib\site-packages\xarray\backends\api.py", line 701, in open_dataarray
    dataset = open_dataset(
  File "C:\Miniconda3\envs\test\lib\site-packages\xarray\backends\api.py", line 572, in open_dataset
    store = opener(filename_or_obj, **extra_kwargs, **backend_kwargs)
[...]

PR #769, opened 2020-12-28
Update test_coast_aliases

willschlitzer:
The current `test_coast_aliases()` in `test_coast.py` does not include all of the current aliases in PyGMT v0.2.1; I updated the function to include all of the aliases and changed the test format to use `@check_figures_equal`.

### DIFF
--- a/pygmt/tests/test_coast.py
+++ b/pygmt/tests/test_coast.py
@@ -37,23 +37,49 @@ def test_coast_iceland():
     return fig_ref, fig_test
 
 
-@pytest.mark.mpl_image_compare
+@check_figures_equal()
 def test_coast_aliases():
     "Test that all aliases work"
-    fig = Figure()
-    fig.coast(
-        region="-30/30/-40/40",
-        projection="m0.1i",
-        frame="afg",
-        rivers="1/1p,black",
-        borders="1/0.5p,-",
-        shorelines="0.25p,white",
-        land="moccasin",
-        water="skyblue",
-        resolution="i",
-        area_thresh=1000,
+    fig_ref, fig_test = Figure(), Figure()
+    fig_ref.coast(
+        R="-30/30/-40/40",
+        J="M25c",
+        B="afg",
+        I="1/1p,black",
+        N="1/0.5p,-",
+        W="0.25p,white",
+        G="moccasin",
+        S="skyblue",
+        D="i",
+        A=1000,
+        L="jMM+c1+w1000k+f+l",
+        U=True,
+        V="q",
+        X="a4c",
+        Y="a10c",
+        p="135/25",
+        t=13,
     )
-    return fig
+    fig_test.coast(
+        region=[-30, 30, -40, 40],  # R
+        projection="M25c",  # J
+        frame="afg",  # B
+        rivers="1/1p,black",  # I
+        borders="1/0.5p,-",  # N
+        shorelines="0.25p,white",  # W
+        land="moccasin",  # G
+        water="skyblue",  # S
+        resolution="i",  # D
+        area_thresh=1000,  # A
+        map_scale="jMM+c1+w1000k+f+l",  # L
+        timestamp=True,  # U
+        verbose="q",  # V
+        xshift="a4c",  # X
+        yshift="a10c",  # Y
+        perspective=[135, 25],  # p
+        transparency=13,  # t
+    )
+    return fig_ref, fig_test
 
 
 @pytest.mark.mpl_image_compare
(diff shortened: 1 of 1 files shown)


seisman:
Following https://github.com/GenericMappingTools/pygmt/issues/771#issuecomment-751854298, I think we don't want the changes in this PR. Am I right?

willschlitzer:
@seisman Sorry if my poor timing of submitting this pull request and #771 at the same time made this confusing. In my opinion, I think the changes in this PR are valid. I'll admit that I'm not too experienced with writing tests, but I think one of the minimum standards for PyGMT tests should be including a test that makes sure all of the aliases are successfully passed to the GMT API. My thought process here is this ensures that any changes in the GMT argument names, or any accidental changes to the PyGMT aliases, are caught. The test already existed, but my assumption is that `Figure.coast` has expanded to include more arguments since the test was written, so I wanted to make sure that tests all of the aliases/parameters that `Figure.coast` accepts.

pygmt/tests/test_coast.py
```diff
@@ -37,23 +37,49 @@ def test_coast_iceland():
...
+        J="M25c",
+        B="afg",
+        I="1/1p,black",
+        N="1/0.5p,-",
+        W="0.25p,white",
+        G="moccasin",
+        S="skyblue",
+        D="i",
+        A=1000,
+        L="jMM+c1+w1000k+f+l",
```
seisman, line 55: `MM` is not a valid anchor. Do you mean `CM`?

willschlitzer: @seisman I did mean `CM`, and changed it in my most recent commit. I don't know enough about how GMT handles a different input, but it looks like `MM` is a valid anchor; both argument variations look the same on the figure. I ran a `check_figures_equal` test for the two and it passed.

seisman: I guess that GMT may default to "CM" if the given anchor can not be recognized.

pygmt/tests/test_coast.py
```diff
@@ -37,23 +37,49 @@ def test_coast_iceland():
...
+        I="1/1p,black",
+        N="1/0.5p,-",
+        W="0.25p,white",
+        G="moccasin",
+        S="skyblue",
+        D="i",
+        A=1000,
+        L="jCM+c1+w1000k+f+l",
+        U=True,
+        V="q",
```
seisman, line 57:
```suggestion
```

pygmt/tests/test_coast.py
```diff
@@ -37,23 +37,49 @@ def test_coast_iceland():
...
+        rivers="1/1p,black",  # I
+        borders="1/0.5p,-",  # N
+        shorelines="0.25p,white",  # W
+        land="moccasin",  # G
+        water="skyblue",  # S
+        resolution="i",  # D
+        area_thresh=1000,  # A
+        map_scale="jCM+c1+w1000k+f+l",  # L
+        timestamp=True,  # U
+        verbose="q",  # V
```
seisman, line 76:
```suggestion
```

seisman (commented):
The test looks OK, but it's a bad idea to use **-Vq** which disables warnings and errors.

pygmt/tests/test_coast.py
willschlitzer: I included that to make sure that all of the aliases are tested; I'm guessing we don't need to test `verbose`?

seisman (approved):
@willschlitzer The test looks good to me.

As you're now a member of the python-maintainers team, you should be able to merge the  PR when the CI tests finish. Please try it.

It's OK if one or two Windows CI tests fail. That's a known issue (https://github.com/GenericMappingTools/pygmt/pull/758) and we don't have a solution yet .

willschlitzer:
@seisman sounds good!

pygmt/tests/test_coast.py
```diff
@@ -37,23 +37,47 @@ def test_coast_iceland():
...
+        B="afg",
+        I="1/1p,black",
+        N="1/0.5p,-",
+        W="0.25p,white",
+        G="moccasin",
+        S="skyblue",
+        D="i",
+        A=1000,
+        L="jCM+c1+w1000k+f+l",
+        U=True,
```
seisman, line 56: @willschlitzer It's problematic to use `U=True` here, as it creates a timestamp and the timestep can be different for the reference and test images, which can cause failure.

seisman:
@willschlitzer It's problematic to use `U=True` here, as it creates a timestamp and the timestep can be different for the reference and test images, which can cause failure.

PR #772, opened 2020-12-28
Add an example for plotting multi-parameter symbols

michaelgrund:
I added rot_rectangle.py to have an example showing the use of rotated rectangles (which I heavily used during my PhD to plot splitting parameters in map view).


seisman:
> plot splitting parameters

@michaelgrund Do you mean S-wave splitting results? Orientation for the fast axis and length for the splitting time?

michaelgrund:
> 
> 
> > plot splitting parameters
> 
> @michaelgrund Do you mean S-wave splitting results? Orientation for the fast axis and length for the splitting time?
> 
> EDIT: I just read your example. I think the answer is yes. My question is, what does the width of the rectangle mean? I thought most people plot the S-wave splitting results by just plotting lines or using **velo**.

@seisman Yes, I mean S-wave splitting. Depends on how you want to show the results (orientation for the fast axis and length for the delay time), either as simple line as you suggested (I also used that various times) or as a "bar" with a specific fill and edge color which can be done with the rotated rectangle. The provided example is based on the latter one and the width has no physical meaning (only length and orientation). Of course we could remove that note about S-wave splitting to avoid confusion. What do you mean?

seisman:
GMT can plot so many different symbols (see https://docs.generic-mapping-tools.org/dev/plot.html), and I'm hesitant to add an example for each symbol (that would be more than 20 examples for symbols). 

I think it would be better if we can show some symbols in one example. For example, the following symbols may have similar input and could be grouped in one example:

image

michaelgrund:
> 
> 
> GMT can plot so many different symbols (see https://docs.generic-mapping-tools.org/dev/plot.html), and I'm hesitant to add an example for each symbol (that would be more than 20 examples for symbols).
> 
> I think it would be better if we can show some symbols in one example. For example, the following symbols may have similar input and could be grouped in one example:
> 
> image

totally agree with you @seisman, will prepare are corresponding example ;)

examples/gallery/plot/rot_rectangle.py
```diff
@@ -0,0 +1,37 @@
+"""
```
seisman, line 1: Please remove this example.

examples/gallery/plot/multi_parameters.py
```diff
@@ -0,0 +1,62 @@
+"""
+Multi-parameter symbols
+----------------
```
seisman, line 3:
```suggestion
Multi-parameter symbols
-------------------------
```

examples/gallery/plot/multi_parameters.py
```diff
@@ -0,0 +1,62 @@
...
+
+The :meth:`pygmt.Figure.plot` method can plot individual multi-parameter symbols by passing 
+the corresponding shortcuts listed below to the ``style`` option. Additionally, we must define 
+the required parameters in a numpy array (or use an appropriately formatted input file) and pass it
```
seisman, line 7: This is not 100% correct. A 2D list should also work.

examples/gallery/plot/multi_parameters.py
```diff
@@ -0,0 +1,62 @@
+"""
```
seisman, line 1: Please rename the file to a better name, maybe "multi-parameter-symbols"?

examples/gallery/plot/multi_parameters.py
michaelgrund: > 
> 
> This is not 100% correct. A 2D list should also work.

It needs to be always a 2D array right?

examples/gallery/plot/multi_parameters.py
michaelgrund: done!

examples/gallery/plot/multi_parameters.py
seisman: > It needs to be always a 2D array, right?

I think so.

michaelgrund: > 
> 
> > It needs to be always a 2D array, right?
> 
> I think so.

ok fine

examples/gallery/plot/multi-parameter-symbols.py
```diff
@@ -0,0 +1,63 @@
+"""
+Multi-parameter symbols
+-------------------------
+
+The :meth:`pygmt.Figure.plot` method can plot individual multi-parameter symbols by passing 
+the corresponding shortcuts listed below to the ``style`` option. Additionally, we must define 
+the required parameters in a 2d numpy array (``np.array([[parameters]])`` for a single symbol 
+or ``np.array([[parameters_1],[parameters_2],[parameters_i]])`` for several ones) or use an 
+appropriately formatted input file and pass it to ``data``.
```
seisman, line 9:
```suggestion
The :meth:`pygmt.Figure.plot` method can plot individual multi-parameter symbols by passing 
the corresponding shortcuts listed below to the ``style`` option. Additionally, we must define 
the required parameters in a 2d list or numpy array (``[[parameters]]`` for a single symbol 
or ``[[parameters_1],[parameters_2],[parameters_i]]`` for several ones) or use an 
appropriately formatted input file and pass it to ``data``. 
```

examples/gallery/plot/multi-parameter-symbols.py
```diff
@@ -0,0 +1,63 @@
...
+appropriately formatted input file and pass it to ``data``. 
+
+The following symbols are available:
+
+- **e**: ellipse, ``np.array([[lon, lat, direction, major_axis, minor_axis]])``
+- **j**: rotated rectangle, ``np.array([[lon, lat, direction, width, height]])``
+- **r**: rectangle, ``np.array([[lon, lat, width, height]])``
+- **R**: rounded rectangle, ``np.array([[lon, lat, width, height, radius]])``
+- **w**: pie wedge, ``np.array([[lon, lat, radius, startdir, stopdir]])``, the last two arguments are 
+  directions given in degrees counter-clockwise from horizontal
```
seisman, line 18:
```suggestion
- **e**: ellipse, ``[[lon, lat, direction, major_axis, minor_axis]]``
- **j**: rotated rectangle, ``[[lon, lat, direction, width, height]]``
- **r**: rectangle, ``[[lon, lat, width, height]]``
- **R**: rounded rectangle, ``[[lon, lat, width, height, radius]]``
- **w**: pie wedge, ``[[lon, lat, radius, startdir, stopdir]]``, the last two arguments are 
  directions given in degrees counter-clockwise from horizontal 
```

Issue #631, opened 2020-09-23
How to reproduce original GMT arguments in PyGMT documentation


weiji14:
**Description of the desired feature**

The GMT documentation has many instances where bold and italicized fonts are mixed, looking something like:

<pre><b>+n</b><i>args</i></pre>

The PyGMT documentation has been rather inconsistent with this (see mentions at https://github.com/GenericMappingTools/pygmt/pull/525#discussion_r454414687 and https://github.com/GenericMappingTools/pygmt/pull/620#discussion_r493088431).

What we should do is to standardize how we want **`+n`** *`args`* to be written in the documentation. This is a bit complicated, but boils down to (at least) two options:

1. `r" **+n**\ *args* "`. Use a raw-string in Python, allowing us to follow upstream GMT (as per https://github.com/GenericMappingTools/pygmt/pull/620#discussion_r493088431)

2. `" **+n**\\ *args* "`. Use a double-backslash to avoid the flake8 error `W605 invalid escape sequence '\ '`.

An alternative would be to find a way to ignore the flake8 error somehow for the ReST docstrings.

Also, should we find a way to put these in code-formatted blocks (i.e. with backticks), as mentioned at https://github.com/GenericMappingTools/pygmt/pull/525/files#r454746311? 

**Are you willing to help implement and maintain this feature?** Somewhat short on time, but will look into it.

References:
- https://meta.stackexchange.com/questions/183610/how-to-combine-bold-and-code-sample-in-markdown

seisman:
This is what [PEP-257](https://www.python.org/dev/peps/pep-0257/) says:

> For consistency, always use """triple double quotes""" around docstrings. Use r"""raw triple double quotes""" if you use any backslashes in your docstrings. For Unicode docstrings, use u"""Unicode triple-quoted strings""".

So, instead of using `r" **+n**\ *args* "` for each individual cases, we should use `r`  for the whole docstring, e.g., 

```
def function():
    r"""
    A long docstring string.

    We now can write doctsrings with backslashes **+n**\ *args*
    """
```

seisman:
As pointed out by @weiji14, the PyGMT documentation is inconsistently using _italics_, **bold**, and `code` strings.

I think we can following these conventions:

1. Use `code` for Python function arguments
2. Use **bold** and *italics* for GMT syntax like **+n***args*, in which bold **+n** means it should be typed as it is, and italic *args* should be given a specific value.

[...]

PR #774, opened 2020-12-29
Add a gallery example of setting line colors with a custom CPT

willschlitzer:
As discussed in #493 (and using code provided by @weiji14), this is a gallery example that shows how to use a custom CPT to set specific z-values for lines.

### DIFF
--- a/examples/gallery/line/line-custom-cpt.py
+++ b/examples/gallery/line/line-custom-cpt.py
@@ -1,0 +1,31 @@
+"""
+Line colors with a custom CPT
+-----------------------------
+
+The color of the lines made by :meth:`pygmt.Figure.plot` can be set according to a
+custom CPT and assigned with the *pen* argument.
+
+The custom CPT can be used by setting the plot command's `cmap` argument to *True*. The
+`zvalue` argument sets the z-value (color) to be used from the custom CPT, and the line
+color is set as the z-value by using **+z** when setting the `pen` color.
+
+"""
+
+import numpy as np
+import pygmt
+
+# Create a custom CPT with the batlow CPT and 10 discrete z-values (colors)
+pygmt.makecpt(cmap="batlow", series=[0, 10, 1])
+# Create a list of values between 20 and 30 with at 0.2 intervals
+x = np.arange(start=20, stop=30, step=0.2)
+
+fig = pygmt.Figure()
+fig.basemap(frame=["WSne", "af"], region=[20, 30, -10, 10])
+
+# Plot 10 lines and set a different z-value for each line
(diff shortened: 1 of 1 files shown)


examples/gallery/line/line-custom-cpt.py
```diff
@@ -0,0 +1,31 @@
+"""
+Line colors with a custom CPT
+-----------------------------
+
+The color of the lines made by :meth:`pygmt.Figure.plot` can be set according to a
+custom CPT and assigned with the *pen* argument.
+
+The custom CPT can be used by setting the plot command's `cmap` argument to *True*. The
+`zvalue` argument sets the z-value (color) to be used from the custom CPT, and the line
+color is set as the z-value by using **+z** when setting the `pen` color.
```
seisman, line 10: Following the convention in #631.
```suggestion
The custom CPT can be used by setting the plot command's ``cmap`` argument to ``True``. The
``zvalue`` argument sets the z-value (color) to be used from the custom CPT, and the line
color is set as the z-value by using **+z** when setting the ``pen`` color.
```

examples/gallery/line/line-custom-cpt.py
```diff
@@ -0,0 +1,31 @@
+"""
+Line colors with a custom CPT
+-----------------------------
+
+The color of the lines made by :meth:`pygmt.Figure.plot` can be set according to a
+custom CPT and assigned with the *pen* argument.
```
seisman, line 6:
```suggestion
custom CPT and assigned with the ``pen`` argument.
```

examples/gallery/line/line-custom-cpt.py
```diff
@@ -0,0 +1,31 @@
...
+`zvalue` argument sets the z-value (color) to be used from the custom CPT, and the line
+color is set as the z-value by using **+z** when setting the `pen` color.
+
+"""
+
+import numpy as np
+import pygmt
+
+# Create a custom CPT with the batlow CPT and 10 discrete z-values (colors)
+pygmt.makecpt(cmap="batlow", series=[0, 10, 1])
```
seisman, line 18: Please move the `pygmt.makecpt` after calling `pygmt.Figure()`. The current `pygmt.makecpt` creates a CPT for the current session. It works for the current example but may lead to some confusion when users try to adapt the example to more complicated ones. 

See https://docs.generic-mapping-tools.org/dev/makecpt.html#note-on-cpts-in-modern-mode for details.

willschlitzer: Done

seisman (approved):

PR #775, opened 2020-12-30
Update CONTRIBUTING.md with formatting conventions

willschlitzer:
As discussed in #631, setting the contributing guidelines to show the PyGMT documentation standard for formatting Python and GMT arguments.


CONTRIBUTING.md
```diff
@@ -159,6 +159,22 @@ Guidelines for a good tutorial:
...
 the documentation.
 
+### Example code standards
+
+When editing documentation, use the following standards to demonstrate the example code:
+
+1. Python arguments, such as import statements, Boolean expressions, and function 
+   arguments should be wrapped as ``code`` by using \`\` on both sides of the code.<br>
+   Example: \`\`import pygmt\`\` results in ``import pygmt``
+<br><br>
```
seisman, line 169: Why do you need `<br>` here?

willschlitzer: When PyCharm would render the Markdown, there wouldn't be a line break without `<br>`.

seisman: I don't think it's required by standard or github markdown.

willschlitzer: Removed the `<br>` tags

CONTRIBUTING.md
```diff
@@ -161,6 +161,21 @@ Guidelines for a good tutorial:
...
+When editing documentation, use the following standards to demonstrate the example code:
+
+1. Python arguments, such as import statements, Boolean expressions, and function 
+   arguments should be wrapped as ``code`` by using \`\` on both sides of the code.
+   Example: \`\`import pygmt\`\` results in ``import pygmt``
+   
+2.  Literal GMT arguments should be **bold** by wrapping the arguments with \*\* 
+    (two asterisks) on both sides. The argument description should be in *italicized* 
+    with \* (single asterisk) on both sides.
+    Example: \*\*+l\*\*\*label\* results in **+l***label*
```
seisman, line 175: ReST's `**+l**\ *label*` results in markdown's **+l***label*. Perhaps we can just use `code` to make it simpler.

```suggestion
    Example: `**+l**\ *label*` results in **+l***label*
```

seisman (approved):

PR #777, opened 2020-12-30
Add a gallery example for the fig.image() method

michaelgrund:
Here's a simple gallery example that shows how to use the fig.image() method. As input I used the GMT logo from the main webpage, hope that's fine.

### DIFF
--- a/examples/gallery/plot/image.py
+++ b/examples/gallery/plot/image.py
@@ -16,7 +16,7 @@ import pygmt
 
 fig = pygmt.Figure()
 
-fig.basemap(region=[0, 2, 0, 2], projection="x3c", frame=True)
+fig.basemap(region=[0, 2, 0, 2], projection="X6c", frame=True)
 
 # place and center the GMT logo from the GMT website to the position 1/1
 # on a basemap and draw a rectangular border around the image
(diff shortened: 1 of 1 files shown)


examples/gallery/plot/image.py
```diff
@@ -0,0 +1,29 @@
...
+
+For more advanced style options, see the full option list 
+at :gmt-docs:`image.html`.
+"""
+
+import pygmt
+
+fig = pygmt.Figure()
+
+fig.basemap(region=[0, 2, 0, 2], projection="x3c", frame=True)
```
seisman, line 19: I feel that `X6c` is more readable than `x3c`.
```suggestion
fig.basemap(region=[0, 2, 0, 2], projection="X6c", frame=True)
```

examples/gallery/plot/image.py
```diff
@@ -0,0 +1,29 @@
...
+
+# place and center the GMT logo from the GMT website to the position 1/1
+# on a basemap and draw a rectangular border around the image
+fig.image(
+    imagefile="https://www.generic-mapping-tools.org/_static/gmt-logo.png",
+    position="g1/1+w1.5i+jCM",
```
seisman, line 25: Use SI units, not US units.
```suggestion
    position="g1/1+w3c+jCM",
```

seisman (approved):

PR #781, opened 2021-01-02
Add lakes alias to Figure.coast()

willschlitzer:
Adding the `lakes` alias for the **-C** argument to `Figure.coast()`.

### DIFF
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -114,9 +114,10 @@ class BasePlotting:
         lakes : str or list
             *fill*\ [**+l**\ |**+r**\ ]
             Set the shade, color, or pattern for lakes and river-lakes. The
-            default is the fill chosen for wet areas set by the ``water`` argument.
-            Optionally, specify separate fills by appending **+l** for lakes or
-            **+r** for river-lakes, and passing multiple strings in a list.
+            default is the fill chosen for wet areas set by the ``water``
+            argument. Optionally, specify separate fills by appending
+            **+l** for lakes or **+r** for river-lakes, and passing multiple
+            strings in a list.
         resolution : str
             Selects the resolution of the data set to use ((f)ull, (h)igh,
             (i)ntermediate, (l)ow, and (c)rude).
--- a/pygmt/base_plotting.py
+++ b/pygmt/base_plotting.py
@@ -60,6 +60,7 @@ class BasePlotting:
         R="region",
         J="projection",
         A="area_thresh",
+        C="lakes",
         B="frame",
         D="resolution",
         I="rivers",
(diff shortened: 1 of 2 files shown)


pygmt/base_plotting.py
```diff
@@ -110,8 +111,12 @@ def coast(self, **kwargs):
             hierarchical level that is lower than min_level or higher than
             max_level will not be plotted.
         {B}
-        C : str
-            Set the shade, color, or pattern for lakes and river-lakes.
+        lakes : str or list
+            *fill*\ [**+l**\ |**+r**\ ]
+            Set the shade, color, or pattern for lakes and river-lakes. The
+            default is the fill chosen for wet areas set by the ``water``.
```
seisman, line 117:
```suggestion
            default is the fill chosen for wet areas set by the ``water`` argument.
```

pygmt/base_plotting.py
```diff
@@ -110,8 +111,12 @@ def coast(self, **kwargs):
...
+            Optionally, specify separate fills by appending +l for lakes or
+            +r for river-lakes, and passing multiple strings in a list.
```
seisman, line 119:
```suggestion
            Optionally, specify separate fills by appending **+l** for lakes or
            **+r** for river-lakes, and passing multiple strings in a list.
```

seisman:
Also see some discussions in https://github.com/GenericMappingTools/pygmt/pull/474#discussion_r434281630.

Ping @GenericMappingTools/python for comments on the argument name.

willschlitzer:
@seisman Looking at that PR, I'm not sure what the issue with using a long-name alias for `C` is.  It seems like the main concern is backwards compatibility, but wouldn't scripts still be able to use `C` instead of the alias?

seisman:
> It seems like the main concern is backwards compatibility, but wouldn't scripts still be able to use `C` instead of the alias?

Yes, `C` always works.

We need to be very careful about choosing these long names. Sometimes we may find that the long names are inappropriate and have to change them to other names, which may break backward compatibility. For example, @weiji14 and I were thinking about changing `sizes` to `size` in `plot()` (https://github.com/GenericMappingTools/pygmt/pull/666#issuecomment-716084019).

seisman:
Just checked the documentation of `coast()`, I think `lakes` is a good name, considering that we already use "rivers", "shorelines" and "borders".

seisman (approved):
OK to merge after you fix this style issue:
```
pygmt/base_plotting.py:117:80: W505 doc line too long (83 > 79 characters)
```

seisman:
@willschlitzer When merging a PR, it would be better to simplify the commit description. For example, this is how this PR looks like after merging into the master branch. 

image

A better one is:

image

willschlitzer:
@seisman Sounds good; my apologies for just leaving the defaults in.

PR #783, opened 2021-01-02
Update COMMON_OPTIONS arguments to GMT standard

willschlitzer:
Update the argument doc strings in the COMMON_OPTIONS dictionary in decorators.py to the standard set in #631.


willschlitzer:
I can't figure out why this is failing so many tests, as it is changing the content of strings and no functions. It appears to be rendering correctly when it is built in Sphinx.

seisman:
One doctest in `pygmt.helpers.decorators.fmt_docstring` fails.
```
=================================== FAILURES ===================================
_______________ [doctest] pygmt.helpers.decorators.fmt_docstring _______________
160     ...
161     ...     Parameters
162     ...     ----------
163     ...     {R}
164     ...     {J}
165     ...
166     ...     {aliases}
167     ...     '''
168     ...     pass
169     >>> print(gmtinfo.__doc__)
Differences (unified diff with -expected +actual):
    @@ -5,10 +5,11 @@
     ----------
     region : str or list
    -    *Required if this is the first plot command*.
    -    ``'xmin/xmax/ymin/ymax[+r][+uunit]'``.
    -    Specify the region of interest.
    +    *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*]
    +    Specify the region of interest. This is a required argument if this 
    +    is the first plot command.
     projection : str
    -    *Required if this is the first plot command*.
    -    Select map projection.
    +    *projection*\ [*projection-specific arguments*\ ]*figure size*
    +    Select map projection. This is a required argument if this 
    +    is the first plot command.
     <BLANKLINE>
     **Aliases:**
```

willschlitzer:
I misinterpreted the `12 xfailed` as test failures then. Regardless, I can't see how this is causing a doctest failure, so I'm not sure how to fix the strings to make it pass.

seisman:
> I misinterpreted the `12 xfailed` as test failures then.

They are "known" failures.

> I can't see how this is causing a doctest failure, so I'm not sure how to fix the strings to make it pass.

You need to update these lines:

https://github.com/GenericMappingTools/pygmt/blob/547a6fc1cf00255fe2e93af97859bae7f9e8b3cc/pygmt/helpers/decorators.py#L172-L180

pygmt/helpers/decorators.py
```diff
@@ -15,13 +15,14 @@
 COMMON_OPTIONS = {
     "R": """\
         region : str or list
-            *Required if this is the first plot command*.
-            ``'xmin/xmax/ymin/ymax[+r][+uunit]'``.
-            Specify the region of interest.""",
+            *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*]
```
weiji14, line 18: As I mentioned in https://github.com/GenericMappingTools/pygmt/issues/631#issuecomment-764968391, we need to find a good, nice-looking way to separate the `[...*unit*]` argument from the `Specify...` text:

image

Either add a fullstop/period (.) as suggested below, or a linebreak like in the GMT docs (does `\n` work?), or something else.

```suggestion
            *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*].
```
For reference, pygmt v0.2.1 looked like this:

image

pygmt/helpers/decorators.py
```diff
@@ -15,13 +15,14 @@
 COMMON_OPTIONS = {
     "R": """\
```
weiji14, line 16:
```suggestion
    "R": r"""\
```
Need to make this a raw r-string to avoid the pylint warning (W605 invalid escape sequence '\ ')

pygmt/helpers/decorators.py
```diff
@@ -15,13 +15,14 @@
...
         region : str or list
-            *Required if this is the first plot command*.
-            ``'xmin/xmax/ymin/ymax[+r][+uunit]'``.
-            Specify the region of interest.""",
+            *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*].
+            Specify the region of interest. This is a required argument if this
+            is the first plot command.""",
     "J": """\
```
weiji14, line 21:
```suggestion
    "J": r"""\
```
Same here, use r-string.

weiji14 (commented):
I reckon this PR is about done, just a few more changes needed.

pygmt/helpers/decorators.py
willschlitzer: It rendered incorrectly when I would have raw strings in both COMMON_OPTIONS and the function doc strings (check out the deployment for 62e8aad at the top).  My guess is some issue with a raw string with escape characters being passed into another raw string.

weiji14: I see, might need to rethink this a bit then or just ignore the `pylint` error if it's harmless.

willschlitzer: Should I add a `pylint: disable=invalid-escape-sequence` (is that the correct wording/syntax?)?

pygmt/helpers/decorators.py
```diff
@@ -15,13 +15,14 @@
 COMMON_OPTIONS = {
     "R": """\
         region : str or list
```
weiji14, line 17: I *think* I might have figured it out, can you try the below? I.e. Use an r-string while removing the backslash and putting the `region : str or list` line right after the triple quotes.

```suggestion
    "R": r"""region : str or list
```

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,91 @@
...
+    *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*].
+    Specify the region of interest. This is a required argument if this
+    is the first plot command.""",
+    "J": r"""projection : str
+    *projection*\ [*projection-specific arguments*\ ]\ *figure size*.
+    Select map projection. This is a required argument if this
+    is the first plot command.""",
+    "B": r"""frame : str or list
+    Set map boundary frame and axes attributes.""",
+    "U": r"""timestamp : bool or str
```
weiji14, line 26: Probably don't need a raw string for the ones without backslashes. Can you revert some of these changes?

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,91 @@
...
+        - **e** - Error messages only
+        - **w** - Warnings [Default]
+        - **t** - Timings (report runtimes for time-intensive algorthms);
+        - **i** - Informational messages (same as "verbose=True")
+        - **c** - Compatibility warnings
+        - **d** - Debugging messages""",
+    "W": r"""pen : str
+    Set pen attributes for lines or the outline of symbols.""",
+    "XY": r"""xshift : str
+    ``[a|c|f|r][xshift]``.
```
weiji14, line 47: These still need to change. Have  look at the GMT documentation source.

willschlitzer: Changed the subsequent docstrings in the commits between 2fe9720 and a82bf6c

pygmt/helpers/decorators.py
willschlitzer: I removed the unnecessary raw strings in a82bf6c

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,94 @@
...
-            be truncated to the maximum cores available). Finally, give a
-            negative number *-n* to select (all - n) cores (or at least 1 if
-            n equals or exceeds all).
-            """,
+    "R": r"""region : str or list
+    *xmin/xmax/ymin/ymax*\ [**+r**\ ][**+u**\ *unit*].
+    Specify the region of interest. This is a required argument if this
+    is the first plot command.""",
+    "J": r"""projection : str
+    *projection*\ [*projection-specific arguments*\ ]\ *figure size*.
```
seisman, line 21: **-J** option is very difficult to read.
image

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,94 @@
...
+    Select map projection. This is a required argument if this
+    is the first plot command.""",
+    "B": """frame : bool or str or list
+    Set map boundary frame and axes attributes.""",
+    "U": """timestamp : bool or str
+    Draw GMT time stamp logo on plot.""",
+    "CPT": """cmap : str
+    File name of a CPT file or ``C='color1,color2[,color3,...]'`` to
```
seisman, line 29:
```suggestion
    File name of a CPT file or a series of comma-separated colors (e.g., *color1*,\ *color2*,\ *color3*) to
```

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,94 @@
...
+    build a linear continuous CPT from those colors automatically.""",
+    "G": """color : str
+    Select color or pattern for filling of symbols or polygons. Default
+    is no fill.""",
+    "V": """verbose : str
+    Select verbosity level [Default is w], which modulates the messages
```
seisman, line 35:
```suggestion
    Select verbosity level [Default is **w**], which modulates the messages
```

pygmt/helpers/decorators.py
```diff
@@ -13,111 +13,94 @@
...
+    "W": """pen : str
+    Set pen attributes for lines or the outline of symbols.""",
+    "XY": r"""xshift : str
+    [**a**\|\ **c**\|\ **f**\|\ **r**\][*xshift*].
+    Shift plot origin in x-direction.
+
+    yshift : str
+    [**a**\|\ **c**\|\ **f**\|\ **r**\][*yshift*].
+    Shift plot origin in y-direction. Full documentation is at
+    :gmt-docs:`gmt.html#xy-full`.""",
```
seisman, line 53: The **Y** option is not correctly rendered.
image

willschlitzer: I tried adding blank lines to create line breaks in bc6f1e8, but that just caused tests to fail. Any idea how to fix this?

willschlitzer:
This problem was fixed by @seisman in #862. Closing this PR.

Issue #771, opened 2020-12-28
Guidance on types of tests to write for plotting functions


willschlitzer:
The testing section on the contributing guide does a good job of explaining how to write tests, but doesn't really explain what should and should not be tested with the plotting functions. Looking through the tests (specifically the ones testing functions in `base_plotting.py`) there doesn't appear to be a standard for testing functions that pass arguments to the GMT API. My limited understanding makes me think the most important thing to test is the function's ability to take arguments for its aliases and pass them to GMT. There are some tests (such as `test_basemap_winkel_tripel()` and `test_basemap_power_axis()` in `test_basemap.py` and many of the tests in `test_colorbar.py`) that test passing different arguments to the same parameters.  While I understand the importance of testing that these will return functions, any errors seem like they would be upstream issues with GMT and not an issue with PyGMT itself. In an effort to standardize writing tests and reducing the amount of time spent testing, could there be some community consensus (and formal guidance) on how each wrapped GMT function should be tested?

I'm happy to implement/change tests once a standard has been decided.

weiji14:
Yeah, we do need to spell this out. I would think of testing as a spectrum from:

1. Absolute minimum - every line of code is covered by a test, see https://codecov.io/gh/GenericMappingTools/pygmt. Note that this does not even need to cover all aliases! Just bare functionality.

2. Full stress testing - Unit tests checking every alias, every possible combination of arguments, hypothesis/mutation testing, behavioural-driven/integration tests, you name it.

Now you'll probably realize that we fall somewhere in the middle, for practical reasons. Not everyone knows how to write tests, hence the range of style. If I could mention 2 things about testing in PyGMT:

1. Don't test everything that GMT already tests (or that you think it tests, e.g. funky combinations of arguments). There are exceptions to this, such as testing to cover against regressions, and testing important functions used by thousands of people (e.g. `plot`).

2. Do test the Python specific stuff. The virtualfile mechanism, numpy/pandas/xarray input/output is a prime example. 

Above all, remember the golden rule of testing, which is to not delete or change tests unless absolutely necessary. These tests may be sacredly guarding some obscure logic and we want to ensure things work years into thd future.

That said, this is just one perspective, and not a fixed rule. I appreciate this being brought up, but you'll realize that there's no hard and fast rules, only guidelines.

seisman:
[...]

PR #796, opened 2021-01-21
Add guidelines for types of tests to write

willschlitzer:
As discussed in #771, this PR adds recommended guidelines to CONTRIBUTING.md for types of tests to write. It emphasizes testing on Python-specific code over GMT aliases and functions.


willschlitzer:
@weiji14 @seisman I added the recommendations we discussed, but I'm still a little confused on how we should address testing aliases and the basic creation of plots. I could see how someone could get confused looking at the testing guidelines, which don't recommend focusing on GMT aliases and functions, and then seeing that the majority of test files focus on testing the creation of plots and how the aliases work.

seisman:
> then seeing that the majority of test files focus on testing the creation of plots and how the aliases work.

There may be some misunderstanding here. Could you point out some tests which you think don't follow the guidelines?

willschlitzer:
@seisman I probably should have specified that I'm looking at the plotting functions, as the majority of the tests that work with GMT under the hood don't focus on testing plotting aliases. Looking at `test_colorbar.py` there are a lot of tests passing different location inputs, `test_coast.py` has a lot of tests for the different arguments (some of which I've written), `test_basemap.py` has numerous tests creating different types of plots. I'm not opposed to these types of tests, but I want to avoid any confusion with the guidelines suggesting that those types of tests are unnecessary.

I

seisman:
> @seisman I probably should have specified that I'm looking at the plotting functions, as the majority of the tests that work with GMT under the hood don't focus on testing plotting aliases. Looking at `test_colorbar.py` there are a lot of tests passing different location inputs, `test_coast.py` has a lot of tests for the different arguments (some of which I've written), `test_basemap.py` has numerous tests creating different types of plots. I'm not opposed to these types of tests, but I want to avoid any confusion with the guidelines suggesting that those types of tests are unnecessary.

I see your points. Some of the tests were written years ago, and we didn't have any guidelines at that time (we're still learning  to make the project better and developer-friendly). 

If you look at the core codes of the `colorbar()` function, you will realize that the colorbar wrapper is pretty simple and almost has no Python-specific codes. However, we still need/want to add a few tests to make sure `colorbar()` works as expected. But I agree that some of the tests may be unnecessary.

seisman (approved):

PR #797, opened 2021-01-21
Use args_in_kwargs for basemap error raising

willschlitzer:
Modify the `basemap` function in `Figure` to use `args_in_kwargs()` to determine if one of the required arguments is passed. This also adds a required arguments test to `test_basemap.py`.


willschlitzer:
@seisman Should the test functions that aren't using `@check_figures_equal` use the long-form aliases or the single-character GMT aliases?

seisman (approved):

seisman:
> Should the test functions that aren't using `@check_figures_equal` use the long-form arguments or the single-character GMT aliases?

We should use long-form arguments unless we're testing the GMT aliases.

Issue #372, opened 2019-11-06
Allow for "makecpt" style colormap functionality in grdimage and colobar


MarkWieczorek:
In classic gmt the function `makecpt` is used to create a "cpt" file that acts as a color map for plotting data in `grdimage` and in plotting a colorbar.

In its present state, pygmt does not (to the best of my knowledge) allow for creating a custom colormap for use in both `grdimage` and `colorbar`, which greatly limits the utility of pygmt. Such functionality could be achieved in the following way:

```
fig.makecpt(limits=[min, max], cmap=..., transparency=False, cpt='name of temporary output cpt file')
fig.grdimage(cpt='name of temporary cpt file')
fig.colorbar(cpt='name of temporary cpt file')
```

In practice, a single image could have several cpt files associated with it.

weiji14:
It's not very well documented, but we do have a [makecpt](https://www.pygmt.org/dev/api/generated/pygmt.makecpt.html#pygmt.makecpt) function as of #329. It actually works via `pygmt.makecpt` rather than `fig.makecpt`. Also GMT6 modern mode (which PyGMT uses) allows you to create a hidden Color Palette Table (CPT) and use it in subsequent commands automatically (see https://docs.generic-mapping-tools.org/6.0/makecpt.html#note-on-cpts-in-modern-mode) so we don't actually need to set the cpt in `grdimage`/`colorbar` (if this sounds confusing, let me know).

The way you would do this in your code snippet then is as follows:

```python
pygmt.makecpt(series=[min, max], cmap=...)
fig.grdimage(...)
fig.colorbar(...)
```

MarkWieczorek:
After playing around with `pygmt.makecpt`, I have run into two problems.

1. After you create a colormap with `makecpt`, it is not possible to change the session colormap to a different one. Creating a new colormap has no effect on any subsequent figure. The only way to make a new image with a different colormap is to kill the python kernel and start over again. This occurs with both the terminal and notebook environments.

2. `makecpt` gives you the option to save the generated colormap to a file, however, it would be more useful if you could instead save it to a python variable, and then specify this variable in subsequent calls to grdimage and colorbar. It is hard for me to imagine the case where you would want to save the cpt data to a file.

weiji14:
I agree that `pygmt.makecpt` could definitely be a lot better, it's not intuitive the way things work.

[...]

PR #803, opened 2021-01-23
Wrap grd2cpt

willschlitzer:
Wraps the GMT function `grd2cpt`. Code and documentation borrowed heavily from `grdcut` and `makecpt`.


willschlitzer:
Screenshot from 2021-01-23 16-17-57
For reasons I'm not sure, I'm getting a `GMTImageComparisonFailure: images not close` error when I run `test_grd2cpt_grdimage`.  I'm confused because the code for both `fig_ref` and `fig_test` are identical (as I copy and pasted to make sure I didn't have a typo). When I ran the same code in a Jupyter notebook, I also had images that looked very different. Any idea what is causing this inconsistency?

seisman:
I haven't checked it carefully, but it may be related to https://github.com/GenericMappingTools/pygmt/issues/372

willschlitzer:
May have solved the issue. I'm guessing it was the order that I was calling the different figures and `grd2cpt`. I established `fig_ref` and `fig_test` using `Figure.basemap` prior to calling `grd2cpt`, and then called `Figure.grdimage` afterwards. The image and colorbar came out the same and the test passed.

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,342 @@
...
+    # Use single-character arguments for the reference image
+    fig_ref.basemap(B="a", J="W0/15c", R="d")
+    grd2cpt(grid=grid)
+    fig_ref.grdimage(grid=grid)
+    fig_ref.colorbar(B="a2000")
+    fig_test.basemap(frame="a", projection="W0/15c", region="d")
+    grd2cpt(grid=grid)
+    fig_test.grdimage(grid=grid)
+    fig_test.colorbar(frame="a2000")
+    return fig_ref, fig_test
```
weiji14, line 58: Just one quick suggestion on unit testing. Best thing would be to **only** have `grd2cpt` in the test function and exclude other functions (e.g. `grdimage`, `basemap`, etc). That way our tests will run fast (`grdimage` can be slow) and changes other functions like `basemap` won't affect the `grd2cpt` test.

Personally would be ok to write the tests with just `colorbar` if possible (as with the newer `makecpt` tests written at https://github.com/GenericMappingTools/pygmt/blame/1303188ab7d777085a01b468606244e2f4f39d05/pygmt/tests/test_makecpt.py#L187-L222). But if there's a way to write a 'pure' `grd2cpt` only test using something like `output=temp.cpt`, then go for it!

willschlitzer: I removed `grdimage` from the tests, and it appears to be about 1 second faster to run the file (6 seconds vs. 7 seconds).

weiji14:
Cc @lhoupert, I remember you wanted a `grd2cpt` function at https://github.com/GenericMappingTools/pygmt/issues/508#issuecomment-656027462, do you want to test this branch out using `pip install git+https://github.com/willschlitzer/pygmt.git@wrap-grd2cpt` and add your two cents?

```
import pygmt

pygmt.grd2cpt(...)
```

pygmt/gridops.py
```diff
@@ -254,3 +254,171 @@ def grdfilter(grid, **kwargs):
...
+        a categorical color table, like ``cmap='categorical'``.
+    cyclic : bool
+        Produce a wrapped (cyclic) color table that endlessly repeats its
+        range. Note that ``cyclic=True`` cannot be set together with
+        ``categorical=True``.
+    {V}
+    """
+    kind = data_kind(grid)
+
+    with GMTTempFile(suffix=".nc"):
```
seisman, line 404: I don't understand, why `GMTTempFile` is needed here?

willschlitzer: I had copied it from another function and didn't realize I never used it. I removed it in 81e6e90 and all of the tests are passing.

willschlitzer:
I'm working on moving this function to grd2cpt.py.

doc/api/index.rst
```diff
@@ -74,6 +74,7 @@ Operations on grids:
     grdcut
     grdfilter
     grdtrack
+    grd2cpt
```
weiji14, line 77: Maybe shift this to be under the `Color palette table generation:` section (above `makecpt`).

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,316 @@
...
+    fig_ref.basemap(R=region, J="X15c", B="a")
+    grd2cpt(grid=grid, C="rainbow")
+    fig_ref.plot(
+        x=points[:, 0],
+        y=points[:, 1],
+        G=points[:, 2],
+        R=region,
+        S="c1c",
+        C=True,
+    )
```
weiji14, line 75:
```suggestion
    fig_ref.plot(
        x=points[:, 0], y=points[:, 1], G=points[:, 2], R=region, S="c1c", C=True
    )
```
This can fit on one line.

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    return [10, 70, -5, 10]
+
+
+@pytest.fixture(scope="module", name="grid")
+def fixture_grid():
+    """
+    Load the grid data from the sample earth_relief file.
+    """
+    return load_earth_relief(registration="gridline")
```
weiji14, line 39:
```suggestion
    return load_earth_relief()
```
No need to hardcode `registration="gridline"` here. The tests in [`test_makecpt.py`](https://github.com/GenericMappingTools/pygmt/blob/v0.2.1/pygmt/tests/test_makecpt.py#L34) uses "gridline" registration because it was the default in GMT 6.0 and we didn't want to update the baseline images for the tests using `@pytest.mark.mpl_image_compare` (see #451 for context). Default for GMT 6.1 (what we're using now) is "pixel" registration (see also longer discussion at #487 on what those two mean).

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,316 @@
...
+@check_figures_equal()
+def test_grd2cpt(grid):
+    """
+    Test creating a CPT with grd2cpt to create a CPT based off a grid input and
+    plot it with a color bar.
+    """
+    fig_ref, fig_test = Figure(), Figure()
+    # Use single-character arguments for the reference image
+    fig_ref.basemap(B="a", J="W0/15c", R="d")
+    grd2cpt(grid=grid)
```
weiji14, line 51:
```suggestion
    grd2cpt(grid="@earth_relief_01d")
```

This change should cover the missing code coverage in https://github.com/GenericMappingTools/pygmt/pull/803/checks?check_run_id=1825050187.

weiji14 (commented):
> I'm working on moving this function to grd2cpt.py.

Cool, will wait for the move to complete before reviewing `def grd2cpt`. In the meantime, I've looked through the tests and they look comprehensive enough (just a few minor comments) with each one averaging 1.3-1.5 seconds of runtime on my local machine.

doc/api/index.rst
willschlitzer: Moved it in 1812b93

willschlitzer:
@GenericMappingTools/python I have moved grd2cpt() to grd2cpt.py

lhoupert:
> Cc @lhoupert, I remember you wanted a `grd2cpt` function at [#508 (comment)](https://github.com/GenericMappingTools/pygmt/issues/508#issuecomment-656027462), do you want to test this branch out using `pip install git+https://github.com/willschlitzer/pygmt.git@wrap-grd2cpt` and add your two cents?
> 
> ```python
> import pygmt
> 
> fig = pygmt.Figure()
> fig.basemap(frame=True)
> grd2cpt(grid=grid, cmap="rainbow")
> fig.grdimage(grid=grid, cmap=True)
> fig.show()
> ```

HI @weiji14 , sorry I will not have time to do anything about that over the next two weeks... I am moving to a new job and I don't have my new computer yet 😆

willschlitzer:
@GenericMappingTools/python Could I get a review/feedback for this pull request? It should be the last priority function to wrap for v0.3.

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,180 @@
...
+    :gmt-term:`COLOR_FOREGROUND`, and :gmt-term:`COLOR_NAN` from the
+    :gmt-docs:`gmt.conf <gmt.conf>` file or the command line will be used. This
+    default behavior can be overruled using the options *background*,
+    *overrule_bg* or *no_bg*.
+
+    The color model (RGB, HSV or CMYK) of the palette created by **grdcpt**
+    will be the same as specified in the header of the master CPT. When there
+    is no :gmt-term:`COLOR_MODEL` entry in the master CPT, the
+    :gmt-term:`COLOR_MODEL` specified in the :gmt-docs:`gmt.conf <gmt.conf>`
+    file or on the command line will be used.
```
weiji14, line 69:
```suggestion
    The color model (RGB, HSV or CMYK) of the palette created by
    :meth:`pygmt.grd2cpt` will be the same as specified in the header of the
    master CPT. When there is no :gmt-term:`COLOR_MODEL` entry in the master
    CPT, the :gmt-term:`COLOR_MODEL` specified in the
    :gmt-docs:`gmt.conf <gmt.conf>` file or the *color_model* option will be
    used.
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,180 @@
...
+
+    This is a module that will help you make static color palette tables
+    (CPTs). By default, the CPT will simply be saved to the current session,
+    but you can use *output* to save it to a file. You define an equidistant
+    set of contour intervals or pass your own z-table or list, and create a new
+    CPT based on an existing master (dynamic) CPT. The resulting CPT can be
+    reversed relative to the master cpt, and can be made continuous or
+    discrete. For color tables beyond the standard GMT offerings, visit
+    `cpt-city <http://soliton.vm.bytemark.co.uk/pub/cpt-city/>`_ and
+    `Scientific Colour-Maps <http://www.fabiocrameri.ch/colourmaps.php>`_.
```
weiji14, line 50:
```suggestion
    but you can use *output* to save it to a file. The CPT is based on an
    existing dynamic master CPT of your choice, and the mapping from data value
    to colors is through the data's cumulative distribution function (CDF), so
    that the colors are histogram equalized. Thus if the grid(s) and the
    resulting CPT are used in :meth:`pygmt.Figure.grdimage` with a linear
    projection, the colors will be uniformly distributed in area on the plot.
    Let z be the data values in the grid. Define CDF(Z) = (# of z < Z) / (# of
    z in grid). (NaNs are ignored). These z-values are then normalized to the
    master CPT and colors are sampled at the desired intervals.
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,180 @@
...
+
+    The CPT includes three additional colors beyond the range of z-values.
+    These are the background color (B) assigned to values lower than the lowest
+    *z*-value, the foreground color (F) assigned to values higher than the
+    highest *z*-value, and the NaN color (N) painted wherever values are
+    undefined.
```
weiji14, line 56:
```suggestion
    undefined. For color tables beyond the standard GMT offerings, visit
    `cpt-city <http://soliton.vm.bytemark.co.uk/pub/cpt-city/>`_ and
    `Scientific Colour-Maps <http://www.fabiocrameri.ch/colourmaps.php>`_.
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,180 @@
...
+    background : bool or str
+        Select the back- and foreground colors to match the colors for lowest
+        and highest *z*-values in the output CPT [Default (``background=True``
+        or ``background='o'``) uses the colors specified in the master file, or
+        those defined by the parameters :gmt-term:`COLOR_BACKGROUND`,
+        :gmt-term:`COLOR_FOREGROUND`, and :gmt-term:`COLOR_NAN`]. Use
+        ``background='i'`` to match the colors for the lowest and highest
+        values in the input (instead of the output) CPT.
+    color_model :
+        ``[R|r|h|c][+c[label]]``.
```
weiji14, line 96:
```suggestion
        [**R**\|\ **r**\|\ **h**\|\ **c**][**+c**\ [*label*]].
```

Update this for the other docstrings below too, see https://github.com/GenericMappingTools/gmt/edit/master/doc/rst/source/grd2cpt.rst.

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,180 @@
...
+    use_alias,
+)
+
+
+@fmt_docstring
+@use_alias(
+    A="transparency",
+    C="cmap",
+    D="background",
+    F="color_model",
```
weiji14, line 22: Missing `-E` (see https://docs.generic-mapping-tools.org/6.1/grd2cpt.html#e). Shall we call the alias `nlevels` like GMT.jl at https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.grd2cpt?

weiji14 (commented):
It's getting late for me, but here are some comments to get you started, you'll need to update some docstrings in line from https://github.com/GenericMappingTools/gmt/edit/master/doc/rst/source/makecpt.rst, and I noticed the `-E` argument is missing.

Also, someone really needs to find a way to make an automated script to copy the docstring over from GMT and find a way to regex substitute the long aliases for the short arguments, this would speed up module wrapping so much!

pygmt/src/grd2cpt.py
willschlitzer: Wrapped `-E` in 3e4b017

pygmt/src/grd2cpt.py
willschlitzer: Formatted in 5924a8e

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+    Z="continuous",
+)
+@kwargs_to_strings(T="sequence", G="sequence", L="sequence", R="sequence")
+def grd2cpt(grid, **kwargs):
+    r"""
+    Make GMT color palette tables from a grid file.
+
+    This is a module that will help you make static color palette tables
+    (CPTs). By default, the CPT will simply be saved to the current session,
+    but you can use *output* to save it to a file. The CPT is based on an
```
seisman, line 45:
```suggestion
    but you can use ``output`` to save it to a file. The CPT is based on an
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+    :gmt-term:`COLOR_FOREGROUND`, and :gmt-term:`COLOR_NAN` from the
+    :gmt-docs:`gmt.conf <gmt.conf>` file or the command line will be used. This
+    default behavior can be overruled using the options *background*,
+    *overrule_bg* or *no_bg*.
+
+    The color model (RGB, HSV or CMYK) of the palette created by
+    :meth:`pygmt.grd2cpt` will be the same as specified in the header of the
+    master CPT. When there is no :gmt-term:`COLOR_MODEL` entry in the master
+    CPT, the :gmt-term:`COLOR_MODEL` specified in the
+    :gmt-docs:`gmt.conf <gmt.conf>` file or the *color_model* option will be
```
seisman, line 74:
```suggestion
    :gmt-docs:`gmt.conf <gmt.conf>` file or the ``color_model`` option will be
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Full option list at :gmt-docs:`grd2cpt.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    transparency : str
```
seisman, line 85:
```suggestion
    transparency : int or float or str
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+    color_model :
+        [**R**\|\ **r**\|\ **h**\|\ **c**][**+c**\ [*label*]].
+        Force output CPT to be written with r/g/b codes, gray-scale values or
+        color name (**R**, default) or r/g/b codes only (**r**), or h-s-v codes
+        (**h**), or c/m/y/k codes (**c**).  Optionally or alternatively, append
+        **+c** to write discrete palettes in categorical format. If *label* is
+        appended then we create labels for each category to be used when the
+        CPT is plotted. The *label* may be a comma-separated list of category
+        names (you can skip a category by not giving a name), or give
+        *start*[-], where we automatically build monotonically increasing
```
seisman, line 110:
```suggestion
        *start*\[-], where we automatically build monotonically increasing
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+        labels from *start* (a single letter or an integer). Append - to build
+        ranges *start*-*start+1* instead.
```
seisman, line 112:
```suggestion
        labels from *start* (a single letter or an integer). Append ``-`` to build
        ranges *start*-*start+1* instead.
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+        with file_context as infile:
+            if "W" in kwargs and "Ww" in kwargs:
+                raise GMTInvalidInput(
+                    "Set only categorical or cyclic to True, not both."
+                )
+
+            if "H" in kwargs.keys():  # if output file is set
+                outfile = kwargs.pop("H")
+                if not outfile or not isinstance(outfile, str):
+                    raise GMTInvalidInput("'output' should be a proper file name.")
```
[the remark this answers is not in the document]
willschlitzer: @seisman As I understand it, `grd2cpt` can specify can output file that saves the CPT instead of just having it be the session CPT. I figured this was the same process for `grd2cpt` and `makecpt` and I copied the code for this. Is this not the way it would work?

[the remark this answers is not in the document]
seisman: To save the CPT to a file, you need to call `grd2cpt ... -H > out.cpt`. See the lines 157 and 158 in `makecpt.py`.

https://github.com/GenericMappingTools/pygmt/blob/921e7a6d7a036ab6ef7975d7449f9627d4f56c05/pygmt/src/makecpt.py#L148-L158

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,344 @@
...
+        grd2cpt(grid=grid, output=["some.cpt"])
+
+
+def test_grd2cpt_output_to_cpt_file(grid):
+    """
+    Save the generated static color palette table to a .cpt file.
+    """
+    with GMTTempFile(suffix=".cpt") as cptfile:
+        grd2cpt(grid=grid, output=cptfile.name)
+        assert os.path.exists(cptfile.name)
```
seisman, line 295: This test doesn't work, because when calling `with GMTTempFile(suffix=".cpt")`, an empty temporary file is always created, no matter you call `grd2cpt` or not.

willschlitzer: Removed in 72130a9

pygmt/src/grd2cpt.py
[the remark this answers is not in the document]
willschlitzer: I think I fixed it in 201f0f6.

pygmt/tests/test_grd2cpt.py
seisman: @willschlitzer Sorry if my previous comment is misleading. The test is still necessary. The point is we can't check if the CPT is written using `assert os.path.exists(cptfile.name)`, because the temporary file `cptfile.name` always exists but with a zero size. I think at least we can check if the file size is non-zero.

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,193 @@
...
+        elif kind == "grid":
+            file_context = lib.virtualfile_from_grid(grid)
+        else:
+            raise GMTInvalidInput(f"Unrecognized data type: {type(grid)}")
+
+        with file_context as infile:
+            if "W" in kwargs and "Ww" in kwargs:
+                raise GMTInvalidInput(
+                    "Set only categorical or cyclic to True, not both."
+                )
```
seisman, line 183: Can we move the if test outside, e.g., before Line 170?

pygmt/tests/test_grd2cpt.py
willschlitzer: I changed it to `assert os.path.getsize(cptfile.name) > 0` and that is passing! Just to confirm is wasn't a fluke I tested `assert os.path.getsize(cptfile.name) == 0` and the test failed.

pygmt/src/grd2cpt.py
willschlitzer: Moved it outside.

pygmt/tests/test_grd2cpt.py
seisman: > I changed it to `assert os.path.getsize(cptfile.name) > 0` and that is passing!

It looks good to me. Ideally we should check the content of the CPT file, but it needs more coding, and I think we can leave it to future.

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+    undefined. For color tables beyond the standard GMT offerings, visit
+    `cpt-city <http://soliton.vm.bytemark.co.uk/pub/cpt-city/>`_ and
+    `Scientific Colour-Maps <http://www.fabiocrameri.ch/colourmaps.php>`_.
+
+    If the master CPT includes B, F, and N entries, these will be copied into
+    the new master file. If not, the parameters :gmt-term:`COLOR_BACKGROUND`,
+    :gmt-term:`COLOR_FOREGROUND`, and :gmt-term:`COLOR_NAN` from the
+    :gmt-docs:`gmt.conf <gmt.conf>` file or the command line will be used. This
+    default behavior can be overruled using the options *background*,
+    *overrule_bg* or *no_bg*.
```
seisman, line 68:
```suggestion
    default behavior can be overruled using the options ``background``,
    ``overrule_bg`` or ``no_bg``.
```

seisman (commented):
I'm OK with the wrapper and the aliases, but most of the tests don't make sense to me.

For example, in the following test, if `cmap="rainbow"` doesn't work as expected (although unlikely as the wrapper is very simple), it would affect both the reference and test images, thus the tests always pass.
```
@check_figures_equal()
def test_grd2cpt_set_cpt(grid):
    """
    Test function grd2cpt to create a CPT based off a grid input and a set CPT.
    """
    fig_ref, fig_test = Figure(), Figure()
    # Use single-character arguments for the reference image
    fig_ref.basemap(B="a", J="W0/15c", R="d")
    grd2cpt(grid=grid, cmap="rainbow")
    fig_ref.colorbar(B="a2000")
    fig_test.basemap(frame="a", projection="W0/15c", region="d")
    grd2cpt(grid=grid, cmap="rainbow")
    fig_test.colorbar(frame="a2000")
    return fig_ref, fig_test
```

willschlitzer:
I'm fine with deleting tests. I was mostly copying the format of test_makecpt.py, since that seems like the most similar module.

seisman:
> I'm fine with deleting tests. I was mostly copying the format of test_makecpt.py, since that seems like the most similar module.

Some of the `makecpt` tests make sense because they're using `@pytest.mark.mpl_image_compare`. I think we need to think more about these tests. 

@weiji14 As you wrote the `makecpt` wrapper, I'll let you decide what to do.

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,344 @@
...
+from pygmt.exceptions import GMTInvalidInput
+from pygmt.helpers import GMTTempFile
+from pygmt.helpers.testing import check_figures_equal
+from pygmt.src.grd2cpt import grd2cpt
+
+TEST_DATA_DIR = os.path.join(os.path.dirname(__file__), "data")
+POINTS_DATA = os.path.join(TEST_DATA_DIR, "points.txt")
+
+
+@pytest.fixture(scope="module", name="points")
```
seisman, line 19: I think we can remove most tests, except:

- `test_grd2cpt`
- `test_grd2cpt_blank_output`
- `test_grd2cpt_invalid_output`
- `test_grd2cpt_output_to_cpt_file`
- `test_grd2cpt_unrecognized_data_type`
- `test_grd2cpt_categorical_and_cyclic`

willschlitzer: Removed in bf68c85

pygmt/tests/test_grd2cpt.py
```diff
@@ -0,0 +1,99 @@
...
+    return np.loadtxt(POINTS_DATA)
+
+
+@pytest.fixture(scope="module", name="region")
+def fixture_region():
+    """
+    The data region.
+    """
+    return [10, 70, -5, 10]
+
```
seisman, line 33: These lines can also be removed.
```suggestion
```

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+        the generated CPT file. If not given or False (default), saves the CPT
+        as the session current CPT.
+    reverse : str
+        Set this to True or c [Default] to reverse the sense of color
+        progression in the master CPT. Set this to z to reverse the sign of
+        z-values in the color table. Note that this change of z-direction
+        happens before *truncate* and *series* values are used so the latter
+        must be compatible with the changed *z*-range. See also
+        :gmt-docs:`cookbook/features.html#manipulating-cpts`.
+    overrule_bg :
```
weiji14, line 143:
```suggestion
    overrule_bg : str
```
Seems like this is missing in `makecpt` too at https://github.com/GenericMappingTools/pygmt/blob/5b80c64bb97a6b84c5d8dcfc539ca1b097c2a5aa/pygmt/src/makecpt.py#L121, could you update the docstring there too?

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+    N="no_bg",
+    Q="log",
+    R="region",
+    T="series",
+    V="verbose",
+    W="categorical",
+    Ww="cyclic",
+    Z="continuous",
+)
+@kwargs_to_strings(T="sequence", G="sequence", L="sequence", R="sequence")
```
weiji14, line 38:
```suggestion
@kwargs_to_strings(G="sequence", L="sequence", R="sequence", T="sequence")
```
Make alphabetical.

pygmt/src/grd2cpt.py
```diff
@@ -0,0 +1,190 @@
...
+        appended then we create labels for each category to be used when the
+        CPT is plotted. The *label* may be a comma-separated list of category
+        names (you can skip a category by not giving a name), or give
+        *start*\[-], where we automatically build monotonically increasing
+        labels from *start* (a single letter or an integer). Append ``-`` to
+        build ranges *start*-*start+1* instead.
+    nlevels : bool or int or str
+        Set to ``True`` to create a linear color table by using the grid
+        z-range as the new  limits in the CPT. Alternatively, set *nlevels*
+        and to resample the color table into nlevels equidistant slices.
```
weiji14, line 116:
```suggestion
        z-range as the new limits in the CPT. Alternatively, set *nlevels*
        to resample the color table into *nlevels* equidistant slices.
```

GMT 6.2.0 (unreleased) has some new options here (see https://github.com/GenericMappingTools/gmt/pull/3693, so we might need to update this in PyGMT v0.4.0.

weiji14 (commented):
Just a few minor comments, getting close!

pygmt/src/grd2cpt.py
willschlitzer: Updated for `makecpt` in e0fd4ab

weiji14 (approved):
Great! Thanks for your patience @willschlitzer, and also thank you @seisman for the comprehensive review!

Issue #182, opened 2018-05-24
Make it clear in the tutorials that they assume that you're using a notebook


leouieda:
The tutorials all assume that the user is running Jupyter with the many `fig.show()` calls. There should be a warning in the beginning of each tutorial stating that this works on notebooks and giving options for the terminal prompt (`fig.show(external=True)`) and in scripts `fig.savefig('figname.pdf', show=True)`.

willschlitzer:
@GenericMappingTools/python Is this something we still want, and if so, should it be in the doc strings/lead section of every tutorial and gallery example?

seisman:
The issue will become invalid after #529 is merged. That PR is almost working, however, I haven't found much time to finish it.

willschlitzer:
Sounds good. Do you think it will be done before the v0.3 release? I'm not trying to hurry you, but I think if it's not the tutorials should at least be updated for the next release, and then removed once your pull request is merged.

seisman:
> Do you think it will be done before the v0.3 release?

Not sure if I can find some time before v0.3 release. 

> I can submit a PR for this if it looks like it's the case that it won't be ready in time.

Go ahead, please.

PR #806, opened 2021-01-26
Tutorial notebook explanation

willschlitzer:
As discussed in #182, this is to update the doc strings for the tutorials to state they assume the use of a Python notebook, how to modify the code to use a Pythons script, and how to save figures. These lines should be removed once #529 is merged.


willschlitzer:
@seisman Not sure if I messed up the fetch/force push, but I'm style seeing all of the old commits in this PR.

willschlitzer:
@GenericMappingTools/python-maintainers Do we think the explanation lines for using a notebook should be on all gallery examples as well? I think tutorials are sufficient, but I can add it in galleries as well.

willschlitzer:
@GenericMappingTools/python Could I get a review/feedback for this pull request? I'm hoping to get it in for the documentation on v0.3.

examples/tutorials/3d-perspective-image.py
```diff
@@ -4,6 +4,13 @@
 
 Create 3-D perspective image or surface mesh from a grid
 using :meth:`pygmt.Figure.grdview`.
+
+This tutorial assumes the use of a Python notebook, such as IPython or Jupyter Notebook.
+To see the figures while using a Python script instead, use
+``fig.show(method="external)`` to display the figure in the default PDF viewer.
+
+To save the figure, use ``fig.savefig("figname.pdf")`` where ``"figname.pdf"``
+is the desired name and file extension for the saved figure.
```
seisman, line 13: Perhaps looks better using a `note` directive?
image


```suggestion
.. note::

    This tutorial assumes the use of a Python notebook, such as IPython or Jupyter Notebook.
    To see the figures while using a Python script instead, use
    ``fig.show(method="external)`` to display the figure in the default PDF viewer.

    To save the figure, use ``fig.savefig("figname.pdf")`` where ``"figname.pdf"``
    is the desired name and file extension for the saved figure.
```

weiji14:
Is there a way to insert a common block of text in multiple tutorial pages? Better than copying and pasting so much text around everytime we make a change.

willschlitzer:
> Is there a way to insert a common block of text in multiple tutorial pages? Better than copying and pasting so much text around everytime we make a change.

I think this is a good idea, but I don't think we'll be able to get it done before v0.3, so I would rather have this note in here as is, for now, rather than no note at all.

examples/tutorials/3d-perspective-image.py
willschlitzer: I changed this in the tutorial files in dd3ab29

seisman:
> Is there a way to insert a common block of text in multiple tutorial pages? Better than copying and pasting so much text around everytime we make a change.

It's a good idea. I tried to explore the possible ways but didn't find one. As we will delete these notes after we finish #529, I feel that the current PR is the simplest solution.

examples/tutorials/3d-perspective-image.py
```diff
@@ -4,6 +4,15 @@
 
 Create 3-D perspective image or surface mesh from a grid
 using :meth:`pygmt.Figure.grdview`.
+
+.. note::
+
+    This tutorial assumes the use of a Python notebook, such as IPython or Jupyter Notebook.
+    To see the figures while using a Python script instead, use
+    ``fig.show(method="external)`` to display the figure in the default PDF viewer.
```
weiji14, line 12:
```suggestion
    ``fig.show(method="external)`` to display the figure in the default image viewer.
```
Is the default PDF or PNG?

weiji14 (commented):
Ok. Since it's temporary let's go with it. But one small comment.

examples/tutorials/3d-perspective-image.py
willschlitzer: > ```suggestion
>     ``fig.show(method="external)`` to display the figure in the default image viewer.
> ```
> Is the default PDF or PNG?

The Figure.show() documentation makes it seem like it is the PDF viewer.

weiji14: Ah ok, all good then.

weiji14 (approved):
Feel free to merge :)

PR #788, opened 2021-01-03
Wrap inset

willschlitzer:
**Description of proposed changes**

As discussed in #663, this is wrapping the GMT function `inset`. After some back and forth (and an attempt at Option 2), this PR goes with Option 1 to use a context manager and `with` statement to call `inset`.

Fixes #663

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Notes**

- You can write `/format` in the first line of a comment to lint the code automatically

PR #731, opened 2020-12-14
Wrap triangulate

weiji14:
Wrapping the [triangulate](https://docs.generic-mapping-tools.org/6.1/triangulate.html) function which does "Delaunay triangulation or Voronoi partitioning and gridding of Cartesian data", implemented under gridding.py.

Live documentation preview for this Pull Request is at https://pygmt-git-gridding-triangulate.gmt.vercel.app/api/generated/pygmt.triangulate.html.

**Description of proposed changes**

Note that this is just a minimal implementation of `triangulate` for now, and I don't expect to wrap and test all the aliases here as they are super complicated. Just that I've just been using `triangulate` secretly at https://github.com/weiji14/pyissm/blob/5c9e3be4c4983ccc79a67cc6471421922e8c3af7/plot_figures.py#L68-L83 and thought it should get into `pygmt` proper :smile:

The complicated thing about `triangulate` is that it generates either text (table) or NetCDF (grid) outputs, and I've implemented it as so:

- [x] When `outgrid=None` (default), output table as `pandas.DataFrame`
- [ ] (Won't be implemented?), output table as textfile 
- [x] When `outgrid=True`, output grid as 'xarray.DataArray`
- [x] When `outgrid=None`, output grid to NetCDF file

Welcome to other suggestions on implementation details.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Notes**

- You can write `/format` in the first line of a comment to lint the code automatically

PR #808, opened 2021-01-28
Move plotting functions to separate modules

willschlitzer:
As mentioned in #685, this is to split up `base_plotting.py` and move the plotting functions to separate scripts, following the method used by @weiji14 in #686.

**Reminders**

**Notes**


willschlitzer:
@GenericMappingTools/python I moved all of the plotting functions to their own modules. I'm running into a few style errors that I'm not sure how to address.

`pygmt/src/contour.py:98:13: W0212: Access to a protected member _preprocess of a client class (protected-access)`
I'm guessing this is because the functions use `_preprocess` but don't import it, but the tests successfully running makes me assume that it works because `_preprocess` is in the class that these modules are imported into.

`pygmt/base_plotting.py:8:0: R0903: Too few public methods (0/2) (too-few-public-methods)`
I'm assuming this is because the class doesn't really "do" all that much specifically, since the functions are all housed in their own modules. I'm assuming this is acceptable, otherwise functions would just be arbitrarily moved back to `base_plotting.py`.

`pygmt/src/text.py:48:4: W0621: Redefining name 'text' from outer scope (line 40) (redefined-outer-name)`
I'm guessing "text" is a common/special word, so having it in the name of both a function and module is causing difficulties. Not sure how it should be addressed.

pygmt/src/text.py
```diff
@@ -0,0 +1,206 @@
+"""
+text - Plot text on a figure figure.
```
seisman, line 2:
```suggestion
text - Plot text on a figure.
```

pygmt/src/basemap.py
```diff
@@ -0,0 +1,78 @@
+"""
+basemap - Plot a base map and frames.
+"""
```
seisman, line 3: See line 36 of this file. It would be better if the function description is the same as the one shown in the API reference (i.e., line 36).

willschlitzer: Copied line 36.

seisman:
> @GenericMappingTools/python I moved all of the plotting functions to their own modules. 

Great work!

> `pygmt/src/contour.py:98:13: W0212: Access to a protected member _preprocess of a client class (protected-access)`
> I'm guessing this is because the functions use `_preprocess` but don't import it, but the tests successfully running makes me assume that it works because `_preprocess` is in the class that these modules are imported into.

Functions with a leading underscore mean they're protected, and _should_ not be accessed outside the class. This function does nothing, and it should be safe to remove it from all functions. I'm not sure if we still need it in the future.
 
> `pygmt/base_plotting.py:8:0: R0903: Too few public methods (0/2) (too-few-public-methods)`
> I'm assuming this is because the class doesn't really "do" all that much specifically, since the functions are all housed in their own modules. I'm assuming this is acceptable, otherwise functions would just be arbitrarily moved back to `base_plotting.py`.

OK to disable the warning.
 
> `pygmt/src/text.py:48:4: W0621: Redefining name 'text' from outer scope (line 40) (redefined-outer-name)`
> I'm guessing "text" is a common/special word, so having it in the name of both a function and module is causing difficulties. Not sure how it should be addressed.

The function is called `text`. Can we also use `text` in the arguments? 😕 
```
def text(
    self,
    textfiles=None,
    x=None,
    y=None,
    position=None,
    text=None,
    angle=None,
    font=None,
    justify=None,
    **kwargs,
):
```

willschlitzer:
@seisman So should I change the function name or the argument name?

willschlitzer:
@seisman When I try removing `_preprocess` all of my tests fail. I get:

`pygmt.exceptions.GMTCLibError: Module 'psconvert' failed with status code 78`

I can't tell what it does, but it has some effect.

seisman:
> @seisman So should I change the function name or the argument name?

Need more thoughts and comments from @GenericMappingTools/python-maintainers


> @seisman When I try removing `_preprocess` all of my tests fail. I get:
> 
> `pygmt.exceptions.GMTCLibError: Module 'psconvert' failed with status code 78`
> 
> I can't tell what it does, but it has some effect.

I think you can't remove `_preprocess` from the functions in `pygmt/figure.py`.

weiji14:
> > @seisman So should I change the function name or the argument name?
> 
> Need more thoughts and comments from @GenericMappingTools/python-maintainers

Maybe just keep the `text` function in `base_plotting.py` for now :slightly_smiling_face:

> I think you can't remove `_preprocess` from the functions in `pygmt/figure.py`.

I feel like the `_preprocess` function might be needed for some obscure reason, e.g. when people do something like:

```python
fig1 = pygmt.Figure()
fig2 = pygmt.Figure()

fig2.basemap()
fig1.basemap()

fig1.show()
fig2.show()
```

Do things get passed to the correct figure in this case?

willschlitzer:
@GenericMappingTools/python Could I get additional feedback/comments for this pull request?

seisman:
> I feel like the `_preprocess` function might be needed for some obscure reason, e.g. when people do something like:
> 
> ```python
> fig1 = pygmt.Figure()
> fig2 = pygmt.Figure()
> 
> fig2.basemap()
> fig1.basemap()
> 
> fig1.show()
> fig2.show()
> ```
> 
> Do things get passed to the correct figure in this case?

Can you remove `_preprocess` from `basemap()` and try if the above example works? I think it doesn't matter.

pygmt/base_plotting.py
```diff
@@ -55,1592 +40,17 @@ def _preprocess(self, **kwargs):  # pylint: disable=no-self-use
...
+    from pygmt.src import grdcontour  # pylint: disable=import-outside-toplevel
+    from pygmt.src import grdimage  # pylint: disable=import-outside-toplevel
+    from pygmt.src import grdview  # pylint: disable=import-outside-toplevel
+    from pygmt.src import image  # pylint: disable=import-outside-toplevel
+    from pygmt.src import legend  # pylint: disable=import-outside-toplevel
+    from pygmt.src import logo  # pylint: disable=import-outside-toplevel
     from pygmt.src import meca  # pylint: disable=import-outside-toplevel
+    from pygmt.src import plot  # pylint: disable=import-outside-toplevel
+    from pygmt.src import plot3d  # pylint: disable=import-outside-toplevel
+    from pygmt.src import text  # pylint: disable=import-outside-toplevel
```
weiji14, line 56: You know, if we're going to move **every** plotting function to the `src/` folder, we might as well delete `base_plotting.py` and import everything in `figure.py` instead. Or we could keep a **few** core functions in `base_plotting.py` (e.g. `basemap`, and `inset` perhaps as mentioned in https://github.com/GenericMappingTools/pygmt/pull/788#issuecomment-766760050). What do people think?

```suggestion
from pygmt.src import (  # pylint: disable=import-outside-toplevel
        basemap,
        coast,
        colorbar,
        contour,
        grdcontour,
        grdimage,
        grdview,
        image,
        legend,
        logo,
        meca,
        plot,
        plot3d,
        text,
    )
```

seisman: > You know, if we're going to move **every** plotting function to the `src/` folder, we might as well delete `base_plotting.py` and import everything in `figure.py` instead. 

I like this one.

willschlitzer: I committed this changed in 76c4969 but it led to 191 tests failing when I ran `make test`. I'm not sure what the issue is, but I've reverted for now.

willschlitzer:
> > I feel like the `_preprocess` function might be needed for some obscure reason, e.g. when people do something like:
> > ```python
> > fig1 = pygmt.Figure()
> > fig2 = pygmt.Figure()
> > 
> > fig2.basemap()
> > fig1.basemap()
> > 
> > fig1.show()
> > fig2.show()
> > ```
> > 
> > 
> > Do things get passed to the correct figure in this case?
> 
> Can you remove `_preprocess` from `basemap()` and try if the above example works? I think it doesn't matter.
 ```python
 fig1 = pygmt.Figure()
fig2 = pygmt.Figure()
fig2.basemap(frame="a", region=[0, 20, 0, 20])
fig1.basemap(frame="a", region=[0, 10, 0, 10])
fig1.show()
fig2.show()
```
When I ran the above code after removing `_preprocess` I received `GMTCLibError: Module 'psconvert' failed with status code 78:
psconvert [ERROR]: No hidden PS file /home/will/.gmt/sessions/gmt_session.38138/gmt_1.ps- found` and it resulted in 10 failed tests.

File: pygmt/base_plotting.py (hunk 3 above)
Comment (weiji14) [reply pygmt/base_plotting.py:56]: Oops sorry, just realized it was missing an indentation... Try this instead:

```suggestion
    from pygmt.src import (  # pylint: disable=import-outside-toplevel
        basemap,
        coast,
        colorbar,
        contour,
        grdcontour,
        grdimage,
        grdview,
        image,
        legend,
        logo,
        meca,
        plot,
        plot3d,
        text,
    )
```

weiji14: Ok, I applied https://github.com/GenericMappingTools/pygmt/pull/808#discussion_r570775763 in 127b42874466c66eb970fb53a8846db9ff1f73e8.

pygmt/src/text.py
```diff
@@ -0,0 +1,206 @@
...
+)
+@kwargs_to_strings(
+    R="sequence",
+    textfiles="sequence_space",
+    angle="sequence_comma",
+    font="sequence_comma",
+    justify="sequence_comma",
+    p="sequence",
+)
+def text(
```
weiji14, line 42:
```suggestion
def text_(
```

How about this to fix the pylint `W0621: Redefining name 'text' from outer scope (line 40) (redefined-outer-name)` warning? And then use `from pygmt.src.text import text_ as text` in `pygmt/src/__init__.py`.

pygmt/src/__init__.py
```diff
@@ -2,13 +2,26 @@
...
 from pygmt.src.info import info
 from pygmt.src.inset import inset
+from pygmt.src.legend import legend
+from pygmt.src.logo import logo
 from pygmt.src.makecpt import makecpt
 from pygmt.src.meca import meca
+from pygmt.src.plot import plot
+from pygmt.src.plot3d import plot3d
 from pygmt.src.surface import surface
+from pygmt.src.text import text
```
weiji14, line 26:
```suggestion
from pygmt.src.text import text_ as text
```

pygmt/base_plotting.py
```diff
@@ -55,1591 +40,20 @@ def _preprocess(self, **kwargs):  # pylint: disable=no-self-use
...
+        grdview,
+        image,
+        inset,
+        legend,
+        logo,
+        meca,
+        plot,
+        plot3d,
+        text,
     )
```
weiji14, line 59: As mentioned in https://github.com/GenericMappingTools/pygmt/pull/808#discussion_r570575569, could you delete `base_plotting.py` (can't believe I'm saying this!) and move this import chunk to `figure.py`? This will fix the pylint `R0903: Too few public methods (0/2) (too-few-public-methods)` recommendation. Feel free to copy any of the docstrings if needed.

willschlitzer: @weiji14 I just moved this import string to `figure.py` and removed imports for `base_plotting.py`, but when I ran `make test` I had 196 test failures. I'm not sure what the need to `_preprocess` is, but it seems to play some sort of role.

weiji14: Just double checking that you placed the imports under `class Figure():`? I.e. indented 4 spaces? The `_preprocess` class function activates the correct figure to plot in.

willschlitzer: ...nope 🤦‍♂️

Removed `base_plotting.py` in 5ac2e07 and actually put the imports in `Figure()` and it worked (or at least passed all of the tests)!

seisman: > As mentioned in [#808 (comment)](https://github.com/GenericMappingTools/pygmt/pull/808#discussion_r570575569), could you delete `base_plotting.py` (can't believe I'm saying this!) and move this import chunk to `figure.py`? This will fix the pylint `R0903: Too few public methods (0/2) (too-few-public-methods)` recommendation. Feel free to copy any of the docstrings if needed.

We should be careful with the changes. There are two different `_preprocess()` functions in `Figure()` and `BasePlotting`. If I understand it correctly, the `_preprocess` function if `BasePlotting` override the one in `Figure()`.

willschlitzer: @seisman The tests are passing; is there an issue with deleting base_plotting.py?

weiji14: > We should be careful with the changes. There are two different `_preprocess()` functions in `Figure()` and `BasePlotting`. If I understand it correctly, the `_preprocess` function if `BasePlotting` override the one in `Figure()`.

I feel like the `_preprocess` in base_plotting.py was just a dummy placeholder, and the one in figure.py is the actual one used. So probably the other way around? Might need to read up on subclassing, I'm not terribly good at object orientated conventions.

weiji14 (approved):
I think this is good to merge now (after updating the branch), what do you think @seisman?

Deleting `base_plotting.py` and moving every module to `src` will cause conflicts in several open PRs (e.g. `velo` at #525, `triangulate` at #731, the `grdimage` work at #750, etc), but we can sort those out later :upside_down_face:

pygmt/base_plotting.py
seisman: I thought `BasePlotting()` is a subclass of `Figure()`, now it makes sense to me.

pygmt/src/__init__.py
```diff
@@ -2,13 +2,26 @@
...
 from pygmt.src.info import info
 from pygmt.src.inset import inset
+from pygmt.src.legend import legend
+from pygmt.src.logo import logo
 from pygmt.src.makecpt import makecpt
 from pygmt.src.meca import meca
+from pygmt.src.plot import plot
+from pygmt.src.plot3d import plot3d
 from pygmt.src.surface import surface
+from pygmt.src.text import text_ as text
```
seisman, line 26: Please add a comment explaining why `text` is different.

pygmt/src/text.py
seisman, line 2:
```suggestion
text - Plot text on a figure.
```

seisman (commented):
Looks good to me, except a few tiny suggestions.

Issue #640, opened 2020-10-04
Decorator kwargs_to_strings is too aggressive when converting boolean arguments


seisman:
**Description of the problem**

https://github.com/GenericMappingTools/pygmt/blob/c191d3e09a43a0536b2ae14a7c8380969f4646a4/pygmt/helpers/decorators.py#L279

Decorator `kwargs_to_strings` can deal with boolean arguments. For example, dict `{'A': True, 'B': False, 'R': '0/10/0/10'}` will be converted to `{'A': '', 'R': '0/10/0/10'}` first, and then `build_arg_strings` will further convert it to valid GMT arguments `-A -R0/10/0/10`. 

The decorator works well for arguments that will be passed to GMT, but doesn't work for pure Python arguments.

**Full code that generated the error**

```python
from pprint import pprint
from pygmt.helpers.decorators import kwargs_to_strings

@kwargs_to_strings(R='sequence', files='sequence_space')
def module(nongmt_args=True, *args, **kwargs):
    "A module that prints the arguments it received"
    pprint(args)
    pprint(kwargs)
    if nongmt_args is True:
        print("nongmt_args is still True. Do something in PyGMT.")
    elif nongmt_args == "":
        print("the kwargs_to_strings decorator converts nongmt_args to an empty string.")
    else:
        print(f"nongmt_args is '{nongmt_args}'")

module(R=[1, 2, 3, 4], nongmt_args=True)
```

**Output**
```
()
{'R': '1/2/3/4'}
the kwargs_to_strings decorator converts nongmt_args to an empty string.
```
As shown in the output above, the boolean argument `nongmt_args` is converted from `True` to an empty string `""`. Thus, checking `nongmt_args is True` no longer works.

If removing the `kwargs_to_strings` decorator before the `module` function, the output is:
```
()
{'R': [1, 2, 3, 4]}
nongmt_args is still True. Do something in PyGMT.
```

**Possible solution**

Like what we do in #639, the possible solution is to convert boolean arguments in `build_arg_strings` function before passing arguments to GMT, not in the `kwargs_to_strings` decorator.

**System information**

Please paste the output of `python -c "import pygmt; pygmt.show_versions()"`:

```
PyGMT information:
  version: v0.2.0+19.gc191d3e0
System information:
  python: 3.8.3 (default, Jul  2 2020, 11:26:31)  [Clang 10.0.0 ]
  executable: /Users/seisman/.anaconda/bin/python
[...]

PR #822, opened 2021-01-30
Wrap subplot using with statement

### LINKED ISSUE #20: Implement subplot command

weiji14:

Yet another attempt at wrapping the `subplot` function! Preview the tutorial at https://pygmt-git-subplotv3.gmt.vercel.app/tutorials/subplots.html, translated and adapted from https://docs.gmt-china.org/latest/tutorial/subplot/ (in Chinese).

Original GMT `subplot` function can be found at https://docs.generic-mapping-tools.org/6.1/subplot.html. This implementation is inspired by the `inset` syntax at #788, namely wrapping things in a `with` statement like so:

Example code:

```python
import pygmt

fig = pygmt.Figure()
with fig.subplot(nrows=1, ncols=2, figsize=("6c", "3c"), frame="WSne"):
    with fig.set_panel(panel=[0, 0]):
        fig.basemap(region=[0, 3, 0, 3], frame="+tplot0")
    with fig.set_panel(panel=[0,1]):
        fig.basemap(region=[0, 3, 0, 3], frame="+tplot1")
fig.show()
```

Produces:

Basic PyGMT subplot figure

Alternatively, this works too:

```python
fig = pygmt.Figure()
with fig.subplot(nrows=1, ncols=2, figsize=("6c", "3c"), frame="WSne"):
    fig.basemap(region=[0, 3, 0, 3], frame="+tplot0", panel=[0, 0])
    fig.basemap(region=[0, 3, 0, 3], frame="+tplot1", panel=[0, 1])
fig.show()
```

Note that the `with` statement does `gmt subplot begin`, and automatically exits using `gmt subplot end` outside the with-block. Users can set the current subplot using `fig.set_panel` (similar to matplotlib's `plt.sca()`) because `set` is a reserved Python name.

`fig.subplot()` parameters/aliases to wrap:

- [x] Ff = figsize
- [x] Fs = subsize
- [x] A = autolabel
- [x] B = frame
- [x] C = clearance
- [x] J = projection
- [x] M = margins
- [x] R = region
- [x] SC = sharex
- [x] SR = sharey
- [x] T = title
- [x] V = verbose
- [x] X/Y = xshift/yshift

`fig.set_panel()` parameters/aliases to wrap:

- [x] A = fixedlabel
- [x] C = clearance
- [x] V = verbose

Fixes #20, Supersedes and closes #412 and closes #427

**Reminders**

**Notes**


willschlitzer:
@weiji14 Any thought to using `sca` in a context manager? I know it doesn't have a `sca_end` equivalent the way that `inset` and `subplot` do, but I think using it with `with` statements will make it more clear that the code is creating different subplots, as opposed to a long line of similar looking plotting functions.

seisman:
@weiji14 It's really a nice job! This is almost what I expect, although I haven't read the details.

A quick question about `ax` and `sca`. The long-form name of **-c** option is **panel** in GMT (see https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_common_longoptions.h). Do we want to follow GMT or matplotlib?

weiji14:
> @weiji14 Any thought to using `sca` in a context manager? I know it doesn't have a `sca_end` equivalent the way that `inset` and `subplot` do, but I think using it with `with` statements will make it more clear that the code is creating different subplots, as opposed to a long line of similar looking plotting functions.

I did think of this, but it's technically quite  hard to implement. What should a `with fig.sca()` block reset to? The first subplot? The next subplot? There might be a way using function decorators (?) to insert `ax=1` for all `fig.*` functions under the with block but I'll need more time to confirm that it's possible. Or feel free to suggest ideas :)

> @weiji14 It's really a nice job! This is almost what I expect, although I haven't read the details.

Thanks, it's really looking at @willschlitzer's `inset` wrapper that gave me the inspiration!

> A quick question about `ax` and `sca`. The long-form name of **-c** option is **panel** in GMT (see https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_common_longoptions.h). Do we want to follow GMT or matplotlib?

Tough question. I think `panel` does seem like a name that conveys what is being done. But `ax` is shorter and more familiar to Python users :/ If only we could do both! Check with @GenericMappingTools/python and @willschlitzer.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,200 @@
...
+        in the initial **subplot** call.
+
+    {V}
+    """
+    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+
+    with Session() as lib:
+        arg_str = " ".join(["set", f"{ax}", build_arg_string(kwargs)])
+        lib.call_module(module="subplot", args=arg_str)
+        yield
```
weiji14, line 200: > > @weiji14 Any thought to using `sca` in a context manager? I know it doesn't have a `sca_end` equivalent the way that `inset` and `subplot` do, but I think using it with `with` statements will make it more clear that the code is creating different subplots, as opposed to a long line of similar looking plotting functions.
> 
> I did think of this, but it's technically quite hard to implement. What should a `with fig.sca()` block reset to? The first subplot? The next subplot? There might be a way using function decorators (?) to insert `ax=1` for all `fig.*` functions under the with block but I'll need more time to confirm that it's possible. Or feel free to suggest ideas :)

`fig.sca` is now a context manager as implemented here. There's not `try... finally` block because there's nothing to reset to. Have a look at the unit tests and tutorials and let me know what you think (i.e. if `with fig.sca():` is a good syntax to have, or if we should just stick to `fig.sca()`).

seisman:
Quick comments. In the 2nd example of the tutorial, you use:
```
fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=axs[0, 0])
```

It seems passing the *index*, or *row,col* also works, 
```
fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=0)
fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax="0,0")
```
but passing `[row, col]` doesn't:
```
fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=[0, 0])
```
I think we can just set `ax` as the alias of `-c`, and convert `[0, 0]` to `0,0`, similar to what we're doing to `region` (converting `[0, 10, 0, 10]` to `0/10/0/10`.

BTW, it's still unclear to me how `ax=0` works in `basemap()`, as I don't see how you pass **-c** to `basemap`, or calling `subplot set`. Could you please explain it a little bit?

weiji14:
> Quick comments. In the 2nd example of the tutorial, you use:
> 
> ```
> fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=axs[0, 0])
> ```
> 
> It seems passing the _index_, or _row,col_ also works,
> 
> ```
> fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=0)
> fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax="0,0")
> ```

Yes, in pure GMT, you would use something like `-c0` or `-c0,1`. I'm debating whether to even have the `axs` numpy array thing, it's meant to be like matplotlib's `Axes` object but we can't do `ax.plot` so maybe that will confuse users?

> but passing `[row, col]` doesn't:
> 
> ```
> fig.basemap(region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], ax=[0, 0])
> ```
> 
> I think we can just set `ax` as the alias of `-c`, and convert `[0, 0]` to `0,0`, similar to what we're doing to `region` (converting `[0, 10, 0, 10]` to `0/10/0/10`.

Ok, I'll read this as aliasing `c` to `ax` (rather than using `panel`), and we just need a `sequence (slash)` converter in place.

> BTW, it's still unclear to me how `ax=0` works in `basemap()`, as I don't see how you pass **-c** to `basemap`, or calling `subplot set`. Could you please explain it a little bit?

`ax` is an alias for `c`, so `ax=0` is equivalent to `-c0` in pure GMT. Using `fig.basemap(..., ax=0)` would tell the program to directly plot in the first (0) subplot, so `subplot set`/`fig.sca()` is not needed. See also https://docs.generic-mapping-tools.org/latest/gmt.html#c-full.

seisman:
> `ax` is an alias for `c`, so `ax=0` is equivalent to `-c0` in pure GMT. Using `fig.basemap(..., ax=0)` would tell the program to directly plot in the first (0) subplot, so `subplot set`/`fig.sca()` is not needed. See also https://docs.generic-mapping-tools.org/latest/gmt.html#c-full.

I see. I was fooled by the github web UI. When I review the changes, I thought you make the alias (`ax`<=>`c`) in `contour()` but can pass `ax` to `basemap()`. What magic! Now it all make sense to me.

> Yes, in pure GMT, you would use something like `-c0` or `-c0,1`. I'm debating whether to even have the `axs` numpy array thing, it's meant to be like matplotlib's `Axes` object but we can't do `ax.plot` so maybe that will confuse users?

Yes, especially that if `ax = [0, 0]` works, why do we bother to type `ax=axs[0, 0]`.

pygmt/src/subplot.py
seisman: Yes, the `fig.sca` context manager makes no sense here. I'm wondering if we can ask for upstream changes to be able to reset the ax number?

weiji14: But that would have to wait until GMT 6.2.0 :slightly_smiling_face: Also as mentioned at https://github.com/GenericMappingTools/pygmt/pull/822#issuecomment-770345781, what do we reset to? The first subplot? The previous one (-1)?

pygmt/base_plotting.py
```diff
@@ -1226,7 +1226,7 @@ def contour(self, x=None, y=None, z=None, data=None, **kwargs):
         p="perspective",
         t="transparency",
     )
-    @kwargs_to_strings(R="sequence", p="sequence")
+    @kwargs_to_strings(R="sequence", c="sequence", p="sequence")
```
seisman, line 1229: It should be `sequence_comma`.
```suggestion
    @kwargs_to_strings(R="sequence", c="sequence_comma", p="sequence")
```

pygmt/src/subplot.py
seisman: > The first subplot? The previous one (-1)?

Both doesn't make sense to me. It's weird and confusing to have `sca` as a context manager, but can add plotting elements to panels outside the context manager.

pygmt/base_plotting.py
weiji14: Oh right. Somehow my local tests passed still...

pygmt/src/subplot.py
```diff
@@ -0,0 +1,203 @@
...
+    want the scale to apply equally to both dimensions then you must specify
+    ``projection="x"`` [The default ``projection="X"`` will fill the subplot by
+    using unequal scales].
+
+    {aliases}
+
+    Parameters
+    ----------
+    ax : str
+        *row,col*\|\ *index*.
```
weiji14, line 169: I think it might be good to allow list/tuple inputs like `fig.sca(ax=[0, 1], ...)`, but `@kwargs_to_string(ax="sequence_comma")` won't work here because of #640.

pygmt/src/subplot.py
weiji14: Yeah, but map elements outside a `with fig.sca()` block can't just go into the main figure, as it would need to be outside the `fig.subplot()` indent. It's hard to come up with a good solution here.

So should we 1) just go back to having `fig.sca` as an ordinary function (i.e. **not** a context manager)? Or 2) keep `fig.sca()` as a context manager so that people can organize their subplots using indentation (if they choose to, or not)?

weiji14 (commented):
> Yes, especially that if `ax = [0, 0]` works, why do we bother to type `ax=axs[0, 0]`.

Ok, no more `axs`, see commit dbd4677409ff5a7ba37268e6862c22cb46c97493.

willschlitzer:
> 
> Yeah, but map elements outside a with fig.sca() block can't just go into the main figure, as it would need to be outside the       
> fig.subplot() indent. It's hard to come up with a good solution here.

I hope I didn't cause too much trouble with suggesting a context manager!

> So should we 1) just go back to having fig.sca as an ordinary function (i.e. not a context manager)? Or 2) keep fig.sca() as a 
> context manager so that people can organize their subplots using indentation (if they choose to, or not)?

I'm a fan of option 2. While I understand that `fig.sca()` doesn't need to be "closed" the way that `inset` and `subplot` do, I still think it makes it visually easier to see different `with` statements being used to plot on different subplots. The good news here is that if someone forgets to indent a line underneath a context manager it will still plot to the appropriate subplot figure, so it's not too punishing if they have a syntax typo.

weiji14:
> > Yeah, but map elements outside a with fig.sca() block can't just go into the main figure, as it would need to be outside the
> > fig.subplot() indent. It's hard to come up with a good solution here.
> 
> I hope I didn't cause too much trouble with suggesting a context manager!

It's actually a good idea, just that we're technically limited by GMT! We're just trying to find a way that's less confusing than matplotlib if possible.

> > So should we 1) just go back to having fig.sca as an ordinary function (i.e. not a context manager)? Or 2) keep fig.sca() as a
> > context manager so that people can organize their subplots using indentation (if they choose to, or not)?
> 
> I'm a fan of option 2. While I understand that `fig.sca()` doesn't need to be "closed" the way that `inset` and `subplot` do, I still think it makes it visually easier to see different `with` statements being used to plot on different subplots. The good news here is that if someone forgets to indent a line underneath a context manager it will still plot to the appropriate subplot figure, so it's not too punishing if they have a syntax typo.

I'm inclined to agree with @willschlitzer here, but after re-reading through the [Zen of Python](https://www.python.org/dev/peps/pep-0020) (my favourite is "There should be one-- and preferably only one --obvious way to do it."), I'm a bit split on what to choose. Having the `with` block option is nice as it allows for more flexibility, but is it too much freedom?

One thing I'll point out too is that the 'many' ways of making a subplot depends on how fancy people want to go, and maybe we could just document it better in the tutorial:

0. Use `fig.shift_origin()` #289.

1. The `ax` way for fairly simple subplots, E.g.:

```python
fig.basemap(..., ax=[0, 1])  # This is in subplot 1
fig.coast(...)  # This is also placed in subplot 1
fig.basemap(..., ax=[0, 1])  # This is placed in subplot 2
```

2. The `with fig.sca(ax=...):` way, better for complicated subplots if you want to override the title. E.g.:

```python
with fig.sca(ax=0, fixedlabel="subplot 0"):
    fig.basemap(...)
with fig.sca(ax=1, fixedlabel="subplot 1"):
    fig.basemap(...)
```

pygmt/src/subplot.py
weiji14: Ok, `fig.sca(ax=[0, 1], ...)` works now in commit 6eb58eada1e7ff6ac2aa0405470ee370523f81b1. Did a one-liner workaround using `ax = ",".join(map(str, ax)) if is_nonstr_iter(ax) else ax`.

willschlitzer:
@weiji14 Not to throw a last-minute change right before we're also trying to get this PR closed out, but did you ever look at making a Subplot() class that is a sub-class of Figure(), with every subplot being it's own Python object? I realize this is a complete change over what I suggested previously, but I think that would provide a clear visual breakdown of which plotting elements are applied to which subplot without needing a context manager/with statement.

weiji14:
> @weiji14 Not to throw a last-minute change right before we're also trying to get this PR closed out, but did you ever look at making a Subplot() class that is a sub-class of Figure(), with every subplot being it's own Python object?

Yes, this (Subplot() sub-class) was pretty much the previous v2 design at #427, though not exactly if I follow your reasoning. One problem with that previous design was that everything in a Subplot instance was a subplot, and you couldn't add say, a colorbar or legend outside of the subplot for the whole figure (after `subplot end`). The thing is that a 'Subplot' (all panels) exists **inside** of a larger 'Figure'.

> I realize this is a complete change over what I suggested previously, but I think that would provide a clear visual breakdown of which plotting elements are applied to which subplot without needing a context manager/with statement.

Actually, could you make a rough skeleton code sketch of how this would work? I'm actually not good with Python classes, more of a functional programming type :blush:

willschlitzer:
I'll take a look at it! I'm not too smart about Python classes as well, so I could be thinking of class inheritance that doesn't actually work.

For the sake of getting this wrapped, since I would consider it the flagship new function of v0.3, I like the function as it currently stands and don't think it's bad to have multiple ways to create subplots, as long as we create sufficient tutorials for it.

weiji14:
> For the sake of getting this wrapped, since I would consider it the flagship new function of v0.3, I like the function as it currently stands and don't think it's bad to have multiple ways to create subplots, as long as we create sufficient tutorials for it.

Yes, it's been a long time coming since the original implementation in #412 (Apr 2020), and I really want to close our oldest issue #20 (Jul 2017)!

willschlitzer:
I'm waiting for the tests to finish running and I'll squash and merge #808 in the next few minutes.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+        plotting but can be accessed by modules that plot scales, bars, text,
+        etc.  This setting overrides the common clearances set by **clearance**
+        in the initial :meth:`pygmt.Figure.subplot` call.
+
+    {V}
+    """
+    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+    kwargs["A"] = f'"{kwargs.get("A")}"' if kwargs.get("A") else None
+    # convert tuple or list to comma-separated str
+    panel = ",".join(map(str, panel)) if is_nonstr_iter(panel) else panel
```
seisman, line 210: I'm OK with the codes, but I feel we can do better checking if `panel` is valid. For example, here are some invalid `panel`:

- `"0, 0"`: space is not allowed if `panel` is string
- `[0, 0, 0]`: when the list is longer than 2, it seems GMT only reads the first two, so it works. Perhaps we can check and give a warning?

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+
+    with Session() as lib:
+        arg_str = " ".join(["set", f"{panel}", build_arg_string(kwargs)]).strip()
```
seisman, line 213: Is `.strip()` necessary?
```suggestion
        arg_str = " ".join(["set", f"{panel}", build_arg_string(kwargs)])
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+    kwargs["T"] = f'"{kwargs.get("T")}"' if kwargs.get("T") else None
+
+    with Session() as lib:
+        try:
+            arg_str = " ".join(["begin", f"{nrows}x{ncols}", build_arg_string(kwargs)])
+            lib.call_module(module="subplot", args=arg_str)
+            yield
+        finally:
+            v_arg = build_arg_string({"V": kwargs.get("V")})
+            lib.call_module("subplot", f"end {v_arg}".strip())
```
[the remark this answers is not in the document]
weiji14: Not really. It's to strip out the trailing whitespace if V is None.

pygmt/src/subplot.py
weiji14: > - `"0, 0"`: space is not allowed if `panel` is string

Ok with checking this, could do something like `' ' not in panel`.

> - `[0, 0, 0]`: when the list is longer than 2, it seems GMT only reads the first two, so it works. Perhaps we can check and give a warning?

Probably don't need to add extra logic to check for this. I doubt it'll be very common, and as you said, things still work so it isn't fatal.

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,239 @@
...
+    fig.basemap(
+        region=[0, 20, 0, 10], projection="X?", frame=["af", "WSne"], panel=[0, 1]
+    )
+    fig.basemap(
+        region=[0, 10, 0, 20], projection="X?", frame=["af", "WSne"], panel=[1, 0]
+    )
+    fig.basemap(
+        region=[0, 20, 0, 20], projection="X?", frame=["af", "WSne"], panel=[1, 1]
+    )
+fig.show()
```
seisman, line 117: Some suggestions on this example:

1. this example uses a new way to specify the panel. We should mention that all plotting functions can use the `panel` argument to specify the panel, and subsequent plotting commands without giving `panel` will add more elements to the same panel.
2. The `frame` arguments are the same for all subplots. For this case, we usually can move it to `Figure.subplot()`. Do you think we should do that way instead?

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,239 @@
...
+#     In fact, we can just use ``fig.basemap(..., panel=True)`` without
+#     specifying any subplot index number, and GMT will automatically activate
+#     the next subplot panel.
+
+###############################################################################
+# Shared X and Y axis labels
+# --------------------------
+# In the example above with the four subplots, the two subplots for each row
+# have the same Y-axis range, and the two subplots for each column have the
+# same X-axis range. You can use the **layout** option to set a common X and/or
```
seisman, line 157:
```suggestion
# same X-axis range. You can use the ``layout`` argument to set a common X and/or
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+          margins). E.g. ['5c', '3c'].
+        - a set of four values (for setting separate left, right, bottom, and
+          top margins). E.g. ['1c', '2c', '3c', '4c'].
+
+        The actual gap created is always a sum of the margins for the two
+        opposing sides (e.g., east plus west or south plus north margins)
+        [Default is half the primary annotation font size, giving the full
+        annotation font size as the default gap].
+    {R}
+    layout : str or list
```
seisman, line 113: Can we do `sharex` and `sharey`, similar to matplotlib subplot?

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+fig.show()
+
+###############################################################################
+# In this example, we define a 2-row, 2-column (2x2) subplot layout using
+# :meth:`pygmt.Figure.subplot`. The overall figure dimensions is set to be 15cm
+# wide and 6cm high (``figsize=["15c", "6c"]``). In addition, we used some
+# optional parameters to fine tune some details of the figure creation:
+#
+# - ``autolabel=True``: Each subplot is automatically labelled abcd
```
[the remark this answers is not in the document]
weiji14: Great catch, was a bug introduced in 4126c16160c67f8d2c405e7d5da56145adc989d4, I've fixed it in eadb8470065da5436802bc796e868e1ac28d8ac7 and modified a test slightly to make sure we don't regress.

examples/tutorials/subplots.py
weiji14: Ok, done in f568982d6bcef13f51c125dfd1cf8aeb18360e6f. Added a note for (1) and moved the `frame` to `fig.subplot` for (2).

pygmt/src/subplot.py
weiji14: > > * `"0, 0"`: space is not allowed if `panel` is string
> 
> Ok with checking this, could do something like `' ' not in panel`.

Actually, do we want to check this even? I can add the logic in `fig.set_panel`, but we will be inconsistent if this check isn't added to other plotting modules that accept `panel` (e.g. `fig.basemap`, `fig.coast`, etc).

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+    use_alias,
+)
+
+
+@fmt_docstring
+@contextlib.contextmanager
+@use_alias(
+    Ff="figsize",
+    Fs="subsize",
+    A="autolabel",
```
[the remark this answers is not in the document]
weiji14: I'd prefer to keep `autolabel` which is consistent with GMT.jl (see https://www.generic-mapping-tools.org/GMT.jl/v0.28/#GMT.subplot). If we change it to `autotag`, we should change `fixedlabel` to `fixedtag` too (which sounds strange), and also a lot of work to update upstream GMT docs.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,215 @@
...
+    B="frame",
+    C="clearance",
+    J="projecton",
+    M="margins",
+    R="region",
+    S="layout",
+    T="title",
```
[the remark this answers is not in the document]
weiji14: Same here, prefer to keep T=title as with GMT.jl at https://www.generic-mapping-tools.org/GMT.jl/v0.28/#GMT.subplot. Heading can also imply "heading in some direction", and isn't as good a name :slightly_smiling_face:

pygmt/src/subplot.py
weiji14: Good idea, I'll look into this, would be good to split up the layout (S) docs actually into x and y components.

pygmt/src/subplot.py
seisman: I don't like the inconsistency. We may implement it after we get rid of the alias mechanism (see https://github.com/GenericMappingTools/pygmt/issues/262#issue-399849306)

pygmt/src/subplot.py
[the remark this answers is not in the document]
seisman: OK, I don't have a strong preference.

pygmt/src/subplot.py
weiji14: Ok, so sticking with no checks then.

weiji14:
Any other comments @GenericMappingTools/python-maintainers? I'd like to get this done by the weekend if possible (but don't hesitate to ping me on any quality issues). Check the following documentation pages and see if there's any mistakes:

- https://pygmt-git-subplotv3.gmt.vercel.app/tutorials/subplots.html
- https://pygmt-git-subplotv3.gmt.vercel.app/api/generated/pygmt.Figure.subplot.html

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    V="verbose",
+    X="xshift",
+    Y="yshift",
+)
+@kwargs_to_strings(Ff="sequence", Fs="sequence", M="sequence", R="sequence")
+def subplot(self, nrows=1, ncols=1, **kwargs):
+    r"""
+    Create multi-panel subplot figures.
+
+    The **subplot** module is used to split the current figure into a
```
seisman, line 39:
```suggestion
    This function is used to split the current figure into a
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    rectangular layout of subplots that each may contain a single self-
+    contained figure. Begin by defining the layout of the entire multi-panel
+    illustration. Several options are available to specify the systematic
```
seisman, line 42:
```suggestion
    illustration. Several arguments are available to specify the systematic
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        *refpoint*][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+r**\|\ **R**]
+        [**+v**].
+        Specify automatic tagging of each subplot. Append either a number or
+        letter [a]. This sets the tag of the first, top-left subplot and others
+        follow sequentially. Surround the number or letter by parentheses on
+        any side if these should be typeset as part of the tag. Use **+j|J**\
+        *refpoint* to specify where the tag should be placed in the subplot
+        [TL]. Note: **+j** sets the justification of the tag to *refpoint*
+        (suitable for interior tags) while **+J** instead selects the mirror
+        opposite (suitable for exterior tags). Append **+c**\ *dx*[/*dy*] to
```
seisman, line 71:
```suggestion
        opposite (suitable for exterior tags). Append **+c**\ *dx*\[/*dy*] to
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        *fill* [no painting]. Append **+o**\ *dx*\ [/*dy*] to offset the tag's
+        reference point in the direction implied by the justification [4p/4p,
+        i.e., 20% of the :gmt-term:`FONT_TAG` size]. Append **+p**\ *pen* to
+        draw the outline of the tag's text box using selected *pen* [no
+        outline]. Append **+r** to typeset your tag numbers using lowercase
+        Roman numerals; use **+R** for uppercase Roman numerals [Arabic
+        numerals]. Append **+v** to increase tag numbers vertically down
+        columns [horizontally across rows].
+    {B}
+    clearance : str
```
seisman, line 84:
```suggestion
    clearance : str or list
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        [*side*]\ *clearance*.
+        Reserve a space of dimension *clearance* between the margin and the
+        subplot on the specified side, using *side* values from **w**, **e**,
+        **s**, or **n**, or **x** for both **w** and **e** or **y** for both
+        **s** and **n**.  No *side* means all sides. The option is repeatable
+        to set aside space on more than one side. Such space will be left
```
seisman, line 90: >  The option is repeatable to set aside space on more than one side.

This sentence doesn't make sense for PyGMT functions. We should rephrase it to a list of *clearance*.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        untouched by the main map plotting but can be accessed by modules that
+        plot scales, bars, text, etc.  Settings specified under **begin**
+        directive apply to all subplots, while settings under **set** only
+        apply to the selected (active) subplot. **Note**: Common options
+        **x_offset** and **y_offset** are not available during subplots; use
+        **clearance** instead.
```
seisman, line 96: This sentence doesn't make sense to PyGMT.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    sharey : bool or str
+        Set subplot layout for shared y-axes. Use when all subplots in a row
+        share a common *y*-range. If ``sharey=True``, the first (i.e.,
+        **l**\ eft) and the last (i.e., **r**\ ight) columns will have
+        *y*-annotations; use ``sharey='l'`` or ``sharey='r'`` to select only
+        one of those two columns [both]. Append **+l** if annotated *y*-axes
+        will have a label [none]; optionally, append the label if it is the
+        same for the entire subplot. Append **+p** to make all annotations
+        axis-parallel [horizontal]; if not used you may have to set
+        **clearance** to secure extra space for long horizontal annotations.
```
seisman, line 133: Use code blocks for PyGMT arguments.
```suggestion
        ``clearance`` to secure extra space for long horizontal annotations.
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+
+        Notes for sharex/sharey:
+
+        - Labels and titles that depends on which row or column are specified
+          as usual via a subplot's own **frame** setting.
+        - Append **+w** to the **figsize** or **subsize** argument to draw
+          horizontal and vertical lines between interior panels using selected
+          pen [no lines].
```
seisman, line 141:
```suggestion
        - Labels and titles that depends on which row or column are specified
          as usual via a subplot's own ``frame`` setting.
        - Append **+w** to the ``figsize`` or ``subsize`` argument to draw
          horizontal and vertical lines between interior panels using selected
          pen [no lines].
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    title : str
+        While individual subplots can have titles (see **layout** or
+        **frame**), the entire figure may also have an overarching *heading*
+        [no heading]. Font is determined by setting :gmt-term:`FONT_HEADING`.
```
seisman, line 145:
```suggestion
    title : str
        While individual subplots can have titles (see ``layout`` or
        ``frame``), the entire figure may also have an overarching *heading*
        [no heading]. Font is determined by setting :gmt-term:`FONT_HEADING`.
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        *row,col*\|\ *index*.
+        Sets the current subplot until further notice. **Note**: First *row*
+        or *col* is 0, not 1. If not given we go to the next subplot by order
+        specified via **autolabel** in :meth:`pygmt.Figure.subplot`. As an
+        alternative, you may bypass the **sca** mode and instead supply the
+        common option **panel**=\ [*row,col*] to the first plot command you
+        issue in that subplot. GMT maintains information about the current
+        figure and subplot. Also, you may give the one-dimensional *index*
+        instead which starts at 0 and follows the row or column order set via
+        **autolabel** in :meth:`pygmt.Figure.subplot`.
```
seisman, line 193: Incorrectly rendered, and `sca` no longer makes sense.

pygmt/src/subplot.py
weiji14: Ok, I'll rephrase it and add some examples in code-blocks.

pygmt/src/subplot.py
weiji14: Yeah, I'll delete the sentences "Settings specified under **begin** ... use **clearance** instead".

pygmt/src/subplot.py
weiji14: Ok, fixed rendering using `**panel**\ =[*row,col*]`, and swapped `sca` to `:meth:pygmt.Figure.set_panel`

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    Parameters
+    ----------
+    nrows : int
+        Number of vertical rows of the subplot grid.
+    ncols : int
+        Number of horizontal columns of the subplot grid.
+    figsize : tuple
+        Specify the final figure dimensions as ``(width, height)``.
+    subsize : tuple
+        Specify the dimensions of each subplot directly as ``(width, height)``.
```
seisman, line 58: Add a check to make sure that users don't give both `figsize` and `subsize`?

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+    Full option list at :gmt-docs:`subplot.html#synopsis-begin-mode`
+
+    {aliases}
+
+    Parameters
+    ----------
+    nrows : int
+        Number of vertical rows of the subplot grid.
+    ncols : int
+        Number of horizontal columns of the subplot grid.
```
seisman, line 54: Check *nrows* and *ncols* are intergers (>0)?

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+        share a common *y*-range. If ``sharey=True``, the first (i.e.,
+        **l**\ eft) and the last (i.e., **r**\ ight) columns will have
+        *y*-annotations; use ``sharey='l'`` or ``sharey='r'`` to select only
+        one of those two columns [both]. Append **+l** if annotated *y*-axes
+        will have a label [none]; optionally, append the label if it is the
+        same for the entire subplot. Append **+p** to make all annotations
+        axis-parallel [horizontal]; if not used you may have to set
+        ``clearance`` to secure extra space for long horizontal annotations.
+
+        Notes for sharex/sharey:
```
seisman, line 133:
```suggestion
        Notes for ``sharex``/``sharey``:
```

pygmt/src/subplot.py
```diff
@@ -0,0 +1,222 @@
...
+
+        - Labels and titles that depends on which row or column are specified
+          as usual via a subplot's own ``frame`` setting.
+        - Append **+w** to the ``figsize`` or ``subsize`` argument to draw
+          horizontal and vertical lines between interior panels using selected
+          pen [no lines].
+    title : str
+        While individual subplots can have titles (see ``layout`` or
```
seisman, line 141: What's `layout`?

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,244 @@
...
+# ``sharex="b"`` indicates that subplots in a column will share the x-axis, and
+# only the **b**\ ottom axis is displayed. ``sharey="l"`` indicates that
+# subplots within a **R**\ ow will share the y-axis, and only the **l**\ eft
+# axis is displayed.
+#
+# Of course, instead of using the **sharex**/**sharey** option, you can also
+# set a different **frame** for each subplot to control the axis properties
+# individually for each subplot.
+
+###############################################################################
```
seisman, line 191: Considering the bugs and the tricks, do you think we should call `subplot` twice to get the same advanced layout? 
```python
import pygmt
fig = pygmt.Figure()

with fig.subplot(nrows=1, ncols=2, figsize=("15c", "6c"), autolabel="b)"):
    fig.basemap(
        region=[0, 5, 0, 5], projection="X?", frame=["af", "WSne"], panel=[0, 0]
    )
    fig.basemap(
        region=[0, 5, 0, 5], projection="X?", frame=["af", "WSne"], panel=[0, 1]
    )
    
with fig.subplot(nrows=1, ncols=1, figsize=("15c", "6c"), autolabel="a)", yshift="h+2c"):
    fig.basemap(
        region=[0, 10, 0, 10], projection="X?", frame=["af", "WSne"], panel=[0, 0]
    )
    fig.text(text="TEXT", x=5, y=5)

fig.show()
```

pygmt/tests/test_subplot.py
```diff
@@ -0,0 +1,87 @@
...
+            fig_test.basemap(region=[0, 3, 0, 3], frame="+tplot0")
+        with fig_test.set_panel(panel=[0, 1]):
+            fig_test.basemap(region=[0, 3, 0, 3], frame="+tplot1")
+    return fig_ref, fig_test
+
+
+@check_figures_equal()
+def test_subplot_direct():
+    """
+    Plot map elements to subplot directly using ax argument.
```
seisman, line 31:
```suggestion
    Plot map elements to subplot directly using panel argument.
```

pygmt/src/subplot.py
weiji14: Good catch, it's changed to `sharex`/`sharey` now.

pygmt/src/subplot.py
weiji14: Done at c475861383423bfa312e2bd625a4ea61682c14be.

pygmt/src/subplot.py
weiji14: Done at c475861383423bfa312e2bd625a4ea61682c14be.

examples/tutorials/subplots.py
weiji14: I was actually thinking whether to take this section out. But your code example looks good so I think I'll switch to yours. Will edit the documentation notes too.

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,236 @@
...
+# on the top row by using ``fig.subplot(..., yshift="h+1c")`` which shifts the
+# plot origin 1 cm beyond the **h**\ eight of the previous (bottom row) plot.
+# You may need to adjust this ``yshift`` parameter to make your plot look nice.
+# This top row uses ``autolabel="a)"``, and we also plotted some text inside.
+# Note that ``projection="X?"`` was used to let GMT automatically determine the
+# size of the subplot according to the size of the subplot area.
+
+###############################################################################
+# You can also manually override the ``autolabel`` for each subplot using for
+# example, ``fig.set_panel(..., fixedlabel="b) Panel 2"`` which would allow you
```
seisman, line 234:
```suggestion
# example, ``fig.set_panel(..., fixedlabel="b) Panel 2")`` which would allow you
```

pygmt/tests/test_subplot.py
```diff
@@ -0,0 +1,110 @@
+"""
```
seisman, line 1: I'm afraid some of the tests may always pass even though we make some backward-incompatible changes. We can explore how to write better and robust tests later.

examples/tutorials/subplots.py
```diff
@@ -0,0 +1,236 @@
...
+    fig.text(text="TEXT", x=5, y=5)
+
+fig.show()
+
+###############################################################################
+#
+# We start by drawing the bottom two subplots, setting ``autolabel="b)"`` so
+# that the subplots are labelled 'b)' and 'c)'. Then, we plot a single subplot
+# on the top row by using ``fig.subplot(..., yshift="h+1c")`` which shifts the
+# plot origin 1 cm beyond the **h**\ eight of the previous (bottom row) plot.
```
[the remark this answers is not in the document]
weiji14: Yep, good point. Updated in e0900e9f18df0d73a754c336cbbdebf05c089ff4.

pygmt/tests/test_subplot.py
weiji14: Yeah, definitely not my best attempt as it's mostly testing the aliases and a few subtle quirks. There are probably more complicated tests we can write (especially around `fig.set_panel`), but probably best to stress test things in the real world and make changes as we go along.

pygmt/src/subplot.py
```diff
@@ -0,0 +1,232 @@
...
+    autolabel : bool or str
+        [*autolabel*][**+c**\ *dx*\ [/*dy*]][**+g**\ *fill*][**+j**\|\ **J**\
+        *refpoint*][**+o**\ *dx*\ [/*dy*]][**+p**\ *pen*][**+r**\|\ **R**]
+        [**+v**].
+        Specify automatic tagging of each subplot. Append either a number or
+        letter [a]. This sets the tag of the first, top-left subplot and others
+        follow sequentially. Surround the number or letter by parentheses on
+        any side if these should be typeset as part of the tag. Use **+j|J**\
+        *refpoint* to specify where the tag should be placed in the subplot
+        [TL]. Note: **+j** sets the justification of the tag to *refpoint*
```
seisman, line 72: When I review the PR #881 from @meghanrjones, I realize that the default values are not documented well. Most readers may find `[TL]` weird, and don't understand what it is, especially that `[ ]` means optional arguments in GMT's syntax. I feel that we should use `[Default is TL]`, rather than just `[TL]` for better clarification. 

If you agree, then you need to update the docstrings again, and better to document it in the contributing guides.


```suggestion
        [Default is **TL**]. Note: **+j** sets the justification of the tag to *refpoint*
```

seisman (commented):
@weiji14 I believe this PR is very close to merging!

pygmt/src/subplot.py
weiji14: > When I review the PR #881 from @meghanrjones, I realize that the default values are not documented well. Most readers may find `[TL]` confusing, especially that `[ ]` means optional arguments in GMT's syntax. I feel that we should use `[Default is TL]`, rather than just `[TL]` for better clarification.

Yes, I hear you that using `[Default is TL]` is clearer, but:

1. There are a lot of these `[ ]`, and it's not easy to CTRL+F and replace all instances.
2. We have mostly been copying and pasting from upstream GMT, and it will be a lot of work checking through them all (c.f. #895).

> If you agree, then you need to update the docstrings again, and better to document it in the contributing guides.

I think it's ok for `[ ]` to mean both optional and default as long as it's documented, or like you say, use `[Default is TL]` instead. I would rather we settle on the standard we want before applying it here in this PR.

seisman: OK, we can discuss it later.

seisman (approved):

weiji14:
All green and merged! Thanks @seisman for your patience in all of this :smile:

liamtoney:
Big cheers on this @weiji14 — this is going to be a huge feature for many users! 🎉

PR #344, opened 2019-10-11
Continuous Documentation

weiji14:
**Description of proposed changes**

Making it easier to preview documentation changes in Pull Requests via 'Continuous Documentation'. Currently deployed to serverless infrastructure on [Zeit Now](https://zeit.co/now).

Zeit Now is meant to be a zero-config/serverless, but that's for the javascript world. I've hacked the `package.json` to get `sphinx` and `pygmt` up and running to build the docs. The static files are then copied to a `public` folder where it gets deployed automatically to a url that looks like `pygmt-abcdef.now.sh`. You can see an example at https://pygmt-l8nmyn8mx.now.sh/.

Merge after #343, Fixes #342

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #830, opened 2021-02-01
WIP: Parallelize Continuous Integration test and documentation build

weiji14:

Split the unit tests and documentation build into two separate and parallel paths to reduce total time spent on Continuous Integration. Also only performing documentation build on Python 3.9 only to conserve CI resources. Edit: doesn't work yet...

image

This should save an estimated ~4 min of CI time (the unit tests and docs build take about 4 min respectively, i.e. 8 mins in serial, 4 min in parallel), so the longest running Windows CI should go from ~18min to ~14min.

References:
- https://hanxiao.io/2021/01/24/Speedup-CI-Workflow-in-Github-Actions-via-Strategy-Matrix/
- See example at https://github.com/jina-ai/jina/actions/runs/507130336
- https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/16
- https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/

Part of #584

**Reminders**

**Notes**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -47,7 +47,7 @@ jobs:
           find . -type f -not -path '*/\.git/*' -exec grep -Iq . {} \; -exec chmod 644 {} \;
           if [[ $(git ls-files -m) ]]; then git --no-pager diff HEAD; exit 1; fi
 
-  test:
+  install-deps:
     name: ${{ matrix.os }} - Python ${{ matrix.python-version }}
     runs-on: ${{ matrix.os }}
     strategy:
@@ -55,10 +55,6 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9]
         os: [ubuntu-latest, macOS-latest, windows-latest]
-    # environmental variables used in coverage
-    env:
-      OS: ${{ matrix.os }}
-      PYTHON: ${{ matrix.python-version }}
 
     steps:
       # Cancel previous runs that are not completed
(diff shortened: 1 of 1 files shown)


weiji14:
Just resurfacing this comment by Leo from https://github.com/GenericMappingTools/pygmt/pull/344#issuecomment-542090505:

> > We could probably disable those documentation builds in a separate PR to speed them up?

> I’m not too keen on that since he docs builds are a sort of test. But to all builds need to do this (probably only one per OS would be good).

And also at https://github.com/GenericMappingTools/pygmt/pull/964#pullrequestreview-599073327:

> - Can we set up a matrix build on all 3 OSes in this action? I.e. remove the "Build the documentation" section in `ci_tests.yaml` and put it here instead.

> Rationale: Saves ~4min of CI time. I wanted to separate the tests from the documentation builds in #830, but maybe we can do it in this "Continuous Documentation" workflow instead. I know Leo mentioned using the documentation build as a 'test' before (lazy to dig up that comment), but our >20 min of CI tests is getting wayy too long.

So we should probably run the docs build on only one Python version (e.g. Python 3.9) instead of three (3.7, 3.8, 3.9) to further save on CI resources.

seisman:
I agree that we should separate "Building the documentation" and "Full tests" into two separate workflows. We can have faster jobs runs and also can control the jobs runs based on files, e.g., we don't want to run the full tests if only gallery examples are changed.

I also agree that we should only build the documentation for one Python version (e.g., 3.9), not three versions.

weiji14:
Closing in favour of alternative implementation at #1033

PR #839, opened 2021-02-05
Skip workflows in PRs if only non-source-code files are changed

### LINKED ISSUE #837: Do not run workflows if only non-source-code files are changed

Sometimes, we make changes to plain text files (.md, .rst), without changing any source codes. We know that the changes won't break our tests, but the workflow still runs 9 jobs (Ubuntu/macOS/Windows + Python 3.7/3.8/3.9). It's a waste of CI resources. For example, #726, #802, and #836.

GitHub Actions actually can skip workflows based on the files that are changed, 

> When using the `push` and `pull_request` events, you can configure a workflow to run when at least one file does not match `paths-ignore` or at least one modified file matches the configured `paths`. Path filters are not evaluated for pushes to tags.

see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths for details.

We may explore this to reduce our CI usages. IMHO, we can skip running workflows if only the following files are changes:

- `*.md`
- `**/*.rst`
- `doc/**`
- `examples/**` (?)

Note that, the documentation is still built by vercel, so documentation preview still works.

**Are you willing to help implement and maintain this feature?** Yes, but let's discuss it first and contributions are welcomed.

seisman:

Skip workflows in PRs if **ONLY** the following are changed:

```
- 'doc/**'
- '*.md'
- '*.json'
- 'README.rst'
- 'LICENSE.txt'
```

Any other changes (especially changes in `pygmt/` and `examples/`) still trigger the workflows.

**Notes:**

- `examples/` directory is not excluded because some examples may pass on one OS but fail on another one.
- the workflow is always triggered in the master branch even though only plaintext files are changed

Fixes #837.

Reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

**Reminders**

**Slash Commands**


seisman:
If I understand the documentation correctly, it should work. However, I can't test it.

.github/workflows/ci_tests.yaml
```diff
@@ -7,6 +7,12 @@ on:
   push:
     branches: [ master ]
   pull_request:
+    paths-ignore:
+      - 'doc/**'
+      - '*.md'
+      - '*.json'
+      - 'README.rst'
+      - 'LICENSE.txt'
```
weiji14, line 15: So if a git commit updates any file that matches these patterns (`doc/**`, `*.md`, etc), that means the workflow won't run? I guess that's ok, as long as we don't commit changes to a *.py and *.md file at the same time, or make sure to trigger the test somehow by editing only a *.py file.

 A problem is that someone with administrative privileges might be required to merge the commit if the 'required checks' are not triggered. But I guess that's ok.

weiji14 (approved):

.github/workflows/ci_tests.yaml
seisman: > So if a git commit updates any file that matches these patterns (`doc/**`, `*.md`, etc), that means the workflow won't run?

No. This is what the documentation says:

> When using the push and pull_request events, you can configure a workflow to run when at least one file does not match paths-ignore or at least one modified file matches the configured paths.

If we make changes to a *.py and a .md file, the workflow still runs.

weiji14: > If we make changes to a *.py and a .md file, the workflow still runs.

Ah ok, perfect!

Issue #374, opened 2019-11-08
Integration of pygmt with xarray and pyshtools


MarkWieczorek:
# Plotting of projected images using xarray and/or pyshtools

It would be desirable to add a method such as `plot_gmt()` or `gmt.plot()` to xarray DataArrays and pyshtools SHGrids (the latter of which are easily converted to DataArrays). Through a single command, it would thus be possible to create a publication quality projected image. Such a built in method would have a set of default plotting parameters defined, and would require only minimal modifications on the part of the user.

# pyshtools implementation

I have just made a pull request at pyshtools https://github.com/SHTOOLS/SHTOOLS/pull/203 which is a first attempt of plotting global grids. This method implements all global and hemispherical projects, and latter will be updated to work with non-global projections. By simply calling `plot_gmt()` a global mollweide projected image is created.

This method defines a large number of human-readable parameters that differ from the standard gmt syntax (which is only understandable by an expert), but also differs in places by what pygmt uses. Ideally, it would be useful if we could try to standardize the syntax among the various gmt-related projects. Here is what I am currently using, but I could be easily convinced to change these:

* `projection`: The name of a global or hemispherical projection (see Notes). Only the first three characters are necessary to identify the projection.
* `region`: The map region, consisting of a list [West, East, South, North] in degrees. The default 'g' specifies the entire sphere.
* `width`: The wdith of the projected image.
* `unit`:  The measurement unit of the figure width: 'i' for inches or 'c' for cm.
* `longitude`: The central meridian or center of the projection.
* `latitude`: The center of the projection for some hemispheric projections.
* `grid`:  If True, plot grid lines.
* `grid_interval`: Grid line interval [latitude, longitude] in degrees.
* `annotate`: If True, plot annotation labels on axes.
* `annotate_interval`: Annotation label interval [latitude, longitude] in degrees.
* `ticks`: If True, plot ticks on axes.
* `tick_interval`:  Tick interval [latitude, longitude] in degrees.
* `axes`: Specify which plot axes should be drawn and annotated. Capital letters draw the axes, ticks, and annotations, whereas small letters draw only the axes and ticks.
* `title`: The title to be displayed above the plot.
* `cmap` : The color map to use when plotting the data and colorbar.
* `cmap_reverse`: Set to True to reverse the sense of the color progression in the color table.
* `continuous`: If True, create a continuous colormap. Default behavior is to use contant colors for each interval.
[...]

PR #853, opened 2021-02-08
Add common alias panel (-c) to all plotting functions

weiji14:

Used to advance to the selected subplot panel. See https://docs.generic-mapping-tools.org/6.1/gmt.html#c-full,
https://github.com/GenericMappingTools/gmt/blob/6.1.1/doc/rst/source/explain_-c_full.rst_, and https://docs.generic-mapping-tools.org/6.1/cookbook/options.html#selecting-subplot-panels-the-c-option. Needed for the `subplot` wrapper at #822 to resolve issue #20.

**Questions**:

1. ~~Just confirming that we want to go with `ax` (as with matplotlib) rather than `panel` (used by GMT). See https://github.com/GenericMappingTools/pygmt/pull/822#issuecomment-770341473.~~ Ok, going with `panel`
2. Haven't tested to see if `-c` works on `inset` (https://docs.generic-mapping-tools.org/6.1/inset.html), i.e. putting an inset in a subplot panel. If it works, then I'll add the alias to the `pygmt/src/inset.py` file.

This can be merged before #822 to reduce the diff there.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -60,16 +60,14 @@ COMMON_OPTIONS = {
             Shift plot origin in y-direction. Full documentation is at
             :gmt-docs:`gmt.html#xy-full`.
          """,
-    "c": r"""panel : int or list
+    "c": r"""panel : int or list or bool
             [*row*\ ,\ *col*\|\ *index*].
-            Used to advance to the selected subplot panel. Only allowed when in
-            subplot mode. Available to all plot modules. If no arguments are
-            given then we advance to the next panel in the selected order. If
-            no **panel** is given and we just entered subplot mode then the
-            first panel (top, left) is selected. Instead of *row, col* you may
-            give the one-dimensional *index* which depends on the order you set
-            via **autolabel** when the subplot was defined. **Note**: *row*,
-            *col*, and *index* all start at 0.
+            Selects a specific subplot panel. Only allowed when in subplot
+            mode. Use ``panel=True`` to advance to the next panel in the
+            selected order. Instead of *row,col* you may also give a scalar
+            value *index* which depends on the order you set via ``autolabel``
+            when the subplot was defined. **Note**: *row*, *col*, and *index*
+            all start at 0.
          """,
     "j": """\
         distcalc : str
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -60,6 +60,17 @@ COMMON_OPTIONS = {
             Shift plot origin in y-direction. Full documentation is at
             :gmt-docs:`gmt.html#xy-full`.
          """,
+    "c": r"""ax : int or list
+            [*row*\ ,\ *col*\|\ *index*].
+            Used to advance to the selected subplot panel. Only allowed when in
+            subplot mode. Available to all plot modules. If no arguments are
+            given then we advance to the next panel in the selected order. If
+            no **ax** is given and we just entered subplot mode then the first
+            panel (top, left) is selected. Instead of *row, col* you may give
+            the one-dimensional *index* which depends on the order you set via
+            **autolabel** when the subplot was defined. **Note**: *row*, *col*,
+            and *index* all start at 0.
+         """,
     "j": """\
         distcalc : str
             ``e|f|g``.
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -60,16 +60,16 @@ COMMON_OPTIONS = {
             Shift plot origin in y-direction. Full documentation is at
             :gmt-docs:`gmt.html#xy-full`.
          """,
-    "c": r"""ax : int or list
+    "c": r"""panel : int or list
             [*row*\ ,\ *col*\|\ *index*].
             Used to advance to the selected subplot panel. Only allowed when in
             subplot mode. Available to all plot modules. If no arguments are
             given then we advance to the next panel in the selected order. If
-            no **ax** is given and we just entered subplot mode then the first
-            panel (top, left) is selected. Instead of *row, col* you may give
-            the one-dimensional *index* which depends on the order you set via
-            **autolabel** when the subplot was defined. **Note**: *row*, *col*,
-            and *index* all start at 0.
+            no **panel** is given and we just entered subplot mode then the
+            first panel (top, left) is selected. Instead of *row, col* you may
+            give the one-dimensional *index* which depends on the order you set
+            via **autolabel** when the subplot was defined. **Note**: *row*,
+            *col*, and *index* all start at 0.
          """,
     "j": """\
         distcalc : str
--- a/.pylintrc
+++ b/.pylintrc
@@ -441,7 +441,6 @@ function-naming-style=snake_case
 good-names=i,
            j,
            k,
-           ax,
            ex,
            Run,
            _,
--- a/.pylintrc
+++ b/.pylintrc
@@ -441,6 +441,7 @@ function-naming-style=snake_case
 good-names=i,
            j,
            k,
+           ax,
            ex,
            Run,
            _,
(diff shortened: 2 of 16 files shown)


seisman:
> * Just confirming that we want to go with `ax` (as with matplotlib) rather than `panel` (used by GMT). See [#822 (comment)](https://github.com/GenericMappingTools/pygmt/pull/822#issuecomment-770341473).

I'm OK with both, although I prefer `panel`. When I learn matplotlib, I find it difficult to understand the differences between "axes" and "axis" (likely because I'm not a native English speaker). BTW, it seems **Julia Plot.jl** uses `subplot` (http://docs.juliaplots.org/latest/layouts/), 

> * Haven't tested to see if `-c` works on `inset` (https://docs.generic-mapping-tools.org/6.1/inset.html), i.e. putting an inset in a subplot panel. If it works, then I'll add the alias to the `pygmt/src/inset.py` file.

I think it doesn't work, to call `inset`, you need to have a larger map first.

seisman:
I also find this (https://matplotlib.org/tutorials/intermediate/artists.html):

> Most of you are probably familiar with the Subplot, which is just a special case of an Axes that lives on a regular rows by columns grid of Subplot instances. 

If I understand it correctly, GMT's subplots also live on a regular rows by columns grid, so it should be called `subplot`, not `axes`, right?

weiji14:
> > * Just confirming that we want to go with `ax` (as with matplotlib) rather than `panel` (used by GMT). See [#822 (comment)](https://github.com/GenericMappingTools/pygmt/pull/822#issuecomment-770341473).
> 
> I'm OK with both, although I prefer `panel`. When I learn matplotlib, I find it difficult to understand the differences between "axes" and "axis" (likely because I'm not a native English speaker).

Oh, and just FYI, there's a Python library called [Panel](https://panel.holoviz.org/) for making dashboards (i.e. interactive subplots). To be fair, I don't really get the "axes" and "axis" terminology used by matplotlib either, "axes" should be the plural of "axis" but they don't use it that way.

> > * Haven't tested to see if `-c` works on `inset` (https://docs.generic-mapping-tools.org/6.1/inset.html), i.e. putting an inset in a subplot panel. If it works, then I'll add the alias to the `pygmt/src/inset.py` file.
> 
> I think it doesn't work, to call `inset`, you need to have a larger map first.

Ok, will leave out `inset` for now then.

> I also find this (https://matplotlib.org/tutorials/intermediate/artists.html):
> 
> > Most of you are probably familiar with the Subplot, which is just a special case of an Axes that lives on a regular rows by columns grid of Subplot instances.
> 
> If I understand it correctly, GMT's subplots also live on a regular rows by columns grid, so it should be called `subplot`, not `axes`, right?

Not sure if it's a good idea to use 'subplot' as an alias for 'c' (if that's what you mean). We'll hit into the same problem as with `def text(text=...)` as in https://github.com/GenericMappingTools/pygmt/pull/808#issuecomment-769018773.

Will all that said, I would still argue that using `ax` is the way to go, considering that both [`pandas.plot`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html) and [`xarray.plot`](http://xarray.pydata.org/en/v0.16.2/generated/xarray.DataArray.plot.html) has a kwarg called `ax`. It might be useful to use `ax` too (or perhaps confusing) to integrate with other libraries (c.f. #374, https://github.com/GenericMappingTools/pygmt/issues/366#issuecomment-559428840) in the future.

seisman:
> Not sure if it's a good idea to use 'subplot' as an alias for 'c' (if that's what you mean). We'll hit into the same problem as with `def text(text=...)` as in [#808 (comment)](https://github.com/GenericMappingTools/pygmt/pull/808#issuecomment-769018773).

Yes, I agree, `subplot` may be a bad idea.

> With all that said, I would still argue that using `ax` is the way to go, considering that both [`pandas.plot`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html) and [`xarray.plot`](http://xarray.pydata.org/en/v0.16.2/generated/xarray.DataArray.plot.html) has a kwarg called `ax`. It might be useful to use `ax` too (or perhaps confusing) to integrate with other Python libraries (c.f. #374, [#366 (comment)](https://github.com/GenericMappingTools/pygmt/issues/366#issuecomment-559428840)) in the future.

`pandas.plot` and `xarray.plot` have the `ax` kwarg because they are integrating well with matplotlib. However, `pygmt` can't insert a GMT figure to a matplot axes, right?

willschlitzer:
Little late to the party, but my vote would be `panel` over `ax`. I understand the potential conflicts of `panel` and the familiarity of `ax` with other plotting packages, but I think `panel` makes more sense in this context, as the specific panel to be plotting on is being set, not just the axis/axes (even though they refer to the same thing).

weiji14:
> Little late to the party, but my vote would be `panel` over `ax`. I understand the potential conflicts of `panel` and the familiarity of `ax` with other plotting packages, but I think `panel` makes more sense in this context, as the specific panel to be plotting on is being set, not just the axis/axes (even though they refer to the same thing).

Ok, if we want to go with `panel`, then we should change `fig.sca` too (short for set current ax) in #822 to aomething else. `fig.set_panel` perhaps?

willschlitzer:
> > Little late to the party, but my vote would be `panel` over `ax`. I understand the potential conflicts of `panel` and the familiarity of `ax` with other plotting packages, but I think `panel` makes more sense in this context, as the specific panel to be plotting on is being set, not just the axis/axes (even though they refer to the same thing).
> 
> Ok, if we want to go with `panel`, then we should change `fig.sca` too (short for set current ax) in #822 to something else. `fig.set_panel` perhaps?

I think `set_panel` works well. I particularly like that it doesn't use an acronym.

seisman:
> > Little late to the party, but my vote would be `panel` over `ax`. I understand the potential conflicts of `panel` and the familiarity of `ax` with other plotting packages, but I think `panel` makes more sense in this context, as the specific panel to be plotting on is being set, not just the axis/axes (even though they refer to the same thing).
> 
> Ok, if we want to go with `panel`, then we should change `fig.sca` too (short for set current ax) in #822 to something else. `fig.set_panel` perhaps?

`fig.set_panel` looks good to me.

pygmt/helpers/decorators.py
```diff
@@ -60,6 +60,17 @@
...
+    "c": r"""panel : int or list
+            [*row*\ ,\ *col*\|\ *index*].
+            Used to advance to the selected subplot panel. Only allowed when in
+            subplot mode. Available to all plot modules. If no arguments are
+            given then we advance to the next panel in the selected order. If
+            no **panel** is given and we just entered subplot mode then the
+            first panel (top, left) is selected. Instead of *row, col* you may
+            give the one-dimensional *index* which depends on the order you set
+            via **autolabel** when the subplot was defined. **Note**: *row*,
+            *col*, and *index* all start at 0.
```
weiji14, line 72: Feel like we should cut out some sentences in this, any ideas?

weiji14 (commented):
Ok I've renamed ax -> panel. Will do the `fig.set_panel` rename later in #822, but just one comment/question.

pygmt/helpers/decorators.py
```diff
@@ -60,6 +60,17 @@
             Shift plot origin in y-direction. Full documentation is at
             :gmt-docs:`gmt.html#xy-full`.
          """,
+    "c": r"""panel : int or list
```
seisman, line 63: `panel=True` is also allowed, right?
```suggestion
    "c": r"""panel : int or list or bool
```

pygmt/helpers/decorators.py
seisman, line 72: A draft change:
```suggestion
            Select a specific subplot panel. Only allowed when in
            subplot mode. Use ``panel=True`` to advance to the next panel 
            in the selected order. Instead of *row,col* you may
            give the one-dimensional *index* which depends on the order you set
            via ``autolabel`` when the subplot was defined. **Note**: *row*,
            *col*, and *index* all start at 0.
```

weiji14: Ok, revised at commit 65fe92f0bc2ba13a3270b2bb800e4a1cc584856f. I've edited it slightly and wrapped things to 79 characters.

seisman (approved):

PR #869, opened 2021-02-12
Add test file for draft pull requests

willschlitzer:
As discussed in #584, this is a change of the .yaml files to create a test for draft pull requests. In a draft, the tests only run Python 3.9 on Ubuntu-latest. The entire test suite for different operating systems and Python versions only runs when the pull request is marked ready for review.

**Reminders**

**Slash Commands**


willschlitzer:
@seisman @weiji14 I think I did this correctly and modified the appropriate .yaml files, but I'm not sure how to test it.

seisman:
> In a draft, the tests only run Python 3.9 on Ubuntu-latest.

I can't say I like this.

willschlitzer:
> > In a draft, the tests only run Python 3.9 on Ubuntu-latest.
> 
> I can't say I like this.

Would you recommend something differently, or not reducing the tests at all for drafts? My thought is that ubuntu typically runs the fastest, and this will provide a check on all the tests without requiring so many CI jobs for ever minor change, but still running the gamut of tests when the PR is marked ready for review.

seisman:
@willschlitzer See #906 for the better implementation and don't forget to push the "Delete branch" button if you no longer need it.

PR #877, opened 2021-02-12
CI: Add GMT's bin to PATH for GMT dev tests

### LINKED ISSUE #874: test_load_libgmt_with_a_bad_library_path fails with GMT dev versions on Linux and macOS

In recent runs (https://github.com/GenericMappingTools/pygmt/actions/runs/560287188), `test_load_libgmt_with_a_bad_library_path` fails with GMT dev on Linux and macOS, but pass on Windows, possibly because we merged #702. However, it passed when @weiji14 triggered the dev tests before merging (https://github.com/GenericMappingTools/pygmt/actions/runs/560286224).

It's unclear to me what happens. We can wait for today's scheduled runs and see if it works or not.

See the error messages below:
```
___________________ test_load_libgmt_with_a_bad_library_path ___________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f462205dca0>

    def test_load_libgmt_with_a_bad_library_path(monkeypatch):
        """
        Test that loading still works when given a bad library path.
        """
        # Set a fake "GMT_LIBRARY_PATH"
        monkeypatch.setenv("GMT_LIBRARY_PATH", "/not/a/real/path")
>       assert check_libgmt(load_libgmt()) is None

/usr/share/miniconda3/envs/test/lib/python3.9/site-packages/pygmt/tests/test_clib_loading.py:48: 

    def load_libgmt():
        """
        Find and load ``libgmt`` as a :py:class:`ctypes.CDLL`.
    
        By default, will look for the shared library in the directory specified by
        the environment variable ``GMT_LIBRARY_PATH``. If it's not set, will let
        ctypes try to find the library.
    
        Returns
        -------
        :py:class:`ctypes.CDLL` object
            The loaded shared library.
    
        Raises
        ------
        GMTCLibNotFoundError
            If there was any problem loading the library (couldn't find it or
            couldn't access the functions).
        """
        lib_fullnames = []
        error = True
        for libname in clib_full_names():
            lib_fullnames.append(libname)
            try:
                libgmt = ctypes.CDLL(libname)
                check_libgmt(libgmt)
                error = False
                break
            except OSError as err:
                error = err
        if error:
>           raise GMTCLibNotFoundError(
                "Error loading the GMT shared library "
                f"{', '.join(lib_fullnames)}.\n {error}."
            )
E           pygmt.exceptions.GMTCLibNotFoundError: Error loading the GMT shared library libgmt.so.
E            libgmt.so: cannot open shared object file: No such file or directory.

```

seisman:

See #874 for context.

Fixes #874.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -118,6 +118,10 @@ jobs:
           python setup.py sdist --formats=zip
           pip install dist/*
 
+      - name: Add GMT's bin to PATH (Linux/macOS)
+        run: echo ${GITHUB_WORKSPACE}/gmt-install-dir/bin >> $GITHUB_PATH
+        if: runner.os != 'Windows'
+
       # Run the tests
       - name: Test with pytest (Linux/macOS)
         run: make test PYTEST_EXTRA="-r P"


.github/workflows/ci_tests_dev.yaml
```diff
@@ -118,6 +118,10 @@ jobs:
           python setup.py sdist --formats=zip
           pip install dist/*
 
+      - name: Add GMT's bin to PATH (Linux/macOS)
+        run: echo ${GITHUB_WORKSPACE}/gmt-install-dir/bin >> $GITHUB_PATH
+        if: runner.os != 'Windows'
+
       # Run the tests
       - name: Test with pytest (Linux/macOS)
         run: make test PYTEST_EXTRA="-r P"
```
weiji14, line 127: Can we just add it inline here?

```suggestion
      # Run the tests
      - name: Test with pytest (Linux/macOS)
        run: |
          echo ${GITHUB_WORKSPACE}/gmt-install-dir/bin >> $GITHUB_PATH  # Add GMT's bin to PATH
          make test PYTEST_EXTRA="-r P"
```

seisman: I don't think it will work.

> Prepends a directory to the system PATH variable and makes it available to all subsequent actions in the current job; the currently running action cannot access the updated path variable.

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#adding-a-system-path

weiji14: Ah, that's annoying, nevermind then.

weiji14 (approved):

Issue #845, opened 2021-02-06
Release PyGMT v0.3.0


willschlitzer:
**Release**: [v0.3.0](https://github.com/GenericMappingTools/pygmt/milestone/3)
**Scheduled Date**: 2021/02/15
**Pull request due date**: 2021/02/13

**Priority PRs/issues to complete prior to release**
- [x] Wrap subplot (#822)
- [x] Wrap grd2cpt (#803)
- [x] Wrap grdinfo aliases (#799)
- [x] Move plotting functions to separate modules (#808)
- [x] Add affiliation and ORCID to AUTHORS.md (#857)
- [x] Fail to install the dev version following the install instructions (#856)

**Before release**:
- [x] Reserve a DOI on [Zenodo](https://zenodo.org) by clicking on "New Version" (@weiji14)
- [x] Update Changelog (#887) (@willschlitzer)
- [x] Add a new entry in `doc/_static/version_switch.js` for documentation switcher
- [x] Update citation information https://github.com/GenericMappingTools/pygmt#citing-pygmt

**Release**:
- [x] Go to [GitHub Release](https://github.com/GenericMappingTools/pygmt/releases) and make a release, this will automatically create a tag too (@willschlitzer)
- [x] Manually upload the pygmt-vX.Y.Z.zip file to https://zenodo.org/deposit, make sure you file it under the correct reserved DOI (@weiji14)

**After release**:
- [x] Update conda-forge [pygmt-feedstock](https://github.com/conda-forge/pygmt-feedstock) [Usually done automatically by conda-forge's bot] (https://github.com/conda-forge/pygmt-feedstock/pull/14)
- [x] Bump PyGMT version on https://github.com/GenericMappingTools/try-gmt (@seisman) (https://github.com/GenericMappingTools/try-gmt/commit/bd8090ce9fe07c6a083cfacfe5ae0e0a9a77b4a1)
- [x] Announce the release on:
  - [x] GMT [forum](https://forum.generic-mapping-tools.org/c/news/) (https://forum.generic-mapping-tools.org/t/pygmt-v0-3-0-released)
  - [x] GMT [website](https://github.com/GenericMappingTools/website) (News) (@michaelgrund) (https://github.com/GenericMappingTools/website/pull/95)
[...]

PR #900, opened 2021-02-15
Ensure proper non-dev version string when publishing to PyPI

### LINKED ISSUE #901: Issues of the "Publish to PyPI" workflow

The [`publish-to-pypi.yml`](https://github.com/GenericMappingTools/pygmt/blob/master/.github/workflows/publish-to-pypi.yml) workflow uploads the PyGMT packages to PyPI and/or TestPyPI.

During the release v0.3.0, we found an issue with the workflow and had a quick fix in #900. 

When I reiviwed the workflow, I found two more issues:

1. We're using secrets `test_pypi_password` and `pypi_password` in the `pypa/gh-action-pypi-publish` action. 
https://github.com/GenericMappingTools/pygmt/blob/65b099a5873b4f9e76d7fccdbc8882d9b37870e3/.github/workflows/publish-to-pypi.yml#L54-L64

The names of these two secrets are confusing. They are API tokens (generated by me, if I remember correctly), not someone's passwords. We should rename them to better names, `TEST_PYPI_API_TOKEN` and `PYPI_API_TOKEN` following the [official guide](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/).

2. The workflow always fails on forks, because forks don't have these secrets. We should skip the workflow for forks.

weiji14:

Ensures that the PyPI release Github Action workflow has a proper version string like `v0.3.0` instead of `v0.3.1.dev0`. Also edit the Publish-to-PyPI workflow to only run on GenericMappingTools/pygmt and rename secrets from `*_PYPI_PASSWORD` to `*_PYPI_API_TOKEN`

Publish Publish Github Action Documentation is at https://github.com/pypa/gh-action-pypi-publish/tree/v1.4.2

Addresses https://github.com/GenericMappingTools/pygmt/issues/845#issuecomment-779148771, also fixes #901.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -55,11 +55,11 @@ jobs:
     - name: Publish to Test PyPI
       uses: pypa/gh-action-pypi-publish@v1.4.2
       with:
-        password: ${{ secrets.test_pypi_password }}
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
         repository_url: https://test.pypi.org/legacy/
 
     - name: Publish to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@v1.4.2
       with:
-        password: ${{ secrets.pypi_password }}
+        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -36,6 +36,7 @@ jobs:
       run: python -m pip install setuptools wheel
 
     - name: Fix up version string
+      if: !startsWith(github.ref, 'refs/tags')
       run: |
         # Change setuptools-scm local_scheme to "no-local-version" so the
         # local part of the version isn't included, making the version string
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -19,6 +19,7 @@ jobs:
   publish-pypi:
     name: Publish to PyPI
     runs-on: ubuntu-latest
+    if: github.repository == 'GenericMappingTools/pygmt'
 
     steps:
     - name: Checkout
@@ -35,8 +36,8 @@ jobs:
     - name: Install dependencies
       run: python -m pip install setuptools wheel
 
+    # This step is only necessary for testing purposes and for TestPyPI
     - name: Fix up version string for TestPyPI
-      # This step is only necessary for testing purposes and for TestPyPI
       if: !startsWith(github.ref, 'refs/tags')
       run: |
         # Change setuptools-scm local_scheme to "no-local-version" so the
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       run: python -m pip install setuptools wheel
 
     - name: Fix up version string
+      if: !startsWith(github.ref, 'refs/tags')
```
seisman, line 39: Can you also move the comment "The step is only necessary for testing purpose" here, so we can better why we use `!startsWith(github.ref, 'refs/tags')`.

seisman:
Please see #901, and let me know if you want to fix it in this PR.

.github/workflows/publish-to-pypi.yml
weiji14: Ok, done at 82290d66716a097d12c135ad713f2f8f3f8057c9

weiji14:
> Please see #901, and let me know if you want to fix it in this PR.

Ok, I think the issues in #901 have been addressed, but please double check. **Note**: Will need to rename `TEST_PYPI_PASSWORD` -> `TEST_PYPI_API_TOKEN` and `PYPI_PASSWORD` -> `PYPI_API_TOKEN` before merging this PR.

.github/workflows/publish-to-pypi.yml
```diff
@@ -35,13 +36,13 @@ jobs:
     - name: Install dependencies
       run: python -m pip install setuptools wheel
 
-    - name: Fix up version string
+    # This step is only necessary for testing purposes and for TestPyPI
+    - name: Fix up version string for TestPyPI
+      if: ${{ !startsWith(github.ref, 'refs/tags') }}
```
seisman, line 41: If I understand it correctly, when we make a release/tag, this sed command won't run. So, we are pushing the same versions (e.g., `v0.3.1`) to both PyPI and TestPyPI, right?

weiji14: Yes, correct. But I think pushing v0.3.1 to both PyPI and TestPyPI is fine right?

seisman: Yes, it looks good.

seisman (approved):
Things to do before merging:

1. Rename the secrets in https://github.com/GenericMappingTools/pygmt/settings/secrets/actions
2. Update the PR title

weiji14:
> Things to do before merging:
> 
>     1. Rename the secrets in https://github.com/GenericMappingTools/pygmt/settings/secrets/actions

Ok, generated new tokens and created new secrets for PyPI and TestPyPI. Also deleted the old `TEST_PYPI_PASSWORD` and `PYPI_PASSWORD` secrets at https://github.com/GenericMappingTools/pygmt/settings/secrets/actions.

>     2. Update the PR title

See if it looks ok, having trouble describing this PR :sweat_smile:

seisman (approved):
Looks good.

weiji14:
I might trigger a quick test (to TestPyPI) just to make sure things work.

PR #921, opened 2021-02-17
Replace argument by parameter in image.py

michaelgrund:
Replaced  all appearances of _argument_ by _parameter_.

### DIFF
--- a/examples/gallery/plot/image.py
+++ b/examples/gallery/plot/image.py
@@ -4,7 +4,7 @@ Images or EPS files on maps
 The :meth:`pygmt.Figure.image` method can be used to read and
 place a raster image file or an Encapsulated PostScript file
 on a map. We must specify the file as *str* via the ``imagefile``
-parameter or simply use the filename as the first parameter. You can
+parameter, or simply use the filename as the first argument. You can
 also use a full URL pointing to your desired image. The ``position``
 parameter allows us to set a reference point on the map for the image.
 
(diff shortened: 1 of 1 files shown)


examples/gallery/plot/image.py
```diff
@@ -4,9 +4,9 @@
 The :meth:`pygmt.Figure.image` method can be used to read and
 place a raster image file or an Encapsulated PostScript file
 on a map. We must specify the file as *str* via the ``imagefile``
-argument or simply use the filename as the first argument. You can
+parameter or simply use the filename as the first parameter. You can
```
weiji14, line 7:
```suggestion
parameter, or simply use the filename as the first argument. You can
```
I feel like this sentence is repeating itself somehow. The str input is an argument, which is passed to the `imagefile` parameter. Alternatively, the filename (also a str) can be passed in as an argument to the function?!

michaelgrund: You're right, thanks!

seisman:
Just a quick comment: I don't quite like this example, because it's so similar to the "logo" example (https://www.pygmt.org/dev/gallery/plot/logo.html). Perhaps we can change to a better image?

michaelgrund:
> 
> 
> Just a quick comment: I don't quite like this example, because it's so similar to the "logo" example (https://www.pygmt.org/dev/gallery/plot/logo.html). Perhaps we can change to a better image?

Agree with you @seisman , any suggestions which image we could use instead? Originally I selected this image since it is availabe from the GMT website, so we would know if its origin changes. Alternatively, we could store an example image in the repo.

seisman:
Close this one as the changes has been merged into #943. We may still want to improve the example using a better figure.

PR #938, opened 2021-02-20
Update web font to Atkinson Hyperlegible

### LINKED ISSUE #933: Better web font for I (capital i) and l (small L)

See this page as an example: https://pygmt-6lvxfpcmx.vercel.app/gallery/line/vector-heads-tails.html

image

The two "I" in the two red boxes are different. The first one is capital `i`, and the second one is small `L`. However, they look the same in the current font (I'm using Microsoft Edge on macOS. Not sure what they look like on other web browsers and platforms).

Should we switch to a better web font? Does the current theme "sphinx_rtd_theme" support custom web fonts?

weiji14:

Use [Atkinson Hyperlegible Font](https://brailleinstitute.org/freefont). Preview at https://pygmt-git-docs-fonts-legible-gmt.vercel.app/gallery/line/vector-heads-tails.html#sphx-glr-gallery-line-vector-heads-tails-py. Sample:

Atkinson Hyperlegible font sample preview

Pros: Letters like small letter L has a little slant at the bottom that differentiates it from i,
Cons: The number 0 character has a slash in the middle which may (or may not) look ugly.

License appears to be modified from Open Font License, see diff at https://github.com/google/fonts/issues/2816#issuecomment-765465539

References:
- https://brailleinstitute.org/freefont
- https://brailleinstitute.org/wp-content/uploads/2020/02/BIA_AtkinsonHyerlegible-Specimen_200210.pdf

Fixes #933

**Reminders**

**Slash Commands**


liamtoney (approved):
Looks good to me.

doc/_static/style.css
```diff
@@ -131,3 +136,34 @@ a.copybtn {
     user-select: none;
 }
 
+/* Atkinson Hyperlegible regular */
+@font-face {
+    font-family: "Atkinson Hyperlegible";
+    src: url("./fonts/Atkinson-Hyperlegible-Regular-102a.woff2") format('woff2');
```
weiji14, line 142: Just a note on the Web Open Font Format 2 (woff2) format, it's smaller to serve than woff, but not supported on Intetnet Explorer (see https://caniuse.com/woff2). Can we check our Google Analytics added in #314 to see how many users this equates to? Just don't want to accidentally affect a significant proportion of our users.

seisman: https://en.wikipedia.org/wiki/Internet_Explorer#End_of_life

> Microsoft Teams no longer support IE11 as of November 30, 2020, and the remaining O365 applications will no longer support IE11 on August 17, 2021.

We don't need to worry about IE.

weiji14: Cool, and if there is anyone using IE, they will still have a 'pretty' default sans-serif font to look at.

PR #907, opened 2021-02-16
Update release process in template post v0.3.0

weiji14:
**Description of proposed changes**

Documenting some of the lessons learned from Release v0.3.0 (#845, #887) so that the next one will be smoother. The Release Drafter bot will also automatically increment to the next minor version (instead of a patch version) when a PR tagged 'feature' is merged into the master branch.

Addresses the documentation part of #885 and also closes #945.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #941, opened 2021-02-20
Write changelog in markdown using MyST

### LINKED ISSUE #885: Update the release drafter workflow and configurations
This project uses the [Release Drafter workflow](https://github.com/GenericMappingTools/pygmt/blob/master/.github/workflows/release-drafter.yml) to draft the [next release notes](https://github.com/GenericMappingTools/pygmt/releases) (only visible to maintainers) from PRs.

- [x] We're using the [`release-drafter/release-drafter@v5.8.0`](https://github.com/GenericMappingTools/pygmt/blob/805f533751cf3b0928c2523758984b3ba31c2161/.github/workflows/release-drafter.yml#L14), but the latest version is [`v5.13.0`](https://github.com/release-drafter/release-drafter/releases). We should bump it to the latest version (we have set up the dependbot to bump the actions automatically, but the bot runs on every Tuesday, so we won't see it happen before v0.3.0 release). [**Done** in #889]
 
- [ ] Some new features were added since v5.8.0. One feature that may be useful to us is https://github.com/release-drafter/release-drafter/pull/582, which means we can write the full URL of PRs in the release notes, instead of running the `sed` command to replace `#NUMBER` to `#NUMBER + URL`. 

- [ ] The release notes on GitHub are written in markdown, but the [changelog](https://github.com/GenericMappingTools/pygmt/blob/master/doc/changes.rst) are written in ReST. When we're going to make a release, we need to copy the draft release notes to `changes.rst`, convert it to ReST, create a PR and remove some items with low importance, merge the PR and then update the github release notes again. I found it very inconvenient because the two different markdown languages. I feel it would be much easier if we also use markdown for changelogs, i.e., `changes.md` instead of `changes.rst`.

- [ ] Changes are grouped into several categories: "Highlights", "New Features", "Enhancements", "Documentation", "Bug Fixes", "Maintenance",  and "Deprecations". I don't like the current order of these categories. IMHO, these categories should be sorted by importance from the readers' point of view. Thus, the order should be 
   1. "Highlights"
   2. "New Features"
   3. "Enhancements"
   4. "Deprecations"
   5. "Bug Fixes"
   6. "Documentation"
   7. "Maintenance"

weiji14:

[MyST (Markedly Structured Text)](https://myst-parser.readthedocs.io/en/v0.13.5/index.html) allows us to write Sphinx documentation entirely in markdown (technically using [Commonmark](https://commonmark.org/)). I.e. think ReST (Restructured Text) but in Markdown! MyST is a fairly young project (first published [PyPI release](https://pypi.org/project/myst-parser/#history) on 6 Mar 2020), but it looks pretty awesome already!

This MyST parser should address @seisman's comment at https://github.com/GenericMappingTools/pygmt/issues/885#issue-807847528 on the pain of having to convert the markdown changelog generated by [Release Drafter](https://github.com/release-drafter/release-drafter) into ReST syntax. In other words, we should be able to just copy and paste things directly from https://github.com/GenericMappingTools/pygmt/releases, all in markdown!

Documentation preview is at https://pygmt-git-myst-changelog.gmt.vercel.app/changes.html. This should look pretty much the same as https://www.pygmt.org/v0.3.0/changes.html (point out any inconsistencies if you spot any).

TODO:
- [x] Install [myst-parser](https://github.com/executablebooks/myst-parser) following https://myst-parser.readthedocs.io/en/v0.13.5/using/intro.html
- [x] Use h2 title (##) for each Release vX.Y.Z header title
- [x] Use h3 title (###) for sub-titles under each Release (i.e. New Changes, Bug Fixes, etc)
- [x] Sort order of "Highlights", "New Features", "Enhancements", "Documentation", "Bug Fixes", "Maintenance", and "Deprecations" subsections (see https://github.com/GenericMappingTools/pygmt/issues/885#issue-807847528)

Fixes #885, companion PR to #907.

**Reminders**

**Slash Commands**


.github/release-drafter.yml
```diff
@@ -17,7 +17,9 @@ exclude-labels:
   - 'skip-changelog'
 change-template: '* $TITLE (#$NUMBER)'
```
weiji14, line 18:
```suggestion
change-template: '* $TITLE [#$NUMBER]($URL)'
```

This change should format the PRs into `[#123](https://github.com/GenericMappingTools/pygmt/pull/123)`, and so `sed` shouldn't be needed anymore!

.github/release-drafter.yml
```diff
@@ -15,16 +15,19 @@ categories:
     label: 'deprecation'
 exclude-labels:
   - 'skip-changelog'
+category-template: '### $TITLE'
```
weiji14, line 18: Need to find a way to make the subheaders (e.g. for v0.2.1 enhancements) a proper permalink. I.e. something like https://pygmt-git-myst-changelog.gmt.vercel.app/changes.html#release-v0-2-1-enhancements instead of https://pygmt-git-myst-changelog.gmt.vercel.app/changes.html#id3

seisman:
Why not using recommonmark (https://www.sphinx-doc.org/en/master/usage/markdown.html)?

weiji14:
> Why not using recommonmark (https://www.sphinx-doc.org/en/master/usage/markdown.html)?

The markdown-it parser used in MyST is about two times faster than the commonmark_py parser used in recommonmark (see https://markdown-it-py.readthedocs.io/en/latest/other.html#performance). Also because I didn't know about Recommonmark, though MyST does seem better documented and actively maintained?

seisman:
> > Why not using recommonmark ([sphinx-doc.org/en/master/usage/markdown.html](https://www.sphinx-doc.org/en/master/usage/markdown.html))?
> 
> The markdown-it parser used in MyST is about two times faster than the commonmark_py parser used in recommonmark (see [markdown-it-py.readthedocs.io/en/latest/other.html#performance](https://markdown-it-py.readthedocs.io/en/latest/other.html#performance)). Also because I didn't know about Recommonmark, though MyST does seem better documented and actively maintained?
> 
> See also [stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rst](https://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rst)

Fair enough.

.github/release-drafter.yml
seisman: I think the `$URL` is pretty useful. The only con is that it makes lines much longer than before, so it would be a little difficult when we try to merge some changelog entries.

weiji14: We might be able to shorten things using Jinja2 substitutions in MyST (using `{{ }}`), though I haven't tested it yet.

References:
- https://myst-parser.readthedocs.io/en/v0.13.5/using/syntax-optional.html#substitutions-and-urls
- https://myst-parser.readthedocs.io/en/v0.13.5/develop/_changelog.html#new

seisman: I think it won't work well on github release page.

weiji14: True, we do want to just copy and paste things across, ideally without changing anything.

seisman: Not a big issue for me, but I'm wondering if we can change each entry to two lines
```
change-template: > 
    * $TITLE 
      ([#$NUMBER]($URL))
```
so it will be:
```
* Improve how PyGMT finds the GMT library 
  ([#702](https://github.com/GenericMappingTools/pygmt/pull/702))
```

seisman: Tried it in my own fork. It doesn't work:

image

weiji14: What if you add a backslash?

seisman: No changes at all.

weiji14: Nevermind then :sweat_smile: Should I commit this change still (i.e. let release drafter do `[#$NUMBER]($URL)` directly) or stick with the current state using `sed` (i.e. so we can review using short `#123` PR numbers and then run `sed` after)?

seisman: If we use short PR numbers (`#123`), we need to strictly follow the following steps:

1. Review and finalize all changelog entries
2. Make a backup of the current changelog (git history already does this)
3. Run `sed` to convert the PR links 
4. Merge the changelog into master
5. Find the backup at step 2 and update the github release notes

In this case, using ReST or Markdown makes no difference to us.

If we use full PR numbers and links, 

1. Review and finalize all changelog entries
2. Merge the changelog into master
3. Copy the changelog from `changes.md` and update the github release notes

The new way (full PR numbers and links) seems easier and also doesn't rely on external tools (`sed`).

weiji14: Ok, go with the new way then (`[#$NUMBER]($URL)`). Great that @willschlitzer was clever enough to do a backup of the changelog for the PyGMT v0.3.0 release, but let's not count on our luck so much!

P.S. I'm drafting the new simplified Release Note curation/editing process in MAINTENANCE.md at https://github.com/GenericMappingTools/pygmt/pull/907/files#diff-87a0c4623bffa64c60b62a39eeab1c40bd0f329cebc177f5e4492555fb5d42ae.

.github/release-drafter.yml
```diff
@@ -15,16 +15,19 @@ categories:
...
   - 'skip-changelog'
-change-template: '* $TITLE (#$NUMBER)'
+category-template: '### $TITLE'
+change-template: '* $TITLE [#$NUMBER]($URL)'
 template: |
-  [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3781524.svg)](https://doi.org/10.5281/zenodo.3781524)
+  ## Release v$NEXT_PATCH_VERSION (20YY/MM/DD)
 
-  ## Highlights
+  [![Digital Object Identifier for PyGMT v$NEXT_PATCH_VERSION](https://zenodo.org/badge/DOI/10.5281/zenodo.3781524.svg)](https://doi.org/10.5281/zenodo.3781524)
```
seisman, line 23: I think `DOI` is well known.
```suggestion
  [![DOI for PyGMT v$NEXT_PATCH_VERSION](https://zenodo.org/badge/DOI/10.5281/zenodo.3781524.svg)](https://doi.org/10.5281/zenodo.3781524)
```

.github/release-drafter.yml
```diff
@@ -15,16 +15,19 @@ categories:
     label: 'deprecation'
 exclude-labels:
   - 'skip-changelog'
-change-template: '* $TITLE (#$NUMBER)'
+category-template: '### $TITLE'
+change-template: '* $TITLE [#$NUMBER]($URL)'
```
seisman, line 19:
```suggestion
change-template: '* $TITLE ([#$NUMBER]($URL))'
```

.github/release-drafter.yml
weiji14: Prefer to keep this alt-text in long form for accessibility (e.g. those using screen readers). Also, the previous ReST alt-text used "Digital Object Identifier" too.

seisman: OK with the long-form.

.github/release-drafter.yml
seisman: It seems commonmark still doesn't support anchors for headers, so maybe we can do nothing here.

weiji14: Maybe do it in sphinx instead? I see https://stackoverflow.com/questions/15394347/adding-a-cross-reference-to-a-subheading-or-anchor-in-another-page#comment79146992_40225340 pointing to https://www.sphinx-doc.org/en/master/usage/extensions/autosectionlabel.html#confval-autosectionlabel_prefix_document.

seisman: > The autosectionlabel_prefix_document configuration variable can be used to make headings which appear **multiple times but in different documents** unique.

> Useful for avoiding ambiguity when the same section heading appears in different documents.

I don't think it works for us.

.github/release-drafter.yml
```diff
@@ -5,26 +5,29 @@ categories:
...
+  - title: 'Deprecations'
+    label: 'deprecation'
   - title: 'Bug Fixes'
     label: 'bug'
+  - title: 'Documentation'
+    label: 'documentation'
   - title: 'Maintenance'
     label: 'maintenance'
-  - title: 'Deprecations'
-    label: 'deprecation'
```
weiji14, line 15: Categories have been reordered following https://github.com/GenericMappingTools/pygmt/issues/885#issue-807847528. Do we want to reorder things in the previous release changelogs (i.e. v0.1.0 to v0.3.0) or just keep them as they were?

weiji14 (commented):
> > The autosectionlabel_prefix_document configuration variable can be used to make headings which appear **multiple times but in different documents** unique.
> 
> > Useful for avoiding ambiguity when the same section heading appears in different documents.
> 
> I don't think it works for us.

Ok, this PR is ready for final review @GenericMappingTools/python-maintainers. Sorry that the `changes.rst` to `changes.md` rename doesn't have a nice diff, but you can preview things at https://pygmt-git-myst-changelog.gmt.vercel.app/changes.html. The permalink/anchor issue mentioned at https://github.com/GenericMappingTools/pygmt/pull/941#discussion_r579727065 is fairly minor so I'll just let it be.

.github/release-drafter.yml
seisman: > Categories have been reordered following [#885 (comment)](https://github.com/GenericMappingTools/pygmt/issues/885#issue-807847528). 

Looks good to me.

> Do we want to reorder things in the previous release changelogs (i.e. v0.1.0 to v0.3.0) or just keep them as they were?

No need to reorder the old entries.

.github/release-drafter.yml
seisman: @weiji14 As we have no solution for the custom anchors, do you think we should not use h3 headers, but just use **bold** headings, i.e., use `**Documentation**`, not `### Documentation`.

weiji14: Would stick with h3 headers. Even without the permalink anchors, it'll still be easier for users to navigate quickly to the subsection on the sidebar:

Changelog subsections in navbar

And people could still point to the v0.3.0 highlights using https://www.pygmt.org/v0.3.0/changes.html#highlights, and so on for v0.4.0 and beyond.

seisman:
It seems that recent release-drafter releases have more features, for example [`version-resolver`](https://github.com/release-drafter/release-drafter#version-resolver) probably can determine if the next version is v0.3.1 or v0.4.0.

seisman:
> It seems that recent release-drafter releases have more features, for example [`version-resolver`](https://github.com/release-drafter/release-drafter#version-resolver) probably can determine if the next version is v0.3.1 or v0.4.0.

If I understand it correctly, the following setting means that if any PR labeled with "feature" or "enhancement" is merged into master branch, the next version will be a minor release (i.e., v0.4.0), otherwise it's a patch release (i.e., v0.3.1). However, sometimes we merge PRs with "enhancement" but may don't want to bump the minor version (e.g., #952).
```
version-resolver:
  minor:
    labels:
      - 'feature'
      - 'enhancement'
  default: patch
```

weiji14:
> It seems that recent release-drafter releases have more features, for example [`version-resolver`](https://github.com/release-drafter/release-drafter#version-resolver) probably can determine if the next version is v0.3.1 or v0.4.0.

Oo, that does look fancy! But let's leave that to a separate Pull Request as we'll need to document that '$RESOLVED_VERSION` feature in MAINTENANCE.md too. Or I can include it in #907.

seisman (approved):
This PR is good to merge. We can change any settings in future PRs.

seisman (approved):

PR #952, opened 2021-02-22
Open figures using the associated application on Windows

seisman:

This PR improves the `launch_external_viewer` function to open images
using the associated application (just like double-click the images) on
Windows.

The implementation is done by calling the
[`os.startfile`](docs.python.org/3/library/os.html#os.startfile) function.

`os.startfile` is only available on Windows, so need to disable the pylint
error `no-member`.

The code was originally written in #528 by @leouieda, re-used in #529, 
but I'd like to open this PR separately so that changes in #529 can be 
smaller.

BTW, the function is not tested at all because we can't open GUI programs 
in CI machines.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -1,6 +1,7 @@
 """
 Utilities and common tasks for wrapping the GMT modules.
 """
+import os
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


weiji14 (approved):
Probably should get a Windows user to test this (or login to Windows myself), but I think [`os.startfile`](https://docs.python.org/3.8/library/os.html#os.startfile) should work.

seisman:
Do we have any Windows developers who can try it? @GenericMappingTools/python @GenericMappingTools/python-maintainers

Issue #593, opened 2020-09-09
Better return values for grdinfo


weiji14:
**Description of the desired feature**

Certain [`pygmt.grdinfo`](https://www.pygmt.org/v0.1.2/api/generated/pygmt.grdinfo.html) return values can be in an array form rather than a string. This would be a good first issue to take up, and would be similar to #575 but for grids instead of tables.

_Originally posted by @liamtoney in https://github.com/GenericMappingTools/pygmt/issues/147#issuecomment-450018956_

> **tl;dr**
> 
> After reading the above discussion, I would like to work on more fully wrapping `grdinfo`. It seems like a potentially good "first issue," and I can see a benefit in wrapping it even with Python containers also available. See below, for example.
> 
> **Motivation**
> 
> In my use case, I want to easily **see** the min/max z-values of a DEM so that I can choose my colorbar limits.
> 
> **1. No specified region**
> 
> With `GRIDFILE` being the full path to my grid file, I could use
> ```python
> import gmt
> print(gmt.grdinfo(GRIDFILE).strip())
> ```
> ```
> ...
> ../srtm15_plus_w180n90.nc: z_min: -7840.125 z_max: 6057.64355469 name: z
> ...
> ```
> or (with `xarray`, for example)
> ```python
> import xarray
> print(xarray.open_dataset(GRIDFILE).z.actual_range)
> ```
> ```
> [-7840.125       6057.64355469]
> ```
> These are about equal effort (and `xarray` returns things in a more usable output). 
> 
> **2. Specified region of interest**
> 
> For viewing the min/maz z-values within a specified `REGION = (lonmin, lonmax, latmin, latmax)`, I think that `grdinfo` makes this easier:
> ```python
> import gmt
> print(gmt.grdinfo(GRIDFILE, R='{}/{}/{}/{}'.format(*REGION)).strip())
> ```
> ```
> ...
> ../srtm15_plus_w180n90.nc: z_min: -4996.11816406 z_max: 6057.64355469 name: z
> ...
> ```
> vs.
> ```python
> import xarray
> z = xarray.open_dataset(GRIDFILE).sel(lon=slice(*REGION[0:2]), lat=slice(*REGION[2:4])).z.values
> print(z.min(), z.max())
> ```
> ```
> -4996.118 6057.6436
> ```
> Plus `grdinfo` is more format agnostic. By minimally including the `@use_alias` and `@kwargs_to_strings` decorators, the `grdinfo` command above could be just
> ```python
[...]

PR #960, opened 2021-02-24
Let pygmt.info load datetime columns into a str dtype array

weiji14:

Fixes problem with pygmt.info not being able to handle datetime64 inputs. I.e. `ValueError: could not convert string to float: '2021-01-01T12:34:56'`.

However, users will still need to use`pygmt.info(..., f="0T")` until upstream issue at https://github.com/GenericMappingTools/gmt/issues/4241 is resolved. Something like so:

```python
import pandas as pd
import pygmt

table = pd.date_range(start="2020-01-01", periods=5).to_numpy()
output = pygmt.info(table=table, per_column=True, f="0T")
print(output)
# ['2020-01-01T00:00:00' '2020-01-05T00:00:00']
```

Also added two extra unit tests using numpy datetime64 types.

References:
- https://stackoverflow.com/questions/31553983/python-using-numpy-loadtxt-reading-a-csv-file-with-different-data-types-for-ea/31554777#31554777

Addresses #597 on the pygmt side using fix mentioned in https://github.com/GenericMappingTools/pygmt/issues/597#issuecomment-725091467. Assuming that https://github.com/GenericMappingTools/gmt/issues/4241 is fixed in GMT 6.2.0, then issue #597 can be closed once PyGMT bumps minimum required GMT version to 6.2.0.

**Reminders**

**Slash Commands**


pygmt/tests/test_info.py
```diff
@@ -41,6 +41,18 @@ def test_info_dataframe():
     assert output == expected_output
 
 
+def test_info_numpy_array_time_column():
+    """
+    Make sure info works on a numpy.ndarray input with a datetime type.
+    """
+    table = pd.date_range(start="2020-01-01", periods=5).to_numpy()
+    output = info(table=table, f="0T")
```
weiji14, line 49: The `f="0T"` argument can be removed once upstream GMT fixes https://github.com/GenericMappingTools/gmt/issues/4241.

pygmt/tests/test_info.py
```diff
@@ -115,6 +127,17 @@ def test_info_per_column():
     )
 
 
+def test_info_per_column_with_time_inputs():
+    """
+    Make sure the per_column option works with time inputs.
+    """
+    table = pd.date_range(start="2020-01-01", periods=5).to_numpy()
+    output = info(table=table, per_column=True, f="0T")
```
weiji14, line 135: Actually, do we want a proper long alias for [`f`](https://docs.generic-mapping-tools.org/6.1/gmt.html#f-full) here?

pygmt/src/info.py
```diff
@@ -94,6 +94,10 @@ def info(table, **kwargs):
             # instead of a raw string that is less useful.
             if result.startswith(("-R", "-T")):  # e.g. -R0/1/2/3 or -T0/9/1
                 result = result[2:].replace("/", " ")
-            result = np.loadtxt(result.splitlines())
+            try:
+                result = np.loadtxt(result.splitlines())
+            except ValueError:
+                # Load non-numerical outputs in str type, e.g. for datetime
+                result = np.loadtxt(result.splitlines(), dtype="str")
```
seisman, line 101: It means all the output columns will be in str type, including numerical columns. I'm not sure if I like it.

weiji14: Yes, but there's no way to have mixed types in numpy. An array with str (e.g. datetime) and numerical (e.g. int/float) types will have to be made into a `str` or `object` type.

pygmt/tests/test_info.py
seisman: Yes.

pygmt/src/info.py
seisman: I'm OK with a string-type array. We can think more about the return values later (https://github.com/GenericMappingTools/pygmt/issues/593).

pygmt/tests/test_info.py
weiji14: Ok, will do in a separate PR.

weiji14:
Hmm, that crash on the Windows GMT dev test at https://github.com/GenericMappingTools/pygmt/runs/2025836161?check_suite_focus=true#step:15:86 at the start (before any tests got run) doesn't look nice.

seisman:
/test-gmt-dev

> Hmm, that crash on the Windows GMT dev test at [GenericMappingTools/pygmt/runs/2025836161?check_suite_focus=true#step:15:86](https://github.com/GenericMappingTools/pygmt/runs/2025836161?check_suite_focus=true#step:15:86) at the start (before any tests got run) doesn't look nice.

It may be just a sporadic crash. Just re-triggered the dev tests.

weiji14:
Ok, Windows tests have passed at https://github.com/GenericMappingTools/pygmt/runs/2026019634?check_suite_focus=true#step:15:867. Any other comment on this PR?

seisman (approved):

Issue #949, opened 2021-02-21
Universal virtualfile_from_data function to replace virtualfile_from_grid/matrix/vectors


weiji14:
**Description of the desired feature**

As a user, I'd like to be able to pass in any type of data into a PyGMT function in a consistent way, be it:

- a `str` type filename
- a `numpy.ndarray`
- a `pandas.DataFrame` matrix
- a `geopandas.GeoDataFrame` (#608)
- an `xarray.DataArray` grid, or even 1d arrays!
- etc

### Current state

We have a not so nice `if... then` block that currently handles the Input/Output (I/O) of data into PyGMT. This results in an incosistent API. E.g. [`fig.plot`](https://www.pygmt.org/v0.3.0/api/generated/pygmt.Figure.plot.html) requires `numpy.ndarray` inputs while [`pygmt.grdtrack`](https://www.pygmt.org/v0.3.0/api/generated/pygmt.grdtrack.html) requires `pandas.DataFrame` inputs. Users (e.g. at https://forum.generic-mapping-tools.org/t/pygmt-plot-errorbars) should not have to figure out which [PyData](https://pydata.org/) format is needed.

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/issues/946#issuecomment-782925067_

> I was going to create a `virtualfile_from_data()` function to replace this common block of code in our functions:

> ```python
>     with Session() as lib:
>         # Choose how data will be passed in to the module
>         if kind == "file":
>             file_context = dummy_context(data)
>         elif kind == "matrix":
>             file_context = lib.virtualfile_from_matrix(data)
>         elif kind == "vectors":
>             file_context = lib.virtualfile_from_vectors(...)
> ```

### Future state

The idea is to have a `virtualfile_from_data()` function (open to other names) that wraps around the current `dummy_context`/`virtualfile_from_matrix`/`virtualfile_from_vectors`/`virtualfile_from_grid` functions, which makes things easier for both end users and developers wrapping new modules that handle data.

This universal `virtualfile_from_data()` may pave the way to support more data inputs in the future:

- `geopandas.GeoDataFrame` inputs, useful for `fig.plot` #608 
[...]

PR #961, opened 2021-02-24
Refactor info and grdinfo to use virtualfile_from_data

weiji14:

Create a universal `virtualfile_from_data` function that can handle any raster or vector data input. This allows us to centralize the data validation logic in a single place, resulting in a cleaner API for PyGMT modules to handle different PyData types (e.g. numpy/pandas/xarray/etc) seamlessly. This is the beginning of resolving #949.

Preview API docs at https://pygmt-git-virtualfilefromdata-gmt.vercel.app/api/generated/pygmt.clib.Session.virtualfile_from_data.html

### Before:

```python
kind = data_kind(grid, None, None)
with Session() as lib:
    if kind == "file":
        file_context = dummy_context(grid)
    elif kind == "grid":
        file_context = lib.virtualfile_from_grid(grid)
    else:
        raise GMTInvalidInput("Unrecognized data type: {}".format(type(grid)))
    with file_context as infile:
        ...
```

### After

```python
with Session() as lib:
```

As a start, both `info` and `grdinfo` have been refactored to use this new convenience function. Other modules will be refactored in separate PRs later.

Addresses #949.

**Reminders**

**Slash Commands**


pygmt/clib/session.py
```diff
@@ -1359,6 +1360,62 @@ def virtualfile_from_grid(self, grid):
         with self.open_virtual_file(*args) as vfile:
             yield vfile
 
+    def virtualfile_from_data(self, data, x=None, y=None, z=None, check_kind=None):
```
seisman, line 1363: A quick question: does it work for modules that requires more than three columns? For example, in `Figure.plot()`, we need to pass `extra_arrays` to GMT API.

weiji14: Excellent point. Perhaps put the `check_kind` parameter in front, so that it becomes `virtualfile_from_data(self, check_kind, data, x, y, z, *extra_arrays)`?

seisman: I'm thinking if having two functions can make it easier:

- virtualfile_from_table_data: `data` can be `file` or `matrix`, or `x, y, z, extra_arrays` can be `vector`.
- virtualfile_from_grid_data: `data` can be `file`, or `xr.DataArray`

pygmt/clib/session.py
```diff
@@ -1359,6 +1360,62 @@ def virtualfile_from_grid(self, grid):
...
+
+        This convenience function automatically detects the kind of data passed
+        into it, and produces a virtualfile that can be passed into GMT later
+        on.
+
+        Parameters
+        ----------
+        data : str, xarray.DataArray, 2d array, or None
+            The vectors that will be included in the array. All must be of the
+            same size.
```
seisman, line 1375: The description of the `data` parameter is very confusing.

pygmt/clib/session.py
```diff
@@ -1359,6 +1360,62 @@ def virtualfile_from_grid(self, grid):
...
+            x, y and z columns as numpy arrays.
+        check_kind : str
+            Used to validate the type of data that can be passed in. Choose
+            from 'raster', 'vector' or None. Default is None (no validation).
+
+        Returns
+        -------
+        file_context : contextlib._GeneratorContextManager
+            The virtual file stored inside a context manager. Access the file
+            name of this virtualfile using `with file_context as fname: ...`.
```
seisman, line 1386:
```suggestion
            name of this virtualfile using ``with file_context as fname: ...``.
```

pygmt/clib/session.py
seisman: I'm OK with one universal function, but you may want to look at some more complicated modules (e.g., `grdtrack` and `plot`). In these modules, it seems the data `kind` is needed outside the `virtualfile_from_data` function.

weiji14: Yeah, `grdtrack` needs a good refactor, and so does `plot`. I think we can refactor this `virtualfile_from_data` function when we refactor those complicated modules, so need to make sure the implementation here is ok to be extended to those functions.

pygmt/clib/session.py
weiji14: I've changed the docstring, see if it looks ok.

pygmt/clib/session.py
seisman: OK to me.

seisman (approved):

seisman:
There are maybe two lines not hit by the tests (see https://codecov.io/gh/GenericMappingTools/pygmt/src/7c60cb304950aacb0ed0e1b3c76fac24a242b65d/pygmt/clib/session.py).

weiji14:
> There are maybe two lines not hit by the tests (see https://codecov.io/gh/GenericMappingTools/pygmt/src/7c60cb304950aacb0ed0e1b3c76fac24a242b65d/pygmt/clib/session.py).

Yes I'm aware of them. They will be covered when I refactor other functions (e.g. `plot`, `grdtrack`) to use `virtualfile_from_data`, so just a temporary thing.

seisman (approved):
OK to merge.

weiji14:
Cool, just want to make sure the full tests pass and I'll merge after that.

Issue #904, opened 2021-02-15
Unnecessary files in the source distribution?


seisman:
Running `python setup.py sdist` will build a PyGMT source distribution in the `dist` directory (e.g., `dist/pygmt-0.3.0.tar.gz`.

The tarball contains the PyGMT source codes. Below is the list of top-level files and directories:
```
.codecov.yml
.github
.gitignore
.pylintrc
AUTHORS.md
AUTHORSHIP.md
CODE_OF_CONDUCT.md
CONTRIBUTING.md
LICENSE.txt
MAINTENANCE.md
MANIFEST.in
Makefile
PKG-INFO
README.rst
doc
environment.yml
examples
package.json
pygmt
pygmt.egg-info
pyproject.toml
requirements-dev.txt
requirements.txt
setup.cfg
setup.py
vercel.json
```

I'm not familiar with Python source distributions, but some files seem unnecessary (e.g., `vercel.json`, `package.json`, `.github`). 

Reading the [Python package guide](https://packaging.python.org/guides/using-manifest-in/), I thought most plaintext files won't be added into the source distribution, unless explicitly specified in the `MANIFEST.in` file. Perhaps my understanding is wrong.

weiji14:
Yeah we should fix this by the next release (v0.3.1). Not that we'll reach it anytime soon, but PyPI has a 10GB total project size limit, and a 100 MB project upload limit (i.e. per release/version).

![image](https://user-images.githubusercontent.com/23487320/108116247-259e4480-7100-11eb-88de-14e61ed96176.png)

seisman:
So setuptools_scm includes all files in the source distribution by default.

See https://github.com/pypa/setuptools_scm#file-finders-hook-makes-most-of-manifestin-unnecessary and https://github.com/pypa/setuptools_scm/issues/516.

I think we need to explicitly exclude unnecessary files in the `MANIFEST.in` file.

PR #982, opened 2021-03-01
Add 'make package' target for building source and binary distributions

seisman:

Add the `make package` target for building packages.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -47,7 +47,7 @@ jobs:
 
     - name: Build source and wheel distributions
       run: |
-        python setup.py sdist bdist_wheel
+        make package
         echo ""
         echo "Generated files:"
         ls -lh dist/
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ help:
 	@echo "Commands:"
 	@echo ""
 	@echo "  install   install in editable mode"
+	@echo "  package   build source and wheel distributions"
 	@echo "  test      run the test suite (including doctests) and report coverage"
 	@echo "  format    run black, blackdoc, docformatter and isort to automatically format the code"
 	@echo "  check     run code style and quality checks (black, blackdoc, docformatter, flake8 and isort)"
@@ -25,6 +26,9 @@ help:
 
 install:
 	pip install --no-deps -e .
+
+package:
+	python setup.py sdist bdist_wheel
 
 test:
 	# Run a tmp folder to make sure the tests are run on the installed version


.github/workflows/publish-to-pypi.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
     - name: Build source and wheel distributions
       run: |
-        python setup.py sdist bdist_wheel
+        make package
```
weiji14, line 50: I almost want to keep this command as a pure python command, that way people without knowledge of `make` (e.g. Windows users) can understand what is happening as well.

Makefile
```diff
@@ -26,6 +27,9 @@ help:
 install:
 	pip install --no-deps -e .
 
+package:
+	python setup.py sdist bdist_wheel
```
weiji14, line 31: Just wondering, when do we want to use this `make package` command? And what is the different between `python setup.py sdist bdist_wheel` and `python setup.py sdist --formats=zip` used in 'ci_tests.yaml'?

.github/workflows/publish-to-pypi.yml
seisman: We're heavily using `make` commands in our workflows. The reason for replacing `python setup.py` with `make` is to keep it simple when we switch to other tools for building packages.

Makefile
seisman: > Just wondering, when do we want to use this `make package` command? 

I was trying to address issue #904, so had to run the commands many times. `make pacakge` is easier to type.

> And what is the different between `python setup.py sdist bdist_wheel` and `python setup.py sdist --formats=zip` used in 'ci_tests.yaml'?

`sdist` is the source distribution, and `bdist_wheel` is the binary distribution (in wheel format). They are uploaded to PyPI (https://pypi.org/project/pygmt/#files). 

In `ci_tests.yaml`, we need to install PyGMT, so we only need the source distribution (`sdist`).

weiji14: > > Just wondering, when do we want to use this `make package` command?
> 
> I was trying to address issue #904, so had to run the commands many times. `make pacakge` is easier to type.

Ok, fair enough. I would just use the [CTRL+R reverse-i-search](https://www.chriswrites.com/search-terminal-history/) command to save on typing an old command but `make package` works too.

> > And what is the different between `python setup.py sdist bdist_wheel` and `python setup.py sdist --formats=zip` used in 'ci_tests.yaml'?
> 
> `sdist` is the source distribution, and `bdist_wheel` is the binary distribution (in wheel format). They are uploaded to PyPI (https://pypi.org/project/pygmt/#files).
> 
> In `ci_tests.yaml`, we need to install PyGMT, so we only need the source distribution (`sdist`).

Got it, thanks for the explanation.

.github/workflows/publish-to-pypi.yml
weiji14: Let's not jinx things, I'm hoping Github Actions will last for a couple more years at least! Or maybe you mean switching to use `poetry` or something? EIther way, ok with using `make` here even though it's biased towards Linux/macOS.

weiji14 (approved):

.github/workflows/publish-to-pypi.yml
seisman: I meant switching to `build` (https://github.com/pypa/build) or poetry.

PR #976, opened 2021-02-27
Allow load_earth_relief() to load the original land-only 01s or 03s SRTM tiles

core-man:
**Description of proposed changes**

Allow the [load_earth_relief](https://github.com/GenericMappingTools/pygmt/blob/6a6171e3efc112f2a8a86088e24caf932149e928/pygmt/datasets/earth_relief.py#L14) function to load the original land-only SRTM tiles, which have a resolution of  `01s` or `03s`.

A new parameter `use_srtm` (default to `False`) was added to perform this feature.

If `resolution` is  `01s` or `03s`, we then check `use_srtm` to determine the relief data prefix (`earth_relief_prefix`).
- If `use_srtm` is `False`, `earth_relief_prefix` = `earth_relief_`.
- If `use_srtm` is `True`, `earth_relief_prefix` = `srtm_relief_`. 

Fixes #966

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #985, opened 2021-03-01
Cache the 03s earth relief tile @N35E135.earth_relief_03s_g.nc

core-man:

Originally posted in https://github.com/GenericMappingTools/pygmt/pull/976#discussion_r584631722.

33.5MB https://github.com/GenericMappingTools/pygmt/suites/2150973318/artifacts/44049313

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -33,7 +33,8 @@ jobs:
           gmt which -Ga @earth_relief_10m_p @earth_relief_10m_g \
                         @earth_relief_30m_p @earth_relief_30m_g \
                         @earth_relief_01d_p @earth_relief_01d_g \
-                        @earth_relief_05m_p @earth_relief_05m_g
+                        @earth_relief_05m_p @earth_relief_05m_g \
+                        @N35E135.earth_relief_03s_g.nc
           gmt which -Ga @ridge.txt @Table_5_11.txt @test.dat.nc \
                         @tut_bathy.nc @tut_quakes.ngdc @tut_ship.xyz \
                         @usgs_quakes_22.txt
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -3,7 +3,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  # pull_request:
+  pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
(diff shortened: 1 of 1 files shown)


.github/workflows/cache_data.yaml
```diff
@@ -3,7 +3,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  # pull_request:
+  pull_request:
```
core-man, line 6:
```suggestion
  # pull_request:
```

.github/workflows/cache_data.yaml
```diff
@@ -33,7 +33,8 @@ jobs:
           gmt which -Ga @earth_relief_10m_p @earth_relief_10m_g \
                         @earth_relief_30m_p @earth_relief_30m_g \
                         @earth_relief_01d_p @earth_relief_01d_g \
-                        @earth_relief_05m_p @earth_relief_05m_g
+                        @earth_relief_05m_p @earth_relief_05m_g \
+                        @N35E135.earth_relief_03s_g.nc
```
seisman, line 37: :confused: I even didn't know that we can use `gmt which -Ga @N35E135.earth_relief_03s_g.nc`  to download a single tile.

seisman:
@PaulWessel It seems we can use the following command to download a tile of 03s earth relief grid:
```
gmt which -Ga @N35E135.earth_relief_03s_g.nc
```
I never saw it in the GMT documentation. Is it an undocumented feature? Is it the recommended way to download one tile? 

I don't understand why it works, since the tiles are stored in jp2 format on the GMT server.

PaulWessel:
Certainly undocumented since we don't really want to tell people to get specific tiles, except indirectly via a -R setting.  But it works as it does because the filename matches a known tile and off it goes.  It knows what to do with these (jp2->nc).

I think we should keep quiet about this because we may decide to change our tiling structure etc.  It should not be something users should be aware of.

.github/workflows/cache_data.yaml
seisman, line 37: It's OK using the internal name, but at least we should add a comment to better explain it.
```suggestion
                        @earth_relief_05m_p @earth_relief_05m_g
          # Download one tile of the 03s srtm data.
          # @N35E135.earth_relief_03s_g.nc is for internal use only.
          # The naming scheme may change. 
          # DO NOT USE IT IN SCRIPTS.         
          gmt which -Ga @N35E135.earth_relief_03s_g.nc
```

seisman (approved):

Issue #608, opened 2020-09-17
Integration with geopandas to plot shapely geometries


weiji14:
**Description of the desired feature**

[Geopandas](https://github.com/geopandas/geopandas/) is somewhat of a defacto geographic data structure in Python nowadays, and is used to hold vector shapes like Polygons, Lines and Points. What might be useful is if we can have PyGMT plot [`geopandas` Data Structures](https://geopandas.readthedocs.io/en/latest/docs/user_guide/data_structures.html#data-structures) (e.g. [`geopandas.GeoDataFrame`](https://geopandas.readthedocs.io/en/latest/docs/user_guide/data_structures.html#geodataframe)s) directly. Something like so:

```python
import geopandas as gpd
import pygmt

cities: gpd.GeoDataFrame = gpd.read_file(gpd.datasets.get_path("naturalearth_cities"))
cities.head()
# 	name	geometry
# 0	Vatican City	POINT (12.45339 41.90328)
# 1	San Marino	POINT (12.44177 43.93610)
# 2	Vaduz	POINT (9.51667 47.13372)
# 3	Luxembourg	POINT (6.13000 49.61166)
# 4	Palikir	POINT (158.14997 6.91664)

fig = pygmt.Figure()
fig.plot(data=cities)
fig.show()
```

One way would be to change [`plot`](https://www.pygmt.org/v0.2.0/api/generated/pygmt.Figure.plot.html) to accept `geopandas` type objects, look for the 'geometry' column, and plot the ['Shapely'](https://shapely.readthedocs.io/en/latest/manual.html) geometries.

Currently, users need to go through a convoluted process such as with the following (adapted from https://forum.generic-mapping-tools.org/t/shapefile-to-gmt-python/834/21, also xref https://github.com/GenericMappingTools/foss4g2019oceania/pull/7):

```
fig = pygmt.Figure()
fig.coast(region="g", land="gray", water="lightblue")
points = [point for point in cities.geometry[:10]]  # plot 10 cities only
for point in points:
    x, y = point.coords.xy
    fig.plot(x=x, y=y, style="s0.2c", color="black")
fig.show()
```

produces:

![cities](https://user-images.githubusercontent.com/23487320/93409113-3f85f200-f8ea-11ea-8f3b-d3832de24a15.png)

Using `data=geodataframe` seems like the easiest, but we could consider alternative ways of implementing the `geopandas` integration. There might need to be separate code pathways for handling Points, Lines, and Polygons.

[...]

PR #990, opened 2021-03-02
Refactor plot and plot3d to use virtualfile_from_data

### LINKED ISSUE #1021: Support 2D list as input

PyGMT doesn't allow 2D list as input.

The script below doesn't work:
```python
import pygmt
fig = pygmt.Figure()

fig.plot(data=[[0, 0]], region=[-2, 2, -2, 2], frame=True, style="c0.5c", pen="1p")
fig.show()
```

but it works if I convert it to a 2D numpy array:
```python
import pygmt
fig = pygmt.Figure()

fig.plot(data=np.array([[0, 0]]), region=[-2, 2, -2, 2], frame=True, style="c0.5c", pen="1p")
fig.show()
```

**Full error message**

```
<ipython-input-15-cf008b39b139> in <module>
      2 fig = pygmt.Figure()
      3 
----> 4 fig.plot([[0, 0]], region=[-2, 2, -2, 2], frame=True, style="c0.5c", pen="1p")
      5 fig.show()

~/Gits/gmt/pygmt/pygmt/helpers/decorators.py in new_module(*args, **kwargs)
    286                 if alias in kwargs:
    287                     kwargs[arg] = kwargs.pop(alias)
--> 288             return module_func(*args, **kwargs)
    289 
    290         new_module.aliases = aliases

~/Gits/gmt/pygmt/pygmt/helpers/decorators.py in new_module(*args, **kwargs)
    430                         kwargs[arg] = separators[fmt].join(f"{item}" for item in value)
    431             # Execute the original function and return its output
--> 432             return module_func(*args, **kwargs)
    433 
    434         return new_module

~/Gits/gmt/pygmt/pygmt/src/plot.py in plot(self, x, y, data, sizes, direction, **kwargs)
    202     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
    203 
--> 204     kind = data_kind(data, x, y)
    205 
    206     extra_arrays = []

~/Gits/gmt/pygmt/pygmt/helpers/utils.py in data_kind(data, x, y, z)
     62         raise GMTInvalidInput("Too much data. Use either data or x and y.")
     63     if data is None and (x is None or y is None):
---> 64         raise GMTInvalidInput("Must provided both x and y.")
     65 
     66     if isinstance(data, str):

GMTInvalidInput: Must provided both x and y.
```

weiji14:

Added an additional `extra_arrays` parameter to the`virtualfile_from_data` function to accept optional numpy arrays from the plot and plot3d functions.

~~**TODO**: May need to tweak the `data_kind` function a little, to output `matrix` for true numpy array matrices and `vectors` for numpy arrays with mixed dtypes. This is to fix a segmentation fault on `test_plot3d_matrix_color`.~~ Edit: see comment at https://github.com/GenericMappingTools/pygmt/pull/990#discussion_r586959742

Part of #949. Once this is done, it will simplify the  `geopandas` integration (#608) (which will be done in a follow-up Pull Request at #1000).

Also fixes #1021 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1380,7 +1380,7 @@ class Session:
             raster grid, a vector matrix/arrays, or other supported data input.
         x/y/z : 1d arrays or None
             x, y and z columns as numpy arrays.
-        extra_arrays : list
+        extra_arrays : list of 1d arrays
             Optional. A list of numpy arrays in addition to x, y and z. All
             of these arrays must be of the same size as the x/y/z arrays.
 
@@ -1446,9 +1446,10 @@ class Session:
                 _data = [array for _, array in data.items()]
             except AttributeError:
                 try:
-                    # Just use virtualfile_from_matrix for 2D
-                    # numpy.ndarray which are not datetime (M) types
-                    assert data.ndim == 2 and not data.dtype.kind == "M"
+                    # Just use virtualfile_from_matrix for 2D numpy.ndarray
+                    # which are signed integer (i), unsigned integer (u) or
+                    # floating point (f) types
+                    assert data.ndim == 2 and data.dtype.kind in "iuf"
                     _virtualfile_from = self.virtualfile_from_matrix
                     _data = (data,)
                 except (AssertionError, AttributeError):
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1360,7 +1360,9 @@ class Session:
         with self.open_virtual_file(*args) as vfile:
             yield vfile
 
-    def virtualfile_from_data(self, check_kind=None, data=None, x=None, y=None, z=None):
+    def virtualfile_from_data(
+        self, check_kind=None, data=None, x=None, y=None, z=None, extra_arrays=None
+    ):
         """
         Store any data inside a virtual file.
 
@@ -1430,7 +1432,11 @@ class Session:
         if kind in ("file", "grid"):
             _data = (data,)
         elif kind == "vectors":
-            _data = (x, y, z)
+            _data = [np.atleast_1d(x), np.atleast_1d(y)]
+            if z is not None:
+                _data.append(np.atleast_1d(z))
+            if extra_arrays:
+                _data.extend(extra_arrays)
         elif kind == "matrix":  # turn 2D arrays into list of vectors
             try:
                 # pandas.DataFrame and xarray.Dataset types
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1442,8 +1442,15 @@ class Session:
                 # pandas.DataFrame and xarray.Dataset types
                 _data = [array for _, array in data.items()]
             except AttributeError:
-                # Python lists, tuples, and numpy ndarray types
-                _data = np.atleast_2d(np.asanyarray(data).T)
+                try:
+                    # Just use virtualfile_from_matrix for 2D
+                    # numpy.ndarray which are not datetime (M) types
+                    assert data.ndim == 2 and not data.dtype.kind == "M"
+                    _virtualfile_from = self.virtualfile_from_matrix
+                    _data = (data,)
+                except (AssertionError, AttributeError):
+                    # Python lists, tuples, and numpy ndarray types
+                    _data = np.atleast_2d(np.asanyarray(data).T)
 
         # Finally create the virtualfile from the data, to be passed into GMT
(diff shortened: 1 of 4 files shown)


pygmt/clib/session.py
```diff
@@ -1430,14 +1435,25 @@ def virtualfile_from_data(self, check_kind=None, data=None, x=None, y=None, z=No
...
-                _data = np.atleast_2d(np.asanyarray(data).T)
+                try:
+                    # Just use virtualfile_from_matrix for 2D
+                    # numpy.ndarray which are not datetime (M) types
+                    assert data.ndim == 2 and not data.dtype.kind == "M"
+                    _virtualfile_from = self.virtualfile_from_matrix
+                    _data = (data,)
+                except (AssertionError, AttributeError):
+                    # Python lists, tuples, and numpy ndarray types
+                    _data = np.atleast_2d(np.asanyarray(data).T)
```
weiji14, line 1456: Would appreciate advice on improving/refactoring this chunk of code. I'm half thinking whether to move some of the logic to `pygmt/helpers/utils.py`, i.e. have a new 'kind' besides file/grid/matrix/vectors:

https://github.com/GenericMappingTools/pygmt/blob/65b099a5873b4f9e76d7fccdbc8882d9b37870e3/pygmt/helpers/utils.py#L65-L72

Or we could just keep things like this as it is.

seisman: I didn't read the codes carefully, but I think the "matrix" kind is quite confusing. 

The GMT API function `GMT_Put_Matrix()`, the PyGMT wrapper `put_matrix()` and the virtualfile function `virtualfile_from_matrix()` all require a simple 2d matrix with a single dtype (e.g., `np.float` or `np.double`). 

However, currently, data types like pandas.DataFrame are also "matrix". So we have to check the data types to choose either `virtualfile_from_vector` or `virtualfile_from_matrix`.

I agree with you that we can/shoudl add a new kind to distinguish between a GMT-compatible "matrix" and a more complicated data structure.

weiji14: How about we keep this logic as is for now to resolve #1021 (which helps with #1020), and refactor things later to have more specific 'matrix' types?

seisman: Sounds good.

pygmt/clib/session.py
```diff
@@ -1378,6 +1380,9 @@ def virtualfile_from_data(self, check_kind=None, data=None, x=None, y=None, z=No
             raster grid, a vector matrix/arrays, or other supported data input.
         x/y/z : 1d arrays or None
             x, y and z columns as numpy arrays.
+        extra_arrays : list
```
seisman, line 1383:
```suggestion
        extra_arrays : list of 1d arrays
```

pygmt/clib/session.py
```diff
@@ -1430,14 +1435,25 @@ def virtualfile_from_data(self, check_kind=None, data=None, x=None, y=None, z=No
...
             try:
                 # pandas.DataFrame and xarray.Dataset types
                 _data = [array for _, array in data.items()]
             except AttributeError:
-                # Python lists, tuples, and numpy ndarray types
-                _data = np.atleast_2d(np.asanyarray(data).T)
+                try:
+                    # Just use virtualfile_from_matrix for 2D
+                    # numpy.ndarray which are not datetime (M) types
+                    assert data.ndim == 2 and not data.dtype.kind == "M"
```
seisman, line 1451: > assert data.ndim == 2 and not data.dtype.kind == "M"

The [`GMT_Put_Matrix()`](https://docs.generic-mapping-tools.org/dev/api.html#gmt-put-matrix) function only supports a few [numeric data types](https://docs.generic-mapping-tools.org/dev/api.html#tbl-types), but [numpy.dtype.kind](https://numpy.org/doc/stable/reference/generated/numpy.dtype.kind.html) can have values other than `M`, for example, `data.dtype.kind == "O"` will pass the assert statement but is not supported by `virtualfile_from_matrix`. Perhaps check `data.dtype.kind in 'iuf'`?

seisman:
One more comment: perhaps we should add a new test to make sure #1021 is fixed.

pygmt/clib/session.py
weiji14: Ok, I'll try `data.dtype.kind in 'iuf'` and see if it works on the test suite.

seisman (approved):
OK to merge after all tests pass and the styling issue is fixed.

PR #998, opened 2021-03-04
Move codecov.yml into .github directory and increase coverage target to 90 percent

weiji14:

One less configuration file to see in the top level directory.
See https://docs.codecov.io/docs/codecov-yaml#can-i-name-the-file-codecovyml

Helps with #904 

**Reminders**

**Slash Commands**


seisman (approved):
Great!

weiji14:
Just wondering, do we want to increase the target coverage from 80% to 90% (or higher)? We've been tracking at above 95% for the past six months it seems.

7iwb5xau

https://github.com/GenericMappingTools/pygmt/blob/f0a6a64ea49d4ff99de0aeb7f2c10e86f765a4b0/.github/codecov.yml#L5-L9

seisman:
> do we want to increase the target coverage from 80% to 90% (or higher)? We've been tracking at above 95% for the past six months it seems.

I'm fine with 90% or 95% coverage.

weiji14:
Ok, increased to 90%. Merging.

seisman:
image

I think we misunderstand what 90% means. 

We set the target of **patches** to 90%. I think it means at least 90% of the changed lines should be covered.

It's different than the **project** coverage. If I understand it correctly, the **project** coverage only pass if the current coverage is no smaller than the previous coverage.

weiji14:
> image
> 
> I think we misunderstand what 90% means.
> 
> We set the target of **patches** to 90%. I think it means at least 90% of the changed lines should be covered.
> 
> It's different than the **project** coverage. If I understand it correctly, the **project** coverage only pass if the current coverage is no smaller than the previous coverage.

Ah, so we need to set the **project** coverage as well then perhaps, in addition to the **patch** coverage?

seisman:
Maybe, but I didn't read the codecov documentation carefully.

PR #1002, opened 2021-03-04
Refactor Figure.meca() to better determine the data format code

seisman:

The `Figure.meca()` method is too complicated to read and maintain, and also has the lowest code coverage (~75%).

I will try to refactor the method, but in separate PRs. Otherwise the changes would be too big to review.

This PR is the first step to refactor the codes. The old code uses nested if-else-statement to determine the data format (**meca -S** option) from parameters `convention` and `component`.
The new code uses two dicts instead, and also wrap the code in a separate function `data_format_code `, because the function will also be used in `Figure.coupe()` (not wrapped yet).

I also add several doctests to check the function and increase the code coverage.

References: https://docs.generic-mapping-tools.org/latest/supplements/seis/meca.html#s

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -16,6 +16,84 @@ from pygmt.helpers import (
 )
 
 
+def data_format_code(convention, component="full"):
+    """
+    Determine the data format code for meca -S option.
+
+    See the meca() method for explanations of the parameters.
+
+    Examples
+    --------
+    >>> data_format_code("aki")
+    'a'
+    >>> data_format_code("gcmt")
+    'c'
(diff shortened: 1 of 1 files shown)


seisman:
Ping @tjnewton @liamtoney for review as you wrote the initial `Figure.meca()` method in #516.

tjnewton:
> Ping @tjnewton @liamtoney for review as you wrote the initial `Figure.meca()` method in #516.

Good idea. I like your changes.

liamtoney (approved):
Nice.

PR #1013, opened 2021-03-06
Use colorblind-friendly colors in the scatter plots gallery example

### LINKED ISSUE #1011: Make scatter plots gallery example colourblind friendly

The orange and green colours at the ["Scatter plots with a legend"](https://www.pygmt.org/v0.3.0/gallery/plot/scatter.html#sphx-glr-gallery-plot-scatter-py) gallery example can be hard to distinguish for people with red/green colour vision deficiencies. We should choose different colours instead if possible.

| [v0.3.0](https://www.pygmt.org/v0.3.0/gallery/plot/scatter.html#sphx-glr-gallery-plot-scatter-py) original colour | Deuteranopia (no green) simulation |
|--|--|
| original colour | Deuteranopia simulation |

The line that needs to change is this:

https://github.com/GenericMappingTools/pygmt/blob/65b099a5873b4f9e76d7fccdbc8882d9b37870e3/examples/gallery/plot/scatter.py#L24

References:
- https://developer.mozilla.org/en-US/docs/Tools/Accessibility_inspector/Simulation
- https://docs.generic-mapping-tools.org/6.1/gmtcolors.html#list-of-colors
- https://colorbrewer2.org/#type=qualitative&scheme=Set2&n=3

vitorgt:
https://github.com/GenericMappingTools/pygmt/issues/1011

I'm replying to an open good first issue (#1011), using color names defined in [GMT docs](https://docs.generic-mapping-tools.org/6.1/gmtcolors.html#list-of-colors) as referred by @weiji14 (thanks).

Fixes #1011


examples/gallery/plot/scatter.py
```diff
@@ -21,7 +21,7 @@
     projection="X10c/10c",
     frame=["xa0.2fg", "ya0.2fg", "WSrt"],
 )
-for color in ["blue", "orange", "green"]:
+for color in ["seagreen2", "chocolate1", "mediumpurple1"]:
```
michaelgrund, line 24:
```suggestion
for color in ["seagreen2", "chocolate1", "mediumpurple1"]:
```
Not sure if these colors solve the problem (please correct me if I'm worong). Maybe it would be better to select colors from cpts that really focus on the color blindness issue.

michaelgrund:
Converted the PR to draft to save resources.

examples/gallery/plot/scatter.py
vitorgt: I agree, but colorblind friendly colorschemes are very limited, and transparency makes this case even worst, I've changed green to a grey, this change will produce greater differences on protanopia and deuteranopia, but purple will get closer to grey on tritanopia (the rarest of the three).

Everywhere i looked on how to choose colorblind friendly colors they say not to rely only color, but to create other ways to differentiate set, usualy changing marker. So i added this.

Also, i couldn't find colorblind friendly cpts on gmt.

seisman: > transparency makes this case even worst,

Perhaps we can try to increase the transparency slightly?

examples/gallery/plot/scatter.py
```diff
@@ -21,19 +21,19 @@
...
-        style="c",
+        style=style,
         sizes=sizes,
         color=color,
         # Set the legend label,
-        # and set the circle size to be 0.25 cm (+S0.25c) in legend
-        label=f"{color}+S0.25c",
+        # and set the style size to be 0.25 cm (+S0.25{style}) in legend
+        label=f"{color}+S0.25{style}",
         transparency=70,  # set transparency level for all symbols
```
michaelgrund, line 37:
```suggestion
        transparency=50,  # set transparency level for all symbols
```
Perhaps 50 gives a better result.

examples/gallery/plot/scatter.py
michaelgrund: > 
> 
> > transparency makes this case even worst,
> 
> Perhaps we can try to increase the transparency slightly?

Agree, perhaps `transparency=50` gives a better result.

examples/gallery/plot/scatter.py
```diff
@@ -21,19 +21,19 @@
...
         y=y,
-        style="c",
+        style=style,
         sizes=sizes,
         color=color,
         # Set the legend label,
-        # and set the circle size to be 0.25 cm (+S0.25c) in legend
-        label=f"{color}+S0.25c",
+        # and set the style size to be 0.25 cm (+S0.25{style}) in legend
+        label=f"{color}+S0.25{style}",
```
core-man, line 36: See the warning at https://pygmt-ln1ssu8ap-gmt.vercel.app/gallery/plot/scatter.html#sphx-glr-gallery-plot-scatter-py:
```
plot [WARNING]: Length <unit> s not supported - revert to default unit [cm]
plot [WARNING]: Length <unit> t not supported - revert to default unit [cm]
```

`0.25c` means 0.25 cm, so we shouldn't use `style` as the unit. If we'd like to change the size for each symbol in the legend, it'd better to also add it in the for loop. But I feel `0.25c` is okay for those three symbols.
```suggestion
        label=f"{color}+S0.25c",
```

examples/gallery/plot/scatter.py
```diff
@@ -21,19 +21,19 @@
...
         x=x,
         y=y,
-        style="c",
+        style=style,
         sizes=sizes,
         color=color,
         # Set the legend label,
-        # and set the circle size to be 0.25 cm (+S0.25c) in legend
-        label=f"{color}+S0.25c",
+        # and set the style size to be 0.25 cm (+S0.25{style}) in legend
```
core-man, line 35:
```suggestion
        # and set the style size to be 0.25 cm (+S0.25c) in legend
```

seisman, line 35:
```suggestion
        # and set the symbol size to be 0.25 cm (+S0.25{style}) in legend
```

seisman:
> Everywhere i looked on how to choose colorblind friendly colors they say not to rely only color, but to create other ways to differentiate set, usualy changing marker.

I agree with it, but I think the example is focusing on different colors, not "different colors and different symbols" (we could add another example for it). 

I think we should only make changes to colors in this PR.

examples/gallery/plot/scatter.py
vitorgt: No, @michaelgrund, "transparency makes this case even worst" means we sould **remove** transparency.
Like ```transparency=90``` or remove that line completely (which would be equal to ```transparency=100``` (default)).

Transparenncy makes it worst beacuse it blends the colors and changes their luminance when adding two or more symbols with equal **or** different colors, notting that color addition with colorblindness is a little bit tricky. That's why transparency should be avoided if we want it more coverage considering colorblindness.

examples/gallery/plot/scatter.py
vitorgt: Thank you so much, and I'm sorry, I didn't know that, I thought it was the symbol to show, not the size in cm. Appreciated.

vitorgt:
> > Everywhere i looked on how to choose colorblind friendly colors they say not to rely only color, but to create other ways to differentiate set, usualy changing marker.
> 
> I agree with it, but I think the example is focusing on different colors, not "different colors and different symbols" (we could add another example for it).
> 
> I think we should only make changes to colors in this PR.

@seisman, I kinda disagree with that, the [example](https://www.pygmt.org/v0.3.0/gallery/plot/scatter.html#sphx-glr-gallery-plot-scatter-py) says:

> To create a scatter plot with a legend one may use a loop and create one scatter plot per item to appear in the legend and set the label accordingly.

It doesn't specify its focus, nor restrics the "legend" or the "label" to deal only with colors, and I reaffirm, color blindness accessibility must not rely only on colors, but I'm not a member of GMT, you have far more context within the project than I do, I'll do whatever you members think fits.

Before I continue I need to know:

* Do you want symbols to change?
* Do you want transparency to change? ([Please note this new reply](https://github.com/GenericMappingTools/pygmt/pull/1013#discussion_r593921826))
* [Are the new colors ok?](https://github.com/GenericMappingTools/pygmt/pull/1013#discussion_r589074955) (```["gray73", "darkorange", "slateblue"]``` I tried to plot them locally but i'm getting an impossible to solve error with "libgmt.so". I've checked them on color blindness tests but this case is special and a plot proof needs to be made)

examples/gallery/plot/scatter.py
weiji14: I think we're getting a bit confused here. `transparency=0` means opaque while `transparency=100` means invincible. We probably want to **reduce** transparency (i.e. increase opacity).

examples/gallery/plot/scatter.py
weiji14: Yes, please try 50, i.e. less transparency (more opaque). See also https://github.com/GenericMappingTools/pygmt/pull/1013#discussion_r594010939.

weiji14:
> * [Are the new colors ok?](https://github.com/GenericMappingTools/pygmt/pull/1013#discussion_r589074955) (`["gray73", "darkorange", "slateblue"]` I tried to plot them locally but i'm getting an impossible to solve error with "libgmt.so". I've checked them on color blindness tests but this case is special and a plot proof needs to be made)

Apologies for the late reply @vitorgt, the colour look great! But we could decrease the transparency a bit (i.e. increase opacity) as suggested at https://github.com/GenericMappingTools/pygmt/pull/1013#discussion_r589324327

| Original | Deuteranopia |
|--|--|
Original colour | Deuteranopia simulated colour

> * Do you want symbols to change?

Could you try changing it back to circles only to see what it looks like?

Oh, and just FYI, we have a Continuous Documentation setup, so you can check the live preview at https://pygmt-git-fork-vitorgt-patch-1-gmt.vercel.app/gallery/plot/scatter.html. Sorry for not mentioning it earlier.

examples/gallery/plot/scatter.py
michaelgrund: Right @weiji14. I think we should not drop transparency here since it allows to deal with overplotting.

meghanrjones:
Personally, I find that the transparency makes it more difficult to see the data points and would prefer to see the transparency parameter removed as @vitorgt suggested. Why not just reduce the overplotting? Gaussian noise alone isn't that interesting to plot anyways. Something like this could be a good compromise (warning: the plotting parameters are not pretty and the comments aren't accurate - just showing the idea here):

```
import numpy as np
import pygmt

np.random.seed(19680801)
n = 30  # number of random data points

fig = pygmt.Figure()
fig.basemap(
    region=[-10, 110, -10, 350],
    projection="X10c/10c",
    frame=["xa20fg", "ya50fg", "WSrt"],
)
for color, style, slope in [("gray73", "c", 1), ("darkorange", "s", 2), ("slateblue", "t", 3)]:
    x = np.linspace(0, 100, n)
    y = (slope*x) + 3*np.random.randn(n)  # random X and Y data in [0,1]
    sizes = np.random.normal(1,0.2,n) * 0.3  # random size [0,0.5], in cm
    # plot data points as circles (style="c"), with different sizes
    fig.plot(
        x=x,
        y=y,
        style=style,
        sizes=sizes,
        color=color,
        pen="black",
        # Set the legend label,
        # and set the style size to be 0.25 cm (+S0.25{style}) in legend
        label=f"{color}+S0.25c",
    )

fig.legend(position="JTL+o-3/-2c", box="+gwhite+pblack")  # set transparency level for legends
fig.show()
```

image

meghanrjones:
Thanks, @vitorgt for addressing this issue. Since the transparency/overplotting was not part of the initial accessible colors issue, it can be addressed separately and to wrap this up you can commit the suggestions from @core-man and @seisman and the reduced transparency suggestion by @michaelgrund whenever you get a chance. Here are specific answers to your previous questions:

> Before I continue I need to know:
> 
>     Do you want symbols to change?
No, your points are well made and the new symbol selections are fine. 
>     Do you want transparency to change? (Please note this new reply)
Yes, please commit this suggestion: https://github.com/GenericMappingTools/pygmt/pull/1013/files#r589324327. After this gets merged, the transparency and overplotting can be dealt with in a separate PR. You are more than welcome to open the next PR for the gallery example or leave it at this important contribution of improving the colors and adding distinct symbols.
>     Are the new colors ok? (["gray73", "darkorange", "slateblue"] I tried to plot them locally but i'm getting an impossible to solve error with "libgmt.so". I've checked them on color blindness tests but this case is special and a plot proof needs to be made)
Yes, as @weiji14 mentioned the new colors are good.

If you are OK with the suggestions, then you can click 'Files changed' in the PR and add the four suggested revisions to a batch commit using the 'Add suggestion to batch' option before committing them all.

seisman:
Ping @vitorgt to see if you have time to finalize this PR.

vitorgt:
Sorry for the late reply
I've moved back to circles and made them more opaque :)
@meghanrjones sorry, i couldn't do the 'Add suggestion to batch', so I just made a new commit with the changes

michaelgrund (commented):
Could you please move the example into the newly created **gallery/symbols** folder @vitorgt ?

seisman (approved):

michaelgrund (approved):
Looks good to me!

PR #1022, opened 2021-03-08
Add a gallery example showing different front styles

michaelgrund:
Here's another gallery example to highlight the indiviudal front styles one can use with PyGMT. Documentation is now added.

**Reminders**

**Slash Commands**


examples/gallery/line/linefronts.py
```diff
@@ -0,0 +1,84 @@
+"""
+Line fronts
+-----------
+
+Using the :meth:`pygmt.Figure.plot` method you can draw a so-called
+*front* which allows to plot specific symbols distributed along a line
+or curve. Typical use cases are weather fronts, fault lines,
+subduction zones, and more.
+
+A front can be drawn using ``style = "f[+-]gap[/size]"`` where *gap*
```
seisman, line 10:
```suggestion
A front can be drawn by passing **f**\[±]\ *gap*\[/*size*] to the ``style`` parameter where *gap*
```

examples/gallery/line/linefronts.py
```diff
@@ -0,0 +1,84 @@
...
+    # use other pen for the outline of the symbol
+    "f1c/0.4c+r+t+p1.5p,dodgerblue",
+    # line with triangle front style, symbols are plotted on the right side of the front
+    # and offset is set to 0.3 cm, skip the outline
+    "f0.5c/0.3c+r+t+o0.3c+p",
+    # line with triangle front style, symbols are plotted on the right side of the front
+    # and offset is set to 0.3 cm, skip the outline and make the main front line invisible
+    "f0.5c/0.3c+r+t+o0.3c+p+i",
+]:
+    y -= 1  # vove the current line down
```
seisman, line 73:
```suggestion
    y -= 1  # move the current line down
```

examples/gallery/line/linefronts.py
```diff
@@ -0,0 +1,84 @@
...
+    "f1c/0.25c",
+    # line with box front style
+    "f1c/0.25c+b",
+    # line with circle front style
+    "f1c/0.25c+c",
+    # line with triangle front style
+    "f1c/0.3c+t",
+    # line with left-lateral "slip" front style, angle is set to 45 and
+    # offset to 2.25 cm
+    "f5c/1c+l+s45+o2.25c",
```
core-man, line 54: When we use `+s`, we must indicate either `+l` or `+r`.
```suggestion
    # line with left-lateral ("+l") slip ("+s") front style, angle is set to 45 and
    # offset to 2.25 cm
    "f5c/1c+l+s45+o2.25c",
```

examples/gallery/line/linefronts.py
```diff
@@ -0,0 +1,84 @@
...
+    "f1c/0.3c+l+t",
+    # line with triangle front style, symbols are plotted on the right side of the front
+    # use other pen for the outline of the symbol
+    "f1c/0.4c+r+t+p1.5p,dodgerblue",
+    # line with triangle front style, symbols are plotted on the right side of the front
+    # and offset is set to 0.3 cm, skip the outline
+    "f0.5c/0.3c+r+t+o0.3c+p",
+    # line with triangle front style, symbols are plotted on the right side of the front
+    # and offset is set to 0.3 cm, skip the outline and make the main front line invisible
+    "f0.5c/0.3c+r+t+o0.3c+p+i",
```
core-man, line 71: How about using the same `size/gap` (`1c/0.3c`) for those cases to only compare `+p` argument.

We may add another example to only compare `+o` argument after line 62 `"f1c/0.3c+l+t",`:
```
# line with triangle front style, symbols are plotted on the left side of the front
# and offset is set to 0.3 cm
"f1c/0.3c+l+t+o0.3c",
```

examples/gallery/line/linefronts.py
```diff
@@ -0,0 +1,84 @@
...
+or curve. Typical use cases are weather fronts, fault lines,
+subduction zones, and more.
+
+A front can be drawn by passing **f**\[±]\ *gap*\[/*size*] to the ``style`` parameter where *gap*
+defines the distance gap between the symbols and *size* the symbol size.
+If *gap* is negative, it is interpreted to mean the number of symbols
+along the front instead. If *gap* has a leading + then we use the value
+exactly as given [Default will start and end each line with a symbol,
+hence the *gap* is adjusted to fit]. If *size* is missing it is set to 30%
+of the *gap*, except when *gap* is negative and *size* is thus required.
```
core-man, line 16: Do we want to have a list of optional arguments here as https://docs.generic-mapping-tools.org/latest/plot.html#s so that the reader can know the full list while they can read more details of each one?: `[+l|+r][+b+c+f+s+t][+ooffset][+p[pen]].`

```suggestion
of the *gap*, except when *gap* is negative and *size* is thus required.
Append more elements to plot different fronts
[**+l**\|\ **+r**\ ][**+b**\ **+c**\ **+f**\ **+s**\ **+t**\][**+o**\ *offset*][**+p**\[*pen*]].
```
Screenshot from 2021-03-11 16-47-15

examples/gallery/line/linefronts.py
michaelgrund: Correct.

examples/gallery/line/linefronts.py
michaelgrund: In principle that's a good idea, however, I'm not sure if we should add it here since the indiviudal options are described afterwards. What do the others think @GenericMappingTools/python-contributors ?

seisman:
FYI, I just merged the master branch and move the example to the `lines` directory.

examples/gallery/line/linefronts.py
seisman: I think it's fine to use different size and gap, so that readers may better understand how these arguments change the appearances of the fronts.

examples/gallery/line/linefronts.py
seisman: Perhaps not. I don't think we should explain the full syntax of one parameter in a gallery example.

michaelgrund: Ok, than we leave it as is.

examples/gallery/lines/linefronts.py
```diff
@@ -0,0 +1,84 @@
...
+-----------
+
+Using the :meth:`pygmt.Figure.plot` method you can draw a so-called
+*front* which allows to plot specific symbols distributed along a line
+or curve. Typical use cases are weather fronts, fault lines,
+subduction zones, and more.
+
+A front can be drawn by passing **f**\[±]\ *gap*\[/*size*] to the ``style`` parameter where *gap*
+defines the distance gap between the symbols and *size* the symbol size.
+If *gap* is negative, it is interpreted to mean the number of symbols
```
seisman, line 12: Please wrap this line to similar lengths (ideally less than 80 characters).

seisman (approved):
This PR looks good to me.

core-man:
Look great~

examples/gallery/lines/linefronts.py
michaelgrund: Done! Thanks 😉

PR #1006, opened 2021-03-04
Add a gallery example to show coloring of points by categories

michaelgrund:
To close the >2 year old issue #244  here's a gallery example highlighting the coloring of points by categories.

Closes #244.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Write detailed docstrings for all functions/methods.
- [x] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1031, opened 2021-03-09
Support passing a sequence to the spacing parameter of pygmt.info()

michaelgrund:
Allow to pass a tuple to the ``spacing`` parameter as suggested in #1006.

**Slash Commands**

### DIFF
--- a/pygmt/src/info.py
+++ b/pygmt/src/info.py
@@ -7,7 +7,9 @@ from pygmt.helpers import GMTTempFile, build_arg_string, fmt_docstring, use_alia
 
 
 @fmt_docstring
-@use_alias(C="per_column", I="sequence", T="nearest_multiple", V="verbose", f="coltypes")
+@use_alias(
+    C="per_column", I="sequence", T="nearest_multiple", V="verbose", f="coltypes"
+)
 def info(table, **kwargs):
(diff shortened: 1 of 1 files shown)


pygmt/src/info.py
```diff
@@ -7,7 +7,7 @@
 
 
 @fmt_docstring
-@use_alias(C="per_column", I="spacing", T="nearest_multiple", V="verbose", f="coltypes")
+@use_alias(C="per_column", I="sequence", T="nearest_multiple", V="verbose", f="coltypes")
```
seisman, line 10: The change is incorrect. The `@use_alias` define the `spacing` parameter as an alias of **I**. And the `kwargs_to_strings` defined that **I** (i.e., `spacing`) can be a *sequence*, i.e, `spacing=(3, 2)` will be converted to `-I3/2`.
```suggestion
@use_alias(C="per_column", I="spacing", T="nearest_multiple", V="verbose", f="coltypes")
@kwargs_to_strings(I="sequence")
```

michaelgrund: I'm sorry, overlooked the decorator... Thanks for the fix @seisman.

seisman (approved):

PR #1024, opened 2021-03-09
Replace underscores with hyphens in projection gallery file names.

willschlitzer:
As discussed in #995, the project is inconsistent with under underscores or hyphens in file names, and advocates switching every file name to use hyphens. This PR makes that change for the projection gallery examples.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1034, opened 2021-03-10
Add underscore guideline to CONTRIBUTING.md

willschlitzer:
As mentioned by @core-man in #1024, this adds to the convention of underscores between words in file names in the contributing guidelines.

**Reminders**

**Slash Commands**


CONTRIBUTING.md
```diff
@@ -210,6 +210,7 @@ General guidelines for pull requests (PRs):
 * Larger changes should be broken down into smaller components and integrated
   separately.
 * Bug fixes should be submitted in separate PRs.
+* New files should use a hyphen between words in the file name, not an underscore.
```
weiji14, line 213: This seems rather broad in scope. Do we want to restrict this to just the 'doc/' and 'examples/' folder, or include things in '.github/' and 'pygmt/' and the top level directory as well? I don't think it's ok to have hyphens for *.py files that have to be imported for example (e.g. `import pygmt.datasets.earth-relief` won't work on an `earth-relief.py` I think). I think underscores are actually preferred in most Python packages (e.g. https://pyphilly.org/hyphen-and-underscores-when-use-each-django-and-python/).

weiji14: Apologies @willschlitzer, I think we should stick with underscore. I just had a look around the galleries of `matplotlib`, `cartopy` and `geopandas` and they all use underscores:

- https://github.com/matplotlib/matplotlib/tree/v3.3.4/examples/pyplots
- https://github.com/SciTools/cartopy/tree/v0.18.0/lib/cartopy/examples
- https://github.com/geopandas/geopandas/tree/v0.9.0/doc/source/gallery

To be honest, I think we can close this PR #1034 and #1024 too since it's a Python standard in PEP8. Really sorry that I didn't catch this earlier yesterday, I know this must have taken up a lot of your time :bow: :bow: :bow:

seisman: I agree that underscores are better than hyphens for Python scripts. Sorry for the inconvenience.

seisman: BTW, we should still add a guideline like "Use underscores for all python file names, not hyphens".

core-man: If we use underscores, shall we also make it as a rule in contributing.md?or, we assume everyone know it. @weiji14

weiji14: Sure, let's put this 'underscore' rule in the guideline and also hint at [PEP8](https://www.python.org/dev/peps/pep-0008/).

```suggestion
* Use underscore for all Python (*.py) files as per [PEP8](https://www.python.org/dev/peps/pep-0008/), not hyphens
```

CONTRIBUTING.md
```diff
@@ -210,6 +210,7 @@ General guidelines for pull requests (PRs):
 * Larger changes should be broken down into smaller components and integrated
   separately.
 * Bug fixes should be submitted in separate PRs.
+* Use underscore for all Python (*.py) files as per [PEP8](https://www.python.org/dev/peps/pep-0008/), not hyphens
```
seisman, line 213:
```suggestion
* Use underscores for all Python (*.py) files as per [PEP8](https://www.python.org/dev/peps/pep-0008/), 
  not hyphens.
```

weiji14 (approved):

seisman:
I think we should apply the underscore rule to directories. For example, a URL like https://www.pygmt.org/dev/gallery/3d-plots/3d_perspective_view.html is pretty weird.

willschlitzer:
> I think we should apply the underscore rule to directories. For example, a URL like the one below is pretty weird.
> 
> ```
> https://www.pygmt.org/dev/gallery/3d-plots/3d_perspective_view.html
> ```

Added in e9102b3

seisman (approved):

PR #1045, opened 2021-03-13
Improve the image size in the documentation

### LINKED ISSUE #1038: The gallery thumbnails are too small

Here is a comparison between the [PyGMT gallery](https://www.pygmt.org/dev/gallery/index.html) and the [matplotlib gallery](https://matplotlib.org/stable/gallery/index.html)

|PyGMT | Matplotlib |
|---|---|
| image | image |

You may notice that the PyGMT gallery thumbnails are always smaller than the ones in matplotlib and [sphinx-gallery](https://sphinx-gallery.github.io/stable/auto_examples/index.html). The reason is that we limit the max width of images to 80% in the CSS:

https://github.com/GenericMappingTools/pygmt/blob/aff5db7f0f3d13e75313648528a50edcd2500b80/doc/_static/style.css#L41-L43

We made the change in #420 to avoid too large badges in the changelog file (https://github.com/GenericMappingTools/pygmt/pull/420#issuecomment-622162738). Apparently, the 80% CSS style is not ideal, and it affects all images. 

The simplest solution is changing `max-width` to 100%. Here is a comparison if we change the max-width to 100%. The gallery thumbnails look larger and better.

| max-width=80% | max-width=100% |
|---|---|
| image | image |

However, it also increases the size of other images to 100%. For example,

| max-width=80% | max-width=100% |
|---|---|
| image | image |
| image | image |

To me, 80% max-width looks better for these individual large images, but 100% max-width looks better for gallery thumbnails.

We should find a way to control the different settings of the two groups of images.

BTW, looking at the matplotlib gallery, I think we can also:

- add a box to each thumbnail
- centering the titles of the gallery examples

seisman:

See #1038 for context.

For the three different types of images, they now have different settings for max-width:

1. Images included using the `figure` directive: default to 100%, but we can customize the figure width using the `:width:` option of the `figure` directive. **Preview:** https://pygmt-git-thumbnail-gmt.vercel.app/overview.html
2. Single images included by calling `Figure.show()`: max-width is 80%. **Preview:** https://pygmt-git-thumbnail-gmt.vercel.app/gallery/maps/land_and_water.html
3. Gallery thumbnails: max-width is 100%. **Preview**: https://pygmt-git-thumbnail-gmt.vercel.app/gallery/index.html

I also add a thin box for each gallery thumbnail and center the thumbnail titles.

Fixes #1038

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -163,3 +163,13 @@ a.copybtn {
     font-weight: bold;
     font-style: italic;
 }
+
+/* Style for images and thumbnails of sphinx-gallery */
+.sphx-glr-thumbcontainer {
+    border: solid #d6d6d6 1px!important;
+    text-align: center!important;
+}
+
+.sphx-glr-single-img {
+    max-width: 80%!important;
+}
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -38,10 +38,6 @@ p {
     margin-bottom: 50px;
 }
 
-.rst-content img {
-    max-width: 80%;
-}
-
 .sidebar-title {
     margin-top: 10px;
     margin-bottom: 0px;
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -43,6 +43,7 @@ These are conference presentations about the development of PyGMT (previously
 
 .. figure:: https://img.youtube.com/vi/SSIGJEe0BIk/maxresdefault.jpg
     :target: https://www.youtube.com/watch?v=SSIGJEe0BIk
+    :width: 80%
     :align: center
     :alt: ROSES 2020 youtube video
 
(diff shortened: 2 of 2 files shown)


meghanrjones (approved):
Much nicer!

meghanrjones:
@seisman, do you have recommended reference for finding the name of sphinx and/or sphinx-extension elements for CSS? For example, sphx-glr-thumbcontainer is in the sphinx-gallery docs but it's not obvious to me how to find out about sphx-glr-single-img.

seisman:
I didn't read the docs. Instead, I use the web developer tools to find the CSS properties.

Put your mouse on the figure you want to check, right-click and choose "Inspect" (I'm using Microsoft Edge but other web browsers should be similar):
image

Then you can check the HTML code and CSS styles for any elements: 
image

meghanrjones:
> I didn't read the docs. Instead, I use the web developer tools to find the CSS properties.
> 
> Put your mouse on the figure you want to check, right-click and choose "Inspect" (I'm using Microsoft Edge but other web browsers should be similar):
> image
> 
> Then you can check the HTML code and CSS styles for any elements:

This is very helpful to know, thank you!

PR #996, opened 2021-03-03
Improve PyGMT documentation

core-man:
**Description of proposed changes**

Related to #983.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #995, opened 2021-03-03
Reorganize gallery examples

willschlitzer:
This pull request reorganizes the gallery examples to have more specific categories.

Fixes #876

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

Issue #983, opened 2021-03-01
Improve PyGMT documentation


core-man:
We have to decide whether we want to add quotes to all arguments, e.g., `'05m'` or `05m`. Originally posted in https://github.com/GenericMappingTools/pygmt/pull/976#discussion_r584484932. 

We use quotes for many arguments, e.g.,
https://github.com/GenericMappingTools/pygmt/blob/afe9293d8d6705587be9982909820d9ebf5e75eb/pygmt/datasets/earth_relief.py#L34-L39

Meanwhile, we don't use quotes in some cases, e.g.,
https://github.com/GenericMappingTools/pygmt/blob/afe9293d8d6705587be9982909820d9ebf5e75eb/pygmt/datasets/earth_relief.py#L44-L45

We could refer to other projects posted in https://github.com/GenericMappingTools/pygmt/pull/976#discussion_r584759875
> Just checked other projects, numpy (https://numpy.org/doc/stable/reference/generated/numpy.datetime_as_string.html#numpy.datetime_as_string), pandas (https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) and matplotlib (https://matplotlib.org/stable/api/legend_api.html) use `'string'`, not `` ``'string'`` ``.

seisman:
Also see https://github.com/GenericMappingTools/pygmt/issues/631

core-man:
This could be an off-topic issue. I feel I am totally new to the PyGMT documentation (https://www.pygmt.org/latest/). What I am feeling when I first go through the documentation may represent some new PyGMT users. I plan to go through the documentation from the landing page to the end and make some comments about what I feel in a PR/issue.

I quickly read `the landing page`, `Overview`, `Installing`, and `Making your first figure`. I have a few comments or confusion. Just take some for examples:

1. https://www.pygmt.org/latest/install.html#installing-gmt-and-other-dependencies: `conda create --name pygmt python=3.9 pip numpy pandas xarray netcdf4 packaging gmt`.
   The `pip` suddenly appears. It is not in the dependence list or above.
2. https://www.pygmt.org/latest/install.html#using-pip
   I know how to install latest development version from the link, but I don't know how to update the latest dev. I think it could be updated by running the same command again.
3. Do we have to add `$` before Linux/macOS commands? For example, `$ pip install pygmt`.
[...]

PR #1052, opened 2021-03-13
Update the title for the images/EPS on maps gallery example

core-man:

Update the title for the [images/EPS on maps gallery example](https://www.pygmt.org/dev/gallery/images/image.html#sphx-glr-gallery-images-image-py) (Originally posted in https://github.com/GenericMappingTools/pygmt/pull/995#issuecomment-791184821)

>> By the way, shall we rename this gallery to something like Images or EPS files on figures since we can put them on any figure instead of only a map, right?

> Sounds good. Actually, Figure.image() can recognize most image formats (png, jpg, eps, pdf et al.) via GDAL. So I think we could just say "Put images on figures" or similar.

A tiny improvement is also added for the comment of `position` argument.

Related to #983 and #996.

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/images/image.py
+++ b/examples/gallery/images/image.py
@@ -9,7 +9,6 @@ The ``position`` parameter allows us to set a reference point on the map for
 the image.
 """
 import os
-
 import pygmt
 
 fig = pygmt.Figure()
--- a/examples/gallery/images/image.py
+++ b/examples/gallery/images/image.py
@@ -9,6 +9,7 @@ The ``position`` parameter allows us to set a reference point on the map for
 the image.
 """
 import os
+
 import pygmt
 
 fig = pygmt.Figure()
--- a/examples/gallery/images/image.py
+++ b/examples/gallery/images/image.py
@@ -1,24 +1,23 @@
 """
-Images or EPS files on maps
----------------------------
-The :meth:`pygmt.Figure.image` method can be used to read and
-place a raster image file or an Encapsulated PostScript file
-on a map. We must specify the file as *str* via the ``imagefile``
-parameter or simply use the filename as the first argument. You can
-also use a full URL pointing to your desired image. The ``position``
-parameter allows us to set a reference point on the map for the image.
-
+Images on figures
+-----------------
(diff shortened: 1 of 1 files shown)


examples/gallery/images/image.py
```diff
@@ -1,24 +1,23 @@
...
-on a map. We must specify the file as *str* via the ``imagefile``
-parameter or simply use the filename as the first argument. You can
-also use a full URL pointing to your desired image. The ``position``
-parameter allows us to set a reference point on the map for the image.
-
+Images on figures
+-----------------
+The :meth:`pygmt.Figure.image` method can be used to read and place an image
+file in many formats (e.g., png, jpg, eps, pdf) on a figure. We must specify
+the file as *str* via the ``imagefile`` parameter or simply use the filename as
```
seisman, line 6:
```suggestion
the filename via the ``imagefile`` parameter or simply use the filename as
```

examples/gallery/images/image.py
```diff
@@ -1,24 +1,23 @@
...
+The ``position`` parameter allows us to set a reference point on the map for
+the image.
 """
 import os
 
 import pygmt
 
 fig = pygmt.Figure()
-
 fig.basemap(region=[0, 2, 0, 2], projection="X6c", frame=True)
```
seisman, line 16: The example may look better for a larger image size.
```suggestion
fig.basemap(region=[0, 2, 0, 2], projection="X10c", frame=True)
```

examples/gallery/images/image.py
```diff
@@ -1,24 +1,23 @@
...
+Images on figures
+-----------------
+The :meth:`pygmt.Figure.image` method can be used to read and place an image
+file in many formats (e.g., png, jpg, eps, pdf) on a figure. We must specify
+the filename via the ``imagefile`` parameter or simply use the filename as
+the first argument. You can also use a full URL pointing to your desired image.
+The ``position`` parameter allows us to set a reference point on the map for
+the image.
 """
 import os
```
core-man, line 12:
```suggestion
```

seisman (approved):

PR #1033, opened 2021-03-10
Separate workflows for running tests and building documentation

seisman:
**Description of proposed changes**

As mentioned in #830, this PR separates the current "Tests" workflow into two workflows to speedup our CI jobs.

- `ci_test.yaml`: runs the full tests
- `ci_docs.yml`: build the documentation and deploy it on master branch 

**TODO**

- [x] Trigger workflows only if necessary (i.e., add `paths` and `paths-ignore` to `push` and `pull_request`)
- [x] Update MAINTENANCE.md

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1055, opened 2021-03-14
WIP: Setup poetry in pyproject.toml

weiji14:

Initializing [poetry](https://python-poetry.org) to see if it can speed up the install step in our Continuous Integration.

`conda` works in serial and can be a bit slow, while `poetry` can install things in parallel. The idea here is to use `conda` only for non-Python packages (i.e. GMT and make), and `poetry` for Python dependencies (e.g. numpy, pandas, etc).

Notes:
- Use `poetry install --no-dev` to get only the minimum required dependencies for PyGMT
- Use `poetry install` to get both minimum and development dependencies (e.g. pytest, black, codecov, etc)
- Use `poetry install --no-dev --extras docs` to install minimum and dependencies for building docs (e.g. sphinx, etc)
- Use `poetry install --extras docs` to get all of the above

Addresses #584, may help with #690

**Reminders**

**Slash Commands**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,47 @@
+[tool.poetry]
+name = "pygmt"
+version = "0.3.1"
+description = "A Python interface for the Generic Mapping Tools"
+authors = ["The PyGMT Developers"]
+license = "BSD License"
+
+[tool.poetry.dependencies]
+myst-parser = {version = "*", optional = true}
+netCDF4 = "*"
+numpy = "*"
+packaging = "*"
+pandas = "*"
+python = "^3.7"
+sphinx = {version = "*", optional = true}
+sphinx-copybutton = {version = "*", optional = true}
+sphinx-gallery = {version = "*", optional = true}
+sphinx_rtd_theme = {version = "0.4.3", optional = true}
+xarray = "*"
+
+[tool.poetry.dev-dependencies]
+black = "*"
+blackdoc = "*"
+codecov = "*"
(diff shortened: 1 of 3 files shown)


seisman:
Just in case you don't notice it: poetry doesn't work well with setuptools_scm, so the version is `v0.3.1`, not `v0.3.1.dev100+xxxxx`.

pyproject.toml
```diff
@@ -1,3 +1,47 @@
+[tool.poetry]
+name = "pygmt"
+version = "0.3.1"
```
weiji14, line 3: > Just in case you don't notice it: poetry doesn't work well with setuptools_scm, so the version is `v0.3.1`, not `v0.3.1.dev100+xxxxx`.

Yeah I realize that this hardcoding is bad. There's https://github.com/mtkennerly/poetry-dynamic-versioning but doesn't look like a very official solution.

To be honest, this PR isn't intended to replace `setup.py` just yet. Just wanted to see if mixing `conda` and `poetry` might shave a few minutes off the CI test time.

weiji14:
We've got #1033 so this PR probably isn't needed anymore, but I'll keep this open for a while as it'll be nice to cherry-pick the poetry stuff in pyproject.toml from this PR into somewhere else to solve #690 or something else :)

weiji14:
Closing this PR. May reopen it in the future if there's any interest in using `poetry` :slightly_smiling_face:

Issue #891, opened 2021-02-14
plot and plot3d: Allow passing a numpy array as intensity


seisman:
> If using ``intensity=True``, we will instead read *intens* from the first data column after the symbols parameters.

As I understand it, there are at least three or four different ways to specify an intensity/intensities.

- Simply give a constant intensity like `intensity=0.5`
- Pass a numpy array to `intensity` (similar to the `x`, `y`, `size` parameters): I think it's not implemented yet
- Set `intensity=True`, and read data from an input file or from a 2D numpy array (via the `data` parameter).

The format of the input file of **plot** is (https://docs.generic-mapping-tools.org/latest/plot.html#s):

> x y [ z ] [ size ] [ symbol-parameters ] [ symbol ]

So, for the 3rd case, the format of the input file should be:

> x y [ z ] [ size ] [ symbol-parameters ] intensity [ symbol ]

_Originally posted by @seisman in https://github.com/GenericMappingTools/pygmt/pull/881#discussion_r575829230_

core-man:
I'd like to help with this issue.

seisman:
Implemented in #1065 and #1109.

PR #1036, opened 2021-03-11
Initialize data version control for managing test images

weiji14:
**Description of proposed changes**

Using a data version control package called [`dvc`](https://github.com/iterative/dvc) to manage the PNG test images in our repository!

In a nutshell, we will only store the hash of the PNG on GitHub (in a *.png.dvc file) and the actual PNG will be stored at https://dagshub.com/GenericMappingTools/pygmt (see e.g. https://dagshub.com/GenericMappingTools/pygmt/src/data_version_control/pygmt/tests/baseline/test_logo_on_a_map.png).

image

List of commands ran so far on this 'data_version_control' branch (will update as I test things out):

### Installing DVC for developing PyGMT

```
cd ~/path/to/pygmt
conda activate pygmt
make install
pip install dvc
```

### DVC init

```
dvc init # creates .dvcignore file and .dvc/ folder
# remove .dvc/plots folder as won't be used
# git add only the .dvcignore, .dvc/.gitignore and .dvc/config file
git add .dvcignore .dvc/.gitignore .dvc/config
git commit -m "Initialize data version control"
```

### Setup DVC remote

```
dvc remote add upstream https://dagshub.com/GenericMappingTools/pygmt.dvc # updates .dvc/config file with remote URL
dvc remote default upstream  # set default dvc remote to 'upstream'
```

---

### DVC remote authentication config (changes .dvc/config.local)
```
dvc remote modify upstream --local auth basic
dvc remote modify upstream --local user "$DAGSHUB_USER"
dvc remote modify upstream --local password "$DAGSHUB_PASS"  # generate token at https://dagshub.com/user/settings/tokens
```

## Generate new baseline PNG images
```
pytest --mpl-generate-path=baseline pygmt/tests/test_logo.py  # generate new baseline images
mv baseline/*.png pygmt/tests/baseline/
dvc add pygmt/tests/baseline/test_logo.png pygmt/tests/baseline/test_logo_on_a_map.png
git rm -r --cached 'pygmt/tests/baseline/test_logo.png'
git rm -r --cached 'pygmt/tests/baseline/test_logo_on_a_map.png'
# echo "/test_*.png" > pygmt/tests/baseline/.gitignore
```

### Push images to DVC remote
```
dvc add pygmt/tests/baseline/test_logo.png pygmt/tests/baseline/test_logo_on_a_map.png
git add pygmt/tests/baseline/.gitignore pygmt/tests/baseline/test_*.png.dvc
git commit -m "Add test_logo_*.png into DVC"

[...]

Issue #963, opened 2021-02-24
Rethink the testing mechanism for images


seisman:
_If you're unclear about how PyGMT tests images, please read the ["Testing plots" section in the contributing guides](https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#testing-plots) first._

---

In short, for image-based tests, we need to specify the baseline/reference image. When we make any changes to the code, we can generate the new "test" image and compare it with the "baseline" image. If the two images are different, then we know the changes break the tests. The most important thing is, to **ensure that the "baseline" images are correct.** 

Currently, we have two different methods to generate the "baseline" image and compare them:

1. using the decorator `@pytest.mark.mpl_image_compare`
2. using the decorator `@check_figures_equal()`

---

The `@pytest.mark.mpl_image_compare` method is the most straightforward way to do image testing. Using the decorator, we need to generate baseline images, check their correctness, and store them in the repository (https://github.com/GenericMappingTools/pygmt/tree/master/pygmt/tests/baseline).

**Pros**:

1. We can visually check the baseline images to make sure they are correct

**Cons**:

1. Have to store the static PNG images in the repository. The repository size grows quickly.

--- 

To avoid storing many large static images in the repository, we (mainly @weiji14 and @seisman) had some discussions (in #451, #522) and developed the `@check_figures_equal` decorator (#555, #590, #600).

Below is an example test using the `@check_figures_equal()` decorator:

https://github.com/GenericMappingTools/pygmt/blob/e0579275d00c7a69d25a82b4c8936e0b0b1b7577/pygmt/tests/test_basemap.py#L67-L77

In this example, the baseline/reference image `fig_ref` is generated using `basemap(R="0/360/0/1000", J="P6i", B="afg")`, while the test image `fig_test` is generated using `basemap(region=[0, 360, 0, 1000], projection="P6i", frame="afg")`. We can't see what the baseline image looks like, but we're somehow confident that the baseline image is correct, because the `basemap` wrapper is very simple.

**Pros**:

1. Don't need to store static images in the repository, thus keep the repository size small

**Cons**:

1. For each test, we have to generate two images (baseline and test images), which doubles the execution time
2. We can't visually check the correctness of the baseline images
3. If we decided to disable single-character parameters (i.e, `J="X10c/10c"` is disallowed) as proposed in #262 (also related to #256), then most of the code for generating reference images will be invalid.

For some complicated wrappers, we even can't easily know if the reference image is correct. For example,

[...]

PR #1065, opened 2021-03-15
Allow passing a array as intensity for plot

core-man:

Pass a NumPy array to `intensity`  (similar to the `x`, `y`, `size` parameters) for the `plot` method when we use `x`/`y` as the input.  (See #891 for details)

Currently, we can have two ways to specify an intensity/intensities.
- Simply give a constant intensity like `intensity=0.5`, which works for both `x`/`y` and `data` input. This is similar to `-I0.5` in the GMT command line.
- Set `intensity=True`, and read intensities from an *input file* or from a *2D NumPy array*, which is only for the `data` parameter. This is similar to `-I` in the GMT command line.

So, when we want to change the intensities for different points, we can only use the `data` parameter now. This PR tries to pass an array to `intensity` so that we can set different intensities when we use `x`/`y` as point input. This way will not work for `data` input.

Related to #891 

**Reminders**

**Slash Commands**


seisman:
I haven't reviewed your codes, but please do `plot` and `plot3d` in two PRs.

core-man:
> I haven't reviewed your codes, but please do `plot` and `plot3d` in two PRs.

Okay. BTW, I only added a NumPy array support for `intensity` now to test if it works (it works). More work need to be done (e.g., `test`).

pygmt/src/plot.py
```diff
@@ -218,6 +219,14 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
             )
         extra_arrays.append(sizes)
 
+    if "I" in kwargs and isinstance(kwargs["I"], np.ndarray):
```
weiji14, line 222: I think we want to allow any array-like input, not just numpy arrays. The `virtualfile_from_data` function (see #990) will be handle any array-like input automatically.

```suggestion
    if "I" in kwargs and not isinstance(kwargs["I"], str):
```

This would follow the syntax of `kwargs["G"]` above:

https://github.com/GenericMappingTools/pygmt/blob/8e2b2e6fd4900ccae25282d9d50f0ccd34d400ee/pygmt/src/plot.py#L207-L208

weiji14 (commented):
Just one comment for now :)

> Okay. BTW, I only added a NumPy array support for `intensity` now to test if it works (it works). More work need to be done (e.g., add `test`).

Yes, time for you to pick up unit testing! I'm currently revamping the plot testing ways at #1036 but you can start first and we'll refactor later. The PyGMT guidelines are at https://github.com/GenericMappingTools/pygmt/blob/v0.3.1/CONTRIBUTING.md#testing-plots, but if you're new to (or need a refresher on) testing, I found https://automationpanda.com/python/ to be a good resource to understand the 'why' and 'how' of testing.

pygmt/src/plot.py
seisman: I think we should use the `is_nonstr_iter` function:

https://github.com/GenericMappingTools/pygmt/blob/69417b4df78c99c178fa3b47ac4617f181ff3fa5/pygmt/helpers/utils.py#L161

core-man: Using `is_nonstr_iter` to check the input `intensity` (3503319)

core-man:
A test with an array intensity is added (9adf272), referring to `test_plot_varying_transparency()`.
https://github.com/GenericMappingTools/pygmt/blob/1a2289a20519a4dbf1c78543acd4a654b981811e/pygmt/tests/test_plot.py#L264

core-man:
**Two related questions:**
1. The documentation for `color` parameter needs a revision in the API page of [plot](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot) (also [plot3d](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot3d.html#pygmt.Figure.plot3d))?

> **color** (str) – Select color or pattern for filling of symbols or polygons. Default is no fill.

`color` can also be a 1D array just like the `sizes` parameter when we use `x`/`y` parameters. See the last example in [Plotting data points](https://pygmt-rfx68rbxo-gmt.vercel.app/tutorials/plot.html#plotting-data-points).

2. We can now input array-like `sizes`, `color`, `intensity`. Shall we support array-like `symbols`?

The format of the input file of plot is ([**-S**](https://docs.generic-mapping-tools.org/latest/plot.html#s) and [**-I**](https://docs.generic-mapping-tools.org/latest/plot.html#i)):

> x y [ z ] [ size ] [ symbol-parameters ] [ intensity ] [ symbol ]

See an example with `x y  z  size intensity symbol`
```
gmt begin symbols png
gmt makecpt -Chot -T0/3/1
gmt plot -R0/5/0/1 -JX5c/1c -BWSne -Bxa1 -Bya0.5 -S -C -I -W1p,black << EOF
1  0.5   0   0.3  1  c
2  0.5   1   0.8 -1  t
3  0.5   2   0.5  0  i
EOF
gmt end show
```
symbols

pygmt/src/plot.py
```diff
@@ -218,6 +218,14 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
...
+        if kind != "vectors":
+            raise GMTInvalidInput(
+                "Can't use arrays for intensity if data is matrix or file."
+            )
+        extra_arrays.append(kwargs["I"])
+        kwargs["I"] = ""
+
     if "t" in kwargs and is_nonstr_iter(kwargs["t"]):
         extra_arrays.append(kwargs["t"])
         kwargs["t"] = ""
```
seisman, line 231: Two comments here:

- Many parameters (e.g., sizes, colors, transparency) don't work if data is matrix or file. I feel we should have a more general way to do the check, rather than raising the GMTInvalidInput exception for each parameter.

- The two options **-I** and **-t** share very similar codes, and we should be able to simplify it like:
```
for flag in ["I", "t"]:
    if flag in kwargs and is_nonstr_iter(kwargs[flag]):
        extra_arrays.append(kwargs[flag])
        kwargs[flag] = ""
```

seisman:
> 1. We can now input array-like `sizes`, `color`, `intensity`. Shall we support array-like `symbols`?

Yes, please open a new feature request.

pygmt/src/plot.py
core-man: > * Many parameters (e.g., sizes, colors, transparency) don't work if data is matrix or file. I feel we should have a more general way to do the check, rather than raising the GMTInvalidInput exception for each parameter.

We can use the `columns` parameter ([-i](https://docs.generic-mapping-tools.org/latest/gmt.html#icols-full) in GMT) to select which columns in `data` to use, right? Is the reason we raise this error that we don't want to mix using array-like `x`, `y`, `sizes` etc with the `data` parameter because this will make PyGMT maintenance more difficult? We have to parse both the `data` parameter and those array-like parameters. Is my understanding right?

pygmt/tests/test_plot.py
```diff
@@ -231,6 +242,42 @@ def test_plot_colors_sizes_proj(data, region):
     return fig
 
 
+@check_figures_equal()
```
seisman, line 245: @core-man Please rewrite the test following our new workflow. See https://github.com/GenericMappingTools/pygmt/issues/963#issuecomment-801606095 for details.

core-man: @seisman could you please add me as a collaborator (https://dagshub.com/GenericMappingTools/pygmt/settings/collaboration) as said in [CONTRIBUTING.md](https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#using-data-version-control-dvc-to-manage-test-images)? See my DAGsHub account (https://dagshub.com/core-man).

seisman: Done.

core-man: done in 3b24385.

pygmt/src/plot.py
core-man: > * The two options **-I** and **-t** share very similar codes, and we should be able to simplify it like:
> 
> ```
> for flag in ["I", "t"]:
>     if flag in kwargs and is_nonstr_iter(kwargs[flag]):
>         extra_arrays.append(kwargs[flag])
>         kwargs[flag] = ""
> ```

We don't need to raise an `GMTInvalidInput` for **t**.

seisman: We need. The codes don't check it.

core-man: An array-like **t** works for both `data` and `x`/`y`, right? Now we only need to raise `GMTInvalideInput` for `color`, `sizes` and `intensity`. See https://github.com/GenericMappingTools/pygmt/pull/1065#discussion_r596161372

seisman: array-like t doesn't work for data.

core-man:
I cannot preview the test figure `test_plot_varying_intensity.png` on DAGsHub. My feature branch `core-man:intensity-np-array` is not there.

seisman:
> I cannot preview the test figure `test_plot_varying_intensity.png` on DAGsHub. My feature branch `core-man:intensity-np-array` is not there.

Yes, that's a known limitation of the new workflow.

pygmt/src/plot.py
```diff
@@ -143,11 +143,13 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
         the coordinates of a *refpoint* which will serve as a fixed external
         reference point for all groups.
     {G}
-    intensity : float or bool
+    intensity : float or bool or 1d array
         Provide an *intens* value (nominally in the -1 to +1 range) to
```
seisman, line 147:
```suggestion
        Provide an *intensity* value (nominally in the -1 to +1 range) to
```

pygmt/src/plot.py
```diff
@@ -143,11 +143,13 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
...
         modulate the fill color by simulating illumination [None]. If
         using ``intensity=True``, we will instead read *intens* from the
```
seisman, line 149:
```suggestion
        using ``intensity=True``, we will instead read *intensity* from the
```

pygmt/src/plot.py
```diff
@@ -143,11 +143,13 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
...
         first data column after the symbol parameters (if given).
+        *intensity* can also be a 1d array to set varying intensity for
+        symbols, but it is only valid if using ``x``/``y``.
```
seisman, line 152:
```suggestion
        symbols, but it is only valid for ``x``/``y`` pairs.
```

pygmt/tests/test_plot.py
```diff
@@ -231,6 +242,29 @@ def test_plot_colors_sizes_proj(data, region):
...
+    x = np.arange(1, 10)
+    y = np.arange(1, 10)
+    intensity = np.linspace(-1, 1, len(x))
+
+    fig = Figure()
+    fig.plot(
+        x=x,
+        y=y,
+        region=[0, 10, 0, 10],
+        projection="X4i",
```
seisman, line 259: Use SI units.
```suggestion
        projection="X10c",
```

pygmt/tests/test_plot.py
```diff
@@ -231,6 +242,29 @@ def test_plot_colors_sizes_proj(data, region):
...
+        frame=True,
+        style="c0.2c",
```
seisman, line 261: Larger circles.
```suggestion
        style="c0.5c",
```

pygmt/tests/test_plot.py
```diff
@@ -95,7 +95,8 @@ def test_plot_fail_no_data(data):
 
 def test_plot_fail_size_color(data):
     """
-    Should raise an exception if array sizes and color are used with matrix.
+    Should raise an exception if array color, sizes and intensity are used with
+    matrix.
     """
     fig = Figure()
```
seisman, line 101: The three tests can be simplified using (I didn't check the codes carefully, please double-check it that it's correct):
```
kwargs = dict(data=data, region=region, projection="X10c", frame="afg")
    fig.plot(style="c0.2c", color=data[:, 2], **kwargs)
    fig.plot(style="cc", sizes=data[:, 2], color="red", **kwargs)
    fig.plot(style="c0.2c", color="red", intensity=data[:, 2], **kwargs)
```

pygmt/src/plot.py
core-man: You are right. I use the example in [Points with varying transparency](https://www.pygmt.org/latest/gallery/symbols/points_transparency.html#sphx-glr-gallery-symbols-points-transparency-py) to test it.
```Python
import numpy as np
import pygmt

x = np.arange(0, 105, 5)
y = np.ones(x.size)
data = np.column_stack([x, y])
transparency = x

fig = pygmt.Figure()
fig.basemap(region=[-5, 105, 0, 2], frame=['xaf+l"Transparency"+u%', "WSrt"],projection="X15c/6c")
fig.plot(data=data, style="c0.6c", color="blue", pen="1p,red", transparency=transparency)
fig.show()
```
trans test

pygmt/tests/test_plot.py
core-man: done in d6ec990.

pygmt/tests/test_plot.py
```diff
@@ -93,30 +93,19 @@ def test_plot_fail_no_data(data):
         )
 
 
-def test_plot_fail_size_color(data):
+def test_plot_fail_color_size_intensity(data):
```
core-man, line 96: I rename this test function (also see https://github.com/GenericMappingTools/pygmt/pull/1065#discussion_r596161372) Do you think it okay?

pygmt/tests/test_plot.py
```diff
@@ -231,6 +220,29 @@ def test_plot_colors_sizes_proj(data, region):
...
+        x=x,
+        y=y,
+        region=[0, 10, 0, 10],
+        projection="X10c",
+        frame=True,
+        style="c0.5c",
+        color="blue",
+        intensity=intensity,
+    )
+    return fig
```
seisman, line 243: What about rewriting the test like this? In the new test, the x values denote the intensity of circles. I feel that it's easier to verify if the baseline image is correct.
```
import numpy as np
import pygmt

x = np.arange(-1, 1.1, 0.1)
y = np.zeros(x.size)
intensity = x 

fig = pygmt.Figure()
fig.plot(
    x=x,
    y=y,
    region=[-1.1, 1.1, -0.5, 0.5],
    projection="X15c/2c",
    frame=["S", "xaf+lIntensity"],
    style="c0.5c",
    color="blue",
    intensity=intensity,
)
fig.show()
```
image

core-man: Done in b3b7734.

seisman (approved):
This PR looks good to me.

meghanrjones (approved):
Nice work!

pygmt/src/plot.py
```diff
@@ -144,11 +144,13 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
...
         reference point for all groups.
     {G}
-    intensity : float or bool
-        Provide an *intens* value (nominally in the -1 to +1 range) to
-        modulate the fill color by simulating illumination [None]. If
-        using ``intensity=True``, we will instead read *intens* from the
-        first data column after the symbol parameters (if given).
+    intensity : float or bool or 1d array
+        Provide an *intensity* value (nominally in the -1 to +1 range) to
+        modulate the fill color by simulating illumination. If using
```
michaelgrund, line 149:
```suggestion
        modulate the color fill by simulating illumination. If using
```

michaelgrund: Not sure, maybe both versions work. Otherwise the PR looks fine to me, great work @core-man.

michaelgrund (approved):

pygmt/src/plot.py
core-man: GMT plot uses fill color, see https://docs.generic-mapping-tools.org/latest/plot.html#i. So let's keep the original version.

Issue #896, opened 2021-02-15
Handling plotting functions that have an inquire (I) only mode


weiji14:
**Description of the issue**

There are a few GMT modules that have a mode for extracting data (statistics or otherwise) via the inquire (I) flag in addition to plotting:

- [`rose`](https://docs.generic-mapping-tools.org/6.1/rose.html#i) -  #794
- [`solar`](https://docs.generic-mapping-tools.org/6.1/solar.html#i) - #804, also has an `-M` mode where no plotting occurs
- [`histogram`](https://docs.generic-mapping-tools.org/6.1/histogram.html#i) - #563

How do we want to handle these cases? It seems weird to use `fig.histogram(..., inquire=...)` if no figure/plot is actually returned. One suggestion made at https://github.com/GenericMappingTools/pygmt/pull/794#pullrequestreview-584985363 was:

> Ideally, what we might have to do is make two user-facing `rose` functions:
> 
>     1. A plotting-only `fig.rose`
> 
>     2. A non-plotting `pygmt.rose` which will solely use the inquire (I) flag.
> 
> Both of these could share a core `def _rose()` function and live within `pygmt/src/rose.py`. This is just an idea though, and probably quite hard to implement in one Pull Request. For now, we could implement just the plotting side (1) and raise a `NotImplementedError` for inquire (I), and perhaps wrap `pygmt.rose` (2) in a future PR? Or maybe there's a smarter alternative way.

Open to suggestions on how we want the API for this look like.

**Are you willing to help implement and maintain this feature?** Discuss first

michaelgrund:
> How do we want to handle these cases? It seems weird to use `fig.histogram(..., inquire=...)` if no figure/plot is actually returned. 

I totally agree that it feels quite strange if no plotting object is returned when applying these modules. 

> One suggestion made at [#794 (review)](https://github.com/GenericMappingTools/pygmt/pull/794#pullrequestreview-584985363) was:
> 
> > Ideally, what we might have to do is make two user-facing `rose` functions:
> > ```
> > 1. A plotting-only `fig.rose`
> > 
> > 2. A non-plotting `pygmt.rose` which will solely use the inquire (I) flag.
> > ```
> > 
> > 
[...]

PR #1072, opened 2021-03-17
Wrap histogram

### LINKED ISSUE #563: Wrap histogram

The **histogram** module is useful and should be wrapped.

Some useful resources:

- [GMT's histogram manpage](https://docs.generic-mapping-tools.org/dev/histogram.html)
- [GMT.jl's histogram wrapper](https://www.generic-mapping-tools.org/GMT.jl/latest/#GMT.histogram)
- [numpy.histogram](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html)
- [matplotlib.pyplot.hist](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.hist.html)

**Are you willing to help implement and maintain this feature?** 
Yes, but may have no time recently, so contributions are welcomed.

willschlitzer:
Wrapping the [histogram](https://docs.generic-mapping-tools.org/6.1/histogram.html) module

**Preview** at https://pygmt-git-wrap-histogram-gmt.vercel.app/api/generated/pygmt.Figure.histogram.html

Fixes #563

**Reminders**

**Slash Commands**


willschlitzer:
I'm not sure how to use the virtualfile to import a table. I'm trying:
`table = [1,1,1,1,1,1,2,2,2,3,4,5,6,7,8,8,8,8,8,8]`
`np_table = np.array(table)`
`fig.histogram(table=np_table, projection="X10c/25c", T=1, G="green")`

But I get `TypeError: histogram() got multiple values for argument 'table'`

I'm not sure how to fix this, as it seems like I'm passing the correct input in (but clearly not).Any ideas @GenericMappingTools/python ?

pygmt/src/histogram.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@fmt_docstring
+@use_alias(
+    J="projection",
+    G="color",
+    R="region",
+)
+@kwargs_to_strings(R="sequence")
+def histogram(table, **kwargs):
```
seisman, line 21: I believe you forgot `self`:
```suggestion
def histogram(self, table, **kwargs):
```

seisman:
I would suggest you only focus on the plotting part in this PR, and implement the inquire (**-I**) feature in a separate PR (after some discussions as in #896).

pygmt/src/histogram.py
willschlitzer: Yep, I copied directly from grdinfo to get the virtual file example and forgot to include self from the other plotting functions. 🤦🏻‍♂️

weiji14:
> I would suggest you only focus on the plotting part in this PR, and implement the inquire (**-I**) feature in a separate PR (after some discussions as in #896).

Just on this point (and as I've mentioned to Jiayuan at https://github.com/GenericMappingTools/pygmt/pull/1145#pullrequestreview-628868487), I'd recommend wrapping as few aliases as you can get away with in this PR. We can open up separate PRs to wrap any complicated parameter like `inquire=I`, or `transparency=t`, and the gallery example/tutorial can be left for another day too.

To be clear, that means you just have to alias the required arguments (https://docs.generic-mapping-tools.org/dev/histogram.html#required-arguments), and the common aliases (-R, -J, -c, -p, -t, etc) to pass. Maybe add one minimal unit test to hit the code coverage target. I see you've done a few aliases already (and you can leave them in if you want), but don't feel like you have to complete all of them.

Example workflow would be to do: Minimal PR (this one) -> Gallery Example PR -> Complete documentation/aliases PR -> Full Tutorial PR

P.S. Yes, we should document this 'policy' about wrapping modules chunk by chunk somewhere, but thought I'd mention it to you first.

willschlitzer:
> > I would suggest you only focus on the plotting part in this PR, and implement the inquire (**-I**) feature in a separate PR (after some discussions as in #896).
> 
> Just on this point (and as I've mentioned to Jiayuan at [#1145 (review)](https://github.com/GenericMappingTools/pygmt/pull/1145#pullrequestreview-628868487)), I'd recommend wrapping as few aliases as you can get away with in this PR. We can open up separate PRs to wrap any complicated parameter like `inquire=I`, or `transparency=t`, and the gallery example/tutorial can be left for another day too.
> 
> To be clear, that means you just have to alias the required arguments (https://docs.generic-mapping-tools.org/dev/histogram.html#required-arguments), and the common aliases (-R, -J, -c, -p, -t, etc) to pass. Maybe add one minimal unit test to hit the code coverage target. I see you've done a few aliases already (and you can leave them in if you want), but don't feel like you have to complete all of them.
> 
> Example workflow would be to do: Minimal PR (this one) -> Gallery Example PR -> Complete documentation/aliases PR -> Full Tutorial PR
> 
> P.S. Yes, we should document this 'policy' about wrapping modules chunk by chunk somewhere, but thought I'd mention it to you first.

I'm on board with with suggested workflow. I think it keeps the pull request more manageable by not including everything under one review,

pygmt/src/histogram.py
willschlitzer: @seisman The use of `self` is causing a style check failure because it isn't used in the function; should I disable pylint for this error (follow-up question: how would I do that?)? The function itself appears to be working correctly, but needs `self` for it to work.

willschlitzer:
Screenshot from 2021-04-09 07-12-21

The parameters list is coming across like this, and I'm not sure how to fix it. As far as I can tell, the docstring format for `histogram` is the same as other functions, but there is some difference that is causing the parameters to list as indented bullet points.

meghanrjones:
> Screenshot from 2021-04-09 07-12-21
> 
> The parameters list is coming across like this, and I'm not sure how to fix it. As far as I can tell, the docstring format for `histogram` is the same as other functions, but there is some difference that is causing the parameters to list as indented bullet points.

This is because of the formatting updates with the latest version of the sphinx_rtd_theme, added in https://github.com/GenericMappingTools/gmt/pull/5072. I will try to fix this as part of #1151.

pygmt/src/histogram.py
```diff
@@ -0,0 +1,63 @@
...
+    {W}
+    {p}
+    horizontal : bool
+        Plot the histogram using horizonal bars instead of the
+        default vertical bars.
+    interval : int or str
+        [*min*\ /*max*\ /]\ *inc*\ [**+n**\ ]
+        Set the interval for the width of each bar in the histogram.
+
+    """
```
weiji14, line 55: Need to add `self.preprocess` to ensure histogram is plotted to the right figure always.

```suggestion
    """
    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
```

pygmt/src/histogram.py
```diff
@@ -0,0 +1,63 @@
...
+    with GMTTempFile() as outfile:
+        with Session() as lib:
+            file_context = lib.virtualfile_from_data(data=table)
+            with file_context as infile:
+                arg_str = " ".join([infile, build_arg_string(kwargs)])
+                lib.call_module("histogram", arg_str)
+        result = outfile.read()
+    return result
```
weiji14, line 63: Err, I don't think there's any data output from `histogram` right now since the inquire (-I) mode doesn't work yet (still need to discuss at #896). So maybe take this out (also need to remove the `GMTTempFile` import and `with` statement above).

```suggestion
```

weiji14 (commented):
Indentation issue seems to be fixed in the live API docs at https://pygmt-git-wrap-histogram-gmt.vercel.app/api/generated/pygmt.Figure.histogram (thanks Meghan). So here comes the comments!

pygmt/src/histogram.py
```diff
@@ -0,0 +1,63 @@
...
+
+@fmt_docstring
+@use_alias(
+    A="horizontal",
+    B="frame",
+    C="cmap",
+    G="fill",
+    J="projection",
+    R="region",
+    T="interval",
```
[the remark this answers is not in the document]
willschlitzer: Works for me!

pygmt/src/histogram.py
```diff
@@ -0,0 +1,63 @@
...
+        Plot the histogram using horizonal bars instead of the
+        default vertical bars.
+    interval : int or str
+        [*min*\ /*max*\ /]\ *inc*\ [**+n**\ ]
+        Set the interval for the width of each bar in the histogram.
+
+    """
+    with GMTTempFile() as outfile:
+        with Session() as lib:
+            file_context = lib.virtualfile_from_data(data=table)
```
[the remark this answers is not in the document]
willschlitzer: @weiji14 Can you explain what this does? I'm committing it, but not sure what this accomplishes.

pygmt/src/histogram.py
willschlitzer: Sounds good, I'll make this change next time I'm on my home computer.

willschlitzer: @weiji14 fixed in d9b8e39

pygmt/src/histogram.py
```diff
@@ -0,0 +1,60 @@
...
+    J="projection",
+    R="region",
+    T="series",
+    W="pen",
+    c="panel",
+    l="label",
+    p="perspective",
+)
+@kwargs_to_strings(R="sequence", T="sequence")
+def histogram(self, table, **kwargs):
```
seisman, line 23: Rename the parameter to `data`, similar to other functions?
```suggestion
def histogram(self, data, **kwargs):
```

pygmt/src/histogram.py
```diff
@@ -0,0 +1,60 @@
...
+
+@fmt_docstring
+@use_alias(
+    A="horizontal",
+    B="frame",
+    C="cmap",
+    G="fill",
+    J="projection",
+    R="region",
+    T="series",
```
seisman, line 16: Is `series` a good name? Other Python histogram functions (e.g., [`numpy.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html), [`plt.histogram`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html) use `bins` instead.

pygmt/src/histogram.py
weiji14: Some functions use data, some use table, I'd actually prefer table since histogram isn't a geographical plot.

pygmt/src/histogram.py
weiji14: See my comment on https://github.com/GenericMappingTools/pygmt/pull/1072#discussion_r611029660. Probably not a good idea to use `bins` as the input for `-T` is structured quite differently to the `bins` argument in `np.histogram` and `plt.histogram`.

seisman: Actually, it depends on how we want to implement the **-T** option in PyGMT.

histogram **-T** have following usages:

1. -T6
2. -T0/200/5
3. -Tfile
4. -T0,3,4,6,10

The current implementation doesn't work well for usage 4, since users have to pass a string "0,3,4,6,10". 

Instead, if we adopt the syntax of `np.histogram`, we can use more Pythonic syntax
```
bins=6
bins=np.arange(0, 200, 5)
bins="file"
bins=[0, 3, 4, 6, 10)
```

weiji14: Currently we're using `@kwargs_to_strings(T="sequence")` which implies Option 2 (-T0/200/5) and I think this works well for `makecpt`/`grd2cpt` where regular sized bins are the norm, and also `histogram` to some extent. Doing Option 4 would mean using `@kwargs_to_strings(T="sequence_comma")` and I guess this would be fine for people wanting irregular sized bins for `histogram`.

I don't often plot histograms, so I'll leave it up to you all to decide which one is better. Ideally we would standardize the `-T` option across all modules, but we could also use `series` (with T="sequence") ` for `makecpt`/`grd2cpt`, and `bins` (with T="sequence_comma")  for `histogram` if that makes more sense for new users.

seisman: Just a note here: we don't have to use `@use_alias` and `@kwargs_to_strings` for **-T**.

Instead, we can check if the `bins` (or any other parameter name) argument is a string, one value, three values, or a list of values. I think it provides the most Pythonic syntax and also the greatest flexibility.

pygmt/src/histogram.py
willschlitzer: For the sake of consistency with the GMT terminology, my vote is it for to remain `table`

pygmt/src/histogram.py
willschlitzer: I'm not looking to shirk responsibility for wrapping `histogram`, but are many geoscientists creating histograms with GMT? I think it's good to provide the capability in PyGMT in case anyone is looking to incorporate it into their figures, but I imagine almost all PyGMT users will be familiar with numpy and matplotlib and using that instead for plots like this, so I don't wish to make the PyGMT syntax overly different than GMT syntax.

weiji14: Good point. For a single histogram plot, it would be be much easier to use something like the built-in [pandas hist function](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.hist.html) (e.g. just do `df.hist()`). But for say, mixing a `grdview` plot with a `histogram` as an 'inset', then using PyGMT would be better :wink:

Let's get another opinion. Ping @core-man since he seemed interested in `histogram` at https://github.com/GenericMappingTools/pygmt/issues/563#issuecomment-801246659. What syntax do you think PyGMT's `histogram` should follow, GMT's or Matplotlib's?

core-man: During my 10-year seismological studies, I only use GMT to plot figures and sometimes use AI/3D max to plot cartoons. I began to use Python and PyGMT at the same time last year, so I have never learned matplotlib. So for my study, GMT/PyGMT is enough right now. For the PyGMT scope I understand now, we'd like to provide a good Python interface for GMT. So as we go to v1.0.0, I prefer more Pythonic syntax instead of strict GMT's. At that time, I think (also hope) most GMT users will learn GMT by using PyGMT.

For this issue, I think the current implementation is okay and most users can use it well (I never use option 4). However, it's better we could refractor all the similar parameters to be more Pythonic syntax as we go to v1.0.0, e.g.,
```
bins=6
bins=np.arange(0, 200, 5)
bins="file"
bins=[0, 3, 4, 6, 10)
```

willschlitzer: I agree that we should move to a more Pythonic syntax overall. Assuming that is the eventual plan, it's creating more work for our future selves by wrapping new features with a GMT syntax. That being said, I think there are more commonly-used features (setting map projections, updating pen/fill settings, adding a frame) that should be prioritized if the plan is to begin incorporating Pythonic syntax. This is certainly an issue outside of the scope of wrapping `histogram`, but I think if we are changing new features to be in-line with a more Pythonic syntax we should be also updating our old features ASAP.

willschlitzer: Maybe I am missing it, but is there a consensus on how this function should be modified?

pygmt/src/histogram.py
willschlitzer: @seisman Do you have an opinion on this? Is it fine if I leave it as `table`?

pygmt/src/histogram.py
weiji14: For the sake of time and consistency with other PyGMT modules like makecpt/grd2cpt, I say we stick with using `series` (-T) instead of `bins`. For better or worse, we are still heavily skewed towards the style of GMT core rather than NumPy, and decisions of what is 'Pythonic' should go in a separate issue. It is better to be consistently like GMT, rather than confuse users with a mix of NumPy-like and GMT command-line-like syntax. So @willschlitzer, let's keep things as it is for now :)

pygmt/src/histogram.py
```diff
@@ -0,0 +1,60 @@
...
+    Plots a histogram, and can read data from a file or
+    list, array, or dataframe.
+
+    Full option list at :gmt-docs:`histogram.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+
```
weiji14, line 34: Can remove one blank line here.
```suggestion
```

weiji14 (commented):
Sorry for the wait @willschlitzer. I'm just about ready to approve this and put the final-review-call label on. If you could please update the branch, and re-create the `histogram` baseline image using GMT 6.2.0rc1 (or we can do it for you if you don't want to upgrade your conda environment yet) and we'll try and get this merged in the next few days, barring any major issues :crossed_fingers:

willschlitzer:
> Sorry for the wait @willschlitzer. I'm just about ready to approve this and put the final-review-call label on. If you could please update the branch, and re-create the `histogram` baseline image using GMT 6.2.0rc1 (or we can do it for you if you don't want to upgrade your conda environment yet) and we'll try and get this merged in the next few days, barring any major issues crossed_fingers

@weiji14 Updated the baseline image with my new GMT 6.2.0rc1 environment (took much longer than I'm proud to admit to find out to use the current working branch of PyGMT in your conda environment; PR coming soon to add that to contributing/install guidelines).

pygmt/src/histogram.py
michaelgrund: Would also prefer table instead of data here.

weiji14 (approved):
Alright, glad you managed to update to GMT 6.2.0rc1, I agree that the upgrade process could be a lot less painful (but that's for a separate issue). Anyways, time for the final review call!

pygmt/src/histogram.py
```diff
@@ -0,0 +1,59 @@
...
+    {G}
+    {W}
+    {c}
+    label : str
+        Add a legend entry for the symbol or line being plotted.
+    {p}
+    horizontal : bool
+        Plot the histogram using horizonal bars instead of the
+        default vertical bars.
+    interval : int or str
```
weiji14, line 49:
```suggestion
    series : int or str or list
```

pygmt/tests/test_histogram.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+from pygmt import Figure
+
+
+@pytest.fixture(scope="module")
+def table():
+    """
+    Returns a list of integers to be used in the histogram.
+    """
+    return [1, 1, 1, 1, 1, 1, 2, 2, 2, 3, 4, 5, 6, 7, 8, 8, 8, 8, 8, 8]
```
seisman, line 14: image

Is the baseline image correct? I expect to see y=6 for x=8.

BTW, please change `projection` to a smaller dimension like `X10c/5c`.

willschlitzer: @seisman I just tried it on both GMT 6.1.1 and 6.2.0rc1, with the figure coming out the same way. Any idea what might be causing the miscount for 8?

Also, changing the projection to `X3c/10c` to give it a larger vertical scale; does that work?

seisman: It looks like an upstream bug.

willschlitzer: I'll try and recreate it in GMT and open an upstream issue

seisman: GMT automatically determines the x range from input data. In this example, xmin=1, xmax=8, but the vertical bar for x=8 is outside of the [1, 8] range, so it's not plotted.

You should be able to see the x=8 bar, if you manually set `region=[0, 9, 0, 6]`. I think we can do it temporarily so that this PR can be merged, and report it to upstream and wait for feedback and potential fix.

seisman: > Also, changing the projection to `X3c/10c` to give it a larger vertical scale; does that work?

BTW, the large vertical scale makes the image difficult to read, why not just use "X10c/10c" for a square basemap?

willschlitzer: @seisman I made your suggested changes

seisman (approved):

seisman:
The "Style Checks" fail but is unrelated to changes in this PR:
```
************* Module pygmt.figure
pygmt/figure.py:82:8: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
```

PR #906, opened 2021-02-16
CI: Run tests in a single CI job (Ubuntu + Python 3.9) for draft PRs

seisman:
**Description of proposed changes**

This PR is inspired by another PR #869 from @willschlitzer.

This PR improves [the "Tests" workflow](https://github.com/GenericMappingTools/pygmt/blob/master/.github/workflows/ci_tests.yaml) to save CI resources. 

Expected behavior (see the comments below for how it actually works):

- For draft PRs: only run tests on Ubuntu + Python 3.9 (~10 minutes)
- When draft PRs are ready for review, run all CI jobs (9 jobs, 10-30 minutes)
- If converting PRs back to draft mode, pushes still only trigger one job (~10 minutes)
- After merged into master, all 9 jobs will run.

References:

1. https://github.com/asyncapi/parser-js/pull/204
2. https://github.community/t/how-to-conditionally-include-exclude-items-in-matrix-eg-based-on-branch/16853/6

Closes #869.

PR #1074, opened 2021-03-18
NEP29: Set minimum required versions to NumPy 1.17+ and Python 3.7+

### LINKED ISSUE #690: Adopt NEP29 - drop support for Python 3.6 and NumPy 1.16

weiji14:

Start adopting NEP29 - Recommend Python and NumPy version support as a community policy standard, see https://numpy.org/neps/nep-0029-deprecation_policy.html. Also modifies the Github Actions CI test matrix to test only minimum (NumPy1.17/Python3.7) and maximum (NumPy1.20/Python3.9) versions only.

Bumps minimum supported NumPy version to 1.17 in the setup.py, requirements.txt and environment.yml files. Also set minimum required Python version to be 3.7+.

**Note**: Will also need to modify the conda-forge pygmt-feedstock recipe to set these minimum required versions, specifically at https://github.com/conda-forge/pygmt-feedstock/blob/823be1f1bf96e6a33725ef7d9c016a0da645b1a9/recipe/meta.yaml#L27-L30

TODO also:

- [ ] Change [branch protection rules](https://github.com/GenericMappingTools/pygmt/settings/branches) to use the NumPy version specifc tests (e.g. 'ubuntu-latest - Python 3.7 / NumPy 1.17 ')
- [x] Document the NEP29 policy somewhere as per https://numpy.org/neps/nep-0029-deprecation_policy.html#implementation

Fixes #690

**Reminders**

**Slash Commands**


.github/workflows/ci_tests.yaml
```diff
@@ -41,20 +41,24 @@ jobs:
...
             isDraft: true
-          - os: ubuntu-latest
-            python-version: 3.7
-            isDraft: true
-          - os: ubuntu-latest
-            python-version: 3.8
-            isDraft: true
+          # - os: ubuntu-latest
+          #   python-version: 3.7
+          #   isDraft: true
```
weiji14, line 46: For some reason, these lines can't be uncommented so that only one workflow (Ubuntu - Py3.9) runs. It fails with `
Error when evaluating 'runs-on' for job 'test'. (Line: 29, Col: 14): Unexpected value ''
` (see e.g. https://github.com/GenericMappingTools/pygmt/actions/runs/663292971), though that Line 29 appears totally unrelated.

```suggestion
          # - os: ubuntu-latest
          #   python-version: 3.7
          #   isDraft: true
```

Might have something to do with the 'include:' lines below? But maybe that's ok, we could just run 2 Linux tests (Python 3.7/NumPy 1.17 and Python 3.9/NumPy 1.20) in draft mode (i.e. +1 extra test compared to #906).

.github/workflows/ci_tests.yaml
```diff
@@ -41,20 +41,24 @@ jobs:
...
+        # Pair Python 3.7 with NumPy 1.17 and Python 3.9 with NumPy 1.20
+        include:
+          - python-version: 3.7
+            numpy-version: '1.17'
+          - python-version: 3.9
+            numpy-version: '1.20'
```
weiji14, line 52: The NumPy versions are paired with Python versions here using 'environment variables' according to https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix.

doc/install.rst
```diff
@@ -79,19 +77,26 @@ We recommend following the instructions further on to install GMT 6.
...
+* `IPython <https://ipython.org>`__: For embedding the figures in Jupyter notebooks.
 
+Note that PyGMT has adopted
+`NEP29 <https://numpy.org/neps/nep-0029-deprecation_policy.html>`__ alongside the rest
+of the Scientific Python ecosystem, and therefore supports:
+
+* All minor versions of Python released 42 months prior to the project,
+  and at minimum the two latest minor versions.
+* All minor versions of NumPy released in the 24 months prior to the project,
+  and at minimum the last three minor versions.
```
weiji14, line 99: Is this the best place to put the NEP29 policy? Or should we restructure the Install page a bit more (e.g. put the 'Using Python' (and 'Using GMT'?) section under 'Dependencies' too).

On the first point, `xarray` placed the NEP29 policy language on their install page (see https://xarray.pydata.org/en/v0.17.0/installing.html#minimum-dependency-versions), but not so sure about other projects.

doc/install.rst
```diff
@@ -46,15 +46,13 @@ Start by looking at the tutorials on our sidebar, good luck!
...
+PyGMT is tested to run on **Python 3.7 or greater**.
 
 We recommend using the `Anaconda <https://www.anaconda.com/distribution>`__
 Python distribution to ensure you have all dependencies installed and the
 `conda <https://docs.conda.io/projects/conda/en/latest/>`__
-package manager is available. Installing Anaconda does not require administrative
-rights to your computer and doesn't interfere with any other Python
-installations on your system.
+package manager is available. Installing Anaconda does not require administrative rights
+to your computer and won't interfere with any other Python installations on your system.
```
seisman, line 55: What's the reason for the changes here? Most lines in `doc/install.rst` are less than 80 characters.

weiji14: Ok, I'll revert the change.

doc/install.rst
seisman: NEP29 was proposed [by maintainers of Matplotlib, scikit-learn, IPython, Jupyter, yt, SciPy, NumPy, and scikit-image](https://numpy.org/neps/nep-0029-deprecation_policy.html#abstract), but it seems none of them list the policy on their sites. Perhaps remove the policy and just change `numpy` to `numpy (>=1.17)`, like [matplotlib](https://matplotlib.org/stable/users/installing.html#dependencies).

Instead, we can mention the policy in the maintenance guides.

.github/workflows/ci_tests.yaml
weiji14: Yeah it's a tricky problem :sweat_smile: I'll leave it as it is and revise the MAINTENANCE.md docs to mention that two jobs are run instead of just one.

doc/install.rst
weiji14: Matplotlib states it at https://matplotlib.org/3.3.4/devel/min_dep_policy.html.

seisman: Yes, but it's in "Developers' Guide", and most users won't read it.

weiji14: True, I'll move the policy out of the installation page then.

MAINTENANCE.md
```diff
@@ -135,6 +137,22 @@ The actual script `package.json` is used by Vercel to install the necessary pack
...
+alongside the rest of the Scientific Python ecosystem, and therefore supports:
+
+* All minor versions of Python released 42 months prior to the project,
+  and at minimum the two latest minor versions.
+* All minor versions of NumPy released in the 24 months prior to the project,
+  and at minimum the last three minor versions.
+
+In `setup.py`, the `python_requires` variable should be set to the minimum
+supported version of Python. Minimum Python and NumPy version support should be
+adjusted upward on every major and minor release, but never on a patch release.
```
weiji14, line 153: I didn't add the line "All supported minor versions of Python should be in the test matrix and have binary artifacts built for the release." as in https://numpy.org/neps/nep-0029-deprecation_policy#implementation because:

1. We're not testing the middle Python version (Python 3.8)
2. We're using no-arch builds on conda-forge and a single [PyGMT Python 3 wheel on PyPI](https://pypi.org/project/pygmt/0.3.1/#files), instead of having specific binary artifacts for each Python version.

README.rst
```diff
@@ -229,14 +229,15 @@ Documentation for other versions
...
+Compatibility with GMT and Python/NumPy versions
+------------------------------------------------
 
-======= ========== =========
-PyGMT   GMT        Python
-======= ========== =========
+======= ========== ========= =========
+PyGMT   GMT        Python    NumPy
+======= ========== ========= =========
+0.4.x   >=6.2.0    >=3.7     >=1.17.0
```
weiji14, line 238: Is it a good idea to add this here now, or should we do it during the v0.4.0 release process instead?

seisman: We _should_ add it when we release v0.4.0, but I'm afraid we may forget it during the release process. I think it's fine to add this here.

MAINTENANCE.md
```diff
@@ -77,8 +78,9 @@ There are 9 configuration files located in `.github/workflows`:
 
    This is run on every commit to the *master* and Pull Request branches.
    It is also scheduled to run daily on the *master* branch.
-   In draft Pull Requests, only one job (Linux + Python latest)
-   is triggered to save on Continuous Integration resources.
+   In draft Pull Requests, only two jobs on Linux (Python 3.7 / NumPy 1.17 and
+   Python 3.9/ NumPy 1.20) are triggered to save on Continuous Integration
+   resources.
```
seisman, line 83: Do you really want to hardcode the version numbers here? It means we have to remember to update this file when we drop support for any old versions. Perhaps we can rephrase it like

"minimum Python / NumPy versions and latest Python / NumPy versions)"

MAINTENANCE.md
seisman: Sounds good to me.

MAINTENANCE.md
weiji14: Done in 1016331f6ae94ec7ac744588dd2b64fdf685a77a.

seisman (approved):
Looks good!

weiji14:
> - [x] Change [branch protection rules](https://github.com/GenericMappingTools/pygmt/settings/branches) to use the NumPy version specifc tests (e.g. 'ubuntu-latest - Python 3.7 / NumPy 1.17 ')

Ok, I've updated the branch protection rules so our test matrix now has 6 instead of 9 tests. Hopefully this decreases the chances of those random Windows failures!

Issue #1091, opened 2021-03-21
Wrap blockmean and blockmode


meghanrjones:
**Description of the desired feature**

Wrap [blockmean](https://docs.generic-mapping-tools.org/dev/blockmean.html) and [blockmode](https://docs.generic-mapping-tools.org/dev/blockmode.html). As @weiji14 mentioned previously, much of blockmedian can be moved to a generic _blockm* function which can be used for all three block average estimation methods.

**Are you willing to help implement and maintain this feature?** Yes- will submit a PR with blockmean for discussion.

meghanrjones:
Assigning this issue to @arleaman in the ESWN/PyGMT workshop

seisman:
The two modules have been wrapped in #1456 and #1092. Closing the issue.

Issue #807, opened 2021-01-27
Should data processing functions have their own modules?


willschlitzer:
In #685, @weiji14 asked the question of if data processing modules like `makecpt` and `grd2cpt` should be split into their own Python scripts, much like the plotting modules will be. My opinion is they should be split up. The other files aren't nearly as big as `base_plotting.py`, but I think it's best to standardize all functions in this way, and it makes it easier to find the function you're looking for (I wasn't sure if `grd2cpt` should be under `mathops.py` or `grdops.py`?).

I'm happy to do it. I'm assuming the process would be similar to the changes in `base_plotting.py`, and just involve copying the functions to their own scripts in `pygmt.src`, but how do we "tell" PyGMT where to look for it beyond the init file, as there isn't a base_plotting-equivalent script to import into?

michaelgrund:
I completely agree with you @willschlitzer. Splitting up all the modules allows to arrange everything in a consistent way. Would also support you in this intention if you need help.

seisman:
> The other files aren't nearly as big as `base_plotting.py`, but I think it's best to standardize all functions in this way, and it makes it easier to find the function you're looking for

I agree with you.

> how do we "tell" PyGMT where to look for it beyond the init file, as there isn't a base_plotting-equivalent script to import into

I think we need to change these lines in `__init__.py`:

https://github.com/GenericMappingTools/pygmt/blob/23454867d4785112281b397a08d7dac109d28f74/pygmt/__init__.py#L14-L24

willschlitzer:
@seisman So would we completely eliminate the files like `mathops.py` and `gridops.py` and just import directly into `pygmt/__init__.py`? If so, how will that affect Sphinx building the documentation?

seisman:
> So would we completely eliminate the files like `mathops.py` and `gridops.py` and just import directly into `pygmt/__init__.py` with something like `from pygmt.src.makecpt import makecpt`?

Yes to me.

> If so, how will that affect Sphinx building the documentation?

I'm not sure. We need to give it a try.

weiji14:
Do we want to get this in for PyGMT v0.3.0 or the next release (v0.3.1 or v0.4.0)?

seisman:
I'm OK to both. I think I can finish PR #832 soon.

weiji14:
Ok, but we should merge #808 before #832. One step at a time ;)

seisman:
These two PRs won't cause conflicts.

[...]

PR #1092, opened 2021-03-21
Wrap blockmean

meghanrjones:

This PR wraps the [blockmean module](https://docs.generic-mapping-tools.org/dev/blockmean.html). Some changes involved in this PR:

- Most of the functionality from the `blockmedian` method is moved to `_blockm`
- `blockmedian` and `blockmean` call `_blockm` with the `block_method` parameter
- `pygmt/src/blockmedian.py` is renamed to `pygmt/src/blockm.py` for simplicity given the similarities between these three methods. I think this is simpler (more pythonic) than having each in separate modules and importing `_blockm` from a fourth module, but feedback is welcome if others feel that #807 should be more strictly adhered to.
- New tests for blockmean are added

Addresses #1091 

**Reminders**

**Slash Commands**


pygmt/src/blockm.py
```diff
@@ -73,6 +134,30 @@ def blockmedian(table, outfile=None, **kwargs):
...
+
+def _blockm(block_method, table, outfile, **kwargs):
+    r"""
+    Block average (x,y,z) data tables by mean or median estimation.
+
+    Reads arbitrarily located (x,y,z) triples [or optionally weighted
+    quadruples (x,y,z,w)] from a table and writes to the output a mean or
+    median (depending on ``block_method``) position and value for every
+    non-empty block in a grid region defined by the ``region`` and ``spacing``
+    parameters.
```
seisman, line 148: Add a "Parameters" section and explain the "block_method" parameter.

meghanrjones: Added in https://github.com/GenericMappingTools/pygmt/pull/1092/commits/62a763bc68a91e2521e9c77db74cd55696335f7a

pygmt/src/blockm.py
```diff
@@ -0,0 +1,215 @@
...
+        *xinc* [and optionally *yinc*] is the grid spacing.
+
+    region : str or list
+        *xmin/xmax/ymin/ymax*\[\ **+r**\][**+u**\ *unit*].
+        Specify the region of interest.
+
+    {V}
+    {a}
+    {f}
+    {r}
```
seisman, line 176: Sorry, I didn't mean to explain all parameters. As this is an internal function, explaining the unique parameter `block_method` is enough. 
```suggestion
```

meghanrjones: I read the comment in email and did not notice the in-line suggestion. The removal was done in https://github.com/GenericMappingTools/pygmt/pull/1092/commits/55a97be3aad14bd177fa56eb972ea2df244cc2a1

pygmt/tests/test_blockmean.py
```diff
@@ -0,0 +1,78 @@
...
+
+    return output
+
+
+def test_blockmean_without_outfile_setting():
+    """
+    Run blockmean by not passing in outfile parameter setting.
+    """
+    with pytest.raises(GMTInvalidInput):
+        blockmean(table="@tut_ship.xyz", spacing="5m", region=[245, 255, 20, 30])
```
weiji14, line 78: One thing I've been contemplating is whether to just give people a `pandas.DataFrame` when no argument is passed into the `outfile` parameter. I feel like the current way is too restrictive, why can't people just pass in a filename (str) input to `table` and get a `pandas.DataFrame` directly? If they set an `outfile`, then return `None` by all means, but I don't think we should force `table=filename` inputs to always go into the `outfile`. What do you think?

seisman: I agree it's too restrictive. The input could be a file, a matrix, a pandas.DataFrame or x/y/z.
The output could be a file or a pandas.DataFrame.

meghanrjones: I also agree. Do you want me to add that to this PR? I would tend towards implementing it separately, since it will impact both blockmean and blockmedian.

seisman: I was trying to improve the blockmean wrapper yesterday and realized that it will cause conflicts with this PR. I think we should improve it in a separate PR, either before or after this PR is merged.

meghanrjones: After seems better to me since this PR is done as far as I can tell and that PR is not open yet. Do you want to work on improving the wrapper? I'm glad to help with adding DataFrame output support, but also fine with leaving it to you in which case I would wait until after you are done to open the PR for wrapping blockmode.

seisman: I'll leave it to you or @weiji14, as I have never used blockm*.

meghanrjones: Sounds good. I'll open an issue to keep track of this proposal.

pygmt/src/blockm.py
```diff
@@ -73,6 +134,35 @@ def blockmedian(table, outfile=None, **kwargs):
         - None if ``outfile`` is set (filtered output will be stored in file
           set by ``outfile``)
     """
+    return _blockm(block_method="blockmedian", table=table, outfile=outfile, **kwargs)
+
+
+def _blockm(block_method, table, outfile, **kwargs):
```
weiji14, line 140: More just a matter of style, but I feel like this `_blockm` function should be placed near the top of the file (before the actual `blockmean` and `blockmedian` functions).

weiji14 (commented):
Just one very minor point on moving `_blockm` to the top of the file, otherwise well done!

pygmt/src/blockm.py
meghanrjones: OK, done in https://github.com/GenericMappingTools/pygmt/pull/1092/commits/f34e77842ab870e3124cf8b15df8a1080ab946f5

weiji14 (approved):
Thanks Meghan! Really appreciate having another data processing module added to PyGMT's toolkit :smile:

seisman (approved):
Looks good to me.

PR #1096, opened 2021-03-22
Migrate Figure.basemap tests to dvc

seisman:

- Remove three tests that only test the aliases
- Rewrite two tests using mpl_image_compare
- Rewrite tests using long-form parameters, SI units and small figure dimensions
- Migrate images to dvc

It would be easier to review each commit separately.

Address #1131.

**Reminders**

**Slash Commands**


willschlitzer:
@seisman Do you want us to begin updating the current tests to use `dvc`? I spoke to @weiji14 about it and his guidance was to wait until GMT 6.2 has been released to avoid having to update the reference images.

seisman:
> @seisman Do you want us to begin updating the current tests to use `dvc`? I spoke to @weiji14 about it and his guidance was to wait until GMT 6.2 has been released to avoid having to update the reference images.

See my comment https://github.com/GenericMappingTools/pygmt/issues/963#issuecomment-802376769:

> Do we want to do the migration before GMT 6.2.0 or after? I prefer to do the migration before v6.2.0, although it means more work for us. After bumping GMT to 6.2.0, most tests will fail due to the changes in GMT 6.2.0, but I feel it's also a good opportunity for us to learn the GMT changes and find potential bugs by comparing the images generated by GMT 6.1.1 and 6.2.0.
> 
> FYI, one week ago, I built the PyGMT documentation using GMT 6.2.0, and found several issues with the GMT dev version (GenericMappingTools/gmt#4955), and they were all fixed in less than one week!

pygmt/tests/test_basemap.py
```diff
@@ -59,69 +46,28 @@ def test_basemap_power_axis():
...
-def test_basemap_map_scale():
-    """
-    Create a map with coast and use basemap to add a map scale.
-    """
-    fig_ref, fig_test = Figure(), Figure()
-    fig_ref.coast(R="127.5/128.5/26/27", W="1/0.5p")
-    fig_ref.basemap(L="jMB+c1+w10k+l+f")
-    fig_test.coast(region=[127.5, 128.5, 26, 27], shorelines="1/0.5p")
-    fig_test.basemap(map_scale="jMB+c1+w10k+f+l")
-    return fig_ref, fig_test
```
weiji14, line 127: You sure about removing these 3 tests? I know they're just aliases, but rose/compass/map_scale seems like a key part of many maps that we should test.

seisman: The "contributing guides" were written based on the existing tests at that time, so I don't think we should take it as a strict rule. 

Of course, I can add these tests if you think they're necessary.

weiji14: I would add the tests back, but maybe take away the `fig.coast` part and make it purely `fig.basemap` if possible.

weiji14 (approved):
/format

seisman:
So typing `/format` in the review box doesn't work.

PR #1100, opened 2021-03-22
Fix documentation for alias 'aspatial'

meghanrjones:

This PR updates an error that @PaulWessel found in the ReST docs for GMT in [this PR](https://github.com/GenericMappingTools/gmt/pull/4796), in that the *cols*=*name* option for 'aspatial' is optional.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -64,7 +64,7 @@ COMMON_OPTIONS = {
          """,
     "a": r"""
         aspatial : str
-            [*col*\ =]\ *name*\ [,...].
+            [[*col*\ =]\ *name*\ [,...]].
             Control how aspatial data are handled during input and output.
             Full documentation is at :gmt-docs:`gmt.html#aspatial-full`.
          """,
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -64,7 +64,7 @@ COMMON_OPTIONS = {
          """,
     "a": r"""
         aspatial : bool or str
-            [[*col*\ =]\ *name*\ [,...]].
+            [*col*\ =]\ *name*\ [,...].
             Control how aspatial data are handled during input and output.
             Full documentation is at :gmt-docs:`gmt.html#aspatial-full`.
          """,
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -63,7 +63,7 @@ COMMON_OPTIONS = {
             :gmt-docs:`gmt.html#xy-full`.
          """,
     "a": r"""
-        aspatial : str
+        aspatial : bool or str
             [[*col*\ =]\ *name*\ [,...]].
             Control how aspatial data are handled during input and output.
             Full documentation is at :gmt-docs:`gmt.html#aspatial-full`.
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -112,14 +112,14 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
         Draw symmetrical error bars. Full documentation is at
         :gmt-docs:`plot.html#e`.
     connection : str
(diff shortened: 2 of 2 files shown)


meghanrjones:
Changed to draft because the new documentation for ``Figure.plot()`` does not work as expected. I'll update when it's ready for review.

pygmt/helpers/decorators.py
```diff
@@ -64,7 +64,7 @@
          """,
     "a": r"""
         aspatial : str
-            [*col*\ =]\ *name*\ [,...].
+            [[*col*\ =]\ *name*\ [,...]].
```
seisman, line 67: Does it mean that we can use `aspatial=True` without giving string type arguments?

meghanrjones: I'll check

meghanrjones: Yes, `aspatial=True` does work. I will update accordingly. An example:
```
import pygmt
pygmt.info(table='points.txt',aspatial=True)
```
where points.txt contains:
```
# @VGMT1.0 @GPOINT @Nname|depth|id
# @Tstring|double|integer
# @R178.43/178.5/-57.98/-34.5
# @Je4326
# @Jp"+proj=longlat +ellps=WGS84 +datum=WGS84+no_defs"
# FEATURE_DATA
# @D"point 1"|-34.5|1
178.5 -45.7
# @D"Point 2"|-57.98|2
178.43 -46.8
```

PaulWessel: I am looking at a case in GMT core right now where -a (no args, the equivalent of your True) only prints out the aspatial values and not the coordinates.  The intent is that it should work (and it used to).

meghanrjones: Yes the output currently includes both spatial and spatial info, as you noted in the other PR. So, a bug.
`'Aspatial columns: name[string] depth[double] id[integer]\npoints.txt: N = 2 <178.43/178.5> <-46.8/-45.7> <-57.98/-34.5> <1/2>\n'`

meghanrjones: I added that `aspatial` can handle `bool` as an input, even though it doesn't work quite right because it is still an option.

pygmt/helpers/decorators.py
```diff
@@ -63,8 +63,8 @@
             :gmt-docs:`gmt.html#xy-full`.
          """,
     "a": r"""
-        aspatial : str
-            [*col*\ =]\ *name*\ [,...].
+        aspatial : bool or str
+            [[*col*\ =]\ *name*\ [,...]].
```
seisman, line 67: I think we should revert the changes. `[ ]` means the whole argument is optional, so it's equivalent to `aspatial=True` or `aspatial=""`. 
```suggestion
            [*col*\ =]\ *name*\ [,...].
```

meghanrjones: I do not understand. The outer brackets are what indicate that `aspatial=""` is allowed, so I do not see why they should be removed. It does not seem to me that `bool` alone is sufficient, because I would think that this
```
aspatial : bool or str
    [*col*\ =]\ *name*\ [,...].
```
would mean that ``aspatial=True`` is allowed, but if a str is passed it must contain arguments (i.e., `aspatial=""` is not allowed).

seisman: Both `aspatial=""` and `aspatial=True` are converted to GMT arguments `-a` internally, so they're the same. Similarly, `frame=""` is also allowed, but we use `frame=True` in all the examples, perhaps `frame=True` is more readable?

meghanrjones: OK, I will revert the change to be consistent with the other parts of the PyGMT documentation, even though it still seems to me that including the brackets is more technically correct because while `parameter=""` and `parameter=True` internally are the same, booleans and strings are not the same to the external user.

seisman (approved):

PR #794, opened 2021-01-17
Wrap rose

michaelgrund:
**Description of proposed changes**

As discussed in #786 this is wrapping the GMT function **rose**. Would be happy to receive any kind of feedback.

**Reminders**

 * [x]  Run `make format` and `make check` to make sure the code follows the style guide.
 * [x]  Add tests for new features or tests that would have caught the bug that you're fixing.
 * [x]  Add new public functions/methods/classes to `doc/api/index.rst`.
 * [x]  Write detailed docstrings for all functions/methods.
 * [x]  If adding new functionality, add an example to docstrings or tutorials.

**Notes**

* You can write `/format` in the first line of a comment to lint the code automatically

PR #1101, opened 2021-03-22
Add new function to load fractures sample data

michaelgrund:

As suggested in #794 here's the function implementation to load the fractures_06.txt sample data.

**Reminders**

**Slash Commands**


meghanrjones:
Are you sure that fracture_06.txt includes lengths? It seems too coincidental that column 0 increments by one (viewable here: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/cache/fractures_06.txt)

meghanrjones:
> Are you sure that fracture_06.txt includes lengths? It seems too coincidental that column 0 increments by one (viewable here: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/cache/fractures_06.txt)

I noticed that this is copied from GMT's example 6. @PaulWessel, do the lengths in @fracture_06.txt represent real fractures or should we specify that the lengths are synthetic and the directions are digitized from maps?

michaelgrund:
> 
> 
> Are you sure that fracture_06.txt includes lengths? It seems too coincidental that column 0 increments by one (viewable here: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/cache/fractures_06.txt)

If I remember correctly the lengths are located in column 1 and directions in column 0 (therefore 0 to 360°).

meghanrjones:
> > Are you sure that fracture_06.txt includes lengths? It seems too coincidental that column 0 increments by one (viewable here: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/cache/fractures_06.txt)
> 
> If I remember correctly the lengths are located in column 1 and directions in column 0 (therefore 0 to 360°).



> > Are you sure that fracture_06.txt includes lengths? It seems too coincidental that column 0 increments by one (viewable here: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/cache/fractures_06.txt)
> 
> If I remember correctly the lengths are located in column 1 and directions in column 0 (therefore 0 to 360°). That's also the reason why to use `columns=[1, 0]` in the rose gallery example. Otherwise I cannot reproduce the figure from the GMT docs (https://docs.generic-mapping-tools.org/latest/gallery/ex06.html). Maybe @PaulWessel can shed some light on this.

Yes, the order makes sense since the `gmt rose` example uses `-:` to switch the column order. The directions being evenly incremented still seems a bit odd to me unless the 'compilation' refers to a binned data set.

PaulWessel:
The data set is fake I think but I cannot recall for sure.  This takes us back to 1990.  While some guys give roses to their love-interests, I wrote **psrose** for my future wife to impress her (she needed to make these plots for her thesis).  Even nerds know that women likes flowers.

michaelgrund:
> 
> 
> The data set is fake I think but I cannot recall for sure. This takes us back to 1990. While some guys give roses to their love-interests, I wrote **psrose** for my future wife to impress her (she needed to make these plots for her thesis). Even nerds know that women likes flowers.

Funny to hear some things from the very beginning of **psrose** @PaulWessel, especially since the corresponding example was the first I was faced with when I started using GMT in 2011 😄.

meghanrjones:
> The data set is fake I think but I cannot recall for sure. This takes us back to 1990. While some guys give roses to their love-interests, I wrote **psrose** for my future wife to impress her (she needed to make these plots for her thesis). Even nerds know that women likes flowers.

I'm glad to have asked the question to learn this story about the origins of the `psrose` module. My spouse and I have given up on flowers due to allergies - perhaps next anniversary we should just use GMT instead :smile:

@michaelgrund - up to you about whether it needs to be specified that the data may be made up. You could add something like 'hypothetically digitized'

meghanrjones (commented):
Can you add a test to 'pygmt/tests/test_datasets_tutorial.py`?

michaelgrund:
> 
> 
> > The data set is fake I think but I cannot recall for sure. This takes us back to 1990. While some guys give roses to their love-interests, I wrote **psrose** for my future wife to impress her (she needed to make these plots for her thesis). Even nerds know that women likes flowers.
> 
> I'm glad to have asked the question to learn this story about the origins of the `psrose` module. My spouse and I have given up on flowers due to allergies - perhaps next anniversary we should just use GMT instead 😄
> 
> @michaelgrund - up to you about whether it needs to be specified that the data may be made up. You could add something like 'hypothetically digitized'

Adjusted 😉.

michaelgrund:
> 
> 
> Can you add a test to 'pygmt/tests/test_datasets_tutorial.py`?

Done! Thanks for mentioning @meghanrjones.

seisman:
> If I remember correctly the lengths are located in column 1 and directions in column 0 (therefore 0 to 360°). That's also the reason why to use `columns=[1, 0]` in the rose gallery example. Otherwise I cannot reproduce the figure from the GMT docs ([docs.generic-mapping-tools.org/latest/gallery/ex06.html](https://docs.generic-mapping-tools.org/latest/gallery/ex06.html)).

If I understand it correctly, **rose** expects *length* and *azimuth*, but the data gives *azimuth* and *length*. Do you think we should swap the column order returned by the function?

PaulWessel:
From man page: rose reads (length, azimuth) pairs from file [or standard input] and plot a windrose diagram. Add -i0 if your file only has azimuth values

This is a bit unfortunate as it would make lots more sense that _azimuth_ would be the first column since azimuths are always required and lengths are optional.  I am not sure if there is a way to fix this in a backwards compatible way though.

michaelgrund:
> 
> 
> > If I remember correctly the lengths are located in column 1 and directions in column 0 (therefore 0 to 360°). That's also the reason why to use `columns=[1, 0]` in the rose gallery example. Otherwise I cannot reproduce the figure from the GMT docs ([docs.generic-mapping-tools.org/latest/gallery/ex06.html](https://docs.generic-mapping-tools.org/latest/gallery/ex06.html)).
> 
> If I understand it correctly, **rose** expects _length_ and _azimuth_, but the data gives _azimuth_ and _length_. Do you think we should swap the column order returned by the function?

Would be fine for me, however, for advanced GMT users which are familiar with this "old" example there may be some confusion.



> 
> 

> 
> This is a bit unfortunate as it would make lots more sense that _azimuth_ would be the first column since azimuths are always required and lengths are optional. I am not sure if there is a way to fix this in a backwards compatible way though.

It would be easy to fix that for PyGMT only, however, I would prefer to ensure consistency with GMT.  Any comments on that?

PaulWessel:
At least for GMT core we need to maintain backwards compatibility.  Here are the options we have I think:

1. Add a new option  to **psrose** that explicitly selects azimuth or azimuth, length input.  Unfortunately, most letters are already used, leaving perhaps only **-H**. With this solution, a missing -H would mean the old format, hence backwards compatible.  It would be listed as a required option, except we would not enforce the requirement so that backwardness would work.
2. Add a new module **pscircular** that uses the input order we wish to have.  **pscircular** is just a few lines of C code setting up the input and then calling the module **psrose** via _GMT_Call_Module_.  **psrose** would remain as a module but no longer documented, hence not breaking anyone's old scripts.

I am not sure if the problem is severe enought to consider those options though.  With **pscircular** we get a fresh page and can redesign those messy options into something better since backwards compatibility is no longer an issue (as long as we can construct **psrose** commands from them)..

pygmt/datasets/tutorial.py
```diff
@@ -101,3 +101,25 @@ def load_usgs_quakes():
...
+    the cache. So you'll need an internet connection the first time around.
+
+    Returns
+    -------
+    data : pandas.Dataframe
+        The data table. Use ``print(data.describe())`` to see the available
+        columns.
+    """
+    fname = which("@fractures_06.txt", download="c")
+    data = pd.read_csv(fname)
```
meghanrjones, line 124:
```suggestion
data = pd.read_csv(fname, header=None, sep=r"\s+", names=["direction","length"])
```
Test fails, probably related to the separator. I think the final size will be 361,2 because the direction goes 0-360

meghanrjones:
> At least for GMT core we need to maintain backwards compatibility. Here are the options we have I think:
> 
>     1. Add a new option  to **psrose** that explicitly selects azimuth or azimuth, length input.  Unfortunately, most letters are already used, leaving perhaps only **-H**. With this solution, a missing -H would mean the old format, hence backwards compatible.  It would be listed as a required option, except we would not enforce the requirement so that backwardness would work.
> 
>     2. Add a new module **pscircular** that uses the input order we wish to have.  **pscircular** is just a few lines of C code setting up the input and then calling the module **psrose** via _GMT_Call_Module_.  **psrose** would remain as a module but no longer documented, hence not breaking anyone's old scripts.
> 
> 
> I am not sure if the problem is severe enought to consider those options though. With **pscircular** we get a fresh page and can redesign those messy options into something better since backwards compatibility is no longer an issue (as long as we can construct **psrose** commands from them)..

I don't think any core gmt changes are needed here. I think @seisman was just suggesting that the PyGMT function to load in the fractures data could switch the columns, so that **-:** isn't needed.

seisman:
> > At least for GMT core we need to maintain backwards compatibility. Here are the options we have I think:
> > ```
> > 1. Add a new option  to **psrose** that explicitly selects azimuth or azimuth, length input.  Unfortunately, most letters are already used, leaving perhaps only **-H**. With this solution, a missing -H would mean the old format, hence backwards compatible.  It would be listed as a required option, except we would not enforce the requirement so that backwardness would work.
> > 
> > 2. Add a new module **pscircular** that uses the input order we wish to have.  **pscircular** is just a few lines of C code setting up the input and then calling the module **psrose** via _GMT_Call_Module_.  **psrose** would remain as a module but no longer documented, hence not breaking anyone's old scripts.
> > ```
> > 
> > 
> > I am not sure if the problem is severe enought to consider those options though. With **pscircular** we get a fresh page and can redesign those messy options into something better since backwards compatibility is no longer an issue (as long as we can construct **psrose** commands from them)..
> 
> I don't think any core gmt changes are needed here. I think @seisman was just suggesting that the PyGMT function to load in the fractures data could switch the columns, so that **-:** isn't needed.

Yes, I agree. I think the recent suggestions (https://github.com/GenericMappingTools/pygmt/pull/1101#discussion_r599005559) made by @meghanrjones can help remove **-:** or **-i1,0**.

pygmt/datasets/tutorial.py
seisman: Please double check whether it's direction (CCW from East) or azimuth (CW from North).

pygmt/tests/test_datasets_tutorial.py
```diff
@@ -58,3 +59,11 @@ def test_usgs_quakes():
     """
     data = load_usgs_quakes()
     assert data.shape == (1197, 22)
+
+
+def test_fractures_compilation():
+    """
+    Check that the dataset loads without errors.
```
weiji14, line 66:
```suggestion
    Check that the @fractures_06.txt dataset loads without errors.
```

pygmt/datasets/tutorial.py
```diff
@@ -101,3 +101,25 @@ def load_usgs_quakes():
...
+    return data
```
weiji14, line 125: In addition to Meghan's suggestion at https://github.com/GenericMappingTools/pygmt/pull/1101/files#r599005559, we can re-order the columns as follows:

```suggestion
    return data[["length", "direction"]]
```

pygmt/tests/test_datasets_tutorial.py
```diff
@@ -58,3 +59,11 @@ def test_usgs_quakes():
...
+    """
+    data = load_fractures_compilation()
+    assert data.shape == (360, 2)
```
weiji14, line 69: Would be nice to add a few more assertion tests like so:

```suggestion
    assert data.shape == (360, 2)
    summary = data.describe()
    assert summary.loc["min", "length"] == 98.6561
    assert summary.loc["max", "length"] == 984.652
    assert summary.loc["min", "direction"] == 0.0
    assert summary.loc["max", "direction"] == 360.0

```

seisman:
A little off-topic: Should we rename the `datasets/tutorial.py` to `datasets/samples.py` or similar?

pygmt/datasets/tutorial.py
seisman: Actually, in the gallery example in #794, I would recommend using 
```
Figure.rose(length=data.length, azimuth=data.direction, ...)
```
rather than
```
Figure.rose(data=data, ...)
```

pygmt/datasets/tutorial.py
michaelgrund: > 
> 
> Test fails, probably related to the separator. I think the final size will be 361,2 because the direction goes 0-360

You're right @meghanrjones , totally overlooked the separator :sweat_smile: . Thanks!

pygmt/tests/test_datasets_tutorial.py
michaelgrund: Thanks for the suggestions @weiji14.

michaelgrund:
> 
> 
> A little off-topic: Should we rename the `datasets/tutorial.py` to `datasets/samples.py` or similar?

I believe `datasets/samples.py` would make more sense in this context.

weiji14:
> > A little off-topic: Should we rename the `datasets/tutorial.py` to `datasets/samples.py` or similar?
> 
> I believe `datasets/samples.py` would make more sense in this context.

Ok, then the test file also needs to be renamed from 'test_datasets_tutorial.py' to 'test_datasets_samples.py'.

P.S. Don't forget to add `load_fractures_compilation` to `doc/api/index.rst`.

michaelgrund:
> 
> 
> > > A little off-topic: Should we rename the `datasets/tutorial.py` to `datasets/samples.py` or similar?
> > 
> > 
> > I believe `datasets/samples.py` would make more sense in this context.
> 
> Ok, then the test file also needs to be renamed from 'test_datasets_tutorial.py' to 'test_datasets_samples.py'.
> 
> P.S. Don't forget to add `load_fractures_compilation` to `doc/api/index.rst`.

Should we handle this in a separate PR?

weiji14:
Yes, probably better to rename things in a separate PR.

michaelgrund:
> 
> 
> Yes, probably better to rename things in a separate PR.

Opened #1110 for this.

weiji14 (approved):
This looks good enough to me, thanks @michaelgrund!

pygmt/datasets/samples.py
```diff
@@ -101,3 +101,25 @@ def load_usgs_quakes():
     fname = which("@usgs_quakes_22.txt", download="c")
     data = pd.read_csv(fname)
     return data
+
+
+def load_fractures_compilation():
+    """
+    Load a table of fracture lengths and directions as hypothetically digitized
+    from geological maps as a pandas.Dataframe.
```
seisman, line 109:
```suggestion
    from geological maps as a pandas.DataFrame.
```

pygmt/datasets/samples.py
```diff
@@ -101,3 +101,25 @@ def load_usgs_quakes():
...
+
+    This is the ``@fractures_06.txt`` dataset used in the GMT tutorials.
+
+    The data are downloaded to a cache directory (usually ``~/.gmt/cache``) the
+    first time you invoke this function. Afterwards, it will load the data from
+    the cache. So you'll need an internet connection the first time around.
+
+    Returns
+    -------
+    data : pandas.Dataframe
```
seisman, line 119:
```suggestion
    data : pandas.DataFrame
```

pygmt/datasets/samples.py
```diff
@@ -101,3 +101,25 @@ def load_usgs_quakes():
...
+        The data table. Use ``print(data.describe())`` to see the available
+        columns.
+    """
+    fname = which("@fractures_06.txt", download="c")
+    data = pd.read_csv(fname, header=None, sep=r"\s+", names=["direction", "length"])
+    return data[["length", "direction"]]
```
seisman, line 125: I still think we should name the two columns by *azimuth* and *length*, not *direction* and *length*.

michaelgrund: Yes, you're right @seisman, will change it.

seisman (approved):

PR #1109, opened 2021-03-23
Allow passing an array as intensity for plot3d

core-man:

Currently, we can have two ways to specify an intensity/intensities using the [plot3d](https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot3d.html) method.
- Simply give a constant intensity like `intensity=0.5`, which works for both `x`/`y`/`z` and `data` input. This is similar to `-I0.5` in the GMT command line.
- Set `intensity=True`, and read intensities from an *input file* or from a *2D array*, which is only for the `data` parameter. This is similar to `-I` in the GMT command line.

So, when we want to change the intensities for different points, we can only use the `data` parameter now. This PR tries to pass an array to `intensity` so that we can set different intensities when we use `x`/`y`/`z` as point input. This way will not work for `data` input.

Related to #891 

**Reminders**

**Slash Commands**


pygmt/src/plot3d.py
```diff
@@ -111,11 +111,14 @@ def plot3d(
...
-        modulate the fill color by simulating illumination [Default is None].
-        If using ``intensity=True``, we will instead read *intens* from the
-        first data column after the symbol parameters (if given).
+    intensity : float or bool or 1d array
+        Provide an *intensity* value (nominally in the -1 to +1 range) to
+        modulate the fill color by simulating illumination. If using
+        ``intensity=True``, we will instead read *intensity* from the first
+        data column after the symbol parameters (if given). *intensity* can
+        also be a 1d array to set varying intensity for symbols, but it is only
+        valid for ``x``/``y`` pairs.
```
seisman, line 120:
```suggestion
        valid for ``x``/``y``/``z``.
```

pygmt/tests/test_plot3d.py
```diff
@@ -329,6 +318,34 @@ def test_plot3d_colors_sizes_proj(data, region):
...
+    z = y
+    intensity = x
+
+    fig = Figure()
+    fig.plot3d(
+        x=x,
+        y=y,
+        z=z,
+        region=[-1.1, 1.1, -0.5, 0.5, -0.5, 0.5],
+        projection="x15c",
```
seisman, line 338: Any particular reason that you want to use `x15c` (not `X15c`) here?
```suggestion
        projection="x15c",
```

core-man: Revised in 3c12707.

pygmt/tests/test_plot3d.py
```diff
@@ -329,6 +318,34 @@ def test_plot3d_colors_sizes_proj(data, region):
...
+    z = y
+    intensity = x
+
+    fig = Figure()
+    fig.plot3d(
+        x=x,
+        y=y,
+        z=z,
+        region=[-1.1, 1.1, -0.5, 0.5, -0.5, 0.5],
+        projection="X15c",
```
seisman, line 338:
```suggestion
        projection="X15c/5c",
```

pygmt/tests/test_plot3d.py
```diff
@@ -329,6 +318,34 @@ def test_plot3d_colors_sizes_proj(data, region):
...
+        zscale=0.1,
```
seisman, line 339:
```suggestion
        zsize="5c",
```

seisman (approved):

pygmt/tests/test_plot3d.py
```diff
@@ -329,6 +318,34 @@ def test_plot3d_colors_sizes_proj(data, region):
     return fig_ref, fig_test
 
 
+@pytest.mark.mpl_image_compare
+def test_plot3d_varying_intensity():
+    """
+    Plot the data with array-like intensity.
+    """
+
```
seisman, line 326:
```suggestion
    """
```

pygmt/tests/test_plot3d.py
```diff
@@ -329,6 +318,34 @@ def test_plot3d_colors_sizes_proj(data, region):
...
+    x = np.arange(-1, 1.1, 0.1)
```
seisman, line 327:
```suggestion
    """
    x = np.arange(-1, 1.1, 0.1)
```

PR #804, opened 2021-01-23
Wrap solar

willschlitzer:
**Description of proposed changes**

Wrap the GMT module `solar`.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [x] If adding new functionality, add an example to docstrings or tutorials.

**Notes**

- You can write `/format` in the first line of a comment to lint the code automatically

PR #1116, opened 2021-03-24
Let Figure.savefig support filenames with spaces

seisman:

`Figure.savefig()` doesn't work if filenames contain spaces.

Here is an minimal example to reproduce the issue:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(projection="X10c/10c", frame=True, region=[0, 10, 0, 10])
fig.savefig("map 2.png")
```

This PR fixes the issue by adding double quotes around filenames. 

**NOTE:** Using filenames with spaces is NOT recommended. 
However, we have to support it, because sometimes filenames of some 
baseline images must contain spaces. 
See https://github.com/GenericMappingTools/pygmt/runs/2179626599 for an example failure.

**Reminders**

**Slash Commands**


weiji14:
> **NOTE:** Using filenames with spaces is NOT recommended.
> However, we have to support it, because sometimes filenames of some
> baseline images must contain spaces.

See my comment at https://github.com/GenericMappingTools/pygmt/pull/804#discussion_r600953519 on how to fix the space issue. If the tests pass, do we still need this PR?

seisman:
Nice fix in https://github.com/GenericMappingTools/pygmt/pull/804#discussion_r600968043.

I think this PR is still needed for people who want to use spaces. It's allowed in GMT (https://docs.generic-mapping-tools.org/dev/begin.html#examples).

pygmt/figure.py
```diff
@@ -170,6 +170,9 @@ def psconvert(self, **kwargs):
         # Default cropping the figure to True
         if "A" not in kwargs:
             kwargs["A"] = ""
+        # allow for spaces in figure name
+        if kwargs.get("F"):
+            kwargs["F"] = f'"{kwargs.get("F")}"'
```
weiji14, line 175: > I think this PR is still needed for people who want to use spaces. It's allowed in GMT (https://docs.generic-mapping-tools.org/dev/begin.html#examples).

Yes, but this is just a patchy solution to the main underlying problem which is #247. I mitigated it in 4126c16160c67f8d2c405e7d5da56145adc989d4 for subplot's `title` parameter too, but I wonder if it's time we have a real, standardized solution that works to fix #247 entirely:

```suggestion
        kwargs["F"] = f'"{kwargs.get("F")}"' if kwargs.get("F") else None
```

seisman: subprocess provides a function [`list2cmdline`](https://github.com/python/cpython/blob/3.9/Lib/subprocess.py#L533) which may be useful. I think I tried it once. It works but may break backward-compatibility.

weiji14: Oo, that looks promising. How does it work, and why do you think it might break backward-compatibility?

seisman: I can't find my old testing codes. If I remember it correctly.

Here is the current behavior:
```
# 1st way: FAIL
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B='x20+lCrust Age')
# 2nd way: SUCCEED
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B='x20+l"Crust Age"')
# 3rd way: FAIL
fig.basemap(R=[0, 100, 0, 50], J="X10c/5c", B="x20+l'Crust Age'")
```
After using `list2cmdline`, all the three options work, but options 2 and 3 will have single or double quotes in the X labels.

weiji14: I see, but maybe we can leave the complicated frame/B option as it is now and only use `list2cmdline` for less complicated parameters, like `fig.text(text=...)` and `fig.savefig(fname=...)`? That itself would be a step in the right direction.

seisman: We need more investigation on the function and how it can be useful in our project.

seisman:
@weiji14 Any other concerns?

seisman:
Ping @GenericMappingTools/pygmt-contributors for comments if we should allow whitespace in figure names.

core-man:
> Ping @GenericMappingTools/pygmt-contributors for comments if we should allow whitespace in figure names.

I think so and this PR is simple enough.

pygmt/tests/test_figure.py
```diff
@@ -96,6 +96,18 @@ def test_figure_savefig_transparent():
...
+def test_figure_savefig_filename_with_spaces():
+    """
+    Check if savefig (or psconvert) supports filenames with spaces.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
+    fname = "test_figure_savefig filename with spaces.png"
+    fig.savefig(fname)
+    assert os.path.exists(fname)
+    os.remove(fname)
```
weiji14, line 108: Probably better to use GMTTempfile if possible here.

weiji14 (commented):
I'm still reluctant to allow spaces since it adds an if-clause to the code, but ok with supporting it if three or more people agree.

willschlitzer:
> Ping @GenericMappingTools/pygmt-contributors for comments if we should allow whitespace in figure names.

I think it is good to allow this. I'm glad there is the disclaimer that spaces in filenames isn't recommended, but better to increase compatibility.

pygmt/tests/test_figure.py
seisman: Done in 4d86e7816cce4b6cf297278e56f20228ab976c4e.

meghanrjones (approved):

Issue #1098, opened 2021-03-22
The transparency parameter of plot/plot3d methods can be 1d array


core-man:
**Description of the problem**

We need to update the documentation for the ` transparency` parameter in [plot](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot) and [plot3d](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot3d.html#pygmt.Figure.plot3d) because `transparency` can also be `1d array`.

---

See the description  of the `transparency` parameter of [plot](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot):
> **transparency**  (int or float) – Set transparency level, in [0-100] percent range. Default is 0, i.e., opaque. Only visible when PDF or raster format output is selected. Only the PNG format selection adds a transparency layer in the image (for further processing). transparency can also be a 1d array to set varying transparency for symbols.

See the description of the `transparency` parameter of [plot3d](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot3d.html#pygmt.Figure.plot3d):
> **transparency** (int or float) – Set transparency level, in [0-100] percent range. Default is 0, i.e., opaque. Only visible when PDF or raster format output is selected. Only the PNG format selection adds a transparency layer in the image (for further processing). transparency can also be a 1d array to set varying transparency for symbols.

We can see that `transparency` can be 1d array at the end of the above description, but it is missing at the beginning. See the example of [Points with varying transparency](https://www.pygmt.org/latest/gallery/symbols/points_transparency.html#sphx-glr-gallery-symbols-points-transparency-py) using a 1d array as `transparency`.

---

Meanwhile, array-like `transparency` doesn't work for the `data` parameter. It is only valid for `x`/`y` like the `sizes` parameter. (originally posted in https://github.com/GenericMappingTools/pygmt/pull/1065#discussion_r598100163).

A `GMTInvalidInput` is raised for an array-like `transparency` when the `data` parameter is used in #1065. We may add a test for this kind of invalid input. A test called `test_plot_fail_color_size_intensity` is updated in #1065 for 1d-array `color`, `sizes`, and `intensity` when `data` parameter is used. It should be easy to add `transparency` in this test function, which may have a more general name, e.g.,`test_plot_fail_1d_array_with_data`

seisman:
Sounds good to me.

Issue #1076, opened 2021-03-18
plot and plot3d: Allow passing an array as symbol


core-man:
**Description of the desired feature**

We can now use a 1d array for `sizes` and `color` parameters, and `intensity` can also accept a 1d array after #1065. Therefore, we may also support a 1d array as `symbol` or `symbols` (originally posted in https://github.com/GenericMappingTools/pygmt/pull/1065#issuecomment-801212175).

---

The format of the input file of GMT `plot` (including both [**-S**](https://docs.generic-mapping-tools.org/latest/plot.html#s) and [**-I**](https://docs.generic-mapping-tools.org/latest/plot.html#i)):

> x y [ z ] [ size ] [ symbol-parameters ] [ intensity ] [ symbol ]

- **z** corresponds to the `color` parameter in PyGMT
- **size** corresponds to the `sizes` parameter in PyGMT (I don't know why we add a "s")
- **intensity** corresponds to the `intensity` (the name may be updated) parameter in PyGMT (#1065 is working on it)

The format of the input file of GMT `plot3d` (including both [**-S**](https://docs.generic-mapping-tools.org/latest/plot3d.html#s) and [**-I**](https://docs.generic-mapping-tools.org/latest/plot3d.html#i)):

> x y z [ w ] [ size ] [ symbol-parameters ] [ intensity ] [ symbol ]

- **w** corresponds to the `color` parameter in PyGMT
- **size** corresponds to the `sizes` parameter in PyGMT (I don't know why we add a "s")
- **intensity** will also accept an array input as requested in  #891

Therefore, we'd also like to support an array as `symbol` when we use `x`/`y` in `plot` and `x`/`y`/`z` in `plot3d`.

---

See an example with `x y z size intensity symbol` for the  GMT `plot` method.
```
gmt begin symbols png
gmt makecpt -Chot -T0/3/1
gmt plot -R0/5/0/1 -JX5c/1c -BWSne -Bxa1 -Bya0.5 -S -C -I -W1p,black << EOF
1  0.5   0   0.3  1  c
2  0.5   1   0.8 -1  t
3  0.5   2   0.5  0  i
EOF
gmt end show
```
![symbols](https://user-images.githubusercontent.com/50591376/111499717-303f0e80-877e-11eb-81cc-a361baa3cbfe.png)

**Are you willing to help implement and maintain this feature?** Yes, but volunteers are welcome.

weiji14:
[...]

PR #1117, opened 2021-03-24
Figure.plot: Add the "symbol" parameter to support plotting data points with varying symbols

core-man:

We can only plot different symbols via the `data` parameter in the [plot](https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot) method. This PR allows passing an array as symbols via the new `symbol` parameter. Meanwhile, we have to set `style="0.2c"`, or set 1d-array sizes using `style=True` and `sizes=sizes`. Therefore, we now have four ways to set symbols and sizes, e.g.,
- constant symbol and size: `style="c0.5c"`
- constant symbol and different sizes: `style="cc"` + `sizes=sizes`
- different symbols and constant size: `style="0.5c"` + `symbol=symbol`
- different symbols and sizes: `style=True` + `sizes=sizes` + `symbol=symbol`

Two tests are added:
- `test_plot_fail_1d_array_using_data`: Raise `GMTInvalidInput` if `symbol` is used with `data` parameter. Actually, I rename `test_plot_fail_color_size_intensity` to be this test function since we should raise such an error for 1d-array `color`, `size`, `intensity` and `symbol` parameters if they are used with the `data` parameter.  (originally posted in https://github.com/GenericMappingTools/pygmt/issues/1098#issue-837692935)
- `test_plot_symbols`: test 1d-array `symbol` (all are `"c"`) with `style="0.5c"`.

Related #1076

**Reminders**

**Slash Commands**


core-man:
[`1c48b39`](https://github.com/GenericMappingTools/pygmt/pull/1117/commits/1c48b390024336ffacf07e66b76992d57cd490be): If the `symbol` parameter is used, we will automatically set `style` to be **True** if it is not set by the user. Therefore, now we can plot different symbols and sizes like this: `symbol=symbol, sizes=sizes`. Note that the `sizes` parameter only accepts int and float (e.g., 1, 0.5) and does not accept string (e.g., **0.5c**), so the unit can have to be set by the [PROJ_LENGTH_UNIT](https://docs.generic-mapping-tools.org/latest/gmt.conf.html#term-PROJ_LENGTH_UNIT) using the `config` method if we don't want to use the default cm.

pygmt/src/plot.py
```diff
@@ -206,15 +210,18 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
     kind = data_kind(data, x, y)
 
     extra_arrays = []
-    if "S" in kwargs and kwargs["S"][0] in "vV" and direction is not None:
-        extra_arrays.extend(direction)
+    if "S" in kwargs and isinstance(kwargs["S"], str):
```
core-man, line 213: I added `isinstance(kwargs["S"], str)` in case that the `style` parameter is **True**, **False**, **None**, **int**, or **float**.

seisman:
@core-man I think the co-existence of `style` and `symbol` parameters will lead to more confusion. We should discuss what we would like to use about the `plot()` and `plot3d()` functions.

core-man:
> @core-man I think the co-existence of `style` and `symbol` parameters will lead to more confusion. We should discuss what we would like to use about the `plot()` and `plot3d()` functions.

Actually, I'd like to fully avoid using the `style` parameter, and use a more Pythonic way, e.g., `sizes`, `symbol`, `direction`. But now I don't know if how much work will have to be done before the `style` parameter is avoided since `plot`/`plot3d` still have many unimplemented and complicated features/options.

seisman:
I believe we need more discussions about what we would like to see about the `plot()` function. 
I'll convert this PR to draft.

pygmt/tests/test_plot.py
```diff
@@ -295,6 +299,25 @@ def test_plot_sizes_colors_transparencies():
...
+    """
+    fig = Figure()
+    fig.plot(
+        x=data[:, 0],
+        y=data[:, 1],
+        region=region,
+        projection="X4c",
+        color="blue",
+        sizes=np.full(len(data[:, 0]), 0.5),
+        symbol=np.full(len(data[:, 0]), "c"),
```
weiji14, line 315: At the moment, this test is checking that plotting all symbols as circles (c) work. A better test would be to test a mix of symbols such as circles (c), triangles (t), squares (s), etc (like your example in https://github.com/GenericMappingTools/pygmt/issues/1076#issue-834922951).

weiji14:
@core-man, ok if one of us picks this PR up again? There's been quite a few changes to the `plot` function including renaming of aliases (see https://github.com/GenericMappingTools/pygmt/commits/v0.9.0/pygmt/src/plot.py), so we might need to resolve a few conflicts first.

seisman:
The CI failures are caused by the codecov step. It's likely that codecov can't get the OIDC token for forks.

pygmt/src/plot.py
```diff
@@ -87,6 +89,8 @@ def plot(self, data=None, x=None, y=None, size=None, direction=None, **kwargs):
     size : 1-D array
         The size of the data points in units specified using ``style``.
         Only valid if using ``x``/``y``.
+    symbol : 1-D array
```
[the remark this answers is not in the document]
seisman: No, similar to the size parameter, symbol must be a 1-D array.

weiji14 (approved):
Cheers @seisman for getting this over the line, I've tested this locally and it looks good. The changes look simple but there were a lot of refactorings in the PyGMT codebase to get this to work! Also thanks again @core-man for starting this effort 3 years ago, we'll finally be able to release this in PyGMT v0.14.0 :tada:

P.S. We should also port these changes to `plot3d` later.

pygmt/src/plot.py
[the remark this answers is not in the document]
yvonnefroehlich: Hm. Maybe writing "list of str" is clearer here?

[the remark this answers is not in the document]
seisman: 1-D array is used here to match the types of other parameters. We probably can change it to Sequence[str] when we add typing hints.

pygmt/src/plot.py
```diff
@@ -87,6 +89,8 @@ def plot(self, data=None, x=None, y=None, size=None, direction=None, **kwargs):
     size : 1-D array
         The size of the data points in units specified using ``style``.
```
yvonnefroehlich, line 90: Hm. But if `size` is used with `symbol` the unit for `size` is given by the GMT / PyGMT defaults, correct?
Using `style` for specifing the unit in this case would be super confusing, e.g.:
```python
...
size=[0.5, 0.8, 0.9],
symbol=["s", "c", "d"],
style="c",  # now centimeters, not circle
...
```

pygmt/tests/test_plot.py
```diff
@@ -297,6 +299,25 @@ def test_plot_sizes_colors_transparencies():
...
+    """
+    fig = Figure()
+    fig.plot(
+        x=[1, 2, 3, 4],
+        y=[1, 1, 1, 1],
+        region=[0, 5, 0, 5],
+        projection="X4c",
+        fill="blue",
+        size=[0.1, 0.2, 0.3, 0.4],
+        symbol=["c", "t", "i", "s"],
```
yvonnefroehlich, line 315: Do `size` and `symbol` need to have to same length? Or is something like:

```python
size=[0.1, 0.2, 0.3, 0.4],  # different sizes but same symbol
symbol=["c"],
```
or
```python
size=[0.4],  # same size but different symbols
symbol=["c", "t", "i", "s"],
```
allowed?

yvonnefroehlich (commented):
I feel this introduction part in the docs for `Figure.plot` needs some adjustment to account for this newly added `symbol` parameter and to avoid confusions:

https://github.com/GenericMappingTools/pygmt/blob/ce89ccbea56452392c976a827d29774249c4c6f1/pygmt/src/plot.py#L65-L70

yvonnefroehlich (commented):
This is definitely a nice enhancement!

So far, I only made some comments (will be back to work at the end of this week).

I feel there can be some confusions regarding the `style`, `symbol`, and `size` parameters. Maybe we should also extend or add a gallery example for this new parameter?

pygmt/src/plot.py
seisman: Actually, it produces three circles with different sizes.

pygmt/tests/test_plot.py
seisman: Currently, they must be the same size.

seisman:
> I feel this introduction part in the docs for `Figure.plot` needs some adjustment to account for this newly added `symbol` parameter and to avoid confusions:
> 
> https://github.com/GenericMappingTools/pygmt/blob/ce89ccbea56452392c976a827d29774249c4c6f1/pygmt/src/plot.py#L65-L70

Currently, docstrings of most PyGMT methods/functions are copied literally from the upstream GMT documentation and may make no sense to PyGMT. I think we need to polish most docstrings carefully.

> I feel there can be some confusions regarding the `style`, `symbol`, and `size` parameters. Maybe we should also extend or add a gallery example for this new parameter?

Yes, definitely.

pygmt/tests/test_plot.py
yvonnefroehlich: Now I figured it out: One has to use the ``size`` or ``symbol`` parameter together with the ``style`` parameter 🙂.

PR #1121, opened 2021-03-25
Refactor text to use virtualfile_from_data

seisman:

- Refactor text to use `virtualfile_from_data`
- Raise exceptions for invalid inputs
- Add more tests for invalid inputs

Related to #949.

**Reminders**

**Slash Commands**


pygmt/src/text.py
```diff
@@ -157,15 +156,17 @@ def text_(
     # Ensure inputs are either textfiles, x/y/text, or position/text
     if position is None:
         kind = data_kind(textfiles, x, y, text)
```
meghanrjones, line 158: It it worth protecting against passing both textfiles and x/y/text to data_kind()? Just affects what error the user gets.
```suggestion
        if (x is not None or y is not None) and textfiles is not None:
            raise GMTInvalidInput( "Provide either position only, or x/y pairs, or textfiles.")
        kind = data_kind(textfiles, x, y, text)
```

pygmt/src/text.py
```diff
@@ -157,15 +156,17 @@ def text_(
...
-                "Provide either position only, or x/y pairs, not both"
+                "Provide either position only, or x/y pairs, or textfiles."
             )
-        kind = "vectors"
-
-    if kind == "vectors" and text is None:
-        raise GMTInvalidInput("Must provide text with x/y pairs or position")
+        if text is None or is_nonstr_iter(text):
+            raise GMTInvalidInput("Text can't be None or array.")
+        kind = "file"
```
meghanrjones, line 168: It doesn't really matter, but "file" does not seem like an accurate descriptor here to me.
```suggestion
        kind = ""
```

meghanrjones (approved):
Looks good to me. I only had a couple small comments.

seisman:
Thanks for your suggestions. I'll mark this PR as draft, because I don't like the current implemention.

pygmt/src/text.py
```diff
@@ -203,19 +204,15 @@ def text_(
         extra_arrays.append(kwargs["t"])
         kwargs["t"] = ""
 
+    # Append text as the last column
+    # text must be in str type, see issue #706
```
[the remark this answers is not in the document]
seisman: > text must be in str type

This statement is still True.

pygmt/src/text.py
```diff
@@ -219,18 +220,14 @@ def text_(
...
```
seisman, line 224:
```suggestion
    # Append text as the last column. Text must be passed in str type.
```

pygmt/src/text.py
```diff
@@ -219,18 +221,13 @@ def text_(
         extra_arrays.append(kwargs["t"])
         kwargs["t"] = ""
 
+    # Append text as the last column. Text must be passed in str type.
```
weiji14, line 224:
```suggestion
    # Append text at last column. Text must be passed in as str type.
```

pygmt/src/text.py
```diff
@@ -164,24 +163,27 @@ def text_(
...
+                "Provide either position only, or x/y pairs, or textfiles."
             )
-        kind = "vectors"
-
-    if kind == "vectors" and text is None:
-        raise GMTInvalidInput("Must provide text with x/y pairs or position")
+        if text is None or is_nonstr_iter(text):
+            raise GMTInvalidInput("Text can't be None or array.")
+        kind = ""
+        textfiles = ""
```
weiji14, line 186: Why are the `kind` and `textfiles` variables set to blank strings here?

weiji14 (commented):
Had a quick glance and can't see any major issues with this refactor. Is there anything else remaining you want to work on here @seisman?

seisman:
> Is there anything else remaining you want to work on here?

I plan to not make more reactors in this PR to make it easier to review.

pygmt/src/text.py
weiji14: Ok with `textfiles = ""`, didn't know about the `+tSTRING` syntax!

Looking at the `kind` variable, it seems to be an internal PyGMT-only variable that is set to bypass the if-condition `if kind == "vectors":` at L225. Maybe use `kind=None` instead?

```suggestion
        kind = None
        textfiles = ""
```

meghanrjones:
Sorry for not getting back to this in time to be helpful

seisman:
No worries. You can still leave your comments if there are any :smile:

PR #1122, opened 2021-03-25
Wrap project

claudiodsf:

⚠️ This is my first try at wrapping a GMT function. ⚠️ 

Here I propose a wrapper for [project](https://docs.generic-mapping-tools.org/latest/project.html).

Still some stuff to do (tests!) and many questions, which I'll ask in a comment below.

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -69,6 +69,7 @@ Operations on tabular data:
 
     blockmean
     blockmedian
+    project
     surface
 
 Operations on grids:
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -39,6 +39,7 @@ from pygmt.src import (
     grdtrack,
     info,
     makecpt,
+    project,
     surface,
     which,
     x2sys_cross,
--- a/pygmt/src/__init__.py
+++ b/pygmt/src/__init__.py
@@ -25,6 +25,7 @@ from pygmt.src.makecpt import makecpt
 from pygmt.src.meca import meca
 from pygmt.src.plot import plot
 from pygmt.src.plot3d import plot3d
+from pygmt.src.project import project
 from pygmt.src.subplot import set_panel, subplot
 from pygmt.src.surface import surface
 from pygmt.src.text import text_ as text  # "text" is an argument within "text_"
--- a/pygmt/src/project.py
+++ b/pygmt/src/project.py
@@ -1,0 +1,271 @@
+"""
+project - Project data onto lines or great circles, or generate tracks.
+"""
+import pandas as pd
+from pygmt.clib import Session
+from pygmt.exceptions import GMTInvalidInput
+from pygmt.helpers import (
+    GMTTempFile,
+    build_arg_string,
+    data_kind,
+    dummy_context,
+    fmt_docstring,
+    kwargs_to_strings,
+    use_alias,
+)
+
+
+@fmt_docstring
+@use_alias(
+    A="azimuth",
+    E="endpoint",
+    F="flags",
+    G="generate",
+    L="length",
+    N="flatearth",
+    Q="maptypeunits",
+    S="sort",
+    T="rotationpole",
+    V="verbose",
+    W="width",
+    Z="ellipse",
+    f="coltypes",
+)
+@kwargs_to_strings(E="sequence", L="sequence", T="sequence", W="sequence")
+def project(points, center, outfile=None, **kwargs):
+    """
+    Project data onto lines or great circles, or generate tracks.
+
+    Project reads arbitrary (x, y [, z]) data and returns any combination of
+    (x, y, z, p, q, r, s), where (p, q) are the coordinates in the projection,
+    (r, s) is the position in the (x, y) coordinate system of the point on the
+    profile (q = 0 path) closest to (x, y), and z is all remaining columns in
+    the input (beyond the required x and y columns).
+
+    Alternatively, Project may be used to generate (r, s, p) triples at equal
+    increments ``dist`` along a profile. In this case (``generate`` option),
+    the value of ``points`` is ignored (you can use, e.g., ``points=None``).
+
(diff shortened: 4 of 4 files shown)


claudiodsf:
Let's go with the questions!

1. I'm confused on whether I should use a pandas dataframe or a NumPy array as input. I saw both in the existing code and I could not understand the rationale between using one or the other (so, for the moment, I'm using pandas).
2. I'm in a special case where I have a required keyword parameter (`center`): this prevents me to use the `@kwargs_to_strings` decorator and I had to do a little hack. Any better idea?
3. I put `points` as first parameter for consistency with other wrapped functions. However this parameter is ignored if `generate` is used. Not sure if the solution I came with is nice...

seisman:
> 1. I'm confused on whether I should use a pandas dataframe or a NumPy array as input. I saw both in the existing code and I could not understand the rationale between using one or the other (so, for the moment, I'm using pandas).

Good question. I think we should always use NumPy if possible, but some modules (e.g., `blockmean`) may only support pandas dataframe. I think we should fix it.

> 2. I'm in a special case where I have a required keyword parameter (`center`): this prevents me to use the `@kwargs_to_strings` decorator and I had to do a little hack. Any better idea?

If you add `C="center"` to `@use_alias` and `C="sequence"` to `@kwargs_to_strings`, it might work.

> 3. I put `points` as first parameter for consistency with other wrapped functions. However this parameter is ignored if `generate` is used. Not sure if the solution I came with is nice...

Perhaps you can set `points=None` and check if `points` is given when `generate` is not used?

claudiodsf:
Thanks @seisman for your comments.

Let me reply to each of them:

> Good question. I think we should always use NumPy if possible, but some modules (e.g., blockmean) may only support pandas dataframe. I think we should fix it.

What about accepting both NumPy arrays and pandas dataframes? The function return value could be adjusted accordingly. For the specific case of `project`, dataframes are quite handy to keep track of column names. 

> If you add C="center" to @use_alias and C="sequence" to @kwargs_to_strings, it might work.

It doesn't work because `center` is a required argument and has to be positional, not keyword.
My understanding is that the decorator `@kwargs_to_strings` only deals with keyword arguments.

Maybe a solution would be using required keyword-only arguments: 
```python
def project(points, *, center, outfile=None, **kwargs)
```
but it doesn't seem to be supported by `@kwargs_to_strings`.

> Perhaps you can set points=None and check if points is given when generate is not used?

This will conflict with the second argument (`center`) which is required. A solution would be moving `points` to second argument, but this would be inconsistent with other functions. Let me know what you prefer.

seisman:
> > Good question. I think we should always use NumPy if possible, but some modules (e.g., blockmean) may only support pandas dataframe. I think we should fix it.
> 
> What about accepting both NumPy arrays and pandas dataframes? The function return value could be adjusted accordingly. For the specific case of `project`, dataframes are quite handy to keep track of column names.

I meant the input should be general enough to support string (filenames), 2D list, 2D numpy array and pandas.DataFrame. For output, pandas.DataFrame sounds good. For people who want numpy, perhaps they should just convert the pandas.DataFrame to numpy using `pandas.DataFrame.to_numpy()`?

willschlitzer:
Ping @claudiodsf to check in and see if you need any help getting this pull request to completion

claudiodsf:
> Ping @claudiodsf to check in and see if you need any help getting this pull request to completion

Hi @willschlitzer and thanks for your help.

There are two pending questions for this PR:

1. Should `points` be a NumPy array, a Pandas data frame, or should we support every kind of input. This seems to vary from function to function in PyGMT, but maybe you have now defined a strategy?
2. Should we keep `points` as first argument (for consistency with other PyGMT functions) or move it after `center`, given that `center` is a required argument and `points` is not?

Please note that I'm leaving on summer break next Wednesday and will have little time to work on this PR till then. I'll be back on the second half of August.

willschlitzer:
> 1. Should `points` be a NumPy array, a Pandas data frame, or should we support every kind of input. This seems to vary from function to function in PyGMT, but maybe you have now defined a strategy?

I would give it the option to accept multiple types of table-like inputs, such as a list, numpy array, or DataFrame. Have a look at `grdtrack` on how to accept multiple types of table-like inputs.

> 2. Should we keep `points` as first argument (for consistency with other PyGMT functions) or move it after `center`, given that `center` is a required argument and `points` is not?

Since neither `points` or `center` are in kwargs, I don't think the order matters too much. That being said, to update the discussion from [your comments](https://github.com/GenericMappingTools/pygmt/pull/1122#issuecomment-808761956) earlier about having `C="center"` in the kwargs, I would still recommend including it in one of the aliases, and then have some sort of check to ensure that `C` is in `kwargs`; having required arguments in the kwargs is common for PyGMT modules. Something like:

    if "C" not in kwargs.keys():
        raise GMTInvalidInput("'center' must be specified.")


> Please note that I'm leaving on summer break next Wednesday and will have little time to work on this PR till then. I'll be back on the second half of August.

No worries! Just wanted to check in to make sure this PR isn't forgotten about. Enjoy your summer break; PyGMT will still need work to be done when you get back!

weiji14:
Hi @claudiodsf, we're planning to release PyGMT v0.5.0 in about two weeks (~15 October 2021). Let us know if you have time to finish this Pull Request, otherwise we can get one of the maintainers to finalize things and get this feature in :smile:

meghanrjones:
I implemented @weiji14 and @willschlitzer's recommendations of using ``kwargs_to_strings`` for `center` and expanding the input options. I also made a few formatting adjustments to the docstrings and adjusted alias names to follow conventions.

One complication with expanding the input options and allowing dataframe output for file input is that assigning output column names is much harder, since we do not know the number or type of columns for all input types. For now, I removed the automatic naming unless the ``generate`` mode is used, in which case we know the output is ``r,s,p``. I think sorting out this would slow down this PR past v0.5.0 such that this issue could be resolved separately after the release.

PR #1125, opened 2021-03-26
Convert booleans arguments in build_arg_string, not in kwargs_to_strings

### LINKED ISSUE #640: Decorator kwargs_to_strings is too aggressive when converting boolean arguments

seisman:

See #640 and #639 (the discussions and the fix) for context.

Changes in this PR:

- 06bee27e4dd3e57b3d37078feb2e97b771cb057c: Remove `convert_bools` parameter from `kwargs_to_strings`. So `kwargs_to_strings` won't convert boolean arguments.
- b7210ff3d855903b5277b3cada3964f57120af4a: Handle boolean arguments in the `build_arg_string` function. 
- 59a2b281bdba1f72dd3ceb83611b58bea7a88679: The boolean `use_srtm` parameter of `load_earth_relief` won't be converted to an empty string or removed anymore. Now it's safe to check `if use_strm`.
- a9d167d9ab35a249660e58cfdade2eb88fd930a3: The `autolabel` parameter of `subplot` function can be either bool or str, but we had to check if `autolabel is not None`. It's not intuitive. Actually, we made a mistake in the initial version (4126c16160c67f8d2c405e7d5da56145adc989d4) and then fixed it (eadb8470065da5436802bc796e868e1ac28d8ac7).
- aac61c25a404b17665a227f362c99d11e6bf2b26: Similarly, `justify`, `angle`, and `font` can be boolean, but we have to check if it's `None`. This commit also fixes the weird codes.

Fixes #640.

**Reminders**

**Slash Commands**


pygmt/helpers/utils.py
```diff
@@ -128,10 +129,20 @@ def build_arg_string(kwargs):
...
+    ...         dict(
+    ...             R="1/2/3/4",
+    ...             J="X4i",
+    ...             P="",
+    ...             E=200,
+    ...             X=None,
+    ...             Y=None,
+    ...             A=True,
+    ...             B=False,
+    ...             Z=0,
```
weiji14, line 141: Sort alphabetically and use SI units :slightly_smiling_face:

```suggestion
    ...             A=True,
    ...             B=False,
    ...             E=200,
    ...             J="X4c",
    ...             P="",
    ...             R="1/2/3/4",
    ...             X=None,
    ...             Y=None,
    ...             Z=0,
```

pygmt/helpers/utils.py
```diff
@@ -128,10 +129,20 @@ def build_arg_string(kwargs):
...
+    ...         )
     ...     )
     ... )
-    -E200 -JX4i -P -R1/2/3/4
+    -A -E200 -JX4i -P -R1/2/3/4 -Z0
```
weiji14, line 145: Use SI units.
```suggestion
    -A -E200 -JX4c -P -R1/2/3/4 -Z0
```

weiji14 (commented):
Just some minor comments for now. I think there's some room for improvement but will suggest more later (possibly after the weekend).

pygmt/src/text.py
```diff
@@ -177,13 +177,23 @@ def text_(
         )
     ):
         kwargs.update({"F": ""})
-    if angle is not None and isinstance(angle, (int, float, str)):
+
+    if angle is True:
+        kwargs["F"] += "+a"
+    elif isinstance(angle, (int, float, str)):
         kwargs["F"] += f"+a{str(angle)}"
```
weiji14, line 184: Just thinking ahead to resolve #483 (allowing list inputs to angle/list/justify), we could simply use `if angle` (returns True for list/int/float/str inputs) like so:

```suggestion
    if angle:
        kwargs["F"] += "+a"
        if isinstance(angle, (int, float, str)):
            kwargs["F"] += str(angle)
```

pygmt/src/text.py
```diff
@@ -177,13 +177,23 @@ def text_(
...
-    if font is not None and isinstance(font, str):
+
+    if font is True:
+        kwargs["F"] += "+f"
+    elif isinstance(font, str):
         kwargs["F"] += f"+f{font}"
```
weiji14, line 189:
```suggestion
    if font:
        kwargs["F"] += "+f"
        if isinstance(font, str):
            kwargs["F"] += font
```

pygmt/src/text.py
```diff
@@ -177,13 +177,23 @@ def text_(
...
-    if justify is not None and isinstance(justify, str):
+
+    if justify is True:
+        kwargs["F"] += "+j"
+    elif isinstance(justify, str):
         kwargs["F"] += f"+j{justify}"
```
weiji14, line 194:
```suggestion
    if justify:
        kwargs["F"] += "+j"
        if isinstance(justify, str):
            kwargs["F"] += justify
```

weiji14 (commented):
A few suggestions, mostly on refactoring the if-then blocks. Feel free to disagree if it feels less readable to you.

pygmt/src/text.py
seisman: Actually, I was planing to work on #483, but I realized that the `Figure.text()` function needs more refactoring, so I opened PR #1121.

Personally, I prefer to make the changes to `text.py` as small as possible, and will address #483 after #1121 is merged.

weiji14: Ok, as you wish then.

weiji14 (approved):
Thanks @seisman, good to finally close #640, and this change will open up a lot of new possibilities with PyGMT functions!

PR #1126, opened 2021-03-26
Add more example code standards

### LINKED ISSUE #983: Improve PyGMT documentation

core-man:

A code standard about string value is added in [CONTRIBUTING.md](https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md) (Originally posted in https://github.com/GenericMappingTools/pygmt/issues/983#issue-818987282. See also https://github.com/GenericMappingTools/pygmt/issues/983#issuecomment-806716900 and https://github.com/GenericMappingTools/pygmt/issues/983#issuecomment-807224037):

- Double-quotes are used to indicate a string value instead of single-quotes. Example: "05m".
- An argument with a string value set should be added \`\` on both sides of the expressions. Example: \`\`style="v"\`\` results in ``style="v"``.

Fixes #983 

**Reminders**

**Slash Commands**


meghanrjones (approved):
This guideline seems fine to me, but we can leave it open for a bit longer to see if anyone else has other preferences.

CONTRIBUTING.md
```diff
@@ -188,6 +188,9 @@ When editing documentation, use the following standards to demonstrate the examp
    Example: `**+l**\ *label*` results in **+l***label*.
 3. Optional arguments are placed wrapped with [ ] (square brackets).
 4. Arguments that are mutually exclusive are separated with a | (bar) to denote "or".
+5. Double-quotes are used to indicate a string value. Example: "05m". An argument
+   with a string value set should be added \`\` on both sides of the expressions.
+   Example: \`\`style="v"\`\` results in ``style="v"``.
```
seisman, line 193: This file uses markdown syntax:
```suggestion
   Example: \`\`style="v"\`\` results in `style="v"`.
```

CONTRIBUTING.md
```diff
@@ -188,6 +188,9 @@ When editing documentation, use the following standards to demonstrate the examp
    Example: `**+l**\ *label*` results in **+l***label*.
 3. Optional arguments are placed wrapped with [ ] (square brackets).
 4. Arguments that are mutually exclusive are separated with a | (bar) to denote "or".
+5. Double-quotes are used to indicate a string value. Example: "05m". An argument
```
seisman, line 191: As mentioned in https://github.com/GenericMappingTools/pygmt/issues/983#issuecomment-807224037:

> Another alternative way is using bold text, for example, **05m**, as bold text means literal arguments in both GMT and PyGMT documentation.

This standard recommend double quotes for strings ("05m"), but some GMT arguments use bold text. For example, the `verbose` parameter can accept strings like "w", "d" and others, and we use bold strings following the GMT convention (see https://www.pygmt.org/dev/api/generated/pygmt.Figure.basemap.html#pygmt.Figure.basemap). 

Will it cause more confusion?

meghanrjones: GMT and PyGMT both use bold for literal arguments in the html documentation, but GMT uses different formatting for the command line documentation. I think this is the main downside of using the formatting standards from GMT's RST documentation in the python docstrings and where using bold inside strings could get confusing. For example, here is the python console help for the `verbose` parameter in Figure.basemap()
image
relative to the command line documentation for the **-V** option in GMT's basemap module:
image

core-man: Actually, I am fine with bold strings. I think Python users know whether it is a string and how to indicate a string value.

seisman: I'm still hesitate to adopt either style: **bold** or "double-quotes".

meghanrjones: > I'm still hesitant to adopt either style: **bold** or "double-quotes".

Why? Or what do you prefer?

seisman: We have to make a choice. Let's adopt **bold** texts.

meghanrjones: I understand using bold text for stand-along strings (for example, the list of verbose options that you originally shared), but when it is within a code block I think it needs to be double quotes. For example, do a search for single quotes in this latest commit https://github.com/GenericMappingTools/pygmt/commit/e131d26f9666bb28aa11d938f232f30036791976 and you will see that most are part of a code block inside docstrings.

seisman: > when it is within a code block I think it needs to be double quotes.

Yes, I agree.

seisman: > 2. Literal GMT arguments should be **bold** by wrapping the arguments with \*\*
   (two asterisks) on both sides. The argument description should be in *italicized*
   with \* (single asterisk) on both sides.

If I understand it correctly, we all agree that this rule also applies to string-type arguments like **05m**, right? If yes, we probably don't need to add a new rule, but extend this rule by adding one more example.

core-man: Add an example in 6af2deb.

core-man:
> 1. Python arguments, such as import statements, Boolean expressions, and function arguments should be wrapped as code by using `` on both sides of the code. Example: \`\`import pygmt\`\` results in `import pygmt`.

I find we use both \`True\` (`True`) and `**True**` (**True**). E.g,

- https://www.pygmt.org/dev/gallery/lines/line_custom_cpt.html?highlight=true
- https://www.pygmt.org/dev/api/generated/pygmt.datasets.load_earth_relief.html?highlight=true

Shall we add an example in the first rule? e.g., \`\`True\`\` results in `True`. This corresponds to Boolean expressions, right?

Meanwhile, shall we move this example in the first rule: \`\`style="v"\`\` results in `style="v"`? This is for function arguments.

seisman:
> Shall we add an example in the first rule? e.g., ``True`` results in `True`. This corresponds to Boolean expressions, right?

Yes.

> Meanwhile, shall we move this example in the first rule: ``style="v"`` results in `style="v"`? This is for function arguments.

Yes.

seisman (approved):

CONTRIBUTING.md
```diff
@@ -181,11 +181,12 @@ When editing documentation, use the following standards to demonstrate the examp
 
 1. Python arguments, such as import statements, Boolean expressions, and function
    arguments should be wrapped as ``code`` by using \`\` on both sides of the code.
-   Example: \`\`import pygmt\`\` results in ``import pygmt``.
+   Example: \`\`import pygmt\`\` results in ``import pygmt``, \`\`True\`\` results in `True`,
```
meghanrjones, line 184:
```suggestion
   Examples: \`\`import pygmt\`\` results in ``import pygmt``, \`\`True\`\` results in `True`,
```

CONTRIBUTING.md
```diff
@@ -181,11 +181,12 @@ When editing documentation, use the following standards to demonstrate the examp
...
+    \`\`style="v"\`\` results in `style="v"`.
 2. Literal GMT arguments should be **bold** by wrapping the arguments with \*\*
    (two asterisks) on both sides. The argument description should be in *italicized*
    with \* (single asterisk) on both sides.
-   Example: `**+l**\ *label*` results in **+l***label*.
+   Example: `**+l**\ *label*` results in **+l***label*, `**05m**` results in **05m**.
```
meghanrjones, line 189:
```suggestion
   Examples: `**+l**\ *label*` results in **+l***label*, `**05m**` results in **05m**.
```

meghanrjones (approved):

PR #1137, opened 2021-03-27
Use smaller figure dimensions in the "Plot lines" tutorial

### LINKED ISSUE #1135: Use smaller figure dimensions in the "Plot lines" tutorial
Examples in the [Plot lines](https://www.pygmt.org/dev/tutorials/lines.html) tutorial all use the `projection="X25c/20c"` argument, which sets a 25cm X 20cm figure. 

The figure dimensions (25cm X 20cm) is too large, making the annotations too small to see. 

Additionally, figures with large dimensions also have larger sizes, and are slower to generate. We should update the examples to use smaller figure dimensions.

In short, we need to change `projection="X25c/20c"` to `projection="X15c/10c"` for all examples.

The source code of the tutorial is at https://github.com/GenericMappingTools/pygmt/blob/master/examples/tutorials/lines.py.

shahid-0:
I change the projection from projection="X25c/20c" to projection="X15c/10c" 

Fixes #1135

### DIFF
--- a/examples/tutorials/lines.py
+++ b/examples/tutorials/lines.py
@@ -28,7 +28,7 @@ import pygmt
 fig = pygmt.Figure()
 fig.plot(
     region=[0, 10, 0, 10],
-    projection="X25c/20c",
+    projection="X15c/10c",
     frame="a",
     x=[1, 8],
     y=[5, 9],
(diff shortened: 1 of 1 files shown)


seisman (approved):
Looks good. Thanks.

Issue #1131, opened 2021-03-26
Migrate tests to use dvc-tracked baseline images


seisman:
After #1036, we start to use the new `dvc` based workflow for baseline images. We need to migrate existing baseline images to dvc, and also rewrite some tests. Please read the [Testing Plots](https://github.com/GenericMappingTools/pygmt/blob/master/CONTRIBUTING.md#testing-plots) section in the contributing guides for the details of the new `dvc` based workflow. (Related to #963)

This issue tracks the progress of the migration.

### Main Tasks

1. Migrate existing baseline images to dvc
2. Migrate `@check_figures_equal` tests to `@pytest.mark.mpl_image_compare`

### Migrate existing baseline images to dvc

For each item, you should see the test file `pygmt/tests/test_xxx.py` and baseline images `pygmt/tests/baseline/test_xxx_*.png`.

- [x] basemap (#1096, @seisman)
- [x] coast (#1103, @willschlitzer)
- [x] colorbar (#1143 @willschlitzer)
- [x] config (#1165 @seisman)
- [x] contour (#1146 @michaelgrund) 
- [x] grdcontour (#1158 @michaelgrund) 
- [x] grdimage (@willschlitzer)
- [ ] grdview (#1154 @weiji14) 
- [x] legend (#1166 @willschlitzer)
- [x] test_makecpt.py (#1168 @willschlitzer)
- [x] meca (#1153 @seisman)
- [x] plot (#1142 @michaelgrund)
- [x] text (#1167 @michaelgrund) 
- [x] shift_origin (in `pygmt/tests/test_figure.py`) (#1133, @seisman)

### Migrate `@check_figures_equal` tests to `@pytest.mark.mpl_image_compare`

Some tests in the following files use `@check_figures_equal`, and should be migrated to `@pytest.mark.mpl_image_compare`.

- [x] test_basemap.py (#1096, @seisman)
- [x] test_coast.py (#1103, @willschlitzer)
- [x] test_colorbar.py (#1143 @willschlitzer)
- [x] test_config.py (#1165 @seisman)
- [x] test_grd2cpt.py (#1162 @willschlitzer)
- [x] test_grdcontour.py (#1158 @michaelgrund) 
- [x] test_grdimage.py (@willschlitzer)
- [ ] test_grdview.py  (#1154 @weiji14) 
- [x] test_inset.py (#1163 @willschlitzer)
- [x] test_makecpt.py (#1168 @willschlitzer)
[...]

PR #1103, opened 2021-03-22
Migrate Figure.coast tests to dvc

willschlitzer:
Changing the tests in test_coast.py to use `dvc` and `mpl_image_compare`. An update/fix to #1102.

Related to #1131.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1143, opened 2021-03-28
Migrate Figure.colorbar tests to use dvc

willschlitzer:
As discussed in #1131, this moves the reference figures in `fig.colorbar` to use `dvc` and replaces the tests using `@check_figures_equal`.

**Reminders**

**Slash Commands**


willschlitzer:
@GenericMappingTools/python `test_colorbar.py` originally had 18 tests, which I've shortened so far to 15 tests. Much like `fig.coast` in #1103, this is a pretty straightforward module, so I'm assuming I can remove tests further. My thought is to leave the three tests that pass different types of parameters for `shading=`, the ones that pass arguments to `scale=` and `truncate=`, and combine the tests that use variants of arguments passed to `box=` and `position=`. Thoughts?

weiji14:
> 1. Add basemaps?
> 
> image
> 
> Do you think we should add a basemap to these tests, so that at least we know the colorbar is plotted at the correct positions.

Nah, I think it's fine as it is.

> 2. If I understand it correctly, calling `Figure.colorbar()` without any arguments also works. I think we should also test it. I think most people usually call `Figure.colorbar()` first and add more arguments to change the appearance.

Yes, but is it important enough to test this? Also seems beyond the scope of this PR.

> 3. 
> 
> image
> 
> Colorbars in two tests have messy annotations. Perhaps we should add `frame=True` to `Figure.colorbar()`?

Yes, probably good to change the annotation interval so the axis labels are visible.

pygmt/tests/test_colorbar.py
```diff
@@ -156,61 +45,23 @@ def test_colorbar_scaled_z_values():
...
-    # Use single-character arguments for the reference image
-    fig_ref.basemap(R="0/10/0/10", J="X15c", B="a")
-    fig_ref.colorbar(C="geo", I="-0.7/0.2")
-
-    fig_test.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
-    fig_test.colorbar(cmap="geo", shading="-0.7/0.2")
-    return fig_ref, fig_test
+    fig = Figure()
+    fig.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
+    fig.colorbar(cmap="geo", shading=True)
```
seisman, line 55:
```suggestion
    fig.colorbar(cmap="geo", shading=True, frame=True)
```

pygmt/tests/test_colorbar.py
```diff
@@ -156,61 +45,23 @@ def test_colorbar_scaled_z_values():
...
-    # Use single-character arguments for the reference image
-    fig_ref.basemap(R="0/10/0/10", J="X15c", B="a")
-    fig_ref.colorbar(C="geo", I="-0.7/0.2")
-
-    fig_test.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
-    fig_test.colorbar(cmap="geo", shading=[-0.7, 0.2])
-    return fig_ref, fig_test
+    fig = Figure()
+    fig.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
+    fig.colorbar(cmap="geo", shading=[-0.7, 0.2])
```
seisman, line 66:
```suggestion
    fig.colorbar(cmap="geo", shading=[-0.7, 0.2], frame=True)
```

seisman (approved):

seisman (approved):

PR #1104, opened 2021-03-22
Create Github Action workflow for reporting DVC image diffs

weiji14:
**Description of proposed changes**

To make reviewing new baseline test images (*.png) easier, this workflow checks what images have been added or modified in a Pull Request. The changes are published in a table and as a series of images by a bot-generated GitHub comment.

References:
- https://dvc.org/doc/cml/start-github
- https://github.com/iterative/cml#using-cml-with-dvc
- https://github.com/iterative/cml_dvc_case/blob/master/.github/workflows/cml.yaml
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths
- https://linuxize.com/post/how-to-read-a-file-line-by-line-in-bash/
- https://github.com/peter-evans/create-or-update-comment

Addresses a painpoint of #963

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1144, opened 2021-03-29
Fix dvc-diff workflow for multiple changes

seisman:

Two problems in the original dvc-diff workflow:

1. `while` loop stops after the first iteration, so only one image is shown. The solution is adding `< /dev/null` to the `cml-publish` command (reference https://stackoverflow.com/questions/13800225/)
2. The `dvc diff --show-md master HEAD` command outputs a blank line at the end. It causes trouble because `cml-publish` will try to upload an image with an empty name. The solution is checking `NF==5` in `awk`, (i.e., checking that the **N**umber of **F**ields equals to 5).

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/dvc-diff.yml
+++ b/.github/workflows/dvc-diff.yml
@@ -48,7 +48,7 @@ jobs:
       id: image-diff
       run: |
         # Get just the filename of the changed image from the report
-        awk 'NR>=7 {print $4}' report.md > diff_files.txt
+        awk 'NF==5 && NR>=7 {print $4}' report.md > diff_files.txt
 
         # Append each image to the markdown report
         echo -e "## Image diff(s)\n" >> report.md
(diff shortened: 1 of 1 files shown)


seisman:
https://github.com/GenericMappingTools/pygmt/pull/1144#issuecomment-809091420 is the comment when I have the changes of #1096 in this branch.

weiji14 (approved):
> [#1144 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1144#issuecomment-809091420) is the comment when I have the changes of #1096 in this branch.

Cool, thanks @seisman! I really should have tested things better in #1104! Perhaps we should revert the changes made to test_basemap_* since they will be covered in #1096, i.e. let there only be a diff on `.github/workflows/dvc-diff.yml`.

seisman:
> > [#1144 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1144#issuecomment-809091420) is the comment when I have the changes of #1096 in this branch.
> 
> Cool, thanks @seisman! I really should have tested things better in #1104! Perhaps we should revert the changes made to test_basemap_* since they will be covered in #1096, i.e. let there only be a diff on `.github/workflows/dvc-diff.yml`.

Rebased and now only the changes of `.github/workflows/dvc-diff.yml` are committed.

Issue #1120, opened 2021-03-25
Backwards compatibility and deprecation policy


seisman:
As stated in many places, PyGMT is still not a finished project:

> All of the API (functions/classes/interfaces) is subject to change until we reach v1.0.0 as per the semantic versioning specification. There may be non-backward compatible changes as we experiment with new design ideas and implement new features.

So we're allowed to break backward-compatibility in any major and minor versions. However, from users' point of view, they would be disappointed when they find that their scripts break after upgrading PyGMT. There are always be some trade-offs between users' and developers' experiences.

We now already have some open issues that might break backward-compatibility, e.g.,

- https://github.com/GenericMappingTools/pygmt/issues/1118
- https://github.com/GenericMappingTools/pygmt/issues/1094
- https://github.com/GenericMappingTools/pygmt/issues/1042
- https://github.com/GenericMappingTools/pygmt/issues/924
- https://github.com/GenericMappingTools/pygmt/issues/262
- https://github.com/GenericMappingTools/pygmt/issues/247
- https://github.com/GenericMappingTools/pygmt/issues/593

It's time for us to discuss about how we will handle backward compatibility and how to deprecate/remove existing features.

**Resources:**

- https://www.python.org/dev/peps/pep-0387/
- https://numpy.org/neps/nep-0023-backwards-compatibility.html

michaelgrund:
I'm definitely a fan of backwards compatibility, however, if some points are changed to improve the overall product one has to think about how this should be handled. For example, when I started using PyGMT in 2019 several aliases were not implemented but now they are. Of course we support both, original GMT shortcuts and aliases and thus my notebooks will still work, however, some other things like how data is passed into a module slightly changed and there might appear issues. Therefore, as @seisman mentioned, supporting backwards compatibility is (in my opinion) necessary, especially to avoid disappointed early stage users.

seisman:
IMHO, we should try our best to:

1. Keep backward-compatibility
2. Raise deprecation warnings if users are using the old syntax 
3. Fully deprecate old syntax or remove unsupported feature after _X_ minor versions or _X_ months.

seisman:
> IMHO, we should try our best to:
> 
> 1. Keep backward-compatibility
> 2. Raise deprecation warnings if users are using the old syntax
> 3. Fully deprecate old syntax or remove unsupported features after _X_ minor versions or _X_ months.

Ping @GenericMappingTools/pygmt-contributors for thoughts.

[...]

PR #1160, opened 2021-04-02
Document the deprecation policy and addd the deprecate_parameter decorator to deprecate parameters

seisman:

This is a draft implementation to deprecate a parameter but keep backward compatibility.

For example, the `Figure.plot()` function has the `sizes` parameter, but we want to rename it to `size`, and keep backward compatibility. We can add the `deprecate_parameter` decorator to the `Figure.plot()` method:

```
@deprecate_parameter("sizes", "size", "v0.4.0", "v0.6.0")
def plot():
   ...
```
It means the old parameter name `sizes` is deprecated since v0.4.0, and will be removed in v0.6.0.
The new parameter name is `size`.

When users use `sizes`, a warning will be raised.

References:

- https://stackoverflow.com/questions/2536307
- https://docs.python.org/3/library/warnings.html
- https://www.python.org/dev/peps/pep-0565/

Related to #1120.

**Reminders**

**Slash Commands**


pygmt/helpers/decorators.py
```diff
@@ -469,3 +470,65 @@ def remove_bools(kwargs):
...
+            """
+            New module instance that converts old parameters to new parameters.
+            """
+            if oldname in kwargs:
+                if newname in kwargs:
+                    raise GMTInvalidInput(
+                        f"Can't provide both '{newname}' and '{oldname}'"
+                    )
+                msg = (
+                    f"'{oldname}' is deprecated since {deprecate_version} and will"
```
weiji14, line 523:
```suggestion
                    f"The '{oldname}' parameter is deprecated since {deprecate_version} and will"
```

pygmt/helpers/decorators.py
```diff
@@ -469,3 +470,65 @@ def remove_bools(kwargs):
...
+                    f" be removed in {remove_version}; use '{newname}' instead."
```
weiji14, line 524:
```suggestion
                    f" be removed in {remove_version}; please use '{newname}' instead."
```

pygmt/src/plot.py
```diff
@@ -43,7 +44,8 @@
     t="transparency",
 )
 @kwargs_to_strings(R="sequence", c="sequence_comma", i="sequence_comma", p="sequence")
-def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
+@deprecate_parameter("sizes", "size", "v0.4.0", "v0.6.0")
```
weiji14, line 47: I would use the keyword name here, make it easier to regex match when the time comes to deprecate something. Also, do we want to deprecate this at v0.6.0 or v0.5.0?

```suggestion
@deprecate_parameter("sizes", "size", "v0.4.0", remove_version="v0.6.0")
```

pygmt/helpers/decorators.py
```diff
@@ -469,3 +470,65 @@ def remove_bools(kwargs):
...
+                )
+                warnings.simplefilter("always", DeprecationWarning)  # turn off filter
+                warnings.warn(msg, DeprecationWarning, 2)
+                warnings.simplefilter("default", DeprecationWarning)  # reset filter
```
weiji14, line 528: One more thing, I don't quite get what the `warnings.simplefilter` is here for? I see it's copied from the stackoverflow answer, but I'm still wondering why we can't just do `warnings.warn`.

pygmt/src/plot.py
seisman: > Also, do we want to deprecate this at v0.6.0 or v0.5.0?

As `sizes` is heavily used, I'd like to give users more time to change the old names to new names.

pygmt/helpers/decorators.py
seisman: I haven't read the full documentation of `warnings`, so I can't answer the question now.

seisman: > One more thing, I don't quite get what the `warnings.simplefilter` is here for?

The `DeprecationWarning` is ignored by default. So, if you remove the `warnings.simplefilter` calls and run the example below, you won't see the warnings. That's why `warnings.simplefilter` are necessary here.
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[-5, 5, -5, 5], projection="X10c/10c", frame=True)
fig.plot(x=[-2, 2], y=[0, 0], style="c", sizes=[1, 2])
fig.plot(x=[-2, 2], y=[2, 2], style="c", size=[1, 2])
fig.savefig("symbols.pdf")
```

---

Here are some related descriptions about `DeprecationWarning` and `FutureWarning`. If I understand it correctly, we should use `FutureWarning` instead, right?

https://docs.python.org/3/library/warnings.html

> **DeprecationWarning**
> 
> Base category for warnings about deprecated features when those warnings are intended for other **Python developers** (ignored by default, unless triggered by code in `__main__`).
> 
> **FutureWarning**
> 
> Base category for warnings about deprecated features when those warnings are **intended for end users** of applications that are written in Python.
> 
> Changed in version 3.7: Previously DeprecationWarning and FutureWarning were distinguished based on whether a feature was being removed entirely or changing its behaviour. They are now distinguished based on their intended audience and the way they’re handled by the default warnings filters.

https://www.python.org/dev/peps/pep-0565/#id13

> DeprecationWarning: reported by default for code that runs directly in the __main__ module (as such code is considered relatively unlikely to have a dedicated test suite), but hidden by default for code in other modules. The intended audience is Python developers that are at risk of upgrades to their dependencies (including upgrades to Python itself) breaking their software (e.g. developers using Python to script environments where someone else is in control of the timing of dependency upgrades).
> 
> FutureWarning: reported by default for all code. The intended audience is users of applications written in Python, rather than other Python developers (e.g. warning about use of a deprecated setting in a configuration file format).
>

weiji14: Yeah, looks like `FutureWarning` does look easier, but PEP 0565 states:

> For library and framework authors that want to ensure their API compatibility warnings are more reliably seen by their users, the recommendation is to use a custom warning class that derives from `DeprecationWarning` in Python 3.7+, and from `FutureWarning` in earlier versions.

So maybe we should stick to DeprecationWarning instead since we're Python 3.7+? I'm not so familiar with the history of DeprecationWarning/FutureWarning in Python to be honest, we should look at how other projects handle things e.g. `pandas`/`geopandas` perhaps.

seisman: PEP 0565 also says:

> The standard library documentation will be updated to explicitly recommend the use of FutureWarning (rather than DeprecationWarning) for backwards compatibility warnings that are intended to be seen by users of an application.

seisman: It seems both pandas and xarray use `FutureWarning` a lot:

- https://github.com/pandas-dev/pandas/blob/master/pandas/util/_depr_module.py
- https://github.com/pydata/xarray/blob/master/xarray/core/common.py

weiji14: Ok, let's use FutureWarning then, would be good to remove the `warnings.simplefilter` lines too :smile:

pygmt/src/plot.py
```diff
@@ -6,6 +6,7 @@
 from pygmt.helpers import (
```
seisman, line 6: FYI, I plan to revert the changes in `plot.py` before merging this PR, because the `deprecate_parameter` decorator is already tested in its docstrings, and the deprecation of `sizes` parameter need more changes and should be done in a separate PR.

pygmt/helpers/decorators.py
```diff
@@ -439,3 +440,71 @@ def new_module(*args, **kwargs):
         return new_module
 
     return converter
+
+
+def deprecate_parameter(oldname, newname, deprecate_version, remove_version):
+    """
+    Decorator to deprecate a parameter.
```
weiji14, line 447: Maybe add an extra note here, feel free to edit my English.

```suggestion
    Decorator to deprecate a parameter.
    
    The old parameter name will be swapped automatically to the new parameter
    name, and users will receive a FutureWarning to inform them of the pending
    deprecation.
```

weiji14 (commented):
Looks good, I would add a note about our deprecation policy (and how to use `@deprecate_parameter`) in the maintainer docs, maybe after #1185 is completed.

pygmt/helpers/decorators.py
seisman: I also added one more note "Use this decorator below the ``use_alias`` decorator.", because `use_alias` will convert long-form parameters to short-form parameters.

seisman:
> Looks good, I would add a note about our deprecation policy (and how to use `@deprecate_parameter`) in the maintainer docs, maybe after #1185 is completed.

#1185 is already merged. Do you want me to document the deprecation policy in this PR or in a separate PR?

weiji14:
Document it here please.

doc/maintenance.md
```diff
@@ -146,6 +146,46 @@ supported version of Python. Minimum Python and NumPy version support should be
...
+
+PyGMT is still undergoing rapid developement. All the API is subject to change
+until the v1.0.0 release.
+
+Basic policy for backward compatibility:
+
+- Any incompatible changes should go through the deprecation process below.
+- Incompatible changes are allowed in major and minor releases, not in patch
+  releases.
+- Incompatible changes should be documente in the release notes.
```
meghanrjones, line 159:
```suggestion
- Incompatible changes should be documented in the release notes.
```

seisman:
Ping @GenericMappingTools/pygmt-contributors for reviews and comments on this PR.

doc/maintenance.md
```diff
@@ -146,6 +146,46 @@ supported version of Python. Minimum Python and NumPy version support should be
...
+## Backwards compatibility and deprecation policy
+
+PyGMT is still undergoing rapid developement. All of the API is subject to change
+until the v1.0.0 release.
+
+Basic policy for backwards compatibility:
+
+- Any incompatible changes should go through the deprecation process below.
+- Incompatible changes are allowed in major and minor releases, not in patch
+  releases.
```
weiji14, line 158:
```suggestion
- Incompatible changes are only allowed in major and minor releases, not in
  patch releases.
```

doc/maintenance.md
```diff
@@ -146,6 +146,46 @@ supported version of Python. Minimum Python and NumPy version support should be
...
+- Incompatible changes should be documented in the release notes.
+
+When making incompatible changes, we should follow the process:
+
+- Discuss whether the incompatible changes are necessary on GitHub.
+- Make the changes in a backwards compatible way, and raise a `FutureWarning`
+  warning for old usage. At least one test using the old usage should be added.
+- The warning message should clearly explain the changes and include the versions
+  in which the old usage is deprecated and is expected to be removed.
+- The `FutureWarning` warning should appear in 2-4 minor versions, depending on
```
weiji14, line 168: May want to wait for votes to come in on https://github.com/GenericMappingTools/pygmt/issues/1120#issuecomment-818377643 before committing.

```suggestion
- The `FutureWarning` warning should appear for 2-4 minor versions, depending on
```

doc/maintenance.md
```diff
@@ -146,6 +146,46 @@ supported version of Python. Minimum Python and NumPy version support should be
...
+  the impact of the changes. It means the deprecation period usually lasts
+  3-12 months.
+- Remove the old usage and warning when reaching the declared version.
+
+To rename a function parameter, we can simply add the `@deprecated_parameter`
+decorator before the function (after the `@use_alias` decorator if any).
```
weiji14, line 174:
```suggestion
To rename a function parameter, add the `@deprecated_parameter` decorator
before the function definition (but after the `@use_alias` decorator if it exists).
```

doc/maintenance.md
```diff
@@ -146,6 +146,46 @@ supported version of Python. Minimum Python and NumPy version support should be
...
+
+```
+@fmt_docstring
+@use_alias(J="projection", R="region", V="verbose")
+@kwargs_to_strings(R="sequence")
+@deprecate_parameter("sizes", "size", "v0.4.0", remove_version="v0.6.0")
+def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
+    pass
+```
+It means the old parameter name `sizes` is deprecated since v0.4.0, and will be
```
weiji14, line 185:
```suggestion

In this case, the old parameter name `sizes` is deprecated since v0.4.0, and will be
```

pygmt/helpers/decorators.py
```diff
@@ -439,3 +440,77 @@ def new_module(*args, **kwargs):
...
+
+    The old parameter name will be swapped automatically to the new parameter
```
weiji14, line 449:
```suggestion
    The old parameter name will be automatically swapped to the new parameter
```

weiji14 (commented):
Just some proofread edits. No outstanding issues on the `@deprecate_parameter` function.

pygmt/helpers/decorators.py
```diff
@@ -439,3 +440,77 @@ def new_module(*args, **kwargs):
...
+            """
+            if oldname in kwargs:
+                if newname in kwargs:
+                    raise GMTInvalidInput(
+                        f"Can't provide both '{newname}' and '{oldname}'."
+                    )
+                msg = (
+                    f"The '{oldname}' parameter is deprecated since {deprecate_version}"
+                    f" and will be removed in {remove_version};"
+                    f" please use '{newname}' instead."
```
core-man, line 508:
```suggestion
                    f" and will be removed in {remove_version}."
                    f" Please use '{newname}' instead."
```

pygmt/helpers/decorators.py
```diff
@@ -439,3 +440,77 @@ def new_module(*args, **kwargs):
...
+            """
+            New module instance that converts old parameters to new parameters.
+            """
+            if oldname in kwargs:
+                if newname in kwargs:
+                    raise GMTInvalidInput(
+                        f"Can't provide both '{newname}' and '{oldname}'."
+                    )
+                msg = (
+                    f"The '{oldname}' parameter is deprecated since {deprecate_version}"
```
core-man, line 506:
```suggestion
                    f"The '{oldname}' parameter has been deprecated since {deprecate_version}"
```

core-man:
Woow~ Looks great :smile: Only some minor suggestions.

pygmt/helpers/decorators.py
```diff
@@ -439,3 +440,77 @@ def new_module(*args, **kwargs):
...
+        """
+
+        @functools.wraps(module_func)
+        def new_module(*args, **kwargs):
+            """
+            New module instance that converts old parameters to new parameters.
+            """
+            if oldname in kwargs:
+                if newname in kwargs:
+                    raise GMTInvalidInput(
```
core-man, line 502: Add a test for this?

seisman: Done in c73b80bdcb4f64bffd71a5eb1427ea9138954100.

core-man (approved):

weiji14 (approved):

PR #1168, opened 2021-04-03
Migrate makecpt tests to use dvc

willschlitzer:
Per #1131, this moves the reference figures in `makecpt` to use `dvc` and replaces the tests using `@check_figures_equal`.

Fixes #

**Reminders**

**Slash Commands**


pygmt/tests/test_makecpt.py
```diff
@@ -68,21 +67,16 @@ def test_makecpt_to_plot_grid(grid):
...
     of grid.
     """
-    # Use single-character arguments for the reference image
-    fig_ref = Figure()
-    makecpt(C="oleron", T="-4500/4500")
-    fig_ref.grdimage(grid, J="W0/6i")
-
-    fig_test = Figure()
+    fig = Figure()
     makecpt(cmap="oleron", series="-4500/4500")
```
weiji14, line 77: Use a list here.
```suggestion
    makecpt(cmap="oleron", series=[-4500, 4500])
```

pygmt/tests/test_makecpt.py
```diff
@@ -154,29 +126,6 @@ def test_makecpt_reverse_color_only(grid):
...
-@pytest.mark.mpl_image_compare
-def test_makecpt_reverse_color_and_zsign(grid):
-    """
-    Use static color palette table with both its colors and z-value sign
-    reversed.
-    """
-    fig = Figure()
-    makecpt(cmap="earth", reverse="cz")
-    fig.grdimage(grid, projection="W0/6i")
-    return fig
```
weiji14, line 177: What do you think about keeping this test_makecpt_reverse_color_and_zsign since it's more advanced, and maybe removing test_makecpt_reverse_color_only instead?

weiji14 (commented):
Just a few minor comments. Also, could you change the inches to centimetres instead? Thanks!

pygmt/tests/test_makecpt.py
willschlitzer: That's fine. I figured we wanted to keep `test_makecpt_reverse_color_only` since it was using a Boolean, so it seemed like a more Python-specific thing to test, whereas passing strings is testing passing a literal string to the GMT API.

weiji14: Hmm, you're right on the Boolean thing. How about we keep both then? I feel like `makecpt` is one of those GMT modules (like `basemap` and `plot`) we want to overtest rather than undertest. I get nervous about removing tests (goes against all my job training on code testing).

willschlitzer: Re-added the function in 571c1ef

weiji14 (approved):

PR #1169, opened 2021-04-03
Migrate Figure.grdimage tests to use dvc

willschlitzer:
Per #1131, this pull request moves reference images to use `dvc` and replaces tests that use `@check_figures_equal`. It also removes some tests, including those that use `pytest.mark.parameterize` to repeatedly pass different arguments to a test.

Fixes #

**Reminders**

**Slash Commands**


pygmt/tests/test_grdimage.py
```diff
@@ -146,18 +146,16 @@ def test_grdimage_over_dateline(xrgrid):
 
 
 @check_figures_equal()
-@pytest.mark.parametrize("lon0", [0, 123, 180])
-@pytest.mark.parametrize("proj_type", ["H", "W"])
-def test_grdimage_central_meridians(grid, proj_type, lon0):
+def test_grdimage_central_meridians(grid):
```
weiji14, line 149: Err, could you revert this change please. This parametrize test was set up explicitly at https://github.com/GenericMappingTools/pygmt/pull/560#discussion_r484011483 to catch a bug (#390).

willschlitzer: Will do; I had just assumed it was repeatedly testing the same arguments. On that note, how should this handle a reference image then, since it produces 6 different images?

weiji14: Actually, I would recommend to keep this test on `@check_figures_equal` for now because it's quite a serious bug we want resolved at some point (differences between plotting NetCDF and xarray grids should not exist).

willschlitzer: Sounds good!

weiji14: Yes, just for now please :smile:

willschlitzer: Should be fixed (took me too long to realize there would be errors when I merged with GitHub)!

weiji14:
Let's hold on to this for a while @willschlitzer, there seems to be a bug at #1172 for both `fig.grdimage` and `fig.grdview` with plotting geographic grids. I'll re-mark it as draft for now.

willschlitzer:
> Let's hold on to this for a while @willschlitzer, there seems to be a bug at #1172 for both `fig.grdimage` and `fig.grdview` with plotting geographic grids. I'll re-mark it as draft for now.

@weiji14 It looks like the upstream bug has been fixed, according to the comment from @seisman. Should I mark this as ready for review, or are we still waiting on additional changes?

weiji14:
> > Let's hold on to this for a while @willschlitzer, there seems to be a bug at #1172 for both `fig.grdimage` and `fig.grdview` with plotting geographic grids. I'll re-mark it as draft for now.
> 
> @weiji14 It looks like the upstream bug has been fixed, according to the comment from @seisman. Should I mark this as ready for review, or are we still waiting on additional changes?

The question is, do we want to 1) store these 'incorrect' GMT 6.1.1 images now, or 2) wait for GMT 6.2.0rc1 to come out and use that to create the 'correct' grdimage test images.

willschlitzer:
> > > Let's hold on to this for a while @willschlitzer, there seems to be a bug at #1172 for both `fig.grdimage` and `fig.grdview` with plotting geographic grids. I'll re-mark it as draft for now.
> > 
> > 
> > @weiji14 It looks like the upstream bug has been fixed, according to the comment from @seisman. Should I mark this as ready for review, or are we still waiting on additional changes?
> 
> The question is, do we want to 1) store these 'incorrect' GMT 6.1.1 images now, or 2) wait for GMT 6.2.0rc1 to come out and use that to create the 'correct' grdimage test images.

I'll defer to you on this one. In my opinion (which I don't feel strongly about), this test should be merged as the reference images are "wrong" because of a GMT problem. Since, as I understand it, we'll have to update most of our reference images when we switch to GMT 6.2, I figure we can update the `grdimage` and `grdview` images then, along with the other tests that are affected by GMT themes.

Side note: How will testing using reference images work if we're still on PyGMT v0.3.1, as we'll want to be be testing our compatibility with both GMT 6.1.1 and GMT 6.2 that may produce different images from each other, but only have a single reference image per test?

weiji14 (approved):
> > The question is, do we want to 1) store these 'incorrect' GMT 6.1.1 images now, or 2) wait for GMT 6.2.0rc1 to come out and use that to create the 'correct' grdimage test images.
> 
> I'll defer to you on this one. In my opinion (which I don't feel strongly about), this test should be merged as the reference images are "wrong" because of a GMT problem, not a PyGMT one. Since, as I understand it, we'll have to update most of our reference images when we switch to GMT 6.2, I figure we can update the `grdimage` and `grdview` images then, along with the other tests that are affected by GMT themes.

Ok, let's merge this in then, and we can update the baseline images when GMT 6.2.0 is out (the GMT 6.2.0rc1 release should be close :crossed_fingers:). The `grdview` test migration to dvc doesn't quite work but that's not relevant to this PR.

> Side note: How will testing using reference images work if we're still on PyGMT v0.3.1, as we'll want to be be testing our compatibility with both GMT 6.1.1 and GMT 6.2 that may produce different images from each other, but only have a single reference image per test?

Well, I'd like to think that PyGMT v0.3.1 should work with GMT 6.2.0 for 99% of use cases, and if it doesn't, people should just upgrade to GMT 6.2.0 :imp:. By 'work', I mean run without crashing, not that the exact same figure is produced (a user won't typically care if their image is a few pixels different between versions). Yes, in a perfect world, we would have a matrix build with GMT 6.1.1 and GMT 6.2.0, and baseline images being pulled from the DVC remote matching these 2 versions, but that's a lot of work to maintain.

Anyways, LGTM!

seisman:
The upstream bug #1172 should not affect these images, because these tests all explicitly specify geographic projections.

seisman (approved):

willschlitzer:
> Anyways, LGTM!

Felt particularly old when I had to look up what "LGTM" means.

weiji14:
> > Anyways, LGTM!
> 
> Felt particularly old when I had to look up what "LGTM" means.

I like to think of it as "Let's get this merged" :twisted_rightwards_arrows: Wait until you try looking up GMT.

willschlitzer:
> > > Anyways, LGTM!
> 
> > 
> 
> > Felt particularly old when I had to look up what "LGTM" means.
> 
> 
> 
> I like to think of it as "Let's get this merged" :twisted_rightwards_arrows: Wait until you try looking up GMT.

The internet told me it was "looks good to me" which works too!

Definitely typed in GMT to Google and realized it most certainly wasn't the GMT I was looking for 😂

Issue #1151, opened 2021-03-31
Update CSS for sphinx_rtd_theme 0.5.1


meghanrjones:
**Description of the problem**

#1150 updated sphinx_rtd_theme to 0.5.1, which fixed the search problem but added unnecessary whitespace to the parameters section (see https://github.com/GenericMappingTools/pygmt/pull/629 for context). The custom CSS in `doc/_static/style.css` should be updated for improved appearance with the latest sphinx_rtd_theme release.

meghanrjones:
I can update the custom CSS, but probably will not have time to do so right away. Anyone is welcome to comment or self-assign if they are interested in addressing this issue instead.

weiji14:
Just mentioning that #1174 (see https://github.com/GenericMappingTools/pygmt/pull/1174#discussion_r606914578) may need to be backported to the gh-pages too, and can be done alongside the backport of #1150 (see https://github.com/GenericMappingTools/pygmt/pull/1150#issuecomment-811053780). But no rush on this, GMT 6.2.0 comes first :slightly_smiling_face:

PR #1150, opened 2021-03-30
Bump sphinx_rtd_theme from 0.4.3 to 0.5.1

meghanrjones:
**Description of proposed changes**

This PR updates sphinx_rtd_theme. Fixes #1149.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1174, opened 2021-04-04
Set __commit__ to an empty string for non-dev releases

### LINKED ISSUE #1171: Broken link to the commit hash in the latest documentation
To reproduce the issue:

1. Go to https://www.pygmt.org/latest/
2. Scroll down to the bottom
3. Click the "Revision v0.3.1" link:

image

"v0.3.1" is linked to a nonexistent commit hash (https://github.com/GenericMappingTools/pygmt/commit/v0.3.1).

seisman:

For non-dev versions, `__version__` is a string like `v0.3.1`, without commit hashes. In this case, we should set `__commit__` to `""`.

When `__commit__` is empty, the "Revision xxx" in the documentation footer won't show, so this PR also fixes #1171.

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -151,7 +151,9 @@ html_theme_options = {}
 repository = "GenericMappingTools/pygmt"
 repository_url = "https://github.com/GenericMappingTools/pygmt"
 if __commit__:
-    commit_link = f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+    commit_link = (
+        f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+    )
 else:
     commit_link = ""
 html_context = {
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -155,7 +155,9 @@ if __commit__:
         f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
     )
 else:
-    commit_link = ""
+    commit_link = (
+        f'<a href="{repository_url}/releases/tag/{ __version__ }">{ __version__ }</a>'
+    )
 html_context = {
     "menu_links": [
         (
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -150,7 +150,10 @@ html_theme = "sphinx_rtd_theme"
 html_theme_options = {}
 repository = "GenericMappingTools/pygmt"
 repository_url = "https://github.com/GenericMappingTools/pygmt"
-commit_link = f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+if __commit__:
+    commit_link = f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+else:
+    commit_link = ""
 html_context = {
     "menu_links": [
         (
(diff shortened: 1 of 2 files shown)


pygmt/__init__.py
```diff
@@ -47,7 +47,7 @@
 
 # Get semantic version through setuptools-scm
 __version__ = f'v{get_distribution("pygmt").version}'  # e.g. v0.1.2.dev3+g0ab3cd78
-__commit__ = __version__.split("+g")[-1]  # 0ab3cd78
+__commit__ = __version__.split("+g")[-1] if "+g" in __version__ else ""  # 0ab3cd78
```
weiji14, line 50: If we set the commit hash as an empty string, then won't the link mentioned at #1171 just become https://github.com/GenericMappingTools/pygmt/commit/ which redirects to the master branch?

seisman: If `__commit__` is empty, the "Revision xxxx" won't show.

See https://github.com/readthedocs/sphinx_rtd_theme/blob/aa28eecc3a0ea7a28efa0880bfcec6abaabdb1fa/sphinx_rtd_theme/footer.html#L32

weiji14 (approved):

pygmt/__init__.py
seisman: @weiji14 You're right here, so I have to set the variable `commit_link` in `conf.py` to an empty string. See the changes in 6b42847bcefad3ded0356cd3238cf323eeeacee6.

Preview for a local build when `__version__ == "v0.3.1"`

image

doc/conf.py
```diff
@@ -150,7 +150,12 @@
 html_theme_options = {}
 repository = "GenericMappingTools/pygmt"
 repository_url = "https://github.com/GenericMappingTools/pygmt"
-commit_link = f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+if __commit__:
+    commit_link = (
+        f'<a href="{repository_url}/commit/{ __commit__ }">{ __commit__[:8] }</a>'
+    )
+else:
+    commit_link = ""
```
weiji14, line 158: Since we have this if-else block, could we be smart and point to the tag on GitHub?

weiji14 (changes requested):

doc/conf.py
weiji14: We'll need to detect whether the commit string is a hash like `a1b2c4d4` or a version tag like `v0.X.Y`, but the idea is to have something like so (note, don't commit as it won't work):

```suggestion
    commit_link = f'<a href="{repository_url}/tree/{ __commit__ }">{ __commit__}</a>'
```

which would e.g. point to https://github.com/GenericMappingTools/pygmt/tree/v0.3.1

seisman: I was thinking about it too. It's OK to point to the specific tag, but the text would be "Revision v0.3.1", not "Release v0.3.1".

weiji14: "Revision v0.3.1" doesn't sound too bad actually (though it's not perfect either).

seisman: Screenshot:
image

It links to https://github.com/GenericMappingTools/pygmt/releases/tag/v0.3.1

weiji14: Does this need to be backported to the gh-pages branch?

seisman: Perhaps do the backport together with #1151, as mentioned in https://github.com/GenericMappingTools/pygmt/pull/1150#issuecomment-811053780.

weiji14: Ok, sounds like a good plan.

weiji14 (approved):

Issue #1217, opened 2021-04-20
Bump to GMT 6.2.0rc1


seisman:
The first release candidate of GMT 6.2.0 has been released (https://github.com/GenericMappingTools/gmt/releases/tag/6.2.0rc1) today.

Now GMT 6.2.0rc1 is available in the devel channel on conda-forge (https://anaconda.org/conda-forge/gmt/) and can be installed using:
```
conda install -c conda-forge/label/dev gmt=6.2.0rc1
```

We need to:

- [x] Bump the GMT version to 6.2.0rc1 used in the CI tests (#1218)
- [ ] Some tests fail and we need to fix them
- [ ] Add some new tests for [upstream bugs](https://github.com/GenericMappingTools/pygmt/labels/upstream) that were fixed in 6.2.0rc1

weiji14:
Bumping to GMT 6.2.0rc1 will also affect several existing PRs. I think we should go ahead with bumping the version as soon as possible though (so that we can test and fix any new bugs). Is there any PR we would like to merge/finish **before** bumping PyGMT to use GMT 6.2.0rc1? Otherwise it will get lost under a torrent of PRs relating to fixing broken unit tests :laughing:

I was actually thinking of updating https://github.com/GenericMappingTools/pygmt/blob/e4cb775346814dce46555a27c0c94f8e0aed6f04/.github/workflows/dvc-diff.yml to have a side-by-side before/after comparison of the images (instead of just having a single new image shown). Unsure if I have time this week to update that workflow (have a presentation and my PhD defence to prep for) but will see.

seisman:
> I was actually thinking of updating [`e4cb775`/.github/workflows/dvc-diff.yml](https://github.com/GenericMappingTools/pygmt/blob/e4cb775346814dce46555a27c0c94f8e0aed6f04/.github/workflows/dvc-diff.yml) to have a side-by-side before/after comparison of the images (instead of just having a single new image shown).

I can work on this.

seisman:
@GenericMappingTools/pygmt-maintainers please give #1219 and #1218 a fast review and approval, otherwise, we won't finish the migration before the next 6.2.0rc2 or 6.2.0 release.

meghanrjones:
> @GenericMappingTools/pygmt-maintainers please give #1219 and #1218 a fast review and approval, otherwise, we won't finish the migration before the next 6.2.0rc2 or 6.2.0 release.

[...]

PR #1221, opened 2021-04-21
Update baseline images for basemap

seisman:

Address #1217.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/dvc-diff.yml
+++ b/.github/workflows/dvc-diff.yml
@@ -65,7 +65,7 @@ jobs:
         echo -e "### Modified images\n" >> report.md
         # Upload new images
         while IFS= read -r line; do
-          cml-publish --title $line --md "$line" > modified_images_new.md < /dev/null
+          cml-publish --title $line --md "$line" >> modified_images_new.md < /dev/null
         done < modified_files.txt
 
         # Pull images in the master branch from cloud storage
@@ -73,7 +73,7 @@ jobs:
         dvc pull --remote upstream
         # Upload old images
         while IFS= read -r line; do
-          cml-publish --title $line --md "$line" > modified_images_old.md < /dev/null
+          cml-publish --title $line --md "$line" >> modified_images_old.md < /dev/null
         done < modified_files.txt
 
         # Append image report for modified images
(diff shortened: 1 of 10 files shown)


seisman:
All the new images make sense to me, except the `test_basemap_polar.png` one. I think it's a bug in GMT v6.2.0rc1, agree?

meghanrjones:
Yes, I agree that it shows this bug: https://github.com/GenericMappingTools/gmt/issues/5120. I do not think it should be allowed to pass with the new image.

seisman:
> Yes, I agree that it shows this bug: [GenericMappingTools/gmt#5120](https://github.com/GenericMappingTools/gmt/issues/5120). 

I don't think they're exactly the same bug. The equivalent GMT CLI is:
```
gmt basemap -R0/360/0/1000 -JP8c -Bafg -pdf map
```
It should NOT append any suffix (e.g., `E`) since it's a Cartesian polar projection.

image

> I do not think it should be allowed to pass with the new image.

Yes, I agree. I think I should keep the old `test_basemap_polar.png` image, and mark the test "xfail" for GMT 6.2.0rc1, right?

meghanrjones:
> > Yes, I agree that it shows this bug: [GenericMappingTools/gmt#5120](https://github.com/GenericMappingTools/gmt/issues/5120).
> 
> I don't think they're exactly the same bug. The equivalent GMT CLI is:
> 
> ```
> gmt basemap -R0/360/0/1000 -JP8c -Bafg -pdf map
> ```
> 
> It should NOT append any suffix (e.g., `E`) since it's a Cartesian polar projection.
> 

I agree that no units should be appended. My understanding is that right now GMT assumes that all angles given to -JP are longitude, which is a problem for both #5120 and the PyGMT test. Up to you if you want to post this in the same issue or a new one.

> 
> > I do not think it should be allowed to pass with the new image.
> 
> Yes, I agree. I think I should keep the old `test_basemap_polar.png` image, and mark the test "xfail" for GMT 6.2.0rc1, right?

Yes to me.

seisman:
> I agree that no units should be appended. My understanding is that right now GMT assumes that all angles given to -JP are longitude, which is a problem for both #5120 and the PyGMT test. Up to you if you want to post this in the same issue or a new one.

I think it's a new one. GMT v6.1.1 does not make this assumption. I'll post a new bug report.

meghanrjones (approved):

PR #1235, opened 2021-04-22
Update pygmt.config baseline images for GMT 6.2.0rc1

weiji14:

Updates baseline images for the `pygmt.config` tests.

Addresses #1217

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/baseline/test_config.png.dvc
+++ b/pygmt/tests/baseline/test_config.png.dvc
@@ -1,4 +1,4 @@
 outs:
-- md5: e473079025646ea07e44493f6ab396da
-  size: 33464
+- md5: 1f6d2bb449a5329d0208627dc293fd42
+  size: 19614
   path: test_config.png
--- a/pygmt/tests/baseline/test_config_font_annot.png.dvc
+++ b/pygmt/tests/baseline/test_config_font_annot.png.dvc
@@ -1,4 +1,4 @@
 outs:
-- md5: e1a3887151ea389de97f177c19488c5e
-  size: 80854
+- md5: 5db66e9f22ef896d1689513026036ec1
+  size: 69927
   path: test_config_font_annot.png
--- a/pygmt/tests/baseline/test_config_font_one.png.dvc
+++ b/pygmt/tests/baseline/test_config_font_one.png.dvc
@@ -1,4 +1,4 @@
 outs:
-- md5: b011328f9e88799e3af073da4f2b3835
-  size: 97085
+- md5: 6969208b1b1852a0b08cd67263966fdc
+  size: 83228
   path: test_config_font_one.png
--- a/pygmt/tests/baseline/test_config_format_time_map.png.dvc
+++ b/pygmt/tests/baseline/test_config_format_time_map.png.dvc
@@ -1,4 +1,4 @@
 outs:
-- md5: 3d1d843c5150f2f775fdb9382284fa31
-  size: 10906
+- md5: d95be055306d25aa619e816a9a4b147e
+  size: 7550
   path: test_config_format_time_map.png
--- a/pygmt/tests/baseline/test_config_map_annot_offset.png.dvc
+++ b/pygmt/tests/baseline/test_config_map_annot_offset.png.dvc
@@ -1,4 +1,4 @@
 outs:
-- md5: 40a89f9014ac054f5f196ace06a00361
-  size: 9389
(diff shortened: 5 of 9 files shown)


seisman:
`test_config_font_annot.png`: the annotations near the needle are wrong, right?

weiji14:
> `test_config_font_annot.png`: the annotations near the needle are wrong, right?

Yeah, why is it -4 instead of 4? Upstream bug?

meghanrjones:
> > `test_config_font_annot.png`: the annotations near the needle are wrong, right?
> 
> Yeah, why is it 4 instead of -4? Upstream bug?
> 
> https://github.com/GenericMappingTools/pygmt/blob/3a4f2f5ad546e2b7196d84099c126e1c5ca113a1/pygmt/tests/test_config.py#L55-L64
> 
> Same case with test_config_font_one.png

This is a consequence of FORMAT_GEO_MAP changing from ddd:mm:ss to ddd:mm:ssF. Feel free to open a GMT issue if you think FORMAT_GEO_MAP should not change these annotations.

weiji14:
> This is a consequence of FORMAT_GEO_MAP changing from ddd:mm:ss to ddd:mm:ssF. Feel free to open a GMT issue if you think FORMAT_GEO_MAP should not change these annotations.

Ok, I see at https://docs.generic-mapping-tools.org/6.2.0rc1/theme-settings.html#default-settings-for-build-in-themes that [`FORMAT_GEO_MAP`](https://docs.generic-mapping-tools.org/6.2.0rc1/gmt.conf.html#term-FORMAT_GEO_MAP) defaults to ddd:mm:ssF for the new 'modern' GMT theme. where `F` is the 'Encode sign using WESN suffix'.

image

@meghanrjones, could you link to the discussion on why `ddd:mm:ssF` is used? Just want to have an informed opinion on this.

meghanrjones:
> @meghanrjones, could you link to the discussion on why `ddd:mm:ssF` is used? Just want to have an informed opinion on this.

If there was a discussion about this, it would have happened in this PR: https://github.com/GenericMappingTools/gmt/pull/3344. I do not remember any discussion after Paul's initial suggestion that maps should have the indicators on them, but I may have missed it.

It might be possible to have an `auto` option for FORMAT_GEO_MAP similar to the new [auto selection](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#automatic-gmt-settings) for some other parameters, which would add WESN suffix for map annotations but not for the declination annotation. But, I haven't looked through how this would work when for example both basemap frame and compass are used so it's just an off-hand idea.

weiji14:
> > @meghanrjones, could you link to the discussion on why `ddd:mm:ssF` is used? Just want to have an informed opinion on this.
> 
> If there was a discussion about this, it would have happened in this PR: [GenericMappingTools/gmt#3344](https://github.com/GenericMappingTools/gmt/pull/3344). I do not remember any discussion after Paul's initial suggestion that maps should have the indicators on them, but I may have missed it.
> 
> It might be possible to have an `auto` option for FORMAT_GEO_MAP similar to the new [auto selection](https://docs.generic-mapping-tools.org/dev/cookbook/features.html#automatic-gmt-settings) for some other parameters, which would add WESN suffix for map annotations but not for the declination annotation. But, I haven't looked through how this would work when for example both basemap frame and compass are used so it's just an off-hand idea.

Thanks Meghan. Combing through the ~250 comments, the only reference I can find on `ddd.:mm:ssF` is Paul's initial comment at https://github.com/GenericMappingTools/gmt/pull/3344#issuecomment-629737443. I think it relates to his point on:

> 4. I think maps should have W-E-S-N indicators on them

which makes sense. Looking more closely at the changed figure, I think it's fine without the minus sign, the plot is actually correct because the declination annotation says 4°30**W** to indicate that it is pointing to the **W**est. Before (in GMT 6.1.1), it would just use -4°30'.

image

In this case, can I get an approval for this PR? Unless there are other concerns.

seisman:
I think it's weird to add **E** or **W** suffix to declination angles. Ping @PaulWessel to confirm whether it's a bug or expected.

seisman (approved):
After reading the [magnetic declination wikipedia page](https://en.wikipedia.org/wiki/Magnetic_declination), I realize that both `-4°30'` and `4°30'W` are correct expressions for magnetic declination. 

Now this PR is good to merge.

michaelgrund (approved):
Looks good to me.

Issue #1118, opened 2021-03-24
Rename the sizes parameter in plot/plot3d to be size


core-man:
**Description of the problem**

Both [plot](https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot) and [plot3d](https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot3d.html#pygmt.Figure.plot3d) have the `sizes` parameter that are used to plot symbols with different sizes. However, `size` is used in GMT documentation. Shall we rename this parameter to be `size`? If yes, I think we should do it as early as possible cause this parameter is often used in PyGMT.

Originally posted in https://github.com/GenericMappingTools/pygmt/issues/1076#issuecomment-804986728: 
> > like the `sizes` parameter (I dont' understand why we add a **s** at the end) of the `plot`/`plot3d` method
> 
> `plot` is one of the oldest wrappers in PyGMT, and the parameters are not carefully chosen. We had some discussions in the GMT Community meeting, and I think we agree that we will change `sizes` to `size`.

----

See the format of the [**-S**](https://docs.generic-mapping-tools.org/latest/plot.html#s) option in GMT `plot`:

> x y [ z ] [ size ] [ symbol-parameters ] [ symbol ]

See the format of the [**-S**](https://docs.generic-mapping-tools.org/latest/plot3d.html#s) option in GMT `plot3d`:

> x y z [ w ] [ size ] [ symbol-parameters ] [ symbol ]

GMT plot | PyGMT plot | GMT plot3d | PyGMT plot3d
:---: | :---:|:---: | :---:
*z* | `color` | *w* | `color`
*size* | `sizes` | *size* | `sizes`
*intensity* | `intensity` |*intensity* | `intensity` 
*symbol* | `symbol` (requested in #1076) | *symbol* | `symbol` (requested in #1076)

---

Now, `sizes` is used in the following files:
- `src/plot.py`
- `tests/test_plot.py`
- `examples/tutorials/plot.py`
- `examples/gallery/symbols/scatter.py`
- `examples/gallery/symbols/points_categorical.py`
- `src/plot3d.py`
- `tests/test_plot3d.py`
- `examples/gallery/3d_plots/scatter3d.py`

seisman:
As the `sizes` parameter is used a lot, we should be careful with backward-compatibility.

PR #1254, opened 2021-05-03
Figure.plot: Deprecate parameter "sizes" to "size" (remove in v0.6.0)

seisman:

- Rename the parameter `sizes` to `size` in Figure.plot()
- Updates tests and examples to use the new parameter name
- Use the deprecate_parameter decorator for backward-compatibility
- Add a test for checking 'sizes' backward compatibility

Address #1118.

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/tutorials/plot.py
+++ b/examples/tutorials/plot.py
@@ -63,7 +63,7 @@ fig.show()
 
 ########################################################################################
 # Notice that we didn't include the size in the ``style`` parameter this time, just the
-# symbol ``c`` (circles) and the unit ``c`` (centimeter). So in this case, the sizes
+# symbol ``c`` (circles) and the unit ``c`` (centimeter). So in this case, the size
 # will be interpreted as being in centimeters.
 #
 # We can also map the colors of the markers to the depths by passing an array to the
--- a/examples/tutorials/plot.py
+++ b/examples/tutorials/plot.py
@@ -45,7 +45,7 @@ fig.show()
 # parameter controls the outline of the symbols and the ``color`` parameter controls the fill.
 #
 # We can map the size of the circles to the earthquake magnitude by passing an array to
-# the ``sizes`` parameter. Because the magnitude is on a logarithmic scale, it helps to
+# the ``size`` parameter. Because the magnitude is on a logarithmic scale, it helps to
 # show the differences by scaling the values using a power law.
 
 fig = pygmt.Figure()
@@ -54,7 +54,7 @@ fig.coast(land="black", water="skyblue")
 fig.plot(
     x=data.longitude,
     y=data.latitude,
-    sizes=0.02 * (2 ** data.magnitude),
+    size=0.02 * (2 ** data.magnitude),
     style="cc",
     color="white",
     pen="black",
@@ -82,7 +82,7 @@ pygmt.makecpt(cmap="viridis", series=[data.depth_km.min(), data.depth_km.max()])
 fig.plot(
     x=data.longitude,
     y=data.latitude,
-    sizes=0.02 * 2 ** data.magnitude,
+    size=0.02 * 2 ** data.magnitude,
     color=data.depth_km,
     cmap=True,
     style="cc",
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -43,7 +43,7 @@ from pygmt.helpers import (
     t="transparency",
 )
 @kwargs_to_strings(R="sequence", c="sequence_comma", i="sequence_comma", p="sequence")
-def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
+def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
     r"""
     Plot lines, polygons, and symbols in 2-D.
 
@@ -78,7 +78,7 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
         Either a data file name or a 2d numpy array with the tabular data.
         Use parameter ``columns`` to choose which columns are x, y, color,
         and size, respectively.
-    sizes : 1d array
+    size : 1d array
         The sizes of the data points in units specified using ``style``.
         Only valid if using ``x``/``y``.
     direction : list of two 1d arrays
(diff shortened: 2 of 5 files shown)


examples/tutorials/plot.py
```diff
@@ -82,7 +82,7 @@
 fig.plot(
     x=data.longitude,
     y=data.latitude,
-    sizes=0.02 * 2 ** data.magnitude,
+    size=0.02 * 2 ** data.magnitude,
```
core-man, line 85: Line 66:
> So in this case, the sizes

`sizes` to be `size`?

core-man (approved):
All are good except one minor comment.

michaelgrund (approved):
Except the comment provided by @core-man, everything looks fine to me!

pygmt/src/plot.py
```diff
@@ -78,7 +80,7 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
         Either a data file name or a 2d numpy array with the tabular data.
         Use parameter ``columns`` to choose which columns are x, y, color,
         and size, respectively.
-    sizes : 1d array
+    size : 1d array
         The sizes of the data points in units specified using ``style``.
```
weiji14, line 84:
```suggestion
        The size of the data points in units specified using ``style``.
```

pygmt/src/plot.py
```diff
@@ -215,12 +217,12 @@ def plot(self, x=None, y=None, data=None, sizes=None, direction=None, **kwargs):
             )
         extra_arrays.append(kwargs["G"])
         del kwargs["G"]
-    if sizes is not None:
+    if size is not None:
         if kind != "vectors":
             raise GMTInvalidInput(
                 "Can't use arrays for sizes if data is matrix or file."
```
weiji14, line 223:
```suggestion
                "Can't use arrays for 'size' if data is a matrix or file."
```

pygmt/tests/test_plot.py
```diff
@@ -152,7 +153,7 @@ def test_plot_sizes(data, region):
     fig.plot(
```
weiji14, line 153: Can you also rename other instances of 'sizes' to 'size' in the `""" """` docstring? E.g. "Plot the data using z as sizes." becomes "Plot the data using z as size.". Also debating whether to keep the test names as `test_plot_*_sizes` or rename to `test_plot_*_size` (which would mean renaming a lot of *.png.dvc files).

examples/tutorials/plot.py
```diff
@@ -54,7 +54,7 @@
 fig.plot(
     x=data.longitude,
     y=data.latitude,
-    sizes=0.02 * (2 ** data.magnitude),
+    size=0.02 * (2 ** data.magnitude),
     style="cc",
     color="white",
     pen="black",
```
weiji14, line 60: Please also change 'sizes' to 'size' at Line 66 in this `examples/tutorials/plot.py` file.

pygmt/tests/test_plot.py
seisman: > E.g. "Plot the data using z as sizes." becomes "Plot the data using z as size."

In this sentence, "sizes" means "sizes" of symbols, not the `size` parameter. I think it is grammatically correct, right? 

BTW, we also use "Plot the data using z as colors" even though the parameter name is `color`.

examples/tutorials/plot.py
seisman: Similarly, at line 66, the sentence is about the symbol sizes, not the `size` parameter.

weiji14: The full paragraph from L64-66 reads:

```
# Notice that we didn't include the size in the ``style`` parameter this time, just the
# symbol ``c`` (circles) and the unit ``c`` (centimeter). So in this case, the sizes
# will be interpreted as being in centimeters.
```

So either change the `size` in L64 to `sizes`, or the `sizes` in L66 to `size`?

pygmt/tests/test_plot.py
weiji14: Hmm, I think either `sizes` or `size` is grammatically correct. Let's just keep it on `sizes` to reduce the `git diff` then.

seisman:
Since the PR title will be used in the release notes, I just changed the PR title to 

> Figure.plot: Deprecate parameter "sizes" to "size" (will be removed in v0.6.0)

to make it more readable. 

I expect to have a "consistent" PR title for parameter deprecations like this, so suggestions on the PR title format are also welcomed.

weiji14 (approved):
> Since the PR title will be used in the release notes, I just changed the PR title to
> 
> > Figure.plot: Deprecate parameter "sizes" to "size" (will be removed in v0.6.0)
> 
> to make it more readable.
> 
> I expect to have a "consistent" PR title for parameter deprecations like this, so suggestions on the PR title format are also welcomed.

Looks good. I would just change the 'removed' to 'remove' in the title, i.e. 'Figure.plot: Deprecate parameter "sizes" to "size" (remove in v0.6.0)'

pygmt/tests/test_plot.py
```diff
@@ -446,3 +447,28 @@ def test_plot_datetime():
...
+            y=data[:, 1],
+            sizes=0.5 * data[:, 2],
+            region=region,
+            projection="X10c",
+            style="cc",
+            color="blue",
+            frame="af",
+        )
+        assert len(w) == 1
+        assert issubclass(w[0].category, FutureWarning)
```
weiji14, line 473: Actually, would it be better to use `pytest.warns` here? See https://docs.pytest.org/en/6.2.x/warnings.html#warns.

weiji14: Might need to remove the `import warnings` line at the top.

```suggestion
    with pytest.warns(expected_warning=FutureWarning) as record:
        fig.plot(
            x=data[:, 0],
            y=data[:, 1],
            sizes=0.5 * data[:, 2],
            region=region,
            projection="X10c",
            style="cc",
            color="blue",
            frame="af",
        )
        assert len(record) == 1  # check that only one warning was raised
```

pygmt/tests/test_plot.py
```diff
@@ -4,6 +4,7 @@
 """
 import datetime
 import os
+import warnings
```
seisman, line 7:
```suggestion
import os
```

seisman, line 7:
```suggestion
```

PR #1255, opened 2021-05-03
Update Figure.meca 'offset' alias to match GMT 6.2.0rc1 syntax

### LINKED ISSUE #1094: Incorrect parameter in Figure.meca()

[Figure.meca](https://www.pygmt.org/dev/api/generated/pygmt.Figure.meca.html#pygmt.Figure.meca) defines an alias as ``C = offset``. Based on the description of the ``offset`` parameter, I think this alias should be for the **-A** option in GMT's meca (https://docs.generic-mapping-tools.org/dev/supplements/seis/meca.html#a); **-C** is for giving a cmap for the compressive part of the beachball. 

This is both a code and documentation issue, since there are these two checks for **-C** in kwargs (which should not be required for meca):

```
            # Add in plotting options, if given, otherwise add 0s
            for arg in plot_longitude, plot_latitude:
                if arg is None:
                    spec.append(0)
                else:
                    if "C" not in kwargs:
                        kwargs["C"] = True
                    spec.append(arg)
```

Relates to the proposed refactoring in https://github.com/GenericMappingTools/pygmt/pull/1002

seisman:

`meca` in GMT <=6.1.1 used **-C** option for offset.
Since https://github.com/GenericMappingTools/gmt/pull/4873, the **-C** option is used
to specify the CPT of the compressional part, and the old **-C** option is
renamed to **-A**, although it's still backward-compatible.

This PR changes the 'offset' alias from **C** to **A**, to match the GMT 6.2.0rc1
syntax.

Fixes #1094.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -99,7 +99,7 @@ def data_format_code(convention, component="full"):
     R="region",
     J="projection",
     B="frame",
-    C="offset",
+    A="offset",
     N="no_clip",
     V="verbose",
     X="xshift",
@@ -334,8 +334,8 @@ def meca(
                 if arg is None:
                     spec.append(0)
                 else:
-                    if "C" not in kwargs:
-                        kwargs["C"] = True
+                    if "A" not in kwargs:
+                        kwargs["A"] = True
                     spec.append(arg)
 
         # or assemble the 2D array for the case of lists as values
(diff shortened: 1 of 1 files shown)


meghanrjones (approved):

michaelgrund (approved):
Looks good to me!

pygmt/src/meca.py
```diff
@@ -99,7 +99,7 @@ def data_format_code(convention, component="full"):
     R="region",
     J="projection",
     B="frame",
-    C="offset",
+    A="offset",
```
weiji14, line 102: Could you maybe move A above B to make it alphabetically ordered?

seisman: Done in fca2520e443f6c0b4aeeca5719a725617188e8f6.

PR #1261, opened 2021-05-06
Wrap grdclip

willschlitzer:
Wraps the gmt function [grdclip](https://docs.generic-mapping-tools.org/6.1/grdclip.html).

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -55,7 +55,7 @@ def grdclip(grid, **kwargs):
         in.
     {R}
     above : str or list or tuple
-        *high/above*.
+        [*high*, *above*].
         Set all data[i] > *high* to *above*.
     below : str or list or tuple
         *low/below*.
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -25,8 +25,26 @@ from pygmt.helpers import (
 )
 @kwargs_to_strings(R="sequence")
 def grdclip(grid, **kwargs):
-    """
-    Grdclip.
+    r"""
+    Sets specific values, or values above/below a set number, to a given value.
+
+    Produce a new ``outgrid`` file clipped version of``grid``.
+
+    The parameters ``above`` and ``below`` allow for a given value to be set
+    for values above or below a set amount, respectively. This allows for
+    extreme values in a grid, such as points below a certain depth when
+    plotting Earth relief, to all be set to the same value.
+    subregion is specified with ``region``; the specified range must not exceed
+    the range of ``grid`` (but see ``extend``). If in doubt, run
+    :meth:`pygmt.grdinfo` to check range. Alternatively, define the subregion
+    indirectly via a range check on the node values or via distances from a
+    given point. Finally, you can give ``projection`` for oblique projections
+    to determine the corresponding rectangular ``region`` that will give a grid
+    that fully covers the oblique domain.
+
+    Full option list at :gmt-docs:`grdcut.html`
+
+    {aliases}
     """
     with GMTTempFile(suffix=".nc") as tmpfile:
         with Session() as lib:
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -1,5 +1,5 @@
 """
-grdclip - Clip the range of grid values.
+grdclip - Change the range and extremes of grid values.
 """
 
 import xarray as xr
@@ -32,9 +32,10 @@ from pygmt.helpers import (
 )
 def grdclip(grid, **kwargs):
     r"""
-    Sets specific values, or values above/below a set number, to a given value.
+    Sets values in a grid that meet certain criteria to a new value.
 
-    Produce a new ``outgrid`` file clipped version of``grid``.
+    Produce a clipped ``outgrid`` or xarray.Data Array version of the
+    input ``grid`` file.
 
     The parameters ``above`` and ``below`` allow for a given value to be set
     for values above or below a set amount, respectively. This allows for
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -1,5 +1,5 @@
 """
-Grdclip - Clip a grid.
+grdclip - Clip a grid.
 """
 
 import xarray as xr
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -58,14 +58,14 @@ def grdclip(grid, **kwargs):
         [*high*, *above*].
         Set all data[i] > *high* to *above*.
     below : str or list or tuple
-        *low/below*.
+        [*low*, *below*].
         Set all data[i] < *low* to *below*.
     interval : str or list or tuple
-        *low/high/between*.
+        [*low*, *high*, *between*].
         Set all data[i] >= *low* and <= *high* to *between*.
         Repeat the option for as many intervals as are needed.
     old: str or list or tuple
-        *old/new*.
+        [*old*, *new*].
         Set all data[i] == *old* to *new*. This is mostly useful when
         your data are known to be integer values. Repeat the option
         for as many replacements as are needed.
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -1,5 +1,5 @@
 """
-grdclip - Clip a grid.
+grdclip - Clip the range of grid values.
 """
 
 import xarray as xr
@@ -44,6 +44,7 @@ def grdclip(grid, **kwargs):
     Full option list at :gmt-docs:`grdclip.html`
 
     {aliases}
+
     Parameters
     ----------
     grid : str or xarray.DataArray
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -1,20 +1,32 @@
 """
-Grdclip - Wrap a grid
+Grdclip - Clip a grid.
 """
+
+import xarray as xr
 from pygmt.clib import Session
-from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
+from pygmt.helpers import (
+    GMTTempFile,
+    build_arg_string,
+    fmt_docstring,
+    kwargs_to_strings,
+    use_alias,
+)
+
 
 @fmt_docstring
 @use_alias(
     G="outgrid",
     R="region",
-    Z="z_subregion",
-    f="coltypes",
+    Sa="above",
+    Sb="below",
+    Si="interval",
+    Sr="old",
+    V="verbose",
 )
 @kwargs_to_strings(R="sequence")
 def grdclip(grid, **kwargs):
     """
-    Grdclip
+    Grdclip.
     """
     with GMTTempFile(suffix=".nc") as tmpfile:
         with Session() as lib:
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -63,12 +63,10 @@ def grdclip(grid, **kwargs):
     between : str or list or tuple
         [*low*, *high*, *between*].
         Set all data[i] >= *low* and <= *high* to *between*.
-        Repeat the option for as many intervals as are needed.
     new : str or list or tuple
         [*old*, *new*].
         Set all data[i] == *old* to *new*. This is mostly useful when
-        your data are known to be integer values. Repeat the option
-        for as many replacements as are needed.
+        your data are known to be integer values.
     {V}
     """
     with GMTTempFile(suffix=".nc") as tmpfile:
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -23,7 +23,13 @@ from pygmt.helpers import (
     Sr="old",
     V="verbose",
 )
-@kwargs_to_strings(R="sequence")
+@kwargs_to_strings(
+    R="sequence",
+    Sa="sequence",
+    Sb="sequence",
+    Si="sequence",
+    Sr="sequence",
+)
 def grdclip(grid, **kwargs):
     r"""
     Sets specific values, or values above/below a set number, to a given value.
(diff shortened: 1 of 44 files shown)


pygmt/src/grdclip.py
```diff
@@ -0,0 +1,83 @@
+"""
+Grdclip - Clip a grid.
```
seisman, line 2:
```suggestion
grdclip - Clip a grid.
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,83 @@
...
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    {R}
+    above : str
+        *high/above*.
+        Set all data[i] > *high* to *above*.
```
seisman, line 51: It's more Pythonic to be able to pass `above=(10, 20)` rather than `above="10/20"`.

willschlitzer: fixed in 5b9e83b

pygmt/tests/test_grdclip.py
```diff
@@ -0,0 +1,44 @@
...
+    return load_earth_relief(resolution="10m", region=[-5, 5, -5, 5])
+
+
+def test_grdclip_outgrid(grid):
+    """
+    Test the below and above parameters for grdclip and creates a test outgrid.
+    """
+    with GMTTempFile(suffix=".nc") as tmpfile:
+        result = grdclip(
+            grid=grid, outgrid=tmpfile.name, below="-1500/-1800", above="30/40"
```
weiji14, line 26: Check that list inputs work after 5b9e83b9f13d1347af9342d4535893300f21b003.
```suggestion
            grid=grid, outgrid=tmpfile.name, below=[-1500, -1800], above=[30, 40]
```

pygmt/tests/test_grdclip.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_grdclip_no_outgrid(grid):
+    """
+    Test the below and above parameters for grdclip with no set outgrid.
+    """
+    temp_grid = grdclip(grid=grid, below=[-1500, -1800], above=[30, 40])
+    result = grdinfo(grid=temp_grid, force_scan=0, per_column="n").strip().split()
+    assert int(result[4]) == -1800
+    assert int(result[5]) == 40
```
weiji14, line 44: Typically with unit tests, it is best to test just the function we want to test (i.e. `grdclip` here) and avoid calling other functions (e.g. `grdinfo`) if possible. This helps to prevent breakages in other functions (e.g. in `grdinfo`) causing breakages in the actual unit we're testing (i.e. `grdclip` here).

Since this test returns an `xarray.DataArray`, we can use the built-in min/max functions to get the information needed.

```suggestion
    assert temp_grid.dims == ("lat", "lon")
    assert temp_grid.gmt.gtype == 1  # Geographic grid
    assert temp_grid.gmt.registration == 1  # Pixel registration
    npt.assert_allclose(temp_grid.min(), -1800)
    npt.assert_allclose(temp_grid.max(), 40)
```

Oh, and I've also added a few more checks to ensure the dims, and GMT grid type and registration type is correct :smile:

pygmt/tests/test_grdclip.py
```diff
@@ -0,0 +1,44 @@
+"""
+Tests for grdclip.
+"""
+import os
+
+import pytest
```
weiji14, line 6:
```suggestion
import numpy.testing as npt
import pytest
```

pygmt/tests/test_grdclip.py
```diff
@@ -0,0 +1,44 @@
...
+        result = grdclip(
+            grid=grid, outgrid=tmpfile.name, below="-1500/-1800", above="30/40"
+        )
+        assert result is None  # return value is None
+        assert os.path.exists(path=tmpfile.name)  # check that outgrid exists
+        result = (
+            grdinfo(grid=tmpfile.name, force_scan=0, per_column="n").strip().split()
+        )
+    assert int(result[4]) == -1800
+    assert int(result[5]) == 40
```
weiji14, line 34: Ok with using `grdinfo` here since the output is a NetCDF file instead of an `xarray.DataArray` (though it would be nice to use `xarray` to read the grid to check the min/max values). As it is though, perhaps just add an inline comment to indicate that -1800 and 40 are the min and max values respectively.

```suggestion
        result = (
            grdinfo(grid=tmpfile.name, force_scan=0, per_column="n").strip().split()
        )
    assert int(result[4]) == -1800  # minimum value
    assert int(result[5]) == 40  # maximum value
```

weiji14 (commented):
Nice and minimal! I'll focus my feedback on just the unit tests.

pygmt/tests/test_grdclip.py
```diff
@@ -0,0 +1,47 @@
...
+from pygmt.datasets import load_earth_relief
+from pygmt.helpers import GMTTempFile
+
+
+@pytest.fixture(scope="module", name="grid")
+def fixture_grid():
+    """
+    Load the grid data from the sample earth_relief file.
+    """
+    return load_earth_relief(resolution="10m", region=[-5, 5, -5, 5])
```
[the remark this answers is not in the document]
willschlitzer: Switched to `01d` in e2027a6

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,89 @@
+"""
+grdclip - Clip a grid.
```
weiji14, line 2:
```suggestion
grdclip - Clip the range of grid values.
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,89 @@
...
+@kwargs_to_strings(
+    R="sequence",
+    Sa="sequence",
+    Sb="sequence",
+    Si="sequence",
+    Sr="sequence",
+)
+def grdclip(grid, **kwargs):
+    r"""
+    Sets specific values, or values above/below a set number, to a given value.
```
weiji14, line 35: 'Clip' can mean different things to different people. Might need to wordsmith this sentence a bit (don't commit it yet). 

```suggestion
    Sets values in a grid that meet a certain criteria to a new value.
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,89 @@
...
+
+    Produce a new ``outgrid`` file clipped version of``grid``.
```
weiji14, line 37: Just another suggestion, feel free to improve the wording a bit before committing.
```suggestion
    Produce a new ``outgrid`` file or xarray.DataArray grid which is a clipped
    version of the input ``grid``.
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,89 @@
...
+
+    The parameters ``above`` and ``below`` allow for a given value to be set
+    for values above or below a set amount, respectively. This allows for
+    extreme values in a grid, such as points below a certain depth when
+    plotting Earth relief, to all be set to the same value.
+
+    Full option list at :gmt-docs:`grdclip.html`
+
+    {aliases}
+    Parameters
```
weiji14, line 47: Add an extra newline here.
```suggestion
    {aliases}

    Parameters
```

weiji14 (commented):
The code implementation and unit tests of `grdclip` looks ok to me. The docstrings might need a couple more iterations of improvement though, will let the others chime in on this.

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,90 @@
...
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    {R}
+    above : str or list
+        *high/above*.
```
seisman, line 57: Should we rewrite it like this?
```suggestion
    above : tuple
        (*high*, *above*).
```

willschlitzer: Made an update in d233ab2

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    {R}
+    above : str or list or tuple
+        *high/above*.
```
seisman, line 58: Although *str* type is supported, the most Pythonic way is passing a tuple or a list. I prefer to the following syntax, rather than *high/above*.
```suggestion
        [*high*, *above*].
```

willschlitzer: @seisman Sounds good, but should we then include example code of how to input a string?

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    {R}
+    above : str or list or tuple
+        [*high*, *above*].
+        Set all data[i] > *high* to *above*.
+    below : str or list or tuple
+        [*low*, *below*].
+        Set all data[i] < *low* to *below*.
+    interval : str or list or tuple
```
seisman, line 63: Is `between` better than `interval`?
```suggestion
    between : str or list or tuple
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+        [*low*, *high*, *between*].
+        Set all data[i] >= *low* and <= *high* to *between*.
+        Repeat the option for as many intervals as are needed.
+    old: str or list or tuple
```
seisman, line 67: I think we should use *new*, rather than *old*, because for *high/above*, we use *above*.
```suggestion
    new : str or list or tuple
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+        in.
+    {R}
+    above : str or list or tuple
+        [*high*, *above*].
+        Set all data[i] > *high* to *above*.
+    below : str or list or tuple
+        [*low*, *below*].
+        Set all data[i] < *low* to *below*.
+    interval : str or list or tuple
+        [*low*, *high*, *between*].
```
seisman, line 64: Is *between* better than *interval*, because we use *above* for *high/above*.
```suggestion
    between : str or list or tuple
        [*low*, *high*, *between*].
```

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+    above : str or list or tuple
+        [*high*, *above*].
+        Set all data[i] > *high* to *above*.
+    below : str or list or tuple
+        [*low*, *below*].
+        Set all data[i] < *low* to *below*.
+    between : str or list or tuple
+        [*low*, *high*, *between*].
+        Set all data[i] >= *low* and <= *high* to *between*.
+        Repeat the option for as many intervals as are needed.
```
seisman, line 66: > Repeat the option for as many intervals as are needed.

It's impossible to repeat the option in Python.

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,91 @@
...
+    new : str or list or tuple
+        [*old*, *new*].
+        Set all data[i] == *old* to *new*. This is mostly useful when
+        your data are known to be integer values. Repeat the option
+        for as many replacements as are needed.
```
seisman, line 71: > Repeat the option for as many replacements as are needed.

It's impossible to repeat the option in Python.

pygmt/src/grdclip.py
weiji14: What if we could do it like frame (-B) in `fig.basemap` and pass in a list? Something like `between=['0/2/1', '3/5/4']`? Or a list of tuples like `between=[(0, 2 ,1), (3, 5 ,4)]` that expands to `-Sb0/2/1 -Sb3/5/4` in GMT.

seisman: `between=[(0, 2 ,1), (3, 5 ,4)]` makes sense to me, but it needs more coding afforts.

pygmt/src/grdclip.py
willschlitzer: fixed in 1046419

pygmt/src/grdclip.py
willschlitzer: This is fixed in 1046419. Regarding @weiji14's suggestion, I'm assuming that would require a change to `kwargs_to_strings()`? Assuming so, I would prefer for this current method to be accepted as is, and a future pull request can be submitted once `kwargs_to_strings` is updated.

weiji14: Actually I was thinking of handling this feature by changing `Sb='sequence'`. Perhaps implementing a `sequence_tuple`? But yes, this can be done in a separate enhancement PR.

pygmt/src/grdclip.py
```diff
@@ -0,0 +1,89 @@
...
+    Sa="sequence",
+    Sb="sequence",
+    Si="sequence",
+    Sr="sequence",
+)
+def grdclip(grid, **kwargs):
+    r"""
+    Sets values in a grid that meet certain criteria to a new value.
+
+    Produce a clipped ``outgrid`` or xarray.Data Array version of the
```
seisman, line 37:
```suggestion
    Produce a clipped ``outgrid`` or :class:`xarray.DataArray` version of the
```

seisman (approved):

pygmt/src/grdclip.py
seisman: > this can be done in a separate enhancement PR.

I agree.

weiji14 (approved):
Good work @willschlitzer! After you merge this in (remember to tidy the commit message), please open up an issue to handle the enhancement mentioned at https://github.com/GenericMappingTools/pygmt/pull/1261#discussion_r633053990. Thanks!

PR #1276, opened 2021-05-17
Wrap grdfill

willschlitzer:
This pull request wraps the `grdfill` module.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -44,7 +44,7 @@ def grdfill(grid, **kwargs):
         Specify the hole-filling algorithm to use.  Choose from **c** for
         constant fill and append the constant value, **n** for nearest
         neighbor (and optionally append a search radius in
-        pixels [default radius is :math:`r^2 = \sqrt{X^2 + Y^2}`,
+        pixels [default radius is `r^2 = sqrt(X^2 + Y^2)`,
         where (*X,Y*) are the node dimensions of the grid]).
     {R}
 
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -44,7 +44,7 @@ def grdfill(grid, **kwargs):
         Specify the hole-filling algorithm to use.  Choose from **c** for
         constant fill and append the constant value, **n** for nearest
         neighbor (and optionally append a search radius in
-        pixels [default radius is `r^2 = sqrt(X^2 + Y^2)`,
+        pixels [default radius is :math:`r^2 = \sqrt{{ X^2 + Y^2 }}`,
         where (*X,Y*) are the node dimensions of the grid]).
     {R}
 
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -22,7 +22,12 @@ from pygmt.helpers import (
 @kwargs_to_strings(R="sequence")
 def grdfill(grid, **kwargs):
     r"""
-    Fill blank areas from a grid.
+    Fill blank areas from a grid file.
+
+    Read a grid that presumably has unfilled holes that the user
+    wants to fill in some fashion.  Holes are identified by NaN values but
+    this criteria can be changed.  There are several different algorithms that
+    can be used to replace the hole values.
 
     Full option list at :gmt-docs:`grdfill.html`
 
@@ -35,6 +40,12 @@ def grdfill(grid, **kwargs):
     outgrid : str or None
         The name of the output netCDF file with extension .nc to store the grid
         in.
+    mode : str
+        Specify the hole-filling algorithm to use.  Choose from **c** for
+        constant fill and append the constant value, **n** for nearest
+        neighbor (and optionally append a search radius in
+        pixels [default radius is :math:`r^2 = \sqrt{X^2 + Y^2}`,
+        where (*X,Y*) are the node dimensions of the grid]).
     {R}
 
     Returns
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -25,8 +25,8 @@ def grdfill(grid, **kwargs):
     Fill blank areas from a grid file.
 
     Read a grid that presumably has unfilled holes that the user
-    wants to fill in some fashion.  Holes are identified by NaN values but
-    this criteria can be changed.  There are several different algorithms that
+    wants to fill in some fashion. Holes are identified by NaN values but
+    this criteria can be changed. There are several different algorithms that
     can be used to replace the hole values.
 
     Full option list at :gmt-docs:`grdfill.html`
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -1,0 +1,66 @@
+"""
+grdfill - Fill blank areas from a grid.
+"""
+
+import xarray as xr
+from pygmt.clib import Session
+from pygmt.helpers import (
+    GMTTempFile,
+    build_arg_string,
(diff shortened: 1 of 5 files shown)


willschlitzer:
@GenericMappingTools/pygmt-contributors Any ideas for how to test this? I have downloaded grid files from scientific cruises to ensure I have gaps in the data that need to be filled, but I'm not sure how to write a test that retrieves an unfilled grid from something like `pygmt.datasets`.

meghanrjones:
> @GenericMappingTools/pygmt-contributors Any ideas for how to test this? I have downloaded grid files from scientific cruises to ensure I have gaps in the data that need to be filled, but I'm not sure how to write a test that retrieves an unfilled grid from something like `pygmt.datasets`.

You can manual set one or more of the data values in grids from `pygmt.datasets` to NaN to test the result from `grdfill`. For example:
```
import numpy as np
import pytest
import xarray as xr
from pygmt import grdfill
from pygmt.datasets import load_earth_relief


@pytest.fixture(scope="module", name="grid")
def fixture_grid():
    """
    Load the grid data from the sample earth_relief file and set value(s) to NaN.
    """
    grid = load_earth_relief(registration="pixel")
    grid[10,10] = np.nan
    return grid

def test_grdfilter_dataarray_in_dataarray_out(grid):
    """
    grdfill an input DataArray, and output as DataArray.
    """
    result = grdfill(grid=grid,mode="c20")
    # check information of the output grid
    assert isinstance(result, xr.DataArray)
    assert result[10,10] == 20

````

**Review by seisman — APPROVED** [review]
Looks good to me.

**Review by michaelgrund — APPROVED** [review]
Looks good to me. Great work!

File: pygmt/src/grdfill.py [hunk 1]
```diff
@@ -0,0 +1,77 @@
...
+    R="region",
+)
+@kwargs_to_strings(R="sequence")
+def grdfill(grid, **kwargs):
+    r"""
+    Fill blank areas from a grid file.
+
+    Read a grid that presumably has unfilled holes that the user
+    wants to fill in some fashion.  Holes are identified by NaN values but
+    this criteria can be changed.  There are several different algorithms that
```
core-man, line 29:
```suggestion
    wants to fill in some fashion. Holes are identified by NaN values but
    this criteria can be changed. There are several different algorithms that
```

pygmt/src/grdfill.py
```diff
@@ -0,0 +1,77 @@
...
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    mode : str
+        Specify the hole-filling algorithm to use.  Choose from **c** for
+        constant fill and append the constant value, **n** for nearest
+        neighbor (and optionally append a search radius in
+        pixels [default radius is `r^2 = sqrt(X^2 + Y^2)`,
```
core-man, line 47: LaTeX formula?
```suggestion
        pixels [default radius is :math:`r^2 = sqrt(X^2 + Y^2)`,
```

core-man (commented):
Looks great. Only two minor comments.

core-man (approved):

pygmt/src/grdfill.py
seisman: If you want to use the `math` directive, it should be:
```
:math:`r^2 = \sqrt{X^2 + Y^2}`
```

core-man: > If you want to use the `math` directive, it should be:
> 
> ```
> :math:`r^2 = \sqrt{X^2 + Y^2}`
> ```

updated~ Thanks

meghanrjones: @willschlitzer previously had this in there - it did not work because `fmt_docstring` tries to insert common text into the brackets. So, it either needs to stay as non-sphinx math or `fmt_docstring` needs to be modified.

seisman: If we want LaTeX math, we may need to use double curly braces `:math:`r^2 = \sqrt{{ X^2 + Y^2 }}`` . See https://stackoverflow.com/questions/5466451

pygmt/tests/test_grdfill.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.fixture(scope="module", name="grid")
+def fixture_grid():
+    """
+    Load the grid data from the sample earth_relief file and set value(s) to
+    NaN.
+    """
+    grid = load_earth_relief(registration="pixel")
+    grid[10, 10] = np.nan
```
weiji14, line 21: Prefer to use a smaller 10x10 region instead of a 180x360 pixel grid, and set more than one pixel to NaN.

```suggestion
    grid = load_earth_relief(registration="pixel", region=[-5, 5, -5, 5])
    grid[3:5, 3:5] = np.nan
    grid[5:7, 5:7] = np.inf
```

I've also tried setting a 2x2 group of pixels to [`np.inf`](https://numpy.org/devdocs/reference/constants.html#numpy.inf) (infinity), and these don't seem to be filled in by `grdfill`, which is to be expected I think?

pygmt/tests/test_grdfill.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_grdfill_dataarray_out(grid):
+    """
+    grdfill with a DataArray output.
+    """
+    result = grdfill(grid=grid, mode="c20")
+    # check information of the output grid
+    assert isinstance(result, xr.DataArray)
+    assert result[10, 10] == 20
```
weiji14, line 32: Add some extra checks to see that the whole grid does not have NaN values.

```suggestion
    assert result[4, 4] == 20
    assert result[5, 5] == np.inf
    assert not result.isnull().all()  # check that no NaN values exists
    assert result.gmt.gtype == 1  # Geographic grid
    assert result.gmt.registration == 1  # Pixel registration
```

pygmt/tests/test_grdfill.py
```diff
@@ -0,0 +1,44 @@
...
+def test_grdfill_file_out(grid):
+    """
+    grdfill with an outgrid set.
+    """
+    with GMTTempFile(suffix=".nc") as tmpfile:
+        result = grdfill(grid=grid, mode="c20", outgrid=tmpfile.name)
+        assert result is None  # return value is None
+        assert os.path.exists(path=tmpfile.name)  # check that outgrid exists
+        result = grdinfo(tmpfile.name, per_column=True).strip()
+        assert result == "-180 180 -90 90 -8182 5651.5 1 1 360 180 1 1"
```
weiji14, line 44:
```suggestion
        assert result == "-5 5 -5 5 -5130.5 inf 1 1 10 10 1 1"
```

weiji14 (commented):
Just some suggestions to improve the unit tests (smaller grid for speed + more robust checks) :grin: I've also thrown in some `np.inf` values just to see if GMT would fill them in.

> What is the preferred was for denoting required parameters? I think there should be some indication in the method docs that `mode` (**-A**) is required.

Perhaps just a one-line description in the docstring wll suffice?

pygmt/src/grdfill.py
willschlitzer: > If we want LaTeX math, we may need to use double curly braces
> 
> ```
> :math:`r^2 = \sqrt{{ X^2 + Y^2 }}`
> ```
> 
> See https://stackoverflow.com/questions/5466451

This appears to be working!

Issue #1099, opened 2021-03-22
Improve input/output options for blockm* and grdtrack


maxrjones:
**Description of the desired feature**

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/1092#discussion_r598541899_

- [x] Improve the **input** options for blockmedian, blockmode, and blockmean from only supporting pandas.DataFrame or file to supporting a file, matrix, pandas.DataFrame, or x/y/z.
- [ ] Improve the **output** options to support pandas.DataFrame or a file, independent of the input choice (current implementation requires ``outfile`` if a data file is provided as the input).

**Are you willing to help implement and maintain this feature?** Yes

weiji14:
I'll take the blame for this one. Just for context, the " pd.DataFrame in, pd.DataFrame out; filename in, filename out" restriction was added in a37fb0aa3fef025eb1c723039060caa0489fcab1 when I was wrapping the `grdtrack` function at #308. Some of the `blockmedian` code was based on `grdtrack`, and was not really questioned ever since.

So we should also update the `grdtrack` code as well to have the same I/O design. I actually had a friend using `grdtrack` last week who would have loved to have file-in and pandas.DataFrame-out for `grdtrack` so I'll work on refactoring `grdtrack` while @meghanrjones can work on the `blockm*` I/O refactor.

meghanrjones:
@weiji14, do you mind if I work on a PR for x/y input for grdtrack?

meghanrjones:
Just an update - I will hold off until after https://github.com/GenericMappingTools/pygmt/pull/1353 and https://github.com/GenericMappingTools/pygmt/pull/1336 are completed to submit the x/y input PR for grdtrack.

seisman:
The issue will be tracked in #2731 and #1318.

PR #1280, opened 2021-05-20
Refactor blockm* to use virtualfile_from_data and improve i/o

meghanrjones:

This PR makes the changes corresponding to #1189 for blockmean and blockmedian.

> Enables blockm* to work with table-like inputs besides pandas.DataFrame. Also, the outfile parameter has become optional, and the output data will be loaded as a pandas.DataFrame when outfile is unset. Unit tests have been updated accordingly too.

Related to #949. Addresses #1099

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_blockmean.py
+++ b/pygmt/tests/test_blockmean.py
@@ -23,8 +23,6 @@ def test_blockmean_input_dataframe():
     assert output.shape == (5849, 3)
     npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
 
-    return output
-
 
 def test_blockmean_input_table_matrix():
     """
@@ -37,8 +35,6 @@ def test_blockmean_input_table_matrix():
     assert isinstance(output, pd.DataFrame)
     assert output.shape == (5849, 3)
     npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
-
-    return output
 
 
 def test_blockmean_wrong_kind_of_input_table_grid():
@@ -70,8 +66,6 @@ def test_blockmean_input_filename():
         assert output.shape == (5849, 3)
         npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
 
-    return output
-
 
 def test_blockmean_without_outfile_setting():
     """
@@ -81,5 +75,3 @@ def test_blockmean_without_outfile_setting():
     assert isinstance(output, pd.DataFrame)
     assert output.shape == (5849, 3)
     npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
-
-    return output
--- a/pygmt/tests/test_blockmean.py
+++ b/pygmt/tests/test_blockmean.py
@@ -26,16 +26,19 @@ def test_blockmean_input_dataframe():
     return output
 
 
-def test_blockmean_wrong_kind_of_input_table_matrix():
+def test_blockmean_input_table_matrix():
     """
     Run blockmean using table input that is not a pandas.DataFrame but still a
     matrix.
     """
     dataframe = load_sample_bathymetry()
-    invalid_table = dataframe.values
-    assert data_kind(invalid_table) == "matrix"
-    with pytest.raises(GMTInvalidInput):
-        blockmean(table=invalid_table, spacing="5m", region=[245, 255, 20, 30])
+    table = dataframe.values
+    output = blockmean(table=table, spacing="5m", region=[245, 255, 20, 30])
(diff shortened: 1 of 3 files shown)


pygmt/tests/test_blockmean.py
```diff
@@ -26,16 +26,19 @@ def test_blockmean_input_dataframe():
...
-    assert data_kind(invalid_table) == "matrix"
-    with pytest.raises(GMTInvalidInput):
-        blockmean(table=invalid_table, spacing="5m", region=[245, 255, 20, 30])
+    table = dataframe.values
+    output = blockmean(table=table, spacing="5m", region=[245, 255, 20, 30])
+    assert isinstance(output, pd.DataFrame)
+    assert output.shape == (5849, 3)
+    npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
+
+    return output
```
seisman, line 41: No need to return the output in tests, right?
```suggestion
    npt.assert_allclose(output.iloc[0], [245.888877, 29.978707, -384.0])
```

meghanrjones: All removed in https://github.com/GenericMappingTools/pygmt/pull/1280/commits/fb0d9a0d9bdc3e1e4e4a13f36cffa74a84eaf58b

pygmt/tests/test_blockmean.py
```diff
@@ -23,19 +23,18 @@ def test_blockmean_input_dataframe():
...
-    return output
 
-
-def test_blockmean_wrong_kind_of_input_table_matrix():
+def test_blockmean_input_table_matrix():
     """
     Run blockmean using table input that is not a pandas.DataFrame but still a
     matrix.
     """
     dataframe = load_sample_bathymetry()
```
seisman, line 32: Please use [`pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html) so that the `dataframe` can be reused. 

See the `test_plot.py` for examples if it's new to you:
https://github.com/GenericMappingTools/pygmt/blob/cc26838d5183e3c71ec1374da5915983ee6ca3ae/pygmt/tests/test_plot.py#L19-L25

meghanrjones: Fixed in https://github.com/GenericMappingTools/pygmt/pull/1280/commits/cf7383741156a4a3664ed687b81daf1e86700a64

weiji14:
The docstring for `blockmean` and `blockmedian` also needs to be updated to mention that other table-like inputs are accepted besides `pandas.DataFrame`.

meghanrjones:
> The docstring for `blockmean` and `blockmedian` also needs to be updated to mention that other table-like inputs are accepted besides `pandas.DataFrame`.

Since geopandas is now supported and included in {table-like} filler text, do I also need to add tests for geopandas input?

seisman:
> Since geopandas is now supported and included in {table-like} filler text, do I also need to add tests for geopandas input?

I think it's unnecessary. The geopandas support is independent with any specific modules and we already have it tested in the `test_geopandas.py` file.

seisman (approved):
Looks good to me.

weiji14 (approved):
Thanks Meghan for the work, I think we can close #1099 with this PR too no?

> > Since geopandas is now supported and included in {table-like} filler text, do I also need to add tests for geopandas input?
> 
> I think it's unnecessary. The geopandas support is independent with any specific modules and we already have it tested in the `test_geopandas.py` file.

Agree that we don't need a `geopandas` input test here.

However, this got me to think about how the PyGMT I/O should work if a `geopandas.GeoDataFrame` is used as an input to `blockm*`. Should we yield a `pandas.DataFrame` output, or would users prefer a `geopandas.GeoDataFrame` to be returned here? Something to think about in a separate issue.

meghanrjones:
> I think we can close #1099 with this PR too no?
> 
Yes, it is now linked.

> However, this got me to think about how the PyGMT I/O should work if a `geopandas.GeoDataFrame` is used as an input to `blockm*`. Should we yield a `pandas.DataFrame` output, or would users prefer a `geopandas.GeoDataFrame` to be returned here? Something to think about in a separate issue.

Some users would probably like the option. This could be a good SciPy sprint issue.

seisman:
> I think we can close #1099 with this PR too no?

What's still missing is `x/y/z` type input like other functions.

weiji14:
> > I think we can close #1099 with this PR too no?
> 
> What's still missing is `x/y/z` type input like other functions.

Ah ok, I've unlinked #1099 from being closed. We'll handle the x/y/z inputs for `blockmean`, `blockmedian` and `grdtrack` in a separate PR. Feel free to merge this one in Meghan since tests are passing now.

Issue #48, opened 2017-09-19
List of GMT aliases


leouieda:
Paul compiled this list of aliases and GMT vocabulary. We should include this somewhere in the docs and in the code somewhere to avoid repetition.

GMT Vocabulary for Python Aliases
---------------------------------------------

The idea is to agree on a set of nouns that can serve as the Python aliases for GMT common options as well as commonly used options (such as pens, color).

Common options:

Option   | Alias
:------: | :---------
B        | frame
Bx,By,Bz | axes settings
J        | projection, proj
P        | portrait [PS only - goes away I think]
R        | region
U        | timestamp
V        | verbosity
X        | xshift
y        | yshift
b        | binary_in, binary_out, binary
d        | nodata
e        | search
f        | coltypes,type?
g        | gap
h        | header
i        | cols_in
n        | bc
o        | cols_out
p        | view
r        | reg, registration
s        | skip
t        | transparency
x        | cores

Common themes that could use fixed aliases

Option   | Alias
:------: | :---------
G        | fill,color,clip
W        | pen, thickness,color,style
C        | cpt
F        | font=, justification=, offset=,

weiji14:
Crossreferencing https://github.com/GenericMappingTools/gmt/issues/2079 and https://github.com/GenericMappingTools/gmt/issues/2079. I see that https://github.com/GenericMappingTools/gmt/pull/230 was merged last year (Dec 2019), does that mean there is a standard source for long options now?

seisman:
https://github.com/GenericMappingTools/gmt/pull/230 was merged, but the long name feature is disabled by default. So the feature is still experimental and the long names can still change (for example, GMT uses `verbosity`, but PyGMT uses `verbose` for **-V**).

meghanrjones:
I think we can solve this by adding a link to https://docs.generic-mapping-tools.org/dev/devdocs/long_options.html in the contributing guide.

seisman:
Why do we want to add the link in the contributing guide?

Should we just wrap up all aliases in the `COMMON_OPTIONS` dict?

https://github.com/GenericMappingTools/pygmt/blob/1fb9e2e4e523baaf3aaedb90d71872ede6d03bb7/pygmt/helpers/decorators.py#L15

meghanrjones:
> Why do we want to add the link in the contributing guide?
> 
> Should we just wrap up all aliases in the `COMMON_OPTIONS` dict?
> 
[...]

Issue #256, opened 2018-12-07
Fail for invalid parameters


seisman:
**Description of the problem**

We should be more careful to pass unrecognized keyword parameters to gmt command.

**Full code that generated the error**

```python
import gmt

# Load sample earthquake data in a pandas.DataFrame
quakes = gmt.datasets.load_usgs_quakes()

# Load the builtin Earth relief grid as an xarray.DataArray.
relief = gmt.datasets.load_earth_relief(resolution="30m")

# The Figure object controls all plotting functions
fig = gmt.Figure()
# Setup a map with a global region, a Mollweide projection, and automatic ticks
fig.basemap(region="g", projection="W200/8i", frame=True)
# Plot the Earth relief grid in pseudo-color.
fig.grdimage(relief, cmap="geo")
# Plot earthquakes as circles. Size maps to magnitude and color to depth.
fig.plot(x=quakes.longitude, y=quakes.latitude, sizes=0.01*2**quakes.mag,
         colors=quakes.depth/quakes.depth.max(), cmap="viridis", style="cc")
# Show a preview of the image (inline if in a Jupyter notebook).
fig.show()
```

The code above is the same as the code on gmt-python's homepage, except that in fig.plot(), I use the keyword `colors` instead of `color`. I know it's a user's mistake, but gmt-python reports some strange errors, which is very confusing.

**Full error message**

```
gmt-python-session [ERROR]: Current panel has row or column outsiden range!
psxy [ERROR]: Found no history for option -c
[Session gmt-python-session (9)]: Error returned from GMT API: GMT_OPTION_HISTORY_ERROR (63)
---------------------------------------------------------------------------
GMTCLibError                              Traceback (most recent call last)
<ipython-input-1-5a482bf0dd97> in <module>()
     15 # Plot earthquakes as circles. Size maps to magnitude and color to depth.
     16 fig.plot(x=quakes.longitude, y=quakes.latitude, sizes=0.01*2**quakes.mag,
---> 17          colors=quakes.depth/quakes.depth.max(), cmap="viridis", style="cc")
     18 # Show a preview of the image (inline if in a Jupyter notebook).
     19 fig.show()

~/Gits/gmt/gmt-python/gmt/helpers/decorators.py in new_module(*args, **kwargs)
    197                 if alias in kwargs:
    198                     kwargs[arg] = kwargs.pop(alias)
--> 199             return module_func(*args, **kwargs)
    200
    201         new_module.aliases = aliases

[...]

PR #1281, opened 2021-05-20
 Add decorator to simplify wrapping of common options 

meghanrjones:

This suggestion adds a decorator for including common options in pygmt methods without needing to include a dictionary pair for each option in `use_alias` and placeholders for each option in the docstrings. It works by taking a list of the common options available to the method, adding them to the aliases attribute, and inserting placeholders into a {common} placeholder in the docstring.

I have adjusted the blockm* functions to use the new decorator as an example to show that the usage and resulting documentation are the same as before.

I think this would be a helpful addition for two reasons:
1. It would simplify the completion of wrapping the remaining common options (https://github.com/GenericMappingTools/pygmt/issues/48) by greatly reducing the number of repeated lines.
2. It would simplify the wrapping of new modules, because the alias name is taken from `COMMON_OPTIONS` in `pygmt/helpers/decorators.py` rather than adding the keyword value pair separately for each module.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/blockm.py
+++ b/pygmt/src/blockm.py
@@ -12,6 +12,7 @@ from pygmt.helpers import (
     fmt_docstring,
     kwargs_to_strings,
     use_alias,
+    use_common,
 )
 
 
@@ -71,13 +72,10 @@ def _blockm(block_method, table, outfile, **kwargs):
 
 
 @fmt_docstring
+@use_common(options=["V", "a", "f", "r"])
 @use_alias(
     I="spacing",
     R="region",
-    V="verbose",
-    a="aspatial",
-    f="coltypes",
-    r="registration",
 )
 @kwargs_to_strings(R="sequence")
 def blockmean(table, outfile=None, **kwargs):
@@ -113,10 +111,7 @@ def blockmean(table, outfile=None, **kwargs):
         Required if ``table`` is a file. The file name for the output ASCII
         file.
 
-    {V}
-    {a}
-    {f}
-    {r}
+    {common}
 
     Returns
     -------
@@ -132,13 +127,10 @@ def blockmean(table, outfile=None, **kwargs):
 
 
 @fmt_docstring
+@use_common(options=["V", "a", "f", "r"])
 @use_alias(
     I="spacing",
     R="region",
-    V="verbose",
-    a="aspatial",
-    f="coltypes",
-    r="registration",
 )
 @kwargs_to_strings(R="sequence")
 def blockmedian(table, outfile=None, **kwargs):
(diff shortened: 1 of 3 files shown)


pygmt/src/blockm.py
```diff
@@ -132,13 +127,10 @@ def blockmean(table, outfile=None, **kwargs):
 
 
 @fmt_docstring
+@use_common(options=["V", "a", "f", "r"])
```
weiji14, line 130: My first impression is that this reduces some readability. What about using the full name as in `@use_common(options=["verbose", "aspatial", "coltypes", "registration"])`? Or do you think it's ok to use the short alias as is (I realize that it's easier to lookup the dictionary using `COMMON_OPTIONS[option]` where option is the short alias, compared to doing a reverse lookup).

pygmt/src/blockm.py
```diff
@@ -132,13 +127,10 @@ def blockmean(table, outfile=None, **kwargs):
...
 @use_alias(
     I="spacing",
     R="region",
-    V="verbose",
-    a="aspatial",
-    f="coltypes",
-    r="registration",
```
weiji14, line 141: Another question I have is the placement of common options like region (R) and frame (B). The current implementation seems to place them all at the very end after the non-common options. Is the `@use_common` decorator meant to be used only for parameters like `verbose`/`aspatial`/`panel`/etc placed at the end? Or do we want to use them for other common options like `region`/`frame` too?

I guess what I'm trying to ask is when we should use `@use_alias` vs `@use_common` :slightly_smiling_face:

weiji14 (commented):
Thanks for starting this Meghan! I have just two discussion points for now. Also converted this PR to draft mode to save on CI resources.

pygmt/src/blockm.py
meghanrjones: It could be an option to write a decorator that sorts the parameters section of the docstrings alphabetically (I would enjoy that, but maybe I am getting carried away here). Otherwise, I guess any common options that should go at the top or need unique descriptions for the given module (e.g., `region` in blockm*) could use `use_alias` while less important ones could use `use_common`.

pygmt/src/blockm.py
meghanrjones: It would be helpful to get more opinions on this. You are correct that the current implementation is simpler but less readable. Either seems fine to me, but I do not have a great perspective on how problematic the readability could be for future contributors/maintainers who have more experience with PyGMT than GMT.

meghanrjones:
I'm closing this PR since it hasn't gotten a lot of traction. Apart from saving a bit of work, the main motivation was to make sure the common options have consistent alias names. Rather than this solution, we could instead add a check for common option aliases when addressing https://github.com/GenericMappingTools/pygmt/issues/256.

Issue #1203, opened 2021-04-11
Let PyGMT functions support tab completion


seisman:
Jupyter and most modern text editors support tab autocompletion, useful to long function and parameter names. However, currently most PyGMT functions don't support tab autocompletion, since we use the alias system (`use_alias`), `*args`, and `**kwargs`.

The only way I know to enable tab autocompletion is using normal Python parameters instead of relying on the `**kwargs` and `use_alias` decorator.

The solution seems straightforward. For example, for the [`pygmt.info()`](https://github.com/GenericMappingTools/pygmt/blob/master/pygmt/src/info.py) function, the current function is defined as:
```python
@fmt_docstring
@use_alias(
    C="per_column",
    I="spacing",
    T="nearest_multiple",
    V="verbose",
    a="aspatial",
    f="coltypes",
    r="registration",
)
@kwargs_to_strings(I="sequence")
def info(table, **kwargs):
```

To make it support tab autocompletion, we just need to change it to:
```python
@fmt_docstring
@use_alias(
    C="per_column",
    I="spacing",
    T="nearest_multiple",
    V="verbose",
    a="aspatial",
    f="coltypes",
    r="registration",
)
@kwargs_to_strings(I="sequence")
def info(
    table,
    per_column=False,
    spacing=None,
    nearest_multiple=None,
    verbose=None,
    aspatial=None,
    coltypes=None,
    registration=None,
    **kwargs
):
```
Since all parameters are defined as normal Python parameters, tab autocompletion should work. The new function definition still keeps the `**kwargs` parameter, so that single-letter options are still supported (but we are likely to disallow single-letter options in future releases). Thus, this would be **backward compatible**.

The screenshots below show how the autocompletion work after making the changes in PR #1202:

| Current `pygmt.info()` | Updated `pygmt.info()` |
|---|---|
| ![image](https://user-images.githubusercontent.com/3974108/114293337-454c4c00-9a63-11eb-840d-104fbfd3ef4d.png) | ![image](https://user-images.githubusercontent.com/3974108/114293309-0c13dc00-9a63-11eb-8633-aa18165f3747.png) |
| ![image](https://user-images.githubusercontent.com/3974108/114293345-4da48700-9a63-11eb-9123-b6c16cf1a1a6.png) | ![image](https://user-images.githubusercontent.com/3974108/114293317-1504ad80-9a63-11eb-94b4-4879ca6ebb9e.png) |

weiji14:
[...]

PR #1282, opened 2021-05-21
Support tab completion in Jupyter by inserting aliases into the method signature

meghanrjones:

This PR adds a decorator that adds aliases as parameters to the signature of PyGMT methods to support tab completion. It is motivated by [this suggestion](https://github.com/GenericMappingTools/pygmt/issues/1203#issuecomment-817378035), but with a different implementation. The decorator was added to blockmean and blockmedian as examples. (edit: the decorator was added to all methods that use `use_alias`, but this can easily be reverted if needed). Here is what the result looks like in the docs and a jupyter notebook:
image
image

I tested that tab completing the alias keywords works in a Jupyter notebook, but not in any other environments.

Addresses #1203

**Reminders**

**Slash Commands**


seisman:
Great work! For me, it works well in IPython console, Jupyter notebook, but not in VSCode.

pygmt/helpers/decorators.py
```diff
@@ -327,6 +328,30 @@ def new_module(*args, **kwargs):
     return alias_decorator
 
 
+def insert_alias(module_func):
```
seisman, line 331: Perhaps we should move the code into the `use_alias` decorator, because we only need to "insert aliases into the signature" when aliases are defined by `@use_alias`? It also means we all function support tab completion without any further actions (i.e., `adding @insert_alias`).

willschlitzer: I agree that combinging `use_alias` and `insert_alias` is the smart move, since I don't see how `insert_alias` would be used on its own.

meghanrjones: I like the idea of implementing this inside `use_alias` to simplify the diff, but would prefer to not combine by simply moving the code inside `use_alias`. As an alternate solution, my latest commit calls `insert_alias` from within `use_alias`. There are two reasons why I prefer this option to having all the code within `use_alias`. First, I think that it is simpler (i.e., easier to maintain) if individual functions do smaller, isolated tasks. Second, this implementation is compatible with the changes proposed in #1282. Even if that PR doesn't get merged, it seems to me that having `insert_alias` as a separate function could be more compatible with future updates to the decorators (reason 1).

meghanrjones:
I learned that parameter order matters for tab completion when trying this out. For example, I can tab complete each parameter in this use of basemap:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[-180,180,-80,80],projection="M6i",frame=True)
```
but cannot tab complete region if I select the projection parameter first:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(projection="M6i",region=[-180,180,-80,80],frame=True)
```
It seems like this is the same for other libraries in Jupyter notebooks (e.g., numpy), so I am not sure if there is a way around this but will look into it more. If not, it motivates ordering the aliases in some way that reflects a likely usage order as much as possible.

seisman:
> I learned that parameter order matters for tab completion when trying this out. For example, I can tab complete each parameter in this use of basemap:
> 
> ```
> import pygmt
> fig = pygmt.Figure()
> fig.basemap(region=[-180,180,-80,80],projection="M6i",frame=True)
> ```
> 
> but cannot tab complete region if I select the projection parameter first:
> 
> ```
> import pygmt
> fig = pygmt.Figure()
> fig.basemap(projection="M6i",region=[-180,180,-80,80],frame=True)
> ```
> 
> It seems like this is the same for other libraries in Jupyter notebooks (e.g., numpy), so I am not sure if there is a way around this but will look into it more. If not, it motivates ordering the aliases in some way that reflects a likely usage order as much as possible.

Is it related to the `Parameter.POSITIONAL_OR_KEYWORD` setting?

meghanrjones:
> 
> Is it related to the `Parameter.POSITIONAL_OR_KEYWORD` setting?

Yes,  thanks. I went with the default setting of `POSITIONAL_OR_KEYWORD`, but we could support position independent tab completion by using `KEYWORD_ONLY`.

Actually, it might be worth considering setting other parameters to `KEYWORD_ONLY`. For example, I think all the parameters in `pygmt.Figure.text` are the default of `POSITIONAL_OR_KEYWORD`:
```
def text_(
    self,
    textfiles=None,
    x=None,
    y=None,
    position=None,
    text=None,
    angle=None,
    font=None,
    justify=None,
    **kwargs,
):
```
But `KEYWORD_ONLY` would be more appropriate for most because `textfiles` cannot be provided along with `x` and `y` so we do not really support positional arguments here:
```
def text_(
self,
textfiles=None,
*,
x=None,
y=None,
position=None,
text=None,
angle=None,
font=None,
justify=None,
**kwargs,
):
```
Or similarly only supporting keyword arguments for `data` onward for `plot` since `x`,`y`, and`data` are not supported together:
`def plot(self, x=None, y=None, *, data=None, size=None, direction=None, **kwargs):`
Tab completion does seem to work for plot independent of parameter order, so I guess the only real benefit here is being more explicit about what is expected.

meghanrjones:
One weakness to the current implementation is that all parameters defined by `use_alias` that are not already in the method signature are given a default value of `None`, but I think that required parameters (e.g, `blockmean spacing` and `grdfill mode`) should not have a default value.

weiji14:
> One weakness to the current implementation is that all parameters defined by `use_alias` that are not already in the method signature are given a default value of `None`, but I think that required parameters (e.g, `blockmean spacing` and `grdfill mode`) should not have a default value.

But any parameter with a value of `None` will disappear after being parsed by [`build_arg_str`](https://github.com/GenericMappingTools/pygmt/blob/8e2b2e6fd4900ccae25282d9d50f0ccd34d400ee/pygmt/helpers/utils.py#L104-L158) right? So it should technically be ok to have a default `None` value?

meghanrjones:
> > One weakness to the current implementation is that all parameters defined by `use_alias` that are not already in the method signature are given a default value of `None`, but I think that required parameters (e.g, `blockmean spacing` and `grdfill mode`) should not have a default value.
> 
> But any parameter with a value of `None` will disappear after being parsed by [`build_arg_str`](https://github.com/GenericMappingTools/pygmt/blob/8e2b2e6fd4900ccae25282d9d50f0ccd34d400ee/pygmt/helpers/utils.py#L104-L158) right? See also #639. So it should technically be ok to have a default `None` value?

You are correct. I had thought that required arguments should not have a default `None` value from a readability standpoint, in that parameter=None implies an optional argument. But I cannot find a source for that opinion.

pygmt/helpers/decorators.py
```diff
@@ -322,6 +323,30 @@ def fmt_docstring(module_func):
     return module_func
 
 
+def insert_alias(module_func, default_value=None):
+    """
+    Decorator insertings aliases into the signature of a method.
```
weiji14, line 328: I'm thinking that this function might be best left for internal use only, so maybe put an underscore in the front (see https://stackoverflow.com/a/8689983 for naming convention)? 

```suggestion
def _insert_alias(module_func, default_value=None):
    """
    Function to insert PyGMT long aliases into the signature of a method.
```

pygmt/helpers/__init__.py
```diff
@@ -4,6 +4,7 @@
 from pygmt.helpers.decorators import (
     deprecate_parameter,
     fmt_docstring,
+    insert_alias,
```
weiji14, line 7: Probably take this out of the top level import since it's not used anywhere else.

```suggestion
```

pygmt/helpers/decorators.py
```diff
@@ -383,6 +408,8 @@ def new_module(*args, **kwargs):
 
         new_module.aliases = aliases
 
+        new_module = insert_alias(new_module)
```
weiji14, line 411:
```suggestion
        new_module = _insert_alias(new_module)
```

weiji14 (approved):
Thanks @meghanrjones. I've tested that this works in Jupyter (see below) but not in my Atom IDE as expected (but I can live with that). We can refine this in the future (post PyGMT v0.4.0) to fully resolve #1203.

image

seisman (approved):
Great!

michaelgrund (approved):
That's a great improvement!

weiji14:
Made a small change to the title to clarify that tab completion only works on Jupyter (and IPython too, but that's implied). Feel free to squash and merge this in whenever @meghanrjones.

PR #1284, opened 2021-05-22
Wrap grd2xyz

willschlitzer:
This pull request wraps the [`grd2xyz`](https://docs.generic-mapping-tools.org/6.2/grd2xyz.html) module.

Fixes #

**Reminders**

**Slash Commands**


pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    return load_earth_relief(resolution="01d", region=[-1, 1, -1, 1])
+
+
+def test_grd2xyz(grid):
+    """
+    Make sure grd2xyz works as expected.
+    """
+    xyz_data = grd2xyz(grid=grid)
+    assert xyz_data.strip().split("\n") == [
```
seisman, line 22: Quick comment: I don't think a string-type output is useful. The function should either return  numpy 1d/2d arrays or pandas.DataFrame.

willschlitzer: @seisman could you look at what I just added? As far as I can tell it is working correctly.

weiji14: We might need to open an issue to discuss the output format. Currently, the blockm*/grdtrack modules simply outputs pandas.DataFrame/filenames (see #1099) automatically depending on the input type (2 options). The implementation @willschlitzer made at bf58dc2defc3b6ff67e4769322505d63e4ad941d and 68fab7ceed26083a07ef23681733c33e8dabe827 returns either numpy/pandas/str depending on the user input (3 options. Best to be consistent across PyGMT on the types of outputs users can expect.

weiji14:
@willschlitzer, I'll set this PR to 'draft' mode to conserve CI resources since it might take a while to settle on the output format types (as mentioned in https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r639581006).

Also on a related note, please go easy on clicking the 'Update branch' button since it uses CI resources and also triggers notifications to all maintainers (double notifications if the tests happen to fail). I know you're keen on getting more PRs in (and we do welcome new features!), but there are better ways to merge in a sequence of PRs one after another.

For example with 5 PRs, instead of clicking 'Update branch' 5+4+3+2+1=15 times (updating every PR after one is merged), we can click 'Update branch' 1+1+1+1+1=5 times (updating only the 1 PR that is going to be merged next).

willschlitzer:
As I don't think #1336 will be merged anytime soon, I'm marking this pull request Ready for Review, as figure it'll be easier to incorporate something like `return_table()` once it is completed rather than have multiple PRs waiting on it to be finished.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+    Full option list at :gmt-docs:`grd2xyz.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+        This is the only required parameter.
+    xyz_format : str
```
meghanrjones, line 39:
```suggestion
    output_type : str
```
I would prefer output_type or data_format so that the parameter name can be consistent across all functions that use the future return_table helper

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+            **d**: pandas DataFrame
+            **s**: string
+
+    {R}
+    {V}
+
+    Returns
+    -------
+    info : str
+        A string with information about the grid.
```
meghanrjones, line 51: This needs to be updated because the return is xyz triplets and the default type is not str.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+        This is the only required parameter.
+    xyz_format : str
+        Determine the format the xyz data will be returned in:
+            **a**: numpy array [Default option]
+            **d**: pandas DataFrame
```
meghanrjones, line 42: https://github.com/GenericMappingTools/pygmt/issues/1318#issuecomment-855317785 suggested that a DataFrame should be the default option until a mirroring behavior is implemented.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+            **s**: string
```
meghanrjones, line 43: I agree with https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r637948876 in that I cannot imagine when a str output would be useful.

pygmt/src/grd2xyz.py
willschlitzer: Changed in 0b52451

pygmt/src/grd2xyz.py
willschlitzer: Updated in 37d366d

pygmt/src/grd2xyz.py
willschlitzer: I agree that it's not likely to be used, but I don't think it costs us anything to leave it in, as the output is originally a string before it becomes an array or DataFrame. My vote is to leave it as an option.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+        for i in string_list:
+            float_entry.append(float(i))
+        data_list.append(float_entry)
+    data_array = np.array(data_list)
+    if output_type == "a":
+        result = data_array
+    elif output_type == "d":
+        result = pd.DataFrame(data_array)
+    else:
+        raise GMTInvalidInput("""Must specify format as either a, d, or s.""")
```
meghanrjones, line 77:
```suggestion
    if output_type not in ["a", "d", "s"]:
        raise GMTInvalidInput("""Must specify format as either a, d, or s.""")
    if output_type == "s" and outfile is None:
        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")

    with GMTTempFile() as tmpfile:
        with Session() as lib:
            file_context = lib.virtualfile_from_data(check_kind="raster", data=grid)
            with file_context as infile:
                if outfile is None:
                    outfile = tmpfile.name
                arg_str = " ".join([infile, build_arg_string(kwargs), "->" + outfile])
                lib.call_module("grd2xyz", arg_str)

        # Read temporary csv output to a pandas table
        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
            result = pd.read_csv(tmpfile.name, sep="\t", header=None, comment=">")
        elif outfile != tmpfile.name:  # return None if outfile set, output in outfile
            result = None

        if output_type == "a":
            result = result.to_numpy()
```

I tested this implementation (which is based on the blockmean code), which supports ASCII file output when output_type is "s" and uses pandas.read_csv to get a DataFrame output rather than outfile.read() and string formatting. It's a lot faster for numpy and pandas outputs (~130ms versus ~180ms on average) for running grd2xyz a 240 x 240 grid. I also think it's a bit more intuitive for maintenance. Using this implementation would require updating /pygmt/tests/test_grd2xyz.py::test_grd2xyz and removing the numpy import.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,78 @@
...
+            **a**: numpy array
+            **d**: pandas DataFrame [Default option]
+            **s**: string
+
+    {R}
+    {V}
+
+    Returns
+    -------
+    info : pandas.DataFrame or numpy.array or str
```
meghanrjones, line 50: Why is the name info used here?

willschlitzer: Not sure why I used that originally; changed to `data` in 01613ff

willschlitzer:
Looks like this branch was rebased with `master` at some point; not sure when this happened but that explains all of the extra files changed.

meghanrjones:
> Looks like this branch was rebased with `master` at some point; not sure when this happened but that explains all of the extra files changed.

I think it should be fixed if you merge the 'master' branch into this 'wrap-grd2xyz' branch.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+)
+
+
+@fmt_docstring
+@use_alias(
+    R="region",
+    V="verbose",
+)
+@kwargs_to_strings(R="sequence")
+def grd2xyz(grid, output_type="d", outfile=None, **kwargs):
```
meghanrjones, line 22:
```suggestion
def grd2xyz(grid, output_type="pandas", outfile=None, **kwargs):
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+    r"""
+    Convert grid to data table.
+
+    Read a grid and output xyz-triplets as a numpy array,
+    pandas DataFrame, string, or ASCII [or binary] file.
```
meghanrjones, line 27:
```suggestion
    Read a grid and output xyz-triplets as a :class:`numpy.ndarray`,
    :class:`pandas.DataFrame`, or ASCII file.
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Full option list at :gmt-docs:`grd2xyz.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+        This is the only required parameter.
```
meghanrjones, line 37:
```suggestion
        The file name of the input grid or the grid loaded as a
        :class:`xarray.DataArray`. This is the only required parameter.
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+    output_type : str
+        Determine the format the xyz data will be returned in:
+            **a**: numpy array
+            **d**: pandas DataFrame [Default option]
+            **s**: string
```
meghanrjones, line 42:
```suggestion
        Determine the format the xyz data will be returned in [Default is
        ``pandas``]:

            - ``numpy`` - :class:`numpy.ndarray`
            - ``pandas``-  :class:`pandas.DataFrame`
            - ``file`` - ASCII file (requires ``outfile``)
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+    outfile : str
+        The file name for the output ASCII file.
+    {R}
+    {V}
+
+    Returns
+    -------
+    data : pandas.DataFrame or numpy.array or str
+        The xyz triplet data in a pandas DataFrame, numpy array, or string.
```
meghanrjones, line 51:
```suggestion
    ret : pandas.DataFrame or numpy.ndarray or None
        Return type depends on ``outfile`` and ``output_type``:

        - None if ``outfile`` is set (output will be stored in file set by
          ``outfile``)
        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
          not set (depends on ``output_type`` [Default is
          :class:`pandas.DataFrame`])

```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    if output_type not in ["a", "d", "s"]:
+        raise GMTInvalidInput("""Must specify format as either a, d, or s.""")
+    if output_type == "s" and outfile is None:
```
meghanrjones, line 55:
```suggestion
    if output_type not in ["numpy", "pandas", "file"]:
        raise GMTInvalidInput(
            """Must specify format as either numpy, pandas, or file."""
        )
    if output_type == "file" and outfile is None:
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,75 @@
...
+                arg_str = " ".join([infile, build_arg_string(kwargs), "->" + outfile])
+                lib.call_module("grd2xyz", arg_str)
+
+        # Read temporary csv output to a pandas table
+        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
+            result = pd.read_csv(tmpfile.name, sep="\t", header=None, comment=">")
+        elif outfile != tmpfile.name:  # return None if outfile set, output in outfile
+            result = None
+
+        if output_type == "a":
```
meghanrjones, line 73:
```suggestion
        if output_type == "numpy":
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,53 @@
...
+    Load the grid data from the sample earth_relief file.
+    """
+    return load_earth_relief(resolution="01d", region=[-1, 1, -1, 1])
+
+
+def test_grd2xyz(grid):
+    """
+    Make sure grd2xyz works as expected.
+    """
+    xyz_data = grd2xyz(grid=grid, output_type="a")
```
meghanrjones, line 24:
```suggestion
    xyz_data = grd2xyz(grid=grid, output_type="numpy")
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,53 @@
...
+    assert xyz_data.shape == (4, 3)
+
+
+def test_grd2xyz_format(grid):
+    """
+    Test that correct formats are returned.
+    """
+    xyz_default = grd2xyz(grid=grid)
+    assert isinstance(xyz_default, pd.DataFrame)
+    xyz_array = grd2xyz(grid=grid, output_type="a")
```
meghanrjones, line 34:
```suggestion
    xyz_array = grd2xyz(grid=grid, output_type="numpy")
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,53 @@
...
+    assert isinstance(xyz_array, np.ndarray)
+    xyz_df = grd2xyz(grid=grid, output_type="d")
```
meghanrjones, line 36:
```suggestion
    xyz_df = grd2xyz(grid=grid, output_type="pandas")
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,53 @@
...
+    with pytest.raises(GMTInvalidInput):
+        grd2xyz(grid=grid, output_type=1)
+
+
+def test_grd2xyz_no_outfile(grid):
+    """
+    Test that grd2xyz fails when a string output is set with no outfile.
+    """
+    with pytest.raises(GMTInvalidInput):
+        grd2xyz(grid=grid, output_type="s")
```
meghanrjones, line 53:
```suggestion
        grd2xyz(grid=grid, output_type="file")
```

meghanrjones (commented):
Sorry for the slow review! I am excited about seeing this added to PyGMT!

Currently, providing ``outfile`` and ``output_type="numpy"`` results in ``AttributeError: 'NoneType' object has no attribute 'to_numpy'``, which we can make cleaner. Some options for improving this are:

1. Consolidate ``output_type`` and ``outfile`` so that rather than providing ``output_type="file"`` and the filename to ``outfile``, the user can provide the filename directly to ``output_type``.
2. Raise ``GMTInvalidInput`` if outfile is not None and output_type is not "file"
3. Override ``output_type`` to "file" if ``outfile`` is not None
4. Refactor to support returning a numpy.ndarray or pandas.DataFrame even if ``outfile`` is provided.

My preference would be for option 1 or 4, since I think 2 and 3 are less smooth for the user. Any opinions? Or other suggestions?

willschlitzer:
> Sorry for the slow review! I am excited about seeing this added to PyGMT!
> 
> Currently, providing `outfile` and `output_type="numpy"` results in `AttributeError: 'NoneType' object has no attribute 'to_numpy'`, which we can make cleaner. Some options for improving this are:
> 
>     1. Consolidate `output_type` and `outfile` so that rather than providing `output_type="file"` and the filename to `outfile`, the user can provide the filename directly to `output_type`.
> 
>     2. Raise `GMTInvalidInput` if outfile is not None and output_type is not "file"
> 
>     3. Override `output_type` to "file" if `outfile` is not None
> 
>     4. Refactor to support returning a numpy.ndarray or pandas.DataFrame even if `outfile` is provided.
> 
> 
> My preference would be for option 1 or 4, since I think 2 and 3 are less smooth for the user. Any opinions? Or other suggestions?

@meghanrjones I think my preference is for 3, with the only addition being that a warning will be raised that states that the `output_type` is being changed to "file" because an argument was passed to `output_file` so that it is made explicitly clear to the user. My rationale is that it is simpler to have different parameters accomplish different things (one to set the output type, one to name it) rather than having a single parameter that makes multiple changes (sets output type, creates a filename).

seisman:
> @meghanrjones I think my preference is for 3, with the only addition being that a warning will be raised that states that the `output_type` is being changed to "file" because an argument was passed to `output_file` so that it is made explicitly clear to the user. My rationale is that it is simpler to have different parameters accomplish different things (one to set the output type, one to name it) rather than having a single parameter that makes multiple changes (sets output type, creates a filename).

I agree with @willschlitzer.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,86 @@
...
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a
+        :class:`xarray.DataArray`. This is the only required parameter.
+    output_type : str
+        Determine the format the xyz data will be returned in [Default is
+        ``pandas``]:
+
+            - ``numpy`` - :class:`numpy.ndarray`
+            - ``pandas``-  :class:`pandas.DataFrame`
```
weiji14, line 43:
```suggestion
            - ``pandas``- :class:`pandas.DataFrame`
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,86 @@
...
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
+
+    """
+    if output_type not in ["numpy", "pandas", "file"]:
+        raise GMTInvalidInput(
+            """Must specify format as either numpy, pandas, or file."""
+        )
+    if output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
```
weiji14, line 67: > > @meghanrjones I think my preference is for 3, with the only addition being that a warning will be raised that states that the `output_type` is being changed to "file" because an argument was passed to `output_file` so that it is made explicitly clear to the user. My rationale is that it is simpler to have different parameters accomplish different things (one to set the output type, one to name it) rather than having a single parameter that makes multiple changes (sets output type, creates a filename).
> 
> I agree with @willschlitzer.

If we're going for option 3 and adding the warning, then try something like this (debating whether to use `SyntaxWarning` or `RuntimeWarning`, xref https://docs.python.org/3/library/warnings.html#warning-categories and #1316).

```suggestion
    if outfile is not None and output_type != "file":
        msg = (
            f"Changing `output_type` of grd2xyz from '{output_type}' to 'file' "
            "since `outfile` parameter is set. Please use `output_type='file'` "
            "to silence this warning."
        )
        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
    if output_type == "file" and outfile is None:
        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,86 @@
+"""
+grd2xyz - Convert grid to data table
+"""
+import pandas as pd
```
weiji14, line 4:
```suggestion
import warnings

import pandas as pd
```

pygmt/src/grd2xyz.py
willschlitzer: @weiji14 Why would this be different than just using a `GMTInvalidInput` error? I agree that this falls under the description of a `RuntimeWarning` as described in the Python docs, but it's not clear to me why it isn't better suited for `GMTInvalid Input`.

weiji14: The `GMTInvalidInput` is blocking, whereas the `RuntimeWarning` is non-blocking (i.e. the module will continue to run). Did you want to make this an error or just a warning?

willschlitzer: Makes sense. It looks like it will still output to a file as long as an `outfile` name is set, regardless of the argument passed to `output_type`.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,95 @@
...
+            """Must specify format as either numpy, pandas, or file."""
+        )
+    if outfile is not None and output_type != "file":
+        msg = (
+            f"Changing `output_type` of grd2xyz from '{output_type}' to 'file' "
+            "since `outfile` parameter is set. Please use `output_type='file'` "
+            "to silence this warning."
+        )
+        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
+    if output_type == "file" and outfile is None:
```
weiji14, line 75: Sorry, forgot that we need to set `output_type="file"` after raising the warning... And maybe this should be an `elif` instead?
```suggestion
        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
        output_type = "file"
    elif output_type == "file" and outfile is None:
```

meghanrjones:
@willschlitzer, sounds good regarding the preference for option 3! Feel free to ping when you are ready for a (hopefully final) review.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,95 @@
...
+        Return type depends on ``outfile`` and ``output_type``:
+
+        - None if ``outfile`` is set (output will be stored in file set by
+          ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
+
+    """
+    if output_type not in ["numpy", "pandas", "file"]:
```
seisman, line 64: Because the same code block will be used in many other modules (e.g., `project`, `grdvolume`), should we move it to a new function like `check_output_type`.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,95 @@
...
+                lib.call_module("grd2xyz", arg_str)
+
+        # Read temporary csv output to a pandas table
+        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
+            result = pd.read_csv(tmpfile.name, sep="\t", header=None, comment=">")
+        elif outfile != tmpfile.name:  # return None if outfile set, output in outfile
+            result = None
+
+        if output_type == "numpy":
+            result = result.to_numpy()
```
seisman, line 94: This code block is also reusable in other functions.

pygmt/src/grd2xyz.py
willschlitzer: I would prefer that this pull request is merged before refactoring it to add a function like `check_output_type`.

pygmt/src/grd2xyz.py
willschlitzer: @weiji14 I'm struggling to write a test for this warning. This is what I have so far:

```def test_grd2xyz_outfile_incorrect_output_type(grid):
    """
    Test that grd2xyz raises a warning when an outfile name is set but the
    output_type is not set to file is set with no outfile.
    """
    with pytest.raises(RuntimeWarning):
        with GMTTempFile(suffix=".xyz") as tmpfile:
            result = grd2xyz(grid=grid, outfile=tmpfile.name, output_type="numpy")
```

It returns `Failed: DID NOT RAISE <class 'RuntimeWarning'>`. 

I'm looking through the [pytest warning documentation](https://docs.pytest.org/en/6.2.x/warnings.html) but not finding something that describes this type of test.

pygmt/src/grd2xyz.py
seisman: It's OK to me.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,96 @@
...
+            file_context = lib.virtualfile_from_data(check_kind="raster", data=grid)
+            with file_context as infile:
+                if outfile is None:
+                    outfile = tmpfile.name
+                arg_str = " ".join([infile, build_arg_string(kwargs), "->" + outfile])
+                lib.call_module("grd2xyz", arg_str)
+
+        # Read temporary csv output to a pandas table
+        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
+            result = pd.read_csv(tmpfile.name, sep="\t", header=None, comment=">")
```
seisman, line 90: For the pandas.DataFrame output, now the output looks like:

image

I think it's better to name these columns, e.g., `x`, `y`, `z`, rather than 0, 1, 2.

weiji14: Good point. The input xarray.DataArray may have names for the coordinates, maybe we should pick it up from there? But fallback to xyz if not possible.

seisman: > Good point. The input xarray.DataArray may have names for the coordinates, maybe we should pick it up from there? But fallback to xyz if not possible.

Yes, it sounds better.

willschlitzer: Would it make sense to include a parameter that allows the user to set headers for the DataFrame output, but would default to the DataArray names, or, as a last resort, xyz?

I'm thinking something like `dataframe_header=None`, followed by:
```
if dataframe_header is None:
   if type(grid) == xr.DataArray:
        dataframe_header = grid.dims
    else:
         dataframe_header = ['x', 'y', 'z']
```

weiji14: It's not too difficult the rename the columns of the output using [`pandas.DataFrame.rename`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.rename.html), so I would just keep the `if type(grid) == xr.DataArray: ... else: ...` part. No need for extra complexity.

willschlitzer:
@GenericMappingTools/pygmt-contributors Could I get some guidance on writing the test for when a `RuntimeWarning` is raised?

seisman:
> @weiji14 I'm struggling to write a test for this warning. This is what I have so far:
> 
> ```
>     """
>     Test that grd2xyz raises a warning when an outfile name is set but the
>     output_type is not set to file is set with no outfile.
>     """
>     with pytest.raises(RuntimeWarning):
>         with GMTTempFile(suffix=".xyz") as tmpfile:
>             result = grd2xyz(grid=grid, outfile=tmpfile.name, output_type="numpy")
> ```
> 
> It returns `Failed: DID NOT RAISE <class 'RuntimeWarning'>`.
> 
> I'm looking through the [pytest warning documentation](https://docs.pytest.org/en/6.2.x/warnings.html) but not finding something that describes this type of test. As far as I can tell, `RuntimeWarning` is built in to Python, so it shouldn't need to be a separate import statement?

I think you should use `with pytest.warns(RuntimeWarning):`. See https://docs.pytest.org/en/6.2.x/warnings.html#warns for details.

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_grd2xyz_outfile_incorrect_output_type(grid):
+    """
+    Test that grd2xyz raises a warning when an outfile name is set but the
+    output_type is not set to file is set with no outfile.
+    """
+    with pytest.warns(RuntimeWarning):
+        with GMTTempFile(suffix=".xyz") as tmpfile:
+            grd2xyz(grid=grid, outfile=tmpfile.name, output_type="numpy")
```
weiji14, line 79: Still need to check that the resulting output is indeed `None`.

```suggestion
    with pytest.warns(RuntimeWarning):
        with GMTTempFile(suffix=".xyz") as tmpfile:
            result = grd2xyz(grid=grid, outfile=tmpfile.name, output_type="numpy")
            assert result is None  # return value is None
            assert os.path.exists(path=tmpfile.name)  # check that outfile exists
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,105 @@
...
+            "to silence this warning."
+        )
+        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
+    # Set the default column names for the pandas dataframe header
+    dataframe_header = ["x", "y", "z"]
+    # Set the column names to match an input DataArray as the grid
+    if isinstance(grid, xr.DataArray) == xr.DataArray and output_type == "pandas":
```
weiji14, line 82:
```suggestion
    if isinstance(grid, xr.DataArray) and output_type == "pandas":
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_grd2xyz_format(grid):
+    """
+    Test that correct formats are returned.
+    """
+    xyz_default = grd2xyz(grid=grid)
+    assert isinstance(xyz_default, pd.DataFrame)
+    print(type(list(xyz_default.columns)))
+    assert list(xyz_default.columns) == ["x", "y", "z"]
```
weiji14, line 38: The earth_relief grid dimensions are `lat` and `lon`, with the 'z' variable being called `elevation`. Oh, and the print statement should be left out.

```suggestion
    assert list(xyz_default.columns) == ["lon", "lat", "elevation"]
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,105 @@
...
+        if len(grid.dims) == 3:
+            dataframe_header = grid.dims
```
weiji14, line 84: I think you meant `if grid.ndim == 2` here instead of 3. The `grid.dims` would just produce `('lat', 'lon')`. That said, do we need to check that the grid is 2-dimensional? Or can we just set the header like so?
```suggestion
        dataframe_header = grid.dims + (grid.name,)
```

weiji14 (commented):
Just following on from the discussion about automatic column names at https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r699053388, I think we might need some extra tests to make sure the x, y, z dimensions in the `xarray.DataArray` map correctly to x, y, z in the `pandas.DataFrame`.

Specifically, we want to make sure yxz becomes yxz, and not xyz. The `earth_relief` grid has 'lat', 'lon', 'elevation' as dimension names (i.e. yxz), so will need to make sure `grd2xyz` handles the conversion correctly. Hopefully I'm getting the idea across..

willschlitzer:
> Just following on from the discussion about automatic column names at [#1284 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r699053388), I think we might need some extra tests to make sure the x, y, z dimensions in the `xarray.DataArray` map correctly to x, y, z in the `pandas.DataFrame`.
> 
> Specifically, we want to make sure yxz becomes yxz, and not xyz. The `earth_relief` grid has 'lat', 'lon', 'elevation' as dimension names (i.e. yxz), so will need to make sure `grd2xyz` handles the conversion correctly. Hopefully I'm getting the idea across..

@weiji14 How would I go about testing this? I'm not sure how I could write a test that would automatically determine if a value in gridded data is on the x-, y-, or z-axis.

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,80 @@
...
+from pygmt.exceptions import GMTInvalidInput
+from pygmt.helpers import GMTTempFile
+
+
+@pytest.fixture(scope="module", name="grid")
+def fixture_grid():
+    """
+    Load the grid data from the sample earth_relief file.
+    """
+    return load_earth_relief(resolution="01d", region=[-1, 1, -1, 1])
```
meghanrjones, line 20:
```suggestion
    return load_earth_relief(resolution="01d", region=[-1, 1, 3, 5])
```
Recommend not having the lon and lat range identical for testing the columns are converted correctly

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,80 @@
...
+    assert xyz_data.shape == (4, 3)
+
+
+def test_grd2xyz_format(grid):
+    """
+    Test that correct formats are returned.
+    """
+    xyz_default = grd2xyz(grid=grid)
+    assert isinstance(xyz_default, pd.DataFrame)
+    assert list(xyz_default.columns) == ["lon", "lat", "elevation"]
```
meghanrjones, line 37:
```suggestion
    lon = -0.5
    lat = 3.5
    orig_val = grid.sel(lon=lon, lat=lat).to_numpy()
    xyz_default = grd2xyz(grid=grid)
    xyz_val = xyz_default[(xyz_default["lon"] == lon) & (xyz_default["lat"] == lat)][
        "elevation"
    ].to_numpy()
    assert isinstance(xyz_default, pd.DataFrame)
    assert orig_val.size == 1
    assert xyz_val.size == 1
    np.testing.assert_allclose(orig_val, xyz_val)
```
Here is an example that would work for testing Wei Ji's suggestion. The test currently fails because 'lon' and 'lat' are switched for the output DataFrame.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
+    # Set the default column names for the pandas dataframe header
+    dataframe_header = ["x", "y", "z"]
+    # Set the column names to match an input DataArray as the grid
+    if isinstance(grid, xr.DataArray) and output_type == "pandas":
+        dataframe_header = grid.dims + (grid.name,)
```
meghanrjones, line 83:
```suggestion
    if "o" not in kwargs.keys(): # Only set column names for standard output
        # Set the default column names for the pandas dataframe header
        dataframe_header = ["x", "y", "z"]
        # Set the column names to match an input DataArray as the grid
        if isinstance(grid, xr.DataArray) and output_type == "pandas":
            # Reverse the dims because it is rows, columns ordered.
            dataframe_header = (grid.dims[1],) + (grid.dims[0],) + (grid.name,)
```
Maybe support for auto-determination of columns names when `outcols` is used could be included as part of https://github.com/GenericMappingTools/pygmt/issues/1318, but for now we could not name the columns if **-o** (`outcols`) is set since we do not have a parser for string arguments and the common option is not added to this function.

Based on this code in `pygmt/clib/converstion.py`, I expect that switching the order of the dims should work for the DataArray input case.
```
    # Extract region and inc from the grid
    region = []
    inc = []
    # Reverse the dims because it is rows, columns ordered. In geographic
    # grids, this would be North-South, East-West. GMT's region and inc are
    # East-West, North-South.
    for dim in grid.dims[::-1]:
        coord = grid.coords[dim].values
        coord_incs = coord[1:] - coord[0:-1]
        coord_inc = coord_incs[0]
        if not np.allclose(coord_incs, coord_inc):
            raise GMTInvalidInput(
                f"Grid appears to have irregular spacing in the '{dim}' dimension."
            )
        if coord_inc == 0:
            raise GMTInvalidInput(
                f"Grid has a zero increment in the '{dim}' dimension."
            )
        region.extend(
            [
                coord.min() - coord_inc / 2 * grid.gmt.registration,
                coord.max() + coord_inc / 2 * grid.gmt.registration,
            ]
        )
```

meghanrjones:
The tests need to be updated for the new automatic column names.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,106 @@
...
+    output_type : str
+        Determine the format the xyz data will be returned in [Default is
+        ``pandas``]:
+
+            - ``numpy`` - :class:`numpy.ndarray`
+            - ``pandas``- :class:`pandas.DataFrame`
+            - ``file`` - ASCII file (requires ``outfile``)
+    outfile : str
+        The file name for the output ASCII file.
+    {R}
```
seisman, line 50: The grd2xyz documentation explains the `-R` option in this way:

> Specify the region of interest. Using the -R option will select a subsection of the grid. If this subsection exceeds the boundaries of the grid, only the common region will be output.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,106 @@
...
+        - None if ``outfile`` is set (output will be stored in file set by
+          ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
+
+    """
+    if output_type not in ["numpy", "pandas", "file"]:
+        raise GMTInvalidInput(
+            """Must specify format as either numpy, pandas, or file."""
```
seisman, line 67:
```suggestion
            "Must specify format as either numpy, pandas, or file."
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,106 @@
...
+        output_type = "file"
+    elif output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
+    if "o" not in kwargs.keys(): # Only set column names for standard output
+        # Set the default column names for the pandas dataframe header
+        dataframe_header = ["x", "y", "z"]
+        # Set the column names to match an input DataArray as the grid
+        if isinstance(grid, xr.DataArray) and output_type == "pandas":
+            # Reverse the dims because it is rows, columns ordered.
+            dataframe_header = (grid.dims[1],) + (grid.dims[0],) + (grid.name,)
```
seisman, line 85: Does the following code works in the same way?
```suggestion
            dataframe_header = [grid.dims[1], grid.dims[0], grid.name]
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,106 @@
...
+                if outfile is None:
+                    outfile = tmpfile.name
+                arg_str = " ".join([infile, build_arg_string(kwargs), "->" + outfile])
+                lib.call_module("grd2xyz", arg_str)
+
+        # Read temporary csv output to a pandas table
+        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
+            result = pd.read_csv(
+                tmpfile.name, sep="\t", names=dataframe_header, comment=">"
+            )
```
seisman, line 100: See line 79. If `-o` is used, `dataframe_header` is undefined, right?

pygmt/src/grd2xyz.py
willschlitzer: Updated docstring

pygmt/src/grd2xyz.py
willschlitzer: I'm not really sure what this is asking me to do. I added an if statement that doesn't pass an argument to `names` if `o` is in `kwargs`.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,110 @@
...
+    if outfile is not None and output_type != "file":
+        msg = (
+            f"Changing `output_type` of grd2xyz from '{output_type}' to 'file' "
+            "since `outfile` parameter is set. Please use `output_type='file'` "
+            "to silence this warning."
+        )
+        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
```
weiji14, line 79:
```suggestion
    elif outfile is None and output_type == "file":
        raise GMTInvalidInput("Must specify `outfile` for ASCII output.")

```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,110 @@
...
+    if "o" not in kwargs.keys():  # Only set column names for standard output
```
weiji14, line 80: No need for `.keys()` here, pylint might complain :smile: Also, the sentence "Only set column names for standard output" sounds confusing, because standard output sounds like stdout in the command line but that doesn't translate well into the Python world. Suggest rewording it.

```suggestion
    if "o" not in kwargs:  # Only set column names for standard output
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,110 @@
...
+        # Set the default column names for the pandas dataframe header
+        dataframe_header = ["x", "y", "z"]
+        # Set the column names to match an input DataArray as the grid
```
weiji14, line 83:
```suggestion
        # Let output pandas column names match input DataArray dimension names
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,110 @@
...
+        if isinstance(grid, xr.DataArray) and output_type == "pandas":
+            # Reverse the dims because it is rows, columns ordered.
+            dataframe_header = [grid.dims[1], grid.dims[0], grid.name]
```
weiji14, line 86: I'm debating whether we should just parse the dimensions names out using `grdinfo`. Maybe in a separate PR since this is getting long, but it would be nice if passing an input NetCDF file through `pygmt.grd2xyz` can output proper pandas column names based on the NetCDF dimensions!

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,110 @@
...
+                lib.call_module("grd2xyz", arg_str)
+
+        # Read temporary csv output to a pandas table
+        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
+            if "o" not in kwargs.keys():
+                result = pd.read_csv(
+                    tmpfile.name, sep="\t", names=dataframe_header, comment=">"
+                )
+            else:
+                result = pd.read_csv(tmpfile.name, sep="\t", comment=">")
```
weiji14, line 104: Need a test with `o` option to ensure this line has code coverage.

willschlitzer: @weiji14 @seisman I'm struggling to come up with a good test for this. Should it just assert that some of the columns will be `NaN` if an argument for `o` is provided?

seisman: When `-o` is used, only some of the `x`, `y`, `z` columns will be written to the temporary file. For this case, it's difficult to guess which columns are written unless PyGMT can parse the `-o` option. 

I'm thinking if it's simpler to disable the `-o` option for pandas output.

willschlitzer: > I'm thinking if it's simpler to disable the `-o` option for pandas output.

I like this. I don't know why someone wouldn't want all of the information in the output dataframe; the user can use `pandas` functions to select some of the columns.

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,115 @@
...
+            "to silence this warning."
+        )
+        warnings.warn(message=msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif outfile is None and output_type == "file":
+        raise GMTInvalidInput("Must specify `outfile` for ASCII output.")
+
+    if "o" in kwargs and output_type == "pandas":
+        raise GMTInvalidInput(
+            "If 'o' is specified, `output_type` must be either as 'numpy' or 'file'."
```
weiji14, line 86: Prefer to use long alias 'outcols' rather than just 'o'.
```suggestion
            "If 'outcols' is specified, `output_type` must be either 'numpy'"
            "or 'file'."
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,115 @@
...
+            - ``numpy`` - :class:`numpy.ndarray`
+            - ``pandas``- :class:`pandas.DataFrame`
+            - ``file`` - ASCII file (requires ``outfile``)
+    outfile : str
+        The file name for the output ASCII file.
+    {R}
+        Adding `region` will select a subsection of the grid. If this
+        subsection exceeds the boundaries of the grid, only the common region
+        will be output.
+    {V}
```
weiji14, line 54:
```suggestion
    {V}
    {o}
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,115 @@
...
+        Determine the format the xyz data will be returned in [Default is
+        ``pandas``]:
+
+            - ``numpy`` - :class:`numpy.ndarray`
+            - ``pandas``- :class:`pandas.DataFrame`
+            - ``file`` - ASCII file (requires ``outfile``)
+    outfile : str
+        The file name for the output ASCII file.
+    {R}
+        Adding `region` will select a subsection of the grid. If this
```
seisman, line 51:
```suggestion
        Adding ``region`` will select a subsection of the grid. If this
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,115 @@
...
+        - None if ``outfile`` is set (output will be stored in file set by
+          ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
+
+    """
+    if output_type not in ["numpy", "pandas", "file"]:
+        raise GMTInvalidInput(
+            "Must specify `output_type` either as 'numpy', 'pandas' or 'file'."
```
seisman, line 70: I think we should always use single quotes for parameters and arguments.
```suggestion
            "Must specify 'output_type' either as 'numpy', 'pandas' or 'file'."
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,118 @@
...
+    if output_type not in ["numpy", "pandas", "file"]:
+        raise GMTInvalidInput(
+            "Must specify 'output_type' either as 'numpy', 'pandas' or 'file'."
+        )
+
+    if outfile is not None and output_type != "file":
+        msg = (
+            f"Changing `output_type` of grd2xyz from '{output_type}' to 'file' "
+            "since `outfile` parameter is set. Please use `output_type='file'` "
+            "to silence this warning."
```
seisman, line 79:
```suggestion
            f"Changing 'output_type' of grd2xyz from '{output_type}' to 'file' "
            "since 'outfile' parameter is set. Please use output_type='file' "
            "to silence this warning."
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        warnings.warn(message=msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif outfile is None and output_type == "file":
+        raise GMTInvalidInput("Must specify `outfile` for ASCII output.")
```
seisman, line 84:
```suggestion
        raise GMTInvalidInput("Must specify 'outfile' for ASCII output.")
```

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if "o" in kwargs and output_type == "pandas":
+        raise GMTInvalidInput(
+            "If 'outcols' is specified, `output_type` must be either 'numpy'"
```
seisman, line 88:
```suggestion
            "If 'outcols' is specified, 'output_type' must be either 'numpy'"
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,96 @@
...
+            assert result is None  # return value is None
+            assert os.path.exists(path=tmpfile.name)  # check that outfile exists
+
+
+def test_grd2xyz_pandas_output_with_o(grid):
+    """
+    Test that grd2xyz fails when 'o' is set and output_type is set to 'pandas'.
+    """
+    with pytest.raises(GMTInvalidInput):
+        grd2xyz(grid=grid, output_type="pandas", o="2")
```
seisman, line 96:
```suggestion
        grd2xyz(grid=grid, output_type="pandas", outcols="2")
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,96 @@
...
+    with pytest.warns(RuntimeWarning):
+        with GMTTempFile(suffix=".xyz") as tmpfile:
+            result = grd2xyz(grid=grid, outfile=tmpfile.name, output_type="numpy")
+            assert result is None  # return value is None
+            assert os.path.exists(path=tmpfile.name)  # check that outfile exists
+
+
+def test_grd2xyz_pandas_output_with_o(grid):
+    """
+    Test that grd2xyz fails when 'o' is set and output_type is set to 'pandas'.
```
seisman, line 93:
```suggestion
    Test that grd2xyz fails when outcols is set and output_type is set to 'pandas'.
```

pygmt/src/grd2xyz.py
willschlitzer: I think it would be best for parsing dimension names to come out of a separate PR, but I think it's a good feature to add to PyGMT!

pygmt/src/grd2xyz.py
```diff
@@ -0,0 +1,118 @@
...
+    Returns
+    -------
+    ret : pandas.DataFrame or numpy.ndarray or None
+        Return type depends on ``outfile`` and ``output_type``:
+
+        - None if ``outfile`` is set (output will be stored in file set by
+          ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
```
seisman, line 67: We already documented the default value of `output_type` above. I feel it's unnecessary to repeat the default value here.
```suggestion
          not set (depends on ``output_type``)
```

pygmt/tests/test_grd2xyz.py
```diff
@@ -0,0 +1,96 @@
...
+    Test that grd2xyz fails when a string output is set with no outfile.
+    """
+    with pytest.raises(GMTInvalidInput):
+        grd2xyz(grid=grid, output_type="file")
+
+
+def test_grd2xyz_outfile_incorrect_output_type(grid):
+    """
+    Test that grd2xyz raises a warning when an outfile name is set but the
+    output_type is not set to file is set with no outfile.
```
seisman, line 82: I can't understand this sentence. Can you please rephrase it?

seisman (approved):

pygmt/tests/test_grd2xyz.py
willschlitzer: Looks like my laptop went a little crazy when I was writing that and copy/pasted text from something else!

willschlitzer: fixed in f49dac0

seisman:
I think this PR is already good for merging, but it would be better if @meghanrjones can also give it a final review, since @meghanrjones also made a lot of suggestions in this PR.

meghanrjones (approved):
Great work, Will!

PR #1287, opened 2021-05-23
Update the inset gallery example

michaelgrund:

Increase the margin pen size of the inset box since after updating to GMT 6.2.0rc1 it appears to be too thin. 

**Reminders**

**Slash Commands**


core-man (approved):
Looks great~

meghanrjones:
Do you think that the inset box pen thickness is universally to thin with 6.2.0rc1, or just in this case since the pen color is not very bold? If it's a universal problem, we could try to fix the auto-scaling on the GMT side before 6.2.0.

willschlitzer:
> Do you think that the inset box pen thickness is universally too thin with 6.2.0rc1, or just in this case since the pen color is not very bold? If it's a universal problem, we could try to fix the auto-scaling on the GMT side before 6.2.0.

I think the inset box thickness is fine for the gallery example, and just looks thin since it is a light border. Looking at the tutorial examples, the border does look too thin when I compare the version in this deployment to the PyGMT v0.3.1 version.

willschlitzer (approved):
Can't say I notice that much of a difference between `red` and `red3`, but I think thickening the inset border is a smart move.

michaelgrund:
> 
> 
> Do you think that the inset box pen thickness is universally too thin with 6.2.0rc1, or just in this case since the pen color is not very bold? If it's a universal problem, we could try to fix the auto-scaling on the GMT side before 6.2.0.

So far, I only noticed it in this example.

michaelgrund:
> 
> 
> Can't say I notice that much of a difference between `red` and `red3`, but I think thickening the inset border is a smart move.

It's not a big difference, only looks a little bit more modern in my opinion 😄 , however, we could also leave it as is.

meghanrjones:
FYI, the PR at https://github.com/GenericMappingTools/gmt/pull/5265 will revert the default behavior for pen with +p to the 6.1 standard. Of course, this doesn't prevent setting it manually anyways as done here.

examples/gallery/embellishments/inset.py
```diff
@@ -14,12 +14,17 @@
...
+# with a pen size of 1.5p.
+with fig.inset(position="jTL+w3.5c+o0.2c", margin=0, box="+p1.5p,gold"):
     # Create a figure in the inset using coast. This example uses the azimuthal
     # orthogonal projection centered at 47E, 20S. The land color is set to "gray" and
-    # Madagascar is highlighted in "red".
+    # Madagascar is highlighted in "red3".
     fig.coast(
-        region="g", projection="G47/-20/3.5c", land="gray", water="white", dcw="MG+gred"
+        region="g",
+        projection="G47/-20/3.5c",
```
seisman, line 25: We should use `?` to let GMT automatically determine the map width in the inset.
```suggestion
        projection="G47/-20/?",
```

seisman (approved):

PR #1295, opened 2021-05-25
Add projection and region to grdview docstring

meghanrjones:

The `projection` parameter is included in the list of aliases but not the docstring for grdview. This PR adds the projection description to the docstring.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdview.py
+++ b/pygmt/src/grdview.py
@@ -58,6 +58,7 @@ def grdview(self, grid, **kwargs):
     grid : str or xarray.DataArray
         The file name of the input relief grid or the grid loaded as a
         DataArray.
+    {R}
     {J}
     zscale/zsize : float or str
         Set z-axis scaling or z-axis size.
--- a/pygmt/src/grdview.py
+++ b/pygmt/src/grdview.py
@@ -58,7 +58,12 @@ def grdview(self, grid, **kwargs):
     grid : str or xarray.DataArray
         The file name of the input relief grid or the grid loaded as a
         DataArray.
-    {R}
+    region : str or list
+        *xmin/xmax/ymin/ymax*\ [**+r**][**+u**\ *unit*].
+        Specify the :doc:`region </tutorials/regions>` of interest. When used
+        with ``perspective``, optionally append */zmin/zmax* to indicate the
+        range to use for the 3-D axes [Default is the region in the input
+        grid].
     {J}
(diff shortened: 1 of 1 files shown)


weiji14 (approved):

pygmt/src/grdview.py
```diff
@@ -58,6 +58,7 @@ def grdview(self, grid, **kwargs):
     grid : str or xarray.DataArray
         The file name of the input relief grid or the grid loaded as a
         DataArray.
+    {J}
```
weiji14, line 61:
```suggestion
    {R}
    {J}
```

weiji14 (changes requested):
Oh wait, actually, we're missing the region (R) docstring as well for this!

pygmt/src/grdview.py
meghanrjones: Good catch, thanks

weiji14 (approved):
So region (-R) isn't actually a required argument for `grdview` according to https://docs.generic-mapping-tools.org/6.2.0rc2/grdview.html, but using `{R}` in the docstring here says that it is:

image

Wondering if we should edit the wording slightly at https://github.com/GenericMappingTools/pygmt/blob/01e2febb77f08370cb8ef8e23f7b1c42e9709c99/pygmt/helpers/decorators.py#L17-L21

Can open a separate issue/PR for this.

meghanrjones:
> 
> Wondering if we should edit the wording slightly
> 


I think the current wording is accurate for most plotting modules, so maybe I can just not use the replacement text for `grdview` but keep the wording from decorators.py as is to not affect other method docs?

meghanrjones:
Is the latest version OK with you, @weiji14?

weiji14:
> Is the latest version OK with you, @weiji14?

Yep, looks good! I like that it explicitly specifies zmin/zmax too. Remember to change the PR title to include 'region' before merging.

meghanrjones:
Sounds good. One more question - when the checks fail for reasons clearly unrelated to the PR (like now), do you prefer for me to re-run all checks or admin merge the PR?

weiji14:
> when the checks fail for reasons clearly unrelated to the PR (like now), do you prefer for me to re-run all checks or admin merge the PR?

I would just admin merge in this case to conserve CI resources.

Issue #764, opened 2020-12-25
Description for the "columns" arguments is missing


michaelgrund:
Please describe what could be improved about this page or the typo/mistake that you found:

I would suggest to add a short note regarding the supported syntax for the `columns` option when using e.g. a data file as input, something like 

* **data** (_str_ or _2d array_) – Either a data file name or a 2d numpy array with the tabular data. Use option _columns_ (i) to choose which columns are x, y, color, and size, respectively (e.g. `columns = [0, 1]` if the x values are stored in the first column and y values in the second one, note: zero-based indexing is used).

I tried to fix that suggestion by my own following the instructions given in the **Contributing Guidelines** under _Editing the Documentation_, however, the corresponding `.rst` file is not available in the `doc` folder.

seisman:
> the corresponding .rst file is not available in the doc folder.

The documentation is automatically generated from the docstrings in the Python codes, see
https://github.com/GenericMappingTools/pygmt/blob/master/pygmt/base_plotting.py#L649

> I would suggest to add a short note regarding the supported syntax for the columns option when using e.g. a data file as input, something like

> data (str or 2d array) – Either a data file name or a 2d numpy array with the tabular data. Use option columns (i) to choose which columns are x, y, color, and size, respectively (e.g. columns = [0, 1] if the x values are stored in the first column and y values in the second one, note: zero-based indexing is used).

The description of the `columns` argument is missing. I think we should add an entry for the "columns" argument, instead of explaining `columns` in the description of the `data` argument.

michaelgrund:
opened a pull request with the suggested changes (according to your comment @seisman ).

weiji14:
Reopening this because the "columns" description is missing for `plot3d` at https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot3d.html.

I'm marking this as a good first issue as it should be easy enough to just copy and paste the description from https://github.com/GenericMappingTools/pygmt/blob/v0.3.0/pygmt/src/plot.py to https://github.com/GenericMappingTools/pygmt/blob/v0.3.0/pygmt/src/plot3d.py.

seisman:
I'm not sure if we discussed it before (github search shows nothing). Is "columns" a good alias for **-i** option?

GMT provides both **-i** and **-o**, for selecting columns of input and output, respectively (see https://docs.generic-mapping-tools.org/latest/std-opts.html).

[...]

PR #1040, opened 2021-03-12
Figure.plot3d: Deprecate parameter "columns" to "incols" (remove in v0.6.0)

willschlitzer:
As discussed by @weiji14, this is to add the parameter description for `incols` in `plot3d.py`.

Fixes #764 

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1298, opened 2021-05-26
Figure.plot: Deprecate parameter "columns" to "incols" (remove in v0.6.0)

michaelgrund:

As discussed in #764, this PR replaces the **columns** parameter by **incols** for  Figure.plot().

- [x] Rename the parameter `columns` to `incols` in Figure.plot()
- [x] Updates tests and examples to use the new parameter name
- [x] Use the deprecate_parameter decorator for backward-compatibility
- [x] Add a test for checking 'columns' backward compatibility

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -193,12 +193,7 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
     {a}
     {c}
     {f}
-    incols : str or 1d array
-        Choose which columns are x, y, color, and size, respectively if
-        input is provided via *data*. E.g. ``incols = [0, 1]`` or
-        ``incols = '0,1'`` if the *x* values are stored in the first
-        column and *y* values in the second one. Note: zero-based
-        indexing is used.
+    {i}
     label : str
         Add a legend entry for the symbol or line being plotted.
 
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -36,7 +36,7 @@ from pygmt.helpers import (
     Y="yshift",
     Z="zvalue",
     a="aspatial",
-    i="columns",
+    i="incols",
     l="label",
     c="panel",
     f="coltypes",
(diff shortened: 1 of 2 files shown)


meghanrjones:
The alias name and description for `incols` should go in `COMMON_OPTIONS` in decorators.py. I will submit a separate PR for adding the `incols` common option to decorators.py so that it does not get complicated coordinating the timing of this PR and #1040.

michaelgrund:
> 
> 
> The alias name and description for `incols` should go in `COMMON_OPTIONS` in decorators.py. I will submit a separate PR for adding the `incols` common option to decorators.py so that it does not get complicated coordinating the timing of this PR and #1040.

Sounds great!

michaelgrund:
Any idea why the newly added test fails here @seisman ?

seisman:
> Any idea why the newly added test fails here @seisman ?

I think I misunderstood the order Python executes decorators. I thought the execution order is bottom-to-top, but the actual execution order is top-to-bottom (see https://gist.github.com/kirsle/bf24622fc5f255256c6e for detailed explanations).

If the top-to-bottom execution order is correct, the `deprecate_parameter` decorator should always be put before the `use_alias` decorator.

pygmt/src/plot.py
```diff
@@ -45,6 +45,7 @@
 )
 @kwargs_to_strings(R="sequence", c="sequence_comma", i="sequence_comma", p="sequence")
 @deprecate_parameter("sizes", "size", "v0.4.0", remove_version="v0.6.0")
+@deprecate_parameter("columns", "incols", "v0.4.0", remove_version="v0.6.0")
```
weiji14, line 48: The problem is that `test_plot_deprecate_columns_to_incols` uses a list instead of str for the `columns` parameter. The order should be:

1. Rename 'columns' to 'incols' via `@deprecate_parameter`
2. Convert long alias 'incols' to GMT short alias 'i' via `@use_alias`
3. Let the list type argument `[0, 1]` be converted to str type `0,1` via `@kwargs_to_strings`
4. Pass in `-i0,1` to `lib.call_module("plot", ...)`

weiji14 (commented):
Python decorators are applied top-down according to https://realpython.com/primer-on-python-decorators/#nesting-decorators. The implementation of `@deprecate_parameter` is correct-ish, but we need to move the `@deprecate_parameter` decorator to be after `@fmt_docstring` and before `@use_alias`.

pygmt/src/plot.py
michaelgrund: Thanks for your replies and the clarification @seisman and @weiji14.

michaelgrund:
> 
> 
> The alias name and description for `incols` should go in `COMMON_OPTIONS` in decorators.py. I will submit a separate PR for adding the `incols` common option to decorators.py so that it does not get complicated coordinating the timing of this PR and #1040.

Now the description for `incols` from `COMMON_OPTIONS`  is used.

seisman (approved):

weiji14 (approved):

PR #1302, opened 2021-05-26
Clarify position of deprecate_parameter decorator to be above use_alias

weiji14:

Need to move the `@deprecate_parameter` decorator to be before `@use_alias` but after `@fmt_docstring` so that the `@kwargs_to_strings` decorator can do its job properly.

In code, it should look something like this.

```python
@fmt_docstring
@deprecate_parameter("columns", "incols", "v0.4.0", remove_version="v0.6.0")
@use_alias(J="projection", R="region", V="verbose", i="incols")
@kwargs_to_strings(R="sequence", i='sequence_comma')
def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
    pass
```

Explanation:

1. Rename 'columns' to 'incols' via `@deprecate_parameter`
2. Convert long alias 'incols' to GMT short alias 'i' via `@use_alias`
3. Let the list type argument `[0, 1]` be converted to str type `0,1` via `@kwargs_to_strings`
4. Pass in `-i0,1` to `lib.call_module("plot", ...)`

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/1298#discussion_r640162801_

Addresses https://github.com/GenericMappingTools/pygmt/pull/1298#discussion_r640162801

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/maintenance.md
+++ b/doc/maintenance.md
@@ -173,21 +173,21 @@ When making incompatible changes, we should follow the process:
   3-12 months.
 - Remove the old usage and warning when reaching the declared version.
 
-To rename a function parameter, add the `@deprecated_parameter` decorator
-before the function definition (but after the `@use_alias` decorator if it exists).
-Here is an example:
+To rename a function parameter, add the `@deprecate_parameter` decorator near
+the top after the `@fmt_docstring` decorator but before the `@use_alias`
+decorator (if those two exists). Here is an example:
 
 ```
 @fmt_docstring
-@use_alias(J="projection", R="region", V="verbose")
-@kwargs_to_strings(R="sequence")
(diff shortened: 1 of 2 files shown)


seisman:
The `@deprecate_parameter` decorator is already used in plot and plot3d. Do you want to fix the order of `deprecate_parameter` in these two functions?

weiji14:
> The `@deprecate_parameter` decorator is already used in plot and plot3d. Do you want to fix the order of `deprecate_parameter` in these two functions?

I'll let @michaelgrund fix the position of the decorator for `plot` in #1298, and @willschlitzer can fix it for `plot3d` in #1040.

seisman (approved):

michaelgrund (approved):

PR #1303, opened 2021-05-27
Figure.contour: Deprecate parameter "columns" to "incols" (remove in v0.6.0)

### LINKED ISSUE #1313: The "incols" (-i) parameter doesn't works for vector input correctly

Using the ``incols`` parameter does not allow to swap columns (seems like the parameter is ignored at all so far). I believe it's an upstream GMT bug. This was discovered in https://github.com/GenericMappingTools/pygmt/pull/1303#discussion_r642213335.

```python
import pygmt
import numpy as np

# generate sample data
region = [10, 70, -5, 10]
x, y = np.meshgrid(
np.linspace(region[0], region[1]), np.linspace(region[2], region[3])
    )
x = x.flatten()
y = y.flatten()
z = (x - 0.5 * (region[0] + region[1])) ** 2 + 4 * y ** 2
z = np.exp(-z / 10 ** 2 * np.log(2))

# generate dataframe
data = np.array([x, y, z]).T

# correct figure
fig1 = pygmt.Figure()
fig1.contour(data=data,
            projection="X10c",
            frame="a",
            pen=True)

fig1.show()

# generate second dataframe, switch x and y from here onwards to simulate 
# different column order
data = np.array([y, x, z]).T

# wrong figure
fig2 = pygmt.Figure()
fig2.contour(data=data,
            incols=[1, 0, 2], # use incols to assign column order
            projection="X10c",
            frame="a",
            pen=True)

fig2.show()
```

**Full output**

| fig1: expected | fig2: actual (wrong) |
|--|--|
| correct | incorrect |

**System information**

```
PyGMT information:
  version: v0.2.2.dev80+g132757a
System information:
  python: 3.6.12 | packaged by conda-forge | (default, Dec  9 2020, 00:36:02)  [GCC 9.3.0]
  executable: /home/mgrund/anaconda3/envs/pygmt/bin/python
  machine: Linux-4.4.0-19041-Microsoft-x86_64-with-debian-buster-sid
Dependency information:
  numpy: 1.19.4
  pandas: 1.1.5
  xarray: 0.16.2
  netCDF4: 1.5.6
  packaging: 20.8
  ghostscript: 9.53.3
  gmt: 6.2.0rc1
GMT library information:
  binary dir: /home/mgrund/anaconda3/envs/pygmt/bin
  cores: 8
  grid layout: rows
  library path: /home/mgrund/anaconda3/envs/pygmt/lib/libgmt.so
  padding: 2
  plugin dir: /home/mgrund/anaconda3/envs/pygmt/lib/gmt/plugins
  share dir: /home/mgrund/anaconda3/envs/pygmt/share/gmt
  version: 6.2.0rc1
```
@PaulWessel @meghanrjones could you please check if you can debug it?

michaelgrund:

This PR adds the `incols` parameter description to Figure.contour() and adds the deprecate_parameter decorator for backward-compatibility.

Fixes #1313.

- [x] Updates tests and examples to use the new parameter name
- [x] Use the deprecate_parameter decorator for backward-compatibility
- [x] Add a test for checking 'columns' backward compatibility

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_contour.py
+++ b/pygmt/tests/test_contour.py
@@ -129,7 +129,7 @@ def test_contour_deprecate_columns_to_incols(region):
 
     # generate dataframe
     # switch x and y from here onwards to simulate different column order
-    data = np.array([[y], [x], [z]])
+    data = np.array([y, x, z])
     data = pd.DataFrame(data=data)
 
     with pytest.warns(expected_warning=FutureWarning) as record:
--- a/pygmt/tests/test_contour.py
+++ b/pygmt/tests/test_contour.py
@@ -130,11 +130,11 @@ def test_contour_deprecate_columns_to_incols(region):
     # generate dataframe
     # switch x and y from here onwards to simulate different column order
     data = np.array([[y], [x], [z]])
-    df = pd.DataFrame(data=data)
+    data = pd.DataFrame(data=data)
 
     with pytest.warns(expected_warning=FutureWarning) as record:
         fig.contour(
-            data=df,
+            data=data,
             projection="X10c",
             region=region,
             frame="a",
--- a/pygmt/tests/test_contour.py
+++ b/pygmt/tests/test_contour.py
@@ -107,3 +107,34 @@ def test_contour_from_file(region):
         data=POINTS_DATA, projection="X10c", region=region, frame="af", pen="#ffcb87"
     )
     return fig
+
+
+@pytest.mark.mpl_image_compare(filename="test_contour_vec.png")
+def test_contour_deprecate_columns_to_incols(region):
+    """
+    Make sure that the old parameter "columns" is supported and it reports an
+    warning.
+
+    Modified from the test_contour_vec() test.
+    """
+    fig = Figure()
+    x, y = np.meshgrid(
+        np.linspace(region[0], region[1]), np.linspace(region[2], region[3])
+    )
+    y = x.flatten()  # switch x and y here to simulate different column order
+    x = y.flatten()
+    z = (x - 0.5 * (region[0] + region[1])) ** 2 + 4 * y ** 2
+    z = np.exp(-z / 10 ** 2 * np.log(2))
+    with pytest.warns(expected_warning=FutureWarning) as record:
+        fig.contour(
+            x=x,
+            y=y,
+            z=z,
+            projection="X10c",
+            region=region,
+            frame="a",
+            pen=True,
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_contour.py
```diff
@@ -107,3 +107,37 @@ def test_contour_from_file(region):
...
+        fig.contour(
+            data=data,
+            projection="X10c",
+            region=region,
+            frame="a",
+            pen=True,
+            columns=[1, 0, 2],
+        )
+        assert len(record) == 1  # check that only one warning was raised
+    return fig
```
weiji14, line 143: One thing I'm wondering is if we need to add deprecation tests for all of these methods (`contour` here and `rose` in #1306) since it'll be a chore to remove them in v0.6.0. Maybe having them in `fig.plot` is enough?

michaelgrund: In principle I agree with you @weiji14. However, especially the new test added to `test_contour.py` fails and seems like the `incols`/`columns` parameter is ignored at all so far. Thus, in my opinion it makes sense to test the deprecated parameters.

weiji14: Ok, let's keep the test then. And about why it's failing, it seems that the plot is outside of the `region` ([10, 70, -5, 10]). Not sure if this is an upstream bug or 'feature', but the region seems to be parsed before the column positions are swapped by incols (`-i`)?:

| expected | actual (wrong?) |
|--|--|
| correct | incorrect |

michaelgrund: Yes, the columns are not swapped although ``incols`` is correctly given to GMT via ``@GMTAPI@-S-I-D-M-T-N-000000 -Ba -JX10c -R10/70/-5/10 -W -i1,0,2``.

seisman: Looks like an upstream bug to me. Could you please create a new issue with a minimal example to reproduce it (see https://github.com/GenericMappingTools/pygmt/issues/1309 for an example), so that Paul and others can decide if it's an upstream issue and debug it.

seisman (approved):

pygmt/tests/test_contour.py
weiji14: Problem with -i parsing fixed upstream in https://github.com/GenericMappingTools/gmt/pull/5289, and the test passes now :tada:

weiji14 (approved):

PR #1310, opened 2021-05-29
Add download_test_data to download data files used in tests

seisman:
**Description of proposed changes**

Fixes #1111.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1312, opened 2021-06-01
Allow pygmt.which to accept a list of filenames as input

weiji14:

Let the fname parameter in `pygmt.which` accept list inputs. List of filenames are parsed using the `sequence_space` (see #325) and then passed into GMT C API as space delimited filenames. Also added a unit test to check that multi-file inputs work.

Use case for this is to download multiple files at once using `pygmt.which` without going through a for-loop, as in #1310.

Addresses https://github.com/GenericMappingTools/pygmt/pull/1310#discussion_r642747438

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/which.py
+++ b/pygmt/src/which.py
@@ -2,11 +2,18 @@
 which - Find the full path to specified files.
 """
 from pygmt.clib import Session
-from pygmt.helpers import GMTTempFile, build_arg_string, fmt_docstring, use_alias
+from pygmt.helpers import (
+    GMTTempFile,
+    build_arg_string,
+    fmt_docstring,
+    kwargs_to_strings,
+    use_alias,
+)
 
 
 @fmt_docstring
 @use_alias(G="download", V="verbose")
+@kwargs_to_strings(fname="sequence_space")
 def which(fname, **kwargs):
     """
     Find the full path to specified files.
@@ -27,8 +34,8 @@ def which(fname, **kwargs):
 
     Parameters
     ----------
-    fname : str
-        The file name that you want to check.
+    fname : str or list
+        One or more file names of any data type (grids, tables, etc.).
     download : bool or str
         If the file is downloadable and not found, we will try to download the
         it. Use True or 'l' (default) to download to the current directory. Use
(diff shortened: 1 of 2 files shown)


pygmt/src/which.py
```diff
@@ -27,8 +34,8 @@ def which(fname, **kwargs):
 
     Parameters
     ----------
-    fname : str
-        The file name that you want to check.
+    fname : str or list
+        One or more file names of any data type (grids, tables, etc.).
     download : bool or str
         If the file is downloadable and not found, we will try to download the
         it. Use True or 'l' (default) to download to the current directory. Use
```
weiji14, line 41: I'm tempted to update the download ([-G](https://docs.generic-mapping-tools.org/6.1/gmtwhich.html#g)) docstring to include the 'a' option (added in GMT 6.1 I think). Should it be done in this PR or a separate one?

```suggestion
    download : bool or str
        If the file is downloadable and not found, we will try to download
        it. Use True or 'l' (default) to download to the current directory. Use
```

core-man: A separate one is better.

seisman: > A separate one is better.

I agree.

seisman (approved):
Looks good to me.

core-man (approved):
Great~

pygmt/tests/test_which.py
```diff
@@ -18,10 +18,23 @@ def test_which():
         assert os.path.basename(cached_file) == fname
 
 
+def test_which_multiple():
+    """
+    Make sure which returns file paths for multiple @files correctly.
```
michaelgrund, line 23:
```suggestion
    Make sure "which" returns file paths for multiple @files correctly.
```

michaelgrund (approved):

pygmt/tests/test_which.py
weiji14: Good suggestion, 'which' does sound confusing. I've updated the docstrings for the other tests too (and used backticks "`" instead) in bb8206373abf452134294927696f57814947a3c7.

PR #1316, opened 2021-06-02
Raise a warning for the use of short-form parameters when long-forms are available

seisman:

Address the comment in https://github.com/GenericMappingTools/pygmt/issues/1203#issuecomment-845350743.

When long-form parameters are available, we should recommend users not to use
the short-form parameters.

For example, running the following example:
```
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], B=True, J="X10c")
fig.savefig("map.pdf")
```
will raise warnings like:
```
test.py:4: FutureWarning: Short-form parameter (J) is not recommended. Use long-form parameter 'projection' instead.
  fig.basemap(region=[0, 10, 0, 10], B=True, J="X10c")
test.py:4: FutureWarning: Short-form parameter (B) is not recommended. Use long-form parameter 'frame' instead.
  fig.basemap(region=[0, 10, 0, 10], B=True, J="X10c")
```

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -372,17 +372,18 @@ def use_alias(**aliases):
             """
             New module that parses and replaces the registered aliases.
             """
-            for arg, alias in aliases.items():
-                if alias in kwargs and arg in kwargs:
+            for short_param, long_alias in aliases.items():
+                if long_alias in kwargs and short_param in kwargs:
                     raise GMTInvalidInput(
-                        f"Parameters in short-form ({arg}) and long-form ({alias}) can't coexist."
+                        f"Parameters in short-form ({short_param}) and "
+                        f"long-form ({long_alias}) can't coexist."
                     )
-                if alias in kwargs:
-                    kwargs[arg] = kwargs.pop(alias)
-                elif arg in kwargs:
+                if long_alias in kwargs:
+                    kwargs[short_param] = kwargs.pop(long_alias)
+                elif short_param in kwargs:
                     msg = (
-                        f"Short-form parameter ({arg}) is not recommended. "
-                        f"Use long-form parameter '{alias}' instead."
+                        f"Short-form parameter ({short_param}) is not recommended. "
+                        f"Use long-form parameter '{long_alias}' instead."
                     )
                     warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
             return module_func(*args, **kwargs)
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -384,7 +384,7 @@ def use_alias(**aliases):
                         f"Short-form parameter ({arg}) is not recommended. "
                         f"Use long-form parameter '{alias}' instead."
                     )
-                    warnings.warn(msg, category=FutureWarning, stacklevel=2)
+                    warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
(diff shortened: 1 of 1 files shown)


pygmt/helpers/decorators.py
```diff
@@ -375,10 +375,16 @@ def new_module(*args, **kwargs):
...
+                        f"Parameters in short-form ({arg}) and long-form ({alias}) can't coexist."
                     )
                 if alias in kwargs:
                     kwargs[arg] = kwargs.pop(alias)
+                elif arg in kwargs:
+                    msg = (
+                        f"Short-form parameter ({arg}) is not recommended. "
+                        f"Use long-form parameter '{alias}' instead."
+                    )
+                    warnings.warn(msg, category=FutureWarning, stacklevel=2)
```
weiji14, line 387: Using `FutureWarning` implies a deprecation at some future date. Do you want to either:

1. set a specific PyGMT version (e.g. v0.5.0) to deprecate this 'feature'
2. Raise a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) instead
3. Raise a [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError) or GMTInvalidInput error?

pygmt/helpers/decorators.py
```diff
@@ -375,10 +375,16 @@ def new_module(*args, **kwargs):
             for arg, alias in aliases.items():
```
weiji14, line 375: I find the variables 'arg' and 'alias' confusing. Maybe we should be more explicit and call them 'short_param' (e.g. -J) and 'long_alias' (e.g. projection) instead? What do you think?

```suggestion
            for short_param, long_alias in aliases.items():
```

pygmt/helpers/decorators.py
seisman: > Using `FutureWarning` implies a deprecation at some future date. Do you want to either:
> 
> 1. set a specific PyGMT version (e.g. v0.5.0) to deprecate this 'feature'

As mentioned in https://github.com/GenericMappingTools/pygmt/issues/1203#issuecomment-845253702, I don't think we will fully deprecate this feature before PyGMT v1.0.0.

> 2. Raise a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) instead

Sounds good to me.

> 3. Raise a [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError) or GMTInvalidInput error?

I'm afraid it's a little too aggressive. BTW, the "single-letter" parameter is also used a lot in the PyGMT tests. Raising any errors will bring us troubles.

seisman: Changed to `SyntaxWarning` in commit 23e9dd67a999cb0c984dea13fb993434f4593944.

pygmt/helpers/decorators.py
seisman: Done in 30cf3e2cd2f7e33ad2fbc37cf081fdc02442899a.

weiji14: Looks much better, thanks!

weiji14 (approved):
Ok to approve. Technically we should have written a test for this, but there are actually some tests still using short parameter names (as mentioned in https://github.com/GenericMappingTools/pygmt/pull/1316#discussion_r646221803) so I think it's alright to just merge it in as is.

Issue #1318, opened 2021-06-04
Consistent table-like output for PyGMT functions/methods


maxrjones:
Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r639581006:

> Currently, the blockm*/grdtrack modules simply outputs pandas.DataFrame/filenames (see #1099) automatically depending on the input type (2 options). The implementation @willschlitzer made at bf58dc2 and 68fab7c returns either numpy/pandas/str depending on the user input (3 options). Best to be consistent across PyGMT on the types of outputs users can expect.

I am opening this issue to find out what the output format/options should be with the additional of x/y/z in blockmean, blockmedian, and grdtrack, as well as for new methods such as grd2xyz.

----

**Edit by @seisman on Oct 9, 2023:**

The PyGMT team has reach an agreement that PyGMT functions/methods should have consistent behavior for table-like outputs, i.e., the output depends on the `output_type` parameter.

Valid `output_type` values are:
- `file`: output to a file directly (usually need to set `outfile`)
- `numpy`: return a NumPy array (may not always possible because data must be in the same dtype in a 2-D numpy array)
- `pandas`: return a pandas.DataFrame (sometimes need to set `newcolname`)

Here are a list of functions/methods that have table-like outputs:

- [x] `blockm*` #3103
- [x] `filter1d` #3085
- [ ] `grdinfo` [Will be tracked in #593]
- [x] `grdtrack` #3106
- [x] `triangulate` #3107
- [x] `grdvolume` #3102
- [x] `select` #3108
- [ ] `which` [Will be tracked in #3003]
- [x] `grdhisteq` #3109
- [ ] `x2sys_cross` [Will be tracked in #3160]
- [ ] `info` [Will be tracked in #3159]
- [x] `project` #3110
- [x] `grd2xyz` #3097

We need to make sure they behave consistently.

weiji14:
Just wanted to say, this is a similar issue to that in the RAPIDS AI `cuml` library. See https://docs.rapids.ai/api/cuml/0.19/api.html#output-data-type-configuration. Their solution is to:

1. By default: 'mirror' the input data format to be the same as the output format as much as possible, i.e. `numpy` in/`numpy` out, or `pandas.DataFrame` in/`pandas.DataFrame` out.
[...]

PR #1336, opened 2021-06-16
WIP: Add return_table helper function

willschlitzer:
This is a proof-of-concept pull request for the `return_table`helper function in `utils.py`. It accepts the default string output of a table from the GMT C API and can return either a string, numpy array, or pandas DataFrame. This is a possible answer for #1318.

Fixes #

**Reminders**

**Slash Commands**


pygmt/helpers/utils.py
```diff
@@ -267,3 +269,48 @@ def args_in_kwargs(args, kwargs):
         If one of the required arguments is in ``kwargs``.
     """
     return any(arg in kwargs for arg in args)
+
+
+def return_table(result, data_format, format_parameter, df_columns):
```
meghanrjones, line 274: Looks promising! A few comments:
1. I think you will need to use this with one of the existing functions so that we can test it out in this PR. In my opinion, grdtrack would be a good option.
2. My preference would be to add a `format_options` parameter for `return_table`. This could take a list, with defaults values including numpy, pandas, and str. This way, if it doesn't for example make sense to return string values then that could not be given as an option in the individual function documentation/implementation.
3. table-like options for input are now `str or numpy.ndarray or pandas.DataFrame or xarray.Dataset or geopandas.GeoDataFrame`. It would be nice if all of these were output options too.
4. I would prefer a more description argument for requested data format than **a**|**d**|**s**. Something like `numpy`|`pandas`|`str` seems more readable.
5. I think `df_columns` needs to be optional.

willschlitzer: >     3. table-like options for input are now `str or numpy.ndarray or 
>     pandas.DataFrame or xarray.Dataset or geopandas.GeoDataFrame`. 
>     It would be nice if all of these were output options too.

Sounds good; I'll have to get smarter on the last two but I don't see why it should be a problem.

>     4. I would prefer a more description argument for requested data format than **a**|**d**|**s**. 
>     Something like `numpy`|`pandas`|`str` seems more readable.

I like the idea of keeping it short, especially when there is a default option (I anticipate it being a numpy array) and the strings are not also the same word as Python modules or variable types. But I understand how the single letters could be confusing.

> 
>     5. I think `df_columns` needs to be optional.

Since this is a helper function, I envisioned that the argument for `df_columns` would be set up in the GMT function that is using it, such as using `["x", "y", "z"]` when calling this function inside of `grd2xyz`.

willschlitzer:
Added `return_table()` to `grdtrack`. I have a few issues:

1. I'm not sure how to inform the user of the `ValueError` (lines 307-308) if it can't convert a value in the table to a float (such as a title or section header). My initial reaction was to print that it cannot convert the value to float, but that is a TON of prints, as every line of text is split up by word.
2. Specifically for `grdtrack`, I removed the use of `data_kind` since column names don't need to be specified by the user (but can be with the `df_columns=` parameter), but should there still be a check of the information format of the table?

pygmt/helpers/utils.py
willschlitzer: >     3. table-like options for input are now `str or numpy.ndarray or pandas.DataFrame or xarray.Dataset or geopandas.GeoDataFrame`. It would be nice if all of these were output options too.

Added in c8cef5e

willschlitzer:
I can't see why this is failing deployment, but it is running into a `ModuleNotFoundError: No module named 'geopandas'` for the Python 3.7 CI job. I asked the question in #1354 about making `geopandas` a dependency, and assume the fix for both pull requests will be the same.

willschlitzer:
@weiji14 I tried adding in `gpd = pytest.importorskip("geopandas")` but am still running into a `ModuleNotFound` error for geopandas on Python 3.7/NumPy 1.17. Any idea how to fix this to make the tests pass?

willschlitzer:
I'm unable to figure out why this is causing the deployments to fail. My guess is it has something to do with trying to incorporate it into `grdtrack` (as deployment and tests started to fail after that change) but I'm not sure why that is causing a problem.

pygmt/helpers/utils.py
```diff
@@ -10,6 +10,9 @@
 from collections.abc import Iterable
 from contextlib import contextmanager
 
+import geopandas as gpd
```
weiji14, line 13: This `import geopandas` line shouldn't be here at the top-level. I'd suggest importing geopandas in the `return_table` function itself if you need it, and only under `elif data_format=="geopandas"`.
```suggestion
```

pygmt/src/grdtrack.py
```diff
@@ -33,7 +34,14 @@
     n="interpolation",
 )
 @kwargs_to_strings(R="sequence", S="sequence")
-def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
+def grdtrack(
+    points,
+    grid,
+    data_format="d",
+    df_columns=["longitude", "latitude", "z-value"],
```
weiji14, line 41: Not sure if it's a good idea to have default column names, especially for the x (longitude) and y (latitude) columns since someone passing in a `pandas.DataFrame` table with existing column names would have their column names overridden by this default.

pygmt/helpers/utils.py
weiji14: >> 4. I would prefer a more description argument for requested data format than **a**|**d**|**s**. 
    Something like `numpy`|`pandas`|`str` seems more readable.

> I like the idea of keeping it short, especially when there is a default option (I anticipate it being a numpy array) and the strings are not also the same word as Python modules or variable types. But I understand how the single letters could be confusing.

I'll have to agree with Meghan that long descriptive names like `numpy`|`pandas`|`str` are preferable :slightly_smiling_face:

willschlitzer:
Closing this PR; I think there may be better ways to return table-like data, but I think the best move is to wrap some more of those functions that do return table like data before trying to figure out how to best refactor them to use a helper function.

PR #1345, opened 2021-06-21
Remove mention of multiple grid inputs in grdtrack docstring

willschlitzer:
The documentation for `grdtrack` states that it can accept multiple grid inputs for `grid`. As `grid=` can't be repeated in the same function call and it doesn't accept a list of grids, this pull request changes the documentation to state that the `grdtrack` accepts a single grid.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdtrack.py
+++ b/pygmt/src/grdtrack.py
@@ -37,10 +37,10 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
     r"""
     Sample grids at specified (x,y) locations.
 
-    Reads one or more grid files and a table (from file or an array input; but
+    Reads a grid file and a table (from file or an array input; but
     see ``profile`` for exception) with (x,y) [or (lon,lat)] positions in the
     first two columns (more columns may be present). It interpolates the
-    grid(s) at the positions in the table and writes out the table with the
+    grid at the positions in the table and writes out the table with the
     interpolated values added as (one or more) new columns. Alternatively
     (``crossprofile``), the input is considered to be line-segments and we
     create orthogonal cross-profiles at each data point or with an equidistant


pygmt/src/grdtrack.py
```diff
@@ -37,10 +37,10 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
     r"""
     Sample grids at specified (x,y) locations.
 
-    Reads one or more grid files and a table (from file or an array input; but
+    Reads a grid file and a table (from file or an array input; but
     see ``profile`` for exception) with (x,y) [or (lon,lat)] positions in the
     first two columns (more columns may be present). It interpolates the
-    grid(s) at the positions in the table and writes out the table with the
+    grid at the positions in the table and writes out the table with the
     interpolated values added as (one or more) new columns. Alternatively
```
willschlitzer, line 44: Is adding more than one new column possible with a single grid input, which I would assume could happen for a grid input with more than 3 dimensions? Or could this only be the case with multiple grid inputs in a single function?

pygmt/src/grdtrack.py
```diff
@@ -37,10 +37,10 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
     r"""
     Sample grids at specified (x,y) locations.
 
-    Reads one or more grid files and a table (from file or an array input; but
+    Reads a grid file and a table (from file or an array input; but
```
weiji14, line 40: > As grid= can't be repeated in the same function call and it doesn't accept a list of grids, this pull request changes the documentation to state that the grdtrack accepts a single grid.

Technically someone smart enough could figure out that doing `grid="grid1.nc -Ggrid2.nc -Ggrid3.nc"` would work, but it's more of a hacky solution :upside_down_face:

pygmt/src/grdtrack.py
weiji14: One could set `crossprofile`/[`-C`](https://docs.generic-mapping-tools.org/6.1/grdtrack.html#c) and get more columns (with just a single grid as input), but this would only work for file outputs, as 3 columns are currently assumed for pandas.DataFrame output (lon, lat, z).

weiji14 (commented):
Looks like you've stumbled on some limitations of the current pygmt.grdtrack function. Now my question is, do we want to

1. Restrict `pygmt.grdtrack` to only handle one grid? In which case this documentation PR would make sense, and users working on multiple grids would need to write for-loops or use GMT C instead; or
2. Add in an enhancement so that `pygmt.grdtrack` can handle multiple grids? Which would be a bit more involved, but is presumably something that needs to be done anyway at some point. Something related to this is the need to handle multiple column outputs as hinted on in https://github.com/GenericMappingTools/pygmt/pull/1345#discussion_r655171335 for people using `grdtrack -C`.

meghanrjones:
> Looks like you've stumbled on some limitations of the current pygmt.grdtrack function. Now my question is, do we want to
> 
>     1. Restrict `pygmt.grdtrack` to only handle one grid? In which case this documentation PR would make sense, and users working on multiple grids would need to write for-loops or use GMT C instead; or
> 
>     2. Add in an enhancement so that `pygmt.grdtrack` can handle multiple grids? Which would be a bit more involved, but is presumably something that needs to be done anyway at some point. Something related to this is the need to handle multiple column outputs as hinted on in [#1345 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1345#discussion_r655171335) for people using `grdtrack -C`.

I think option 2 is better.

willschlitzer:
> > Looks like you've stumbled on some limitations of the current pygmt.grdtrack function. Now my question is, do we want to
> > ```
> > 1. Restrict `pygmt.grdtrack` to only handle one grid? In which case this documentation PR would make sense, and users working on multiple grids would need to write for-loops or use GMT C instead; or
> > 
> > 2. Add in an enhancement so that `pygmt.grdtrack` can handle multiple grids? Which would be a bit more involved, but is presumably something that needs to be done anyway at some point. Something related to this is the need to handle multiple column outputs as hinted on in [#1345 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1345#discussion_r655171335) for people using `grdtrack -C`.
> > ```
> 
> I think option 2 is better.

While I'm in favor of making this PyGMT module as functionally equivalent to its GMT counterpart as possible, I think these two options are different in scope and not mutually exclusive. I think the documentation should match the capabilities of the current PyGMT module, and can be further updated to reflect when PyGMT's `grdtrack` has been upgraded to accept multiple grid inputs.

willschlitzer:
@weiji14 @meghanrjones Should I close this PR? It looks like you are in favor of modifying `grdtrack` to accept multiple grids.

meghanrjones:
> @weiji14 @meghanrjones Should I close this PR? It looks like you are in favor of modifying `grdtrack` to accept multiple grids.

I agree with your comment about updating the documentation for current usage here, then opening a separate issue/PR for the enhancement.

meghanrjones (approved):
I am fine with this intermediate solution, but an enhancement issue should be opened before merging so we do not lose track of the need to add support for multiple inputs.

meghanrjones:
Actually, I just noticed in the gmt docs the **-G+l***list* option. I have not tested if this would work as part of a str argument to the grid parameter in pygmt's grdtrack, but if it does then documenting that option would be a better intermediate solution for multiple grids IMO.

meghanrjones:
OK so pygmt.grdtrack works with multiple grid inputs but needs some improvements. Here's the test for that case:
In bash shell:
```
gmt grdcut @earth_relief_02m_g -Ggrid1.nc -R-118/-107/-49/-42
gmt grdcut @earth_age_02m_g -Ggrid2.nc -R-118/-107/-49/-42
echo "grid1.nc" > list.txt
echo "grid2.nc" >> list.txt
```
In python console:
```
from pygmt import grdtrack, which
from pygmt.datasets import load_ocean_ridge_points

dataframe = load_ocean_ridge_points()
output = grdtrack(points=dataframe, grid="+llist.txt",newcolname="data")
```

So in my opinion two things that need fixing are documenting grid="+lgrids" and updating newcolname to work with multiple column outputs. While I don't think that the `grid="grid1.nc -Ggrid2.nc -Ggrid3.nc"` hack should be considered a supported feature, since we do not always fully replicate gmt docs the `grid="+lgrids"` could be considered a feature. So, I no longer think we should remove mentions of multiple grid inputs.

willschlitzer:
Per @meghanrjones's comment, this change is no longer valid, so I'm closing the PR.

PR #1353, opened 2021-06-23
Move reused data to pytest fixtures in test_grdtrack.py

willschlitzer:
Five test functions in `test_grdtrack.py` use a dataframe from `load_ocean_ridge_points()` and called it inside the test functions. This pull requests moves it to `fixture_dataframe()`. 

Over five runs of `test_grdtrack.py` on my machine, the average time for the old tests was 906 ms and the new tests averaged 658 ms.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_grdtrack.py
+++ b/pygmt/tests/test_grdtrack.py
@@ -24,12 +24,14 @@ def fixture_dataarray():
         lat=slice(-49, -42), lon=slice(-118, -107)
     )
 
+
 @pytest.fixture(scope="module", name="dataframe")
 def fixture_dataframe():
     """
     Load the ocean ridge file.
     """
     return load_ocean_ridge_points()
+
 
 def test_grdtrack_input_dataframe_and_dataarray(dataarray, dataframe):
     """
--- a/pygmt/tests/test_grdtrack.py
+++ b/pygmt/tests/test_grdtrack.py
@@ -33,6 +33,22 @@ def fixture_dataframe():
     return load_ocean_ridge_points()
 
 
+@pytest.fixture(scope="module", name="csvfile")
+def fixture_csvfile():
+    """
+    Load the csvfile.
+    """
+    return which("@ridge.txt", download="c")
+
+
+@pytest.fixture(scope="module", name="ncfile")
+def fixture_ncfile():
+    """
+    Load the ncfile.
+    """
+    return which("@earth_relief_01d", download="a")
+
+
 def test_grdtrack_input_dataframe_and_dataarray(dataarray, dataframe):
     """
     Run grdtrack by passing in a pandas.DataFrame and xarray.DataArray as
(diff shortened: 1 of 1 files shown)


meghanrjones:
Thanks, @willschlitzer! I would be interested in whether adding the csvfile and ncfile as fixtures speeds it up even more (used 3 times each):

```
@pytest.fixture(scope="module", name="csvfile")
def fixture_csvfile():
    """
    Load the grid data from the sample earth_relief file.
    """
    return which("@ridge.txt", download="c")


@pytest.fixture(scope="module", name="ncfile")
def fixture_ncfile():
    """
    Load the grid data from the sample earth_relief file.
    """
    return which("@earth_relief_01d", download="a")
```

willschlitzer:
> Thanks, @willschlitzer! I would be interested in whether adding the csvfile and ncfile as fixtures speeds it up even more (used 3 times each):

Big time savings! 667 ms vs 411 ms, so with all of the fixtures this cuts down the test time by half!

Side note: Is there any reason we use `@pytest.fixture` instead of creating global variables in the script?

seisman (approved):

michaelgrund (approved):

weiji14 (approved):
Thanks @willschlitzer!

> Side note: Is there any reason we use @pytest.fixture instead of creating global variables in the script?

Good question. I think pytest fixtures (https://docs.pytest.org/en/latest/explanation/fixtures.html) are more explicit than global variables. Other than that, either fixtures or global variables would technically work for test_grdtrack.py. However, if we're going to parallelize tests (using something like [pytest-xdist](https://github.com/pytest-dev/pytest-xdist)) at some point, then fixtures would probably be preferrable.

seisman:
> Good question. I think pytest fixtures ([docs.pytest.org/en/latest/explanation/fixtures.html](https://docs.pytest.org/en/latest/explanation/fixtures.html)) are more explicit than global variables. Other than that, either fixtures or global variables would technically work for test_grdtrack.py. However, if we're going to parallelize tests (using something like [pytest-xdist](https://github.com/pytest-dev/pytest-xdist)) at some point, then fixtures would probably be preferrable.

I'm wondering if there are global pytest fixtures. After a quick search, I find this (https://betterprogramming.pub/understand-5-scopes-of-pytest-fixtures-1b607b5c19ed). It seems a `conftest.py` file is what we may need, but I don't read the full post.

weiji14:
> > Good question. I think pytest fixtures ([docs.pytest.org/en/latest/explanation/fixtures.html](https://docs.pytest.org/en/latest/explanation/fixtures.html)) are more explicit than global variables. Other than that, either fixtures or global variables would technically work for test_grdtrack.py. However, if we're going to parallelize tests (using something like [pytest-xdist](https://github.com/pytest-dev/pytest-xdist)) at some point, then fixtures would probably be preferrable.
> 
> I'm wondering if there are global pytest fixtures. After a quick search, I find this (https://betterprogramming.pub/understand-5-scopes-of-pytest-fixtures-1b607b5c19ed). It seems a `conftest.py` file is what we may need, but I don't read the full post.

Yes there are session, function, class, module and package scoped fixtures (see https://docs.pytest.org/en/latest/reference/fixtures.html#reference-fixtures). I'd prefer to keep the fixtures used by a test in the same test_*.py file and not to use a `conftest.py` file, because this makes it clearer where the fixtures come from (especially to new contributors unfamiliar with pytest).

seisman:
> I'd prefer to keep the fixtures used by a test in the same test_*.py file and not to use a `conftest.py` file, because this makes it clearer where the fixtures come from (especially to new contributors unfamiliar with pytest).

Yes, I agree. 

When I worked on the PR https://github.com/GenericMappingTools/pygmt/pull/1359, I felt that the variable `gmt_version` should be a global variable or a global fixture, so that we don't have to add the following lines when we want to access the GMT library version.
```
from pygmt import clib
with clib.Session() as _lib:
    gmt_version = Version(_lib.info["version"])
```

PR #1376, opened 2021-07-06
Change invalid input conditions in grdtrack

### LINKED ISSUE #1344: grdtrack raises an input error when points are not a DataFrame and no column name is specified
The documentation for `grdtrack` states that an argument for `newcolname` is required if the argument for `points` is a DataFrame. However, a `GMTInvalidInput` error requiring an argument for `newcolname` is raised if `points` is a numpy array or a list as well.

*Import statement*

```python
import pygmt
import numpy as np
grid = pygmt.datasets.load_earth_relief(region=[-5, 5, -5, 5])
```
*Points as a numpy array*
```python
points_matrix = np.array([[1,1], [1,2], [2,2]])
assert type(points_matrix) == np.ndarray
grdtrack_output = pygmt.grdtrack(points=points_matrix, grid=grid)
```
*Points as a list*
```python
points_list = [[1,1], [1,2], [2,2]]
assert type(points_list) == list
grdtrack_output = pygmt.grdtrack(points=points_list, grid=grid)
```

**Full error message**

```
GMTInvalidInput: Please pass in a str to 'newcolname'
```

The error is raised in lines 251-252 in `grdtrack.py`
```python
if data_kind(points) == "matrix" and newcolname is None:
    raise GMTInvalidInput("Please pass in a str to 'newcolname'")
```
It looks like `data_kind()` won't distinguish between a DataFrame and a numpy array and returns "matrix" in both instances. Should `newcolname` be a required parameter for all matrix-type data, since the function still returns a DataFrame if no outfile is specified, or should the error handling check to make sure that `points` is a DataFrame?

willschlitzer:
As discussed in #1344, the documentation for `grdtrack` states that an argument for `newcolname` is required for DataFrame inputs. However, `GMTInvalidInput` is raised if a list or numpy array is passed in. This changes the if statement to only raise the error if the input type is a DataFrame.

Fixes #1344 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdtrack.py
+++ b/pygmt/src/grdtrack.py
@@ -247,7 +247,7 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
         - None if ``outfile`` is set (track output will be stored in file set
           by ``outfile``)
     """
-    if isinstance(points, pd.DataFrame) and newcolname is None:
+    if hasattr(points, "columns") and newcolname is None:
         raise GMTInvalidInput("Please pass in a str to 'newcolname'")
 
     with GMTTempFile(suffix=".csv") as tmpfile:
--- a/pygmt/src/grdtrack.py
+++ b/pygmt/src/grdtrack.py
@@ -7,7 +7,6 @@ from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import (
     GMTTempFile,
     build_arg_string,
-    data_kind,
     fmt_docstring,
     kwargs_to_strings,
     use_alias,
--- a/pygmt/src/grdtrack.py
+++ b/pygmt/src/grdtrack.py
@@ -248,7 +248,7 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
         - None if ``outfile`` is set (track output will be stored in file set
           by ``outfile``)
     """
-    if data_kind(points) == "matrix" and newcolname is None:
+    if isinstance(points, pd.DataFrame) and newcolname is None:
         raise GMTInvalidInput("Please pass in a str to 'newcolname'")
 
     with GMTTempFile(suffix=".csv") as tmpfile:


pygmt/src/grdtrack.py
```diff
@@ -248,7 +247,7 @@ def grdtrack(points, grid, newcolname=None, outfile=None, **kwargs):
         - None if ``outfile`` is set (track output will be stored in file set
           by ``outfile``)
     """
-    if data_kind(points) == "matrix" and newcolname is None:
+    if isinstance(points, pd.DataFrame) and newcolname is None:
```
weiji14, line 250: Minor detail, but I'd prefer checking if the `points` variable has a `columns` attribute so that `pandas.DataFrame`, [`cudf.DataFrame`](https://docs.rapids.ai/api/cudf/stable/api.html#dataframe) and [`dask.DataFrame`](https://docs.dask.org/en/stable/dataframe-api.html#dask.dataframe.DataFrame) objects (or any PyData object which implements 'column') will work.

```suggestion
    if hasattr(points, "columns") and newcolname is None:
```

```python
import numpy as np
import pandas as pd
import dask.dataframe as dd
import cudf

numpy_points = np.array([[0.1, 0.2, 0.3]])
hasattr(numpy_points, "columns")  # False
pandas_points = pd.DataFrame(data=[[0.1, 0.2, 0.3]], columns=["x", "y", "z"])
hasattr(pandas_points, "columns")  # True
dask_points = dd.from_pandas(data=pandas_points, npartitions=1)
hasattr(dask_points, "columns")  # True
cudf_points = cudf.DataFrame(data=[[0.1, 0.2, 0.3]], columns=["x", "y", "z"])
hasattr(cudf_points, "columns")  # True
```

weiji14 (commented):
Thanks @willschlitzer, just one minor point.

pygmt/src/grdtrack.py
willschlitzer: @weiji14 I'm assuming this block of code is something you want added to `test_grdtrack.py`?

If so, how should the test import `dask` and `cudf`, as they are not listed as dependencies in `environment.yml`?

weiji14: Ideally we would test it yes, but the API of `cudf` and `dask` is supposed to mimic pandas so we don''t really need to either...

willschlitzer: So should I leave it out of the test then and just test with pandas DataFrame?

weiji14: > So should I leave it out of the test then and just test with pandas DataFrame?

yep

weiji14 (approved):

michaelgrund (approved):

Issue #1384, opened 2021-07-12
Add gallery example for grdclip


willschlitzer:
**Description of the desired feature**

It would be great to have some **gallery examples** on using the `grdclip` function. This will be a follow up to #1261.

## Examples of what you can do

These are just suggestions of things that you can plot. Feel free to come up with other ideas!

1. Plot seafloor bathymetry and clip all land above sea level
2. Plot an island and remove all seafloor data

## How to make the change :monocle_face:

1. First, read up our contributing guidelines at https://www.pygmt.org/v0.4.0/contributing.html#editing-the-documentation
2. Leave a comment below, stating that you will tackle this issue. We will assign you to this issue and you can then start working on it.
3. Go to e.g. https://github.com/GenericMappingTools/pygmt/tree/master/examples/gallery/maps, and click on 'Add File'. Or you may want to do this in your code editor locally.
4. Submit a pull request, and be sure to fill it up with as much detail as possible.

**Are you willing to help implement and maintain this feature?** Yes! I'm happy to help someone learn how to contribute to PyGMT!

seisman:
Closed by #1396.

PR #1396, opened 2021-07-26
Add gallery example for grdclip

michaelgrund:

Since noone worked on this issue during the scipy-sprint here's a gallery example adressing #1384.

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/images/grdclip.py
+++ b/examples/gallery/images/grdclip.py
@@ -21,8 +21,8 @@ grid = pygmt.datasets.load_earth_relief(resolution="03m", region=region)
 fig.basemap(region=region, projection="M12c", frame=["f", '+t"original grid"'])
 fig.grdimage(grid=grid, cmap="oleron")
 
-# Shift plot origin for second map 12.5 cm in x direction
-fig.shift_origin(xshift="12.5c")
+# Shift plot origin of the second map by "width of the first map + 0.5 cm"
+fig.shift_origin(xshift="w+0.5c")
 
 # Set all grid points < 0 m to a value of -2000 m.
 grid = pygmt.grdclip(grid, below=[0, -2000])
--- a/examples/gallery/images/grdclip.py
+++ b/examples/gallery/images/grdclip.py
@@ -1,0 +1,22 @@
+import pygmt
+
+fig = pygmt.Figure()
+
+# Load sample grid and use area around the Hawaiian Islands
+grid = pygmt.datasets.load_earth_relief(resolution="01m", region = [-162, -153, 18, 23])
+
+# Plot original grid 
+fig.basemap(region = [-162, -153, 18, 23], projection="M12c", frame=["f", '+t"original grid"'])
+fig.grdimage(grid=grid, cmap="oleron")
+
+fig.shift_origin(yshift = "-9c")
+
+# Set all grid points < 0 to a value of -2000.
+grid = pygmt.grdclip(grid, below = [0, -2000])
+
+# Plot clipped grid 
+fig.basemap(region = [-162, -153, 18, 23], projection="M12c", frame=["f", '+t"clipped grid"'])
+fig.grdimage(grid=grid)
+fig.colorbar(frame=["x+lElevation", "y+lm"])
+
+fig.show()
--- a/examples/gallery/images/grdclip.py
+++ b/examples/gallery/images/grdclip.py
@@ -3,19 +3,23 @@ import pygmt
 fig = pygmt.Figure()
 
 # Load sample grid and use area around the Hawaiian Islands
-grid = pygmt.datasets.load_earth_relief(resolution="01m", region = [-162, -153, 18, 23])
+grid = pygmt.datasets.load_earth_relief(resolution="01m", region=[-162, -153, 18, 23])
 
-# Plot original grid 
-fig.basemap(region = [-162, -153, 18, 23], projection="M12c", frame=["f", '+t"original grid"'])
+# Plot original grid
+fig.basemap(
+    region=[-162, -153, 18, 23], projection="M12c", frame=["f", '+t"original grid"']
+)
 fig.grdimage(grid=grid, cmap="oleron")
 
-fig.shift_origin(yshift = "-9c")
+fig.shift_origin(yshift="-9c")
 
 # Set all grid points < 0 to a value of -2000.
-grid = pygmt.grdclip(grid, below = [0, -2000])
+grid = pygmt.grdclip(grid, below=[0, -2000])
 
-# Plot clipped grid 
-fig.basemap(region = [-162, -153, 18, 23], projection="M12c", frame=["f", '+t"clipped grid"'])
+# Plot clipped grid
+fig.basemap(
+    region=[-162, -153, 18, 23], projection="M12c", frame=["f", '+t"clipped grid"']
+)
 fig.grdimage(grid=grid)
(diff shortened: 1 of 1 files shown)


examples/gallery/images/grdclip.py
```diff
@@ -0,0 +1,35 @@
+"""
+Clip grid values
+----------------
```
seisman, line 3:
```suggestion
Clipping grid values
--------------------
```

examples/gallery/images/grdclip.py
```diff
@@ -0,0 +1,35 @@
...
+than 0 m (in general the bathymetric part of the grid) to a common value of
+-2000 m via the ``below`` parameter.
+"""
+
+import pygmt
+
+fig = pygmt.Figure()
+
+# Load sample grid (1 arc minute global relief) and use area around the Hawaiian Islands
+grid = pygmt.datasets.load_earth_relief(resolution="01m", region=[-162, -153, 18, 23])
```
seisman, line 15: The 1 arc minute global relief data is [NOT cached](https://github.com/GenericMappingTools/pygmt/blob/6a4b8715bf00cd42cae8d5405e35e08a4764e97c/pygmt/helpers/testing.py#L145) in the CI. We should either cache the 01m data or use a lower resolution earth relief.

examples/gallery/images/grdclip.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Plot original grid
+fig.basemap(
+    region=[-162, -153, 18, 23], projection="M12c", frame=["f", '+t"original grid"']
+)
+fig.grdimage(grid=grid, cmap="oleron")
+
+fig.shift_origin(yshift="-9c")
```
seisman, line 23: Why not use `subplot`?

examples/gallery/images/grdclip.py
michaelgrund: I would prefer to use a high resolution as the 1 arc minute global relief, otherwise the map looks bad. Will add the 01m data to cache in a separate PR.

examples/gallery/images/grdclip.py
michaelgrund: In my opinion the whole example looks less crowded when using ```fig.shift_origin()``` here. However, I'm open to adjust it.

core-man: I think both are okay. But, do you think it's better to compare the two figures with a 1X2 grid layout (left-right) instead of the current 2X1 (top-bottom)?

examples/gallery/images/grdclip.py
weiji14: Or can we use a different location? I like the Hawaii example, but caching the 1 arc minute data (258MB) isn't very feasible... How about another island like Iceland (or another medium-ish sized one).

seisman: > Or can we use a different location? I like the Hawaii example, but caching the 1 arc minute data (258MB) isn't very feasible... How about another island like Iceland (or another medium-ish sized one).

Sounds a better solution.

examples/gallery/images/grdclip.py
seisman: > do you think it's better to compare the two figures with a 1X2 grid layout (left-right) instead of the current 2X1 (top-bottom)?

I think a 1x2 layout is better. The current 2x1 layout makes the image too long to display without scrolling down the page.

michaelgrund: Made some adjustements, thanks for all the input 😉

examples/gallery/images/grdclip.py
michaelgrund: I changed it to the 3 arc minute grid for the area around Iceland.

examples/gallery/images/grdclip.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Load sample grid (3 arc minute global relief) in target area
+grid = pygmt.datasets.load_earth_relief(resolution="03m", region=region)
+
+# Plot original grid
+fig.basemap(region=region, projection="M12c", frame=["f", '+t"original grid"'])
+fig.grdimage(grid=grid, cmap="oleron")
+
+# Shift plot origin for second map 12.5 cm in x direction
+fig.shift_origin(xshift="12.5c")
```
seisman, line 25: In case you don't know it, we can use **h** in Y shifts and **w** in X shifts, in which **h** and **w** means *height* and *width* of the previous plot. 
```suggestion
# Shift plot origin of the second map by "width of the first map + 0.5 cm"
fig.shift_origin(xshift="w+0.5c")
```

michaelgrund: > 
> 
> In case you don't know it, we can use **h** in Y shifts and **w** in X shifts, in which **h** and **w** means _height_ and _width_ of the previous plot. See https://docs.generic-mapping-tools.org/latest/cookbook/options.html#plot-positioning-and-layout-the-x-y-options for detailed explanations. Feel free to make more changes to the comment to better explain the `xshift` argument.

Indeed, I didn't know that so far, thanks @seisman.

examples/gallery/images/grdclip.py
weiji14: The current cache size is 30.5MB (https://github.com/GenericMappingTools/pygmt/actions/runs/1064597430), so adding 27MB would make it 50.5MB. I'm still a bit hesitant to add it... How about adding just the tile that is downloaded. Running `gmt grdcut @earth_relief_03m -R-28/-10/62/68 -Gtemp.nc` gives me the tile name "N00W090.earth_relief_03m_p" in `.gmt/server/earth/earth_relief/earth_relief_03m_p` so maybe try just caching that?

michaelgrund: > 
> 
> The current cache size is 30.5MB (https://github.com/GenericMappingTools/pygmt/actions/runs/1064597430), so adding 27MB would make it 50.5MB. I'm still a bit hesitant to add it... How about adding just the tile that is downloaded. Running `gmt grdcut @earth_relief_03m -R-28/-10/62/68 -Gtemp.nc` gives me the tile name "N00W090.earth_relief_03m_p" in `.gmt/server/earth/earth_relief/earth_relief_03m_p` so maybe try just caching that?

Does adding "N00W090.earth_relief_03m_p" allow to automatically use only this tile for plotting?

core-man: See another similar case:
https://github.com/GenericMappingTools/pygmt/blob/6a4b8715bf00cd42cae8d5405e35e08a4764e97c/pygmt/helpers/testing.py#L163

I think we could only cache such a sinngle tile.

michaelgrund: > 
> 
> See another similar case:
> 
> https://github.com/GenericMappingTools/pygmt/blob/6a4b8715bf00cd42cae8d5405e35e08a4764e97c/pygmt/helpers/testing.py#L163
> 
> I think we could only cache such a single tile.

I already recognized this one 😉 . My question is: Is that tile automatically used by ```pygmt.datasets.load_earth_relief()``` when the corresponding region is defined?

seisman: > Is that tile automatically used by `pygmt.datasets.load_earth_relief()` when the corresponding region is defined?

Yes.

willschlitzer (approved):

seisman (approved):

core-man (approved):
Great~~~We have one more gallery~ :smile:

PR #984, opened 2021-03-01
Add Jiayuan Yao to AUTHORS.md

core-man:
**Description of proposed changes**

I am happy to help the pygmt developers with this meaningful project although I am a new Python user.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1405, opened 2021-08-02
Plot square or cube by default for geopandas Point/MultiPoint types

### LINKED ISSUE #1373: Plot point type geopandas.GeoDataFrames as points directly instead of as lines

:warning: Note to SciPy 2021 sprinters: this is a bit more of an advanced feature enhancement request, which you are happy to help with if you good with GeoPandas and PyGMT. But if not, please take a look at other [`scipy-sprint`](https://github.com/GenericMappingTools/pygmt/labels/scipy-sprint) issues which are easier :warning:

PyGMT added initial support for plotting geopandas.GeoDataFrame objects in #1000 (see also #608). However, GMT by default will plot lines even if you give it points like so:

```python
import geopandas as gpd
import pygmt

cities: gpd.GeoDataFrame = gpd.read_file(gpd.datasets.get_path("naturalearth_cities"))
cities.head()
# 	name	geometry
# 0	Vatican City	POINT (12.45339 41.90328)
# 1	San Marino	POINT (12.44177 43.93610)
# 2	Vaduz	POINT (9.51667 47.13372)
# 3	Luxembourg	POINT (6.13000 49.61166)
# 4	Palikir	POINT (158.14997 6.91664)

fig = pygmt.Figure()
fig.plot(data=cities, color="black")
fig.show()
```
cities as lines

Having lines show up can be counter-intuitive, especially to new users unfamiliar with how GMT works. The workaround is to use `fig.plot(data=cities, style="s0.2c", color="black")` so that circles are plotted instead of lines. Ideally however, a POINT based GeoDataFrame should just be plotted as that, points.

cities as points

## What needs to be done :monocle_face:

To enable point GeoDataFrames to be plotted directly as points, there are probably a few ways we can update PyGMT's code. 

### Solution 1

Update the `plot.py` code to add a default style when points are given but no `style`/`-S` argument is given. This would involve changing the code somewhere along here:

https://github.com/GenericMappingTools/pygmt/blob/d90b3fc889b53633deab6b4ab77612ac7a247c1b/pygmt/src/plot.py#L210-L211

Note that the code for `plot3d` would also need to be updated too at:

https://github.com/GenericMappingTools/pygmt/blob/d90b3fc889b53633deab6b4ab77612ac7a247c1b/pygmt/src/plot3d.py#L179-L180

### Solution 2

Alternatively, we could set the style in the temporary OGR_GMT file somehow, which would involve modifying the code somewhere along here:

https://github.com/GenericMappingTools/pygmt/blob/d90b3fc889b53633deab6b4ab77612ac7a247c1b/pygmt/helpers/tempfile.py#L133-L134

### Solution 3

Anyone have a better idea? Let's discuss!

**Are you willing to help implement and maintain this feature?** Happy to mentor someone on tackling this issue.

yohaimagen:
In response to issue #1373  adding if statement to the plot function to create desirable default behavior for plotting geopandas dataframe with points data(geometry) to be plotted as points instead of a line.

Fixes #1373


yohaimagen:
I am not very happy with my solution. 
first of all, as we discuss in #1373 it involves the plot function which has many many more utilities than just plotting geopandas DataFrams so it seems a bit too specific for this function. 
second, I need to make sure that `data` is geopandas DataFrame before I am checking if it contains points only. In order to do so without importing geopandas I am checking that data contains the attribute _geometry_column_name `"_geometry_column_name" in data.__dict__` by doing so I am assuming no other data option(string, np.ndarray etc...) will have this attribute. Another option will be to import geopands and to directly check that it is a geopandas DataFeame instant.

So if you have an idea for better implementation ill be happy to know.

pygmt/src/plot.py
```diff
@@ -209,6 +209,14 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
     extra_arrays = []
     if "S" in kwargs and kwargs["S"][0] in "vV" and direction is not None:
         extra_arrays.extend(direction)
+    elif (
+        "S" not in kwargs
+        and data is not None
+        and data is object
+        and "_geometry_column_name" in data.__dict__
+        and (data.geom_type == "Point").all()
```
weiji14, line 217: While this works, I do have two questions:

1. Will MultiPoint geometry types work?
2. Do you think it's ok to set the default Point style if **all** rows are points? I have concerns about how fast the check is say on 1000 rows (it may well be fast but I'm not sure).

pygmt/src/plot.py
```diff
@@ -209,6 +209,14 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
...
+    ):
+        kwargs["S"] = "s0.2c"
```
weiji14, line 219: I'm ok with a 0.2cm square as a default style for points, but wanted to hear from @GenericMappingTools/pygmt-maintainers too. Would other [shapes](https://docs.generic-mapping-tools.org/latest/plot.html#s) be better? E.g. **p**oint (a 1 pixel dot), **c**ircle, etc? Also, is 0.2cm a good size, or should we just go without setting a size, i.e. `kwargs["S"] = "s"`.

weiji14 (commented):
You've got the right idea here to check for an attribute that only geopandas.GeoDataFrame objects have :smile: I've offered a few suggestions that can help to improve things a bit.

Once we settle on the if-statement condition, it would be good to:

1. copy the same chunk of code to `plot3d.py`.
2. Add a unit test to `test_geopandas.py` to ensure this works. Are you interested in learning about unit test writing? If so, we can give you some tips on that.

P.S. I've converted this Pull Request (PR) to draft mode to save on Continuous Integration resources (test will only run on Linux instead of all 3 OS types). We will convert the PR back to non-draft mode once things are more polished.

pygmt/src/plot.py
yohaimagen: 1. the current implementation works only with Point, I don't think we should support mixed geometrys in geopandas DataFrame it's a bad practice. we can add support for MultiPoint alon with `(data.geom_type == "Point").all() or (data.geom_type == "MultiPoint").all()`
2.  We don't need to worry about the execution time of `(data.geom_type == "Point").all()` because the execution time of writing the temporary file is larger (probably also plotting the figure but I didn't check)
from:
`import numpy as np
import pandas as pd
import geopandas as gpd
import timeit

x = np.random.uniform(0, 10, 1000)
y = np.random.uniform(0, 10, 1000)
df = pd.DataFrame(dict(x=x, y=y))
gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df.x, df.y))

num = 1000
print(timeit.timeit(lambda: (gdf.geom_type == 'Point').all(), number=num) / num)
print(timeit.timeit(lambda: gdf.to_file(filename='ttt'), number=num) / num)`

we get:
0.0010915064080000007
0.055181070641

pygmt/src/plot.py
michaelgrund: I would prefer either **p**oint (since no unit is required) or **c**ircle (currently not sure which size would be a good choice).

pygmt/src/plot.py
```diff
@@ -209,6 +209,8 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
     extra_arrays = []
     if "S" in kwargs and kwargs["S"][0] in "vV" and direction is not None:
         extra_arrays.extend(direction)
+    elif "S" not in kwargs and kind == "geojson" and (data.geom_type == "Point").all():
```
weiji14, line 212: > 1. the current implementation works only with Point, I don't think we should support mixed geometrys in geopandas DataFrame it's a bad practice. we can add support for MultiPoint alon with `(data.geom_type == "Point").all() or (data.geom_type == "MultiPoint").all()`

Agree with not supporting mixed geometries. Let's go with using `data.geom_type.isin(["Point", "MultiPoint"]).all()` (see [`pandas.Series.isin` ](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.isin.html)) which will do the same as the `or` operator.

```suggestion
    elif "S" not in kwargs and kind == "geojson" and data.geom_type.isin(["Point", "MultiPoint"]).all():
```

> 2. We don't need to worry about the execution time of `(data.geom_type == "Point").all()` because the execution time of writing the temporary file is larger (probably also plotting the figure but I didn't check)

Good to have those numbers! 1 millisecond is acceptable. I've tried to check both Point and MultiPoint and the timings aren't too bad either.

```python
%%timeit
gdf.geom_type.isin(["Point", "MultiPoint"]).all()
# 1.33 ms ± 13.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

yohaimagen:
I am not sure about the `p` style. at least in my example, you can't see the 1 pixel points

from 
`
fig.plot(
    projection='X10',
    frame='f',
    region=[0, 10, 0, 10, 0, 10],
    data=gdf,
)
`
we get 
plot
 and from 
`
fig.plot(
    projection='X10',
    frame='f',
    region=[0, 10, 0, 10, 0, 10],
    data=gdf,
    style='s0.2c'
)
`
we get 

plot_s0 2c

the same goes for plot3d with and without `style='u0.2c'`

plot3d
plot3d_u0 2c

weiji14:
> I am not sure about the `p` style. at least in my example, you almost can't see the 1-pixel points

Hmm, you're right. Let's go back to square 0.2cm then for `plot`, and perhaps we should use `u0.2c` for `plot3d` as you suggested. Sorry for the back and forth!

yohaimagen:
it's O.K.

now I'll be happy to learn about the tests. thanks!

weiji14:
> now I'll be happy to learn about the tests. thanks!

Alright! Testing involves checking that code produces the expected behaviour. Specifically, you'll need to write some new test functions to put in [`pygmt/tests/test_geopandas.py`](https://github.com/GenericMappingTools/pygmt/blob/master/pygmt/tests/test_geopandas.py), and ensure that the function will generate an image that matches a **baseline**. The guidelines are at https://www.pygmt.org/dev/contributing.html#testing-plots.

The test itself can be adapted from what you did in https://github.com/GenericMappingTools/pygmt/pull/1405#issuecomment-891549918. You might also find the examples in [`test_plot.py`](https://github.com/GenericMappingTools/pygmt/blob/master/pygmt/tests/test_plot.py) useful. Without giving you too much of a full answer, it should look something like this:

```python
def test_geopandas_plot_default_square():
    """
    """
    point = shapely.geometry.Point(...)
    gdf = gpd.GeoDataFrame(...)
    fig = Figure()
    fig.plot(data=gdf, ...)
    return fig
```

Start with the above for a 2D plot with a Point type geometry, and once you've got the hang of it, try making a 3D plot with a MultiPoint type geometry. Good luck!

yohaimagen:
@weiji14 can you please add me as a collaborator at DAGshub.

weiji14:
> @weiji14 can you please add me as a collaborator at DAGshub.

Done! You should now be able to do `dvc push`.

yohaimagen:
I am falling on 2 tests

I've added `from pygmt import Figure` and falling on formatting test with the error 
"Imports are incorrectly sorted and/or formatted"
what are the formatting guidelines for imports?

I am also falling on DVC image diff which I do not quite understand. would be happy for an explanation.

weiji14:
This is an auto-generated report of images that have changed on the DVC remote
  
  | Status | Path |
  |----------|-------------------------------------------------------------|
  | added | pygmt/tests/baseline/test_geopandas_plot3d_default_cube.png |
  | added | pygmt/tests/baseline/test_geopandas_plot_default_square.png |
  
  ## Image diff(s)
  
  <details>
  
  ### Added images
  
  - pygmt/tests/baseline/test_geopandas_plot3d_default_cube.png 
  
  
  - pygmt/tests/baseline/test_geopandas_plot_default_square.png 
  
  
  ### Modified images
  
  | Path | Old | New |
  |---|---|---|
  </details>
  
  Report last updated at commit 1979249c9060111011b3dbf696cb1e73bc2a4a30

weiji14:
> I've added `from pygmt import Figure` and falling on formatting test with the error
> "Imports are incorrectly sorted and/or formatted"
> what are the formatting guidelines for imports?

We use [`isort`](https://pycqa.github.io/isort/) to sort all the import statements as per https://www.pygmt.org/dev/contributing.html?highlight=isort#code-style. If you leave a comment with `/format`, there will be a bot that will sort it for you automatically in GitHub. Alternatively, run `make format` in your terminal locally (assuming you're on macOS/Linux) and that should do the trick too. 

> I am also falling on DVC image diff which I do not quite understand. would be happy for an explanation.

The error is ok. It's just that the DVC image workflow doesn't work from forks due to some permissions issues. I've manually posted what things should look like in https://github.com/GenericMappingTools/pygmt/pull/1405#issuecomment-893052271.

If you're keen on contributing more in the future, I can invite you to the [pygmt-contributors](https://github.com/orgs/GenericMappingTools/teams/pygmt-contributors) team so that you can directly create branches on the main GenericMappingTools/pygmt repository next time which will prevent the error from showing up :grin:

yohaimagen:
O.K I've reformated the imports, from my end its looks like we are done here let me know if there is still anything you think we should add. 
I will be keen to contribute more in the futere :), the link in [pygmt-contributors ](https://github.com/orgs/GenericMappingTools/teams/pygmt-contributors) is broken.

pygmt/tests/test_geopandas.py
```diff
@@ -64,3 +64,40 @@ def test_geopandas_info_shapely(gdf, geomtype, desired):
...
+@pytest.mark.mpl_image_compare
+def test_geopandas_plot_default_square():
+    """
+    Check the default behavior of plotting geopandas DataFrame with Point
+    geometry in 2d.
+    """
+    point = shapely.geometry.Point(5, 5)
+    gdf = gpd.GeoDataFrame(dict(geometry=[point]))
+    fig = Figure()
+    fig.plot(data=gdf, region=[0, 10, 0, 10], projection="X10", frame=True)
```
weiji14, line 78: To speed up the test, we try to keep the figure small-ish. There's a lot of whitespace with a 10x10 plot, so maybe trim it to 2x2cm instead.

```suggestion
    point = shapely.geometry.Point(1, 2)
    gdf = gpd.GeoDataFrame(geometry=[point])
    fig = Figure()
    fig.plot(data=gdf, region=[0, 2, 1, 3], projection="X2c", frame=True)
```

pygmt/tests/test_geopandas.py
```diff
@@ -64,3 +64,40 @@ def test_geopandas_info_shapely(gdf, geomtype, desired):
...
+    fig.plot3d(
+        data=gdf,
+        perspective=[315, 25],
+        region=[0, 10, 0, 10, 0, 10],
+        frame=[
+            "wsnez3",
+            "xafg",
+            "yafg",
+            "zafg",
+        ],
```
weiji14, line 100: Same here, the figure is originally 102KB, but we can reduce it down to 12.1KB using the code below.

```suggestion
    mp = shapely.geometry.MultiPoint([(0.5, 0.5, 0.5), (1.5, 1.5, 1.5)])
    gdf = gpd.GeoDataFrame(geometry=[mp])
    fig = Figure()
    fig.plot3d(
        data=gdf,
        perspective=[315, 25],
        region=[0, 2, 0, 2, 0, 2],
        projection="X2c",
        frame=["WsNeZ1", "xag", "yag", "zag"],
```

weiji14 (commented):
> O.K I've reformated the imports, from my end its looks like we are done here let me know if there is still anything you think we should add.

Good work on the unit tests! I have a few recommendations to reduce the figure size (makes the test run faster), but other than that, I'm happy with it :grinning:

> I will be keen to contribute more in the futere :), the link in [pygmt-contributors ](https://github.com/orgs/GenericMappingTools/teams/pygmt-contributors) is broken.

Ah sorry, that seems to be an internal link. Please open up a separate PR adding your name to AUTHORS.md (see e.g. previous PR at #984 on how to do it), and we'll get onboard the PyGMT team!

yohaimagen:
I've changed the sizes of the figures, will open a separate pull request to add myself to `AUTHERS.md`.

weiji14 (approved):
> I've changed the sizes of the figures, will open a separate pull request to add myself to `AUTHERS.md`.

Excellent! Happy to approve this PR. I'll set a `final-review-call` label so someone else can have a look, otherwise we typically will merge things in ~24-48 hours.

pygmt/tests/test_geopandas.py
```diff
@@ -64,3 +64,36 @@ def test_geopandas_info_shapely(gdf, geomtype, desired):
     geom = gdf.loc[geomtype].geometry
     output = info(table=geom, per_column=True)
     npt.assert_allclose(actual=output, desired=desired)
+
+
+@pytest.mark.mpl_image_compare
+def test_geopandas_plot_default_square():
+    """
+    Check the default behavior of plotting geopandas DataFrame with Point
```
michaelgrund, line 72:
```suggestion
    Check the default behavior of plotting a geopandas DataFrame with Point
```

pygmt/tests/test_geopandas.py
```diff
@@ -64,3 +64,36 @@ def test_geopandas_info_shapely(gdf, geomtype, desired):
...
+    gdf = gpd.GeoDataFrame(geometry=[point])
+    fig = Figure()
+    fig.plot(data=gdf, region=[0, 2, 1, 3], projection="X2c", frame=True)
+    return fig
+
+
+@pytest.mark.mpl_image_compare
+def test_geopandas_plot3d_default_cube():
+    """
+    Check the default behavior of plotting geopandas DataFrame with MultiPoint
```
michaelgrund, line 85:
```suggestion
    Check the default behavior of plotting a geopandas DataFrame with MultiPoint
```

michaelgrund (commented):
The dvc image diff test fails. Is that caused by the draft status or went something wrong with the upload @weiji14 ?

meghanrjones:
> The dvc image diff test fails. Is that caused by the draft status or went something wrong with the upload @weiji14 ?

dvc image failure is not a problem, see this post for an explanation https://github.com/GenericMappingTools/pygmt/pull/1405#issuecomment-893055638.

seisman:
Please hold off merging this PR and see my comment in https://github.com/GenericMappingTools/pygmt/issues/1373#issuecomment-895415166 first.

yohaimagen:
I'll add them

pygmt/src/plot.py
```diff
@@ -207,6 +207,12 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
     extra_arrays = []
     if "S" in kwargs and kwargs["S"][0] in "vV" and direction is not None:
         extra_arrays.extend(direction)
+    elif (
```
seisman, line 210: It would be better if you can add a short comment here to explain what these lines does.

pygmt/src/plot3d.py
```diff
@@ -177,6 +177,12 @@ def plot3d(
     extra_arrays = []
     if "S" in kwargs and kwargs["S"][0] in "vV" and direction is not None:
         extra_arrays.extend(direction)
+    elif (
```
seisman, line 180: Same here.

pygmt/src/plot.py
```diff
@@ -207,6 +207,12 @@ def plot(self, x=None, y=None, data=None, size=None, direction=None, **kwargs):
...
+        "S" not in kwargs
+        and kind == "geojson"
+        and data.geom_type.isin(["Point", "MultiPoint"]).all()
+    ):  # chacking if the geometry of a geoDataFrame is Point or MultiPoint
```
seisman, line 214:
```suggestion
    ):  # checking if the geometry of a geoDataFrame is Point or MultiPoint
```

pygmt/src/plot3d.py
```diff
@@ -177,6 +177,12 @@ def plot3d(
...
+        "S" not in kwargs
+        and kind == "geojson"
+        and data.geom_type.isin(["Point", "MultiPoint"]).all()
+    ):  # chacking if the geometry of a geoDataFrame is Point or MultiPoint
```
seisman, line 184:
```suggestion
    ):  # checking if the geometry of a geoDataFrame is Point or MultiPoint
```

seisman (approved):
Looks good to me.

yohaimagen:
great thanks for the reviwe!

seisman:
Ping @weiji14 and @michaelgrund to give this PR another round of final review, because this PR has changed a little after you approved it.

weiji14:
I think it looks fine, just need to incorporate the suggestions at https://github.com/GenericMappingTools/pygmt/pull/1405#discussion_r684802717 and should be good to merge :smile:

yohaimagen:
Sorry did not notice them, committed the changes.

weiji14:
Cool, I'll merge this PR in now, thanks again @yohaimagen!

Issue #1421, opened 2021-08-08
Missing (required) parameters in grdlandmask


michaelgrund:
**Description of the desired feature**

I thought about generating a gallery example highlighting how to use `grdlandmask` (similar to #1396 for `grdclip`) based on the examples given under https://docs.generic-mapping-tools.org/6.2/grdlandmask.html. Please correct me if I'm wrong but for the case of plotting shoreslines already included in GMT at least the alias (and description) for parameter **-N** is missing, isn't it?

**Are you willing to help implement and maintain this feature?** Yes

meghanrjones:
-N is not a required arguments since there is a default in GMT (setting values over water as 0 and values over land as 1). It is certainly useful since it allows completing tasks such as those in the examples. Feel free to propose an alias in a PR, since there is not already a PR open for adding more aliases to grdlandmask.

PR #1423, opened 2021-08-09
Update/add grdlandmask parameters and aliases

michaelgrund:

As discussed in #1421 this PR adds further parameters and corresponding aliases to ``grdlandmask``.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdlandmask.py
+++ b/pygmt/src/grdlandmask.py
@@ -64,7 +64,7 @@ def grdlandmask(**kwargs):
     bordervalues : bool or str or float or list
         Nodes that fall exactly on a polygon boundary should be
         considered to be outside the polygon [Default considers them to be
-        inside]. Alternatively, append either a list of four values 
+        inside]. Alternatively, append either a list of four values
         [*cborder*, *lborder*, *iborder*, *pborder*] or just the single value *bordervalue*
         (for the case when they should all be the same value). This turns on
         the line-tracking mode. Now, after setting the mask values specified
@@ -75,12 +75,12 @@ def grdlandmask(**kwargs):
         outlines, and *pborder* for ponds-in-islands-in-lakes outlines
         [Default is no line tracing].
     maskvalues : str or list
-        [*wet*, *dry*] or [*ocean*, *land*, *lake*, *island*, *pond*]. 
+        [*wet*, *dry*] or [*ocean*, *land*, *lake*, *island*, *pond*].
         Sets the values that will be assigned to nodes. Values can
-        be any number, including the textstring NaN 
+        be any number, including the textstring NaN
         [Default is 0/1/0/1/0 (i.e., 0/1)]. Also select
         ``bordervalues`` to let nodes exactly on feature boundaries be
-        considered outside [Default is inside]. 
+        considered outside [Default is inside].
     {V}
     {r}
 
--- a/pygmt/src/grdlandmask.py
+++ b/pygmt/src/grdlandmask.py
@@ -16,13 +16,13 @@ from pygmt.helpers import (
 
 @fmt_docstring
 @use_alias(
-    G="outgrid",
-    I="spacing",
-    R="region",
     A="area_thresh",
     D="resolution",
     E="bordervalues",
+    G="outgrid",
+    I="spacing",
     N="maskvalues",
+    R="region",
     V="verbose",
     r="registration",
 )
@@ -51,7 +51,7 @@ def grdlandmask(**kwargs):
     {R}
     {A}
     resolution : str
-        *res*\[\ **+f**\]. Selects the resolution of the data set to use      **f**\|\ **h**\|\ **i**\|\ **l**\|\ **c**
+        *res*\[\ **+f**\]. Selects the resolution of the data set to use
         ((**f**)ull, (**h**)igh, (**i**)ntermediate, (**l**)ow, or
         (**c**)rude). The resolution drops off by ~80% between data sets.
         [Default is **l**]. Append **+f** to automatically select a lower
--- a/pygmt/src/grdlandmask.py
+++ b/pygmt/src/grdlandmask.py
@@ -49,8 +49,18 @@ def grdlandmask(**kwargs):
         in.
     {I}
     {R}
-    area_thresh: todo
-    resolution : todo
+    {A}
+    resolution : str
+        res[**+f**]. Selects the resolution of the data set to use
+        ((**f**)ull, (**h**)igh, (**i**)ntermediate, (**l**)ow, or
+        (**c**)rude). The resolution drops off by ~80% between data sets.
+        [Default is **l**]. Append **+f** to automatically select a lower
+        resolution should the one requested not be available
+        [abort if not found]. Alternatively, choose (**a**)uto to automatically
+        select the best resolution given the chosen region. Note that because
+        the coastlines differ in details a node in a mask file using one
+        resolution is not guaranteed to remain inside [or outside] when a
+        different resolution is selected.
     bordervalues : todo
     mask_geog : todo
     {V}
--- a/pygmt/src/grdlandmask.py
+++ b/pygmt/src/grdlandmask.py
@@ -26,7 +26,7 @@ from pygmt.helpers import (
     V="verbose",
     r="registration",
 )
-@kwargs_to_strings(R="sequence")
+@kwargs_to_strings(R="sequence", N="sequence", E="sequence")
 def grdlandmask(**kwargs):
     r"""
     Create a grid file with set values for land and water.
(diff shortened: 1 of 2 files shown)


pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,41 @@ def grdlandmask(**kwargs):
         in.
     {I}
     {R}
+    {A}
+    resolution : str
+        *res*\[\ **+f**\]. Selects the resolution of the data set to use      **f**\|\ **h**\|\ **i**\|\ **l**\|\ **c**
```
weiji14, line 54: Not sure what happened here...
```suggestion
        *res*\[\ **+f**\]. Selects the resolution of the data set to use
```

weiji14 (commented):
Thanks @michaelgrund! By the way, could you swap the order of `grdlandmask` and `grdgradient` at https://github.com/GenericMappingTools/pygmt/blame/v0.4.1/doc/api/index.rst#L95-L96? Just noticed they weren't in alphabetical order.

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,41 @@ def grdlandmask(**kwargs):
...
+        [abort if not found]. Alternatively, choose (**a**)uto to automatically
+        select the best resolution given the chosen region. Note that because
+        the coastlines differ in details a node in a mask file using one
+        resolution is not guaranteed to remain inside [or outside] when a
+        different resolution is selected.
+    bordervalues : str or list
+        [*values*]. Nodes that fall exactly on a polygon boundary should be
+        considered to be outside the polygon [Default considers them to be
+        inside]. Alternatively, append either the four values *cborder*/
+        *lborder*/*iborder*/*pborder* or just the single value *bordervalue*
```
seisman, line 68: Need a backslash here, otherwise there is a whitespace before `lborder`.
```suggestion
        inside]. Alternatively, append either the four values *cborder*/\
        *lborder*/*iborder*/*pborder* or just the single value *bordervalue*
```

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,41 @@ def grdlandmask(**kwargs):
...
+        (**c**)rude). The resolution drops off by ~80% between data sets.
+        [Default is **l**]. Append **+f** to automatically select a lower
+        resolution should the one requested not be available
+        [abort if not found]. Alternatively, choose (**a**)uto to automatically
+        select the best resolution given the chosen region. Note that because
+        the coastlines differ in details a node in a mask file using one
+        resolution is not guaranteed to remain inside [or outside] when a
+        different resolution is selected.
+    bordervalues : str or list
+        [*values*]. Nodes that fall exactly on a polygon boundary should be
```
seisman, line 65: I'm wondering if the docstring is more clear in this way:
```suggestion
    bordervalues : str or list or bool
        Nodes that fall exactly on a polygon boundary should be
```

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,41 @@ def grdlandmask(**kwargs):
...
+        [Default is no line tracing].
+    maskvalues : str or list
+        *values*. Sets the values that will be assigned to nodes. Values can
+        be any number, including the textstring NaN. Also select
+        ``bordervalues`` to let nodes exactly on feature boundaries be
+        considered outside [Default is inside]. Specify this information
+        using 1 of 2 formats:
+
+        - ``maskvalues`` \*wet*/*dry*.
+        - ``maskvalues`` \*ocean*/*land*/*lake*/*island*/*pond*.
```
seisman, line 85: Maybe better?
```suggestion
        considered outside [Default is inside]. Specify this information
        using either *wet*/*dry* or *ocean*/*land*/*lake*/*island*/*pond*.
```

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,41 @@ def grdlandmask(**kwargs):
...
+        ((**f**)ull, (**h**)igh, (**i**)ntermediate, (**l**)ow, or
+        (**c**)rude). The resolution drops off by ~80% between data sets.
+        [Default is **l**]. Append **+f** to automatically select a lower
+        resolution should the one requested not be available
+        [abort if not found]. Alternatively, choose (**a**)uto to automatically
+        select the best resolution given the chosen region. Note that because
+        the coastlines differ in details a node in a mask file using one
+        resolution is not guaranteed to remain inside [or outside] when a
+        different resolution is selected.
+    bordervalues : str or list
```
seisman, line 64: Need to update the `kwargs_to_strings` decorator to let bordervalues and maskvalues support list.

pygmt/src/grdlandmask.py
michaelgrund: Much better, thanks!

michaelgrund (commented):
Thanks @seisman!

pygmt/src/grdlandmask.py
michaelgrund: Done @seiman, thanks!

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,38 @@ def grdlandmask(**kwargs):
...
+        inside]. Alternatively, append either the four values *cborder*/\
+        *lborder*/*iborder*/*pborder* or just the single value *bordervalue*
+        (for the case when they should all be the same value). This turns on
+        the line-tracking mode. Now, after setting the mask values specified
+        via ``maskvalues`` we trace the lines and change the node values for
+        all cells traversed by a line to the corresponding border value. Here,
+        *cborder* is used for cells traversed by the coastline, *lborder* for
+        cells traversed by a lake outline, *iborder* for islands-in-lakes
+        outlines, and *pborder* for ponds-in-islands-in-lakes outlines
+        [Default is no line tracing].
```
seisman, line 76: Recommend the usage of a list of four values, rather than slash-separated values?
```suggestion
    bordervalues : bool or str or float or list
        Nodes that fall exactly on a polygon boundary should be
        considered to be outside the polygon [Default considers them to be
        inside]. Alternatively, append either a list of four values 
        [*cborder*, *lborder*, *iborder*, *pborder*] or just the single value *bordervalue*
        (for the case when they should all be the same value). This turns on
        the line-tracking mode. Now, after setting the mask values specified
        via ``maskvalues`` we trace the lines and change the node values for
        all cells traversed by a line to the corresponding border value. Here,
        *cborder* is used for cells traversed by the coastline, *lborder* for
        cells traversed by a lake outline, *iborder* for islands-in-lakes
        outlines, and *pborder* for ponds-in-islands-in-lakes outlines
        [Default is no line tracing].
```

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,38 @@ def grdlandmask(**kwargs):
...
+    maskvalues : str or list
+        *values*. Sets the values that will be assigned to nodes. Values can
+        be any number, including the textstring NaN. Also select
+        ``bordervalues`` to let nodes exactly on feature boundaries be
+        considered outside [Default is inside]. Specify this information
+        using either *wet*/*dry* or *ocean*/*land*/*lake*/*island*/*pond*.
+        [Default is 0/1/0/1/0 (i.e., 0/1)].
```
seisman, line 83: Perhaps write the docstring in this way is better.
```suggestion
        [*wet*, *dry*] or [*ocean*, *land*, *lake*, *island*, *pond*]. 
        Sets the values that will be assigned to nodes. Values can
        be any number, including the textstring NaN 
        [Default is 0/1/0/1/0 (i.e., 0/1)]. Also select
        ``bordervalues`` to let nodes exactly on feature boundaries be
        considered outside [Default is inside]. 
```

pygmt/src/grdlandmask.py
michaelgrund: Agree with you @seisman!

pygmt/src/grdlandmask.py
```diff
@@ -45,6 +49,38 @@ def grdlandmask(**kwargs):
...
+        node values for all cells traversed by a line to the corresponding
+        border value. Here, *cborder* is used for cells traversed by the
+        coastline, *lborder* for cells traversed by a lake outline, *iborder*
+        for islands-in-lakes outlines, and *pborder* for
+        ponds-in-islands-in-lakes outlines [Default is no line tracing].
+    maskvalues : str or list
+        [*wet*, *dry*] or [*ocean*, *land*, *lake*, *island*, *pond*].
+        Sets the values that will be assigned to nodes. Values can
+        be any number, including the textstring NaN
+        [Default is 0/1/0/1/0 (i.e., 0/1)]. Also select
```
seisman, line 81:
```suggestion
        [Default is [0, 1, 0, 1, 0] (i.e., [0, 1])]. Also select
```

seisman (approved):

weiji14 (approved):

PR #1426, opened 2021-08-09
Add area_thresh to COMMON_OPTIONS

michaelgrund:

While working on #1423 I recognized that the **-A** alias already implemented in ``coast`` as **area_thresh** is also used in several other modules like ``grdlandmask``, ``grdmath`` and ``gmtselect``. This PR adds **area_thresh** to the common options list.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -31,7 +31,8 @@ COMMON_OPTIONS = {
             [**s**\|\ **S**]][**+l**\|\ **r**][**+p**\ *percent*].
             Features with an area smaller than *min_area* in km\ :sup:`2` or of
             hierarchical level that is lower than *min_level* or higher than
-            *max_level* will not be plotted.""",
+            *max_level* will not be plotted [Default is 0/0/4 (all
+            features)].""",
     "B": r"""
         frame : bool or str or list
             Set map boundary
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -28,7 +28,7 @@ COMMON_OPTIONS = {
     "A": r"""
         area_thresh : int, float, or str
             *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
-            [**s**\|\ **S**][**+l**\|\ **r**][**+p**\ *percent*].
+            [**s**\|\ **S**]][**+l**\|\ **r**][**+p**\ *percent*].
             Features with an area smaller than *min_area* in km\ :sup:`2` or of
             hierarchical level that is lower than *min_level* or higher than
             *max_level* will not be plotted.""",
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -26,7 +26,7 @@ COMMON_OPTIONS = {
             *projcode*\[*projparams*/]\ *width*.
             Select map :doc:`projection </projections/index>`.""",
     "A": r"""
-        area_thresh : int, float, or str
+        area_thresh : int or float or str
             *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
             [**s**\|\ **S**]][**+l**\|\ **r**][**+p**\ *percent*].
             Features with an area smaller than *min_area* in km\ :sup:`2` or of
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -27,11 +27,11 @@ COMMON_OPTIONS = {
             Select map :doc:`projection </projections/index>`.""",
     "A": r"""
         area_thresh : int, float, or str
-        *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
-        [**s**\|\ **S**][**+l**\|\ **r**][**+p**\ *percent*].
-        Features with an area smaller than *min_area* in km\ :sup:`2` or of
-        hierarchical level that is lower than *min_level* or higher than
-        *max_level* will not be plotted.""",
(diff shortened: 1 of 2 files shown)


seisman:
It sounds reasonable to me. Two thoughts:

1. Also update the `area_thresh` docstring of `coast` in this PR?
2. Although these options (e.g, `area_thresh`, `cmap`, `pen`) are used in many modules, they are not "strict" common options. I think it's OK to list the common and "non-common" options in the `COMMON_OPTIONS` dictionary, but it would be better to add a few comments explaining that the dictionary contains both common and non-common options, and it would be better to organize them in two groups (separated by a few comment lines). This could be done in a separate PR if agreed.

seisman:
> 2\. Although these options (e.g, `area_thresh`, `cmap`, `pen`) are used in many modules, they are not real common options. I think it's OK to list the common and "non-common" options in the same `COMMON_OPTIONS` dictionary, but it would be better to add a few comments explaining that the dictionary contains both common and non-common options, and it would be better to organize them in two groups (separated by a few comment lines). This could be done in a separate PR if agreed.

Or we can define two separate dicts (e.g., COMMON_OPTIONS and NONCOMMON_OPTIONS) and another dict that is the combination of both.

michaelgrund:
> 
> 
> It sounds reasonable to me. Two thoughts:
> 
>     1. Also update the `area_thresh` docstring of `coast` in this PR?

Will do that!

> 
>     2. Although these options (e.g, `area_thresh`, `cmap`, `pen`) are used in many modules, they are not real common options. I think it's OK to list the common and "non-common" options in the same `COMMON_OPTIONS` dictionary, but it would be better to add a few comments explaining that the dictionary contains both common and non-common options, and it would be better to organize them in two groups (separated by a few comment lines). This could be done in a separate PR if agreed.

I would go with this option @seisman. Maybe a good first issue to add some comments explaining that the dictionary contains both, common and non-common options?

seisman (approved):
Looks good to me.

pygmt/helpers/decorators.py
```diff
@@ -25,6 +25,13 @@
             *Required if this is the first plot command*.
             *projcode*\[*projparams*/]\ *width*.
             Select map :doc:`projection </projections/index>`.""",
+    "A": r"""
+        area_thresh : int, float, or str
+        *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
+        [**s**\|\ **S**][**+l**\|\ **r**][**+p**\ *percent*].
+        Features with an area smaller than *min_area* in km\ :sup:`2` or of
+        hierarchical level that is lower than *min_level* or higher than
+        *max_level* will not be plotted.""",
```
meghanrjones, line 34:
```suggestion
            *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
            [**s**\|\ **S**][**+l**\|\ **r**][**+p**\ *percent*].
            Features with an area smaller than *min_area* in km\ :sup:`2` or of
            hierarchical level that is lower than *min_level* or higher than
            *max_level* will not be plotted.""",
```
Formatting in the preview is off - I suspect the indentation needs to be further for these lines.

michaelgrund: Of course, thanks @meghanrjones 😉.

pygmt/helpers/decorators.py
```diff
@@ -25,6 +25,13 @@
             *Required if this is the first plot command*.
             *projcode*\[*projparams*/]\ *width*.
             Select map :doc:`projection </projections/index>`.""",
+    "A": r"""
+        area_thresh : int, float, or str
+            *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
+            [**s**\|\ **S**][**+l**\|\ **r**][**+p**\ *percent*].
```
meghanrjones, line 31:
```suggestion
            [**s**\|\ **S**]][**+l**\|\ **r**][**+p**\ *percent*].
```

meghanrjones (approved):
Just missing one close bracket, otherwise looks good. Thanks!

pygmt/helpers/decorators.py
```diff
@@ -25,6 +25,13 @@
             *Required if this is the first plot command*.
             *projcode*\[*projparams*/]\ *width*.
             Select map :doc:`projection </projections/index>`.""",
+    "A": r"""
+        area_thresh : int, float, or str
+            *min_area*\ [/*min_level*/*max_level*][**+a**\[**g**\|\ **i**]\
+            [**s**\|\ **S**]][**+l**\|\ **r**][**+p**\ *percent*].
+            Features with an area smaller than *min_area* in km\ :sup:`2` or of
+            hierarchical level that is lower than *min_level* or higher than
+            *max_level* will not be plotted.""",
```
weiji14, line 34: I see that the default `-A` is `0/0/4` for all the modules you mentioned (`grdlandmask`, `gmtselect`, etc), so maybe we should add this?

```suggestion
            *max_level* will not be plotted [Default is 0/0/4 (all features)].""",
```

michaelgrund: Agree with you!

weiji14 (approved):
:grinning:

pygmt/helpers/decorators.py
```diff
@@ -25,6 +25,14 @@
             *Required if this is the first plot command*.
             *projcode*\[*projparams*/]\ *width*.
             Select map :doc:`projection </projections/index>`.""",
+    "A": r"""
+        area_thresh : int, float, or str
```
weiji14, line 29: Oh wait, just noticed this. Should use 'or' as a separator.
```suggestion
        area_thresh : int or float or str
```

Issue #1427, opened 2021-08-10
Wrapper for gmtselect


weiji14:
**Description of the desired feature**

Implement [`gmtselect`](https://docs.generic-mapping-tools.org/6.2/gmtselect.html) which "Select data table subsets based on multiple spatial criteria".

Mentioned at https://twitter.com/ATiredGradStdnt/status/1417189659365498884, also partially used before at https://forum.generic-mapping-tools.org/t/gmt-example-18-with-pygmt-volumes-and-spatial-selections/1585.

This function will probably be named `select`, i.e. called using `pygmt.select()`. As a table in, table out function, the implementation would be similar to already wrapped functions like `blockmedian`/`blockmean`. The code snippet would look something like so:

https://github.com/GenericMappingTools/pygmt/blob/31980e8af2bc8074af325fd9caa2355040088084/pygmt/src/blockm.py#L41-L64

Xref table I/O discussion at #1268 and #1318. Best to wrap this module after #1426.

**Are you willing to help implement and maintain this feature?** Yes

seisman:
Can we close the issue, since the initial wrapper has been done in #1429? Or do we still want to keep this issue open to tracking missing aliases, and gallery examples? Keeping the issue open makes me think that it has not been wrapped.

Related to https://github.com/GenericMappingTools/pygmt/issues/1599.

weiji14:
Ok, closing as initial implementation done at #1429. Tracking progress on project board https://github.com/GenericMappingTools/pygmt/projects/9#card-72007028 instead.

PR #1429, opened 2021-08-10
Wrap gmtselect

weiji14:

Wrapping the [gmtselect](https://docs.generic-mapping-tools.org/6.2/gmtselect.html) function which "Selects data table subsets based on multiple spatial criteria". For a GMT example, see https://docs.generic-mapping-tools.org/6.2/gallery/ex24.html.

Crossref GMT.jl docs at https://www.generic-mapping-tools.org/GMT.jl/v0.31/#GMT.gmtselect.

Parameters/Aliases to wrap:

- [ ] `A` area_thresh
- [ ] `C` dist2pt?
- [ ] `D` resolution?
- [ ] `E` boundary
- [ ] `F` polygon
- [ ] `G` gridmask
- [x] `I` reverse
- [ ] `L` dist2line?
- [x] `R` region
- [ ] `N` mask
- [x] `Z` z_subregion/zsubregion

To keep this PR small, I would prefer implementing only 3 out of 7 of the possible `gmtselect` filters, specifically `-A` (once #1426 is done), `-R` rectangular region subset, and `-Z` z value data subset. This, plus the reverse (`-I`) option. The others can be done in a follow-up PR.

Other implementation points to consider:
- GMT's `select` function supports multiple file inputs, should we support this in PyGMT? See e.g. `text` at 7a08ebd7a5d1b185f31b84265617d1816d2d8610. Related to this is if we should support multiple table-like inputs (e.g. two `geopandas.GeoDataFrame` objects in a list).
- I've used the long alias `z_subregion` for `-Z` (following [`pygmt.grdcut`](https://www.pygmt.org/v0.4.1/api/generated/pygmt.grdcut.html), xref #492). Should we use `zsubregion` (no underscore) instead, following policy set in https://github.com/GenericMappingTools/pygmt/blame/v0.4.1/doc/contributing.md#L446-L450)?

Addresses first part of #1427.

**Reminders**

**Slash Commands**


willschlitzer:
@weiji14 Is the intent just to have the output of this as the file or a DataFrame, or should their be an option for a numpy array as well?

weiji14:
> @weiji14 Is the intent just to have the output of this as the file or a DataFrame, or should their be an option for a numpy array as well?

Maybe just a `pandas.DataFrame` for now, until someone makes a reusable function to handle #1318.

pygmt/src/select.py
```diff
@@ -0,0 +1,146 @@
...
+
+    Full option list at :gmt-docs:`gmtselect.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    data : str or {table-like}
+        Pass in either a file name to an ASCII data table, a 2D
+        {table-classes}.
```
willschlitzer, line 64:
```suggestion
        Pass in either a file name to an ASCII data table or a 2D
        {table-classes}.
```

I"m not familiar with using `gmtselect`, but can't it accept 3D tables (as indicated by `z_subregion`?

pygmt/src/select.py
```diff
@@ -0,0 +1,146 @@
...
+    outfile : str
+        The file name for the output ASCII file.
+    {A}
+    reverse : str
+        [**cflrsz**].
+        Reverses the sense of the test for each of the criteria specified:
```
willschlitzer, line 70:
```suggestion
        Reverses the criteria of the test for each of the criteria specified:
```
The use of "sense" doesn't make much sense to me. I know it's from the GMT docs, but would there be a better way to phrase this?

PR #1431, opened 2021-08-10
Wrap ternary

willschlitzer:
This pull request wraps the GMT function `ternary`.

**Preview** docs at https://pygmt-git-wrap-ternary-gmt.vercel.app/api/generated/pygmt.Figure.ternary.html

**Reminders**

**Slash Commands**


willschlitzer:
I'm getting a weird problem where the axis labels on the right-side (**b**) axis isn't showing up. I'm not sure what I'm missing, as this follows along with the example on the GMT page for `ternary`, and there are still gridlines for that axis appearing in the final figure.

```
fig = pygmt.Figure()
pygmt.makecpt(cmap="turbo", series=[0, 80, 10])
fig.ternary(table="@ternary.txt",
            region="0/100/0/100/0/100",
            cmap=True,
            projection="X6i",
            frame= ["bafg+lAir",
                   "cafg+lLimestone",
                   "aafg+lWater"],
            S="c0.1c")
fig.show()
```
image

meghanrjones:
> I'm getting a weird problem where the axis labels on the right-side (**b**) axis isn't showing up. I'm not sure what I'm missing, as this follows along with the example on the GMT page for `ternary`, and there are still gridlines for that axis appearing in the final figure.
> 
> ```
> fig = pygmt.Figure()
> pygmt.makecpt(cmap="turbo", series=[0, 80, 10])
> fig.ternary(table="@ternary.txt",
>             region="0/100/0/100/0/100",
>             cmap=True,
>             projection="X6i",
>             frame= ["bafg+lAir",
>                    "cafg+lLimestone",
>                    "aafg+lWater"],
>             S="c0.1c")
> fig.show()
> ```
> 
> image

This was a bug in GMT in 6.2.0, which was fixed by https://github.com/GenericMappingTools/gmt/pull/5431.

meghanrjones:
> > This was a bug in GMT in 6.2.0, which was fixed by [GenericMappingTools/gmt#5431](https://github.com/GenericMappingTools/gmt/pull/5431).
> 
> @meghanrjones Is there a workaround for GMT 6.2.0, or will this plotting bug be the case until GMT releases a new version and PyGMT uses it?

You could try using pygmt.config(MAP_FRAME_AXIS, "WSE"), though using that internally would prevent users from overriding it unless you found a way to check whether the setting has been configured in the session.

willschlitzer:
> 
> You could try using pygmt.config(MAP_FRAME_AXIS, "WSE"), though using that internally would prevent users from overriding it unless you found a way to check whether the setting has been configured in the session.
> 
> Edit: To be more clear, it would need to be in a context manager for pygmt.ternary calls so that it doesn't impact other function calls.

I think calling `ternary` with a context manager will cause more confusion when none of the other plotting functions do, especially since this will presumably go back to normal in the next release of GMT (which I'm assuming will be after PyGMT v0.5.0 release). My thought is to just hold off on merging the upstream fix takes effect, as there are other methods to make `ternary` plots with Python.

weiji14:
Since we've bumped PyGMT's minimum GMT version to 6.3.0, do you want to give this PR another go @willschlitzer?

pygmt/src/ternary.py
```diff
@@ -0,0 +1,59 @@
+"""
+Histogram - Create a histogram
```
weiji14, line 2:
```suggestion
ternary - Plot data on ternary diagrams.
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,59 @@
...
+from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
+
+
+@fmt_docstring
+@use_alias(
+    B="frame",
+    C="cmap",
+    F="center",
+    G="fill",
+    J="projection",
```
weiji14, line 14: Seems like `tenary` uses [`-JX`](https://docs.generic-mapping-tools.org/6.3/ternary.html#j) only, so should we alias it like so?

```suggestion
    JX="projection",
```

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,23 @@
+"""
+Tests for pygmt.ternary.
```
weiji14, line 2:
```suggestion
Tests ternary.
```

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,23 @@
...
+def test_ternary():
+    """
+    Test plotting a ternary chart.
+    """
+    fig = Figure()
+    makecpt(cmap="turbo", series=[0, 80, 10])
+    fig.ternary(
+        table="@ternary.txt",
+        region="0/100/0/100/0/100",
+        cmap=True,
```
weiji14, line 18: Is it possible to modify the test so that it doesn't require `makecpt`? Just want this to be a proper 'unit' test that tests the `ternary` function only.

weiji14 (commented):
First round of review. Implementation look ok, but the docs could use a bit of work. Could you please add `ternary` to `doc/api/index.rst` so that we can see the preview docs? Thanks!

pygmt/tests/test_ternary.py
willschlitzer: I'll see if I can use a color pallete without `makecpt`; I've only tried it once but was unsuccessful.

willschlitzer: It's not the best graph, but I changed the test to use `cmap="red,orange,yellow,green,blue,violet",`. It doesn't apply color to all of the plotted data points, but I want to avoid an exceptionally long string for the CPT.

pygmt/src/ternary.py
```diff
@@ -0,0 +1,59 @@
...
+    p="perspective",
+    t="transparency",
+)
+@kwargs_to_strings(R="sequence", T="sequence", c="sequence_comma", p="sequence")
+def ternary(self, table, **kwargs):
+    r"""
+    Plots a histogram, and can read data from a file or
+    list, array, or dataframe.
+
+    Full option list at :gmt-docs:`histogram.html`
```
weiji14, line 31:
```suggestion
    Full option list at :gmt-docs:`ternary.html`
```

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,22 @@
...
+    Test plotting a ternary chart.
+    """
+    fig = Figure()
+    fig.ternary(
+        table="@ternary.txt",
+        region="0/100/0/100/0/100",
+        cmap="red,orange,yellow,green,blue,violet",
+        projection="6i",
+        frame=["bafg+lAir", "cafg+lLimestone", "aafg+lWater"],
+        style="c0.1c",
```
weiji14, line 20: Set an outline pen so that the white dots still show up.
```suggestion
        style="c0.1c",
        pen="thinnest",
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,59 @@
...
+    X="xshift",
+    Y="yshift",
+    p="perspective",
+    t="transparency",
+)
+@kwargs_to_strings(R="sequence", T="sequence", c="sequence_comma", p="sequence")
+def ternary(self, table, **kwargs):
+    r"""
+    Plots a histogram, and can read data from a file or
+    list, array, or dataframe.
```
weiji14, line 29: Probably need to make this docstring more Pythonic, but here's a first shot copied from https://github.com/GenericMappingTools/gmt/blob/6.3.0/doc/rst/source/ternary.rst
```suggestion
    Reads (*a*,\ *b*,\ *c*\ [,\ *z*]) records from *table* and plots symbols at
    those locations on a ternary diagram. If a symbol is selected and no symbol
    size given, then we will interpret the fourth column of the input data as
    symbol size. Symbols whose *size* is <= 0 are skipped. If no symbols are
    specified then the symbol code (see **style** below) must be present as
    last column in the input.  If **style** is not specified then we instead
    plot lines or polygons.
```

pygmt/tests/baseline/test_ternary.png.dvc
```diff
@@ -0,0 +1,4 @@
+outs:
+- md5: 9f6b57e9dd82aa24a8a013176ba8fb1d
+  size: 128744
```
weiji14, line 3: Make sure to update the plot with the new pen.

willschlitzer: Updated in 33d4cb0

pygmt/src/ternary.py
```diff
@@ -0,0 +1,64 @@
...
+from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
+
+
+@fmt_docstring
+@use_alias(
+    B="frame",
+    C="cmap",
+    F="center",
+    G="fill",
+    JX="projection",
```
seisman, line 14: I don't think `JX="projection"` is a good alias for ternary. For other modules, we usually pass `projection="X10c"`, but in ternary, we need to pass `projection="10c"`. Such inconsistency may confuse users. Can we simply use alias `JX` to `withd` instead?

willschlitzer: updated

pygmt/src/ternary.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Parameters
+    ----------
+    table : str or list or {table-like}
+        Pass in either a file name to an ASCII data table, a Python list, a 2D
+        {table-classes}.
+    {J}
+    width : str
+        *width*. Set the width of the figure by passing a number, followed by
+        a unit (**i** for inches, **c** for centimeters).
```
seisman, line 48: Please update the docstring of `width` following the official documentation at https://docs.generic-mapping-tools.org/latest/ternary.html#j

pygmt/src/ternary.py
```diff
@@ -0,0 +1,66 @@
...
+    U="timestamp",
+    V="verbose",
+    W="pen",
+    X="xshift",
+    Y="yshift",
+    p="perspective",
+    t="transparency",
+)
+@kwargs_to_strings(R="sequence", T="sequence", c="sequence_comma", p="sequence")
+def ternary(self, table, **kwargs):
```
seisman, line 26: We should use `data`, not `table`, following the discussions in https://github.com/GenericMappingTools/pygmt/issues/1479.

willschlitzer: done in ce6de7c

pygmt/src/ternary.py
willschlitzer: I added the second line, but I think the first line doesn't make sense in the context (what would setting the width have to do with a valid linear plot?). I wanted to leave in my original line explaining how the width is set, as that is not clear from the GMT docstring.

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+from pygmt.clib import Session
+from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
+
+
+@fmt_docstring
+@use_alias(
+    B="frame",
+    C="cmap",
+    F="center",
```
seisman, line 12: I don't see the `-F` option in ternary.
```suggestion
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    S="style",
+    U="timestamp",
+    V="verbose",
+    W="pen",
+    X="xshift",
+    Y="yshift",
+    p="perspective",
+    t="transparency",
+)
+@kwargs_to_strings(R="sequence", T="sequence", c="sequence_comma", p="sequence")
```
seisman, line 25:
```suggestion
@kwargs_to_strings(R="sequence", c="sequence_comma", p="sequence")
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    F="center",
+    G="fill",
+    JX="width",
+    R="region",
+    S="style",
+    U="timestamp",
+    V="verbose",
+    W="pen",
+    X="xshift",
+    Y="yshift",
```
seisman, line 21:
```suggestion
    Y="yshift",
    c="panel",
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    p="perspective",
+    t="transparency",
+)
+@kwargs_to_strings(R="sequence", T="sequence", c="sequence_comma", p="sequence")
+def ternary(self, data, **kwargs):
+    r"""
+    Reads (*a*,\ *b*,\ *c*\ [,\ *z*]) records from *table* and plots symbols at
```
seisman, line 28:
```suggestion
    Reads (*a*,\ *b*,\ *c*\ [,\ *z*]) records from *data* and plots symbols at
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    those locations on a ternary diagram. If a symbol is selected and no symbol
+    size given, then we will interpret the fourth column of the input data as
+    symbol size. Symbols whose *size* is <= 0 are skipped. If no symbols are
+    specified then the symbol code (see **style** below) must be present as
```
seisman, line 32:
```suggestion
    specified then the symbol code (see ``style`` below) must be present as
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    last column in the input.  If **style** is not specified then we instead
```
seisman, line 33:
```suggestion
    last column in the input.  If ``style`` is not specified then we instead
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    Full option list at :gmt-docs:`ternary.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    data : str or list or {table-like}
+        Pass in either a file name to an ASCII data table, a Python list, a 2D
+        {table-classes}.
+    {J}
```
seisman, line 45:
```suggestion
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+    width : str
+        *width*. Set the width of the figure by passing a number, followed by
```
seisman, line 47:
```suggestion
        Set the width of the figure by passing a number, followed by
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,69 @@
...
+        counter-clockwise direction].
+    {CPT}
+    {G}
+    style : str
+        Plot symbols (including vectors, pie slices, fronts, decorated or
+        quoted lines).
+    {W}
+    {XY}
+    {U}
+    {V}
```
seisman, line 60:
```suggestion
    {V}
    {c}
```

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,23 @@
...
+
+@pytest.mark.mpl_image_compare
+def test_ternary():
+    """
+    Test plotting a ternary chart.
+    """
+    fig = Figure()
+    fig.ternary(
+        data="@ternary.txt",
+        region="0/100/0/100/0/100",
```
seisman, line 16:
```suggestion
        region=[0, 100, 0, 100, 0, 100],
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,70 @@
...
+        {table-classes}.
+    width : str
+        Set the width of the figure by passing a number, followed by
+        a unit (**i** for inches, **c** for centimeters). Use a negative width
+        to indicate that positive axes directions be clock-wise
+        [Default lets the a, b, c axes be positive in a
+        counter-clockwise direction].
+    {R}
+    {CPT}
+    {G}
```
seisman, line 53: `G` is aliased to `fill`, but the docstring here says `color`. Xref https://github.com/GenericMappingTools/pygmt/issues/1617

pygmt/src/ternary.py
```diff
@@ -0,0 +1,70 @@
...
+    style : str
+        Plot symbols (including vectors, pie slices, fronts, decorated or
+        quoted lines).
```
seisman, line 56: Need to expand the docstring of `style`, so that users understand what `style="c0.2c"` means.

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.mark.mpl_image_compare
+def test_ternary():
+    """
+    Test plotting a ternary chart.
+    """
+    fig = Figure()
+    fig.ternary(
+        data="@ternary.txt",
```
seisman, line 15: Need to cache this file using the new function `load_sample_data`.

pygmt/src/ternary.py
willschlitzer: Is there a standard as to what should be used? `pygmt.plot` has the same docstring for `style`.

pygmt/src/ternary.py
willschlitzer: I tried using `G=["fill", "color"]` but it didn't pass the tesing/documentation build. Am I missing a step?

pygmt/tests/test_ternary.py
willschlitzer: Does this need to be cached for `load_sample_data` or is it enough to cache it for the tests?

pygmt/src/ternary.py
meghanrjones: I think rather than adding a lot of text, we could add a link to the [symbols and markers gallery page](https://www.pygmt.org/dev/gallery/index.html#symbols-and-markers) similar to what is done for the region, frame, and projection docstring sections.

pygmt/src/ternary.py
seisman: > I tried using `G=["fill", "color"]` but it didn't pass the tesing/documentation build. Am I missing a step?

Need the rewrite the `use_alias` decorator to support the multi-alias mechanism like `G=["fill", "color"]`.

pygmt/tests/test_ternary.py
seisman: I think it depends on if you want to add an example for ternary. If the cache data is used in an example, I think it would be better to add it to the `load_sample_data` function. Of course, it can/should be done in a separate PR. So, caching it in CI is enough for this PR.

seisman: Just one more thought on the test: perhaps we should use a much simpler dataset (with only a few records) in a test, so that it's easier to verify if the test is correct or not. The current test can be a good example instead.

pygmt/src/ternary.py
willschlitzer: I think that falls outside of the scope of this pull request. I'll convert this to draft for the time being.

pygmt/tests/test_ternary.py
willschlitzer: I changed this to use a simple dataset

pygmt/src/ternary.py
```diff
@@ -0,0 +1,70 @@
...
+    {c}
+    {p}
+    {t}
+    """
+    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+    with Session() as lib:
+        file_context = lib.virtualfile_from_data(check_kind="vector", data=data)
+        with file_context as infile:
+            arg_str = " ".join([infile, build_arg_string(kwargs)])
+            lib.call_module("ternary", arg_str)
```
seisman, line 70:
```suggestion
            lib.call_module("ternary", build_arg_string(kwargs, infile=infile))
```

pygmt/tests/test_ternary.py
```diff
@@ -0,0 +1,65 @@
...
+        [0.59, 0.4, 0.01, 0.802],
+        [0.64, 0.24, 0.12, 2.423],
+        [0.0, 0.07, 0.93, 3.014],
+        [0.8, 0.07, 0.13, 2.79],
+        [0.04, 0.4, 0.56, 3.863],
+        [0.71, 0.11, 0.18, 1.095],
+        [0.22, 0.61, 0.17, 3.666],
+        [0.43, 0.39, 0.18, 1.887],
+    ]
+
```
seisman, line 46:
```suggestion
```

pygmt/src/ternary.py
```diff
@@ -0,0 +1,70 @@
...
+    data : str or list or {table-like}
+        Pass in either a file name to an ASCII data table, a Python list, a 2D
+        {table-classes}.
+    width : str
+        Set the width of the figure by passing a number, followed by
+        a unit (**i** for inches, **c** for centimeters). Use a negative width
+        to indicate that positive axes directions be clock-wise
+        [Default lets the a, b, c axes be positive in a
+        counter-clockwise direction].
+    {R}
```
seisman, line 51:
```suggestion
    region : str or list
        [*amin*, *amax*, *bmin*, *bmax*, *cmin*, *cmax*].
        Give the min and max limits for each of the three axis **a**, **b**, and **c**.
```

meghanrjones:
Hi @willschlitzer, I think this PR is good to merge after two small in-line suggestions (one for a line length issue to satisfy pylint and one to add a reference to plotting styles). Would you be open to moving forward with a merge rather than closing the PR?

willschlitzer:
@meghanrjones Not sure how I managed this (currently on a work computer), but didn't mean to close it! Thanks

pygmt/src/ternary.py
```diff
@@ -0,0 +1,71 @@
...
+    {V}
+    {c}
+    {p}
+    {t}
+    """
+    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+    with Session() as lib:
+        file_context = lib.virtualfile_from_data(check_kind="vector", data=data)
+        with file_context as infile:
+            lib.call_module("ternary", build_arg_string(kwargs, infile=infile))
```
seisman, line 71:
```suggestion
            lib.call_module(
                module="ternary", 
                args=build_arg_string(kwargs, infile=infile),
            )
```

pygmt/src/ternary.py
seisman: I think we can ignore it now and address the issue in https://github.com/GenericMappingTools/pygmt/issues/1617 instead.

pygmt/src/ternary.py
```diff
@@ -0,0 +1,71 @@
...
+        [Default lets the a, b, c axes be positive in a
+        counter-clockwise direction].
+    region : str or list
+        [*amin*, *amax*, *bmin*, *bmax*, *cmin*, *cmax*].
+        Give the min and max limits for each of the three axis **a**, **b**, and **c**.
+    {CPT}
+    {G}
+    style : str
+        Plot symbols (including vectors, pie slices, fronts, decorated or
+        quoted lines).
```
seisman, line 58:
```suggestion
    style : str
    	*symbol*\[\ *size*].
    	Plot individual symbols in a ternary diagram.
```

seisman:
@willschlitzer I think this PR is almost ready for merge. I'm wondering if you have some time to finalize this PR recently.

pygmt/src/ternary.py
```diff
@@ -0,0 +1,74 @@
...
+        {table-classes}.
+    width : str
+        Set the width of the figure by passing a number, followed by
+        a unit (**i** for inches, **c** for centimeters). Use a negative width
+        to indicate that positive axes directions be clock-wise
+        [Default lets the a, b, c axes be positive in a
+        counter-clockwise direction].
+    region : str or list
+        [*amin*, *amax*, *bmin*, *bmax*, *cmin*, *cmax*].
+        Give the min and max limits for each of the three axis **a**, **b**, and **c**.
```
seisman, line 53:
```suggestion
        Give the min and max limits for each of the three axes **a**, **b**, 
        and **c**.
```

seisman:
The tests fail, likely due to the outdated baseline image. Could you please regenerate the baseline image using GMT 6.4.0 again?

seisman (approved):

Issue #1399, opened 2021-07-29
Add a function wrapping GMT's grdhisteq module


meghanrjones:
**Description of the desired feature**

GMT's [grdhisteq](https://docs.generic-mapping-tools.org/dev/grdhisteq.html) would be a nice addition to PyGMT's grid processing capabilities. I am interested in submitting a PR wrapping pygmt.grdhisteq().

**Are you willing to help implement and maintain this feature?** Yes

PR #1433, opened 2021-08-10
Wrap grdhisteq

meghanrjones:

This PR will wrap the gmt module [`grdhisteq`](https://docs.generic-mapping-tools.org/6.2/grdhisteq.html).

**Preview** docs at https://pygmt-git-grdhisteq-gmt.vercel.app/api/generated/pygmt.grdhisteq.html

Fixes https://github.com/GenericMappingTools/pygmt/issues/1399

**Reminders**

**Slash Commands**


weiji14 (commented):
Gentle push on this. I'm starting a project using Sentinel 2 data so I imagine this function would come in handy at some point for doing histogram equalized plots :smile:

Looking at https://docs.generic-mapping-tools.org/6.2/grdhisteq.html closely, it seems that both grid outputs (`-G`) and table outputs (`-D`) are possible. Happy to only do `grid` outputs for now, but if you want to do `table` outputs too, the `triangulate` implementation at #731 can be a good resource to follow.

P.S. Recommend to update branch and resolve merge conflict before applying suggested changes below.

pygmt/tests/test_grdhisteq.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_grdhisteq_no_outgrid(grid):
+    """
+    Test the azimuth and direction parameters for grdhisteq with no set
+    outgrid.
+    """
+    temp_grid = grdhisteq(grid=grid)
+    assert temp_grid.dims == ("lat", "lon")
+    assert temp_grid.gmt.gtype == 1  # Geographic grid
+    assert temp_grid.gmt.registration == 1  # Pixel registration
```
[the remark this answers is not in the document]
meghanrjones: Exact tests are better in my opinion.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Parameters
+    ----------
+    grid : str or xarray.DataArray
+        The file name of the input grid or the grid loaded as a DataArray.
+    outgrid : str or None
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    divisions : int or str
+        The number of divisions of data range to make [Default is 16].
```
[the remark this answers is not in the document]
meghanrjones: 1. The gmt docs language is confusing here, but grdhisteq defaults to -C16 if -C is not provided. But if -C is provided, *n_cells* must also be appended (e.g., this fails `gmt grdhisteq @earth_relief_05m -R0/1/0/1 -C -D`).

2. I prefer aliases that are one word rather than split, so I prefer `divisions` or `cells`. I think `equalization` is less descriptive for this parameter. What is your preference?

meghanrjones:
> Gentle push on this. I'm starting a project using Sentinel 2 data so I imagine this function would come in handy at some point for doing histogram equalized plots 😄
> 

I'll try to wrap up the remaining tasks (i.e., polishing docstring, adding table output test) tomorrow.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,74 @@
...
+                lib.call_module("grdhisteq", arg_str)
+
+        if outgrid == tmpfile.name:  # if user did not set outgrid, return DataArray
+            with xr.open_dataarray(outgrid) as dataarray:
+                result = dataarray.load()
+                _ = result.gmt  # load GMTDataArray accessor information
+        else:
+            result = None  # if user sets an outgrid, return None
+
+        return result
```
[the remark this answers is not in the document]
meghanrjones: I added support for table output based on your triangulate implementation. I still need to add a test.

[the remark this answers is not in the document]
meghanrjones: Regarding this and your comment that `-h` is supported, we probably have a bug with the functions that include `result = pd.read_csv(tmpfile.name, sep="\t", header=None)` and `h="header"` since the user could specify that GMT should add a header to the output.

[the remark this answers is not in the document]
weiji14: Right, so we might need to do something like `header=None if "h" not in kwargs else kwargs["h"]` (or something more complicated). Also, I just noticed `grdhisteq -h` is not in GMT 6.2, only GMT 6.3, so should we include it here or not?

pygmt/src/grdhisteq.py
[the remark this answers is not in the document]
weiji14: >     1. The gmt docs language is confusing here, but grdhisteq defaults to -C16 if -C is not provided. But if -C is provided, _n_cells_ must also be appended (e.g., this fails `gmt grdhisteq @earth_relief_05m -R0/1/0/1 -C -D`).

Ah that's interesting, so only `int` then and no `bool`. Not sure about the `[Default is 16]` text, might get a bit confusing for users indeed.

>     2. I prefer aliases that are one word rather than split, so I prefer `divisions` or `cells`. I think `equalization` is less descriptive for this parameter. What is your preference?

To be honest, I was sort of leaning on `equalization` because `equalization=True` sounded nice. But since `bool` options (`-C`) might not work, I prefer `divisions`.

pygmt/src/grdhisteq.py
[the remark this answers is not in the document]
meghanrjones: If you type `gmt grdhisteq -?` in GMT 6.2, you will see that `-h` was supported. I have been going through making sure that the UNIX manpage-type docs are consistent with the html documentation, so there may be some options that look new based on the online reference but are just doc fixes.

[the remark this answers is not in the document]
meghanrjones: > Right, so we might need to do something like `header=None if "h" not in kwargs else kwargs["h"]` (or something more complicated)

We might need something more complicated since GMT can write multi-row headers depending on the modifiers. I think I'll just leave ``header`` off this PR and deal with it separately.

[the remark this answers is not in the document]
weiji14: Yep, the `-h` option will need to handled properly for other table functions too, so best to do that separately.

meghanrjones:
I changed the implementation a bit relative to https://github.com/GenericMappingTools/pygmt/pull/731 to support ASCII or pandas.DataFrame output for writing out the equalized histogram.

Still, the code is a bit clunky in order to support four different output types (pandas.DataFrame, xarray.DataArray, netCDF, or ASCII). What would you think about having two PyGMT functions for GMT's grdhisteq module rather than just one? One function could write out the data ranges of histogram equalization to a pd.DataFrame or ASCII table and the other could write out the cumulative distribution statistics to a netCDF file or xarray.DataArray. I guess coming up with the names for these would be harder than the current implementation, but I think it would be more user friendly long-term.

weiji14:
> I changed the implementation a bit relative to #731 to support ASCII or pandas.DataFrame output for writing out the equalized histogram.
> 
> Still, the code is a bit clunky in order to support four different output types (pandas.DataFrame, xarray.DataArray, netCDF, or ASCII). What would you think about having two PyGMT functions for GMT's grdhisteq module rather than just one? One function could write out the data ranges of histogram equalization to a pd.DataFrame or ASCII table and the other could write out the cumulative distribution statistics to a netCDF file or xarray.DataArray. I guess coming up with the names for these would be harder than the current implementation, but I think it would be more user friendly long-term.

Yeah I've debated a bit on whether to have 2 functions too, something like a `pygmt.grdhisteq.to_table()` and `pygmt.grdhisteq.to_grid()` (implemented using a Python class), or maybe with an underscore like `pygmt.grdhisteq_to_table()` and `pygmt.grdhisteq_to_grid()` (implemented purely using Python functions). Tying this to https://github.com/GenericMappingTools/pygmt/issues/1318#issuecomment-855317785, I think the split into 2 may have to happen eventually, especially if we want to support more table-like outputs (ascii/numpy/pandas/geopandas/etc) like what Will is doing at `grd2xyz`.

In terms of this PR, I almost want to focus on just the `grid` output part since that's what I'm interested in. But that would require a bit of work cherry-picking the 'table' output parts into a separate branch :upside_down_face:

meghanrjones:
> Yeah I've debated a bit on whether to have 2 functions too, something like a `pygmt.grdhisteq.to_table()` and `pygmt.grdhisteq.to_grid()` (implemented using Python classmethods), or maybe with an underscore like `pygmt.grdhisteq_to_table()` and `pygmt.grdhisteq_to_grid()` (implemented purely using Python functions). Tying this to [#1318 (comment)](https://github.com/GenericMappingTools/pygmt/issues/1318#issuecomment-855317785), I think the split into 2 may have to happen eventually, especially if we want to support more table-like outputs (ascii/numpy/pandas/geopandas/etc) like what Will is doing at `grd2xyz` #1284.
> 
> In terms of this PR, I almost want to focus on just the `grid` output part since that's what I'm interested in, and also because the bug at [GenericMappingTools/gmt#5785](https://github.com/GenericMappingTools/gmt/issues/5785) might not land until GMT 6.3. But that would require a bit of work cherry-picking the 'table' output parts from this PR into a separate branch 🙃

OK, I'll try out splitting the two tomorrow and will prioritize getting the grid output part set. If I fail we can always go back to this point 😆 I'll probably go for a function oriented implementation since that's more in-line with the rest of the API for processing.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,267 @@
...
+
+    :meth:`pygmt.grdhisteq.equalize_grid` provides a way to write a grid with
+    statistics based on a cumulative distribution function. In this
+    application, the ``outgrid`` has relative highs and lows in the same
+    (x,y) locations as the ``grid``, but the values are changed to reflect
+    their place in the cumulative distribution.
+    """
+
+    @staticmethod
+    def _grdhisteq(grid, **kwargs):
```
weiji14, line 51: Should we put the `@use_alias` decorator here instead? This would 1) prevent duplication, and 2) allow us to set default values on the methods (e.g. `equalize_histogram(grid, outgrid=True, ...)` and `compute_bins(grid, outfile=True, ...)`.

Downside is that users won't have tab-completion to see what parameter options are available :slightly_smiling_face:

weiji14 (commented):
Just a quick glance review. Implementation still needs some more polish, but I quite like the API style!

meghanrjones:
Thanks for the feedback @weiji14! I actually prefer the pure functional style rather than the mixed class structure. I think a path towards deciding involves adding a gallery example to this PR and the functional design at https://github.com/GenericMappingTools/pygmt/tree/grdhisteq-functions. The user-oriented comparison between the two options could help people make informed votes over at https://github.com/GenericMappingTools/pygmt/issues/1536.

After confirming whether to move forward with methods or functions, I will make more detailed responses your helpful comments above.

doc/api/index.rst
```diff
@@ -138,6 +138,9 @@ Operations on grids
     grdfill
     grdfilter
     grdgradient
+    grdhisteq
+    grdhisteq.equalize_grid
+    grdhisteq.compute_bins
```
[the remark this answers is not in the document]
meghanrjones: I'll modify this to keep the top-level class.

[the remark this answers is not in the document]
meghanrjones: I actually couldn't get autodoc to generate the pages for equalize_grid or compute_bins without including them in the toctree. I might look more into it later, but I also don't think it would be that big of deal to list all the class methods.

pygmt/src/grdhisteq.py
meghanrjones: I moved the `@use_alias` decorator to _grdhisteq. Rather than using `**kwargs`, I added positional arguments to the public methods so that tab completion still works. The downside is that single-character arguments do not work for these methods. I think this is fine since the only option that isn't included is **h**, which I would say is a bit unstable in PyGMT anyways. I would prefer to require long-form arguments where possible from the start rather than having more deprecations later on (xref https://github.com/GenericMappingTools/pygmt/issues/262).

meghanrjones:
FYI in case it is helpful for reviewing the implementation, I created a grdhisteq tutorial over at https://github.com/GenericMappingTools/pygmt/blob/grdhisteq-tutorial/examples/tutorials/advanced/grid_equalization.py. I am not including it as part of this PR to keep the changes as small as possible.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,267 @@
...
+            - None if ``outgrid`` is a str (grid output is stored in
+              ``outgrid``)
+
+        See Also
+        -------
+        :meth:`pygmt.grd2cpt`
+        """
+        # Return a xarray.DataArray if ``outgrid`` is not set
+        if "G" not in kwargs:
+            kwargs.update({"G": True})
```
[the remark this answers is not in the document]
meghanrjones: I think this case is avoided with the latest revision, unless someone decides to directly use the internal function `_grdhisteq`.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,341 @@
...
+                arg_str = " ".join([infile, build_arg_string(kwargs)])
+                lib.call_module("grdhisteq", arg_str)
+
+        if output_type == "file":
+            return None
+        if output_type == "xarray":
+            return load_dataarray(tmpfile.name)
+
+        result = pd.read_csv(
+            kwargs["D"], sep="\t", header=None, names=["start", "stop", "bin_id"]
```
weiji14, line 141: Reduce memory footprint of output `pandas.DataFrame`. I suppose uint8/uint32 should work for most cases?

```suggestion
            filepath_or_buffer=kwargs["D"],
            sep="\t",
            header=None,
            names=["start", "stop", "bin_id"],
            dtype={
                "start": np.float32,
                "stop": np.float32,
                "bin_id": np.uint8 if kwargs["C"] < 256 else np.uint32,
            },
```

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,341 @@
+"""
+grdhisteq - Perform histogram equalization for a grid.
+"""
+import warnings
+
+import pandas as pd
```
weiji14, line 6:
```suggestion
import numpy as np
import pandas as pd
```

pygmt/tests/test_grdhisteq.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    return pd.DataFrame(
+        data=np.array(
+            [
+                [345.5, 519.5, 0],
+                [519.5, 726.5, 1],
+            ]
+        ),
+        columns=["start", "stop", "bin_id"],
+    ).astype({"start": np.float64, "stop": np.float64, "bin_id": np.int64})
```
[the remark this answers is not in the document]
weiji14:
```suggestion
    ).astype({"start": np.float32, "stop": np.float32, "bin_id": np.uint8})
```

pygmt/tests/test_grdhisteq.py
```diff
@@ -0,0 +1,103 @@
...
+    Test grdhisteq.compute_bins with ``outfile``.
+    """
+    with GMTTempFile(suffix=".txt") as tmpfile:
+        result = grdhisteq.compute_bins(
+            grid=grid, divisions=2, region=region, outfile=tmpfile.name
+        )
+        assert result is None  # return value is None
+        assert os.path.exists(path=tmpfile.name)
+        temp_df = pd.read_csv(
+            tmpfile.name, sep="\t", header=None, names=["start", "stop", "bin_id"]
```
weiji14, line 101:
```suggestion
            filepath_or_buffer=tmpfile.name,
            sep="\t",
            header=None,
            names=["start", "stop", "bin_id"],
            dtype={"start": np.float32, "stop": np.float32, "bin_id": np.uint8},
```

pygmt/src/grdhisteq.py
meghanrjones: This would require more logic because `kwargs["C"]` can be `None`. What do you think about just having np.uint32 for "bin_id"? This is most compatible with its representation in GMT and also simpler than having it depend on the number of divisions.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,366 @@
...
+        Example
+        -------
+        >>> import pygmt  # doctest: +SKIP
+        >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to
+        >>> # 30, and a y-range of 15 to 25
+        >>> grid = pygmt.datasets.load_earth_relief(
+        ...     resolution="30m", region=[10, 30, 15, 25]
+        ... )  # doctest: +SKIP
+        >>> # Find the elevation intervals that divide the grid into 5
+        >>> # divisions of equal area
```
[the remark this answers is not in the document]
meghanrjones: It actually is referring to spatial area for this example, since it is geographic. From the docs "For geographical grids we do a weighted histogram equalization since the area of each node varies with latitude."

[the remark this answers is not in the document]
meghanrjones: I'll add this note to our docstring as well.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,366 @@
...
+
+        Example
+        -------
+        >>> import pygmt  # doctest: +SKIP
+        >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to
+        >>> # 30, and a y-range of 15 to 25
+        >>> grid = pygmt.datasets.load_earth_relief(
+        ...     resolution="30m", region=[10, 30, 15, 25]
+        ... )  # doctest: +SKIP
+        >>> # Find the elevation intervals that divide the grid into 5
```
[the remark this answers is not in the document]
meghanrjones: I removed 'the' but I think that it should stay 'divide' since 'intervals' is plural.

pygmt/src/grdhisteq.py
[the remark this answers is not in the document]
weiji14: Right, I got a little confused by the sentence as the divisions are 1D elevation divisions (e.g. 0, 100m, 200m, 250m, 275m, etc), but the way those divisions are made is by dividing the 2D grid into 5 equal areas. Maybe we can reword the sentence a little bit?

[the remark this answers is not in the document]
meghanrjones: I revised this in https://github.com/GenericMappingTools/pygmt/pull/1433/commits/32757d88e916e2a1087d69e956481f29727ac3ac. Let me know if it's still not clear (it's a bit difficult to phrase right).

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,358 @@
...
+        >>> # 30, and a y-range of 15 to 25
+        >>> grid = pygmt.datasets.load_earth_relief(
+        ...     resolution="30m", region=[10, 30, 15, 25]
+        ... )  # doctest: +SKIP
+        >>> # Find elevation intervals that divide the data range into
+        >>> # divisions which have an equal area in the original grid.
+        >>> bins = pygmt.grdhisteq.compute_bins(
+        ...     grid=grid, divisions=5
+        ... )  # doctest: +SKIP
+
```
weiji14, line 322: Just to illustrate what `compute_bins` is actually doing.
```suggestion
        >>> print(bins)  # doctest: +SKIP
              start      stop  bin_id
        0 -3.023719 -0.846985       0
        1 -0.846985 -0.257874       1
        2 -0.257874  0.248180       2
        3  0.248180  0.838059       3
        4  0.838059  3.023719       4

```

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,358 @@
...
+        :meth:`pygmt.grd2cpt`
+
+        Notes
+        -----
+        This method does a weighted histogram equalization for geographic
+        grids to account for node area varying with latitude.
+        """
+        # Return a pandas.DataFrame if ``outfile`` is not set
+        if output_type not in ["numpy", "pandas", "file"]:
+            raise GMTInvalidInput(
```
weiji14, line 334: Need to add a failing test.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,358 @@
...
+            header=None,
+            names=["start", "stop", "bin_id"],
+            dtype={
+                "start": np.float32,
+                "stop": np.float32,
+                "bin_id": np.uint32,
+            },
+        )
+        if output_type == "numpy":
+            result = result.to_numpy()
```
weiji14, line 153: Need to add a test for `compute_bins(..., output_type="numpy")`. Can probably turn `test_compute_bins_no_outfile` into a parametrized test (or just make a new one).

pygmt/src/grdhisteq.py
weiji14: That's a really good idea actually. Go for it!

meghanrjones:
> Nice one. Looking forward to using this in my project soon! I'm getting sick of arbitrarily stretching optical satellite images...

Thanks for all your helpful feedback!

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,367 @@
...
+    @staticmethod
+    @fmt_docstring
+    @use_alias(
+        C="divisions",
+        D="outfile",
+        G="outgrid",
+        R="region",
+        N="gaussian",
+        Q="quadratic",
+        V="verbose",
```
seisman, line 60:
```suggestion
        V="verbose",
        h="header",
```

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,367 @@
...
+            The name of the output netCDF file with extension .nc to store the
+            grid in.
+        outfile : str or bool or None
+            The name of the output ASCII file to store the results of the
+            histogram equalization in. Not allowed if ``outgrid`` is used.
+        divisions : int
+            Set the number of divisions of the data range [Default is 16].
+
+        {R}
+        {V}
```
seisman, line 111:
```suggestion
        {V}
        {h}
```

pygmt/src/grdhisteq.py
meghanrjones: ``header`` won't be accessible from ``compute_bins`` yet but could be added in a separate PR when ``output_type="file"``.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,369 @@
...
+        set]. The :class:`pandas.DataFrame` or ASCII file output can be used to
+        make a colormap with :meth:`pygmt.makecpt` and an image with
+        :meth:`pygmt.Figure.grdimage` that has all levels of gray occuring
+        equally.
+
+        :meth:`pygmt.grdhisteq` also provides a way to write a grid with
+        statistics based on a cumulative distribution function. In this
+        application, the ``outgrid`` has relative highs and lows in the same
+        (x,y) locations as the ``grid``, but the values are changed to reflect
+        their place in the cumulative distribution.
```
seisman, line 90: Lines 68-90 are the same as Lines 24-48. Should we remove these lines from `_grdhisteq`.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,369 @@
...
+    @fmt_docstring
+    def compute_bins(
+        grid,
+        *,
+        output_type="pandas",
+        outfile=None,
+        divisions=None,
+        quadratic=None,
+        verbose=None,
+        region=None,
```
seisman, line 261:
```suggestion
        region=None,
        header=None,
```

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,369 @@
...
+
+                - ``numpy`` - :class:`numpy.ndarray`
+                - ``pandas``- :class:`pandas.DataFrame`
+                - ``file`` - ASCII file (requires ``outfile``)
+        divisions : int
+            Set the number of divisions of the data range.
+        quadratic : bool
+            Perform quadratic equalization [Default is linear].
+        {R}
+        {V}
```
seisman, line 301:
```suggestion
        {V}
        {h}
```

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,369 @@
...
+            )
+
+        if isinstance(outfile, str) and output_type != "file":
+            msg = (
+                f"Changing 'output_type' from '{output_type}' to 'file' "
+                "since 'outfile' parameter is set. Please use output_type='file' "
+                "to silence this warning."
+            )
+            warnings.warn(message=msg, category=RuntimeWarning, stacklevel=2)
+            output_type = "file"
```
seisman, line 356: Perhaps we should raise an exception here rather than give a warning? Sometimes people just ignore warnings.

pygmt/src/grdhisteq.py
```diff
@@ -0,0 +1,369 @@
...
+                outfile = tmpfile.name
+            return grdhisteq._grdhisteq(
+                grid,
+                output_type=output_type,
+                tmpfile=tmpfile,
+                outfile=outfile,
+                divisions=divisions,
+                quadratic=quadratic,
+                verbose=verbose,
+                region=region,
```
seisman, line 368:
```suggestion
                region=region,
                header=header,
```

pygmt/src/grdhisteq.py
meghanrjones: There was a bit of a discussion about this following this comment https://github.com/GenericMappingTools/pygmt/pull/1284#pullrequestreview-737690502. See https://github.com/GenericMappingTools/pygmt/pull/1284#discussion_r698128719 for a specific discussion about raising a warning versus exception. I think we should follow that lead, although we could move this to a helper function to make it more clear that the logic is used for multiple modules.

pygmt/src/grdhisteq.py
meghanrjones: I removed these lines.

seisman (approved):

PR #1377, opened 2021-07-07
Wrap grdproject

willschlitzer:
This pull requests wraps the GMT module [`grdproject`](https://docs.generic-mapping-tools.org/6.2/grdproject.html).

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1439, opened 2021-08-11
Add pygmt.load_dataarray function and refactor modules that return None or xarray.dataarray

### LINKED ISSUE #1400: Reduce repeated code for managing the outgrid parameter

The block of code below is currently used seven times throughout PyGMT modules (probably more pending Will's PRs). Based on https://github.com/GenericMappingTools/gmt/issues/5563#issuecomment-889519995, it may need to be used up to 34 times if we continue adding functions 1:1 with GMT modules. I think it would be useful to add an `build_outgrid_args()` method (or something similar) that can be used throughout to build the argument string using either outgrid or tempfile.name. I expect this would be simplest to implement as a new method in the Session class, although it could be a function in utils.

https://github.com/GenericMappingTools/pygmt/blob/6a4b8715bf00cd42cae8d5405e35e08a4764e97c/pygmt/src/grdclip.py#L85-L88

Edit: The original post only mentioned managing the input arguments to call_module, but the same can be applied to the output for the current default behavior of returning a DataArray if outgrid is not set:

https://github.com/GenericMappingTools/pygmt/blob/6a4b8715bf00cd42cae8d5405e35e08a4764e97c/pygmt/src/grdclip.py#L91-L96

meghanrjones:

This PR adds a new pygmt/io.py module that holds process_output_grid, which returns an xarray.DataArray or None depending on whether outgrid matches tmpfile.name.

It replaces duplicate code with the new function.

Fixes #1400 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -9,8 +9,8 @@ def load_dataarray(filename_or_obj, **kwargs):
     Open, load into memory, and close a DataArray from a file or file-like
     object containing a single data variable.
 
-    This is a thin wrapper around :py:meth:`xarray.open_dataarray`. It differs
-    from :py:meth:`xarray.open_dataarray` in that it loads the DataArray into
+    This is a thin wrapper around :py:func:`xarray.open_dataarray`. It differs
+    from :py:func:`xarray.open_dataarray` in that it loads the DataArray into
     memory, gets GMT specific metadata about the grid via
     :py:meth:`GMTDataArrayAccessor`, closes the file, and returns the
     DataArray. In contrast, :py:func:`xarray.open_dataarray` keeps the file
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -4,33 +4,41 @@ PyGMT input/output (I/O) utilities.
 import xarray as xr
 
 
-def process_output_grid(grid_name, tmpfile_name):
+def load_datagrid(filename_or_obj, **kwargs):
     """
-    Processes the output from the GMT API to return an xarray.DataArray if
-    ``grid_name`` matches ``tmpfile_name`` and return None if it does not.
+    Open, load into memory, and close a DataArray from a file or file-like
+    object containing a single data variable.
+
+    This is a thin wrapper around :py:meth:`xarray.open_dataarray`. It differs
+    from `open_dataarray` in that it loads the DataArray into memory, gets GMT
+    specific metadata about the grid via :meth:`GMTDataArrayAccessor`, closes
+    the file, and returns the DataArray. In contrast, `open_dataarray` keeps
+    the file handle open and lazy loads its contents. All parameters are passed
+    directly to `open_dataarray`. See that documentation for further details.
 
     Parameters
     ----------
-    grid_name : str
-        The name of the output netCDF file with extension .nc to store the grid
-        in.
-    tmpfile_name : str
-        The name attribute from a GMTTempFile instance.
+    filename_or_obj : str or Path or file-like or DataStore
+        Strings and Path objects are interpreted as a path to a netCDF file
+        or an OpenDAP URL and opened with python-netCDF4, unless the filename
+        ends with .gz, in which case the file is gunzipped and opened with
+        scipy.io.netcdf (only netCDF3 supported). Byte-strings or file-like
+        objects are opened by scipy.io.netcdf (netCDF3) or h5py (netCDF4/HDF).
 
     Returns
     -------
-    ret: xarray.DataArray or None
-        Return type depends on whether the ``outgrid`` parameter is set:
+    datarray : xarray.DataArray
+        The newly created DataArray.
 
-        - :class:`xarray.DataArray` if ``outgrid`` is not set
-        - None if ``outgrid`` is set (grid output will be stored in file set by
-          ``outgrid``)
+    See Also
+    --------
+    xarray.open_dataarray
     """
-    if grid_name == tmpfile_name:  # Implies user did not set outgrid, return DataArray
-        with xr.open_dataarray(grid_name) as dataarray:
-            result = dataarray.load()
-            _ = result.gmt  # load GMTDataArray accessor information
-    else:
-        result = None  # Implies user set an outgrid, return None
+    if "cache" in kwargs:
+        raise TypeError("cache has no effect in this context")
+
+    with xr.open_dataarray(grid_name, **kwargs) as dataarray:
+        result = dataarray.load()
+        _ = result.gmt  # load GMTDataArray accessor information
 
     return result
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -1,5 +1,5 @@
 """
-PyGMT i/o utilities.
+PyGMT input/output (I/O) utilities.
 """
 import xarray as xr
 
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -13,14 +13,14 @@ def load_dataarray(filename_or_obj, **kwargs):
     from :py:meth:`xarray.open_dataarray` in that it loads the DataArray into
     memory, gets GMT specific metadata about the grid via
     :py:meth:`GMTDataArrayAccessor`, closes the file, and returns the
-    DataArray. In contrast, :py:meth:`xarray.open_dataarray` keeps the file
+    DataArray. In contrast, :py:func:`xarray.open_dataarray` keeps the file
     handle open and lazy loads its contents. All parameters are passed directly
-    to :py:meth:`xarray.open_dataarray`. See that documentation for further
+    to :py:func:`xarray.open_dataarray`. See that documentation for further
     details.
 
     Parameters
     ----------
-    filename_or_obj : str or Path or file-like or DataStore
+    filename_or_obj : str or pathlib.Path or file-like or DataStore
         Strings and Path objects are interpreted as a path to a netCDF file
         or an OpenDAP URL and opened with python-netCDF4, unless the filename
         ends with .gz, in which case the file is gunzipped and opened with
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -10,11 +10,13 @@ def load_dataarray(filename_or_obj, **kwargs):
     object containing a single data variable.
 
     This is a thin wrapper around :py:meth:`xarray.open_dataarray`. It differs
-    from `open_dataarray` in that it loads the DataArray into memory, gets GMT
-    specific metadata about the grid via :meth:`GMTDataArrayAccessor`, closes
-    the file, and returns the DataArray. In contrast, `open_dataarray` keeps
-    the file handle open and lazy loads its contents. All parameters are passed
-    directly to `open_dataarray`. See that documentation for further details.
+    from :py:meth:`xarray.open_dataarray` in that it loads the DataArray into
+    memory, gets GMT specific metadata about the grid via
+    :py:meth:`GMTDataArrayAccessor`, closes the file, and returns the
+    DataArray. In contrast, :py:meth:`xarray.open_dataarray` keeps the file
+    handle open and lazy loads its contents. All parameters are passed directly
+    to :py:meth:`xarray.open_dataarray`. See that documentation for further
+    details.
 
     Parameters
     ----------
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -1,0 +1,36 @@
+"""
+PyGMT i/o utilities.
+"""
+import xarray as xr
+
+
+def process_output_grid(grid_name, tmpfile_name):
+    """
+    Processes the output from the GMT API to return an xarray.DataArray if
+    ``grid_name`` matches ``tmpfile_name`` and return None if it does not.
+
+    Parameters
+    ----------
+    grid_name : str
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    tmpfile_name : str
+        The name attribute from a GMTTempFile instance.
+
+    Returns
+    -------
+    ret: xarray.DataArray or None
+        Return type depends on whether the ``outgrid`` parameter is set:
+
+        - :class:`xarray.DataArray` if ``outgrid`` is not set
+        - None if ``outgrid`` is set (grid output will be stored in file set by
+          ``outgrid``)
+    """
+    if grid_name == tmpfile_name:  # Implies user did not set outgrid, return DataArray
+        with xr.open_dataarray(grid_name) as dataarray:
+            result = dataarray.load()
+            _ = result.gmt  # load GMTDataArray accessor information
+    else:
+        result = None  # Implies user set an outgrid, return None
+
+    return result
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -37,7 +37,7 @@ def load_datagrid(filename_or_obj, **kwargs):
     if "cache" in kwargs:
         raise TypeError("cache has no effect in this context")
 
-    with xr.open_dataarray(grid_name, **kwargs) as dataarray:
+    with xr.open_dataarray(filename_or_obj, **kwargs) as dataarray:
         result = dataarray.load()
         _ = result.gmt  # load GMTDataArray accessor information
 
--- a/pygmt/io.py
+++ b/pygmt/io.py
@@ -4,7 +4,7 @@ PyGMT input/output (I/O) utilities.
 import xarray as xr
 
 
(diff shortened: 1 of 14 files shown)


willschlitzer (approved):
Looks great!

michaelgrund (approved):

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
+"""
+PyGMT i/o utilities.
+"""
+import xarray as xr
+
+
+def process_output_grid(grid_name, tmpfile_name):
```
weiji14, line 7: Do we want this to be a public facing function? If this is meant for internal use only, I suggest renaming the function to `_process_output_grid` as per https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables.

However, if we do want this to be public-facing (i.e. users can do `pygmt.io.process_output_grid`), I would suggest another name like `open_datagrid` (or similar), and a completely different API implementation that follows in the footsteps of [`rioxarray.open_rasterio`](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.open_rasterio).

There is also the option of creating a custom pygmt 'engine' in `xarray`, see e.g. https://github.com/corteva/rioxarray/pull/281, but I'm getting ahead of myself here. I'm just thinking more long-term xarray integration plans (xref #499).

pygmt/src/surface.py
```diff
@@ -99,11 +99,4 @@ def surface(x=None, y=None, z=None, data=None, **kwargs):
...
 
-        if outfile == tmpfile.name:  # if user did not set outfile, return DataArray
-            with xr.open_dataarray(outfile) as dataarray:
-                result = dataarray.load()
-                _ = result.gmt  # load GMTDataArray accessor information
-        elif outfile != tmpfile.name:  # if user sets an outfile, return None
-            result = None
-
-    return result
+        return process_output_grid(outfile, tmpfile.name)
```
weiji14, line 102: Just a note that we should probably standardize on using `outgrid` instead of `outfile` here for `pygmt.surface`. Will need to go through some deprecation cycles :slightly_smiling_face:

pygmt/io.py
meghanrjones: While I think a open_grid() function or pygmt engine in xarray would be excellent longterm (analogous to GMT.jl's [gdalread_function](https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.gdalread)), this function is not for that purpose. So I will rename appropriately.

pygmt/src/surface.py
meghanrjones: Completely agree :smile: Want to open an issue for this?

pygmt/io.py
seisman: Ideally, we should avoid writing the grid to a temporary file. Instead, we should define a ["GMT_GRID" class](https://docs.generic-mapping-tools.org/latest/api.html#gmt-grids), output the grid into a GMT_GRID object, but we may also need a function to convert a GMT_GRID object to xarray.DataArray.

I tried it before but didn't succeed due to my limited knowledge about ctypes.

pygmt/src/surface.py
weiji14: Ok, done at #1441.

pygmt/io.py
meghanrjones: Yes, I can see this being a better option. I do not have any time to look into a more robust refactoring now, so unless someone else has free time soon I suggest we still merge this PR and open a new issue for your idea.

weiji14: Agree with merging this PR (after renaming the function to have a leading underscore). The `GMT_GRID` class can wait, from experience it's probably not a trivial thing to add (xref https://github.com/GenericMappingTools/pygmt/pull/520#discussion_r454096421) though probably something we need to think about longer term.

meghanrjones: Some clarification would be helpful- The PEP8 link is for methods and instances but this is a function. The section for functions does not have similar guidance and most analogous functions in utils.py do not use a leading underscore. Should I still add one here?

weiji14: Sorry, a better link is probably https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces. To quote (emphasis mine):

> ..., internal interfaces (packages, modules, classes, **functions**, attributes or other names) should still be **prefixed with a single leading underscore**.

So you can just rename it like so:

```suggestion
def _process_output_grid(grid_name, tmpfile_name):
```
and make sure to apply it to the other lines in this PR.

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
...
+        The name attribute from a GMTTempFile instance.
+
+    Returns
+    -------
+    ret: xarray.DataArray or None
+        Return type depends on whether the ``outgrid`` parameter is set:
+
+        - :class:`xarray.DataArray` if ``outgrid`` is not set
+        - None if ``outgrid`` is set (grid output will be stored in file set by
+          ``outgrid``)
```
seisman, line 27: This paragraph doesn't make sense in this function.

pygmt/io.py
seisman: As the variable names `grid_name` and `tempfile_name` make little sense in this function, I'm wondering if we can rewrite the function to:
```
def process_output_grid(bool_flag)
```
and call it this way `process_output_grid(outgrid == tmpfile.name)` in other functions.

weiji14: Do you mean something like:

```python
def process_output_grid(outgrid, bool_flag):
```

and used like `return process_output_grid(outgrid, bool_flag=outgrid == tmpfile.name)`?

To be honest, I did think of using a one-liner like `return load_datagrid(outgrid) if outgrid == tmpfile.name else None`, where the `load_datagrid` function would just have the `with xr.open_dataarray()` code-block. The `load_datagrid` function could then be public-facing, and is essentially similar to [`xarray.load_dataarray`](http://xarray.pydata.org/en/v0.19.0/generated/xarray.load_dataarray.html) but with an extra `_ = result.gmt  # load GMTDataArray accessor information` line.

seisman: I like the `return load_datagrid(outgrid) if outgrid == tmpfile.name else None` code. Actually I feel it's a little weird to check `if outgrid == tmpfile.name` in the `process_output_grid` function, which is not easy to understand for new wrapper contributors.

weiji14: Let's see what Meghan thinks when she comes back tomorrow. If we do decide to go with `load_datagrid` though, the code here can be replaced too :grin::

https://github.com/GenericMappingTools/pygmt/blob/ebbb0cc0045454f05db5abdfb3bcb5e1f18e96a7/pygmt/datasets/earth_relief.py#L136-L138

meghanrjones: Sounds good, I'll try out your implementation idea. Going to convert to draft for now and come back to this after more workshop prep work.

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
+"""
+PyGMT input/output (I/O) utilities.
+"""
+import xarray as xr
+
+
+def process_output_grid(grid_name, tmpfile_name):
```
weiji14, line 7: Basing this on [`xarray.load_dataarray`](https://xarray.pydata.org/en/v0.19.0/generated/xarray.load_dataarray.html) at https://github.com/pydata/xarray/blob/a21dead8621480b608827355b50e85af6f88f185/xarray/backends/api.py#L247

```suggestion
def load_datagrid(filename_or_obj, **kwargs):
```

pygmt/src/grdclip.py
```diff
@@ -88,11 +88,4 @@ def grdclip(grid, **kwargs):
...
 
-        if outgrid == tmpfile.name:  # if user did not set outgrid, return DataArray
-            with xr.open_dataarray(outgrid) as dataarray:
-                result = dataarray.load()
-                _ = result.gmt  # load GMTDataArray accessor information
-        else:
-            result = None  # if user sets an outgrid, return None
-
-        return result
+        return process_output_grid(outgrid, tmpfile.name)
```
weiji14, line 91: Remember to apply this change elsewhere too!
```suggestion
        return load_datagrid(outgrid) if outgrid == tmpfile.name else None
```

pygmt/src/grdclip.py
```diff
@@ -11,6 +10,7 @@
     kwargs_to_strings,
     use_alias,
 )
+from pygmt.io import process_output_grid
```
weiji14, line 13: Remember to apply this change elsewhere too!
```suggestion
from pygmt.io import load_datagrid
```

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    Processes the output from the GMT API to return an xarray.DataArray if
+    ``grid_name`` matches ``tmpfile_name`` and return None if it does not.
```
weiji14, line 10: Based on [`xarray.load_dataarray`](https://xarray.pydata.org/en/v0.19.0/generated/xarray.load_dataarray.html) at https://github.com/pydata/xarray/blob/a21dead8621480b608827355b50e85af6f88f185/xarray/backends/api.py#L248-L255, but modified to mention the `GMTDataArrayAccessor`.
```suggestion
    Open, load into memory, and close a DataArray from a file or file-like
    object containing a single data variable.

    This is a thin wrapper around :py:meth:`xarray.open_dataarray`. It differs
    from `open_dataarray` in that it loads the DataArray into memory, gets GMT
    specific metadata about the grid via :meth:`GMTDataArrayAccessor`, closes
    the file, and returns the DataArray. In contrast, `open_dataarray` keeps
    the file handle open and lazy loads its contents. All parameters are passed
    directly to `open_dataarray`. See that documentation for further details.
```

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Parameters
+    ----------
+    grid_name : str
+        The name of the output netCDF file with extension .nc to store the grid
+        in.
+    tmpfile_name : str
+        The name attribute from a GMTTempFile instance.
```
weiji14, line 18: Copied directly from the `filename_or_obj` parameter docs in [`xarray.open_dataarray`](https://xarray.pydata.org/en/v0.19.0/generated/xarray.open_dataarray.html#xarray.open_dataarray) at https://github.com/pydata/xarray/blob/a21dead8621480b608827355b50e85af6f88f185/xarray/backends/api.py#L542-L547

```suggestion
    filename_or_obj : str or Path or file-like or DataStore
        Strings and Path objects are interpreted as a path to a netCDF file
        or an OpenDAP URL and opened with python-netCDF4, unless the filename
        ends with .gz, in which case the file is gunzipped and opened with
        scipy.io.netcdf (only netCDF3 supported). Byte-strings or file-like
        objects are opened by scipy.io.netcdf (netCDF3) or h5py (netCDF4/HDF).
```

pygmt/io.py
weiji14, line 27: Suggest changing this to match https://github.com/pydata/xarray/blob/a21dead8621480b608827355b50e85af6f88f185/xarray/backends/api.py#L259-L264

```suggestion
    datarray : xarray.DataArray
        The newly created DataArray.

    See Also
    --------
    xarray.open_dataarray
```

pygmt/io.py
```diff
@@ -0,0 +1,36 @@
...
+        - :class:`xarray.DataArray` if ``outgrid`` is not set
+        - None if ``outgrid`` is set (grid output will be stored in file set by
+          ``outgrid``)
+    """
+    if grid_name == tmpfile_name:  # Implies user did not set outgrid, return DataArray
+        with xr.open_dataarray(grid_name) as dataarray:
+            result = dataarray.load()
+            _ = result.gmt  # load GMTDataArray accessor information
+    else:
+        result = None  # Implies user set an outgrid, return None
```
weiji14, line 34:
```suggestion
    with xr.open_dataarray(grid_name) as dataarray:
        result = dataarray.load()
        _ = result.gmt  # load GMTDataArray accessor information
```

weiji14 (commented):
Hi @meghanrjones, hoping that you can spend some time working on this PR soon (since there's a lot of open PRs that could benefit from a standardized function)! Based on the discussion above at https://github.com/GenericMappingTools/pygmt/pull/1439#discussion_r687374987, here are some suggested changes. You'll also need to:

1. Update `doc/api/index.rst` to include the new `load_datagrid` function. My thought was to place it under 'Data Processing', but we could start a dedicated Input/output section in the API docs if you prefer, similar to https://geopandas.org/docs/reference/io.html. You'll also need to add `from pygmt.io import load_datagrid` to `pygmt/__init__.py`.
```
Input/Output:

.. autosummary::
    :toctree: generated

    load_datagrid
```

2. Find and replace `process_output_grid` to `load_datagrid` across all the `grd*` functions. Remember to also include `grdproject` added recently in #1377 and `load_earth_relief` at https://github.com/GenericMappingTools/pygmt/blob/ebbb0cc0045454f05db5abdfb3bcb5e1f18e96a7/pygmt/datasets/earth_relief.py#L136-L138.

Let me know if you need any help. I'm happy to push some changes directly to this branch too if you're ok with working on this together.

meghanrjones:
Thanks for the ping @weiji14. I will prioritize this PR tomorrow (9/1) or am fine with you pushing changes if you post a message first.

meghanrjones:
@weiji14, thanks for your detailed comment. I named the function `load_dataarray` rather than `load_datagrid` to be more specific about the output type. Of course, you can comment if you disagree with this decision.

seisman (approved):
Looks a clean solution to me!

pygmt/io.py
```diff
@@ -0,0 +1,46 @@
...
+import xarray as xr
+
+
+def load_dataarray(filename_or_obj, **kwargs):
+    """
+    Open, load into memory, and close a DataArray from a file or file-like
+    object containing a single data variable.
+
+    This is a thin wrapper around :py:meth:`xarray.open_dataarray`. It differs
+    from :py:meth:`xarray.open_dataarray` in that it loads the DataArray into
```
weiji14, line 13: Need to use `:py:func:` instead to link to https://xarray.pydata.org/en/stable/generated/xarray.open_dataarray.html#xarray.open_dataarray properly.
```suggestion
    This is a thin wrapper around :py:func:`xarray.open_dataarray`. It differs
    from :py:func:`xarray.open_dataarray` in that it loads the DataArray into
```

pygmt/io.py
```diff
@@ -0,0 +1,46 @@
...
+    memory, gets GMT specific metadata about the grid via
+    :py:meth:`GMTDataArrayAccessor`, closes the file, and returns the
+    DataArray. In contrast, :py:meth:`xarray.open_dataarray` keeps the file
+    handle open and lazy loads its contents. All parameters are passed directly
+    to :py:meth:`xarray.open_dataarray`. See that documentation for further
```
weiji14, line 18:
```suggestion
    DataArray. In contrast, :py:func:`xarray.open_dataarray` keeps the file
    handle open and lazy loads its contents. All parameters are passed directly
    to :py:func:`xarray.open_dataarray`. See that documentation for further
```

pygmt/io.py
```diff
@@ -0,0 +1,46 @@
...
+    details.
+
+    Parameters
+    ----------
+    filename_or_obj : str or Path or file-like or DataStore
```
weiji14, line 23: I'm not quite sure how to link the `file-like` to https://docs.python.org/3/glossary.html#term-file-like-object. Looking at xarray, they do it at https://github.com/pydata/xarray/blob/f915515d610b4471888fa44dfb00dbae3fd22349/doc/conf.py#L130, so maybe we could add a `napoleon_type_aliases` dictionary to `doc/conf.py`? But we can leave it to a separate PR and just keep it at `file-like` for now.

```suggestion
    filename_or_obj : str or pathlib.Path or file-like or DataStore
```

weiji14 (approved):
> @weiji14, thanks for your detailed comment. I named the function `load_dataarray` rather than `load_datagrid` to be more specific about the output type. Of course, you can comment if you disagree with this decision.

Ok with renaming this to `load_dataarray`. It's not a commonly used function I think, so hopefully the `xarray` folks don't mind :smile_cat: Other than that, only a few more minor suggestions on the docstrings.

meghanrjones:
I would prefer to wait to merge this until after https://github.com/GenericMappingTools/pygmt/pull/1458 so that I can resolve the merge conflict rather than causing one for that PR.

PR #1478, opened 2021-09-01
Raise GMTInvalidInput exception when required z is missing

### LINKED ISSUE #1447: Raise a GMTInvalidInput exception for only x,y input to blockmedian and blockmean

Currently, no exception is raised in the `blockmedian` and `blockmean` functions if the `x` and `y` parameters are used but not the `z` parameter. This should be disallowed because these functions require three columns. For example, GMT will raise an error if a textfile with only two columns is used.

The simplest method to fix this problem would be to add the equivalent of these lines of code (using `table` rather than `data`) to the start of the `_blockm` function:
https://github.com/GenericMappingTools/pygmt/blob/cd822ca2c01f98a1496c63cc0bff8cc6661cb3bb/pygmt/src/surface.py#L81-L83

Another option could be to extend the `virtualfile_from_data` function to (optionally) check the number of columns. This would limit redundancy relative to using the same code block for all functions operating on x,y,z data (e.g., blockm*, surface, nearneighbor). If the test were in virtualfile_from_data, we could also reduce the tests required because the exception would not need to be tested in each module that requires three columns.

**Are you willing to help implement and maintain this feature?** Glad to help someone else with this issue

yohaimagen:
adding an additional if statement to the data_kind utils function that is based on a new parameter `requierd_z` and checks if z vector is provided alongside x and y.

Fixes #1447 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -75,6 +75,8 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     elif hasattr(data, "__geo_interface__"):
         kind = "geojson"
     elif data is not None:
+        if required_z and data.shape[1] < 3:
+            raise GMTInvalidInput("data must provide x, y, and z columns.")
         kind = "matrix"
     else:
         kind = "vectors"
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -66,7 +66,7 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     if data is None and (x is None or y is None):
         raise GMTInvalidInput("Must provided both x and y.")
     if data is None and required_z and (x is None or y is None or z is None):
-        raise GMTInvalidInput("Must provided both x, y, and z.")
+        raise GMTInvalidInput("Must provide x, y, and z.")
 
     if isinstance(data, str):
         kind = "file"
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -64,7 +64,7 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     if data is not None and (x is not None or y is not None or z is not None):
         raise GMTInvalidInput("Too much data. Use either data or x and y.")
     if data is None and (x is None or y is None):
-        raise GMTInvalidInput("Must provided both x and y.")
+        raise GMTInvalidInput("Must provide both x and y.")
     if data is None and required_z and (x is None or y is None or z is None):
         raise GMTInvalidInput("Must provide x, y, and z.")
 
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -14,7 +14,7 @@ import xarray as xr
 from pygmt.exceptions import GMTInvalidInput
 
 
-def data_kind(data, x=None, y=None, z=None):
+def data_kind(data, x=None, y=None, z=None, required_z=False):
     """
     Check what kind of data is provided to a module.
 
@@ -65,6 +65,8 @@ def data_kind(data, x=None, y=None, z=None):
         raise GMTInvalidInput("Too much data. Use either data or x and y.")
     if data is None and (x is None or y is None):
         raise GMTInvalidInput("Must provided both x and y.")
+    if data is None and required_z and (x is None or y is None or z is None):
+        raise GMTInvalidInput("Must provided both x, y, and z.")
 
     if isinstance(data, str):
         kind = "file"
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -65,7 +65,7 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
         raise GMTInvalidInput("Too much data. Use either data or x and y.")
     if data is None and (x is None or y is None):
         raise GMTInvalidInput("Must provide both x and y.")
-    if data is None and required_z and (x is None or y is None or z is None):
+    if data is None and required_z and z is None:
         raise GMTInvalidInput("Must provide x, y, and z.")
 
     if isinstance(data, str):
--- a/pygmt/tests/test_clib.py
+++ b/pygmt/tests/test_clib.py
@@ -410,6 +410,19 @@ def test_virtual_file_bad_direction():
                 print("This should have failed")
 
 
+def test_virtualfile_from_data_required_z():
+    """
+    Test that function fails when needs and do not provide z columns.
+    """
+    x = np.arange(5)
+    y = np.arange(5)
+    z = None
+    with clib.Session() as lib:
+        with pytest.raises(GMTInvalidInput):
+            with lib.virtualfile_from_data(x=x, y=y, z=z, required_z=True):
+                print("This should have failed")
+
+
 def test_virtualfile_from_vectors():
     """
     Test the automation for transforming vectors to virtual file dataset.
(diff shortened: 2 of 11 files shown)


yohaimagen:
Where should I add tests for not providing the `z` vector?

yohaimagen:
I added the test.

If we decide to go the route of requierd_z(see discussion at #1447) we should add `requierd_z = True` in all the functions that do require `z` column. Should we do it over here or in a separate PR for each function?

pygmt/helpers/utils.py
```diff
@@ -65,6 +65,8 @@ def data_kind(data, x=None, y=None, z=None):
         raise GMTInvalidInput("Too much data. Use either data or x and y.")
     if data is None and (x is None or y is None):
         raise GMTInvalidInput("Must provided both x and y.")
+    if data is None and required_z and (x is None or y is None or z is None):
```
meghanrjones, line 68:
```suggestion
    if data is None and required_z and z is None:
```
I do not think you need to check `x` and `y` again here since this is done before.

meghanrjones (commented):
I recommend adding `required_z` to the appropriate functions in this PR.

It would be good to raise the same exception if the data kind is `matrix` (L78) and `data.shape[1] < 3` when `required_z` is set. Would you be able to add this to the PR and add another `pytest.raises(GMTInvalidInput)` check to the new test?

yohaimagen:
I am making a list of functions that should pass `required_z=True` please add functions that I've missed. There are quite a lot of new functions that thas been wrapped recently and will be released in 0.5  please let me know which one of them needs to be set `required_z=True`.

- [ ] `plot3d`
- [ ]  `countour`
- [ ]  `blockmean`
- [ ] `blockmedian`
- [ ] `surface`

yohaimagen:
Should we remove the tests of the relevant functions that test the passing of `z` column?  https://github.com/GenericMappingTools/pygmt/blob/cd822ca2c01f98a1496c63cc0bff8cc6661cb3bb/pygmt/tests/test_contour.py#L42-L56

if so we shuld extend the test to be similar to the above.

meghanrjones:
> Should we remove the tests of the relevant functions that test the passing of `z` column?
> 
> https://github.com/GenericMappingTools/pygmt/blob/cd822ca2c01f98a1496c63cc0bff8cc6661cb3bb/pygmt/tests/test_contour.py#L42-L56
> 
> if so we shuld extend the test to be similar to the above.

Sorry for the slow response. Yes, my opinion is that invalid input tests for `data`, `x`, `y`, and `z` could be included in the `virtualfile_from_data` tests and removed from `surface`, `contour`, and `plot3d`.

Also, no need to take any action on this but just for explanation the contour code is different than the rest because it still needs refactoring for https://github.com/GenericMappingTools/pygmt/issues/949.

As for the other modules, here is a list:

- wiggle
- sphdistance (but hold off on this, since there's a PR open and currently no `x`, `y`, `z` option.)

yohaimagen:
From my end its seems that this branch is ready to be merged into main, let me know if there is anything else to do over here.

pygmt/tests/test_clib.py
```diff
@@ -410,6 +421,59 @@ def test_virtual_file_bad_direction():
...
+    data = np.ones((5, 2))
+    with clib.Session() as lib:
+        with pytest.raises(GMTInvalidInput):
+            with lib.virtualfile_from_data(data=data, required_z=True):
+                print("This should have failed")
+
+
+def test_virtualfile_from_data_fail_non_valid_data(data):
+    """
+    Should raise an exception if to less or to match data is given.
```
meghanrjones, line 439:
```suggestion
    Should raise an exception if too few or too much data is given.
```

pygmt/tests/test_clib.py
```diff
@@ -410,6 +421,59 @@ def test_virtual_file_bad_direction():
                 print("This should have failed")
 
 
+def test_virtualfile_from_data_required_z_matrix():
+    """
+    Test that function fails when needs and do not provide third z columns in a
+    matrix.
```
meghanrjones, line 427:
```suggestion
    Test that function fails when third z column in a matrix is needed but not
    provided.
```

meghanrjones (commented):
Nice work, thank you! :tada: 

As you have seen, we leave approved pull requests up for at least a day so other maintainers have a final opportunity to check it out. In the meantime, please apply the two minor grammatical suggestions.

yohaimagen:
Actually, when writing the test for the case of a 2d array data with less than three columns when z is required, I noticed that we do not check that 2d array data have at least 2 columns in the general case, should we raise an error if a 2d matrix of a shape `nx1` is passed? and if so should I add it here or open another PR?

pygmt/tests/test_clib.py
```diff
@@ -410,6 +421,59 @@ def test_virtual_file_bad_direction():
                 print("This should have failed")
 
 
+def test_virtualfile_from_data_required_z_matrix():
+    """
+    Test that function fails when third z column in a matrix is needed but not
+    provided.
+    """
+
```
seisman, line 429:
```suggestion
```

pygmt/tests/test_clib.py
```diff
@@ -410,6 +421,59 @@ def test_virtual_file_bad_direction():
...
+    data = np.ones((5, 2))
+    with clib.Session() as lib:
+        with pytest.raises(GMTInvalidInput):
+            with lib.virtualfile_from_data(data=data, required_z=True):
+                print("This should have failed")
```
seisman, line 434:
```suggestion
            with lib.virtualfile_from_data(data=data, required_z=True):
                pass
```

seisman:
> Actually, when writing the test for the case of a 2d array data with less than three columns when z is required, I noticed that we do not check that 2d array data have at least 2 columns in the general case, should we raise an error if a 2d matrix of a shape `nx1` is passed? and if so should I add it here or open another PR?

It's more complicated, because for some modules (e.g., `histogrom`), only one column is required.

seisman:
> > Actually, when writing the test for the case of a 2d array data with less than three columns when z is required, I noticed that we do not check that 2d array data have at least 2 columns in the general case, should we raise an error if a 2d matrix of a shape `nx1` is passed? and if so should I add it here or open another PR?
> 
> It's more complicated, because for some modules (e.g., `histogram`), only one column is required.

I think this PR is good for merge and we can deal with the 2-column data check in another issue/PR.

seisman (approved):

meghanrjones (approved):
I agree with @seisman that more changes aren't needed. Great work, thanks!

pygmt/helpers/utils.py
```diff
@@ -78,6 +80,8 @@ def data_kind(data, x=None, y=None, z=None):
     elif hasattr(data, "__geo_interface__"):
         kind = "geojson"
     elif data is not None:
+        if required_z and data.shape[1] < 3:
```
weiji14, line 83: Checking for `data.shape[1]` fails for [xarray.Dataset](https://xarray.pydata.org/en/v0.18.1/generated/xarray.Dataset.html) inputs with `AttributeError: 'Dataset' object has no attribute 'shape'. The `shape` attribute is valid only for `numpy.ndarray`, `pandas.DataFrame`, and maybe a few other PyData objects. Will need to create a bugfix and add some extra tests for `xarray.Dataset` inputs.
`

meghanrjones: I can submit a PR fix since I think that I suggested this. But I feel that we should not label it as a bug since this code has not yet been included in a release, so it might be confusing to list on the changelog as a bugfix relative to v0.4.1.

weiji14: Ok, I've submitted a bugfix PR at #1523. Have put on the `skip-changelog` label so it won't show up in the changelog :)

Issue #655, opened 2020-10-19
-JEPSG:4326 not writing CRS to netcdf in surface


RichardScottOZ:
**Description of the problem**

Using a Spanish government gravity dataset converted to epsg:4326.  Run blockmedian and surface seems to work fine.  If I put a proj string in it seems to write that projection to the netcdf file.

However, -JEPSG:4326 (or -Jepsg:4326 or 4326) do

es not.

I made a script version and put a debug print in the utils library generating the run arguments and that gives this:

RUNARGS -I65.15s -R-9.34335092522/3.46340435922/35.8532670285/43.8612236053
RUNARGS -GD:\Spain\GMTtext15.nc -I65.15s -J4326 -R-9.34335092522/3.46340435922/35.8532670285/43.8612236053 -Vc

**Full code that generated the error**

```python
#!/usr/bin/env python
# coding: utf-8

# In[1]:


import pygmt
import pandas as pd
import geopandas as gpd
import numpy as np
#import rasterio


#import pyvista as pv

from mpl_toolkits.mplot3d import Axes3D  # noqa: F401 unused import

import matplotlib.pyplot as plt
from matplotlib import cm
from matplotlib.ticker import LinearLocator, FormatStrFormatter
from matplotlib.patches import Rectangle

import verde as vd
#import cartopy.crs as ccrs
import dask
import pyproj
import pooch
import xarray as xr
import os

spain = gpd.read_file(r'D:\Spain\Gravimetria.shp')

# In[60]:

spain.head()

# In[3]:

Z = spain['VALU_BOU26'].values
lat =  spain.geometry.y.values
lon =  spain.geometry.x.values

# In[4]:

xmin = spain.total_bounds[0]
xmax = spain.total_bounds[2]
ymin = spain.total_bounds[1]
ymax = spain.total_bounds[3]
region = [xmin, xmax, ymin, ymax]

# In[5]:

datatable = pd.DataFrame()
datatable['longitude'] = lon.tolist()
datatable['latitude'] = lat.tolist()
datatable['bouguer_anomaly'] = Z.tolist()

# In[61]:

region2=[-9.31, 3.430053434000058, 35.92, 43.77780739100007]
region3="-9.34335092522/3.46340435922/35.8532670285/43.8612236053"

datatable_blockmedian = pygmt.blockmedian(table=datatable, spacing="65.15s", region=region3)  #e is metres
[...]

PR #1487, opened 2021-09-07
Allow passing arguments containing spaces into pygmt functions

### LINKED ISSUE #247: Arguments with spaces in strings

weiji14:

To allow for spaces in arguments to PyGMT functions, this pull request modifies the `build_arg_string` function in `pygmt/helpers/utils.py` to replace space characters (` `) with the equivalent octal code (`\040`). This enables users to do this:

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 2, 4, 8], projection="X2c", frame=r"WSne+tPlot Zero")
fig.show()
```

Note that no nested double quotes (`'"some word"'`) are needed anymore!

I have also removed some hacky workarounds that allowed for text inputs with spaces into:
- `subplot`'s `autolabel`(-A) and `title` (-T) parameters - workaround added in #1125 and #822
- `text`'s `-F` parameter
- `psconvert`'s `prefix` (-F) parameter - workaround added in #1116

Also updated some unit tests for `fig.legend` and `fig.rose` that had used arguments like `'"Some text"'`.

### **Special handling** of functions where spaces can't just be converted to `\040`:
- [x] Ensure that `pygmt.config(FORMAT_DATE_MAP="o dd")` works (c29e63229f6c46aeb47cfe068b9f6c959728f64a) 
- [x] Make sure that `fig.savefig()` saves filenames with spaces properly (#1116) without `\040` octal code characters in the filename (83c8c3b0e8c699cd52e7d539c49dabbbf7bf384c)
- [x] Ensure that PROJ4 projections with spaces (e.g. `projection="+proj4=+longlat +ellipsoid=someplanet"`) work, xref https://github.com/GenericMappingTools/pygmt/issues/655#issuecomment-713154812 (3ec772740823a12f906210edeed5284aa101ff12)

Fixes #247 :tada: :tada: :tada:

**Reminders**

**Slash Commands**


pygmt/tests/test_figure.py
```diff
@@ -116,6 +116,7 @@ def test_figure_savefig_filename_with_spaces():
     fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
     with GMTTempFile(prefix="pygmt-filename with spaces", suffix=".png") as imgfile:
         fig.savefig(imgfile.name)
+        assert r"\040" not in os.path.abspath(imgfile.name)
         assert os.path.exists(imgfile.name)
```
weiji14, line 120: A problem with this PR is that running `fig.savefig` with filenames that have spaces (introduced in #1116) will produce an output like "pygmt-filename\040with\040spaces.png". This test currently passes still, but it is a **false positive**, because Python's `os.path.exists` reads "pygmt-filename\040with\040spaces.png" as ""pygmt-filename with spaces.png"".

Any ideas how to improve the test and/or implementation of `psconvert`?

seisman:
xref: https://forum.generic-mapping-tools.org/t/errors-in-pygmt-when-there-are-spaces-in-string-arguments/2494/3

pygmt/src/config.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, **kwargs):
                     self.old_defaults[key] = lib.get_default(key)
 
         # call gmt set to change GMT defaults
-        arg_str = " ".join([f"{key}={value}" for key, value in kwargs.items()])
+        arg_str = " ".join([f'{key}="{value}"' for key, value in kwargs.items()])
```
weiji14, line 58: > xref: https://forum.generic-mapping-tools.org/t/errors-in-pygmt-when-there-are-spaces-in-string-arguments/2494/3

Ok, I've fixed the `FORMAT_DATE_MAP="o dd"` issue reported in the forum by wrapping the arguments in double quotes here in commit c29e63229f6c46aeb47cfe068b9f6c959728f64a. Didn't use `\040` because `o\040dd` doesn't work. The problem with this solution is that the workaround suggested in the forum (use `FORMAT_DATE_MAP='"o dd"'`) will break in PyGMT v0.6.0, but I think that's acceptable since that workaround isn't intuitive anyway.

pygmt/tests/test_rose.py
```diff
@@ -152,7 +152,7 @@ def test_rose_no_sectors(data_fractures_compilation):
         region=[0, 500, 0, 360],
         diameter="10c",
         labels="180/0/90/270",
-        frame=["xg100", "yg45", "+t'Windrose diagram'"],
+        frame=["xg100", "yg45", "+tWindrose diagram"],
```
seisman, line 155: Does `"+t'Windrose diagram'"` still work or not?

weiji14: Yes it does still work, but the single quotes would be printed. I think it was a typo from whoever wrote that test.

willschlitzer:
As far as I can tell, this added function seems to be working. Am I missing some remaining issues before it will be approved?

weiji14:
> As far as I can tell, this added function seems to be working. Am I missing some remaining issues before it will be approved?

Yes it works. But I have a nagging feeling that there might be some edge cases that I haven't handled. Actually, this one 👇🏻 still needs a unit test

＞Ensure that PROJ4 projections with spaces (e.g. projection="+proj4=+longlat +ellipsoid=someplanet") work, xref -JEPSG:4326 not writing CRS to netcdf in surface #655 (comment)

weiji14:
> > As far as I can tell, this added function seems to be working. Am I missing some remaining issues before it will be approved?
> 
> Yes it works. But I have a nagging feeling that there might be some edge cases that I haven't handled. Actually, this one 👇🏻 still needs a unit test
> 
>     * [ ]  Ensure that PROJ4 projections with spaces (e.g. projection="+proj4=+longlat +ellipsoid=someplanet") work, xref -JEPSG:4326 not writing CRS to netcdf in surface [-JEPSG:4326 not writing CRS to netcdf in surface #655](https://github.com/GenericMappingTools/pygmt/issues/655) (comment)

Ok, PROJ4 strings with spaces now work as of 3ec772740823a12f906210edeed5284aa101ff12 :tada: I just converted deleted the spaces for -J (projection) and it works, since GMT allows it (xref https://github.com/GenericMappingTools/gmt/blob/c697c355f39619911ffca266788e818c6ffce73e/doc/rst/source/explain_-Jproj_full.rst_#L9-L11)

pygmt/figure.py
```diff
@@ -235,10 +235,14 @@ def psconvert(self, icc_gray=False, **kwargs):
                 kwargs["N"] = "+i"
             else:
                 kwargs["N"] += "+i"
-        # allow for spaces in figure name
-        kwargs["F"] = f'"{kwargs.get("F")}"' if kwargs.get("F") else None
+
+        # Manually handle prefix -F argument so spaces aren't converted to \040
+        # by build_arg_string function. For more information, see
+        # https://github.com/GenericMappingTools/pygmt/pull/1487
+        prefix = kwargs.pop("F")
```
meghanrjones, line 242: Should there be a check before this that ``prefix`` is provided so that the user doesn't get a KeyError here?

meghanrjones:
I tried for a while to break it. Nice work on this complicated issue! 

The two issues that I came across relate to the ``outgrid`` and ``outfile`` parameters.

#### Outgrid parameter

The following will produce a filename with spaces using the main branch but the filename contains "\040" with this branch:
```
pygmt.grdcut("@earth_relief_01d", region=[0, 1, 0, 1], outgrid="test grid output.nc")
```

#### Outfile parameter

File outputs with spaces still require using the trick of wrapping double quotes in single quotes (this could be addressed separately from this PR).

```
pygmt.blockmean(data="@ship_15.txt", region=[245, 255, 20, 30], spacing="5m", outfile='"Test file output.txt"') # Works
pygmt.blockmean(data="@ship_15.txt", region=[245, 255, 20, 30], spacing="5m", outfile="Test file output.txt") # Doesn't work
```

pygmt/figure.py
weiji14: Good point. Looking at the docs at https://docs.generic-mapping-tools.org/6.3/psconvert.html, it reads like `prefix`/`-F` is optional, but when I tried running `psconvert` without the `-F` option, it throws an error: `psconvert [ERROR]: Modern GMT mode requires the -F option`. See https://github.com/GenericMappingTools/gmt/blob/adb244afa51ca7246cc051080c9d47193087d6c2/src/psconvert.c#L1041-L1042

So since PyGMT is modern mode only, it should be fine to keep this line intact.

weiji14:
> I tried for a while to break it. Nice work on this complicated issue!

Thanks for testing this! I still have a nagging feeling that there's a random edge case that will break this, but we'll find out once this gets released into the wild :smiling_imp:

> The two issues that I came across relate to the `outgrid` and `outfile` parameters.
> #### Outgrid parameter
> 
> The following will produce a filename with spaces using the main branch but the filename contains "\040" with this branch:
> 
> ```
> pygmt.grdcut("@earth_relief_01d", region=[0, 1, 0, 1], outgrid="test grid output.nc")
> ```
> 
> #### Outfile parameter
> 
> File outputs with spaces still require using the trick of wrapping double quotes in single quotes (this could be addressed separately from this PR).
> 
> ```
> pygmt.blockmean(data="@ship_15.txt", region=[245, 255, 20, 30], spacing="5m", outfile='"Test file output.txt"') # Works
> pygmt.blockmean(data="@ship_15.txt", region=[245, 255, 20, 30], spacing="5m", outfile="Test file output.txt") # Doesn't work
> ```

Right, fixing this would require a workaround like the one applied to `fig.savefig` mentioned in https://github.com/GenericMappingTools/pygmt/pull/1487#discussion_r783577450, but it'll be a lot of work to apply it individually to all the functions having an outgrid/outfile parameter. Maybe leave it for another time, and honestly, we really need to discourage people from having spaces in filenames in the first place.

pygmt/figure.py
meghanrjones: The docs say "If no input files are given, will convert the current active figure (see pygmt.figure). In this case, an output name must be given using parameter prefix." How does one provide other input files than the current figure using pygmt.Figure.psconvert?

If we chose not to support operations analogous to ``psconvert -Tg test.ps`` (which uses the same prefix as the original file name), I think we should make ``prefix`` required in the function signature or raise an invalid input exception to avoid the traceback below. I don't think it's obvious to that the error is caused by not using ``prefix``.

image

meghanrjones (approved):
There are still a couple issues to work out as mentioned in https://github.com/GenericMappingTools/pygmt/pull/1487/files#r825371307 and https://github.com/GenericMappingTools/pygmt/pull/1487#issuecomment-1065965868, but this is a big improvement that would benefit users in v0.6.0.

pygmt/figure.py
weiji14: > How does one provide other input files than the current figure using pygmt.Figure.psconvert?

We can't, because it has not been implemented yet. The current `fig.psconvert` is a method of the `pygmt.Figure()` class and is tied to the `fig`, and can only 'psconvert' the active figure.

> If we chose not to support operations analogous to `psconvert -Tg test.ps` (which uses the same prefix as the original file name), I think we should make `prefix` required in the function signature or raise an invalid input exception to avoid the traceback below. I don't think it's obvious to that the error is caused by not using `prefix`.

Ok, I've added a try-except block in 801ba01b3c83a134f6dd15f756223f8e3663513a to raise a GMTInvalidInput if `prefix`/`-F` is not given.

seisman:
The one thing that makes little sense to me is shown in the following example:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 2, 4, 8], projection="X10c", frame=["WSne+tPlot Zero", "xaf+lX label+uK"])
fig.show()
```
As you can see, I have to use `"xaf+lX label+uK"`, not `"xaf+l'X label'+uK"` or `'xaf+l"X label"+uK'`. For me, the last two are much easier to read.

weiji14:
> As you can see, I have to use `"xaf+lX label+uK"`, not `"xaf+l'X label'+uK"` or `'xaf+l"X label"+uK'`. For me, the last two are much easier to read.

I just tested all three examples and only `"xaf+l'X label'+uK"` doesn't work properly (it prints `'X label` instead of `X label`), but that's a know issue with using single quotes. This branch should be backwards compatible with the `'xaf+l"X label"+uK'` method, and it now allows for `"xaf+lX label+uK"` which isn't nice but works if people like that format.

seisman:
> > As you can see, I have to use `"xaf+lX label+uK"`, not `"xaf+l'X label'+uK"` or `'xaf+l"X label"+uK'`. For me, the last two are much easier to read.
> 
> I just tested all three examples and only `"xaf+l'X label'+uK"` doesn't work properly (it prints `'X label` instead of `X label`), but that's a known issue with using single quotes. This branch should be backwards compatible with the `'xaf+l"X label"+uK'` method, and it now allows for `"xaf+lX label+uK"` which isn't nice but works if people like that format.

Ah, you're right. Good to know that I can still use `'xaf+l"X label"+uK'`. 

I'm reading how GMT parses a long text string  https://github.com/GenericMappingTools/gmt/blob/434ed1848a1e8886931694b2de4c352e4c6cddae/src/gmt_parse.c#L600, it seems it ignores double quotes (that's why `'xaf+l"X label"+uK'`), but doesn't ignore single quotes (so `"xaf+l'X label'+uK"` doesn't work).

pygmt/helpers/utils.py
```diff
@@ -131,6 +131,11 @@ def build_arg_string(kwargs):
     same command line argument. For example, the kwargs entry ``'B': ['xa',
     'yaf']`` will be converted to ``-Bxa -Byaf`` in the argument string.
 
+    Note that spaces ` ` in arguments are converted to the equivalent octal
+    code `\040`, except in the case of -J (projection) arguments where PROJ4
```
seisman, line 135: It's ReST, not markdown.
```suggestion
    Note that spaces `` `` in arguments are converted to the equivalent octal
    code ``\040``, except in the case of -J (projection) arguments where PROJ4
```

pygmt/helpers/utils.py
```diff
@@ -185,11 +200,19 @@ def build_arg_string(kwargs):
...
             gmt_args.append(f"-{key}")
         else:
-            gmt_args.append(f"-{key}{kwargs[key]}")
+            if key != "J":  # non-projection parameters
+                _value = str(kwargs[key]).replace(" ", r"\040")
+            else:
+                # special handling if key == "J" (projection)
+                # remove any spaces in PROJ4 string
+                _value = str(kwargs[key]).replace(" ", "")
+            gmt_args.append(rf"-{key}{_value}")
```
seisman, line 215: I know almost nothing about PROJ4, but does a PROJ4 string always contains `+proj=xxx` or `EPSG`? If yes, then we can simplify the if-test to something like:
```
            if key != "J" or (key == "J" and "+proj" not in value and "EPSG" not in value):
                _value = str(kwargs[key]).replace(" ", r"\040")
            else:
                # special handling if -J + PROJ4 string
                # remove any spaces in PROJ4 string
                _value = str(kwargs[key]).replace(" ", "")
            gmt_args.append(rf"-{key}{_value}")
```
The motivation is, since PROJ4 is not commonly used, it seems a waste of time to do call the `replace` method for GMT-style -J.

weiji14: Not sure if they all have either `+proj=xxx` or `EPSG` (see https://proj.org/operations/projections/index.html), but they probably do? I know that there's `+init=EPSG:xxx` which works when I tried here, but that's actually deprecated syntax (see https://pyproj4.github.io/pyproj/stable/gotchas.html#init-auth-auth-code-should-be-replaced-with-auth-auth-code) and not probably something we want to allow.

Also, not sure if 2 extra if-checks saves time, considering that either way, a `str.replace` still needs to happen for the space character - either to `\040` or to nothing ``.

seisman (approved):

weiji14:
Thanks team! I'll merge this in and close that almost 4 year old #247 bug!!

PR #1247, opened 2021-04-27
Fix ModuleNotFoundError error when IPython is not installed

seisman:
**Description of proposed changes**

Fixes #1241.

A test is needed but currently I have no idea how to test it.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1496, opened 2021-09-12
Make IPython partially optional on CI to increase test coverage of figure.py

weiji14:

The [`pygmt/figure.py`](https://github.com/GenericMappingTools/pygmt/blob/v0.4.1/pygmt/figure.py) file has the smallest code coverage in the PyGMT project, with 15 lines uncovered. This is mostly due to if-branches that only get triggered when [`IPython`](https://ipython.org) isn't installed.

image

This PR makes `IPython` an optional dependency in Continuous Integration for the minimum dependency stream (Python 3.7/NumPy 1.18), so that those uncovered if-branches are covered. Tests which require `IPython` will have a [`pytest.importorskip`](https://docs.pytest.org/en/stable/skipping.html#skipping-on-a-missing-import-dependency) line added so that the test suite doesn't break.

See coverage diff at https://app.codecov.io/gh/GenericMappingTools/pygmt/compare/1496/changes. Project test coverage increased by 0.96% to 98.25%!

References:
- https://docs.pytest.org/en/stable/skipping.html#skipping-on-a-missing-import-dependency
- https://sphinx-hxl.readthedocs.io/en/stable/usage/extensions/doctest.html#skipping-tests-conditionally
- https://stackoverflow.com/questions/42708809/python-doctest-skip-a-test-conditionally

Addresses https://github.com/GenericMappingTools/pygmt/pull/1247#issuecomment-827991167 and patches #529 with more code coverage.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -5,11 +5,17 @@ Doesn't include the plotting commands which have their own test files.
 """
 import os
 
+try:
+    import IPython
+except ModuleNotFoundError:
+    IPython = None  # pylint: disable=invalid-name
+
+
 import numpy as np
 import numpy.testing as npt
 import pytest
 from pygmt import Figure, set_display
-from pygmt.exceptions import GMTInvalidInput
+from pygmt.exceptions import GMTError, GMTInvalidInput
 from pygmt.helpers import GMTTempFile
 
 
@@ -198,6 +204,18 @@ def test_figure_show_invalid_method():
     fig.basemap(region="10/70/-300/800", projection="X3i/5i", frame="af")
     with pytest.raises(GMTInvalidInput):
         fig.show(method="test")
+
+
+@pytest.mark.skipif(IPython is not None, reason="run without IPython installed")
+def test_figure_show_notebook_error_without_ipython():
+    """
+    Test to check if an error is raised when display method is 'notebook', but
+    IPython is not installed.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 2, 3], frame=True)
+    with pytest.raises(GMTError):
+        fig.show(method="notebook")
 
 
 def test_figure_set_display_invalid():
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -182,11 +182,11 @@ def test_figure_savefig():
     )
 
 
+@pytest.mark.skipif(IPython is None, reason="run when IPython is installed")
 def test_figure_show():
     """
     Test that show creates the correct file name and deletes the temp dir.
     """
-    pytest.importorskip("IPython")
     fig = Figure()
     fig.basemap(region="10/70/-300/800", projection="X3i/5i", frame="af")
     fig.show()
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -58,6 +58,23 @@ def test_figure_region_country_codes():
     npt.assert_allclose(fig.region, np.array([0.0, 360.0, -90.0, 90.0]))
 
 
+def test_figure_repr():
+    """
+    Make sure that figure output's PNG and HTML printable representations look
+    ok.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 2, 3], frame=True)
+    # Check that correct PNG 8-byte file header is produced
+    # https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_header
+    repr_png = fig._repr_png_()
+    assert repr_png.hex().startswith("89504e470d0a1a0a")
+    # Check that correct HTML image tags are produced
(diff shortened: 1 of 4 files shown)


pygmt/tests/test_figure.py
```diff
@@ -52,6 +58,23 @@ def test_figure_region_country_codes():
...
+def test_figure_repr():
+    """
+    Make sure that figure output's PNG and HTML printable representations look
+    ok.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 2, 3], frame=True)
+    # Check that correct PNG 8-byte file header is produced
+    # https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_header
+    repr_png = fig._repr_png_()  # pylint: disable=protected-access
```
seisman, line 70: Is `_repr_png_` and `_repr_html_` useful? I have no idea how these functions can be used.

pygmt/tests/test_figure.py
```diff
@@ -199,6 +223,27 @@ def test_figure_show_invalid_method():
...
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 2, 3], frame=True)
+    with pytest.raises(GMTError):
+        fig.show(method="notebook")
+
+
+def test_figure_display_external():
+    """
+    Test to check that a figure can displayed in an external window.
```
seisman, line 240:
```suggestion
    Test to check that a figure can be displayed in an external window.
```

pygmt/tests/test_figure.py
weiji14: Good question. They are used by [`IPython.display.display_png`](https://ipython.readthedocs.io/en/7.27.0/api/generated/IPython.display.html#IPython.display.display_png) and [`IPython.display.display_html`](https://ipython.readthedocs.io/en/7.27.0/api/generated/IPython.display.html#IPython.display.display_html) respectively, i.e. for display in Jupyter notebooks.

Technically I should have written these unit test using `IPython.display.display_png(fig)` and `IPython.display.display_html(fig)` instead of testing these non-public functions directly, but it's hard to test a pop-up figure. So I've settled for testing the `__repr_png__` and `__repr_html__` instead (which works with or without IPython installed). Con with this approach is that I need to do `pylint: disable=protected-access` :upside_down_face:

seisman: > They are used by [`IPython.display.display_png`](https://ipython.readthedocs.io/en/7.27.0/api/generated/IPython.display.html#IPython.display.display_png) and [`IPython.display.display_html`](https://ipython.readthedocs.io/en/7.27.0/api/generated/IPython.display.html#IPython.display.display_html) respectively, i.e. for display in Jupyter notebooks.

PyGMT already provides the `Figure.show()` function for display in Jupyter notebooks. Removing `_repr_png_()` and `_repr_html_()` doesn't affect the `Figure.show()` function. Do you think we should remove these two functions?

weiji14: It might not affect `Figure.show()`, but I think we need to keep it. Having `__repr_png__()` and `__repr_html__` means that putting just `fig` at the last line of a Jupyter notebook cell will display the plot. E.g. like this:

image

This works in Jupyter/IPython because of a [`IPython.display.display_html(fig)`](https://ipython.readthedocs.io/en/7.27.0/api/generated/IPython.display.html#IPython.display.display_html) call. Similarly, if you put a `df` (pandas.DataFrame) on the last line of a cell, a nicely formatted html table will be printed because Jupyter/IPython does `IPython.display.display_html(df)` (see the `__repr_html__` at https://github.com/pandas-dev/pandas/blob/73c68257545b5f8530b7044f56647bd2db92e2ba/pandas/core/frame.py#L1007-L1049).

pygmt/tests/test_figure.py
```diff
@@ -199,6 +223,27 @@ def test_figure_show_invalid_method():
...
+        fig.show(method="notebook")
+
+
+def test_figure_display_external():
+    """
+    Test to check that a figure can be displayed in an external window.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 3, 6, 9], projection="X1c", frame=True)
+    fig.show(method="external")
```
seisman, line 244: The CI can't open external viewers. Why does this test work?

seisman (approved):

pygmt/tests/test_figure.py
weiji14: Yeah, the console won't show a window pop-up, but I think the code will still execute fine somehow. Looking at the codecov diff at https://app.codecov.io/gh/GenericMappingTools/pygmt/compare/1496/changes#D2L237, the `launch_external_viewer` function is being called up to the `time.sleep(0.5)` line, so it seems that Linux/macOS/Windows launchers (`xdg-open`/`open`/not-sure-what's-on windows) are available on CI?

image

PR #1506, opened 2021-09-13
WIP: Add missing aliases to xyz2grd

willschlitzer:
This pull request adds missing aliases for `xyz2grd`.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/xyz2grd.py
+++ b/pygmt/src/xyz2grd.py
@@ -14,14 +14,18 @@ from pygmt.io import load_dataarray
 
 @fmt_docstring
 @use_alias(
+    A="duplicate",
+    D="dname",
     G="outgrid",
     I="spacing",
     R="region",
     V="verbose",
+    Z="onecolumn",
+    r="registration",
 )
 @kwargs_to_strings(R="sequence")
 def xyz2grd(table, **kwargs):
-    """
+    r"""
     Create a grid file from table data.
 
     xyz2grd reads one or more z or xyz tables and creates a binary grid file.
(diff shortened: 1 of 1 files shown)


meghanrjones:
Please change:

```
    Full option list at :gmt-docs:`xyz2grd.html`

    Parameters
```

to

```
    Full option list at :gmt-docs:`xyz2grd.html`

    {aliases}

    Parameters
```

willschlitzer:
> Please change:
> 
> ```
>     Full option list at :gmt-docs:`xyz2grd.html`
> 
>     Parameters
> ```
> 
> to
> 
> ```
>     Full option list at :gmt-docs:`xyz2grd.html`
> 
>     {aliases}
> 
>     Parameters
> ```

changed

pygmt/src/xyz2grd.py
```diff
@@ -14,14 +14,18 @@
 
 @fmt_docstring
 @use_alias(
+    A="duplicate",
+    D="dname",
```
seisman, line 18: I don't think `dname` is a good name.

willschlitzer: I just pulled it directly from the GMT docs. What would you recommend instead?

seisman: I don't have a good name to recommend. I just feel that it's difficult to guess what the parameter does from the name `dname`.

pygmt/src/xyz2grd.py
```diff
@@ -40,9 +46,94 @@ def xyz2grd(table, **kwargs):
...
+
+        **d** 8-byte floating point double precision
+
+        Default format is scanline orientation of ASCII numbers: **-ZTLa**.
+        The difference between **A** and **a** is that the latter can decode
+        both *date*\ **T**\ *clock* and *ddd:mm:ss[.xx]* formats but expects
+        each input record to have a single value, while the former can handle
+        multiple values per record but can only parse regular floating point
+        values. Translate incoming *z*-values via the **-i**\ 0 option and
+        needed modifiers.
```
meghanrjones, line 135: I don't think I have ever actually seen a data file that contains x,y,z values in a 1-column ASCII/binary table. Has anyone else? I wonder if this is a relic of some older data standards, in which case it might not be worth including in PyGMT since it's such an obscure argument and would be quite difficult to make pythonic.

meghanrjones:
I think xyz2grd should go under "operations on tabular data" rather than "operations on grids" in `doc/api/index.rst`.

pygmt/src/xyz2grd.py
seisman: > a data file that contains x,y,z values in a 1-column ASCII/binary table.

I think it means the 1-column ASCII/binary file only contains the Z values. The X and Y coordinates are determined from `-R` and `-I` option automatically, because the data are uniformly spaced in X and Y dimensions.

pygmt/src/xyz2grd.py
meghanrjones: It seems to me that xyz2grd -D does the same thing as `grdedit` in setting the header information. Related to past conversations about -X and -Y versus shift_origin, would it be better to leave this out of xyz2grd and instead add a wrapper for grdedit in order to only have one way to do this task?

pygmt/src/xyz2grd.py
meghanrjones: Ah, I misunderstood the purpose. Yes, I agree that is quite common and should be included.

pygmt/src/xyz2grd.py
```diff
@@ -40,9 +46,94 @@ def xyz2grd(table, **kwargs):
...
+
+        **I** uint32_t, unsigned 4-byte integer
+
+        **l** int64_t, long (8-byte) integer
+
+        **L** uint64_t, unsigned long (8-byte) integer
+
+        **f** 4-byte floating point single precision
+
+        **d** 8-byte floating point double precision
```
meghanrjones, line 127:
```suggestion
        - **A** ASCII representation of one or more floating point values per
           record
        - **a** ASCII representation of a single item per record
        - **c** int8_t, signed 1-byte character
        - **u** uint8_t, unsigned 1-byte character
        - **h** int16_t, signed 2-byte integer
        - **H** uint16_t, unsigned 2-byte integer
        - **i** int32_t, signed 4-byte integer
        - **I** uint32_t, unsigned 4-byte integer
        - **l** int64_t, long (8-byte) integer
        - **L** uint64_t, unsigned long (8-byte) integer
        - **f** 4-byte floating point single precision
        - **d** 8-byte floating point double precision
```

willschlitzer:
> I think xyz2grd should go under "operations on tabular data" rather than "operations on grids" in `doc/api/index.rst`.

changed in fda29eb

pygmt/src/xyz2grd.py
```diff
@@ -14,14 +14,18 @@
 
 @fmt_docstring
 @use_alias(
+    A="duplicate",
```
weiji14, line 17: GMT.jl uses ['multiple_nodes' for `-A`](https://www.generic-mapping-tools.org/GMT.jl/v0.31/#GMT.xyz2grd), should we follow?

pygmt/src/xyz2grd.py
```diff
@@ -14,14 +14,18 @@
...
+    D="dname",
     G="outgrid",
     I="spacing",
     R="region",
     V="verbose",
+    Z="onecolumn",
```
weiji14, line 23: Again, GMT.jl uses ['flags' for `-Z`](https://www.generic-mapping-tools.org/GMT.jl/v0.31/#GMT.xyz2grd), should we follow?

pygmt/src/xyz2grd.py
willschlitzer: I prefer `duplicate` but I think having parameters named consistently across wrappers is more important

pygmt/src/xyz2grd.py
seisman: `flags` is a bad name, because "an option" sometimes is also called "a flag".

pygmt/src/xyz2grd.py
willschlitzer: removed in d890a75

pygmt/src/xyz2grd.py
meghanrjones: My preference would be 'z_convention', since this is what the *flags* provided in the argument is supposed to describe.

Issue #1493, opened 2021-09-11
Add region docstring to COMMON_OPTIONS for non-plotting modules


willschlitzer:
The use of `{R}` in docstrings begins with "Required if this is the first plot command.", which doesn't make sense for not plotting modules. I think there should be an `{R-noplot}`-type option that just states that the region can be set.

**Are you willing to help implement and maintain this feature?** Yes

meghanrjones:
Good point. Another factor is that ``region`` and ``projection`` are not always required for the first plot command (not well documented in GMT, I'll look into fixing this but cannot easily find the conversation about when these are and are not required). Here is an example that works fine without ``region`` or ``projection``.

```
import pygmt

data = pygmt.datasets.load_sample_bathymetry()

fig = pygmt.Figure()
fig.plot(data=data, style="p1p")
fig.show()
```

seisman:
> Another factor is that `region` and `projection` are not always required for the first plot command (not well documented in GMT, I'll look into fixing this but cannot easily find the conversation about when these are and are not required).

The `THIS_MODULE_NEEDS` macro in the C code determine if `-J` or `-R` option is required. For example, `JR` means both `-J` and `-R` are required, `d` or `g` means the module can determine the region from the input dataset or grid, thus `-R` option is not required.

For about half of the plotting modules, `-R` option is not required, but for some non-plotting modules (e.g., `grdmask`, `grdlandmask`), `-R` is required.

I think we can simply define `{R}` as:
```
    "R": r"""
        region : str or list
            *xmin/xmax/ymin/ymax*\ [**+r**][**+u**\ *unit*].
            Specify the :doc:`region </tutorials/regions>` of interest.""",
``` 
and add some trailing text if `-R` is required. For example, in `pscoast` docstring, we can use
``` 
{R}
    *Required if this is the first plot command.*
```
in `grdlandmask` docstring, we can use:
```
{R}
    *Required.*
```

meghanrjones:
I agree with @seisman's suggestion and think we should update the docs for ``projection`` similarly. Here's the distribution for modules that are already wrapped and require ``region`` and/or ``projection``.

**Requires only region**
- [ ] blockmean, blockmedian, blockmode
- [ ] surface
- [ ] grdlandmask
- [ ] sphdistance
- [ ] xyz2grd

**Requires only projection**
- [ ] grdproject
- [ ] contour
- [ ] grdcontour
- [ ] grdimage
- [ ] grdview
[...]

PR #1510, opened 2021-09-14
Update region and projection standard docstrings

willschlitzer:
This pull request modifies the COMMON_OPTIONS text for `region` and `projection`. It also adds that it is requirement for certain plotting commands to individual modules. This partially addresses #1493.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/histogram.py
+++ b/pygmt/src/histogram.py
@@ -54,7 +54,9 @@ def histogram(self, table, **kwargs):
         Pass in either a file name to an ASCII data table, a Python list, a 2D
         {table-classes}.
     {J}
+        *Required if this is the first plot command.*
     {R}
+        *Required if this is the first plot command.*
     {B}
     {CPT}
     {G}
--- a/pygmt/src/solar.py
+++ b/pygmt/src/solar.py
@@ -50,7 +50,9 @@ def solar(self, terminator="d", terminator_datetime=None, **kwargs):
         passed as a string or Python datetime object.
         [Default is the current UTC date and time]
     {R}
+        *Required if this is the first plot command.*
     {J}
+        *Required if this is the first plot command.*
     {B}
     fill : str
         Color or pattern for filling of terminators.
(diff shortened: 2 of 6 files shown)


pygmt/src/histogram.py
```diff
@@ -54,7 +54,9 @@ def histogram(self, table, **kwargs):
         Pass in either a file name to an ASCII data table, a Python list, a 2D
         {table-classes}.
     {J}
+        *Required if this is the first plot command.*
     {R}
+        *Required if this is the first plot command.*
```
seisman, line 59: `-J` and `-R` are not required for `histogram`.

This CLI command works well without `-J` and `-R`:
```
gmt histogram @v3206_06.txt -F -T250 -W0.25p -B -pdf plot
```

The comment in the [pshistogram.c source code](https://github.com/GenericMappingTools/gmt/blob/3ebaae9ad5385309e5ec89f80fbce0cad6a30c7c/src/pshistogram.c#L35) explains why `THIS_MODULE_NEEDS` is `JR`, but `-J` and `-R` are not required:

> Note: The NEEDS must be JR.  Although pshistogram can create a region from data, it
> does so indirectly by building the histogram and setting the ymin/ymax that way, NOT by
> reading y data (which is what a NEEDS of d or r would do).

pygmt/src/solar.py
```diff
@@ -50,7 +50,9 @@ def solar(self, terminator="d", terminator_datetime=None, **kwargs):
         passed as a string or Python datetime object.
         [Default is the current UTC date and time]
     {R}
+        *Required if this is the first plot command.*
     {J}
+        *Required if this is the first plot command.*
```
seisman, line 55: `-J` and `-R` are not required for `solar`, because the following CLI command works:
```
gmt solar -W1p -Tdc -pdf map
```

seisman:
The doctest in https://github.com/GenericMappingTools/pygmt/blob/c81555e11d5013aed46b409d6c67c2f42dcff4d1/pygmt/helpers/decorators.py#L451 fails. You need to remove the text `*Required if this is the first plot command*.` from the `fmt_docstring` doctest.

pygmt/src/histogram.py
willschlitzer: fixed in f8f9cf8

pygmt/src/solar.py
willschlitzer: fixed in fd18364

Issue #1436, opened 2021-08-11
Functions that import sample GMT remote files


willschlitzer:
I've recently opened some pull requests (#1386 and #1420) that import the sample remote files, and I've realized there are a lot of sample files on the GMT server that are used in the GMT examples. Should functions be added that import remote files (e.g. "ternary.txt" and "EGM96_to_36.txt" for `ternary` and `sph2grd`) whenever they are used in a GMT module example? I think it's more Pythonic to have the datasets available as a DataFrame/array than just the text file, and I think those sample files are a great resource when writing a test or making an example, but I could also see it as bad practice to be adding all of these functions that import little-used sample files.

meghanrjones:
Functions to load the file are nice, though I am not sure that we should export them or include in the API docs since there's no stability guarantee for those files (see https://docs.generic-mapping-tools.org/dev/datasets/remote-data.html#cache-file-updates).

meghanrjones:
xref https://github.com/GenericMappingTools/gmtserver-admin/issues/100, which has some discussions about alternate options for data files used in examples.

willschlitzer:
It looks like the consensus is that cache files should be used only in examples. Should the PyGMT examples just call the remote files directly or should there be a Python function to load the remote files first before they're used in the examples.

meghanrjones:
> It looks like the consensus is that cache files should be used only in examples. Should the PyGMT examples just call the remote files directly or should there be a Python function to load the remote files first before they're used in the examples?

This is a good question. I am not sure the best solution here, but my current thinking is that we should either have a separate namespace for the functions related to GMT’s cache data relative to the earth relief, age, etc grids or a single public function like `load_cache_data` with a `filename` parameter that can be used to specify the cache file. This way, there would be more information that the data accessed are not 'stable' (which is currently only documented on the GMT side) and the smaller functions for loading each file for the examples would not overwhelm the API docs.

willschlitzer:
> > It looks like the consensus is that cache files should be used only in examples. Should the PyGMT examples just call the remote files directly or should there be a Python function to load the remote files first before they're used in the examples?
> 
[...]

PR #1512, opened 2021-09-15
Wrap filter1d

willschlitzer:
This pull request wraps `filter1d`.

**Reminders**

**Slash Commands**


willschlitzer:
The style checks are getting stuck on the argument ("\s+") passed to the `sep=` parameter for `pd.read_csv`. As far as I can tell, that is the correct way to be importing the MaunaLoa_CO2.txt data file and reading it into a dataframe. Is there a way to get this style check to pass?

pygmt/tests/test_filter1d.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pytest.fixture(scope="module", name="data")
+def fixture_table():
+    """
+    Load the grid data from the sample earth_relief file.
+    """
+    fname = which("@MaunaLoa_CO2.txt", download="c")
+    data = pd.read_csv(
+        fname, header=None, skiprows=1, sep="\s+", names=["date", "co2_ppm"]
```
weiji14, line 23: > The style checks are getting stuck on the argument ("\s+") passed to the `sep=` parameter for `pd.read_csv`. As far as I can tell, that is the correct way to be importing the MaunaLoa_CO2.txt data file and reading it into a dataframe. Is there a way to get this style check to pass?

Try using a raw-string.
```suggestion
        fname, header=None, skiprows=1, sep=r"\s+", names=["date", "co2_ppm"]
```
Similar to here:
https://github.com/GenericMappingTools/pygmt/blob/4b80f3c269f39917c8bb7ee96c5a253279877024/pygmt/datasets/samples.py#L26

pygmt/tests/test_filter1d.py
```diff
@@ -0,0 +1,96 @@
...
+from pygmt.helpers import GMTTempFile
+from pygmt.src import which
+
+
+@pytest.fixture(scope="module", name="data")
+def fixture_table():
+    """
+    Load the grid data from the sample earth_relief file.
+    """
+    fname = which("@MaunaLoa_CO2.txt", download="c")
```
weiji14, line 21: This `@MaunaLoa_CO2.txt` file needs to be added to the cache at https://github.com/GenericMappingTools/pygmt/blob/4b80f3c269f39917c8bb7ee96c5a253279877024/pygmt/helpers/testing.py#L146

Please add it in a separate Pull Request. Also, maybe a good time to implement the seaborn-like sample dataset loader as discussed in #1436.

pygmt/tests/test_filter1d.py
willschlitzer: Didn't realize I could pass a raw string in as an argument. Thanks!

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@fmt_docstring
+@use_alias(
+    E="end",
+    F="filter",
+)
+def filter1d(data, output_type="pandas", outfile=None, **kwargs):
+    r"""
+    Time domain filtering of 1-D data tables
```
michaelgrund, line 24:
```suggestion
    Time domain filtering of 1-D data tables.
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,96 @@
...
+    outliers and the special options are not needed.
+    Read a table and output as a :class:`numpy.ndarray`,
+    :class:`pandas.DataFrame`, or ASCII file.
+
+    Full option list at :gmt-docs:`filter1d.html`
+
+    {aliases}
+
+    Parameters
+    ----------.
```
michaelgrund, line 39:
```suggestion
    ----------
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,91 @@
...
+
+        - None if ``outfile`` is set (output will be stored in file set by
+          ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is
+          not set (depends on ``output_type`` [Default is
+          :class:`pandas.DataFrame`])
+
+    """
+    if "F" not in kwargs:
+        raise GMTInvalidInput("""Pass a required argument to 'filter'.""")
```
seisman, line 58:
```suggestion
        raise GMTInvalidInput("Pass a required argument to 'filter'.")
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,91 @@
...
+    if output_type not in ["numpy", "pandas", "file"]:
+        raise GMTInvalidInput(
+            """Must specify format as either numpy, pandas, or file."""
```
seisman, line 61:
```suggestion
            "Must specify format as either numpy, pandas, or file."
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,91 @@
...
+    if outfile is not None and output_type != "file":
+        msg = (
+            f"Changing `output_type` of filter1d from '{output_type}' to 'file' "
+            "since `outfile` parameter is set. Please use `output_type='file'` "
+            "to silence this warning."
+        )
+        warnings.warn(msg, category=RuntimeWarning, stacklevel=2)
+        output_type = "file"
+    elif output_type == "file" and outfile is None:
+        raise GMTInvalidInput("""Must specify outfile for ASCII output.""")
```
seisman, line 72:
```suggestion
        raise GMTInvalidInput("Must specify outfile for ASCII output.")
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,91 @@
...
+    ----------
+    output_type : str
+        Determine the format the xyz data will be returned in [Default is
+        ``pandas``]:
+
+            - ``numpy`` - :class:`numpy.ndarray`
+            - ``pandas``- :class:`pandas.DataFrame`
+            - ``file`` - ASCII file (requires ``outfile``)
+    outfile : str
+        The file name for the output ASCII file.
```
seisman, line 43: Need to add some docstrings for `E` and `F`.

seisman:
@willschlitzer Sorry for the long delay in reviewing this PR. We will try to finalize this PR in the next few days or weeks.

Could you please resolve the conflicts and merge the main branch?

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+import pandas as pd
+from pygmt.clib import Session
+from pygmt.exceptions import GMTInvalidInput
+from pygmt.helpers import GMTTempFile, build_arg_string, fmt_docstring, use_alias
+
+
+@fmt_docstring
+@use_alias(
+    E="end",
+    F="filter",
```
seisman, line 15: Following the alias in [GMT.jl](https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.filter1d):
```suggestion
    F="filter",
    N="time_col",
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+    r"""
+    Time domain filtering of 1-D data tables.
+
+    Filter a general time domain for multiple column time
+    series data. The user specifies which column is the time (i.e., the
+    independent variable). The fastest operation
+    occurs when the input time series are equally spaced and have no gaps or
+    outliers and the special options are not needed.
+    Read a table and output as a :class:`numpy.ndarray`,
+    :class:`pandas.DataFrame`, or ASCII file.
```
seisman, line 27:
```suggestion
    A general time domain filter for multiple column time
    series data. The user specifies which column is the time (i.e., the
    independent variable) via ``time_cols``. The fastest operation
    occurs when the input time series are equally spaced and have no gaps or
    outliers and the special options are not needed.
    Read a table and output as a :class:`numpy.ndarray`,
    :class:`pandas.DataFrame`, or ASCII file.
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Full option list at :gmt-docs:`filter1d.html`
+
+    {aliases}
+
+    Parameters
+    ----------
+    filter : str
+        **type**\ *width*\ [**+l**\|\ **u**].
```
seisman, line 36: See https://github.com/GenericMappingTools/gmt/issues/6599
```suggestion
        **type**\ *width*\ [**+h**].
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+        value. Append **+l** or **+u** if you rather want
+        to return the lowermost or uppermost of the modal values.
+
+        (**l**) Lower: Return the minimum of all values.
+
+        (**L**) Lower: Return minimum of all positive values only.
+
+        (**u**) Upper: Return maximum of all values.
+
+        (**U**) Upper: Return maximum of all negative values only.
```
seisman, line 69: image

I feel the current documentation is not easy to read, so I reformat the documentation like below:

```suggestion
        Available convolution filter types are:

        - (**b**) Boxcar: All weights are equal.
        - (**c**) Cosine Arch: Weights follow a cosine arch curve.
        - (**g**) Gaussian: Weights are given by the Gaussian function.
        - (**f**) Custom: Instead of *width* give name of a one-column file
          with your own weight coefficients.

        Non-convolution filter types are:

        - (**m**) Median: Returns median value.
        - (**p**) Maximum likelihood probability (a mode estimator): Return
          modal value. If more than one mode is found we return their average
          value. Append **+l** or **+u** if you rather want
          to return the lowermost or uppermost of the modal values.
        - (**l**) Lower: Return the minimum of all values.
        - (**L**) Lower: Return minimum of all positive values only.
        - (**u**) Upper: Return maximum of all values.
        - (**U**) Upper: Return maximum of all negative values only.
```

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+    F="filter",
+    N="time_col",
+)
+def filter1d(data, output_type="pandas", outfile=None, **kwargs):
+    r"""
+    Time domain filtering of 1-D data tables.
+
+    A general time domain filter for multiple column time
+    series data. The user specifies which column is the time (i.e., the
+    independent variable) via ``time_cols``. The fastest operation
```
seisman, line 24:
```suggestion
    independent variable) via ``time_col``. The fastest operation
```

seisman (approved):
Looks good to me except for a typo.

seisman:
Hopefully at least one of @michaelgrund and @weiji14 can give this PR a final review.

pygmt/src/filter1d.py
weiji14, line 15: `filter` is a built-in function in Python, see https://docs.python.org/3/library/functions.html#filter. Sure we want to use this? Or use a parameter name like `filter_type` instead?

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+        required filter width at each output location via interpolation.  For
+        multi-segment data files the filter file must either have the same
+        number of segments or just a single segment to be used for all data
+        segments.
+
+    end : bool
+        Include Ends of time series in output. The default [False] loses
+        half the filter-width of data at each end.
+
+    time_col : int or str
```
weiji14, line 84: Is a `str` allowed for time_col (`-N`)? As in users can pass in the name of the time column in the `pandas.DataFrame` table input?

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+        variable filter widths by passing *width* as a two-column time-series
+        file with filter widths in the second column.  The filter-width file
+        does not need to be co-registered with the data as we obtain the
+        required filter width at each output location via interpolation.  For
+        multi-segment data files the filter file must either have the same
+        number of segments or just a single segment to be used for all data
+        segments.
+
+    end : bool
+        Include Ends of time series in output. The default [False] loses
```
weiji14, line 81:
```suggestion
        Include ends of time series in output. The default [False] loses
```

pygmt/src/filter1d.py
michaelgrund: Agree with @weiji14.

michaelgrund (approved):

pygmt/src/filter1d.py
```diff
@@ -0,0 +1,144 @@
...
+        raise GMTInvalidInput("Must specify outfile for ASCII output.")
+
+    with GMTTempFile() as tmpfile:
+        with Session() as lib:
+            file_context = lib.virtualfile_from_data(check_kind="vector", data=data)
+            with file_context as infile:
+                if outfile is None:
+                    outfile = tmpfile.name
+                lib.call_module(
+                    "filter1d", build_arg_string(kwargs, infile=infile, outfile=outfile)
```
seisman, line 133: Following the style in #1910.
```suggestion
                    module="filter1d", 
                    args=build_arg_string(kwargs, infile=infile, outfile=outfile),
```

pygmt/src/filter1d.py
willschlitzer: That makes sense to me; I'll make the change

pygmt/src/filter1d.py
willschlitzer: I am under the impression that GMT reads a string containing an integer or floating point number as the same value as if it was passed as a number instead. I don't know why a Python user would ever pass a number as a string, so it may just be easier to leave it as `int`.

I'll see if I can try and test it, but `filter1d` is one (of the many) module that I don't understand its application and just wrapped using the provided GMT example and dataset.

seisman: > As in users can pass in the name of the time column in the `pandas.DataFrame` table input?

I think this is a nice feature and can be implemented in a separate PR. In this initial wrapper, a time_col as `int` should be good.

weiji14 (approved):
🚀

PR #1494, opened 2021-09-12
Remove deprecated codecov dependency from CI

weiji14:
**Description of proposed changes**

The `codecov` python package is deprecated in favour of a bash script, so it should be removed from PyGMT's CI dependency lists. See https://github.com/codecov/codecov-python/pull/325 and https://about.codecov.io/blog/codecov-uploader-deprecation-plan/ for more information.

Note that PyGMT's code coverage reports are currently generated using [`pytest-cov`](https://github.com/pytest-dev/pytest-cov) (which produces a `coverge.xml` file), with the reported uploaded to https://app.codecov.io/gh/GenericMappingTools/pygmt/ using [`codecov-action`](https://github.com/codecov/codecov-action). So there hasn't been any real use of installing the `codecov` PyPI package on local development environments.

Closes #1482

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1518, opened 2021-09-17
Add tests for load_dataarray to increase code coverage

weiji14:

Patches missing code coverage of `pygmt.load_dataarray` function wrapped in #1439. This commit was cherry-picked from test_io_load_dataarray in #1494.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_io.py
+++ b/pygmt/tests/test_io.py
@@ -1,0 +1,34 @@
+"""
+Tests for input/output (I/O) utilities.
+"""
+import numpy as np
+import pytest
+import xarray as xr
+from pygmt.helpers import GMTTempFile
+from pygmt.io import load_dataarray
+
+
+def test_io_load_dataarray():
+    """
+    Check that load_dataarray works to read a NetCDF grid with
+    GMTDataArrayAccessor information loaded.
+    """
+    with GMTTempFile(suffix=".nc") as tmpfile:
+        grid = xr.DataArray(
+            data=np.random.rand(2, 2), coords=[[0.1, 0.2], [0.3, 0.4]], dims=("x", "y")
+        )
+        grid.to_netcdf(tmpfile.name)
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_io.py
```diff
@@ -0,0 +1,34 @@
...
+        grid = xr.DataArray(
+            data=np.random.rand(2, 2), coords=[[0.1, 0.2], [0.3, 0.4]], dims=("x", "y")
+        )
+        grid.to_netcdf(tmpfile.name)
+        dataarray = load_dataarray(tmpfile.name)
+        assert dataarray.gmt.gtype == 0  # Cartesian grid
+        assert dataarray.gmt.registration == 1  # Pixel registration
+        # this would fail if we used xr.open_dataarray instead of
+        # load_dataarray
+        dataarray.to_netcdf(tmpfile.name)
```
meghanrjones, line 26: Can you clarify why we need this test? To me, it seems redundant with the other tests that read NetCDF grids with GMTDataArrayAccessor information loaded with the exception of the `to_netcdf` line, but that is xarray functionality rather than pygmt.

weiji14: This test was based on `xarray`'s similar test at https://github.com/pydata/xarray/blob/befd1b98bd84047d62307419a30bcda7a0727926/xarray/tests/test_backends.py#L3729-L3736. I agree that this doesn't increase test coverage and seems redundant, but I feel like we should still have an explicit unit test for `load_dataarray` here, rather than just testing it indirectly in other functions.

meghanrjones: Sounds good

meghanrjones (approved):

seisman (approved):

michaelgrund (approved):

Issue #1268, opened 2021-05-10
Add a tutorial explaining the generally accepted input types


seisman:
_Inspired by PR #1186._

For GMT CLI, there are mainly two types of input: a table or a grid. PyGMT is more powerful and can accept more input types. 

For a table input, PyGMT functions generally accept the following input types:

- a file name to a ASCII table, e.g., `data="input.dat"`
- a 2d numpy array, pandas.DataFrame, e.g., `data=data`
- a single value of int or float or str type to `x`,`y`, e.g., `x=5.0, y=5.0`
- a list to `x` and `y`, e.g., `x=[1.0, 2.0], y=[3.0, 4.0]`
- a 1d numpy array to `x` and `y`, e.g., `x=np.array([1.0, 2.0]), y=np.array([3.0, 4.0])`
- a [`geopandas.GeoDataFrame`](https://geopandas.org/docs/reference/api/geopandas.GeoDataFrame.html#geopandas.GeoDataFrame) (see #608 & #1000)
- an [`xarray.Dataset`](https://xarray.pydata.org/en/v0.19.0/generated/xarray.load_dataarray.html)
- `io.StringIO` (once #576 is done)
- more?

For a grid input, PyGMT functions generally accept the following input types:

- a file name to a grid file, e.g., `grid="input_grid.nc"`
- xarray.DataArray, e.g., `grid=grid`
- more?

Although these various input data types are already used in different tutorials and examples, I think it would be better to have a standalone tutorial, explaining these input types.

weiji14:
Also should mention that plotting shapefiles directly using `plot` and `plot3d` (maybe others too?) is supported.

seisman:
`pathlib.Path` support was added in https://github.com/GenericMappingTools/pygmt/pull/1382.

weiji14:
Working on supporting 3-band `xarray.DataArray` inputs at #2590, this would be in addition to regular 1-band `xarray.DataArray` inputs (e.g. DEMs).

PR #1523, opened 2021-09-19
Let required_z check work on xarray.Dataset inputs

weiji14:

This PR modifies the if-statement check in `data_kind` helper function to use `len(data.data_vars)` to check the shape of [`xarray.Dataset`](https://xarray.pydata.org/en/stable/generated/xarray.Dataset.html#xarray.Dataset) inputs. Added two regression tests in test_clib and test_blockm to ensure this works on both the low-level clib and high-level module APIs.

_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/1478#discussion_r711649040_

> Checking for `data.shape[1]` fails for [xarray.Dataset](https://xarray.pydata.org/en/v0.18.1/generated/xarray.Dataset.html) inputs with `AttributeError: 'Dataset' object has no attribute 'shape'. The `shape` attribute is valid only for `numpy.ndarray`, `pandas.DataFrame`, and maybe a few other PyData objects. Will need to create a bugfix and add some extra tests for `xarray.Dataset` inputs.

For a list of data formats supported by PyGMT, see #1268.

Patches #1478, addressing https://github.com/GenericMappingTools/pygmt/pull/1478#discussion_r711649040.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1375,6 +1375,8 @@ class Session:
         extra_arrays : list of 1d arrays
             Optional. A list of numpy arrays in addition to x, y and z. All
             of these arrays must be of the same size as the x/y/z arrays.
+        required_z : bool
+            State whether the 'z' column is required. [Default is False].
 
         Returns
         -------
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1376,7 +1376,7 @@ class Session:
             Optional. A list of numpy arrays in addition to x, y and z. All
             of these arrays must be of the same size as the x/y/z arrays.
         required_z : bool
-            State whether the 'z' column is required. [Default is False].
+            State whether the 'z' column is required.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


pygmt/clib/session.py
```diff
@@ -1375,6 +1375,8 @@ def virtualfile_from_data(
         extra_arrays : list of 1d arrays
             Optional. A list of numpy arrays in addition to x, y and z. All
             of these arrays must be of the same size as the x/y/z arrays.
+        required_z : bool
+            State whether the 'z' column is required. [Default is False].
```
seisman, line 1379: I don't think we need to document the default values for these Python parameters, because the default values are already indicated in the function definitions.
```suggestion
            State whether the 'z' column is required.
```

pygmt/helpers/utils.py
```diff
@@ -39,8 +39,9 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     x/y : 1d arrays or None
         x and y columns as numpy arrays.
     z : 1d array or None
-        z column as numpy array. To be used optionally when x and y
-        are given.
+        z column as numpy array. To be used optionally when x and y are given.
+    required_z : bool
+        State whether the 'z' column is required. [Default is False].
```
seisman, line 44:
```suggestion
        State whether the 'z' column is required.
```

seisman (approved):

PR #1526, opened 2021-09-19
Allow non-string color when input data is a matrix or a file for plot and plot3d

seisman:

The following script raises an error with PyGMT v0.4.1 and the master branch:

```
import pygmt
fig = pygmt.Figure()
fig.plot(data=[[3, 3], [5, 5]], frame=True, region=[0, 10, 0, 10], style="c1c", color=200)
fig.show()
```
because the following code is too strict about the `color` type.

https://github.com/GenericMappingTools/pygmt/blob/f7d55269aac7585e1d1cb9b10abae70664736af8/pygmt/src/plot.py#L242-L248

This PR fixed the check to make sure that types like string, float, boolean or None can be used.

**Reminders**

**Slash Commands**


pygmt/tests/test_plot.py
```diff
@@ -302,12 +302,13 @@ def test_plot_sizes_colors_transparencies():
     return fig
 
 
-@pytest.mark.mpl_image_compare
+@pytest.mark.mpl_image_compare(filename="test_plot_matrix.png")
+@pytest.mark.parametrize("color", ["#aaaaaa", 170])
```
weiji14, line 306: Should we set `G="sequence"` so that people can input RGB or CMYK colors as a list like `[170, 170, 170]` or [0, 0, 0, 33] which gets converted to `170/170/170` and `0/0/0/33` respectively? Ref https://docs.generic-mapping-tools.org/6.2/gmtcolors.html#description.

weiji14 (approved):

PR #1530, opened 2021-09-21
Raise RuntimeWarning instead of an exception for irregular grid spacing

### LINKED ISSUE #1468: PyGMT is too strict when checking for irregular grids

Since GMT only supports regular grids, PyGMT has a check to make sure all point spacings are the same ([here](https://github.com/GenericMappingTools/pygmt/blob/8a6db70f1a9786007705a3888300028af4044de1/pygmt/clib/conversion.py#L97)):

```python
coord = grid.coords[dim].values
coord_incs = coord[1:] - coord[0:-1]
coord_inc = coord_incs[0]
if not np.allclose(coord_incs, coord_inc):
    raise GMTInvalidInput(
        "Grid appears to have irregular spacing in the '{}' dimension.".format(
            dim
        )
    )
```

The problem is that the `np.allclose` function has variable precision arguments to determine whether it fails or passes. So the degree to which we check for uniform spacing depends on coordinate units. This is not ideal as this check can fail in cases where the grid is evenly spaced within an acceptable accuracy for coordinates but maybe not for `allclose`. 

I'm not sure there is a good solution here since:

1. This will vary with units so it's hard to generalize.
2. This part of the code is deep in the internals so getting the `rtol` and `atol` arguments from the user into this part of the code will be awkward and bug prone.
3. Should this really be an exception if GMT may be OK with the coordinates?

I would propose turning this into a warning instead and let GMT fail if it really thinks it can't handle the grid. What do others think? I'm not sure that would solve the problem or if GMT will be equally strict.

```python
grid = xr.open_dataarray("topography-earth-10arcmin.nc")
fig = pygmt.Figure()
fig.grdimage(grid)
fig.show()
```

Using this grid file: [topography-earth-10arcmin.nc.zip](https://github.com/GenericMappingTools/pygmt/files/7065984/topography-earth-10arcmin.nc.zip) (GitHub required a zip archive for this)

For context, this grid is coming from the [ICGEM](http://icgem.gfz-potsdam.de/home) calculation service. I tracked down the error to them using the equivalent of `np.arange` to generate the grid coordinates instead of `np.linspace` (which is the correct way to do it). 

**Full error message**

<details>

```
/tmp/ipykernel_207888/3701624484.py in <module>
      1 fig = pygmt.Figure()
----> 2 fig.grdimage(grid)
      3 fig.show()

~/bin/conda/envs/rockhound-data/lib/python3.9/site-packages/pygmt/helpers/decorators.py in new_module(*args, **kwargs)
    580                     )
    581                     warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
--> 582             return module_func(*args, **kwargs)
    583 
    584         new_module.aliases = aliases

~/bin/conda/envs/rockhound-data/lib/python3.9/site-packages/pygmt/helpers/decorators.py in new_module(*args, **kwargs)
    723                         kwargs[arg] = separators[fmt].join(f"{item}" for item in value)
    724             # Execute the original function and return its output
--> 725             return module_func(*args, **kwargs)
    726 
    727         return new_module

~/bin/conda/envs/rockhound-data/lib/python3.9/site-packages/pygmt/src/grdimage.py in grdimage(self, grid, **kwargs)
    171                 kwargs["I"] = stack.enter_context(shading_context)
    172 
--> 173             fname = stack.enter_context(file_context)
    174             arg_str = " ".join([fname, build_arg_string(kwargs)])
    175             lib.call_module("grdimage", arg_str)

~/bin/conda/envs/rockhound-data/lib/python3.9/contextlib.py in enter_context(self, cm)
    427         _cm_type = type(cm)
    428         _exit = _cm_type.__exit__
--> 429         result = _cm_type.__enter__(cm)
    430         self._push_cm_exit(cm, _exit)
    431         return result

~/bin/conda/envs/rockhound-data/lib/python3.9/contextlib.py in __enter__(self)
    115         del self.args, self.kwds, self.func
    116         try:
--> 117             return next(self.gen)
    118         except StopIteration:
    119             raise RuntimeError("generator didn't yield") from None

~/bin/conda/envs/rockhound-data/lib/python3.9/site-packages/pygmt/clib/session.py in virtualfile_from_grid(self, grid)
   1344         # guarantees that the copy will be around until the virtual file is
   1345         # closed. The conversion is implicit in dataarray_to_matrix.
-> 1346         matrix, region, inc = dataarray_to_matrix(grid)
   1347 
   1348         family = "GMT_IS_GRID|GMT_VIA_MATRIX"

~/bin/conda/envs/rockhound-data/lib/python3.9/site-packages/pygmt/clib/conversion.py in dataarray_to_matrix(grid)
     96         coord_inc = coord_incs[0]
     97         if not np.allclose(coord_incs, coord_inc):
---> 98             raise GMTInvalidInput(
     99                 "Grid appears to have irregular spacing in the '{}' dimension.".format(
    100                     dim

GMTInvalidInput: Grid appears to have irregular spacing in the 'longitude' dimension.
```

</details>

**System information**

```
PyGMT information:
  version: v0.4.1
System information:
  python: 3.9.0 | packaged by conda-forge | (default, Nov 26 2020, 07:57:39)  [GCC 9.3.0]
  executable: /home/leo/bin/conda/envs/rockhound-data/bin/python
  machine: Linux-5.13.11-1-MANJARO-x86_64-with-glibc2.33
Dependency information:
  numpy: 1.21.2
  pandas: 1.3.2
  xarray: 0.19.0
  netCDF4: 1.5.7
  packaging: 21.0
  ghostscript: 9.54.0
  gmt: 6.2.0
GMT library information:
  binary dir: /home/leo/bin/conda/envs/rockhound-data/bin
  cores: 8
  grid layout: rows
  library path: /home/leo/bin/conda/envs/rockhound-data/lib/libgmt.so
  padding: 2
  plugin dir: /home/leo/bin/conda/envs/rockhound-data/lib/gmt/plugins
  share dir: /home/leo/bin/conda/envs/rockhound-data/share/gmt
  version: 6.2.0
```

seisman:

GMT only supports regular grid spacing. So when passing a grid to GMT API, it's always assumed that the grid has regular spacings in x and y dimensions.

When irregular grid spacings are detected, we raised an exception, but the detection may be too strict (see #1468 for the issue report). In this PR, a warning is raised, instead of an exception.

Fixes #1468

**Reminders**

**Slash Commands**


pygmt/clib/conversion.py
```diff
@@ -95,9 +97,12 @@ def dataarray_to_matrix(grid):
         coord_incs = coord[1:] - coord[0:-1]
         coord_inc = coord_incs[0]
         if not np.allclose(coord_incs, coord_inc):
```
weiji14, line 99: Could we try to work out the [`np.assert_allclose`](https://numpy.org/doc/1.21/reference/generated/numpy.allclose.html) atol or rtol argument values which won't matter for GMT? I remember reading somewhere that GMT uses [single-precision (4-byte/32-bit) floating point](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) to store the grids (see https://github.com/GenericMappingTools/gmt/blob/32296e512b173cfdd2c3e8da1d2735e49b79f02a/doc/rst/source/explain_float.rst_#L5-L7). So how many decimal places is that?

pygmt/clib/conversion.py
```diff
@@ -95,9 +97,12 @@ def dataarray_to_matrix(grid):
         coord_incs = coord[1:] - coord[0:-1]
         coord_inc = coord_incs[0]
```
weiji14, line 98: If the coordinate increments can be slightly different, is it safe to just take the first one, or should we take the mean/average value (inside the `if not np.allclose(coord_incs, coord_inc):` block)?

pygmt/clib/conversion.py
seisman: Although GMT grids are stored using 4-bytes, the grid metadata (range, inc, and others) are stored using double-precision floating points (8 bytes) (see https://github.com/GenericMappingTools/gmt/blob/8ee9060d8620a7d0e599e3f4990303b1517aed02/src/gmt_resources.h#L413).

I'm looking at @leouieda's example data in #1468. The "longitude" dimension has the following values:
```python
>>> grid.coords["longitude"].values
array([-180.    , -179.8333, -179.6667, ...,  179.6667,  179.8333,
        180.    ], dtype=float32)
```
As you can see, the grid spacings can be either 0.1666 or 0.1667 even if we ignore the floating-point precision issue.

pygmt/clib/conversion.py
seisman: Yes, using the mean spacing sounds better.

seisman: Or we can calculate the increment from the minimum, maximum and length of the coordinates? Take the coordinates in https://github.com/GenericMappingTools/pygmt/pull/1530#discussion_r713752741 as an example, the array `grid.coords["longitude"].values` has a size of 2161, and the increment is (180-(-180))/(2161-1) = 0.1666666...

pygmt/tests/test_clib.py
```diff
@@ -808,16 +808,17 @@ def test_dataarray_to_matrix_dims_fails():
         dataarray_to_matrix(grid)
 
 
-def test_dataarray_to_matrix_inc_fails():
+def test_dataarray_to_matrix_irregular_inc_warning():
     """
-    Check that it fails for variable increments.
+    Check that it warns for variable increments.
```
weiji14, line 813: Maybe link to the issue at #1468 so people know why this is being tested like this.
```suggestion
    Check that it warns for variable increments, see also
    https://github.com/GenericMappingTools/pygmt/issues/1468.
```

weiji14 (commented):
Would prefer if @leouieda could give this a review and see if it fixes his problem at #1468.

pygmt/clib/conversion.py
seisman: > Or we can calculate the increment from the minimum, maximum and length of the coordinates? Take the coordinates in [#1530 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1530#discussion_r713752741) as an example, the array `grid.coords["longitude"].values` has a size of 2161, and the increment is (180-(-180))/(2161-1) = 0.1666666...

Implemented in d5d98c8c39f225c361a395a4d1ba275724dcdf7c.

weiji14 (approved):

leouieda:
Hi everyone, sorry just got to this now. I tried out the latest version on `main` and it seems to be working great! Thank you for implementing this @seisman and @weiji14 ❤️

Issue #1479, opened 2021-09-01
Standardize parameter name for table data inputs


weiji14:
**Description of the desired feature**

_Originally posted by @meghanrjones in https://github.com/GenericMappingTools/pygmt/issues/1441#issuecomment-901303058_

> This is somewhat related - I am not sure if it's worth a deprecation but I think we should decide on whether the parameter for tabular input is `data` or `table` for all modules moving forward. Currently, this is the distribution:
> 
> `data` parameter: surface, wiggle, velo, rose, plot3d, plot, contour
> `table` parameter: blockmedian, blockmean, histogram
> other: grdtrack (`points`), text (`textfiles`)
> 
> It would also be worthwhile to set a consistent order for the positional arguments x, y, [z], data, [outfile]

**Are you willing to help implement and maintain this feature?** Discuss first

weiji14:
One option to reduce the amount of deprecation notices is to keep two parameters, but with this rule:

- `data` - use for the `fig.*` plotting functions (e.g. `fig.plot`, `fig.contour`, etc)
- `table` - use for the data processing functions (e.g. `blockmean`, `histogram`, etc)

This way, we'll only need to do the renaming for `surface` ('data' -> 'table'), `grdtrack` ('points' -> 'table') and `text` ('textfiles' -> 'data'). I have some reservations with renaming `text`'s 'textfiles' to 'data' since it's a bit non-standard (and need to wait for #1121), but otherwise, what do people think?

meghanrjones:
> One option to reduce the amount of deprecation notices is to keep two parameters, but with this rule:
> 
>     * `data` - use for the `fig.*` plotting functions (e.g. `fig.plot`, `fig.contour`, etc)
> 
>     * `table` - use for the data processing functions (e.g. `blockmean`, `histogram`, etc)
> 
> 
> This way, we'll only need to do the renaming for `surface` ('data' -> 'table'), `grdtrack` ('points' -> 'table') and `text` ('textfiles' -> 'data'). I have some reservations with renaming `text`'s 'textfiles' to 'data' since it's a bit non-standard (and need to wait for #1121), but otherwise, what do people think?

One complication is that many modules can be plotting or processing (e.g., `histogram` is mostly plotting but can return statistics with **-I**). My preference would be to use `data` moving forward and, if needed, deprecate `table` to `data` in `blockmean`, `blockmedian`, and `histogram`. `grdtrack` and `text` are non-standard enough for different names in my opinion.

[...]

PR #1541, opened 2021-09-24
Add a decorator to raise a FutureWarning when positional arguments are passed for functions that changed parameter orders

seisman:

See https://github.com/GenericMappingTools/pygmt/issues/1479#issuecomment-920320607 for discussions.

This PR adds a decorator to raise a warning when the order of parameters are changed and users are passing positional arguments.

Usage:
```python
@fmt_docstring
@deprecate_parameter("sizes", "sizes", "v0.4.0", remove_version="v0.6.0")
@check_data_input_order("v0.5.0", remove_version="v0.7.0")
@use_alias(...)
@kwargs_to_strings(...)
def plot(self, data=None, x=None, y=None, ....):
    pass
```

Address #1479.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -811,3 +811,61 @@ def deprecate_parameter(oldname, newname, deprecate_version, remove_version):
         return new_module
 
     return deprecator
+
+
+def check_data_input_order(deprecate_version, remove_version):
+    """
+    Decorator to raise a FutureWarning if the order of data input parameters
+    changes and positional arguments are passed.
+
+    Parameters
+    ----------
+    deprecate_version : str
+        The PyGMT version when the order of data input parameters is changed.
+    remove_version : str
+        The PyGMT version when the deprecation warning should be removed.
+
+    Examples
+    --------
+    >>> @check_data_input_order("v0.0.0", "v9.9.9")
+    ... def module(data=None, x=None, y=None, z=None, **kwargs):
+    ...     "A module that prints the arguments it received"
+    ...     print(f"data={data}, x={x}, y={y}, z={z}")
+    >>> module(data="table.txt")
+    data=table.txt, x=None, y=None, z=None
+    >>> module(x=0, y=1, z=2)
+    data=None, x=0, y=1, z=2
+    >>> with warnings.catch_warnings(record=True) as w:
+    ...     module(0, 1, 2)
+    ...     assert len(w) == 1
+    ...     assert issubclass(w[0].category, FutureWarning)
+    ...
+    data=0, x=1, y=2, z=None
+    """
+
+    def data_input_order_checker(module_func):
+        """
+        The decorator that creates the new function to check if positional
+        arguments are passed.
+        """
+
+        @functools.wraps(module_func)
+        def new_module(*args, **kwargs):
+            """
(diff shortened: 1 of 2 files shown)


seisman:
@pygmt-maintainers Need to review/merge this PR before we can change the order of input parameters (e.g., #1547, #1546).

michaelgrund:
> 
> 
> @GenericMappingTools/pygmt-maintainers Need to review/merge this PR before we can change the order of input parameters (e.g., #1547, #1546).

Sorry, totally overlooked this PR before opening #1547 and #1546.

michaelgrund (approved):
Looks good to me!

pygmt/helpers/decorators.py
```diff
@@ -811,3 +811,61 @@ def new_module(*args, **kwargs):
...
+        arguments are passed.
+        """
+
+        @functools.wraps(module_func)
+        def new_module(*args, **kwargs):
+            """
+            New module instance that raises a warning if positional arguments
+            are passed.
+            """
+            if len(args) != 0:  # positional arguments are used
```
weiji14, line 858: Don't think args can be negative, but this looks better I think :slightly_smiling_face:
```suggestion
            if len(args) > 0:  # positional arguments are used
```

weiji14 (commented):
I realize that this is a temporary decorator, which might not have much use after PyGMT v0.7.0, but should we:

1) Add a usage note for `@check_data_input_order` at https://github.com/GenericMappingTools/pygmt/blob/main/doc/maintenance.md#backwards-compatibility-and-deprecation-policy
2) Add a note in the docstring of this decorator function to say that this function can be removed after PyGMT v0.7.0?

seisman:
> I realize that this is a temporary decorator, which might not have much use after PyGMT v0.7.0, but should we:
> 
> 1. Add a usage note for `@check_data_input_order` at [`main`/doc/maintenance.md#backwards-compatibility-and-deprecation-policy](https://github.com/GenericMappingTools/pygmt/blob/main/doc/maintenance.md#backwards-compatibility-and-deprecation-policy)

I prefer not to add the usage, because it's likely that no one will use this decorator again after we finish issue #1479 (definitely before v0.5.0 release). 

> 2. Add a note in the docstring of this decorator function to say that this function can be removed after PyGMT v0.7.0?

Yes, it sounds good.

pygmt/helpers/decorators.py
```diff
@@ -811,3 +811,61 @@ def new_module(*args, **kwargs):
         return new_module
 
     return deprecator
+
+
+def check_data_input_order(deprecate_version, remove_version):
+    """
+    Decorator to raise a FutureWarning if the order of data input parameters
+    changes and positional arguments are passed.
```
seisman, line 819:
```suggestion
    changes and positional arguments are passed.
    
    The decorator is temporary and should be removed in v0.7.0.
```

weiji14 (approved):
Ok to merge once you fix the lint issue. Some whitespace I think.

seisman:
The decorator works well for non-plotting modules, but doesn't work for plotting modules (e.g., #1548, #1547), because for plotting modules, the first argument is `self`, thus, the length of the `args` array is at least 1. 

Perhaps we have to check if the function is plotting or not.

Issue #1309, opened 2021-05-28
grdtrack produces incorrect results for xarray.DataArray grid


seisman:
**Description of the problem**

grdtrack works well for netCDF file, but produces random results for xarray.DataArray grid. If I understand #1304 correctly, it works on Linux, but not on macOS and Windows. I believe it's an upstream GMT bug.

**Full code that generated the error**

```python
from pygmt import grdtrack, grdcut
from pygmt.datasets import load_earth_relief

# load the earth_relief_01d_g file
grid = load_earth_relief(registration="gridline")

# prepare a subgrid in xarray.DataArray
dataarray = grid.sel(lat=slice(-49, -42), lon=slice(-118, -107))
# prepare a subgrid in netcdf
grdcut(grid, region=[-118, -107, -49, -42], outgrid="earth_subset.nc")

# run grdtrack on the netCDF file
output = grdtrack(points="@ridge.txt", grid="earth_subset.nc", newcolname="bathymetry")
print(output)

# run grdtrack on the xarray.DataArray
output = grdtrack(points="@ridge.txt", grid=dataarray, newcolname="bathymetry")
print(output)
```

**Full output**

The first output is correct, and the second output is wrong.
```
grdtrack [WARNING]: Some input points were outside the grid domain(s).
          0        1            2
0 -110.9536 -42.2489 -2797.394987
1 -111.3500 -42.2033 -2715.359945
2 -111.7309 -43.1706 -2711.704748
3 -112.2196 -44.3819 -2787.948153
4 -112.5505 -45.4878 -2746.303734
5 -112.6736 -45.8347 -2685.852247
6 -112.4235 -45.8561 -2785.099618
7 -112.9507 -47.2270 -2672.932596
8 -113.3124 -48.3208 -2706.345811
grdtrack [WARNING]: Some input points were outside the grid domain(s).
          0        1             2
0 -110.9536 -42.2489 -2.974656e+03
1 -111.3500 -42.2033 -2.859961e+03
2 -111.7309 -43.1706 -2.711705e+03
3 -112.2196 -44.3819 -2.787948e+03
4 -112.5505 -45.4878 -2.746304e+03
5 -112.6736 -45.8347 -2.685852e+03
6 -112.4235 -45.8561 -2.785100e+03
7 -112.9507 -47.2270 -2.672933e+03
8 -113.3124 -48.3208  1.123369e+28
```

**System information**

[...]

PR #1608, opened 2021-11-05
Mark two grdtrack tests as xfail due to upstream GMT 6.3 fixes

weiji14:

Using correct output values for `grdtrack` with bicubic interpolation on `xarray.DataArray` grids on GMT 6.3.

Fixes https://github.com/GenericMappingTools/pygmt/issues/1309#issuecomment-953452875, Supersedes #1587.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_grdtrack.py
+++ b/pygmt/tests/test_grdtrack.py
@@ -6,6 +6,7 @@ import os
 import numpy.testing as npt
 import pandas as pd
 import pytest
+from packaging.version import Version
 from pygmt import clib, grdtrack, which
 from pygmt.datasets import load_earth_relief, load_ocean_ridge_points
 from pygmt.exceptions import GMTInvalidInput
--- a/pygmt/tests/test_grdtrack.py
+++ b/pygmt/tests/test_grdtrack.py
@@ -6,7 +6,7 @@ import os
 import numpy.testing as npt
 import pandas as pd
 import pytest
-from pygmt import grdtrack, which
+from pygmt import clib, grdtrack, which
 from pygmt.datasets import load_earth_relief, load_ocean_ridge_points
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import data_kind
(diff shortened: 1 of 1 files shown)


meghanrjones:
In this case the result is updated and the test is marked as xfail with a condition of <= GMT 6.X but for others the current result is kept and the test is marked as xfail with a condition of > GMT 6.X, for example: https://github.com/GenericMappingTools/pygmt/blob/5dc135b3d789c842c959c9abbdb505b07011a68e/pygmt/tests/test_velo.py#L74-L95 I would prefer to use the same strategy for all tests when tracking failures corresponding to upstream commits.

weiji14:
> In this case the result is updated and the test is marked as xfail with a condition of <= GMT 6.X but for others the current result is kept and the test is marked as xfail with a condition of > GMT 6.X, for example:
> 
> I would prefer to use the same strategy for all tests when tracking failures corresponding to upstream commits.
> 
> Edit: The steps in this PR are preferable IMO because updating the tests and marking as xfail for <= GMT 6.X makes it easier to test for regressions in the GMT dev branch between releases.

Agree that using <= GMT6.X is better, I was actually following this:

https://github.com/GenericMappingTools/pygmt/blob/5dc135b3d789c842c959c9abbdb505b07011a68e/pygmt/tests/test_plot.py#L307-L311

We can remove those xfails on the unit tests for `velo` once we bump the minimum GMT version to 6.3, hopefully in the next few weeks. I'm just fixing these 2 grdtrack tests temporarily because they're breaking the ci_tests_dev.yaml workflow.

pygmt/tests/test_grdtrack.py
```diff
@@ -6,14 +6,17 @@
 import numpy.testing as npt
 import pandas as pd
 import pytest
```
meghanrjones, line 8:
```suggestion
import pytest
from packaging.version import Version
```

weiji14: Thanks, seems like I need some coffee today :sweat_smile:

meghanrjones (approved):

Issue #573, opened 2020-08-31
Document how to parse verbose messages to report equivalent GMT commands


seisman:
**Description of the desired feature**

Sometimes we may need the equivalent GMT commands. For example, 

1. some PyGMT issues are actually due to upstream GMT bugs (https://forum.generic-mapping-tools.org/t/829). When we open a GMT bug, we need the equivalent bash script
2. currently, how PyGMT works is a little fragile. Any unknown arguments are directly passed to GMT. See a bug report #256. The error messages are really confusing. But if the equivalent CLI commands are printed out, it would be much easier to understand why it doesn't work as expected.

GMT already provides 7 verbosity levels by [the **-V** option](https://docs.generic-mapping-tools.org/dev/gmt.html#v-full). The 7 verbosity levels (higher levels add more messages) are:
```
q < e < w < t < i < c < d
```
 Here I propose to add one more verbosity level, for example, `verbose='p'`. The verbosity level should be between **q** and **e**, i.e.,

```
q <   < e < w < t < i < c < d
    ^
    |
    p
```
If `verbose='p'` is used, they PyGMT should print the equivalent GMT command when calling the function `call_module()`.

I only spent a few seconds to choose the character `p` and its verbosity level. So I'm open to better choices. 

**Are you willing to help implement and maintain this feature?** 

Yes, it should be easy to implement it in the [`call_module()`](https://github.com/GenericMappingTools/pygmt/blob/fc246e7aa0a136250136495b0259c220a00785f0/pygmt/clib/session.py#L468) function. Contributions are welcomed.

weiji14:
I think the intention is good, it does take a bit of experience translating PyGMT Python commands to GMT bash commands, and any way we can simplify this to improve bug reports will be a plus.

That said, I'm not sure if this should live in the PyGMT library. Presumably other GMT wrappers like GMT.jl and GMTmex might find it useful too?

seisman:
Ping @PaulWessel @joa-quim for thoughts on this.

PaulWessel:
First reaction is that it sounds like a debug option, hence might be part of -Vd, but I am guessing you want to just see those commands.  Of course, it makes no sense from the command line GMT to document and explain p since we already have the commands.  So, first to check, might this go into -Vd as is?

weiji14:
How about just for `-Vi` at the informative level?

PaulWessel:
[...]

PR #1610, opened 2021-11-05
Add instructions for reporting upstream bugs to contributing.md

meghanrjones:

Add a section to the reporting a bug section of the contributing guide with steps to provide helpful information about upstream bugs.

Fixes https://github.com/GenericMappingTools/pygmt/issues/573

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -62,12 +62,12 @@ the problem:
   filter the messages to include only the GMT-equivalent commands using a command
   such as:
 
-      python <test>.py 2>&1 | grep GMT_Call_Command | awk -F': ' '{print "gmt", $3}'
+      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
 
   where `<test>` is the name of your test script.
 * If the bug is produced when passing an in-memory data object (e.g., a
   pandas.DataFrame or xarray.DataArray) to a PyGMT function, try writing the
-  data to a file (e.g., an ASCII file or NetCDF grid) and passing the data file
+  data to a file (e.g., a NetCDF or ASCII txt file) and passing the data file
   to the PyGMT function instead. In the GitHub issue, please share the results
   for both cases along with your code.
 
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -50,6 +50,27 @@ top of the GitHub repository and click *New Issue*.
 * After submitting your bug report, try to answer any follow up questions about the bug
   as best as you can.
 
+#### Reporting upstream bugs
+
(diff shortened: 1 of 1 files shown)


doc/contributing.md
```diff
@@ -50,6 +50,27 @@ top of the GitHub repository and click *New Issue*.
...
+PyGMT-specific issue, you can optionally take the following steps to help resolve
+the problem:
+
+* Add the line `pygmt.config(GMT_VERBOSE='d')` after your import statements, which
+  will report the equivalent GMT commands as one of the debug messages.
+* Either append all messages from running your script to your GitHub issue, or
+  filter the messages to include only the GMT-equivalent commands using a command
+  such as:
+
+      python <test>.py 2>&1 | grep GMT_Call_Command | awk -F': ' '{print "gmt", $3}'
```
seisman, line 65: What about this command? It doesn't call `grep`.
```suggestion
      python test.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
```

seisman: BTW, this command works for both Linux (GNU awk) and macOS (BSD awk).

doc/contributing.md
```diff
@@ -50,6 +50,27 @@ top of the GitHub repository and click *New Issue*.
...
+
+  where `<test>` is the name of your test script.
+* If the bug is produced when passing an in-memory data object (e.g., a
+  pandas.DataFrame or xarray.DataArray) to a PyGMT function, try writing the
+  data to a file (e.g., an ASCII file or NetCDF grid) and passing the data file
```
weiji14, line 70: NetCDFs are not necessarily grids (can also be data cubes), so maybe just say NetCDF.
```suggestion
  data to a file (e.g., a NetCDF or ASCII txt file) and passing the data file
```

doc/contributing.md
weiji14: Do we want to have an equivalent command for Windows users? :slightly_smiling_face:

seisman: Does Windows provide any built-in tools to do tasks like awk?

weiji14: A quick search yields this https://superuser.com/questions/524822/awk-equivalent-functionality-on-windows/524836#524836, but not sure if it works with stderr/stdout from GMT.

meghanrjones: I don't really think we need to spend time on this, since it only applies to the subset of people who are voluntarily providing additional information about upstream bugs without WSL or Git for Windows installed. They can always post the full error message or we could extract the equivalent commands on a *nix system.

seisman (approved):

weiji14 (approved):

PR #1611, opened 2021-11-07
Fix missing gcmt convention keys in pygmt.meca

### LINKED ISSUE #1580: Problem with GCMT convention in meca.py

Originally reported by @philcummins at https://forum.generic-mapping-tools.org/t/pygmt-parameter-check-in-meca-py/1874. Copying here to keep track.

> In the parameters section of the documentation block in meca.py, it says that the GCMT convention is determined if a dictionary with the following keys is passed:
> - "gcmt" — strike1, dip1, rake1, strike2, dip2, rake2, mantissa,exponent
> But later in the code, when the parameters to check against are set (line 250-252), it says:
> param_conventions = {
> “AKI”: [“strike”, “dip”, “rake”, “magnitude”],
> “GCMT”: [“strike1”, “dip1”, “dip2”, “rake2”, “mantissa”,“exponent”],
> …
> 
> So it seems if you pass all the parameters it says it requires, it fails with:
> GMTError: Parameters in spec dictionary do not match known conventions.
> 
> BTW, it says a geopandas dataframe an be used to pass in location info. Are there any planes to pass in the focal mechanism parameters in the geopandas dataframe as well?

Solution would be to modify this line I think:

https://github.com/GenericMappingTools/pygmt/blob/5cb1035dd4e8fdaeec1d0421b4cbc58392dff29b/pygmt/src/meca.py#L252

I.e., need to add `"rake1", "strike2"` in the middle of the list.

```python
PASTE CODE HERE
```

**Full error message**

```
PASTE ERROR MESSAGE HERE
```

**System information**

```
PASTE THE OUTPUT HERE
```

weiji14:

Adds the 'rake1' and 'strike2' keys to the global CMT (gcmt) convention, and add a unit test adapted from https://docs.generic-mapping-tools.org/6.2/supplements/seis/meca.html#examples.

Fixes #1580, Patches #516.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -249,7 +249,16 @@ def meca(
 
         param_conventions = {
             "AKI": ["strike", "dip", "rake", "magnitude"],
-            "GCMT": ["strike1", "dip1", "dip2", "rake2", "mantissa", "exponent"],
+            "GCMT": [
+                "strike1",
+                "dip1",
+                "rake1",
+                "strike2",
+                "dip2",
+                "rake2",
+                "mantissa",
+                "exponent",
+            ],
             "MT": ["mrr", "mtt", "mff", "mrt", "mrf", "mtf", "exponent"],
             "PARTIAL": ["strike1", "dip1", "strike2", "fault_type", "magnitude"],
             "PRINCIPAL_AXIS": [
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -110,7 +110,7 @@ def data_format_code(convention, component="full"):
 )
 @kwargs_to_strings(R="sequence", c="sequence_comma", p="sequence")
 def meca(
-    self,  # pylint: disable=unused-argument
+    self,
     spec,
     scale,
     longitude=None,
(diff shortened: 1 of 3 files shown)


seisman (approved):
Looks good to me.

seisman:
Ping @tjnewton and @liamtoney for reviews, since they were the main authors of the original PR #516.

michaelgrund (approved):

pygmt/src/meca.py
```diff
@@ -275,23 +284,21 @@ def update_pointers(data_pointers):
             "plot_latitude": plot_latitude,
         }
 
-        # make a DataFrame copy to check convention if it contains
-        # other parameters
-        if isinstance(spec, (dict, pd.DataFrame)):
```
weiji14, line 280: Just noticed that this if-statement is unnecessary, because this code block is already nested under an `if isinstance(spec, (dict, pd.DataFrame))` statement at L143 above, and there doesn't seem to be any modifications made to the `spec` variable in between. Please double-check if this looks ok.

seisman: Looks good to me.

tjnewton: Nice catch. Looks great

Issue #1466, opened 2021-08-25
Allow specifying the label angle, font, justification, and offset in pygmt.meca


meghanrjones:
**Description of the desired feature**

Based on [this forum post](https://forum.generic-mapping-tools.org/t/pygmt-how-to-change-meca-label-sizes/2023), it would be useful to add support for customizing the label characteristics for meca, which is done using **-S** in [gmt meca](https://docs.generic-mapping-tools.org/dev/supplements/seis/meca.html#s). Relates to https://github.com/GenericMappingTools/pygmt/issues/1082, in order to make the PyGMT implementation more pythonic than gmt meca **-S**. 

**Are you willing to help implement and maintain this feature?** Not likely that I will find time to implement

weiji14:
Note that there was a previous attempt to handle the text label specification. See @tjnewton's implementation at 2fe87af5adb9403ee17d388d63fb14cf351be1fa, but this was reverted by @liamtoney at ff164e654827024e76ec892f3acc2ee0d4c47bef because we didn't support `GMT_Put_Strings` at that time (full explanation at https://github.com/GenericMappingTools/pygmt/pull/516#issuecomment-657313059.

I'll try and resurrect that old bit of code, and put it in #1613. Or, I could implement the text label and angle/font/justification specification in a separate Pull Request after #1613.

Issue #1129, opened 2021-03-26
pygmt.Figure.meca offset not working


rdcorona:
I am using pygmt to graph focal mechanisms with magnitude scale.
However, when I try to give it an offset so that they don't pile up the function doesn't seem to work.

I'm using :

```python
spec = dict (strike=strikes[0], dip=dips[0], rake=rakes[0], magnitude=magnitude [0])

fig.meca (spec, scale=f ".{0.002 * (2 ** magnitude[0])}c", longitude = lons [0], latitude = lats [0], depth = 25.0, convention = "aki", plot_longitude = lons [0], plot_latitude = lats[0] -10, offset=True) 
```
where all the objects are lists with real (float) values

**System information**
```
# Name                    Version                   Build  Channel
pygmt                     0.3.1              pyhd8ed1ab_0    conda-forge

```

Do I make something wrong?

core-man:
I can re-produce this potential bug using [Focal mechanisms](https://www.pygmt.org/dev/gallery/seismology/meca.html#sphx-glr-gallery-seismology-meca-py) gallery example.

```Python
import pygmt
fig = pygmt.Figure()
fig.coast(region=[-125, -122, 47, 49], projection="M6c", land="grey", water="lightblue", shorelines=True, frame="a")
focal_mechanism = dict(strike=330, dip=30, rake=90, magnitude=3)
fig.meca(focal_mechanism, scale="1c", longitude=-124.3, latitude=48.1, depth=12.0, plot_longitude=-124.3, plot_latitude=47.5, offset=True)
fig.show()
```
![meca](https://user-images.githubusercontent.com/50591376/112717373-32942c00-8f27-11eb-934e-d9924d1e983e.png)

seisman:
I can reproduce the issue, and now know why it fails. Unfortunately, there is no workaround now.

rdcorona:
Hello seisman.
So, we'll have to wait for the 0.4.0 release.🤷🏻‍♂️

Thanks for the reply

seisman:
Currently, offsetting beachballs only works for files. 

I'll see what I can do to fix it.

rdcorona:
Ok, meanwhile I'll try it with a dataFrame and an ASCII file 

Thanks again

hazardgoat:
> Ok, meanwhile I'll try it with a dataFrame and an ASCII file

[...]

Issue #1132, opened 2021-03-27
Let PyGMT accept a 1d array when 2d array is needed


seisman:
The issue was first found in https://github.com/GenericMappingTools/pygmt/pull/1070#discussion_r602617399.

Below is a simple example plotting a circular vector. The `data` parameter expects a 2D array/list. Each row is a record, and the 5 columns are `[x_start, y_start, radius, degree_start, degree_stop]`. 

The following script works as expected because the variable `data` is a 2D array/list:
```python
import numpy as np
import pygmt

fig = pygmt.Figure()
fig.plot(
    region=[-10, 10, -10, 10],
    projection="X10c",
    frame="ag",
    data=[[0, 0, 2.5, 90, 270], [0, 0, 1.5, 90, 270]],
    style="m0.5c+ea",
    pen="2p",
    color="red3",
)
fig.show()
```

When users want to plot a single circular vector, they still have to pass a 2D list, so the code looks like:
```
data=[[0, 0, 2.5, 90, 270]]
```
However, most people may first try a 1D array and then realize it doesn't work as expected.
```
data=[0, 0, 2.5, 90, 270]
```

As I understand it, `data` always expects a 2D array/list. When a 1D array is given, can we convert it to a 2D array? Are there any negative effects?

seisman:
Similarly, parameters like `x` and `y` usually expect a 1D array/list, but PyGMT allow passing scalar values to them.

noorbuchi:
Hello! I tried the example you mentioned in the issue description and I got this error message.
```
'list' object has no attribute 'flags'
```
So it looks like even passing a 2D array would cause an error. To solve it, I had to pass the 2D array into a `numpy` array object to see the plot. 
I don't know if it's important to know but I ran the code using Jupyter Notebook. 
```
fig = pygmt.Figure()
fig.plot(
    region=[-10, 10, -10, 10],
    projection="X10c",
    frame="ag",
    data=numpy.array([[0, 0, 2.5, 90, 270], [0, 0, 1.5, 90, 270]]),
    style="m0.5c+ea",
    pen="2p",
    color="red3",
)
fig.show()
```

seisman:
It works for me. What's your PyGMT and GMT versions?
```
python -c "import pygmt; pygmt.show_versions()"
```

noorbuchi:
This is what I got:
```
PyGMT information:
  version: v0.3.0
System information:
[...]

PR #1613, opened 2021-11-07
Refactor meca to use virtualfile_from_data

weiji14:

Part 1/3 of `meca` refactor as mentioned at https://github.com/GenericMappingTools/pygmt/issues/1129#issuecomment-962592735. Focus of this Pull Request is to clean up the `pygmt.meca` code and add some unit tests so that more refactorings can take place later to fix bugs. The `meca` module could use a lot of refactoring in general, so feel free to suggest changes if it simplifies/improves the code!

Notes:

1. This Pull Request will **neither** resolve the `offset` bug in #1129 nor allow beachballs to be plotted with string labels, c.f. https://github.com/GenericMappingTools/pygmt/pull/516#issuecomment-657313059. Those will be done in a separate PR to keep this one small.
2. I'm not a seismologist, so don't trust what I'm doing :stuck_out_tongue_winking_eye:

Part of #949. Towards resolving the bug at #1129 and enhancement request at #1466.

**Reminders**

**Slash Commands**


pygmt/src/meca.py
```diff
@@ -442,20 +435,19 @@ def update_pointers(data_pointers):
         else:
             raise GMTError("Parameter 'spec' contains values of an unsupported type.")
 
+    # Ensure non-file types are a 2d array
+    if isinstance(spec, (list, np.ndarray)):
+        spec = np.atleast_2d(spec)
```
weiji14, line 440: This is a brute-force solution to make 1d lists and numpy.arrays work (c.f. #1132). Note to self to remove this, because it will break if we allow for text labels (i.e. `str` dtype) on beachballs later.

pygmt/src/meca.py
```diff
@@ -306,95 +322,20 @@ def update_pointers(data_pointers):
...
+                "depth": np.atleast_1d(depth),
+            }
+            _spec.update({key: np.atleast_1d(val) for key, val in spec.items()})
+            spec = pd.DataFrame.from_dict(_spec)
+
+        assert isinstance(spec, pd.DataFrame)
+        dict_type_pointer = spec.values
+
+        # Assemble the array for the case of pd.DataFrames
+        if isinstance(dict_type_pointer, np.ndarray):
```
weiji14, line 338: Just to explain this change (clearer if you look at commit e2230573aa751c61fcb25d997454d3bbcd4e6a45), previously the `spec` could be formatted in 3 different paths depending on the type:

1. `if isinstance(dict_type_pointer, (int, float)):` (i.e. a dictionary of int/float like `{"depth": 0}`
2. `elif isinstance(dict_type_pointer, list):` (i.e. a dictionary of lists like `{"depth": [0, 1, 2]}`
3. `elif isinstance(dict_type_pointer, np.ndarray):` (i.e. a pandas.DataFrame with `np.ndarray` columns)

There was a lot of duplication in these 3 code paths, so I've removed paths 1 and 2 by converting any `spec` provided by the user in dictionary format into `pandas.DataFrame`. So now, everything happens via path 3 only.

Another benefit of handling everything using pandas.DataFrame is that columns can have different data types, so this sets things up for the ability to pass in string labels later in a follow up PR.

pygmt/src/meca.py
```diff
@@ -247,6 +240,29 @@ def update_pointers(data_pointers):
         ):
```
seisman, line 240: Lines 237-241 are unnecessary because line 236 already checks `spec` is dict or pd.DataFrame. Am I right?

weiji14: You might be right, let me recheck the logic.

seisman: @weiji14 Please see my new PR #1784.

weiji14:
Closing this in favour of #1784.

PR #1618, opened 2021-11-08
Allow users to set the waiting time when displaying a preview image using an external viewer

### LINKED ISSUE #1061: Fail to open images in external viewer programs if running PyGMT in a Python script

seisman:

See https://github.com/GenericMappingTools/pygmt/issues/1061 for the detailed issue report and discussions.

This PR adds the `waiting` parameter to the `launch_external_viewer` function. The default value is 0, so no waiting by default.

As mentioned in https://github.com/GenericMappingTools/pygmt/issues/1061#issuecomment-960374177, the `launch_external_viewer` function is used at two places, one in the `Figure.show()` function, another in the `Figure.savefig()` function. 

- The function call in the `Figure.show()` opens the preview image in the temporary directory, thus we need to set the waiting time to a non-zero value (`waiting=0.5` works for most users) so the external viewer can open the preview image before the PyGMT session exits.
- The function call in the `Figure.savefig()` opens the image that was already saved in the current directory or elsewhere, and won't be deleted when the PyGMT session exits. Thus there is no need to set the waiting time here.

Fixes #1061.

**Reminders**

**Slash Commands**


willschlitzer:
I assume this change would slow down running multiple calls of `fig.show()` in a notebook environment? If that is the case, I would think it would slow down processes like running a notebook with multiple figures or when building documentation. Could this be changed to apply the waiting time only when `method="external"`?

pygmt/figure.py
```diff
@@ -284,7 +284,12 @@ def show(self, dpi=300, width=500, method=None):
         This is useful when running unit tests and building the documentation
         in consoles without a Graphical User Interface.
 
-        Note that the external viewer does not block the current process.
+        Note that the external viewer does not block the current process, thus
+        it's necessary to suspend the execution of the current process for a
+        short while after launching the external view, so that the preview
+        image won't be deleted before the external viewer tries to open it. Set
+        the ``waiting`` parameter to a larger number if your computer is old
+        and slow.
```
willschlitzer, line 292:
```suggestion
        the ``waiting`` parameter to a larger number if your computer is slow.
```
I feel like "slow" is enough to let the user know to change the waiting parameter.

pygmt/figure.py
```diff
@@ -298,6 +302,10 @@ def show(self, dpi=300, width=500, method=None):
             - **external**: PDF preview in an external program [default]
             - **notebook**: PNG preview [default in Jupyter notebooks]
             - **none**: Disable image preview
+        waiting : float
+            Suspend the execution of the current process for a given number of
+            seconds after launching an external viewer.
+            Only works if ``method="external"``.
```
weiji14, line 308:
```suggestion
            seconds after launching an external viewer. [Default is 0.5].
            Only works if ``method="external"``.
```

seisman: It's clear that `waiting` has a default value of 0.5 from the function parameter `waiting=0.5`, thus no need to mention it in the docstrings.

weiji14 (approved):

pygmt/figure.py
```diff
@@ -284,7 +284,11 @@ def show(self, dpi=300, width=500, method=None):
         This is useful when running unit tests and building the documentation
         in consoles without a Graphical User Interface.
 
-        Note that the external viewer does not block the current process.
+        Note that the external viewer does not block the current process, thus
+        it's necessary to suspend the execution of the current process for a
+        short while after launching the external view, so that the preview
```
seisman, line 289:
```suggestion
        short while after launching the external viewer, so that the preview
```

michaelgrund (approved):

willschlitzer (approved):

PR #1631, opened 2021-11-22
Add html-noplot target to doc/Makefile

meghanrjones:

We have many gallery examples and tutorials :tada:, such that it takes a long time to generate the html documentation. I find this frustrating when locally editing/reviewing documentation unrelated to the gallery (e.g., when wrapping a new module). This PR adds an `html-noplot` target to `doc/Makefile` to generate the html documentation without running the examples.

Note: there will be a `WARNING: The config value 'plot_gallery' has type 'str', defaults to 'bool'.` when using this target - from looking at the sphinx_gallery issues I do not think there's a way around this.

Ref: https://sphinx-gallery.github.io/stable/advanced.html#build-the-gallery-without-running-any-examples

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -13,11 +13,12 @@ ALLSPHINXOPTS = -d $(BUILDDIR)/doctrees  $(SPHINXOPTS) .
 
 help:
 	@echo "Please use 'make <target>' where <target> is one of"
-	@echo "  all        build the HTML files from the existing rst sources"
-	@echo "  api        generate rst source files of API documentation"
-	@echo "  html       build the HTML files from the existing rst sources"
-	@echo "  server     make a local HTTP server for previewing the built documentation"
-	@echo "  clean      clean up built and generated files"
+	@echo "  all          build the HTML files from the existing rst sources"
+	@echo "  api          generate rst source files of API documentation"
+	@echo "  html         build the HTML files from the existing rst sources"
+	@echo "  html-noplot  build the HTML files without running any examples"
+	@echo "  server       make a local HTTP server for previewing the built documentation"
+	@echo "  clean        clean up built and generated files"
 
 all: html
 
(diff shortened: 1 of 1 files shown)


doc/Makefile
```diff
@@ -36,6 +37,15 @@ html: api
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+html-noplot: api
+	@echo
+	@echo "Building HTML files."
```
willschlitzer, line 42:
```suggestion
	@echo "Building HTML files without example plots."
```
I think it's good to have a different print statement than `make html`.

willschlitzer (approved):
One minor suggestion to differentiate `make html` and `make html-noplot`, but I think this is a good idea to speed up building documentation.

michaelgrund (approved):

doc/Makefile
```diff
@@ -36,6 +37,15 @@ html: api
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+html-noplot: api
+	@echo
+	@echo "Building HTML files without example plots."
+	@echo
+	# Set PYGMT_USE_EXTERNAL_DISPLAY to "false" to disable external display
+	PYGMT_USE_EXTERNAL_DISPLAY="false" $(SPHINXBUILD) -D plot_gallery=0 -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
```
seisman, line 45: Setting `PYGMT_USE_EXTERNAL_DISPLAY` to `false` seems unnecessary because no examples will be executed. Removing it works well for me locally, but please double check it.
```suggestion
	$(SPHINXBUILD) -D plot_gallery=0 -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
```

meghanrjones: Yes, this also works well for me with your suggestion.

seisman (approved):

PR #1669, opened 2021-12-14
Set timeout limit for CI tests

meghanrjones:

Occasionally, the CI tests get caught and timeout after 6 hours. Given that the tests should complete within 30 minutes, this PR proposes to set [timeout-minutes](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) to end after 60 minutes rather than the default of 360.

To-do:
- [x] Set limit for dev tests

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -36,6 +36,7 @@ jobs:
             isDraft: true
           - os: windows-latest
             isDraft: true
+    timeout-minutes: 30
     defaults:
       run:
         shell: bash -l {0}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -53,6 +53,7 @@ jobs:
           - python-version: 3.9
             numpy-version: '1.21'
             optional-packages: 'geopandas ipython'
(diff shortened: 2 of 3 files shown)


seisman:
Perhaps also set the timeout limit for the `ci_docs.yml` workflow?

willschlitzer:
I think this is a good idea, and agree that 30 minutes should be enough.

weiji14:
> Perhaps also set the timeout limit for the `ci_docs.yml` workflow?

Probably don't need to, since the Docs workflow doesn't use `dvc` (which is what the CI gets stuck on usually).

seisman:
> > Perhaps also set the timeout limit for the `ci_docs.yml` workflow?
> 
> Probably don't need to, since the Docs workflow doesn't use `dvc` (which is what the CI gets stuck on usually).

But it does no harm to timeout the workflow after 30 minutes, considering that the workflow should finish in 20 minutes.

meghanrjones:
OK, I added a limit to the docs workflow.

weiji14 (approved):

seisman (approved):

Issue #1644, opened 2021-12-04
Bump to GMT 6.3.0


meghanrjones:
:tada: [GMT 6.3.0](https://github.com/GenericMappingTools/gmt/releases/tag/6.3.0) has been released! It is installable from the [conda-forge channel](https://anaconda.org/conda-forge/gmt/files) using the following command:

`conda install -c conda-forge gmt=6.3.0`

**To-Do**:
- [x] Bump the minimum required GMT version (@meghanrjones - https://github.com/GenericMappingTools/pygmt/pull/1649)
  - [x] Update `.github/workflows/cache_data.yaml` 
  - [x] Update `.github/workflows/ci_docs.yml`
  - [x] Update `.github/workflows/ci_tests.yaml`
  - [x] Update `doc/install.rst`
  - [x] Update `environment.yml`
  - [x] Update `required_version` in `pygmt/clib/session.py`
  - [x] Update `test_get_default` in `pygmt/tests/test_clib.py`
  - [x] Update compatibility table in `README.rst`
- [x] Fix failing tests (1 or more PRs)
  - [x] `pygmt/tests/test_basemap.py::test_basemap_rose` @weiji14 #1658
  - [x] `pygmt/tests/test_basemap.py::test_basemap_compass` @weiji14 #1658
  - [x] `pygmt/tests/test_inset.py::test_inset_context_manager` @meghanrjones https://github.com/GenericMappingTools/pygmt/pull/1653
  - [x] `pygmt/tests/test_legend.py::test_legend_specfile` @meghanrjones https://github.com/GenericMappingTools/pygmt/pull/1654
  - [x] `pygmt/tests/test_rose.py::test_rose_bools` @meghanrjones https://github.com/GenericMappingTools/pygmt/pull/1655
  - [x] `pygmt/tests/test_rose.py::test_rose_deprecate_columns_to_incols` @meghanrjones https://github.com/GenericMappingTools/pygmt/pull/1655
- [x] Remove [xfail](https://docs.pytest.org/en/stable/skipping.html#xfail-mark-test-functions-as-expected-to-fail) pytest markers on tests that are now xpass (@meghanrjones) https://github.com/GenericMappingTools/pygmt/pull/1656
  - [x] `pygmt/tests/test_grdimage.py::test_grdimage_grid_and_shading_with_xarray[png]` (xref https://github.com/GenericMappingTools/pygmt/issues/618#issuecomment-961579394)
[...]

PR #1673, opened 2021-12-17
Figure.psconvert: Add new aliases and deprecate parameter "icc_gray" (remove in v0.8.0)

michaelgrund:

This PR updates the psconvert -A syntax to account for GMT 6.3 change made at GenericMappingTools/gmt#5583.

**Preview** at https://pygmt-git-update-psconvert-syntax-gmt.vercel.app/api/generated/pygmt.Figure.psconvert.html

Part of #1644

**Reminders**

**Slash Commands**


michaelgrund:
I'm not fully sure which parts need to be changed here @GenericMappingTools/pygmt-maintainers . Every change that was introduced in GenericMappingTools/gmt#5583?

weiji14:
> I'm not fully sure which parts need to be changed here @GenericMappingTools/pygmt-maintainers . Every change that was introduced in [GenericMappingTools/gmt#5583](https://github.com/GenericMappingTools/gmt/pull/5583)?

I would also update `-I` (icc_gray) as a minimum since that is aliased. The changes in `-N` and `-W` could be done in a separate PR when we add long aliases for those.

michaelgrund:
> > I'm not fully sure which parts need to be changed here @GenericMappingTools/pygmt-maintainers . Every change that was introduced in [GenericMappingTools/gmt#5583](https://github.com/GenericMappingTools/gmt/pull/5583)?
> 
> I would also update `-I` (icc_gray) as a minimum since that is aliased. The changes in `-N` and `-W` could be done in a separate PR when we add long aliases for those.

I agree, however, it seems like ``icc_gray`` isn't a good alias for ``-I`` anymore @weiji14.

pygmt/figure.py
```diff
@@ -167,8 +172,24 @@ def psconvert(self, **kwargs):
             using the input names as base, which are appended with an
             appropriate extension. Use this option to provide a different name,
             but without extension. Extension is still determined automatically.
-        icc_gray : bool
-            Enforce gray-shades by using ICC profiles.
+        icc_gray : str
```
weiji14, line 175: > > > I'm not fully sure which parts need to be changed here @GenericMappingTools/pygmt-maintainers . Every change that was introduced in [GenericMappingTools/gmt#5583](https://github.com/GenericMappingTools/gmt/pull/5583)?
> > 
> > I would also update `-I` (icc_gray) as a minimum since that is aliased. The changes in `-N` and `-W` could be done in a separate PR when we add long aliases for those.
> 
> I agree, however, it seems like ``icc_gray`` isn't a good alias for ``-I`` anymore @weiji14.

Yeah, this seems completely different. Should we rename it to margin, scale, or something else? This would need to require a deprecation cycle too 😅

seisman: > Should we rename it to margin, scale, or something else?

Is `resize` better?

michaelgrund: Would also go with `resize`. What do @GenericMappingTools/pygmt-maintainers mean?

weiji14: Not perfect, but I think `resize` will do. Unless someone can come up with a better name 🙃

willschlitzer: `resize` works for me

pygmt/figure.py
```diff
@@ -167,8 +172,24 @@ def psconvert(self, **kwargs):
...
+            [**+m**\ *margins*][**+s**\ [**m**]\ *width*\
+            [/\ *height*]][**+S**\ *scale*] ]
```
weiji14, line 177:
```suggestion
            [/\ *height*]][**+S**\ *scale*] ].
```

pygmt/figure.py
weiji14: Ok, go let's go with `resize` then @michaelgrund:

```suggestion
        resize : str
```

Also, please modify:
- L132: `I="icc_gray",` -> `I="resize",`
- L127, add `@deprecate_parameter("icc_gray", "resize", "v0.6.0", remove_version="v0.8.0")` (below `@fmt_docstring`)

Make sure to change the PR title to ''Figure.psconvert: Deprecate parameter "icc_gray" to "resize" (remove in v0.8.0)" and update the PR label to 'deprecation' so that it goes in the changelog. Follow steps detailed at https://www.pygmt.org/v0.5.0/maintenance.html#backwards-compatibility-and-deprecation-policy.

willschlitzer (approved):
looks good!

pygmt/figure.py
```diff
@@ -124,12 +125,13 @@ def region(self):
         return wesn
 
     @fmt_docstring
+    @deprecate_parameter("icc_gray", "resize", "v0.6.0", remove_version="v0.8.0")
```
seisman, line 128: After the changes in this PR, scripts that use `icc_gray` will see a warning:

```
The 'icc_gray' parameter has been deprecated since v0.6.0 and will be removed in v0.8.0.
Please use 'resize' instead.
```
This warning doesn't make sense, because the 'resize' parameter has nothing to do with ICC profiles, so changing `icc_gray` to `resize` still breaks users old scripts.

I think adding the `resize` parameter should be considered an enhancement. As for `icc_gray`, I don't have a good way to deprecate this parameter without breaking backward compatibility, but perhaps we can change the `deprecate_parameter` decorator to allow a `None` new parameter name.
```
@deprecate_parameter("icc_gray", None, "v0.6.0", remove_version="v0.8.0")
```
This new style should give a warning like:

```
The 'icc_gray' parameter has been deprecated since v0.6.0 and will be removed in v0.8.0.
```

michaelgrund: Agree with you @seisman. However, this would mean that we need to support both parameters, ``icc_gray`` as well as ``resize`` for a while, right?

seisman: Perhaps we can change the function definition to:
```
psconvert(self, icc_gray=False, **kwargs):
```
and check if `icc_gray` is set to `True`:
```
if icc_gray:
    warnings.warn("depreaction warning here", category=FutureWarning, stacklevel=2)
    if "N" not in kwargs:
        kwargs["N"] = "+i"
    else: 
        kwargs["N"] += "+i"
```

michaelgrund: Ok, we also need to add an alias for ``-N``. Currently I named it ``bb_style`` (boundingbox style). Feel free to provide other suggestions.

pygmt/figure.py
```diff
@@ -167,8 +174,40 @@ def psconvert(self, **kwargs):
...
+            the default pen of 0.25p,black). Note: If both **+g** and **+f**
+            are used then we use paint as the fade color instead of black.
+            Append **+i** to enforce gray-shades by using ICC profiles.
+            Ghostscript versions >= 9.00 change gray-shades by using ICC
+            profiles. Ghostscript 9.05 and above provide the
+            '-dUseFastColor=true' option to prevent that and that is what
+            **psconvert** does by default, unless modifier **+i** is set.
+            Note that for Ghostscript >= 9.00 and < 9.05 the gray-shade
+            shifting is applied to all but PDF format. We have no solution to
+            offer other than suggesting you upgrade Ghostscript.
```
seisman, line 210: I think it's OK to remove this long description about old ghostscript versions, because these versions are [~10 years old](https://www.ghostscript.com/doc/9.55.0/History9.htm), and I believe very few people will use them.  
```suggestion
```

seisman (approved):
Looks good to me.

pygmt/tests/test_figure.py
```diff
@@ -251,3 +251,12 @@ def test_figure_set_display_invalid():
...
         set_display(method="invalid")
+
+
+def test_figure_icc_gray():
+    """
+    Check if icc_gray parameter works correctly if used.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
+    fig.psconvert(icc_gray=True, prefix="Test")
```
weiji14, line 262: Also need to add `import pytest` at L14 for this to work.
```suggestion
    with pytest.warns(expected_warning=FutureWarning) as record:
        fig.psconvert(icc_gray=True, prefix="Test")
        assert len(record) == 1  # check that only one warning was raised
```

PR #1682, opened 2021-12-21
Use xarray.testing rather than pygmt.grdinfo in xyz2grd tests

willschlitzer:

This PR refactors the pygmt.xyz2grd tests to use xarray.testing.assert_allclose() rather than pygmt.grdinfo() in preparation for the deprecation associated with better return values for grdinfo (https://github.com/GenericMappingTools/pygmt/issues/593).

**Reminders**

**Slash Commands**


pygmt/tests/test_xyz2grd.py
```diff
@@ -19,40 +19,52 @@ def fixture_ship_data():
     return load_sample_bathymetry()
 
 
-def test_xyz2grd_input_file():
```
seisman, line 22: Why do you remove the `test_xyz2grd_input_file` test?

pygmt/tests/test_xyz2grd.py
```diff
@@ -19,40 +19,52 @@ def fixture_ship_data():
...
+                25.0,
+                30.0,
+            ],
+        ),
+        dims=["y", "x"],
+    )
 
 
-def test_xyz2grd_input_array(ship_data):
+def test_xyz2grd_input_array(ship_data, expected_grid):
```
seisman, line 45: The tests `test_xyz2grd_input_array` and `test_xyz2grd_input_df` can be combined into a single test by using the `pytest.mark.parametrize` decorator, similar to https://github.com/GenericMappingTools/pygmt/blob/7061dacfa227ba12a2f6eaa23df928c556928486/pygmt/tests/test_nearneighbor.py#L24

pygmt/tests/test_xyz2grd.py
willschlitzer: I wanted to standardize it with the other modules that import remote files using a dedicated module (in this case, `load_sample_bathymetry()`). I can re-add it, but I figured it was a redundant test, as it is testing the GMT function to call the remote file rather than the PyGMT function to convert the information to a pandas dataframe or numpy array.

pygmt/tests/test_xyz2grd.py
willschlitzer: added in d4be06a

willschlitzer:
@meghanrjones This is listed at #1684 as a test to use `static_earth_relief`. Would it be better to use `load_sample_bathymetry` that already has xyz-data?

meghanrjones:
> @meghanrjones This is listed at #1684 as a test to use `static_earth_relief`. Would it be better to use `load_sample_bathymetry` that already has xyz-data?

Yes, that was an oversight. It would be better to use xyz-data.

pygmt/tests/test_xyz2grd.py
```diff
@@ -6,7 +6,7 @@
 import numpy as np
 import pytest
 import xarray as xr
-from pygmt import grdinfo, xyz2grd
+from pygmt import load_dataarray, xyz2grd
 from pygmt.datasets import load_sample_bathymetry
```
meghanrjones, line 10:
```suggestion
from pygmt.datasets import load_sample_data
```
I think you should use ``load_sample_data(name='bathymetry')`` rather than load_sample_bathymetry since the latter will be deprecated.

pygmt/tests/test_xyz2grd.py
```diff
@@ -16,54 +16,56 @@ def fixture_ship_data():
...
+            [-2546.2512, -1977.8754, -963.23303],
+            [-352.3795, -1025.4508, np.nan],
+        ],
+        coords=dict(
+            x=[245.0, 250.0, 255.0],
+            y=[
+                20.0,
+                25.0,
+                30.0,
+            ],
```
seisman, line 39:
```suggestion
            y=[20.0, 25.0, 30.0],
```

seisman (approved):

PR #1683, opened 2021-12-21
Change test_grdproject.py to use static_earth_relief

willschlitzer:

This PR refactors the pygmt.grdproject tests to use xarray.testing.assert_allclose() rather than pygmt.grdinfo() in preparation for the deprecation associated with better return values for grdinfo (https://github.com/GenericMappingTools/pygmt/issues/593).

**Reminders**

**Slash Commands**


meghanrjones:
> I wasn't able to change the dimensions of the output grid using the `region` parameter. I don't want to use `grdcut` in this test to get a smaller input grid, but I'm not sure why I'm unable to change the 14x8 output. I tried inputting region coordinates both in lat/long and x/y.

The behavior of ``region`` is a bit trickier in grdproject than the other modules. Using ``region`` alone will impact what data are projected but won't impact the number of nodes unless ``spacing`` or ``dpi`` are set. Without one of those two other parameters, the default is to output a grid with the same number of nodes as the input grid such that the x_inc and/or y_inc would be smaller if a subregion is selected.

pygmt/tests/test_grdproject.py
```diff
@@ -3,52 +3,74 @@
...
         assert os.path.exists(path=tmpfile.name)  # check that outgrid exists
-        result = grdinfo(tmpfile.name, per_column=True).strip().split()
-        npt.assert_allclose(float(result[0]), 0)  # x min
-        npt.assert_allclose(float(result[1]), 10)  # x max
-        npt.assert_allclose(float(result[2]), 0, atol=1.0e-10)  # y min
-        npt.assert_allclose(float(result[3]), 9.94585661273)  # y max
-        npt.assert_allclose(float(result[4]), -5130.48193359)  # min
-        npt.assert_allclose(float(result[5]), -152.585281372)  # max
+        temp_grid = load_dataarray(tmpfile.name)
+        print(temp_grid)
```
meghanrjones, line 58:
```suggestion
```

seisman (approved):

meghanrjones (approved):

PR #1685, opened 2021-12-23
Use a single function for loading any sample dataset

meghanrjones:

This PR partially implements the suggestions in #1436, in deprecating individual functions in favor of a single function for loading sample datasets.

It also adds a `list_sample_dataframes` function that lists the available dataset names to provide to `load_sample_dataframe`.

A couple questions:
1. Should we have one function that returns either an xarray.DataArray or pandas.DataFrame depending on the specific file requested or one function for tabular data samples and another for raster data samples?
2. Currently, the user would pass exact name of the file on the GMT server. Should we continue with this implementation or provide different names that are more similar to the existing functions (e.g., `name="japan_quakes"` vs `name="tut_quakes.ngdc"`)?

To do:
- [ ] Deprecate other functions for loading a sample pandas.DataFrame
- [ ] Deprecate functions for loading a sample xarray.DataArray
- [ ] Update syntax in gallery examples and tutorials.

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -210,6 +210,7 @@ and store them in the GMT cache folder.
 .. autosummary::
     :toctree: generated
 
+    datasets.list_sample_dataframes
     datasets.load_earth_age
     datasets.load_earth_relief
     datasets.load_fractures_compilation
@@ -218,6 +219,7 @@ and store them in the GMT cache folder.
     datasets.load_mars_shape
     datasets.load_ocean_ridge_points
     datasets.load_sample_bathymetry
+    datasets.load_sample_dataframe
     datasets.load_usgs_quakes
 
 .. automodule:: pygmt.exceptions
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -213,13 +213,13 @@ and store them in the GMT cache folder.
     datasets.list_sample_data
     datasets.load_earth_age
     datasets.load_earth_relief
+    datasets.load_sample_data
     datasets.load_fractures_compilation
     datasets.load_hotspots
     datasets.load_japan_quakes
     datasets.load_mars_shape
     datasets.load_ocean_ridge_points
     datasets.load_sample_bathymetry
-    datasets.load_sample_data
     datasets.load_usgs_quakes
 
 .. automodule:: pygmt.exceptions
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -210,7 +210,7 @@ and store them in the GMT cache folder.
 .. autosummary::
     :toctree: generated
 
-    datasets.list_sample_dataframes
+    datasets.list_sample_data
     datasets.load_earth_age
     datasets.load_earth_relief
     datasets.load_fractures_compilation
@@ -219,7 +219,7 @@ and store them in the GMT cache folder.
     datasets.load_mars_shape
     datasets.load_ocean_ridge_points
     datasets.load_sample_bathymetry
-    datasets.load_sample_dataframe
+    datasets.load_sample_data
     datasets.load_usgs_quakes
 
 .. automodule:: pygmt.exceptions
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -61,7 +61,7 @@ def load_japan_quakes(**kwargs):
     pandas.DataFrame.
 
     .. warning:: Deprecated since v0.6.0. This function has been replaced with
-       ``load_sample_data(name="tut_quakes.ngdc")`` and will be removed in
+       ``load_sample_data(name="japan_quakes")`` and will be removed in
        v0.9.0.
 
     Data is from the NOAA NGDC database. This is the ``@tut_quakes.ngdc``
@@ -82,7 +82,7 @@ def load_japan_quakes(**kwargs):
         warnings.warn(
             "This function has been deprecated since v0.6.0 and will be "
             "removed in v0.9.0.. Please use "
-            "load_sample_data(name='tut_quakes.ngdc') instead.",
+            "load_sample_data(name='japan_quakes') instead.",
             category=FutureWarning,
             stacklevel=2,
         )
@@ -108,7 +108,7 @@ def load_ocean_ridge_points(**kwargs):
     pandas.DataFrame (Deprecated).
 
     .. warning:: Deprecated since v0.6.0. This function has been replaced with
-       ``load_sample_data(name="ridge.txt")`` and will be removed in
+       ``load_sample_data(name="ocean_ridge_points")`` and will be removed in
        v0.9.0.
 
     This is the ``@ridge.txt`` dataset used in the GMT tutorials.
@@ -126,7 +126,7 @@ def load_ocean_ridge_points(**kwargs):
     if "suppress_warning" not in kwargs:
         warnings.warn(
             "This function has been deprecated since v0.6.0 and will be removed "
-            "in v0.9.0. Please use load_sample_data(name='ridge.txt') "
+            "in v0.9.0. Please use load_sample_data(name='ocean_ridge_points') "
             "instead.",
             category=FutureWarning,
             stacklevel=2,
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -10,16 +10,27 @@ from pygmt.src import which
 
 def list_sample_data():
     """
-    Report tabular datasets available for tests and documentation examples.
+    Report datasets available for tests and documentation examples.
 
     Returns
     -------
     output : dict
-        Names and short descriptions of available sample dataframes.
+        Names and short descriptions of available sample datasets.
+
+    See Also
+    --------
+    load_sample_data : Load an example dataset from the GMT server.
     """
     names = {
         "japan_quakes": "Table of earthquakes around Japan from NOAA NGDC database",
         "ocean_ridge_points": "Table of ocean ridge points for the entire world",
+        "bathymetry": "Table of ship bathymetric observations off Baja California",
+        "usgs_quakes": "Table of global earthquakes from the USGS",
+        "fractures": "Table of hypothetical fracture lengths and azimuths",
+        "hotspots": "Table of locations, names, and symbol sizes of hotpots from "
+        " Mueller et al., 1993",
+        "mars_shape": "Table of topographic signature of the hemispheric dichotomy of "
+        " Mars from Smith and Zuber (1996)",
     }
     return names
 
@@ -39,8 +50,14 @@ def load_sample_data(name):
 
     Returns
     -------
-    output : pandas.DataFrame
-        Tabular dataset.
+    output : pandas.DataFrame or xarray.DataArray
+        Sample dataset loaded as a pandas.DataFrame for tabular data or
+        xarray.DataArray for raster data
+
+    See Also
+    --------
+    list_sample_data : Report datasets available for tests and documentation
+        examples.
     """
     names = list_sample_data()
     if name not in names:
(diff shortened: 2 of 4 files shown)


willschlitzer:
I like the idea of having a single function to call datasets, but I think it may make `load_sample_dataframe` a big and cumbersome function. Do we want some non-user facing functions to handle the dataframe reading?

```
def load_sample_dataframe(name)
    names = list_sample_dataframes()
    if name not in names:
        raise GMTInvalidInput(f"Invalid dataset name '{name}'")

    fname = which("@" + name, download="c")

    if name == "tut_quakes.ngdc":
        return load_japan_quakes(fname=fname)

def load_japan_quakes(fname):
    data = pd.read_csv(fname, header=1, sep=r"\s+")
    data.columns = [
        "year",
        "month",
        "day",
        "latitude",
        "longitude",
        "depth_km",
        "magnitude",
    ]
    return data
```

meghanrjones:
> I like the idea of having a single function to call datasets, but I think it may make `load_sample_dataframe` a big and cumbersome function. Do we want some non-user facing functions to handle the dataframe reading?
> 
> ```
> def load_sample_dataframe(name)
>     names = list_sample_dataframes()
>     if name not in names:
>         raise GMTInvalidInput(f"Invalid dataset name '{name}'")
> 
>     fname = which("@" + name, download="c")
> 
>     if name == "tut_quakes.ngdc":
>         return load_japan_quakes(fname=fname)
> 
> def load_japan_quakes(fname):
>     data = pd.read_csv(fname, header=1, sep=r"\s+")
>     data.columns = [
>         "year",
>         "month",
>         "day",
>         "latitude",
>         "longitude",
>         "depth_km",
>         "magnitude",
>     ]
>     return data
> ```

I agree that separating the parsing code would be more readable. I went with the current implementation anyways because I thought we should issue a deprecation notice if `load_japan_quakes` is called by the user, but not if it is called by `load_sample_dataframe` and could not find any satisfactory solutions for distinguishing between those two cases.

willschlitzer:
> I agree that separating the parsing code would be more readable. I went with the current implementation anyways because I thought we should issue a deprecation notice if `load_japan_quakes` is called by the user, but not if it is called by `load_sample_dataframe` and could not find any satisfactory solutions for distinguishing between those two cases.

My solution would be to just copy and paste the `load_japan_quakes` function to something like `load_japan_quakes_sample_data` (name isn't great; just a placeholder suggestion).  The `load_japan_quakes` function gets a deprecation notice, and we'll be alerted before a release when it is time to be take it out. The `load_japan_quakes_sample_data` function is called by `load_sample_data` and can be eventually renamed to `load_japan_quakes` without concerns for deprecation, as it isn't user-facing.

I know it's a little cumbersome, but I think it's better to have some redundant code to set the precedent that `load_sample_data` will be calling functions.

meghanrjones:
> I know it's a little cumbersome, but I think it's better to have some redundant code to set the precedent that `load_sample_data` will be calling functions.

Sounds good, I updated the code based on your suggestions.

willschlitzer:
> > I know it's a little cumbersome, but I think it's better to have some redundant code to set the precedent that `load_sample_data` will be calling functions.
> 
> Sounds good, I updated the code based on your suggestions.

I think your implementation works better than what I envisioned; I like that the original functions still get a returned value from `load_sample_dataframe`. I'm assuming you want the other functions to be updated on later PRs (before the next release)?

meghanrjones:
> > > I know it's a little cumbersome, but I think it's better to have some redundant code to set the precedent that `load_sample_data` will be calling functions.
> > 
> > 
> > Sounds good, I updated the code based on your suggestions.
> 
> I think your implementation works better than what I envisioned; I like that the original functions still get a returned value from `load_sample_dataframe`. I'm assuming you want the other functions to be updated on later PRs (before the next release)?

I updated these two functions first to get agreement on the structure before putting a lot of work in. I would lean slightly towards updating the other functions in other PRs to keep the PR size down and allow new additions in the meantime (e.g., MaunaLao_CO2 for https://github.com/GenericMappingTools/pygmt/pull/1512), but could update them in this PR (likely after Jan 3) if that's what others prefer.

doc/api/index.rst
```diff
@@ -218,6 +219,7 @@ and store them in the GMT cache folder.
     datasets.load_mars_shape
     datasets.load_ocean_ridge_points
     datasets.load_sample_bathymetry
+    datasets.load_sample_dataframe
```
weiji14, line 222: If we're deprecating the other load_* functions, do we need to keep them here? Or add a 'Deprecated' note somewhere?

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,67 @@
...
+
+
 def load_japan_quakes():
     """
-    Load a table of earthquakes around Japan as a pandas.DataFrame.
+    Load a table of earthquakes around Japan as a pandas.DataFrame (Deprecated)
+
+    .. warning:: Deprecated since v0.6.0. This function has been replaced with
+       ``load_sample_dataframe(name="tut_quakes.ngdc")`` and will be removed in
+       v0.9.0.
```
weiji14, line 64: Shouldn't this ve removed in v0.8.0, i.e. 2 major versions? Same with the other ones below.
```suggestion
       v0.8.0.
```

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,67 @@
...
+    if name == "ocean_ridge_points":
+        data = load_ridge()
+
+    return data
+
+
 def load_japan_quakes():
     """
-    Load a table of earthquakes around Japan as a pandas.DataFrame.
+    Load a table of earthquakes around Japan as a pandas.DataFrame (Deprecated)
```
weiji14, line 60: Maybe put (Deprecated) at the beginning of the sentence so that it stands out in the API docs list?
```suggestion
    (Deprecated) Load a table of earthquakes around Japan as a pandas.DataFrame.
```

pygmt/datasets/samples.py
```diff
@@ -22,6 +76,32 @@ def load_japan_quakes():
...
+        "in v0.9.0. Please use load_sample_dataframe(name='tut_quakes.ngdc') "
+        "instead.",
+        category=FutureWarning,
+        stacklevel=2,
+    )
+
+    return load_sample_dataframe("japan_quakes")
+
+
+def load_tut_quakes():
```
weiji14, line 91: If this function is meant to be removed at some point, maybe add an underscore to indicate it is non-user-facing
```suggestion
def _load_tut_quakes():
```

weiji14 (commented):
>  I went with the current implementation anyways because I thought we should issue a deprecation notice if load_japan_quakes is called by the user, but not if it is called by load_sample_dataframe and could not find any satisfactory solutions for distinguishing between those two cases.

I feel that there should be a way to silence the warning somehow so that we don't need to make new temporary functions. Will look into this in the coming days (I'm currently reviewing this on the plane, haha).

pygmt/datasets/samples.py
meghanrjones: I thought deprecating a whole function is a more major change than the alias updates, so I went on the longer side of the 2-4 releases recommendation in the contributing guide. I'm open to a shorter deprecation cycle though.

meghanrjones:
> > I went with the current implementation anyways because I thought we should issue a deprecation notice if load_japan_quakes is called by the user, but not if it is called by load_sample_dataframe and could not find any satisfactory solutions for distinguishing between those two cases.
> 
> I feel that there should be a way to silence the warning somehow so that we don't need to make new temporary functions. Will look into this in the coming days (I'm currently reviewing this on the plane, haha).

OK, I pushed a revised module that avoids using new functions at the expense of being a bit trickier. Rather than removing the original function (e.g., `load_japan_quakes`) at the end of the deprecation cycle, we could make the function internal (leading underscore, remove from imports and API docs) and remove the kwargs/suppress warning check.

doc/api/index.rst
seisman: I prefer to add "deprecated" notes in the docstrings of these functions.

pygmt/datasets/samples.py
seisman: I'm in favor of 3 releases.

seisman:
> 1. Should we have one function that returns either an xarray.DataArray or pandas.DataFrame depending on the specific file requested or one function for tabular data samples and another for raster data samples?

Personally, I prefer to have a single function `load_sample_data` to load everything.

meghanrjones:
> 



> > 1. Should we have one function that returns either an xarray.DataArray or pandas.DataFrame depending on the specific file requested or one function for tabular data samples and another for raster data samples?
> 
> Personally, I prefer to have a single function `load_sample_data` to load everything.

Sounds good to me, I updated the name accordingly. I actually realized we do not have any functions yet for loading sample DataArray objects from the GMT cache, but I assume that we will in the future.

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,100 @@
...
+
+    See Also
+    --------
+    list_sample_data : Report datasets available for tests and documentation
+        examples.
+    """
+    names = list_sample_data()
+    if name not in names:
+        raise GMTInvalidInput(f"Invalid dataset name '{name}'")
+
```
seisman, line 65: Can we simply have a dict mapping a name to a load function, like this:
```
load_func = {
    "japan_quakes": load_japan_quakes,
    "ocean_ridge_points": load_ocean_ridge_points,
}

data = load_func[name](supress_warning=True)
```
I didn't test it so I don't know if it works or not.

meghanrjones: It works. I think it is simpler, thanks.

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,91 @@
...
+    names = {
+        "japan_quakes": "Table of earthquakes around Japan from NOAA NGDC database",
+        "ocean_ridge_points": "Table of ocean ridge points for the entire world",
+        "bathymetry": "Table of ship bathymetric observations off Baja California",
+        "usgs_quakes": "Table of global earthquakes from the USGS",
+        "fractures": "Table of hypothetical fracture lengths and azimuths",
+        "hotspots": "Table of locations, names, and symbol sizes of hotpots from "
+        " Mueller et al., 1993",
+        "mars_shape": "Table of topographic signature of the hemispheric dichotomy of "
+        " Mars from Smith and Zuber (1996)",
```
seisman, line 33: Should we order the names alphabetically?

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,91 @@
...
+    --------
+    list_sample_data : Report datasets available for tests and documentation
+        examples.
+    """
+    names = list_sample_data()
+    if name not in names:
+        raise GMTInvalidInput(f"Invalid dataset name '{name}'.")
+
+    load_func = {
+        "japan_quakes": load_japan_quakes,
```
seisman, line 67: Same here. Order the names alphabetically.

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,91 @@
...
+        "fractures": "Table of hypothetical fracture lengths and azimuths",
+        "hotspots": "Table of locations, names, and symbol sizes of hotpots from "
+        " Mueller et al., 1993",
+        "mars_shape": "Table of topographic signature of the hemispheric dichotomy of "
+        " Mars from Smith and Zuber (1996)",
+    }
+    return names
+
+
+def load_sample_data(name):
```
seisman, line 38: I'm thinking if we can merge the `list_sample_data()` function into `load_sample_data()`. Calling `load_sample_data()` without giving a name can return the name-description dict.

meghanrjones: I'd prefer to keep them separate even though it requires two dictionaries because I think overall it's simpler to have each function have one purpose.

seisman (approved):

pygmt/datasets/samples.py
weiji14: Ok, stick to 3 minor versions then (i.e. v0.9.0).

pygmt/datasets/samples.py
```diff
@@ -1,13 +1,91 @@
...
+    Parameters
+    ----------
+    name : str
+        Name of the dataset to load.
+
+    Returns
+    -------
+    :class:`pandas.DataFrame` or :class:`xarray.DataArray`
+        Sample dataset loaded as a pandas.DataFrame for tabular data or
+        xarray.DataArray for raster data
```
seisman, line 55:
```suggestion
        xarray.DataArray for raster data.
```

willschlitzer (approved):

Issue #770, opened 2020-12-28
PyGMT tutorial


willschlitzer:
I think there should be a PyGMT tutorial, much like the [GMT tutorial](https://docs.generic-mapping-tools.org/6.1/tutorial.html). I don't think it requires as much explanation as what's happening under the hood, but would ideally recreate the Laboratory Exercises. The PyGMT User Guides cover a variety of topics, but I think following the GMT tutorial gives us a good roadmap on what features to explain for someone trying to learn.

weiji14:
This reminds me of the FOSS4GOceania 2019 workshop material I made at #317, and the ROSES 2020 stuff which @liamtoney made at #613. As a start, we could probably expose those to be more visible somewhere (and update the content since the API has changed).

If you're thinking about a full comprehensive GMT tutorial like the one linked, I'd suggest doing it session by session (and good luck with that)! On a side note, there might still be time (?) to put in an abstract for vEGU21 to gain more exposure via an online workshop. All depends on what we want to focus on ;)

willschlitzer:
I was planning on recreating exercises in the tutorial. I'm not sure how to split them up regarding sessions/pages, but I figure the hard part would be putting together the content, which can be done one exercise at a time.

Regarding the vEGU21 abstract, I was unaware that was something we were hoping to do. According to the website, there are still 16 days for abstract submission. I'm happy to help out on this, but my last conference abstract was for AGU Fall 2011, so I'm not too familiar with whats involved in this process.

liamtoney:
I will likely be providing a PyGMT lesson again for ROSES 2021.

willschlitzer:
@liamtoney Awesome! Your tutorial video was what brought me to PyGMT.

seisman:
Sorry, it's still unclear to me what's the difference between the proposed "PyGMT Tutorial" and the existing "PyGMT User Guides". I thought they're the same.

willschlitzer:
@seisman I'm thinking something more structured like the laboratory exercises that are in the GMT docs, instead of stand-alone user guides. I'm not saying the PyGMT User Guides should be eliminated, but I think having a series of exercises covering different PyGMT functions would be good for someone trying to familiarize themself with the package.

seisman:
Sounds good to me.

weiji14:
[...]

PR #1688, opened 2021-12-30
Add an "add a title" to starter tutorial

willschlitzer:
I originally intended to add a second page to the tutorial that covered adding a title, but changed my mind and think the title section would be more appropriate in first_figure.py rather than adding a separate page, which this PR does.

Additionally, to reduce confusion over switching between `frame=True` and frame=["a", "+tMaine"], this changes the frame example to use `frame="a"`.

Addresses #770 

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/get-started/first_figure.py
+++ b/examples/get-started/first_figure.py
@@ -121,7 +121,7 @@ fig.show()
 # annotate the latitude and longitude of the region.
 #
 # The ``frame`` parameter is used to add a frame to the figure. For now, it
-# will be set to ``"a"`` to **a**nnotate the axes automatically.
+# will be set to ``"a"`` to **a**\ nnotate the axes automatically.
 
 fig = pygmt.Figure()
 fig.coast(
--- a/examples/get-started/first_figure.py
+++ b/examples/get-started/first_figure.py
@@ -147,8 +147,8 @@ fig.show()
 # the figure title.
 #
 # If the figure title has any spaces, the string to set the title needs to
-# use single-quotes, with the actual title set in double quotes (e.g. setting
-# the title to "Coastal Maine" would use the syntax ``'+t"Coastal Maine"'``.
+# be wrapped in single-quotes, while the actual title is set in double quotes
+# (e.g. setting the title to "A Title" would use the syntax ``'+t"A Title"'``.
 
 fig = pygmt.Figure()
 fig.coast(
--- a/examples/get-started/first_figure.py
+++ b/examples/get-started/first_figure.py
@@ -146,7 +146,7 @@ fig.show()
 # example below. This format uses ``frame`` to set both the axes gridlines and
 # the figure title.
 #
-# If the figure title has multiple words, the string to set the title needs to
+# If the figure title has any spaces, the string to set the title needs to
 # use single-quotes, with the actual title set in double quotes (e.g. setting
 # the title to "Coastal Maine" would use the syntax ``'+t"Coastal Maine"'``.
 
--- a/examples/get-started/first_figure.py
+++ b/examples/get-started/first_figure.py
@@ -121,8 +121,7 @@ fig.show()
 # annotate the latitude and longitude of the region.
 #
 # The ``frame`` parameter is used to add a frame to the figure. For now, it
-# will be set to ``True`` to use default settings, but later tutorials will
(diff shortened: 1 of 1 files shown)


examples/get-started/first_figure.py
```diff
@@ -121,8 +121,7 @@
 # annotate the latitude and longitude of the region.
 #
 # The ``frame`` parameter is used to add a frame to the figure. For now, it
-# will be set to ``True`` to use default settings, but later tutorials will
-# show how ``frame`` can be used to customize the axes, gridlines, and titles.
+# will be set to ``"a"`` to use automatic settings.
```
seisman, line 124: `frame="a"` only plots annotations.
```suggestion
# will be set to ``"a"`` to use automatic annotation settings.
```

seisman (approved):

michaelgrund (approved):

examples/get-started/first_figure.py
```diff
@@ -131,7 +130,34 @@
...
+#
+# The ``frame`` parameter can be used to add a title to the figure. The title
+# is set with by passing ``"+t"`` followed by the title (e.g. setting the map
+# title to "Title" would be ``"+tTitle"``).
+#
+# To pass multiple arguments to ``frame``, a list can be used, as shown in the
+# example below. This format uses ``frame`` to set both the axes gridlines and
+# the figure title.
+#
+# If the figure title has multiple words, the string to set the title needs to
```
weiji14, line 149: Maybe just say spaces?
```suggestion
# If the figure title has any spaces, the string to set the title needs to
```

examples/get-started/first_figure.py
```diff
@@ -131,7 +130,34 @@
...
+# use single-quotes, with the actual title set in double quotes (e.g. setting
+# the title to "Coastal Maine" would use the syntax ``'+t"Coastal Maine"'``.
```
weiji14, line 151: Feel like this could be worded a bit better. Something along the lines of double quotes nested inside single quotes?

Or maybe I should finish up #1487 😂

willschlitzer: I think "nested" might be a confusing term for someone not familiar with Python terms.

weiji14: Ok, how about something like this?
```suggestion
# be wrapped in single-quotes, while the actual title is set in double quotes
# (e.g. setting the title to "A Title" would use the syntax ``'+t"A Title"'``.
```

examples/get-started/first_figure.py
```diff
@@ -121,8 +121,7 @@
 # annotate the latitude and longitude of the region.
 #
 # The ``frame`` parameter is used to add a frame to the figure. For now, it
-# will be set to ``True`` to use default settings, but later tutorials will
-# show how ``frame`` can be used to customize the axes, gridlines, and titles.
+# will be set to ``"a"`` to use automatic annotation settings.
```
weiji14, line 124: I want to prevent users from thinking that 'a' stands for automatic (the 'a' refers to 'annotate'). How about a sentence like this?
```suggestion
# will be set to ``"a"`` to **a**nnotate the axes automatically.
```

examples/get-started/first_figure.py
```diff
@@ -121,8 +121,7 @@
 # annotate the latitude and longitude of the region.
 #
 # The ``frame`` parameter is used to add a frame to the figure. For now, it
-# will be set to ``True`` to use default settings, but later tutorials will
-# show how ``frame`` can be used to customize the axes, gridlines, and titles.
+# will be set to ``"a"`` to **a**nnotate the axes automatically.
```
weiji14, line 124: Oops, forgot that this is RST and not markdown. How do you set bold text again?

seisman, line 124:
```suggestion
# will be set to ``"a"`` to **a**\ nnotate the axes automatically.
```

willschlitzer: Changed in 606ee98.

willschlitzer: Didn't see this :facepalm: Already pushed a change in 606ee98

willschlitzer:
@GenericMappingTools/pygmt-admin Can I get someone to merge this? Thanks!

Issue #1686, opened 2021-12-24
Add inline code examples in module docstrings


willschlitzer:
As first discussed in [this comment](https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742), should module docstrings contain inline code examples? It wouldn't create a plot on the module page, but would show a working example of the module with the correct syntax without needing to find an example page

**Are you willing to help implement and maintain this feature?** Yes/No

meghanrjones:
Regarding the CI issue raised in https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742, here are two options:

1. Use the [doctest skip directive](https://docs.pytest.org/en/latest/how-to/doctest.html#skipping-tests) to skip particular examples all the time.
2. Add a new target to the Makefile to [invoke pytest without doctest](https://docs.pytest.org/en/6.2.x/usage.html#disabling-plugins), such that the tests without docstring examples can be run on every PR and doctests could be run occasionally.

willschlitzer:
@GenericMappingTools/pygmt-maintainers Should this only be for data processing modules that can't be as easily demonstrated as a plotting function, or should there be in-line examples for all of our modules?

My vote is for all of them.

meghanrjones:
> Should this only be for data processing modules that can't be as easily demonstrated as a plotting function, or should there be in-line examples for all of our modules?

> My vote is for all of them.

Sounds good to me.

weiji14:
> > Should this only be for data processing modules that can't be as easily demonstrated as a plotting function, or should there be in-line examples for all of our modules?
> 
> > My vote is for all of them.
> 
> Sounds good to me.
> 

Ok with doing inline examples for all, but maybe prioritize those that don't already have a gallery example first. E.g. don't do `grdgradient` until much later.

weiji14:
> 1. Use the [doctest skip directive](https://docs.pytest.org/en/latest/how-to/doctest.html#skipping-tests) to skip particular examples all the time.

Just to cross-reference from #1790, @seisman has found a smart way to skip doctests now using `pytest-doctestplus` and @meghanrjones suggested in https://github.com/GenericMappingTools/pygmt/pull/1790#pullrequestreview-902000046 to document the `__doctest_skip__` and `make fulltest` methods in our contributing guides.

[...]

Issue #1666, opened 2021-12-14
Missing Examples for user facing function


jbusecke:
**Description of the desired feature**

First of all, the existing examples are really, really nice! They helped me a lot understanding the functionality.

I was however not able to find examples for `pygmt.Figure.psconvert`, `pygmt.Figure.grd2cpt`, and `pygmt.set_display`.

Furthermore many of the [processing functions](https://www.pygmt.org/latest/api/index.html#data-processing) are not show in the examples as far as I can tell.

willschlitzer:
> **Action point**: I think inline docstring might be the way to go with some of these examples. Upstream GMT has done this (e.g. at https://docs.generic-mapping-tools.org/6.3/basemap.html#examples), so we can probably translate these to PyGMT. The [`pygmt.grdfilter`](https://www.pygmt.org/v0.5.0/api/generated/pygmt.grdfilter.html) function has an example we can use as a reference. Key challenge might be to prevent these inline code examples from running in CI (but definitely check that the syntax is ok).

I agree with using inline examples. I use the GMT inline examples to demonstrate the syntax for non-plotting modules when writing the tests for module-wrapping.

willschlitzer:
@GenericMappingTools/pygmt-maintainers Would it be a good idea to add inline docstring examples (without an associated plot) to all of our module docstrings? That way the page would be a one-stop-shop without needing to go to an example.

meghanrjones:
> @GenericMappingTools/pygmt-maintainers Would it be a good idea to add inline docstring examples (without an associated plot) to all of our module docstrings? That way the page would be a one-stop-shop without needing to go to an example.

Yes to me, as long as we can skip the inline examples in the CI and locally when needed.

PR #1689, opened 2021-12-30
Add inline example for grdcut

willschlitzer:
This PR adds an inline code example to the docstring for `grdcut`.

Addresses #1686

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdcut.py
+++ b/pygmt/src/grdcut.py
@@ -90,15 +90,15 @@ def grdcut(grid, **kwargs):
 
     Examples
     --------
-    >>> import pygmt
+    >>> import pygmt # doctest: +SKIP
     >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
     >>> # and a y-range of 15 to 25
-    >>> grid = pygmt.datasets.load_earth_relief(
-    ...     resolution="30m", region=[10, 30, 15, 25]
-    ... )
+    >>> grid = pygmt.datasets.load_earth_relief( # doctest: +SKIP
+    ...     resolution="30m", region=[10, 30, 15, 25] # doctest: +SKIP
+    ... ) # doctest: +SKIP
     >>> # Create a new grid from an input grid, with an x-range of 12 to 15,
     >>> # and a y-range of 21 to 24
-    >>> new_grid = pygmt.grdcut(grid=grid, region=[12, 15, 21, 24])
+    >>> new_grid = pygmt.grdcut(grid=grid, region=[12, 15, 21, 24]) # doctest: +SKIP
     """
     with GMTTempFile(suffix=".nc") as tmpfile:
         with Session() as lib:
--- a/pygmt/src/grdcut.py
+++ b/pygmt/src/grdcut.py
@@ -91,7 +91,6 @@ def grdcut(grid, **kwargs):
     Examples
     --------
     >>> import pygmt
-
     >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
     >>> # and a y-range of 15 to 25
     >>> grid = pygmt.datasets.load_earth_relief(
--- a/pygmt/src/grdcut.py
+++ b/pygmt/src/grdcut.py
@@ -90,15 +90,17 @@ def grdcut(grid, **kwargs):
 
     Examples
     --------
-    >>> import pygmt # doctest: +SKIP
+    >>> import pygmt  # doctest: +SKIP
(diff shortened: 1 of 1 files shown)


pygmt/src/grdcut.py
```diff
@@ -87,6 +87,18 @@ def grdcut(grid, **kwargs):
...
+    --------
+    >>> import pygmt
+    >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
+    >>> # and a y-range of 15 to 25
+    >>> grid = pygmt.datasets.load_earth_relief(
+    ...     resolution="30m", region=[10, 30, 15, 25]
+    ... )
+    >>> # Create a new grid from an input grid, with an x-range of -3 to 1,
+    >>> # and a y-range of 2 to 5
+    >>> new_grid = pygmt.grdcut(grid=grid, region=[-3, 1, 2, 5])
```
meghanrjones, line 101: Thanks for moving forward with the in-line examples!

I would prefer that the example using a different dataset than the result from `load_earth_relief`. We could add a raster dataset to the proposed `load_sample_data` function (https://github.com/GenericMappingTools/pygmt/pull/1685). I think that `grid = load_sample_data(name="relief")` (hypothetically for `@tut_relief`) is more intuitive than relying on`load_earth_relief`.

seisman: > I think that `grid = load_sample_data(name="relief")` (hypothetically for `@tut_relief.nc`) is more intuitive than relying on`load_earth_relief`.

Why do you think so? I think cutting a earth relief grid is a very common case, and users can simply use it without changes.

willschlitzer: I agree with @seisman on this. I think the examples are good to demonstrate how a user could get the code running, and it's better to demonstrate using the `load_earth_relief` for a usable example.

meghanrjones, line 101:
```suggestion
    >>> # Create a new grid from an input grid, with an x-range of 12 to 15,
    >>> # and a y-range of 21 to 24
    >>> new_grid = pygmt.grdcut(grid=grid, region=[12, 15, 21, 24])
```
This suggestion should fix the failing tests due to the second region not being in the first.

willschlitzer: :facepalm: numbers are difficult

meghanrjones:
The example looks nice! 

Do you want to figure out the CI separation mentioned [here](https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742) as part of this RFC?

willschlitzer:
> The example looks nice!
> 
> Do you want to figure out the CI separation mentioned [here](https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742) as part of this RFC?

@meghanrjones I decided to use `# doctest: +SKIP`; I think the better long term solution is to edit the makefile, but I don't want to go outside the scope of this PR.

meghanrjones:
> > The example looks nice!
> > Do you want to figure out the CI separation mentioned [here](https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742) as part of this RFC?
> 
> @meghanrjones I decided to use `# doctest: +SKIP`; I think the better long term solution is to edit the makefile, but I don't want to go outside the scope of this PR.

It's nice that the doctest comments are not included in the rendered html docs: https://pygmt-5z64fvi6h-gmt.vercel.app/api/generated/pygmt.grdcut.html#pygmt.grdcut.

meghanrjones (approved):
Looks good to me!

weiji14 (approved):
> > > The example looks nice!
> > > Do you want to figure out the CI separation mentioned [here](https://github.com/GenericMappingTools/pygmt/issues/1666#issuecomment-993962742) as part of this RFC?
> > 
> > @meghanrjones I decided to use `# doctest: +SKIP`; I think the better long term solution is to edit the makefile, but I don't want to go outside the scope of this PR.
> 
> It's nice that the doctest comments are not included in the rendered html docs: https://pygmt-5z64fvi6h-gmt.vercel.app/api/generated/pygmt.grdcut.html#pygmt.grdcut.

Wow, very handy that they hide the `# doctest: +SKIP` comment! So much for overthinking. Let's hope that Sphinx keeps this as a 'feature' rather than a 'bug' :laughing:

seisman:
Is there any way to skip the entire code block using a single directive?

willschlitzer:
> Is there any way to skip the entire code block using a single directive?

The example I found online involves wrapping the example in a function, and then calling that function but then adding `doctest: +SKIP`. I think that would complicate the example code if we tried that.

seisman (approved):

PR #1706, opened 2022-01-17
Add clip parameter to pygmt.Figure.coast

michaelgrund:

This PR adds a ``clip`` parameter to the Figure.coast module:

* ``clip="land"`` to clip dry areas
* ``clip="water"`` to clip wet areas
* ``clip ="end"`` to mark end of existing clip path

Fixes #

**Reminders**

**Slash Commands**


meghanrjones:
I like how GMT.jl handles clipping through the coast module (https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.coast) using a `clip` parameter that controls both starting and ending clip paths. In Python form, `clip="land"` starts clip path for land, `clip="water"` starts a clip path for water, and `clip="end"` ends the clip path. I prefer this over `land="True"` and `water="True"` to start a clip path and `end_clip="True"` to end the clip path.

michaelgrund:
> I like how GMT.jl handles clipping through the coast module (https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.coast) using a `clip` parameter that controls both starting and ending clip paths. In Python form, `clip="land"` starts clip path for land, `clip="water"` starts a clip path for water, and `clip="end"` ends the clip path. I prefer this over `land="True"` and `water="True"` to start a clip path and `end_clip="True"` to end the clip path.

Sounds like a perfect solution to me, thanks @meghanrjones!

pygmt/src/coast.py
```diff
@@ -183,10 +188,22 @@ def coast(self, **kwargs):
...
+
+    if "clip" in kwargs and kwargs["clip"][0] in "land":
+        kwargs["G"] = True
+        kwargs.pop("clip")
+    elif "clip" in kwargs and kwargs["clip"][0] in "water":
+        kwargs["S"] = True
+        kwargs.pop("clip")
+    elif "clip" in kwargs and kwargs["clip"][0] in "end":
+        kwargs["Q"] = True
+        kwargs.pop("clip")
```
weiji14, line 200: Could you explain what `kwargs["clip"][0] in "land"` does? I don't understand why you need to take the first value (`[0]`), and why the check is using `in` rather than `==`.

weiji14 (commented):
Need to also add some unit tests to `test_coast.py` to cover the clip parameter. Maybe one for a land clip and one for a water clip.

pygmt/src/coast.py
michaelgrund: Not sure what I did there, thanks @weiji14 😉.

michaelgrund:
> Need to also add some unit tests to `test_coast.py` to cover the clip parameter. Maybe one for a land clip and one for a water clip.

Tests are added @weiji14.

pygmt/src/coast.py
```diff
@@ -159,6 +159,11 @@ def coast(self, **kwargs):
         lake-in-island-in-lake shore. Pass a list of *level*\ /*pen*
         strings to ``shorelines`` to set multiple levels. When specific
         level pens are set, those not listed will not be drawn.
+    clip : str
+        To clip land do ``clip=land``, ``clip=water`` clips water. Use
+        ``clip=end`` to mark end of existing clip path. No projection
```
seisman, line 164:
```suggestion
        To clip land do ``clip="land"``, ``clip="water"`` clips water. Use
        ``clip="end"`` to mark end of existing clip path. No projection
```

pygmt/src/coast.py
```diff
@@ -183,10 +188,22 @@ def coast(self, **kwargs):
...
+
+    if "clip" in kwargs and kwargs["clip"] == "land":
+        kwargs["G"] = True
+        kwargs.pop("clip")
+    elif "clip" in kwargs and kwargs["clip"] == "water":
+        kwargs["S"] = True
+        kwargs.pop("clip")
+    elif "clip" in kwargs and kwargs["clip"] == "end":
+        kwargs["Q"] = True
+        kwargs.pop("clip")
```
seisman, line 200: No need to check `"clip" in kwargs` in three if-branches:
```suggestion
    if "clip" in kwargs:
        if kwargs["clip"] == "land":
            kwargs["G"] = True
            kwargs.pop("clip")
        elif kwargs["clip"] == "water":
            kwargs["S"] = True
            kwargs.pop("clip")
        elif kwargs["clip"] == "end":
            kwargs["Q"] = True
            kwargs.pop("clip")
```

michaelgrund:
Btw, should we also use the static_earth_relief here?

pygmt/src/coast.py
```diff
@@ -183,10 +188,23 @@ def coast(self, **kwargs):
...
+    if "clip" in kwargs:
+        if kwargs["clip"] == "land":
+            kwargs["G"] = True
+            kwargs.pop("clip")
+        elif kwargs["clip"] == "water":
+            kwargs["S"] = True
+            kwargs.pop("clip")
+        elif kwargs["clip"] == "end":
+            kwargs["Q"] = True
+            kwargs.pop("clip")
```
seisman, line 201: Check if the given clip parameter is valid.
```suggestion
    if "clip" in kwargs:
        if kwargs["clip"] in ["land", "water", "end"]:
            kwargs.pop("clip")
            if kwargs["clip"] == "land":
                kwargs["G"] = True
            elif kwargs["clip"] == "water":
                kwargs["S"] = True
            elif kwargs["clip"] == "end":
                kwargs["Q"] = True
        else:
            raise GMTInvalidInput(
                "Invalid clip parameter. Must be one of 'land', 'water', and 'end'."
            )  
```

pygmt/tests/test_coast.py
```diff
@@ -72,3 +73,35 @@ def test_coast_dcw_list():
...
+    Test to clip dry areas.
+    """
+    region = [-28, -10, 62, 68]
+    grid = load_earth_relief(resolution="10m", region=region)
+
+    fig = Figure()
+    fig.basemap(region=region, projection="M12c", frame=True)
+    fig.coast(resolution="l", clip="land")
+    fig.grdimage(grid=grid, shading=True, cmap="batlow")
+    fig.coast(clip="end", frame=True)
```
seisman, line 90:
```suggestion
    fig.coast(clip="end")
```

pygmt/tests/test_coast.py
```diff
@@ -72,3 +73,35 @@ def test_coast_dcw_list():
...
+    Test to clip wet areas.
+    """
+    region = [-28, -10, 62, 68]
+    grid = load_earth_relief(resolution="10m", region=region)
+
+    fig = Figure()
+    fig.basemap(region=region, projection="M12c", frame=True)
+    fig.coast(resolution="l", clip="water")
+    fig.grdimage(grid=grid, shading=True, cmap="batlow")
+    fig.coast(clip="end", frame=True)
```
seisman, line 106:
```suggestion
    fig.coast(clip="end")
```

seisman:
> Btw, should we also use the static_earth_relief here?

I think yes.

pygmt/tests/test_coast.py
```diff
@@ -72,3 +73,35 @@ def test_coast_dcw_list():
...
+    """
+    Test to clip dry areas.
+    """
+    region = [-28, -10, 62, 68]
+    grid = load_earth_relief(resolution="10m", region=region)
+
+    fig = Figure()
+    fig.basemap(region=region, projection="M12c", frame=True)
+    fig.coast(resolution="l", clip="land")
+    fig.grdimage(grid=grid, shading=True, cmap="batlow")
```
seisman, line 89: Setting `shading=True` is unnecessary and makes the test slower, so remove it.
```suggestion
    fig.grdimage(grid=grid, cmap="batlow")
```

pygmt/tests/test_coast.py
```diff
@@ -72,3 +73,35 @@ def test_coast_dcw_list():
...
+    """
+    Test to clip wet areas.
+    """
+    region = [-28, -10, 62, 68]
+    grid = load_earth_relief(resolution="10m", region=region)
+
+    fig = Figure()
+    fig.basemap(region=region, projection="M12c", frame=True)
+    fig.coast(resolution="l", clip="water")
+    fig.grdimage(grid=grid, shading=True, cmap="batlow")
```
seisman, line 105:
```suggestion
    fig.grdimage(grid=grid, cmap="batlow")
```

pygmt/src/coast.py
```diff
@@ -183,10 +188,26 @@ def coast(self, **kwargs):
...
+                kwargs["G"] = True
+            elif kwargs["clip"] == "water":
+                kwargs["S"] = True
+            elif kwargs["clip"] == "end":
+                kwargs["Q"] = True
+            kwargs.pop("clip")
+        else:
+            raise GMTInvalidInput(
+                "Invalid clip parameter. Must be one of 'land', 'water', and 'end'."
+            )
```
seisman, line 204: Please also add a test to make sure `GMTInvalidInput` is raised when invalid parameter like `clip="invalid"` is given.

pygmt/tests/test_coast.py
```diff
@@ -3,6 +3,7 @@
 """
 import pytest
 from pygmt import Figure
+from pygmt.datasets import load_earth_relief
 from pygmt.exceptions import GMTInvalidInput
```
weiji14, line 7: > >  Btw, should we also use the static_earth_relief here?
> I think yes.

The static_earth_relief covers the whole Earth though, do we need to spatially subset it to some island?

```suggestion
from pygmt.exceptions import GMTInvalidInput
from pygmt.helpers.testing import load_static_earth_relief
```

pygmt/tests/test_coast.py
weiji14: I do like the Scientific Colour Maps, but because those colourmaps have a tendency to get updated with minor tweaks once in a while, maybe use one of the standard GMT CPTs at https://docs.generic-mapping-tools.org/6.3/cookbook/cpts.html#id2 that doesn't change (e.g. 'relief') so we don't need to update this test when a new SCM version lands.

pygmt/src/coast.py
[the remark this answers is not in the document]
seisman: > have `def coast(self, clip=None, **kwargs):`

I like it.

> there might be an issue if someone tries to use fig.coast(..., Q=True) (I know I probably used it somewhere before)

I think we should check if any of `Q`, `G`, `S` is given when `clip` is used.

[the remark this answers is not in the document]
michaelgrund: Sounds good to me. Thanks for all the suggestions 😉

pygmt/tests/test_coast.py
michaelgrund: The static_earth_relief does not cover the whole Earth (please correct me if I'm wrong @weiji14 ). Moreover, no water areas are currently covered. I'm not sure how we should continue. Adjust the static_earth_relief grid or continue with load_earth_relief?

pygmt/tests/test_coast.py
```diff
@@ -72,3 +73,35 @@ def test_coast_dcw_list():
...
     return fig
+
+
+@pytest.mark.mpl_image_compare
+def test_coast_clip_land():
+    """
+    Test to clip dry areas.
+    """
+    region = [-28, -10, 62, 68]
+    grid = load_earth_relief(resolution="10m", region=region)
```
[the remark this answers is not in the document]
michaelgrund: Done!

pygmt/src/coast.py
michaelgrund: Done!

seisman (approved):

pygmt/src/coast.py
```diff
@@ -159,6 +159,11 @@ def coast(self, **kwargs):
         lake-in-island-in-lake shore. Pass a list of *level*\ /*pen*
         strings to ``shorelines`` to set multiple levels. When specific
         level pens are set, those not listed will not be drawn.
+    clip : str
+        To clip land do ``clip="land"``, ``clip="water"`` clips water. Use
+        ``clip="end"`` to mark end of existing clip path. No projection
+        information is needed. Also supply ``xshift`` and ``yshift`` settings
+        if you have moved since the clip started.
```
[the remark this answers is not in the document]
seisman: > clip' actually refers to clipping of the next `grdimage` call.

That's not true. Everything, including symbols, texts, images, contours are clipped. So the clipping affects many plotting modules.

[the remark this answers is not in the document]
michaelgrund: Added an inline example.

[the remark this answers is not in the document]
weiji14: > > clip' actually refers to clipping of the next `grdimage` call.
> 
> That's not true. Everything, including symbols, texts, images, contours, are clipped. So the clipping affects many plotting modules.

Oh really? Then why are we making a unit test that uses `fig.grdimage` and `load_earth_age` instead of a simple `fig.plot`...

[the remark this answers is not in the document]
weiji14: Just to be clear, I think the inline example is good. But maybe do 2 more things:

1. Reword the docstring to explicitly mention that 'clip' is for clipping the next `grdimage`/`plot`/`contour`/etc call
2. Refactor the unit test to just use `fig.plot` or something instead of `fig.grdimage`, so that `load_earth_age` won't be needed (really sorry for the back and forth @michaelgrund!)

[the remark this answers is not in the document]
michaelgrund: > Just to be clear, I think the inline example is good. But maybe do 2 more things:
> 
>     1. Reword the docstring to explicitly mention that 'clip' is for clipping the next `grdimage`/`plot`/`contour`/etc call

Done! Used parts of @meghanrjones'  answer.

>     2. Refactor the unit test to just use `fig.plot` or something instead of `fig.grdimage`, so that `load_earth_age` won't be needed (really sorry for the back and forth @michaelgrund!)

Done! Now, simply plotted circles are used without the need of any dataset 😉

seisman:
> Just a note regarding comment number 1 - the clip path applies to **all** plotting calls between the start of the clip path and the end of a clip bath. tbh I should have suggested a context manager for this case which would have been a more Pythonic way to manage clip="end".

@meghanrjones Do you still think we should use a context manager for clipping?

seisman:
> My one concern with the current implementation is that while backwards compatibility for Q=True is helpful for the creative users who've figured out that undocumented option, I imagine that people could try something like `land=True` expecting some default settings for plotting land masses and be confused when instead all their data are clipped at the land borders. One option around this could be to issue deprecation warnings for Q=True, land=True, or water=True, which could be handled in a separate PR.

I agree.

meghanrjones:
> > Just a note regarding comment number 1 - the clip path applies to **all** plotting calls between the start of the clip path and the end of a clip bath. tbh I should have suggested a context manager for this case which would have been a more Pythonic way to manage clip="end".
> 
> @meghanrjones Do you still think we should use a context manager for clipping?

Possibly, I will look into this idea a bit more based the gmt clip module, which is very similar to these combinations of options to coast.

meghanrjones:
> > Just a note regarding comment number 1 - the clip path applies to all plotting calls between the start of the clip path and the end of a clip bath. tbh I should have suggested a context manager for this case which would have been a more Pythonic way to manage clip="end".

> @meghanrjones Do you still think we should use a context manager for clipping?

Yes, after testing the feasibility with https://github.com/GenericMappingTools/pygmt/pull/1779, I still think that using a context manager for clipping land, water, and countries would be a good idea. @michaelgrund, I am sorry for not having my thoughts in order on the first review of this PR.

Here are the various options that I know of that set clipping paths in GMT:

- clip (polygon clipping)
- cloast -E+c|C (DCW i.e., geographic clipping)
- coast -G (land clipping)
- coast -S (water clipping)
- mask (clip based on data coverage)

I think it would be better to have this functionality organized together in the PyGMT API rather than spread out amongst the other functions, but I'm not sure what the best way to accomplish that would be.

michaelgrund:
> Yes, after testing the feasibility with #1779, I still think that using a context manager for clipping land, water, and countries would be a good idea. @michaelgrund, I am sorry for not having my thoughts in order on the first review of this PR.
> 
> Here are the various options that I know of that set clipping paths in GMT:
> 
>     * clip (polygon clipping)
> 
>     * cloast -E+c|C (DCW i.e., geographic clipping)
> 
>     * coast -G (land clipping)
> 
>     * coast -S (water clipping)
> 
>     * mask (clip based on data coverage)
> 
> 
> I think it would be better to have this functionality organized together in the PyGMT API rather than spread out amongst the other functions, but I'm not sure what the best way to accomplish that would be.

Totally agree with you @meghanrjones, should be bundled together. How should we continue with this PR?

meghanrjones:
> Totally agree with you @meghanrjones, should be bundled together. How should we continue with this PR?

I recommend leaving it as is briefly until we find out whether GMT can internally handle the -X -Y shifts when using clip paths (https://github.com/GenericMappingTools/gmt/issues/6406) because the result from that discussion could impact our design of these functions.

michaelgrund:
I saw the corresponding GMT PR was merged with only a fix in the docs @maxrjones. Do you think we can continue here?

maxrjones:
> I saw the corresponding GMT PR was merged with only a fix in the docs @maxrjones. Do you think we can continue here?

Yes, continuing sounds good. Here's how I imagine the API looking with the various clip options being organized in a single namespace:
```
import pygmt
fig = pygmt.Figure()
with fig.clip.land():
    # Plot data with land clipped
with fig.clip.water():
    # Plot data with water clipped
with fig.clip.polygon(...):
    # Plot data with polygons clipped
with fig.clip.dcw(...):
    # Plot data with clipping based on dcw
```
Optionally, we could split up dcw, e.g., (`fig.clip.country`, `fig.clip.continent`, `fig.clip.collection`)

seisman:
> > > Yes, continuing sounds good. Here's how I imagine the API looking with the various clip options being organized in a single namespace:
> > > ```
> > > import pygmt
> > > fig = pygmt.Figure()
> > > with fig.clip.land():
> > >     # Plot data with land clipped
> > > with fig.clip.water():
> > >     # Plot data with water clipped
> > > with fig.clip.polygon(...):
> > >     # Plot data with polygons clipped
> > > with fig.clip.dcw(...):
> > >     # Plot data with clipping based on dcw
> > > ```
> > >   
> > > Optionally, we could split up dcw, e.g., (`fig.clip.country`, `fig.clip.continent`, `fig.clip.collection`)
> > 
> > 
> > This looks like a good API design. If we decide to follow this API degisn, I think we should close this PR and then work in #1779 instead. Agree?
> 
> Is it possible, or would there be a need though to do a combined clip? E.g. some polygon + dcw?

That's a good question. GMT does support nested clipping paths:
```
gmt begin map
gmt coast -Rg -JH15c -W0.2p -Baf -Y40c
gmt coast -G
	gmt grdimage @earth_relief_01d
gmt coast -Q

gmt coast -Rg -JH15c -W0.2p -Baf -Y-8c
gmt coast -G
	gmt coast -ECN+c
		gmt grdimage @earth_relief_01d
	gmt coast -Q
gmt coast -Q

gmt coast -Rg -JH15c -W0.2p -Baf -Y-8c
gmt coast -G
	gmt clip <<- EOF
	110 30
	110 60
	180 60
	180 30
	EOF
		gmt grdimage @earth_relief_01d
	gmt clip -C
gmt coast -Q

gmt coast -Rg -JH15c -W0.2p -Baf -Y-8c
gmt coast -G
	gmt coast -ECN+c
		gmt clip <<- EOF
		110 30
		110 60
		180 60
		180 30
		EOF
		gmt grdimage @earth_relief_01d
		gmt clip -C
	gmt coast -Q
gmt coast -Q

gmt end show
```
The output is:
map

So nested clipping in PyGMT should be like:
```
with fig.clip.land():
    # Plot data with land clipped
    with fig.clip.polygon(...):
	    # Plot data with polygons clipped
```

seisman:
Closing the issue as it will be superseded by https://github.com/GenericMappingTools/pygmt/pull/3878.

PR #1711, opened 2022-01-28
Add inline example for grdclip

willschlitzer:
This PR adds an inline example for `grdclip`.

Addresses: #1686 

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -77,6 +77,20 @@ def grdclip(grid, **kwargs):
         - :class:`xarray.DataArray` if ``outgrid`` is not set
         - None if ``outgrid`` is set (grid output will be stored in file set by
           ``outgrid``)
+
+    Examples
+    --------
+    >>> import pygmt  # doctest: +SKIP
+    >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
+    >>> # and a y-range of 15 to 25
+    >>> grid = pygmt.datasets.load_earth_relief(
+    ...     resolution="30m", region=[10, 30, 15, 25]
+    ... )  # doctest: +SKIP
+    >>> # Create a new grid from an input grid. Set all values below 1,000 to
+    >>> # 0 and all values above 1,500 to 10,000
+    >>> new_grid = pygmt.grdclip(
+    ...     grid=grid, below=[1000, 0], above=[1500, 10000]
+    ... )  # doctest: +SKIP
     """
     with GMTTempFile(suffix=".nc") as tmpfile:
         with Session() as lib:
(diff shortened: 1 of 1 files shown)


pygmt/src/grdclip.py
```diff
@@ -77,6 +77,20 @@ def grdclip(grid, **kwargs):
...
+    >>> # and a y-range of 15 to 25
+    >>> grid = pygmt.datasets.load_earth_relief(
+    ...     resolution="30m", region=[10, 30, 15, 25]
+    ... )  # doctest: +SKIP
+    >>> # Create a new grid from an input grid. Set all values below 1,000 to
+    >>> # 0 and all values above 1,500 to 10,000
+    >>> new_grid = pygmt.grdclip(
+    ...     grid=grid, below=[1000, 0], above=[1500, 10000]
+    ... )  # doctest: +SKIP
     """
```
meghanrjones, line 94:
```suggestion
    >>> # Report the minimum and maximum data values
    >>> [new_grid.data.min(), new_grid.data.max()]  # doctest: +SKIP
    [0.0, 10000.0]
    """
```

meghanrjones (commented):
What do you think about adding a couple lines to report the range of data values (suggestions below)? This might help people understand the impact on the data and would be essential if we ever want to include these in the tests.

willschlitzer:
> What do you think about adding a couple lines to report the range of data values (suggestions below)? This might help people understand the impact on the data and would be essential if we ever want to include these in the tests.

I'm not a big fan of using pseudo-print statements to show the effect of the `grdclip`. Should we make it more like the GMT example and have more explanation in the text?

meghanrjones:
> > What do you think about adding a couple lines to report the range of data values (suggestions below)? This might help people understand the impact on the data and would be essential if we ever want to include these in the tests.
> 
> I'm not a big fan of using pseudo-print statements to show the effect of the `grdclip`. Should we make it more like the GMT example and have more explanation in the text?

I think this way of displaying the output to show and test results in docstring examples is really common for scientific Python libraries e.g., [geopandas.DataFrame.clip](https://geopandas.org/en/stable/docs/reference/api/geopandas.clip.html) - [docstring source](https://github.com/geopandas/geopandas/blob/a80081e52ed3bd069a75a4caaed44e45e35c647d/geopandas/geodataframe.py#L2106-L2120) and [pandas.DataFrame.clip](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.clip.html) - [dostring source](https://github.com/pandas-dev/pandas/blob/f956bb5e7e6fa5d3a8161442b26b44058d4b61a9/pandas/core/generic.py#L7478-L7536).

seisman:
I agree with @meghanrjones

willschlitzer:
@seisman @meghanrjones Sounds good!

weiji14 (approved):

Issue #1684, opened 2021-12-22
Use a static grid, asymmetric region, and xarray.testing for raster tests


meghanrjones:
This issue was originally for discussing improvements to the PyGMT test suite and is now dedicated to tracking specific improvements:

1. Use xarray.testing rather than pygmt.grdinfo
2. Use `static_earth_relief` rather than `earth_relief`
3. Use asymmetric regions to facilitate identification of lat/lon errors

If anyone wants to help, please edit this comment to add your username next to the module or post a comment below to avoid redundant work.

- [x] Add static_earth_relief to cache (https://github.com/GenericMappingTools/gmtserver-admin/pull/139)
- [x] Add function to load static_earth_relief.nc as an xarray.DataArray (#1727)
- [x] test_grd2cpt.py (static_earth_relief) (@willschlitzer #1736)
- [x] test_grd2xyz.py (static_earth_relief) (@willschlitzer #1737)
- [x] test_grdclip.py (xarray.testing) (@meghanrjones, #1573)
- [x] test_grdclip.py (static_earth_relief) (@willschlitzer #1738)
- [x] text_grdcontour.py (static_earth_relief) (@willschlitzer #1741)
- [x] test_grdcut.py (xarray.testing) (@meghanrjones, #1574)
- [x]  test_grdcut.py (static_earth_relief) (@willschlitzer #1742)
- [x] test_grdfill.py (xarray.testing) (@willschlitzer, #1677)
- [x] test_grdfill.py (static_earth_relief) (@willschlitzer #1743)
- [x] test_grdfilter.py (xarray.testing and static_earth_relief) (@willschlitzer, #1678)
- [x] test_grdgradient.py (xarray.testing and static_earth_relief) (@willschlitzer, #1679)
- [x] test_grdinfo.py (static_earth_relief) (@willschlitzer #1746)
- [x] test_grdlandmask.py (xarray.testing) (@willschlitzer, #1680)
- [x] test_grdproject.py (xarray.testing and static_earth_relief) (@willschlitzer #1683)
- [x] test_grdsample.py (xarray.testing and static_earth_relief) (@willschlitzer, #1681)
- [x] test_grdtrack.py (static_earth_relief) (@willschlitzer, #1762)
- [x] test_grdview.py (static_earth_relief) (@willschlitzer #1752)
- [x] test_grdvolume.py (static_earth_relief) (@willschlitzer #1747)
[...]

PR #1727, opened 2022-02-06
Add load_static_earth_relief function for internal testing

willschlitzer:
This PR adds the `@static_earth_relief` to functions in `samples.py`.

Addresses [#1684 (comment)](https://github.com/GenericMappingTools/pygmt/issues/1684#issuecomment-1030874123)

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -4,7 +4,6 @@ Functions to load sample data.
 import warnings
 
 import pandas as pd
-import xarray as xr
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.src import which
 
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -4,6 +4,7 @@ Functions to load sample data.
 import warnings
 
 import pandas as pd
+import xarray as xr
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.src import which
 
@@ -351,15 +352,11 @@ def _load_static_earth_relief(**kwargs):
     """
     Load the static_earth_relief file for internal testing.
 
-    Parameters
-    ----------
-    kwargs
-
     Returns
     -------
+    data : xarray.DataArray
+        A grid of Earth relief for internal tests.
     """
-    fname = which("@static_earth_relief", download="c")
-    data = pd.read_csv(
-        fname, sep="\t", header=None, names=["longitude", "latitude", "bathymetry"]
-    )
+    fname = which("@static_earth_relief.nc", download="c")
+    data = xr.open_dataarray(fname)
     return data
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -71,7 +71,6 @@ def load_sample_data(name):
         "japan_quakes": load_japan_quakes,
         "mars_shape": load_mars_shape,
         "ocean_ridge_points": load_ocean_ridge_points,
-        "static_earth_relief": _load_static_earth_relief,
         "usgs_quakes": load_usgs_quakes,
     }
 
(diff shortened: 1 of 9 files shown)


pygmt/tests/test_datasets_samples.py
```diff
@@ -145,3 +147,17 @@ def test_hotspots():
...
+    """
+    Check that @static_earth_relief.nc loads without errors.
+    """
+    data = load_sample_data("static_earth_relief")
+    assert data.dims == ("lat", "lon")
+    assert data.shape == (14, 8)
+    assert data.min() == 190
+    assert data.max() == 981
+    assert data.median() == 467
+    npt.assert_allclose(data.mean(), 490.87946)
```
meghanrjones, line 162:
```suggestion
```
I don't think this adds much to the other `min, max, and median` assertions, so we could leave it out to cut down on imports.

pygmt/tests/test_datasets_samples.py
```diff
@@ -1,8 +1,10 @@
 """
 Test basic functionality for loading sample datasets.
 """
+import numpy.testing as npt
```
meghanrjones, line 4:
```suggestion
```

pygmt/datasets/samples.py
```diff
@@ -343,3 +346,17 @@ def load_mars_shape(**kwargs):
...
+    Load the static_earth_relief file for internal testing.
+
+    Returns
+    -------
+    data : xarray.DataArray
+        A grid of Earth relief for internal tests.
+    """
+    fname = which("@static_earth_relief.nc", download="c")
+    data = xr.open_dataarray(fname)
+    return data
```
meghanrjones, line 362:
```suggestion
    return xr.open_dataarray(fname)
```
In addition to consolidating these two lines, I think you should consider using `from pygmt.io import load_dataarray` rather than `import xarray as xr` and `load_arrayarray(fname)` rather than `xr.open_datarray(fname)`. Right now, these two options return the same ouput. However, we could extend the load_dataarray function to add other important metadata (e.g., citations) from the NetCDF global attributes to xarray.DataArray.attrs whereas open_dataarray is less flexible for future enhancements.

pygmt/datasets/samples.py
```diff
@@ -343,3 +346,17 @@ def load_mars_shape(**kwargs):
         fname, sep="\t", header=None, names=["longitude", "latitude", "radius(m)"]
     )
     return data
+
+
+def _load_static_earth_relief(**kwargs):
```
meghanrjones, line 351:
```suggestion
def _load_static_earth_relief(**kwargs):  # pylint: disable=unused-argument
```

pygmt/datasets/samples.py
```diff
@@ -30,6 +31,7 @@ def list_sample_data():
         "mars_shape": "Table of topographic signature of the hemispheric dichotomy of "
         " Mars from Smith and Zuber (1996)",
         "ocean_ridge_points": "Table of ocean ridge points for the entire world",
+        "static_earth_relief": "Sample grid used for testing",
```
meghanrjones, line 34:
```suggestion
        "static_earth_relief": "Sample grid used for testing based on Sandwell et al., 2022",
```

weiji14: Do we actually want `list_sample_data()` to print this `static_earth_relief` file? I thought this was meant to be non-user-facing?

weiji14 (commented):
Also need to add `@static_earth_relief` to the cache.

pygmt/datasets/samples.py
meghanrjones: I suggested adding it to both for consistency with the other functions for loading data from the cache, but I don't have a strong preference for it to be an user-facing feature. If we do have it as a strictly internal feature, would it be better in pygmt.helpers.testing rather than this module?

pygmt/datasets/samples.py
willschlitzer: I'll switch the function to use `load_dataarray`.

pygmt/datasets/samples.py
willschlitzer: I didn't even think about using it in helpers, but I think it makes more sense there. We don't intend for this grid to be used by any users.

meghanrjones (approved):
Looks good, thanks! I think the test function could go in either test_datasets_samples.py or test_helpers.py and am fine with whichever you think is more appropriate.

michaelgrund (approved):

willschlitzer:
> Looks good, thanks! I think the test function could go in either test_datasets_samples.py or test_helpers.py and am fine with whichever you think is more appropriate.

I think test_helpers.py makes more sense; moving the test

seisman (approved):

PR #1743, opened 2022-02-11
Change test_grdfill.py to use static_earth_relief

willschlitzer:
This modifies `test_grdfill.py` to use `load_static_earth_relief`.

Addresses #1684 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_grdfill.py
+++ b/pygmt/tests/test_grdfill.py
@@ -7,20 +7,20 @@ import numpy as np
 import pytest
 import xarray as xr
 from pygmt import grdfill, load_dataarray
-from pygmt.datasets import load_earth_relief
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import GMTTempFile
+from pygmt.helpers.testing import load_static_earth_relief
 
 
 @pytest.fixture(scope="module", name="grid")
 def fixture_grid():
     """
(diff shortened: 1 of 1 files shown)


willschlitzer:
I tried modifying the `region` parameter to have a smaller grid, but I kept getting errors that would crash my IPython/Jupyter kernel. I can't figure out why this is; it's not a problem on other data processing modules, including grdcut.

meghanrjones:
> I tried modifying the region parameter to have a smaller grid, but I kept getting errors that would crash my IPython/Jupyter kernel. I can't figure out why this is; it's not a problem on other data processing modules, including grdcut.

That's not good. Do you want to open a separate issue for this? Since it will probably require a GMT release to be fixed, I think we can just use the full grid for these tests.

meghanrjones (approved):

seisman (approved):

willschlitzer:
> > I tried modifying the region parameter to have a smaller grid, but I kept getting errors that would crash my IPython/Jupyter kernel. I can't figure out why this is; it's not a problem on other data processing modules, including grdcut.
> 
> That's not good. Do you want to open a separate issue for this? Since it will probably require a GMT release to be fixed, I think we can just use the full grid for these tests.

Opened at #1745

PR #1748, opened 2022-02-12
Change test_makecpt.py to use colorbar instead of grdimage

willschlitzer:
This modifies `test_makecpt.py` to use `load_static_earth_relief`.

Addresses #1684 

**Reminders**

**Slash Commands**


seisman:
In most tests, we call `makecpt` to create a CPT file and call `grdimage` to plot the earth relief grid. I feel it's very difficult to verify if the grid image is correct or not.

Why not just call `colorbar` to plot the color bar instead? It's more intuitive to see if the plotted color bar matches the `makecpt` command.

seisman:
BTW, please change the PR title to "Change test_xxxx.py to use static_earth_relief".

pygmt/tests/test_makecpt.py
```diff
@@ -57,25 +48,25 @@ def test_makecpt_plot_points(points, region):
...
 @pytest.mark.mpl_image_compare
-def test_makecpt_plot_grid(grid):
+def test_makecpt_plot_grid(region):
     """
     Use static color palette table to change color of grid.
     """
     fig = Figure()
     makecpt(cmap="relief")
-    fig.grdimage(grid, projection="W0/10c")
+    fig.colorbar(cmap=True, region=region, frame=True, position="JBC")
```
seisman, line 57: Is `region` required here?

willschlitzer: I get this error `colorbar [ERROR]: Option -DJ requires the -R option` when I only use `fig.colorbar(cmap=True, frame=True, position="JBC")`

seisman: Yes, `-DJ` (or `position="JBC"`) needs a `region`. For internal tests, maybe `fig.colorbar(position="x0c/0c+w10c")` is enough.

meghanrjones (commented):
Looks good after removing the ``region`` pytest fixture.

meghanrjones:
> Looks good after removing the `region` pytest fixture.

Actually, I just noticed that this is used in ``test_makecpt_plot_points``, but ``region`` does not need to be a fixture anymore because it is only used once. I recommend removing the fixture and just setting the region using a list in the ``region`` argument for plot.

pygmt/tests/test_makecpt.py
```diff
@@ -57,25 +56,25 @@ def test_makecpt_plot_points(points, region):
 
 
 @pytest.mark.mpl_image_compare
-def test_makecpt_plot_grid(grid):
+def test_makecpt_plot_grid(position):
     """
     Use static color palette table to change color of grid.
     """
```
seisman, line 62: The name and docstring don't make sense anymore, because we're not plotting grids here.

pygmt/tests/test_makecpt.py
```diff
@@ -57,25 +56,25 @@ def test_makecpt_plot_points(points, region):
...
     return fig
 
 
 @pytest.mark.mpl_image_compare
-def test_makecpt_plot_grid_scaled_with_series(grid):
+def test_makecpt_plot_grid_scaled_with_series(position):
     """
     Use static color palette table scaled to a min/max series to change color
     of grid.
     """
```
seisman, line 74: Same here.

pygmt/tests/test_makecpt.py
willschlitzer: updated in 8bd56a3

pygmt/tests/test_makecpt.py
willschlitzer: updated in 8bd56a3

seisman:
@willschlitzer Still need to upload the new images to dvc, because names of these images have been changed in your last commit.

meghanrjones (approved):
Thanks for updating these tests. I still would prefer if `fixture_region()` were removed and `test_makecpt_plot_points()` used a list as the argument type for ``region``, but it is only a suggestion.

weiji14 (commented):
Recommend modifying the PR title to say something like "Change test_makecpt.py to use colorbar instead of grdimage", since `static_earth_relief` isn't actually being used. Oh, and I second Meghan's suggestion to remove `fixture_region` if possible.

pygmt/tests/test_makecpt.py
```diff
@@ -105,70 +96,70 @@ def test_makecpt_invalid_output():
...
+    fig.colorbar(cmap=True, frame=True, position=position)
     return fig
 
 
 @pytest.mark.mpl_image_compare
-def test_makecpt_continuous(grid):
+def test_makecpt_continuous(position):
     """
     Use static color palette table that is continuous from blue to white and
     scaled from -4500 to 4500m.
```
seisman, line 136:
```suggestion
    scaled from 0 to 1000 m.
```

seisman (approved):
Looks good to me after this minor fix.

weiji14 (approved):
Nice work!

PR #1764, opened 2022-02-23
Remove Liam's talk from the overview page

seisman:

The overview page lists conference presentations about the development of PyGMT, but Liam's talk is a tutorial. I think we should remove it from the list.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -41,18 +41,6 @@ Presentations
 These are conference presentations about the development of PyGMT (previously
 "GMT/Python"):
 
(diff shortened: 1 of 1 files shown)


meghanrjones:
I actually find it non-intuitive to find the presentations in the overview section and would prefer moving all the presentations to the external resources page. Liam's talk could still go under tutorials rather than presentations.

seisman:
> I actually find it non-intuitive to find the presentations in the overview section and would prefer moving all the presentations to the external resources page. 

Then the overview page would be too short 🤦

meghanrjones (approved):

michaelgrund (approved):

PR #1773, opened 2022-03-01
Consistently use the load_sample_data function in tests

seisman:

Use the new `load_sample_data` rather than other deprecated load_ functions, so we don't see a lot of deprecation warnings in tests. Some docstrings are also updated.


meghanrjones (approved):
Related to the test reports, do we still need verbose="i" for `test_grdimage_grid_and_shading_with_xarray`?

PR #1790, opened 2022-03-07
Use pytest-doctestplus to skip some inline doctests

seisman:

We're adding more and more inline code examples in function docstrings as tracked in issue #1686. We don't want these new doctests slow down our CI, so we skip these doctests using the `# doctest: +SKIP` directive. Here is an example doctest in `grdlandmask`:
```
    >>> import pygmt  # doctest: +SKIP
    >>> # Create a landmask grid with an x-range of 125 to 130,
    >>> # and a y-range of 30 to 35
    >>> landmask = pygmt.grdlandmask(
    ...     spacing=1, region=[125, 130, 30, 35]
    ... )  # doctest: +SKIP
```
As you can see, there are many downsides for this kind of doctests:

1. `# doctest: +SKIP` only works for one single line, so we have to add `# doctest: +SKIP` many times
2. `  # doctest: +SKIP` has 18 characters. So even a short code like `landmask = pygmt.grdlandmask(spacing=1, region=[125, 130, 30, 35])` must be split into multiple lines
3. there is no simple way to run these tests

In this PR, the [`pytest-doctestplus`](https://github.com/astropy/pytest-doctestplus) plugin is  used to solve the problems above. For each module, we just need to define an internal variable `__doctest_skip__`, which contains a list of functions/classes whose tests should be skipped.
See the changes in `pygmt/src/grdlandmask.py` as an example and also the official documentation of [pytest-doctestplus](https://github.com/astropy/pytest-doctestplus#skip-unconditionally).

As you can see, it solves the problems 1 and 2 above. 

|Before | After |
|---|---|
|image |image|

For problem 3, there is still no easy way, but I think it's possible to comment the `__doctest_skip__` line, run the tests, and then revert the changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ omit = ["*/tests/*", "*pygmt/__init__.py"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-plus --mpl --mpl-results-path=results"
+addopts = "--verbose --durations=0 --durations-min=0.2 --mpl --mpl-results-path=results"
 
 [tool.isort]
 profile = "black"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ omit = ["*/tests/*", "*pygmt/__init__.py"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-modules --mpl --mpl-results-path=results"
+addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-plus --mpl --mpl-results-path=results"
 
 [tool.isort]
 profile = "black"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ omit = ["*/tests/*", "*pygmt/__init__.py"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--verbose --durations=0 --durations-min=0.2 --mpl --mpl-results-path=results"
+addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-modules --mpl --mpl-results-path=results"
 
 [tool.isort]
 profile = "black"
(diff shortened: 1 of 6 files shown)


seisman:
Ping @meghanrjones @weiji14 @willschlitzer @michaelgrund for comments on the new doctest method.

pyproject.toml
```diff
@@ -3,7 +3,7 @@ omit = ["*/tests/*", "*pygmt/__init__.py"]
 
 [tool.pytest.ini_options]
 minversion = "6.0"
-addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-modules --mpl --mpl-results-path=results"
+addopts = "--verbose --durations=0 --durations-min=0.2 --doctest-plus --mpl --mpl-results-path=results"
```
weiji14, line 6: > For problem 3, there is still no easy way, but I think it's possible to comment the __doctest_skip__ line, run the tests, and then revert the changes.

Actually, from reading the doctest-plus docs (https://github.com/astropy/pytest-doctestplus/tree/v0.12.0#setup-and-configuration), it says:

> Using pytest's built-in `--doctest-modules` option will override the behavior of this plugin, even if `doctest_plus = enabled in setup.cfg`, and will cause the default doctest plugin to be used...

So maybe we just need to edit the Makefile, and have have a `make doctest` command that uses `--doctest-modules` instead of `doctest-plus` to run all of these doctests? Could also setup a CI to have it run weekly if needed.

weiji14 (commented):
Great find! Should have found this earlier :laughing: I have one suggestion in regard to Problem 3 below.

pyproject.toml
seisman: Good suggestion. I removed the `--doctest-plus` option from `pyproject.toml`, and added it to `make test` command. I also the `make fulltest` target with the `--doctest-modules` option (I think `make doctest` is misleading).

Now, running `make test` collects 541 tests, and running `make fulltest` collects 542 tests. The extra one in `make fulltest` comes is the `grdlandmask` doctest.

weiji14 (approved):

seisman:
> pytest pygmt/src/<module>.py --doctest-modules

I added the `--doctest-modules` option back to pytest's default options, so that anyone can run `pytest pygmt/src/<module>.py` to run doctests without the `--doctest-modules` option.

So here are three different ways to run tests:

- `make test`: use the default pytest options (i.e., `--doctest-modules`), so that all tests and some doctests will run
- `make fulltest`: use `--doctest-modules --doctest-plus`, all tests and all doctests will run
- `pytest pygmt/src/<module>.py`: any doctests in the module will run

PR #1792, opened 2022-03-08
Raise an exception if the given parameter is not recognized and is longer than 2 characters

seisman:

Address #256.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -172,8 +172,16 @@ def build_arg_string(kwargs):
     ...     )
     ... )
     -BWSen -Bxaf -Byaf -I1/1p,blue -I2/0.25p,blue -JX4i -R1/2/3/4
+    >>> print(build_arg_string(dict(R="1/2/3/4", J="X4i", watre=True)))
+    Traceback (most recent call last):
+      ...
+    pygmt.exceptions.GMTInvalidInput: Unrecognized parameter 'watre'.
     """
     gmt_args = []
+    # raise an exception for unrecognized options
+    for key in kwargs:
+        if len(key) > 2:
+            raise GMTInvalidInput(f"Unrecognized parameter '{key}'.")
     # Exclude arguments that are None and False
     filtered_kwargs = {
         k: v for k, v in kwargs.items() if (v is not None and v is not False)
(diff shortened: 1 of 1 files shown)


pygmt/helpers/utils.py
```diff
@@ -172,8 +172,16 @@ def build_arg_string(kwargs):
...
+    >>> print(build_arg_string(dict(R="1/2/3/4", J="X4i", watre=True)))
+    Traceback (most recent call last):
+      ...
+    pygmt.exceptions.GMTInvalidInput: Unrecognized parameter 'watre'.
     """
     gmt_args = []
+    # raise an exception for unrecognized options
+    for key in kwargs:
+        if len(key) > 2:
+            raise GMTInvalidInput(f"Unrecognized parameter '{key}'.")
```
weiji14, line 184: Why not put this in the `for key in filtered_kwargs` loop below?

seisman: It's OK to put it in the `for key in filtered_kwargs` loop, but it won't catch typos like `watre=False`.

pygmt/helpers/utils.py
```diff
@@ -172,8 +172,16 @@ def build_arg_string(kwargs):
...
     # Exclude arguments that are None and False
```
seisman, line 185: Perhaps we can rewrite the whole function like below, so we don't need to loop over `kwargs` three times.
```python
    for key in filtered_kwargs:
        # raise an exception for unrecognized options
        if len(key) > 2:
            raise GMTInvalidInput(f"Unrecognized parameter '{key}'.")       
        if (kwargs[key] is None or kwargs[key] is False):
            pass  # Exclude arguments that are None and False
        elif is_nonstr_iter(kwargs[key]):
            for value in kwargs[key]:
                gmt_args.append(f"-{key}{value}")
        elif kwargs[key] is True:
            gmt_args.append(f"-{key}")
        else:
            gmt_args.append(f"-{key}{kwargs[key]}")
    return " ".join(sorted(gmt_args))
```

weiji14: Yep, looks good. It might conflict with #1487 but we can fix it later.

weiji14 (approved):
I had a last minute thought that the `len(key) > 2` check should probably go in the `@use_alias` decorator, but I really like this refactor so let's just go with this :stuck_out_tongue_closed_eyes:

willschlitzer (approved):

PR #1823, opened 2022-03-16
Use the build package to build sdist and wheel distributions

### LINKED ISSUE #1797: python setup.py gives deprecation warnings
In our CI workflow, we run commands like `python setup.py sdist --formats=zip` to build source or binary distributions. The command gives the following deprecation warnings:
```
SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
```
It seems `python setup.py` is no longer recommended to build distributions.

Reference: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary

seisman:

Fixes #1797.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -38,13 +38,13 @@ jobs:
       # Install GMT and other required dependencies from conda-forge
       - name: Install dependencies
         run: |
-          mamba install gmt=6.3.0 \
-                        numpy pandas xarray netCDF4 packaging matplotlib
+          mamba install gmt=6.3.0 numpy pandas xarray netCDF4 packaging \
+                        build matplotlib
 
       # Install the package that we want to test
       - name: Install the package
         run: |
-          python setup.py sdist --formats=zip
+          python -m build --sdist
           pip install dist/*
 
       # Download remote files
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -34,7 +34,7 @@ jobs:
         python-version: '3.10'
 
     - name: Install dependencies
-      run: python -m pip install setuptools wheel
+      run: python -m pip install build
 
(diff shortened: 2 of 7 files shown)


.github/workflows/cache_data.yaml
```diff
@@ -38,13 +38,13 @@ jobs:
       # Install GMT and other required dependencies from conda-forge
       - name: Install dependencies
         run: |
-          mamba install gmt=6.3.0 \
-                        numpy pandas xarray netCDF4 packaging matplotlib
+          mamba install gmt=6.3.0 numpy pandas xarray netCDF4 packaging \
+                        build matplotlib
```
weiji14, line 42: A bit unrelated, but why is matplotlib needed in the cache_data workflow? Isn't this GitHub Actions workflow just used to download files?

I see that the `matplotlib` dependency was added at 270e3ea29a72058c277922cc16c2fbf225f63ea1 in #1310, but can't see exactly why this is needed.

.github/workflows/publish-to-pypi.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         python-version: '3.10'
 
     - name: Install dependencies
-      run: python -m pip install setuptools wheel
+      run: python -m pip install build
```
weiji14, line 37: Oh, and just noting that we should double check that the wheel and source files on TestPyPI look ok once this PR is merged.

.github/workflows/cache_data.yaml
weiji14: Ah ok. Seems like that `import` line could just be put inside the `check_figures_equal` function, but let's just keep things the way it is right now.

weiji14 (approved):
Ok for me on Linux. Might pay to know if this works locally on macOS and Windows as well, but the CI does look ok.

PR #1833, opened 2022-03-19
Run full tests only on Wednesday scheduled jobs

seisman:

Address comment https://github.com/GenericMappingTools/pygmt/issues/1686#issuecomment-1061357814.

**Reminders**

**Slash Commands**


seisman:
It should work, but I can't test it.

weiji14:
I was thinking of running these full tests on a less regular basis, maybe weekly or monthly instead?

seisman:
> I was thinking of running these full tests on a less regular basis, maybe weekly or monthly instead?

I agree. Changing the following lines to weekly?

https://github.com/GenericMappingTools/pygmt/blob/813142bbc2ef84b49cff91f20dcb99999ce067d6/.github/workflows/ci_tests.yaml#L21

weiji14:
> > I was thinking of running these full tests on a less regular basis, maybe weekly or monthly instead?
> 
> I agree. Changing the following lines to weekly?
> 
> https://github.com/GenericMappingTools/pygmt/blob/813142bbc2ef84b49cff91f20dcb99999ce067d6/.github/workflows/ci_tests.yaml#L21

But this means the daily unit tests will become weekly. 

Or is there a way to just run the doctests, and not the unit tests? That way we could put it in the docs workflow as a weekly/monthly scheduled run (in addition to the current docs build).

seisman:
> > > I was thinking of running these full tests on a less regular basis, maybe weekly or monthly instead?
> > 
> > 
> > I agree. Changing the following lines to weekly?
> > https://github.com/GenericMappingTools/pygmt/blob/813142bbc2ef84b49cff91f20dcb99999ce067d6/.github/workflows/ci_tests.yaml#L21
> 
> But this means the daily unit tests will become weekly.

We have PRs merging into the main branch every few days, so the unit tests are also run every few days, making the daily unit tests less necessary. So a weekly job is still acceptable to me.

> Or is there a way to just run the doctests, and not the unit tests? That way we could put it in the docs workflow as a weekly/monthly scheduled run (in addition to the current docs build).

Sphinx has its own doctest extension, which can run doctests.
```
cd doc/
sphinx-build -b doctest . _build
```
The above command works, but it reports many "errors" that are not reported by pytest.

.github/workflows/ci_tests.yaml
```diff
@@ -134,9 +134,15 @@ jobs:
...
 
       # Run the tests
-      - name: Test with pytest
+      - name: Run tests
+        if: github.event_name != 'schedule'
         run: make test PYTEST_EXTRA="-r P"
 
+      # Run full tests including doctests
+      - name: Run full tests
+        if: github.event_name == 'schedule'
```
weiji14, line 143: Not sure if GitHub Actions uses `&` or `and`, but try something like this to run full tests on Wednesday only. Xref https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
```suggestion
        if: github.event_name == 'schedule' & github.event.schedule == '0 0 * * 3'
```

seisman: > Not sure if GitHub Actions uses `&` or `and`,

It's `&&` (https://docs.github.com/en/actions/learn-github-actions/expressions#operators).

.github/workflows/ci_tests.yaml
```diff
@@ -134,9 +134,15 @@ jobs:
           pip install dist/*
 
       # Run the tests
-      - name: Test with pytest
+      - name: Run tests
+        if: github.event_name != 'schedule'
```
seisman, line 138: This `if` statement means the tests won't be run in daily jobs, which is not what we want. However, removing this `if` statement means on Wednesday job, most tests will be run twice by both `make test` and `make fulltest`.

weiji14: What about something like
```suggestion
        if: github.event.schedule != '0 0 * * 3'
```
Though not sure if it will error for regular (non-scheduled) tests, e.g. in PRs.

.github/workflows/ci_tests.yaml
```diff
@@ -134,9 +134,15 @@ jobs:
...
         run: make test PYTEST_EXTRA="-r P"
 
+      # Run full tests including doctests
```
weiji14, line 141:
```suggestion
      # Run full tests including doctests on Wednesday
```

weiji14 (commented):
Also need to update `doc/maintenance.md` to mention the new full tests schedule.

.github/workflows/ci_tests.yaml
seisman: No errors in PRs.

.github/workflows/ci_tests.yaml
```diff
@@ -134,9 +134,15 @@ jobs:
           pip install dist/*
 
       # Run the tests
```
weiji14, line 136:
```suggestion
      # Run regular tests
```

weiji14 (approved):
Nice! Just one last suggestion, and wait for second approval to merge.

Issue #1832, opened 2022-03-18
which(), image() and legend() don't support pathlib.Path inputs


seisman:
I'm trying to fix a `pathlib` bug of plot and plot3d in #1831 and find that `which()`, `image()` and `legend` also don't support `pathlib.Path` inputs.

To reproduce the issue, run:
```
>>> import pygmt
>>> from pathlib import Path

>>> pygmt.which(Path("abc.txt"))

>>> fig = pygmt.Figure()
>>> fig.image(Path("abc.png"))

>>> fig.legend(spec=Path("legend.txt"))
```
The error message is: `TypeError: sequence item 0: expected str instance, PosixPath found`.

Other functions work because their input file is wrapped in a context manager, like

https://github.com/GenericMappingTools/pygmt/blob/8fe2d331f04d21fc1ad731dea983c1701c42a662/pygmt/src/histogram.py#L150-L152

To fix the bug, the simplest solution is changing input file to a string type, for example, changing `fname` to `str(fname)`. Do anyone have better and more general solutions?

seisman:
In `pygmt.which()` or other functions, we build the argument string using the following code:
https://github.com/GenericMappingTools/pygmt/blob/813142bbc2ef84b49cff91f20dcb99999ce067d6/pygmt/src/which.py#L63

Perhaps we should refactor the `build_arg_string` function to:
```
build_arg_string(kwargs, infile=None, outfile=None)
```
and in `build_arg_string`, we always make sure that `infile` and `outfile` are converted to string type.

seisman:
Does anyone have comments on the proposed refactor in #1837?

meghanrjones:
> Does anyone have comments on the proposed refactor in https://github.com/GenericMappingTools/pygmt/pull/1837?

I like the idea in general. I would rather that all modules handle input and output using the same strategy rather than only applying this code structure to these three methods/functions. Would you be open to a larger refactor using the new ``build_arg_string`` input and output arguments in other functions/methods for consistency?

A couple other questions:
1. Should we do any type checking before converting to a string?
[...]

PR #1837, opened 2022-03-21
Refractor build_arg_string to also deal with infile and outfile

seisman:

Address the comment in https://github.com/GenericMappingTools/pygmt/issues/1832#issuecomment-1072974748.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -119,7 +119,7 @@ def dummy_context(arg):
     yield arg
 
 
-def build_arg_string(kwargs):
+def build_arg_string(kwargs, infile=None, outfile=None):
     r"""
     Transform keyword arguments into a GMT argument string.
 
@@ -213,6 +213,10 @@ def build_arg_string(kwargs):
                 # remove any spaces in PROJ4 string
                 _value = str(kwargs[key]).replace(" ", "")
             gmt_args.append(rf"-{key}{_value}")
+    if infile:
+        gmt_args.append(str(infile))
+    if outfile:
+        gmt_args.append("->" + str(outfile))
     return " ".join(sorted(gmt_args))
 
 
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -119,12 +119,12 @@ def dummy_context(arg):
     yield arg
 
 
-def build_arg_string(kwargs, infile=None, outfile=None):
+def build_arg_string(kwdict, infile=None, outfile=None):
     r"""
-    Transform keyword arguments into a GMT argument string.
+    Convert a dict and optional input/output files into a GMT argument string.
 
-    Make sure all arguments have been previously converted to a string
-    representation using the ``kwargs_to_strings`` decorator. The only
+    Make sure all values in ``kwdict`` have been previously converted to a
+    string representation using the ``kwargs_to_strings`` decorator. The only
     exceptions are True, False and None.
 
     Any lists or tuples left will be interpreted as multiple entries for the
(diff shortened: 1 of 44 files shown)


pygmt/helpers/utils.py
```diff
@@ -213,6 +217,10 @@ def build_arg_string(kwargs):
                 # remove any spaces in PROJ4 string
                 _value = str(kwargs[key]).replace(" ", "")
             gmt_args.append(rf"-{key}{_value}")
+    if infile:
+        gmt_args.append(str(infile))
+    if outfile:
+        gmt_args.append("->" + str(outfile))
     return " ".join(sorted(gmt_args))
```
seisman, line 224: I don't like how the arguments are sorted. For example, currently, the returned argument string may look like:
```
->outfile -A0 -Bafg -Ctopo -Vd infile
```
but I expect to see a more readable argument string like:
```
infile -A0 -Bafg -Ctopo -Vd ->outfile
```
which is more friendly when debugging.

I can rewrite these lines to 
```
gmt_args = sorted(gmt_args)
if infile:
    gmt_args = [str(infile)] + gmt_args
if outfile:
    gmt_args = gmt_args.append("->" + str(outfile))
return " ".join(gmt_args)
```
but I'm wondering if someone can give a better and cleaner solution.

pygmt/helpers/utils.py
```diff
@@ -119,7 +119,7 @@ def dummy_context(arg):
     yield arg
 
 
-def build_arg_string(kwargs):
+def build_arg_string(kwargs, infile=None, outfile=None):
```
weiji14, line 122: In Python, kwargs is usually placed at the last position, but here it is placed as first. Maybe we should change the ordering of the parameters here? See also @meghanrjones's comment at https://github.com/GenericMappingTools/pygmt/pull/731#discussion_r825477816 and https://peps.python.org/pep-3102 about keyword-only arguments.

seisman: Actually, it's just a dict which has a name `kwargs`. Maybe `kwargs` is not a good name here.

weiji14: How about `kwdict`?

pygmt/helpers/utils.py
weiji14: Something like this? Based on https://stackoverflow.com/questions/3752240/join-string-and-none-string-using-optional-delimiter/3752317#3752317
```suggestion
    return " ".join(filter(None, [infile, sorted(gmt_args), outfile]))
```

pygmt/src/blockm.py
```diff
@@ -51,8 +51,10 @@ def _blockm(block_method, data, x, y, z, outfile, **kwargs):
             with table_context as infile:
                 if outfile is None:
                     outfile = tmpfile.name
-                arg_str = " ".join([infile, build_arg_string(kwargs), "->" + outfile])
-                lib.call_module(module=block_method, args=arg_str)
+                lib.call_module(
+                    module=block_method,
+                    args=build_arg_string(kwargs, infile=infile, outfile=outfile),
```
weiji14, line 56: Still having mixed feeling about having kwargs as the first argument here, but I guess it works :upside_down_face:

weiji14 (approved):

seisman:
This PR will conflict with PR #1857 or #1862, due to changes in grd2cpt and makecpt. We need to decide which PR should be merged first.

weiji14:
> This PR will conflict with PR #1857 or #1862, due to changes in grd2cpt and makecpt. We need to decide which PR should be merged first.

Merge this one (#1837) first, so that you can proceed with #1831/#1832. The conflicts in #1857 and #1862 can be resolved later.

PR #1866, opened 2022-04-04
Let slash command /test-gmt-dev report job URL

### LINKED ISSUE #1816: Format and test-gmt-dev slash commands not working properly

The `/format` and `/test-gmt-dev` slash commands added in #646 and #831 respectively seem to have stopped working like they used to, and probably for a while now to be honest. I think GitHub made some changes in regard to bot permissions some time ago (though I can't pin an exact date), and a lot of these tools with auto-commit functionality just broke.

To be specific:
1. For the `/format` command, an action is triggered and runs to completion, but no formatting commit actually happens. E.g.:
  - https://github.com/GenericMappingTools/pygmt/pull/1719#issuecomment-1065125258
  - https://github.com/GenericMappingTools/pygmt/pull/1812#issuecomment-1066258411
2. The `/test-gmt-dev` command, I think it does run (?), but you have to jump to the 'Actions' tab to see it run which is very inconvenient. From memory, I think there was another problem with how it wasn't testing the changes in the PR branch properly, but some rebased commit instead? E.g.:
  - https://github.com/GenericMappingTools/pygmt/pull/1577#issuecomment-944877093
  - https://github.com/GenericMappingTools/pygmt/pull/517#issuecomment-777194521

So, we'll either need to update these convenience slash commands, or remove them.

seisman:

As reported in #1816:

> The /test-gmt-dev command, I think it does run (?), but you have to jump to the 'Actions' tab to see it run which is very inconvenient.

This PR fixes the problem by adding the job URL to the PR comment.

See the PR https://github.com/seisman/pygmt/pull/2 in my own fork for how it works:

image

Fixes #1816.

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -71,6 +71,14 @@ jobs:
           # fecth all history so that setuptools-scm works
           fetch-depth: 0
 
+      - name: Show job URL
+        uses: peter-evans/create-or-update-comment@v2
+        if: github.event_name == 'repository_dispatch' && (matrix.os == 'ubuntu-latest')
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
+          comment-id: ${{ github.event.client_payload.github.payload.comment.id }}
+
       # Install Mambaforge with conda-forge dependencies
       - name: Setup Mambaforge
         uses: conda-incubator/setup-miniconda@v2.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests_dev.yaml
```diff
@@ -71,6 +71,14 @@ jobs:
           # fecth all history so that setuptools-scm works
           fetch-depth: 0
 
+      - name: Show job URL
+        uses: peter-evans/create-or-update-comment@v2
```
weiji14, line 75:
```suggestion
        uses: peter-evans/create-or-update-comment@v2.0.0
```

weiji14 (approved):
Cool, thanks @seisman! Just one minor suggestion, otherwise looks good to me.

PR #1865, opened 2022-04-03
Pull dvc baseline images from gmt-6.4 branch to test GMT dev

weiji14:
**Description of proposed changes**

Make it easier to catch any changes in upstream GMT that break PyGMT, until PyGMT bumps its minimum required version to GMT 6.4.

This modifies the GMT Dev Tests workflow to `dvc pull` images from @seisman's branch at https://github.com/GenericMappingTools/pygmt/pull/1791 that has the newer baseline images generated using GMT 6.4, so that the daily scheduled GMT Dev tests can pass.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1791, opened 2022-03-07
WIP: Update baseline images for the GMT 6.4 part 1

seisman:
**Description of proposed changes**

Recent upstream GMT changes break some PyGMT tests. Some failures are due to very tiny changes so we just need to update the baseline images when we bump to GMT 6.4.0. However, some failures are caused by new bugs in recent changes (for example https://github.com/GenericMappingTools/gmt/issues/6213#issuecomment-1057191889). It becomes difficult to find these GMT bugs because we have both good and bad failures.

This PR tries to update baseline images for good failures with only tiny changes. Any remaining failures likely reflect new upstream bugs or wrong baseline images.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1882, opened 2022-04-15
Pull baseline images from the gmt-6.4 branch only for scheduled jobs

seisman:

Currently, the workflow always pull baseline images from the gmt-6.4 branch. It makes it difficult to verify if changes in PRs like https://github.com/GenericMappingTools/pygmt/pull/1791 are correct or not.

Patches #1865

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -110,6 +110,7 @@ jobs:
         run: |
           ORIGINAL_BRANCH=$(git branch --show-current)
           # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
+          # https://github.com/GenericMappingTools/pygmt/pull/1883
           git checkout gmt-6.4
           dvc pull
           ls -lhR pygmt/tests/baseline/
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -106,6 +106,7 @@ jobs:
 
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
+        if: github.event_name == 'schedule'
         run: |
           ORIGINAL_BRANCH=$(git branch --show-current)
           # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests_dev.yaml
```diff
@@ -106,6 +106,7 @@ jobs:
 
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
+        if: github.event_name == 'schedule'
         run: |
           ORIGINAL_BRANCH=$(git branch --show-current)
           # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
```
weiji14, line 112: Should the `gmt-6.4` branch have a Pull Request associated with it? Or do we just keep pushing to that `gmt-6.4` branch directly?

seisman: > Should the `gmt-6.4` branch have a Pull Request associated with it?

Opening a PR for gmt-6.4 sounds a good idea.

weiji14: Ok, then we can reference the `gmt-6.4` with a GitHub URL, and add it to the file here:

```suggestion
          # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
          # https://github.com/GenericMappingTools/pygmt/pull/<number>
```

seisman: I can't open a PR for the gmt-6.4 branch because currently it's the same as the main branch. So we can't know the exact PR number now. Maybe we can assume that the next PR (#1883) comes from the gmt-6.4 branch.

weiji14: Could have `gmt-6.4.` target a different branch first (e.g. `update-baseline-images-for-gmt-dev`) and then switch it to target the `main` branch later?

seisman: Great idea. The PR is now at https://github.com/GenericMappingTools/pygmt/pull/1883.

weiji14: Not sure why the above suggestion doesn't work, let's try again.
```suggestion
          # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
          # https://github.com/GenericMappingTools/pygmt/pull/1883
```

seisman, line 112:
```suggestion
          # Pull down GMT 6.4 baseline images from the gmt-6.4 branch
          # https://github.com/GenericMappingTools/pygmt/pull/1883
```

weiji14 (approved):

PR #1894, opened 2022-04-22
Added alias for F in basemap

### LINKED ISSUE #1893: Missing alias for option -F in basemap

I am trying to add a filled background behind my mapscale, similar to the examples in the GMT documentation here: https://docs.generic-mapping-tools.org/dev/cookbook/features.html#placing-map-scales. From what I understand, when adding the map_scale, I should be able to simultaneously add a frame option specifying the fill parameters, like in my example code below, using the options from the GMT documentation. The code below runs when 'frame' isn't specified, but of course no background is generated behind the map_scale. I would be very appreciative if someone could point me in the right direction or let me know if this is in fact a bug. Thanks!

```python
import pygmt

fig = pygmt.Figure()
fig.coast(
    # Set the x-range from 10E to 20E and the y-range to 35N to 45N
    region=[10, 20, 35, 45],
    projection="M12c",
    land="lightgray",
    water="white",
    borders="1/0.5p",
    shorelines="1/0.5p",
    frame="ag",
)
fig.basemap(map_scale='jBR+c36N+w100k+l+f', frame='+p1p+i+gwhite+c0.1i')
fig.show()
```

**Full error message**

```
pygmt.exceptions.GMTCLibError: Module 'basemap' failed with status code 72:
basemap [ERROR]: Option -B: Found frame setting modifiers (+b|g|i|n|o|t|w|x|y|z) mixed with axes settings!
basemap [ERROR]: Option -B parsing failure. Correct syntax:
basemap [ERROR]: Offending option -B+p1p+i+gwhite+c0.1i
```

**System information**

```
PyGMT information:
  version: v0.6.1
System information:
  python: 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:24:38)  [Clang 12.0.1 ]
  executable: /opt/homebrew/Caskroom/miniforge/base/bin/python
  machine: macOS-12.3.1-arm64-arm-64bit
Dependency information:
  numpy: 1.22.3
  pandas: 1.4.2
  xarray: 2022.3.0
  netCDF4: 1.5.7
  packaging: 21.3
  geopandas: 0.10.2
  ghostscript: 9.56.1
  gmt: 6.3.0
GMT library information:
  binary dir: /opt/homebrew/Caskroom/miniforge/base/bin
  cores: 10
  grid layout: rows
  library path: /opt/homebrew/Cellar/gmt/6.3.0_2/lib/libgmt.dylib
  padding: 2
  plugin dir: /opt/homebrew/Cellar/gmt/6.3.0_2/lib/gmt/plugins
  share dir: /opt/homebrew/Cellar/gmt/6.3.0_2/share/gmt
  version: 6.3.0
```

jackbeagley:

Adds an alias for "box" so that backgrounds and frames for map_scales can be generated more easily.

Fixes #1893 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/basemap.py
+++ b/pygmt/src/basemap.py
@@ -65,7 +65,7 @@ def basemap(self, **kwargs):
     box : bool or str
         [**+c**\ *clearances*][**+g**\ *fill*][**+i**\ [[*gap*/]\ *pen*]]\
         [**+p**\ [*pen*]][**+r**\ [*radius*]][**+s**\ [[*dx*/*dy*/][*shade*]]].
-        If set to ``True``, draws a rectangular border around the map inset,
+        If set to ``True``, draws a rectangular border around the
         map scale or rose. Alternatively, specify a different pen with
         **+p**\ *pen*. Add **+g**\ *fill* to fill the scale panel [default is
         no fill]. Append **+c**\ *clearance* where *clearance* is either gap,
--- a/pygmt/src/basemap.py
+++ b/pygmt/src/basemap.py
@@ -65,20 +65,21 @@ def basemap(self, **kwargs):
     box : bool or str
         [**+c**\ *clearances*][**+g**\ *fill*][**+i**\ [[*gap*/]\ *pen*]]\
         [**+p**\ [*pen*]][**+r**\ [*radius*]][**+s**\ [[*dx*/*dy*/][*shade*]]].
-        If set to ``True``, draws a rectangular border around the map scale.
-        Alternatively, specify a different pen with **+p**\ *pen*. Add
-        **+g**\ *fill* to fill the scale panel [default is no fill]. Append
-        **+c**\ *clearance* where *clearance* is either gap, xgap/ygap, or
-        lgap/rgap/bgap/tgap where these items are uniform, separate in x- and
-        y-direction, or individual side spacings between scale and border.
-        Append **+i** to draw a secondary, inner border as well. We use a
-        uniform gap between borders of 2p and the :gmt-term:`MAP_DEFAULTS_PEN`
-        unless other values are specified. Append **+r** to draw rounded
-        rectangular borders instead, with a 6p corner radius. You can override
-        this radius by appending another value. Finally, append **+s** to draw
-        an offset background shaded region. Here, *dx/dy* indicates the shift
-        relative to the foreground frame [4p/-4p] and shade sets the fill
-        style to use for shading [default is gray50].
+        If set to ``True``, draws a rectangular border around the map inset,
+        map scale or rose. Alternatively, specify a different pen with
+        **+p**\ *pen*. Add **+g**\ *fill* to fill the scale panel [default is
+        no fill]. Append **+c**\ *clearance* where *clearance* is either gap,
+        xgap/ygap, or lgap/rgap/bgap/tgap where these items are uniform,
+        separate in x- and y-direction, or individual side spacings between
+        scale and border. Append **+i** to draw a secondary, inner border as
+        well. We use a uniform gap between borders of 2p and the
+        :gmt-term:`MAP_DEFAULTS_PEN` unless other values are specified. Append
+        **+r** to draw rounded rectangular borders instead, with a 6p corner
+        radius. You can override this radius by appending another value.
+        Finally, append **+s** to draw an offset background shaded region.
+        Here, *dx/dy* indicates the shift relative to the foreground frame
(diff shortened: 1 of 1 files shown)


michaelgrund:
Thanks for opening this PR @jackbeagley. Besides adding an alias for ``-F`` could also provide the corresponding documentation part in the docstring? You can orientate yourself on the already existing ones for basemap.

seisman:
> Besides adding an alias for `-F` could you please also provide the corresponding documentation part in the docstring? You can orientate yourself on the already existing ones for basemap.

@jackbeagley FYI, the same `box` parameter has been aliased in the `colorbar` function, so you can simply copy the `box` docstring from `colorbar` to `basemap`:

https://github.com/GenericMappingTools/pygmt/blob/555cd310b0afbb79330df02a5679ccf5358f5343/pygmt/src/colorbar.py#L71

jackbeagley:
@seisman , awesome, I have just copied the description from `colorbar` verbatim, as it seemed very well written and as you say - it is the same scale, so I don't think there is anything additional that needs to be added?

jackbeagley:
Does a test need to be written for this too?

pygmt/src/basemap.py
```diff
@@ -61,6 +62,23 @@ def basemap(self, **kwargs):
         [**g**\|\ **j**\|\ **J**\|\ **n**\|\ **x**]\ *refpoint*\
         **+w**\ *length*.
         Draws a simple map scale centered on the reference point specified.
+    box : bool or str
+        [**+c**\ *clearances*][**+g**\ *fill*][**+i**\ [[*gap*/]\ *pen*]]\
+        [**+p**\ [*pen*]][**+r**\ [*radius*]][**+s**\ [[*dx*/*dy*/][*shade*]]].
+        If set to ``True``, draws a rectangular border around the color scale.
```
seisman, line 68:
```suggestion
        If set to ``True``, draws a rectangular border around the map scale.
```

seisman:
> Does a test need to be written for this too?

I don't think a test is needed for such an alias.

michaelgrund (commented):
I think we have to consider the different types of map objects that can be manipulated such as  map inset, map scale, or map rose.

jackbeagley:
> I think we have to consider the different types of map objects that can be manipulated such as map inset, map scale, or map rose.

Hey @michaelgrund, is that what you had in mind?

pygmt/src/basemap.py
```diff
@@ -61,6 +62,24 @@ def basemap(self, **kwargs):
...
+        xgap/ygap, or lgap/rgap/bgap/tgap where these items are uniform,
+        separate in x- and y-direction, or individual side spacings between
+        scale and border. Append **+i** to draw a secondary, inner border as
+        well. We use a uniform gap between borders of 2p and the
+        :gmt-term:`MAP_DEFAULTS_PEN` unless other values are specified. Append
+        **+r** to draw rounded rectangular borders instead, with a 6p corner
+        radius. You can override this radius by appending another value.
+        Finally, append **+s** to draw an offset background shaded region.
+        Here, *dx/dy* indicates the shift relative to the foreground frame
+        [4p/-4p] and shade sets the fill style to use for shading
```
weiji14, line 81: One quick suggestion here
```suggestion
        [Default is 4p/-4p] and shade sets the fill style to use for shading
```

pygmt/src/basemap.py
```diff
@@ -61,6 +62,24 @@ def basemap(self, **kwargs):
         [**g**\|\ **j**\|\ **J**\|\ **n**\|\ **x**]\ *refpoint*\
         **+w**\ *length*.
         Draws a simple map scale centered on the reference point specified.
+    box : bool or str
+        [**+c**\ *clearances*][**+g**\ *fill*][**+i**\ [[*gap*/]\ *pen*]]\
+        [**+p**\ [*pen*]][**+r**\ [*radius*]][**+s**\ [[*dx*/*dy*/][*shade*]]].
+        If set to ``True``, draws a rectangular border around the map inset,
+        map scale or rose. Alternatively, specify a different pen with
```
seisman, line 69: `map inset` is no longer available in the modern-mode "basemap" module. Users who wants to use `inset` should use the `inset` module instead.
```suggestion
        If set to ``True``, draws a rectangular border around the 
        map scale or rose. Alternatively, specify a different pen with
```

seisman (approved):

michaelgrund (approved):

PR #1910, opened 2022-04-29
Consistently pass keyword arguments when using call_module()

seisman:

When calling the `call_module()` method, we use two different styles:

1. Pass positional arguments
```
lib.call_module("triangulate", build_arg_string(kwargs, infile=infile))
```
2. Pass keyword arguments
```
lib.call_module(module="triangulate", args=build_arg_string(kwargs, infile=infile))
```

The two styles work the same way, but it's better to use a single style in the project.

I prefer the first style (passing positional arguments) because it's shorter.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -19,7 +19,7 @@ jobs:
     steps:
       # Checkout current git repository
       - name: Checkout
-        uses: actions/checkout@v3.0.1
+        uses: actions/checkout@v3.0.2
         with:
           # fecth all history so that setuptools-scm works
           fetch-depth: 0
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -15,12 +15,12 @@ jobs:
 
     steps:
     - name: Checkout the repository
-      uses: actions/checkout@v3.0.1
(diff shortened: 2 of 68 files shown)


weiji14:
I kinda prefer the longer kwargs version though, easier to see what the parameters are, especially for newcomers. Maybe get a vote on this?

seisman:
> I kinda prefer the longer kwargs version though, easier to see what the parameters are, especially for newcomers. Maybe get a vote on this?

Here are the main reasons that I prefer the short version:

- `call_module`'s arguments are positional arguments, so `module=` and `args=` are not required
- The function only has two arguments, and it's easy to guess the meaning of the two arguments
- it's shorter. In some functions, it can be written in a single line rather than 4 lines.

willschlitzer:
I'm also a fan of the keyword arguments. I understand that it's pretty clear which argument is being used as there are only two, but I think it's always best to remove any chance for ambiguity.

seisman:
OK, I've updated all `call_module()` calls using the keyword arguments. The exceptions are a few doctests in `pygmt/clib.session.py` and tests in `pygmt/tests/test_session_management.py` and `pygmt/tests/test_clib.py`.

weiji14 (approved):

PR #1913, opened 2022-05-03
Fix the grid accessor (grid registration and type) for 3D grids

seisman:

As mentioned in https://github.com/GenericMappingTools/gmt/issues/6588#issuecomment-1115541413, the `grdinfo` output for a 3D grid was wrong in old GMT versions.

After the upstream fix, the grdinfo output is:
```
$ gmt grdinfo -Cn eraint_uvz.nc     
grdinfo [WARNING]: No 3-D array in file eraint_uvz.nc.  Selecting first 3-D slice in the 4-D array z
-180	179.25	-90	90	200	850	66825.5	66825.5	0.75	0.75	0	480	241	3	0	1
```
which has 16 columns, instead of 12 columns for 2D grids. Thus, we cannot get the grid registration and type information from columes 11 and 12. Instead, we should use the last two columns instead.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/accessors.py
+++ b/pygmt/accessors.py
@@ -29,10 +29,10 @@ class GMTDataArrayAccessor:
         self._obj = xarray_obj
         try:
             self._source = self._obj.encoding["source"]  # filepath to NetCDF source
-            # From the shortened summary information of `grdinfo`,
-            # get grid registration in column 10, and grid type in column 11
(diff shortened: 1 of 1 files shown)


weiji14 (approved):
Looks great, and it's compatible with GMT 6.3 and 6.4!

PR #1921, opened 2022-05-10
Add earth_relief_holes to load_sample_data()

willschlitzer:
This adds a function to load the remote dataset `@earth_relief_20m_holes.grd` for use with the in-line example for `grdfill`.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -3,7 +3,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  # pull_request:
+  pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -3,7 +3,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  pull_request:
+  # pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -348,7 +348,7 @@ def load_mars_shape(**kwargs):
     return data
 
 
-def _load_earth_relief_holes(**kwargs):
+def _load_earth_relief_holes(**kwargs):  # pylint: disable=unused-argument
     """
     Loads the remote file @earth_relief_20m_holes.grd.
 
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -5,6 +5,7 @@ import warnings
 
 import pandas as pd
 from pygmt.exceptions import GMTInvalidInput
+from pygmt.io import load_dataarray
 from pygmt.src import which
 
 
@@ -23,6 +24,7 @@ def list_sample_data():
     """
     names = {
         "bathymetry": "Table of ship bathymetric observations off Baja California",
+        "earth_relief_holes": "Earth relief grid with holes",
         "fractures": "Table of hypothetical fracture lengths and azimuths",
         "hotspots": "Table of locations, names, and symbol sizes of hotpots from "
         " Mueller et al., 1993",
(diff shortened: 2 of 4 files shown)


weiji14 (commented):
Thanks @willschlitzer! Do you also want to add `@earth_relief_20m_holes` to the cache list at https://github.com/GenericMappingTools/pygmt/blob/951d30aa9578b4108b768983c09208c021ec1d1f/pygmt/helpers/testing.py#L153-L155 ?

pygmt/datasets/samples.py
```diff
@@ -23,6 +24,7 @@ def list_sample_data():
     """
     names = {
         "bathymetry": "Table of ship bathymetric observations off Baja California",
+        "earth_relief_holes": "Earth relief grid with holes",
```
seisman, line 27:
```suggestion
        "earth_relief_holes": "Regional 20 arc-minute Earth relief grid with holes",
```

pygmt/datasets/samples.py
```diff
@@ -343,3 +346,18 @@ def load_mars_shape(**kwargs):
...
+
+    Returns
+    -------
+    grid : :class:`xarray.DataArray`
+        The Earth relief grid. Coordinates are latitude and longitude in
+        degrees. Relief is in meters.
+    """
+    fname = which("@earth_relief_20m_holes.grd", download="c")
+    grid = load_dataarray(fname, engine="netcdf4")
+    return grid
```
seisman, line 363:
```suggestion
    return load_dataarray(fname, engine="netcdf4")
```

.github/workflows/cache_data.yaml
```diff
@@ -3,7 +3,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  # pull_request:
+  pull_request:
```
seisman, line 6: Need to comment this line before merging:
```suggestion
  # pull_request:
```

seisman (approved):

pygmt/datasets/samples.py
```diff
@@ -343,3 +346,18 @@ def load_mars_shape(**kwargs):
         fname, sep="\t", header=None, names=["longitude", "latitude", "radius(m)"]
     )
     return data
+
+
+def _load_earth_relief_holes(**kwargs):
```
seisman, line 351: Pylint reports the following warning. 
```
pygmt/datasets/samples.py:351:0: W0613: Unused argument 'kwargs' (unused-argument)
```
I think we don't need `kwagrs` for new datasets.
```suggestion
def _load_earth_relief_holes():
```

seisman:
> @GenericMappingTools/pygmt-admin It looks like this is failing some tests because of the change to gridline registration for `@earth_relief_20m_holes.grd`. Is this file now stable?

xref: the grid file registration was changed in https://github.com/GenericMappingTools/gmtserver-admin/pull/156

pygmt/datasets/samples.py
willschlitzer: I'll check again to confirm after I write the tests, but I think it needs `**kwargs` because of `data = load_func[name](suppress_warning=True)`. It was getting passed a value for `suppress_warning` but wasn't expecting any arguments.

meghanrjones:
> > @GenericMappingTools/pygmt-admin It looks like this is failing some tests because of the change to gridline registration for `@earth_relief_20m_holes.grd`. Is this file now stable?
> 

Yes, it's now stable so the baseline images can be updated.

pygmt/datasets/samples.py
willschlitzer: I received `TypeError: _load_earth_relief_holes() got an unexpected keyword argument 'suppress_warning'` when I tried running it without `**kwargs`.

seisman: The `suppress_warning` parameter (or `**kwargs`) was added so that users who call the deprecated, old `load_xxx` functions can silent the deprecation warnings by passing `suppress_warning=True`. Thus, `**kwargs` is required for new private `_load_xxx` functions.

Do you think we should rewrite the code to call the old and new load_xxx functions in different ways:
```
    # old public load functions for backward compatibility
    load_func_old = {
        "bathymetry": load_sample_bathymetry,
        "fractures": load_fractures_compilation,
        "hotspots": load_hotspots,
        "japan_quakes": load_japan_quakes,
        "mars_shape": load_mars_shape,
        "ocean_ridge_points": load_ocean_ridge_points,
        "usgs_quakes": load_usgs_quakes,
    }
    # new private load functions
    load_func = {
        "earth_relief_holes": _load_earth_relief_holes,
    }
    if name in load_func_old:
        data = load_func[name](suppress_warning=True)
    else:
        data = load_func[name]
    return data
```

willschlitzer: I think that's a good idea. Should that be a separate PR?

seisman: A separate PR is OK to me.

Issue #1617, opened 2021-11-08
In some functions, G is aliased as "fill" but shown "color" in docstrings


mdtanker:
Noticed a mistake here: -G (fill) is listed as "color" under the "Parameters" section, should be "fill".

weiji14:
Right, so the code aliases `-G` as `fill`:

https://github.com/GenericMappingTools/pygmt/blob/2ebe165373106d7bff5edcea3e5e09e077bc80b6/pygmt/src/histogram.py#L23

But we used `{G}` in the docstring here:

https://github.com/GenericMappingTools/pygmt/blob/2ebe165373106d7bff5edcea3e5e09e077bc80b6/pygmt/src/histogram.py#L69

which means it shows up in the docs as `color`:

https://github.com/GenericMappingTools/pygmt/blob/2ebe165373106d7bff5edcea3e5e09e077bc80b6/pygmt/helpers/decorators.py#L46-L47

![image](https://user-images.githubusercontent.com/23487320/140869026-f98f1129-58e1-46bb-958c-6b96957228d2.png)

So, do we go with using `color` or `fill`?

meghanrjones:
xref https://github.com/GenericMappingTools/gmt/issues/5563 - I think we're leaning towards fill for GMT, but PyGMT already has color in plot and plot3d.

weiji14:
> xref [GenericMappingTools/gmt#5563](https://github.com/GenericMappingTools/gmt/issues/5563) - I think we're leaning towards fill for GMT, but PyGMT already has color in plot and plot3d.

Yeah, I think it would be nice to use `fill` too to standardize with GMT. Just that the `color="G"` alias for plot/psxy has been here for 4 years already since https://github.com/GenericMappingTools/pygmt/blob/aad12e0879bc71a7dbb396298ed87d68c7d53c80/gmt/ps_modules.py#L10 so this would be a pretty significant deprecation. Ok with deprecating this though, because 'color' is a pretty generic term, and 'fill' is more specific to mean just the inside of the symbol/polygon being plotted. What do others think?

willschlitzer:
I think `fill` works better, but it should be changed across the PyGMT modules.

seisman:
I also agree that `fill` is better.

I think it's possible to extend the `use_alias` decorator to support multiple aliases. 
For example, `@use_alias(G=["fill", "color"])` means both `fill` are `color` are accepted.

meghanrjones:
[...]

PR #1932, opened 2022-05-25
Use longname placeholders in the docstrings for common options

seisman:

For "common" options, we have the option description defined in the file [`pygmt/helpers/decorators.py`](https://github.com/GenericMappingTools/pygmt/blob/c70fe880335c1ace9d9fab6ec67d31896edea117/pygmt/helpers/decorators.py#L17). Then in the module wrappers, we can use placeholders like `{R}` in the docstrings, which are automatically replaced by the long docstrings.

Currently, we use single-letter flags as the placeholders (e.g., `{R}`, `{I}`) for most common options, although there are a few exceptions (e.g., `{CPT}`, `{XY}`). It works well but sometimes may cause confusion, especially for non-common options like `-G` and `-I`.

For example, sometimes the "common" option `-G` is aliased to `fill`, but the docstring for placeholder `{G}` indicates it's aliased to `color` (e.g., https://github.com/GenericMappingTools/pygmt/issues/1617). I also remember that someone once incorrectly used the placeholder `{I}` in the docstring, but the `-I` option for the module has very different meanings (for example, `-I` means `inquiry` in some modules).

I find that using long names for these placeholders can avoid these problems and long names are also more readable than short names.

---

In this PR, I rename the placeholder `{I}` to `{spacing}` to show what the changes look like, but I plan to replace all short-name placeholders with long names (e.g., `{R}` => `{region}`).

What do you think about the changes?

References:
- https://docs.generic-mapping-tools.org/6.4/devdocs/long_options.html#common-options


meghanrjones:
I agree this is an improvement. In addition to your comments, this also offers a slight benefit to users of IDEs like vscode which do not process the f-strings when displaying the documentation.

pygmt/tests/test_clib.py
```diff
@@ -532,7 +532,7 @@ def test_virtualfile_from_vectors_one_string_or_object_column(dtype):
             with GMTTempFile() as outfile:
                 lib.call_module("convert", f"{vfile} ->{outfile.name}")
                 output = outfile.read(keep_tabs=True)
-        expected = "".join(f"{i}\t{j}\t{k}\n" for i, j, k in zip(x, y, strings))
+        expected = "".join(f"{incols}\t{distcalc}\t{k}\n" for i, j, k in zip(x, y, strings))
```
weiji14, line 535: These changes in the test files need to be reverted.

pygmt/src/basemap.py
```diff
@@ -82,13 +82,13 @@ def basemap(self, **kwargs):
...
-    {f}
-    {p}
-    {t}
+    {timestamp}
+    {verbose}
+    {xyshift}
+    {panel}
+    {coltypes}
+    {perspective}
+    {transparency}
```
weiji14, line 91: I like the long names, but then there's a small problem with sorting the parameters. Previously we sorted by asciibetical order (e.g. ABCabc), and it made sense since the small-letter GMT parameters like `c`/`f`/`p`/`t` were rarely used. Now that we switch to the long-form, do we still go with the original GMT sorting, or change to sorting by a long-parameter's first character?

pygmt/helpers/decorators.py
```diff
@@ -393,14 +393,14 @@ def fmt_docstring(module_func):
...
-    * ``{W}``: pen
-    * ``{n}``: interpolation
+    * ``{region}``: region (bounding box as west, east, south, north)
+    * ``{projection}``: projection (coordinate system to use)
+    * ``{frame}``: frame (map frame and axes parameters)
+    * ``{timestamp}``: timestamp (insert time stamp logo)
+    * ``{cmap}``: cmap (the color palette table)
+    * ``{color}``: color
+    * ``{pen}``: pen
+    * ``{interpolation}``: interpolation
```
seisman, line 403: I think we can remove this list because the list is incomplete and there is no need to maintain it.

pygmt/src/basemap.py
seisman: I don't have a good answer for this, but packages like numpy and pandas don't sort their parameters alphabetically. I think we also don't have to do that.

pygmt/helpers/decorators.py
seisman: Done in 9dccaa40f5035453cbf64efd3349aa7c1d76f7ca.

pygmt/src/ternary.py
```diff
@@ -52,18 +52,18 @@ def ternary(self, data, **kwargs):
         [*amin*, *amax*, *bmin*, *bmax*, *cmin*, *cmax*].
         Give the min and max limits for each of the three axes **a**, **b**,
         and **c**.
-    {CPT}
-    {G}
+    {cmap}
+    {fill}
```
weiji14:
```suggestion
    {color}
```
Fixes https://github.com/GenericMappingTools/pygmt/actions/runs/3067316855/jobs/4953471467#step:11:28
```python-traceback
  File "/home/runner/work/pygmt/pygmt/pygmt/__init__.py", line 25, in <module>
    from pygmt import datasets
  File "/home/runner/work/pygmt/pygmt/pygmt/datasets/__init__.py", line 5, in <module>
    from pygmt.datasets.earth_age import load_earth_age
  File "/home/runner/work/pygmt/pygmt/pygmt/datasets/earth_age.py", line 10, in <module>
    from pygmt.src import grdcut, which
  File "/home/runner/work/pygmt/pygmt/pygmt/src/__init__.py", line 52, in <module>
    from pygmt.src.ternary import ternary
  File "/home/runner/work/pygmt/pygmt/pygmt/src/ternary.py", line 26, in <module>
    def ternary(self, data, **kwargs):
  File "/home/runner/work/pygmt/pygmt/pygmt/helpers/decorators.py", line 481, in fmt_docstring
    module_func.__doc__ = docstring.format(**filler_text)
KeyError: 'fill'
```

weiji14: Oops nevermind, you beat me to it :laughing:

weiji14 (approved):
Thanks @seisman! Was a little concerned if it's premature to find and replace the not-so-common-options like `{I|` -> `{spacing}` since we're not 100% confident that the longnames for these are considered final for GMT. I.e. that someone won't suddenly decide that `I` should be called `interval` instead of `spacing`. But should be ok since we can always change them again :slightly_smiling_face:  Certainly not a problem for the very-common-options listed in https://docs.generic-mapping-tools.org/6.4/devdocs/long_options.html#common-options such as B/R/J/U/V/X/Y and so on.

Would be good to have a second review (since this is a big PR), but I did scan through each change one by one and they seem ok.

seisman:
Ping @maxrjones @willschlitzer @michaelgrund @yvonnefroehlich

maxrjones:
> Ping @maxrjones @willschlitzer @michaelgrund @yvonnefroehlich

I can take a look later today or tomorrow.

yvonnefroehlich (commented):
So far, all replacements appear correct to me.
The `fill`-`color` inconsistency (`pygmt.Figure.histogram` and `pygmt.Figure.ternary`) seems to be separately discussed in issue https://github.com/GenericMappingTools/pygmt/issues/1617.

pygmt/src/nearneighbor.py
```diff
@@ -84,9 +84,9 @@ def nearneighbor(data=None, x=None, y=None, z=None, **kwargs):
...
         Arrays of x and y coordinates and values z of the data points.
 
-    {I}
+    {spacing}
 
-    {R}
+    {region}
 
     search_radius : str
         Sets the search radius that determines which data points are considered
```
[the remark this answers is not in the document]
seisman: > In a few functions and methods, blank lines are used between the different parameters. Currently, it appears unclear to me when and why this is done.

These blank lines are not necessary. They're there because these docstrings were written by different authors at different time.

maxrjones (approved):

PR #1964, opened 2022-06-16
Figure.text: Deprecate parameter "incols" to "use_word" (remove in v0.10.0) 

### LINKED ISSUE #1947: `pygmt.Figure.text()`: `incols` (`-i`) parameter does not work as described in the documentation

The `incols` (`-i`) parameter of the method `pygmt.Figure.text()` does not provide the functionality as described in the documentation.

In this minimal example the data is provided via a txt-file ([test_text.txt](https://github.com/GenericMappingTools/pygmt/files/8883727/test_text.txt)) with three columns (x,y,text) and two lines and passed through the `textfiles` parameter. I tested both passing a string and a 1-D array through the `incols` parameter. When trying to resort the columns 'x' and 'y' via `incols=[1,0,2]` or `incols='1,0,2'` the below given  error arises (cases (iv), (v)). This problem occurs already when using the `incols` parameter without any resorting of the columns, i. e. `incols=[0,1,2]` or `incols='0,1,2'` (cases (ii),  (iii)). This holds also for skipping columns (no code example provided).

Relation to _GMT_
In the _GMT_ documentation (version 6.3) I can not find **`-i`**_`flags`_ for [text](https://docs.generic-mapping-tools.org/latest/text.html) (in contrast to [plot](https://docs.generic-mapping-tools.org/latest/plot.html)), only **`-it`**_`word`_ and **`-:`**`[i|o]` are listed. However, the _PyGMT_ documentation (version 0.6.1) describes the same functionality of the `incols` parameter for [pygmt.Figure.text()](https://www.pygmt.org/v0.6.1/api/generated/pygmt.Figure.text.html#pygmt.Figure.text) and [pygmt.Figure.plot()](https://www.pygmt.org/v0.6.1/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot). Maybe it is not a bug, but an error in the _PyGMT_ documentation (?), since the error message below says:
```
text [ERROR]: Option -i: Must give -it<word> from 0 (first) to nwords-1
```

```python
import pygmt as gmt
# Is this an upstream bug?
#gmt.config(GMT_VERBOSE='d')

#-----------------------------------------------------------------------------
#-----------------------------------------------------------------------------
# general stuff
size = 10
RR_use = '0/' + str(size) + '/0/' + str(size)
JJ_use = 'X' + str(size) + 'c'

# (x, y[, angle, font, justify], text)
# here only two lines with x,y,text
data_text = 'test_text.txt'

#-----------------------------------------------------------------------------
#-----------------------------------------------------------------------------
# create figure object
fig = gmt.Figure()

fig.basemap(region=RR_use, projection=JJ_use, frame='agf',)

#-----------------------------------------------------------------------------
#%% (i) no usage of incols parameter - for reference, works correctly
fig.text(textfiles=data_text,)
fig.show()
#-----------------------------------------------------------------------------
#%% (ii) passing a string to incols parameter (no re-ordering of columns)
fig.text(textfiles=data_text, incols='0,1,2',)
fig.show()
#-----------------------------------------------------------------------------
#%% (iii) passing a 1-D array to incols parameter (no re-ordering of columns)
fig.text(textfiles=data_text, incols=[0,1,2],)
fig.show()
#-----------------------------------------------------------------------------
#%% (iv) passing a string to incols parameter (re-ordering of columns)
fig.text(textfiles=data_text, incols='1,0,2',)
fig.show()
#-----------------------------------------------------------------------------
#%% (v) passing a 1-D array to incols parameter (re-ordering of columns)
fig.text(textfiles=data_text, incols=[1,0,2],)
fig.show()
```

**Full error message** 

For the cases (ii) to (v):

```
text [ERROR]: Option -i: Must give -it<word> from 0 (first) to nwords-1.

  File ~\C2\EigeneDokumente\Studium\Promotion\E_GMT\00_testing\question_text\text_read_from_file.py:55 in <module>
    fig.text(textfiles=data_text, incols='1,0,2',)

  File C:\ProgramData\Anaconda3\envs\pygmt_env\lib\site-packages\pygmt\helpers\decorators.py:585 in new_module

  File C:\ProgramData\Anaconda3\envs\pygmt_env\lib\site-packages\pygmt\helpers\decorators.py:725 in new_module

  File C:\ProgramData\Anaconda3\envs\pygmt_env\lib\site-packages\pygmt\src\text.py:236 in text_
    lib.call_module("text", build_arg_string(kwargs, infile=fname))

  File C:\ProgramData\Anaconda3\envs\pygmt_env\lib\site-packages\pygmt\clib\session.py:502 in call_module
    raise GMTCLibError(

GMTCLibError: Module 'text' failed with status code 72:
text [ERROR]: Option -i: Must give -it<word> from 0 (first) to nwords-1.
```

**System information**

Output of `python -c "import pygmt; pygmt.show_versions()"`:

```
PyGMT information:
  version: v0.6.1
System information:
  python: 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:17:03) [MSC v.1929 64 bit (AMD64)]
  executable: C:\ProgramData\Anaconda3\envs\pygmt_env\python.exe
  machine: Windows-10-10.0.19043-SP0
Dependency information:
  numpy: 1.22.3
  pandas: 1.4.2
  xarray: 2022.3.0
  netCDF4: 1.5.8
  packaging: 21.3
  geopandas: 0.10.2
  ghostscript: 9.54.0
  gmt: 6.3.0
GMT library information:
  binary dir: C:/ProgramData/Anaconda3/envs/pygmt_env
  cores: 4
  grid layout: rows
  library path: C:/ProgramData/Anaconda3/envs/pygmt_env/Library/bin/gmt.dll
  padding: 2
  plugin dir: C:/ProgramData/Anaconda3/envs/pygmt_env/Library/bin/gmt_plugins
  share dir: C:/Program Files (x86)/gmt6/share
  version: 6.3.0
```

yvonnefroehlich:

Fixes #1947 

**Reminders**

**Slash Commands**


pygmt/src/text.py
```diff
@@ -156,7 +156,12 @@ def text_(
     {e}
     {f}
     {h}
-    {i}
+    incols : str
```
seisman, line 159: Current, we alias `-i` to `incols` and need to use `incols="t1"` to select the 2nd word. Here is an example:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=True)
fig.text(x=5, y=5, text="A B C", incols="t1")
fig.show()
```
We can alias `-it` to `incols` and use `incols=1` instead, which is more Pythonic, although it's not backward-compatible. 

Or we can choose a better alias name like `use_word=1`.

Ping @GenericMappingTools/pygmt-maintainers for more thoughts on this.

weiji14: Either way, it looks like we'll need to go through a deprecation cycle? I don't have a strong preference on whether to keep using `incols` for `-it`  or switch to `use_word`. Had a quick look at https://joa-quim.github.io/GMTjl_doc/documentation/text/index.html and doesn't seem like they aliased `it` for `text` so no guidance to follow.

weiji14: @yvonnefroehlich, what do you think is better? Keep using `incols` as the alias (but change it to `-it` instead of `-i`), or use a different alias like `use_word`?

yvonnefroehlich: > @yvonnefroehlich, what do you think is better? Keep using `incols` as the alias (but change it to `-it` instead of `-i`), or use a different alias like `use_word`?

I think using a different, more descriptive alias, like `use_word`, is better. 

Maybe users (knowing the GMT syntax) are confused, when both `-i` and `-it` are aliased as `incols`. If keeping `incols`, I would stay with `'t`_`word`_`'` as argument. (This were actually my thoughts behind my question in https://github.com/GenericMappingTools/pygmt/issues/1947#issuecomment-1157267031. Thanks @seisman for the nice formulation at the top of this pull request!)
Furthermore, `incols` indicates that the user can specify [numeric] columns, but here a specific word of the whole text in the third column can be specified by passing the word order or position, i. e. for example 0 means not the first column but the first word.

seisman: I think we will go with `use_word`.

@yvonnefroehlich Please make the changes when you have time.

yvonnefroehlich:
> I think we will go with `use_word`.
> 
> @yvonnefroehlich Please make the changes when you have time.

I started working on this. At the moment the deprecation cycle includes (only) two minor releases, starting at 0.8.0 and ending at 1.0.0 (not 0.10.0 corresponding to the [announcement draft]( https://hackmd.io/KZEh0ZOTSbSxC_qUC-5gWA#-Upcoming-deprecations)). I am unsure about how to
- change `i` to `it` or add the `t` later/externally.
- change the format from string to integer (the docstring is already adjusted corresponding to `use_word=<integer>`).
- write a test as proposed in the [Maintainers Guide](https://www.pygmt.org/dev/maintenance.html#backwards-compatibility-and-deprecation-policy).

pygmt/src/text.py
```diff
@@ -15,6 +16,7 @@
 
 
 @fmt_docstring
+@deprecate_parameter("incols", "use_word", "v0.8.0", remove_version="v1.0.0")
```
seisman, line 19: v1.0.0 is our target stable release for PyGMT. It definitely will take more times so it will be v0.10.0 after two minor releases.
```suggestion
@deprecate_parameter("incols", "use_word", "v0.8.0", remove_version="v0.10.0")
```

seisman:
> * change `i` to `it` or add the `t` later/externally.

I think changing `i` to `it` is easier.

> * change the format from string to integer (the docstring is already adjusted corresponding to `use_word=<integer>`).

Yes.

pygmt/src/text.py
```diff
@@ -33,7 +35,7 @@
     e="find",
     f="coltypes",
     h="header",
-    i="incols",
+    i="use_word",
```
seisman, line 38: We are aliasing `-it` to `word`, so need to change `i` to `it` here:
```suggestion
    it="use_word",
```

yvonnefroehlich: I try to work on this over the weekend. I am sorry for the delay.

seisman (approved):
Do you think we should add a test for it?

willschlitzer (approved):
Not necessary for just the deprecation of a parameter, but could a new test be written or an existing test modified to include this new parameter? I think it might be good to make sure that `text` is consistently reading a string from an input file.

seisman:
> Not necessary for just the deprecation of a parameter, but could a new test be written or an existing test modified to include this new parameter? I think it might be good to make sure that `text` is consistently reading a string from an input file.

Yes, I agree. This can be done in a separate PR.

PR #1980, opened 2022-06-26
Exclude researchgate when checking links

seisman:

Closes https://github.com/GenericMappingTools/pygmt/issues/1979


weiji14 (approved):
RIght, so for context, the link checker failed for researchgate in #1948, #1970, #1979 (about once a week), is that why we should skip it? The link to https://www.researchgate.net/project/PyGMT-A-Python-interface-for-the-Generic-Mapping-Tools seems to work fine now though, but I suppose we can skip it :slightly_smiling_face:

PR #2005, opened 2022-07-12
Fix figure labels in grdhisteq tutorial

maxrjones:

This PR fixes the axis labels to agree with the histogram type.

Fixes #

**Reminders**

**Slash Commands**


examples/tutorials/advanced/grid_equalization.py
```diff
@@ -210,12 +211,13 @@
...
-            region=[0, divisions, 0, 40],
+            region=[-1, divisions, 0, 40],
             series=[0, divisions, 1],
-            frame=["wnSE", "xaf+lElevation (m)", "yaf+lCounts"],
+            frame=["wnSE", "xaf+lRelative elevation", "yaf+lPercent frequency"],
             cmap=True,
             histtype=1,
             pen="1p,black",
+            center=True,
         )
```
seisman, line 221: For the last example, I tried to change `region=[-1, divisions, 0, 40],` to `region="a"`, so that we don't have to manually change the x- and y-limits. It gives me the following image, so it looks like the automatic determination of region is not perfect for this case.
map

maxrjones: Yes, there may be a bug. I remember also trying `region=[-0.5, 8.5, 0, 40]` and the xmin still being 0.

Issue #1984, opened 2022-06-29
Unable to use dataarray if its a variable of a dataset


mdtanker:
**Description of the problem**

If a dataarray has been converted to a DataSet with `xarray.DataArray.to_dataset()`, calling the dataarray in PyGMT modules fails. See below that output of `grid` and `dataset.height` are equivalent.

**Full code that generated the error**

```python
import pygmt
import xarray as xr

region = [-28, -10, 62, 68]

grid = pygmt.datasets.load_earth_relief(resolution="03m", region=region)

dataset = grid.to_dataset(name='height')

print(pygmt.grdinfo(grid))
print(pygmt.grdinfo(dataset.height))
```

**Full error message**

```
grdcut [NOTICE]: Remote data courtesy of GMT data server oceania [http://oceania.generic-mapping-tools.org]
grdcut [NOTICE]: SRTM15 Earth Relief at 20x20 arc minutes reduced by Gaussian Cartesian filtering (37.1 km fullwidth) [Tozer et al., 2019].
grdcut [NOTICE]:   -> Download grid file [831K]: earth_relief_20m_p.grd
: Title: 
: Command: 
: Remark: 
: Pixel node registration used [Geographic grid]
: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
: x_min: -28 x_max: -10 x_inc: 0.333333333333 (20 min) name: x n_columns: 54
: y_min: 62 y_max: 68 y_inc: 0.333333333333 (20 min) name: y n_rows: 18
: v_min: -2240 v_max: 1664 name: z
: scale_factor: 1 add_offset: 0
: format: classic
: Default CPT: 

grdinfo [ERROR]: File /tmp/pygmt-yyhpy7cb.nc was not found
grdinfo [ERROR]: Cannot find file /tmp/pygmt-yyhpy7cb.nc
grdinfo [ERROR]: Must specify one or more input files
Output exceeds the size limit. Open the full output data in a text editor
---------------------------------------------------------------------------
GMTCLibError                              Traceback (most recent call last)
Untitled-1.ipynb Cell 1' in <cell line: 11>()
      8 dataset = grid.to_dataset(name='height')
     10 print(pygmt.grdinfo(grid))
---> 11 print(pygmt.grdinfo(dataset.height))

[...]

PR #2009, opened 2022-07-16
Let GMTDataArrayAccessor check if the grid source file exists

seisman:

If the netCDF source file doesn't exist, running `grdinfo` on the netCDF source file gives the following error:
```
grdinfo [ERROR]: Cannot find file /tmp/xxx/xxx.nc
grdinfo [ERROR]: Must specify one or more input files
```
The error is unfriendly and confusing. Instead, we can check if the netCDF source file exists. If it doesn't exist, we raise a `ValueError` exception and fallback to registration=0 and gtype=0/ 

Partially address https://github.com/GenericMappingTools/pygmt/issues/1984 and https://github.com/GenericMappingTools/pygmt/issues/524#issuecomment-1121803929

Please note that this PR doesn't solve the problems in #1984 and #524, because the registration and gtype information are incorrect.

### DIFF
--- a/pygmt/accessors.py
+++ b/pygmt/accessors.py
@@ -1,6 +1,8 @@
 """
 GMT accessor methods.
 """
+from pathlib import Path
+
 import xarray as xr
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.src.grdinfo import grdinfo
@@ -29,6 +31,8 @@ class GMTDataArrayAccessor:
         self._obj = xarray_obj
         try:
             self._source = self._obj.encoding["source"]  # filepath to NetCDF source
+            if not Path(self._source).exists():
+                raise ValueError(f"Grid source file {self._source} doesn't exist.")
             # Get grid registration and grid type from the last two columns of
             # the shortened summary information of `grdinfo`.
             self._registration, self._gtype = map(


seisman:
Ping @maxrjones and @weiji14 for reviewing.

pygmt/accessors.py
```diff
@@ -29,6 +31,8 @@ def __init__(self, xarray_obj):
         self._obj = xarray_obj
         try:
             self._source = self._obj.encoding["source"]  # filepath to NetCDF source
+            if not Path(self._source).exists():
+                raise ValueError(f"Grid source file {self._source} doesn't exist.")
```
weiji14, line 35: Suggest to use https://docs.python.org/3/library/exceptions.html#FileNotFoundError instead. Also need a unit test for this line to ensure code coverage. 
```suggestion
            if not Path(self._source).exists():
                raise FileNotFoundError(f"Grid source file {self._source} doesn't exist.")
```

seisman:
I've added a unit test in 37869d470b45773c02f731ec8c006651d1db4c9b, and the test helps find a new bug.

At the end of the test, I manually set the registration and gtype information and expect the two `assert` statements to pass, but they don't. So there must be another bug.
```
    # manually set the registration and gtype
    dataset.height.gmt.registration = 1
    dataset.height.gmt.gtype = 1
    # the registration and gtype should be correct now.
    assert dataset.height.gmt.registration == 1
    assert dataset.height.gmt.gtype == 1
```

seisman:
> I've added a unit test in [37869d4](https://github.com/GenericMappingTools/pygmt/commit/37869d470b45773c02f731ec8c006651d1db4c9b), and the test helps find a new bug.
> 
> At the end of the test, I manually set the registration and gtype information and expect the two `assert` statements to pass, but they don't. So there must be another bug.
> 
> ```
>     # manually set the registration and gtype
>     dataset.height.gmt.registration = 1
>     dataset.height.gmt.gtype = 1
>     # the registration and gtype should be correct now.
>     assert dataset.height.gmt.registration == 1
>     assert dataset.height.gmt.gtype == 1
> ```

The [documentation of xarray accessor](https://docs.xarray.dev/en/stable/internals/extending-xarray.html) explains why they don't work as expected:

> Accessors are created once per DataArray and Dataset instance. New instances, like those created from arithmetic operations or when accessing a DataArray from a Dataset (ex. ds[var_name]), will have new accessors created.

So every time `dataset.height.gmt` is used, the `__init__` function is called. It's the limitation of the xarray accessors and I think we can do nothing here.

weiji14:
> The [documentation of xarray accessor](https://docs.xarray.dev/en/stable/internals/extending-xarray.html) explains why they don't work as expected:
> 
> > Accessors are created once per DataArray and Dataset instance. New instances, like those created from arithmetic operations or when accessing a DataArray from a Dataset (ex. ds[var_name]), will have new accessors created.
> 
> So every time `dataset.height.gmt` is used, the `__init__` function is called. It's the limitation of the xarray accessors and I think we can do nothing here. 

The accessor `dataset.height.gmt` is created once, but the variable or property should be able to be modified using a setter method. Still think this is an issue, but need time to look into it. Don't we have some tests that actually modify the accessor values?

pygmt/accessors.py
```diff
@@ -29,12 +31,16 @@ def __init__(self, xarray_obj):
...
+                )
             # Get grid registration and grid type from the last two columns of
             # the shortened summary information of `grdinfo`.
             self._registration, self._gtype = map(
                 int, grdinfo(self._source, per_column="n").split()[-2:]
             )
-        except (KeyError, ValueError):
+        except (KeyError, ValueError, FileNotFoundError):
             self._registration = 0  # Default to Gridline registration
             self._gtype = 0  # Default to Cartesian grid type
```
seisman, line 45: I wonder if we should change these lines to the following codes, which I think are more readable:
```
        self._source = self._obj.encoding.get("source")
        if self._source is not None and Path(self._source).exists():
            try:
                # Get grid registration and grid type from the last two columns of
                # the shortened summary information of `grdinfo`.
                self._registration, self._gtype = map(
                    int, grdinfo(self._source, per_column="n").split()[-2:]
                )
            except ValueError:
                self._registration = 0
                self._gtype = 0
        else:
            self._registration = 0
            self._gtype = 0
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers for review.

pygmt/accessors.py
weiji14: So we will silently fallback to registration=0 and gtype=0 if `grdinfo` cannot determine the grid registration and type? Ok I guess, but please add the code comments back :slightly_smiling_face:

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
+    """
+    # load the 05m earth relief grid, which is stored as tiles
+    grid = load_earth_relief(
+        resolution="05m", region=[0, 5, -5, 5], registration="pixel"
+    )
+    # registration and gtype are correct
+    assert grid.gmt.registration == 1
+    assert grid.gmt.gtype == 1
+    # the source grid file is defined but doesn't exists
+    assert len(grid.encoding["source"]) > 0
```
weiji14, line 121: Maybe better to check the file extension is a NetCDF, rather than the length of the filename
```suggestion
    assert grid.encoding["source"].endswith(".nc")
```

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
+    assert not Path(grid.encoding["source"]).exists()
+
+    # For a sliced grid, fallback to default registration and gtype,
+    # because the source grid file doesn't exists.
```
weiji14, line 125:
```suggestion
    # because the source grid file doesn't exist.
```

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
         assert grid.gmt.gtype == 1  # geographic coordinate type
     finally:
         os.remove(fname)
+
+
+def test_accessor_grid_source_file_not_exist():
+    """
+    Check that the accessor fallbacks to the default registration and gtype
+    when the grid source file (i.e., grid.encoding["source"]) doesn't exists.
```
weiji14, line 111:
```suggestion
    when the grid source file (i.e., grid.encoding["source"]) doesn't exist.
```

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
+    """
+    # load the 05m earth relief grid, which is stored as tiles
+    grid = load_earth_relief(
+        resolution="05m", region=[0, 5, -5, 5], registration="pixel"
+    )
+    # registration and gtype are correct
+    assert grid.gmt.registration == 1
+    assert grid.gmt.gtype == 1
+    # the source grid file is defined but doesn't exists
```
weiji14, line 120:
```suggestion
    # the source grid file is defined but doesn't exist
```

weiji14 (approved):
Sorry, 2 more typos, otherwise ok.

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
     finally:
         os.remove(fname)
+
+
+def test_accessor_grid_source_file_not_exist():
+    """
+    Check that the accessor fallbacks to the default registration and gtype
+    when the grid source file (i.e., grid.encoding["source"]) doesn't exist.
+    """
+    # load the 05m earth relief grid, which is stored as tiles
```
michaelgrund, line 113:
```suggestion
    # Load the 05m earth relief grid, which is stored as tiles
```

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
+    grid = load_earth_relief(
+        resolution="05m", region=[0, 5, -5, 5], registration="pixel"
+    )
+    # registration and gtype are correct
```
michaelgrund, line 117:
```suggestion
    # Registration and gtype are correct
```

pygmt/tests/test_accessor.py
```diff
@@ -101,3 +103,32 @@ def test_accessor_sliced_datacube():
...
+    assert grid.gmt.registration == 1
+    assert grid.gmt.gtype == 1
+    # the source grid file is defined but doesn't exist
```
michaelgrund, line 120:
```suggestion
    # The source grid file is defined but doesn't exist
```

michaelgrund (approved):
Only minor consistency changes, otherwise looks good!

PR #2020, opened 2022-07-22
Use the org-wide code of conduct

seisman:

Update the links of the CODE_OF_CONDUCT.md file to https://github.com/GenericMappingTools/.github/blob/main/CODE_OF_CONDUCT.md.

Following https://github.com/GenericMappingTools/.github/pull/1.


maxrjones:
I think that minimizing the contents of `CODE_OF_CONDUCT.md` to something like the following could be better than removing the file completely.
```
All participants in the PyGMT community must abide by the [Generic Mapping Tools organization Code of Conduct](https://github.com/GenericMappingTools/.github/blob/main/CODE_OF_CONDUCT.md)
```
The reason why I think this could be better is that some people may look either in the source distribution or the repo files for a code of conduct without knowing to check the GitHub sidebar. This seems like a good middle ground that doesn't require separate maintenance.

seisman:
Sounds good and I've added the minimized version of COC back.

maxrjones (approved):
Thanks!

weiji14 (approved):
Looks good. Agree with Max to have a placeholder CODE_OF_CONDUCT.md still, if only to have all green ticks on https://github.com/GenericMappingTools/pygmt/community :smile:

michaelgrund (approved):

PR #2057, opened 2022-08-17
NEP29: Test PyGMT on NumPy 1.23 and 1.21

weiji14:

Bumps [numpy](https://github.com/numpy/numpy) from 1.22.4 to 1.23.2. NumPy 1.23.0 was released on 22 Jun 2022.
- [Release notes](https://github.com/numpy/numpy/releases)
- [Changelog](https://github.com/numpy/numpy/blob/v1.23.0/doc/changelog/1.23.0-changelog.rst)
- [Commits](https://github.com/numpy/numpy/compare/v1.22.4...v1.23.2)

This is in line with PyGMT's policy on [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) at https://www.pygmt.org/v0.7.0/maintenance.html#dependencies-policy, xref #1074.

Also setting minimum NumPy version on CI to 1.21, to bypass the `TypeError: <class 'numpy.typing._dtype_like._SupportsDType'> is not a generic class`  problem with NumPy 1.20 and `xarray==2022.6.0` so that unit tests pass. See also https://github.com/pydata/xarray/issues/6818

Note that the branch protection rules at https://github.com/GenericMappingTools/pygmt/settings/branches will need to be changed to use Python 3.10/Numpy 1.22 instead of Python 3.10/Numpy 1.23 before this Pull Request is merged.

Supersedes #1701

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -40,11 +40,11 @@ jobs:
             isDraft: true
           - os: windows-latest
             isDraft: true
-        # Pair Python 3.8 with NumPy 1.20 and Python 3.10 with NumPy 1.23
+        # Pair Python 3.8 with NumPy 1.21 and Python 3.10 with NumPy 1.23
         # Only install optional packages on Python 3.10/NumPy 1.23
         include:
           - python-version: '3.8'
-            numpy-version: '1.20'
+            numpy-version: '1.21'
             optional-packages: ''
           - python-version: '3.10'
             numpy-version: '1.23'
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -40,14 +40,14 @@ jobs:
             isDraft: true
           - os: windows-latest
             isDraft: true
-        # Pair Python 3.8 with NumPy 1.20 and Python 3.10 with NumPy 1.22
-        # Only install optional packages on Python 3.10/NumPy 1.22
+        # Pair Python 3.8 with NumPy 1.20 and Python 3.10 with NumPy 1.23
+        # Only install optional packages on Python 3.10/NumPy 1.23
         include:
           - python-version: '3.8'
             numpy-version: '1.20'
             optional-packages: ''
           - python-version: '3.10'
-            numpy-version: '1.22'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yaml
```diff
@@ -40,14 +40,14 @@ jobs:
             isDraft: true
           - os: windows-latest
             isDraft: true
-        # Pair Python 3.8 with NumPy 1.20 and Python 3.10 with NumPy 1.22
-        # Only install optional packages on Python 3.10/NumPy 1.22
+        # Pair Python 3.8 with NumPy 1.20 and Python 3.10 with NumPy 1.23
+        # Only install optional packages on Python 3.10/NumPy 1.23
         include:
           - python-version: '3.8'
             numpy-version: '1.20'
```
weiji14, line 47: Question. Do we want to just set our minimum tested NumPy version to 1.21 in the CI since `xarray` doesn't work with Numpy 1.20 and they won't be releasing a patch as per https://github.com/pydata/xarray/issues/6818#issuecomment-1215278493? Due date to drop Numpy 1.20 is Jan 31, 2023 (5 months) per https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule.

willschlitzer: I think setting numpy to v1.21 makes sense. I'm not sure what the plan is for the pygmt v0.8 release, but I don't think we'll have 2 releases before January 2023, so our next release should be up to date on minimum versions.

weiji14: Cool. Just to be clear, the suggesting is to just set NumPy 1.21 for the **Continuous Integration (CI)** so that the unit tests pass. We shouldn't change the minimum supported NumPy mentioned in https://github.com/GenericMappingTools/pygmt/tree/v0.7.0#compatibility-with-gmtpythonnumpy-versions until after Jan 2023 (though users using NumPy 1.20 will need to figure out the `xarray` dependency themselves).

seisman (approved):
Looks good to me. 

The two failing tests on Windows are unrelated to this PR and should be addressed separately.

PR #2058, opened 2022-08-17
Fix typos in documentation for Polar projection

yvonnefroehlich:

This PR aims to fix a few typos in the documentation for the [Polar projection](https://www.pygmt.org/dev/projections/nongeo/polar.html).
In case 'r axis' (no hyphen) is preferred instead of 'r-axis', I can change this. Following PR https://github.com/GenericMappingTools/pygmt/pull/2028 'Earth' is  capitalized.

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/projections/nongeo/polar.py
+++ b/examples/projections/nongeo/polar.py
@@ -23,7 +23,7 @@ The following customizing modifiers are available:
   definition); **+a** indicates that the input data is rotated clockwise
   relative to the north direction (geographical azimuth angle).
 
-- **+r**\ *offset*: represents the offset of the r axis. This modifier allows
+- **+r**\ *offset*: represents the offset of the r-axis. This modifier allows
   you to offset the center of the circle from r=0.
 
 - **+t**\ *origin*: sets the angle corresponding to the east direction which is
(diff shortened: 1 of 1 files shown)


maxrjones (approved):
Looks good, thanks!

If you wanted to improve this even further, it might be nice to have a comment describing the different sections rather than `# ============`, which does not render separate sections in https://www.pygmt.org/latest/projections/nongeo/polar.html#sphx-glr-projections-nongeo-polar-py.

seisman (approved):

michaelgrund (approved):

yvonnefroehlich:
> If you wanted to improve this even further, it might be nice to have a comment describing the different sections rather than `# ============`, which does not render separate sections in https://www.pygmt.org/latest/projections/nongeo/polar.html#sphx-glr-projections-nongeo-polar-py.

Thanks @maxrjones! I can try to work on this in a separat PR.

Maybe we can further use the `+t` modifier of the `frame` parameter of  the `pygmt.Figure.basemap` method to set the titles? Currently, this is done via multiple calls of the `pygmt.Figure.text` method.

maxrjones:
> Maybe we can further use the +t modifier of the frame parameter of the pygmt.Figure.basemap method to set the titles? Currently, this is done via multiple calls of the pygmt.Figure.text method.

Sorry for the very slow response. `+t` cannot plot on multiple lines, but it could be possible to use both `+t` and `+s` for a title and subtitle. The `+s` feature may not have been added when the example was originally made. However, It may be more confusing to use those modifiers rather than a couple text calls until the syntax for titles and subtitles is simplified.

yvonnefroehlich:
> Sorry, just noticed https://github.com/GenericMappingTools/pygmt/pull/2110 as I'm slowly trying to catch up on GitHub notifications. I'll review that, thanks for opening the PR!

No worries and no hurry @maxrjones! The related PR is still WIP :wink:.

> +t cannot plot on multiple lines, but it could be possible to use both +t and +s for a title and subtitle. The +s feature may not have been added when the example was originally made. However, It may be more confusing to use those modifiers rather than a couple text calls until the syntax for titles and subtitles is simplified.

Up on PR https://github.com/GenericMappingTools/gmt/pull/4562 it is possible to accomplish line breaks within titles via `@^`. But of course, I am open to discuss this change during review.

PR #1848, opened 2022-03-26
Migrate project metadata from setup.py to pyproject.toml following PEP621

seisman:
**Description of proposed changes**

Fixes #1836.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2069, opened 2022-08-20
Add optional dependencies to pyproject.toml

seisman:

As suggested in https://github.com/GenericMappingTools/pygmt/pull/1848#discussion_r949080477, this PR adds `optional-dependencies` to pyproject.toml so that people can run `pip install pygmt[all]` to install pygmt with optional dependencies included. 

To test it locally, you can check out this branch, change into the root directory, and run
```
pip install .
```
and 
```
pip install .[all]
```
you should see that `pip install .[all]` also check if geopandas is installed while `pip install .` doesn't.

Address https://github.com/GenericMappingTools/pygmt/pull/1848#discussion_r949080477.

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -220,8 +220,8 @@ This installs the latest stable release from
 
 .. tip::
 
-   You can also run ``pip install pygmt[all]`` to install pygmt alongside with
-   all its optional dependencies.
+   You can also run ``pip install pygmt[all]`` to install pygmt with
+   all of its optional dependencies.
 
 Alternatively, you can install the latest development version from
 `TestPyPI <https://test.pypi.org/project/pygmt>`__::
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -218,6 +218,11 @@ This installs the latest stable release from
 
     pip install pygmt
 
+.. tips::
+
+   You can also run ``pip install pygmt[all]`` to install pygmt alongside with
+   all its optional dependencies.
+
 Alternatively, you can install the latest development version from
 `TestPyPI <https://test.pypi.org/project/pygmt>`__::
 
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -218,7 +218,7 @@ This installs the latest stable release from
 
     pip install pygmt
 
-.. tips::
+.. tip::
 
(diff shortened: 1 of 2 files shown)


seisman:
I'm not sure if we should update the `pip install pygmt` command to `pip install pygmt[all]`
in the installation instructions (https://www.pygmt.org/dev/install.html).

pyproject.toml
```diff
@@ -32,6 +32,11 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+[project.optional-dependencies]
+all = [
+    "geopandas"
```
weiji14, line 37: Should we add `ipython` as well as mentioned in https://www.pygmt.org/v0.7.0/install.html#dependencies?
```suggestion
    "geopandas",
    "ipython"
```

weiji14 (commented):
> I'm not sure if we should update the `pip install pygmt` command to `pip install pygmt[all]` in the installation instructions (https://www.pygmt.org/dev/install.html).

Wouldn't recommend replacing it with `pip install pygmt[all]`, but I think we can mention the `pip install pygmt[all]` command in https://www.pygmt.org/v0.7.0/install.html#using-pip and indicate that this will install optional dependencies like `geopandas`.

pyproject.toml
seisman: Yes, because we also try to import IPython in our codes.

doc/install.rst
```diff
@@ -218,6 +218,11 @@ This installs the latest stable release from
 
     pip install pygmt
 
+.. tips::
```
seisman, line 221:
```suggestion
.. tip::
```

doc/install.rst
```diff
@@ -218,6 +218,11 @@ This installs the latest stable release from
 
     pip install pygmt
 
+.. tip::
+
+   You can also run ``pip install pygmt[all]`` to install pygmt alongside with
+   all its optional dependencies.
```
weiji14, line 224:
```suggestion
   You can also run ``pip install pygmt[all]`` to install pygmt with
   all of its optional dependencies.
```

weiji14 (approved):

Issue #924, opened 2021-02-17
Remove the xshift (X) and yshift (Y) aliases from all plotting modules?


seisman:
In [the Zen of Python](https://www.python.org/dev/peps/pep-0020/):

> There should be one-- and preferably only one --obvious way to do it.

Now we have two different methods to shift origins:

1. passing arguments like `xshift="2c", yshift="1c"` to any plotting modules
2. explicitly call `Figure.shift_origin(xshift="2c", yshift="1c")`

GMT users are familiar with method 1, but IMHO method 2 is more pythonic. As I can see, these two methods are equivalent.

Should we remove **xshift** and **yshift** from all plotting modules, so that we can better follow the zen of python, and also make the docstrings shorter.

If yes, we need to do more things:

1. Remove **xshift** and **yshift** alias from docstrings of all plotting functions
2. Raise a warning/error if **xshift**, **yshift**, **X** or **Y** is passed.
3. Improve the docstrings of `Figure.shift_origin()`.

seisman:
Similar to this issue, I also don't like the **timestamp** (-U) alias. I'm in favor of removing **timestamp** too, and add a new method ``Figure.timestamp()``.

We can discuss **xshift** and **yshift** first. If we all agree, I'll open another issue to better track and discuss the ``Figure.timestamp()`` method.

seisman:
Ping @GenericMappingTools/python and @GenericMappingTools/python-maintainers for comments and thoughts.

weiji14:
For context, the xshift/yshift aliases were added in #624. While I agree with that line in the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) "There should be one-- and preferably only one --obvious way to do it.", what comes higher up before that line is "Readability counts.", and that is what the xshift/yshift aliases (and other PyGMT aliases) was originally intended for.

> 3. Improve the docstrings of `Figure.shift_origin()`.

Agree that this should be done, regardless of whether we decide to deprecate xshift (X) and yshift (Y).

> 1. Remove **xshift** and **yshift** alias from docstrings of all plotting functions
> 2. Raise a warning/error if **xshift**, **yshift**, **X** or **Y** is passed.

To be honest, this would be a good opportunity to practice Deprecation Warning cycles in PyGMT. But I'm not keen on handling this work, so my vote would be to keep the existing behaviour (xshift/yshift still works), but I'll respect people's opinions if deprecating xshift/yshift is a more popular opinion.

[...]

PR #2071, opened 2022-08-20
Deprecate xshift (X) and yshift (Y) aliases from all plotting modules (remove in v0.12.0)

michaelgrund:

This PR addresses #924 and removes the xshift (X) and yshift (Y) aliases from all plotting modules and gallery examples.

Modules:

* [x] Figure.basemap
* [x] Figure.coast
* [x] Figure.colorbar
* [x] Figure.contour
* [x] Figure.grdcontour
* [x] Figure.grdimage
* [x] Figure.grdview
* [x] Figure.histogram
* [x] Figure.image
* [x] Figure.legend
* [x] Figure.logo
* [x] Figure.meca
* [x] Figure.plot
* [x] Figure.plot3d
* [x] Figure.rose
* [x] Figure.solar
* [x] Figure.subplot
* [x] Figure.text
* [x] Figure.velo
* [x] Figure.wiggle

Gallery examples:

* [x] scatter3d.py

**Reminders**

**Slash Commands**


michaelgrund:
Should we add the warning/error if xshift, yshift, X or Y are passed directly into each function or create a correspondig decorator @GenericMappingTools/pygmt-maintainers? Exemplary I added one in basemap.py.

seisman:
> Should we add the warning/error if xshift, yshift, X or Y are passed directly into each function or create a correspondig decorator @GenericMappingTools/pygmt-maintainers? Exemplary I added one in basemap.py.

I feel it's possible to add the warning in the `use_alias` decorator, instead of creating a new decorator.

michaelgrund:
> > Should we add the warning/error if xshift, yshift, X or Y are passed directly into each function or create a correspondig decorator @GenericMappingTools/pygmt-maintainers? Exemplary I added one in basemap.py.
> 
> I feel it's possible to add the warning in the `use_alias` decorator, instead of creating a new decorator.

I updated the  `use_alias` decorator. Is this the direction you had in mind @seisman?

seisman:
> > > Should we add the warning/error if xshift, yshift, X or Y are passed directly into each function or create a correspondig decorator @GenericMappingTools/pygmt-maintainers? Exemplary I added one in basemap.py.
> > 
> > 
> > I feel it's possible to add the warning in the `use_alias` decorator, instead of creating a new decorator.
> 
> I updated the `use_alias` decorator. Is this the direction you had in mind @seisman?

Yes. Does it work as expected?

pygmt/helpers/decorators.py
```diff
@@ -574,6 +565,14 @@ def new_module(*args, **kwargs):
                         f"Parameters in short-form ({short_param}) and "
                         f"long-form ({long_alias}) can't coexist."
                     )
+                if (long_alias in kwargs and long_alias in ["xshift", "yshift"]) or (
```
seisman, line 568: I think it doesn't works, because after removing `X="xshift"` and `Y="yshift"` from the `use_alias` decorator, the variable `aliases` at line 571 no longer contain the paramters `xshift` or `yshift`.

Instead, after line 583, you can check if any of `xshift`, `yshift`, `X` and `Y` exists in `kwargs`. If yes, then raise an warning. For backward compatibility, you also need to do the "xshift"->X and "yshift"->"Y" conversions if "xshift" or "yshift" is used.

michaelgrund: So, you mean to leave `X="xshift" and Y="yshift"` as is in each function and add the others at this point, too @seisman ?
```
@use_alias(
    X="xshift",
    Y="yshift",
    xshift="X",
    yshift="Y",
)
```

seisman: > So, you mean to leave `X="xshift" and Y="yshift"` as is in each function and add the others at this point, too @seisman ?
> 
> ```
> @use_alias(
>     X="xshift",
>     Y="yshift",
>     xshift="X",
>     yshift="Y",
> )
> ```

No. Changes to other files (e.g., `basemap.py`) make sense to me. The only thing that is missing is how to _correctly_ check if users use `xshift`, `yshift`, `X` or `Y`. The checking should be done in the `pygmt/helpers/decorators.py` file.

pygmt/helpers/decorators.py
```diff
@@ -574,6 +565,14 @@ def new_module(*args, **kwargs):
...
+                    short_param in kwargs and short_param in ["X", "Y"]
+                ):
+                    raise GMTInvalidInput(
+                        f"Parameters ({short_param}) and "
+                        f"({long_alias}) are not supported anymore."
+                        f" Please use shift_origin() instead!"
+                    )
```
weiji14, line 575: Using an error here will immediately break people's code if they're using `xshift` or `yshift`. Better to go through a deprecation cycle that raises a `FutureWarning` instead.

One idea is to use the [`@deprecate_parameter`](https://github.com/GenericMappingTools/pygmt/blob/d80b986a3f5634e2e68d98d569405dfa3eba1475/pygmt/helpers/decorators.py#L732-L808) function and do something like `@deprecate_parameter(oldname="xshift", newname=None, deprecate_version="v0.8.0", remove_version="0.10.0")`. This will require some modifications to the decorator to handle `newname=None` though.

seisman: > One idea is to use the [`@deprecate_parameter`](https://github.com/GenericMappingTools/pygmt/blob/d80b986a3f5634e2e68d98d569405dfa3eba1475/pygmt/helpers/decorators.py#L732-L808) function and do something like `@deprecate_parameter(oldname="xshift", newname=None, deprecate_version="v0.8.0", remove_version="0.10.0")`. This will require some modifications to the decorator to handle `newname=None` though.

We don't have to use `@deprecate_parameter` for this, which requires adding `@deprecate_parameter` decorators in many Python files.

michaelgrund: I modified the passage @seisman. I understand the concerns @weiji14 has regarding raising an error here. However, we did agree to totally drop this option (not only deprecate it), didn't we?

seisman: > was there a discussion to drop xshift/yshift in a backward incompatible way?

No, we definitely need **backward compatibility**.

The final goal is, for the next 2 or more minor versions, users can still use parameters `xshift`, `yshift`, `X`, `Y`, and they will see a *FutureWarning* like:
```
Parameters xshift, yshift, X and Y are deprecated and will be no longer 
supported in v0.x.x. Please user Figure.shift_origin() instead.
```

> One idea is to use the [`@deprecate_parameter`](https://github.com/GenericMappingTools/pygmt/blob/d80b986a3f5634e2e68d98d569405dfa3eba1475/pygmt/helpers/decorators.py#L732-L808) function and do something like `@deprecate_parameter(oldname="xshift", newname=None, deprecate_version="v0.8.0", remove_version="0.10.0")`. This will require some modifications to the decorator to handle `newname=None` though.

I believe this will work but it's not ideal. It's easy to modify the `deprecate_parameter` decorator to support `newname=None`, but the warning message would be something like:
```
The 'xxxx' parameter has been deprecated since vx.x.x and will be removed in vx.x.x.
```
The error message can't recommend using `Figure.shift_origin()` unless we deal with `xshift` and `yshfit` in a special way.

It's also more confusing when users use the short parameters `X` or `Y`. When they use `X`, they are recommended to use `xshift`. After they modify `X` to `xshift`, then they see `xshfit` is deprecated.

seisman:
@michaelgrund There are a lot of conflicts after merging PR #1932 into the main branch. Please resolve these conflicts first.

pygmt/src/solar.py
```diff
@@ -56,10 +54,9 @@ def solar(self, terminator="d", terminator_datetime=None, **kwargs):
         Color or pattern for filling of terminators.
     pen : str
         Set pen attributes for lines. The default pen
-        is ``"0.25p,black,solid"``.
+        is ``default,black,solid``.
```
seisman, line 57:
```suggestion
        is ``"0.25p,black,solid"``.
```

pygmt/helpers/decorators.py
seisman: I've opened PR #2135, which shows how we can deprecate `timestamp` (`U`) in a backward-compatible and recommend `Figure.timestamp()` (not implemented yet) instead. The same idea can be applied to xshift and yshift in this PR.

michaelgrund: I applied your implemetations for this PR @seisman.

pygmt/helpers/decorators.py
```diff
@@ -575,6 +575,27 @@ def new_module(*args, **kwargs):
...
                     warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
+
+            # xshift (X) is deprecated since v0.8.0.
+            if "X" in kwargs or "xshift" in kwargs:
+                if "xshift" in kwargs:
+                    kwargs["X"] = kwargs.pop("xshift")
+                msg = (
+                    "Parameters 'X' and 'xshift' are deprecated since v0.8.0. "
+                    "Use Figure.shift_origin(xshift=...) instead."
+                )
```
seisman, line 586: Do we want to fully remove these parameters in future versions (e.g., v0.12.0)?
```suggestion
                msg = (
                    "Parameters 'X' and 'xshift' are deprecated since v0.8.0 "
                    "and will be removed in v0.12.0. "
                    "Use Figure.shift_origin(xshift=...) instead."
                )
```

seisman:
@michaelgrund I think this PR is almost ready for the final review. What do you think?

pygmt/helpers/decorators.py
weiji14: 4 minor versions sounds good.

michaelgrund:
> @michaelgrund I think this PR is almost ready for the final review. What do you think?

Yes it is!

seisman:
Some changes need to be made:

- [ ] Need to remove `xshift` and `yshift` from https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/src/ternary.py
- [ ] `xshift` and `yshift` are still used in https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/tests/test_config.py.
- [ ] Placeholder `xyshift` is no longer used and can be removed from COMMON_DOCSTRING: https://github.com/GenericMappingTools/pygmt/blob/6964fe13cd1463452457becf4221a80387b674f6/pygmt/helpers/decorators.py#L100
- [ ] Need to add a test to make sure that a warning is raised if users use xshift/X/yshift/Y

michaelgrund:
> Some changes need to be made:
> 
> * [x]  Need to remove `xshift` and `yshift` from https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/src/ternary.py
> * [x]  `xshift` and `yshift` are still used in https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/tests/test_config.py.
> * [x]  Placeholder `xyshift` is no longer used and can be removed from COMMON_DOCSTRING: https://github.com/GenericMappingTools/pygmt/blob/6964fe13cd1463452457becf4221a80387b674f6/pygmt/helpers/decorators.py#L100
> * [x]  Need to add a test to make sure that a warning is raised if users use xshift/X/yshift/Y

Added 4 tests covering each parameter. Do we need all 4 or is one sufficient @seisman ?

seisman:
> > Some changes need to be made:
> > 
> > * [x]   Need to remove `xshift` and `yshift` from [`main`/pygmt/src/ternary.py](https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/src/ternary.py?rgh-link-date=2022-11-16T20%3A33%3A26Z)
> > * [x]   `xshift` and `yshift` are still used in [`main`/pygmt/tests/test_config.py](https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/tests/test_config.py?rgh-link-date=2022-11-16T20%3A33%3A26Z).
> > * [x]   Placeholder `xyshift` is no longer used and can be removed from COMMON_DOCSTRING: https://github.com/GenericMappingTools/pygmt/blob/6964fe13cd1463452457becf4221a80387b674f6/pygmt/helpers/decorators.py#L100
> > * [x]   Need to add a test to make sure that a warning is raised if users use xshift/X/yshift/Y
> 
> Added 4 tests covering each parameter. Do we need all 4 or is one sufficient @seisman ?

I think we can merge the four tests into one larger test.

pygmt/tests/test_figure.py
```diff
@@ -274,3 +274,47 @@ def test_figure_icc_gray():
...
+
+def test_figure_depr_x():
+    """
+    Check if deprecation of X parameter works correctly if used.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
+    with pytest.warns(expected_warning=SyntaxWarning) as record:
+        fig.plot(x=1, y=1, style="c3c", X="3c")
+        assert len(record) == 1  # check that only one warning was raised
```
willschlitzer, line 298: Does the testing of `X` and `Y` needs to be included if `xshift` and `yshift` are already included, since we trust that the aliases translate to the appropriate arguments?

seisman: > Does the testing of `X` and `Y` needs to be included if `xshift` and `yshift` are already included, since we trust that the aliases translate to the appropriate arguments?

I think we still need to test `X` and `Y`, because we have removed the "X=>xshift" and "Y=>yshift" aliases in this PR.

pygmt/tests/test_figure.py
```diff
@@ -274,3 +274,47 @@ def test_figure_icc_gray():
...
+
+def test_figure_depr_y():
+    """
+    Check if deprecation of Y parameter works correctly if used.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
+    with pytest.warns(expected_warning=SyntaxWarning) as record:
+        fig.plot(x=1, y=1, style="c3c", Y="3c")
+        assert len(record) == 1  # check that only one warning was raised
```
seisman, line 320:
```suggestion
def test_figure_deprecated_xshift_yshift():
    """
    Check if deprecation of parameters X/Y/xshift/yshift work correctly if used.
    """
    fig = Figure()
    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
    with pytest.warns(expected_warning=SyntaxWarning) as record:
        fig.plot(x=1, y=1, style="c3c", xshift="3c")
        assert len(record) == 1  # check that only one warning was raised
    with pytest.warns(expected_warning=SyntaxWarning) as record:
        fig.plot(x=1, y=1, style="c3c", X="3c")
        assert len(record) == 1  # check that only one warning was raised
    with pytest.warns(expected_warning=SyntaxWarning) as record:
        fig.plot(x=1, y=1, style="c3c", yshift="3c")
        assert len(record) == 1  # check that only one warning was raised
    with pytest.warns(expected_warning=SyntaxWarning) as record:
        fig.plot(x=1, y=1, style="c3c", Y="3c")
        assert len(record) == 1  # check that only one warning was raised
```

seisman (approved):

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final reviews.

willschlitzer (approved):
Nice work @michaelgrund!

PR #2078, opened 2022-08-23
Figure.show: Allow keyword arguments passed to Figure.psconvert

### LINKED ISSUE #2038: Add "-G" flag passthrough to psconvert for finding the correct gs version
As discussed [here](https://forum.generic-mapping-tools.org/t/pygmt-finding-old-ghostscript-with-conda-on-windows/3174), if an old version of gs installed on Windows is referred to in the registry, psconvert will use the old gs instead of any newer versions that might be referred to in the path (psconvert checks the registry only and not the path). 

This leads to an error in pygmt.show() if the gs in the registry is too old; currently there is no way to point pygmt.show() to a newer version except by adding the path to the newer version into the registry.

However, pygmt.show_versions() is able to find a newer version of gs referred to in the path, for example if both have been installed through conda. A suggested workaround is to enable pygmt.show() and pygmt.savefig() to use the same method to determine that path to gs, and pass it on to psconvert via the -G flag.

seisman:

Add `**kwargs` to `Figure.show` so that it can pass any keyword arguments to `Figure.psconvert`.

`Figure.savefig()` already have `**kwargs`. 

Fixes #2038 

**Reminders**

**Slash Commands**


pygmt/figure.py
```diff
@@ -289,6 +289,9 @@ def savefig(
         dpi : int
             Set raster resolution in dpi. Default is 720 for PDF, 300 for
             others.
+        **kwargs : dict
+            Additional keyword arguments passed to
+            :meth:`pygmt.Figure.psconvert`.
```
seisman, line 294: Unsure if we should use `kwargs` or `**kwargs` here.

weiji14: I had a look at some projects and there seems to be a mix 😅. Let's go with `**kwargs` following https://docs.xarray.dev/en/v2022.06.0/generated/xarray.open_dataset.html, unless anyone else has a good reason on why not.

weiji14, line 294: I find it a bit confusing that `dpi` was previously an extra option or kwarg, but not there's both `dpi` and `**kwargs` in the `Figure.savefig(...)` method shown in the documentation:
image.

Not sure what's the best solution here. Removing the `dpi` parameter documentation doesn't seem ideal, nor does including all of `psconvert`'s parameters in `savefig`. Would it make sense then to put the `dpi` documentation under `**kwargs`? Or is there a cleaner solution?

seisman: I agree. Actually some of `psconvert`'s parameters (e.g., `prefix`, `fmt`) can **NOT** be used in `Figure.show` or `Figure.savefig`.

Perhaps we should say:
```
**kwargs: dict
    Additional keyword arguments passed to `pygmt.Figure.pscovert`. 
    Valid parameters are `gs_path`, `gs_option`, `resize`, `bb_style`, and `verbose`.
```

weiji14: Yes! Also need to do a similar thing for the fig.show docstring.

weiji14 (approved):

michaelgrund (approved):

PR #1755, opened 2022-02-13
Migrate pylint settings from .pylintrc to pyproject.toml

seisman:
**Description of proposed changes**

Changes in this PR:

- Remove the long pylint configuration file `.pylintrc`
- Migrate custom pylint settings (only five settings) in `pyproject.toml`
- Do not set global disables `import-outside-toplevel` and `attribute-defined-outside-init`
- Set `jobs` to 0 so that pylint can use all available cores (the old value is 4).

Fixes #1740.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2120, opened 2022-09-17
Remove unnecessary typecheck settings for pylint

seisman:

We added pylint settings `generated-members="pandas.*"` in https://github.com/GenericMappingTools/pygmt/pull/1755#discussion_r806882150 to silent annoying pylint messages. It seems recent versions of pylint no longer report these messages, so we can remove the settings.


weiji14 (approved):
Yep, can confirm that running `make lint` locally doesn't throw up any errors. Tested using `pylint=2.15.0` and `astroid=2.12.5`.

Issue #1082, opened 2021-03-20
What's the most Pythonic way for long GMT arguments?


seisman:
For each GMT option, PyGMT usually provides a long-form parameter. For example, GMT CLI option **-B** is equivalent to **frame** in PyGMT. This is done by the alias system (the `use_alias` decorator), which automatically converts long-form PyGMT parameters to short-form GMT options. The alias system works well but has known limitations and issues (e.g., #256, #262).

Sometimes, GMT arguments are still specified using long unreadable strings. For example,

- `projection="S0/90/12c"`: related to #356, #379
- `frame=["WSen", "xaf+lXLABEL", "yaf+lYLABEL"]`: related to #249 

As these common options are used a lot, we may take some time to find and implement a more Pythonic way to specify `projection`, `frame` et al.

However, there are still a lot of other long GMT arguments. Here is a simple example showing how to plot a GMT logo on maps. 
```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 2], projection="X6c", frame=["WSen", "xaf", "yaf"])
fig.logo(position="jTR+o0.3c/0.6c+w3c", box="+p1p+glightblue")
fig.show()
```
The above Python script is equivalent to the following GMT CLI script:
```bash
gmt begin
gmt basemap -R0/10/0/2 -JX6c -B
gmt logo -DjTR+o0.3c/0.6c+w3c -F+p1p+glightblue
gmt end show
``` 
The long argument for **position** and **box** are really difficult to remember and write. There are definitely some better and more Pythonic ways to specify these arguments. As GMT has too many options, we must provide a universal way to specify these long arguments. 

Perhaps the simplest way is to pass these long arguments as dict, i.e.,
```
fig.logo(
    position={
        location: "jTR",
        offset: [0.3, 0.6],
        width: "3c",
    },
    box={
        pen: "1p",
        fill: "lightblue"
    }
)
```
Are there any better ways?

weiji14:
This has come up again and again in different forms, so I've started a project at https://github.com/GenericMappingTools/pygmt/projects/6 to track the various issues and PRs tackling this. Let's make this issue is the central point to gather all the ideas. Here's my attempt at summarize the various candidates (feel free to edit this comment anyone, or suggest alternatives)

## Dictionary

Pass the arguments into the parameter as a Python dictionary. E.g. at https://github.com/GenericMappingTools/pygmt/issues/1082#issue-836710001

```python
[...]

PR #2130, opened 2022-09-21
Add Pythonic argument options for colorbar frame parameters

willschlitzer:
This adds the parameters `xlabel`, `ylabel`, and `annotation` to the `colorbar` module to allow those labels to be set without passing a list to `frame`. It will still allow other arguments to be passed to `frame`.

This is a proof of concept for moving towards more Pythonic arguments; I feel that `colorbar` is pretty straightforward to begin substituting new parameters, as I assume most of the arguments passed to `frame` affect either the axis labels or the annotation interval. It should still maintain backward compatibility for setting all of these values within `frame`.

Addresses #1082 

**Reminders**

**Slash Commands**


pygmt/src/colorbar.py
```diff
@@ -47,6 +47,12 @@ def colorbar(self, **kwargs):
     ----------
     frame : str or list
         Set color bar boundary frame, labels, and axes attributes.
+    annotation : int or float or str
+        Set the interval for annotated gridlines on the colorbar.
+    xlabel : str or int or float
+        Set the label for the x-axis of the colorbar.
+    ylabel : str or int or float
+        Set the label for the y-axis of the colorbar.
```
weiji14, line 55: Probably should just set the type as `str`. Yes they could pass in a number, but that's not very common.
```suggestion
    xlabel : str
        Set the label for the x-axis of the colorbar.
    ylabel : str
        Set the label for the y-axis of the colorbar.
```

pygmt/src/colorbar.py
```diff
@@ -105,6 +111,21 @@ def colorbar(self, **kwargs):
...
+        else:
+            frame = [kwargs.get("B")]
+        if xlabel:
+            frame.append("x+l" + str(xlabel))
+        if ylabel:
+            frame.append("y+l" + str(ylabel))
+        if annotation:
+            frame.append("a" + str(annotation))
+    if frame:
+        kwargs["B"] = frame
```
weiji14, line 128: This is a nice backward compatible way of handling old `frame/B` arguments while introducing new user-friendly parameters. I understand that putting the logic here in `colorbar` is a good proof of concept. Just want to ask though, what's your medium term plan for this chunk of if-then logic? Specifically, do you think we should duplicate this for every PyGMT module that does `frame/B`, or put it somewhere central (e.g. as a decorator under https://github.com/GenericMappingTools/pygmt/blob/v0.7.0/pygmt/helpers/decorators.py) or something else?

I don't want to resurface the discussion in #1082 on dictionary/functions/classes too much, since nothing would get done if just keep talking while not doing anything. But a colleague of mine recently mentioned that using the function based approach in PyGMT (i.e. passing in parameter-arguments like `xlabel="Some label"`) seemed like a good idea, so I can be swayed with going for this approach rather than the over-engineered class-based approach.

pygmt/tests/test_colorbar.py
```diff
@@ -65,3 +65,53 @@ def test_colorbar_shading_list():
...
+def test_colorbar_frame_string_parameters():
+    """
+    Create colorbar with a string passed to frame and arguments passed to
+    xlabel and ylabel.
+    """
+    fig = Figure()
+    fig.colorbar(
+        cmap="rainbow",
+        position="x0c/0c+w1c/0.5c",
+        frame="a.5",
```
weiji14, line 113: > I added that your suggestion and changed the parameters to this, which should be the same output as `test_colorbar_frame_parameters`:
> 
> ```
>     fig.colorbar(
>         cmap="rainbow",
>         position="x0c/0c+w1c/0.5c",
>         frame="a.25",
>         xlabel="test-x",
>         ylabel="test-y",
>     )
> ```
> 
> The test then fails; am I missing a step to make the figures match up?

Should be 0.25 to match `test_colorbar_frame_parameters`.

```suggestion
        frame="a0.25",
```

pygmt/src/colorbar.py
weiji14: > I specifically chose to work with `frame` in `colorbar` since there aren't a ton of edge cases to support, and it doesn't need to be consistent with other modules (as opposed to adding this logic for `frame` with something like `grdcontour` or `plot`). I think it can stay in `colorbar` for now, but I don't feel that strongly.

Agree to keep the logic in colorbar for now. The key part is the unit tests really (which tests what the users will be doing). The implementation, be it as a decorator or some if-then check here can change afterwards.

> I don't think I understand the function based approach method you mention; it looks to me like there would still be a parameter `xlabel` that has an argument passed to it. I'm definitely on board with other approaches; I just want to create proofs of concepts to get the ball rolling on more Pythonic arguments.

By function-based approach, I meant exactly what you're doing here - doing `fig.colorbar(..., xlabel="something")`. Contrast this with the class-based approach I mentioned in https://github.com/GenericMappingTools/pygmt/issues/1082#issuecomment-814040028 that would be like:

```python
frame = pygmt.param.Frame(xlabel="something")
fig.colorbar(..., frame=frame)
```

Function-based approach would be more user friendly (allows for tab-completion). Class-based approach would be more developer friendly (less coding for us).

pygmt/tests/test_colorbar.py
```diff
@@ -65,3 +65,53 @@ def test_colorbar_shading_list():
...
+        cmap="rainbow",
+        position="x0c/0c+w1c/0.5c",
+        frame=["a.25", "y+lylabel"],
+        xlabel="test-x",
+    )
+    return fig
+
+
+@pytest.mark.mpl_image_compare
+def test_colorbar_frame_string_parameters():
```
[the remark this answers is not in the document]
willschlitzer: Getting this test failure when I have those changes:

```
Error: Image dimensions did not match.
  Expected shape: (175, 237)
  Actual shape: (175, 260)
```
I'm not sure why they should produce differently sized figures.

pygmt/tests/baseline/test_colorbar_frame_string_parameters.png.dvc
```diff
@@ -0,0 +1,4 @@
+outs:
+- md5: 7a8bbd25a40214160f00a5c5bf83f69d
+  size: 3764
+  path: test_colorbar_frame_string_parameters.png
```
weiji14, line 4: Could you delete this file and see if it fixes the test failure? Maybe it's still comparing the old image?

willschlitzer: Still no luck, tried deleting both the .dvc file and the image.

pygmt/tests/test_colorbar.py
```diff
@@ -65,3 +65,53 @@ def test_colorbar_shading_list():
...
+    """
+    Create colorbar with a string passed to frame and arguments passed to
+    xlabel and ylabel.
+    """
+    fig = Figure()
+    fig.colorbar(
+        cmap="rainbow",
+        position="x0c/0c+w1c/0.5c",
+        frame="a0.5",
```
weiji14, line 113:
```suggestion
        frame="a0.25",
```

willschlitzer:
@weiji14 I'm not sure why I'm not able to get `test_colorbar_frame_parameters` and `test_colorbar_frame_string_parameters` to match up. I'm going to leave it as two separate reference images.

willschlitzer:
Don't think this has much traction, so I'm going to close the issue.

PR #2139, opened 2022-09-25
Add L alias for ternary.py

michaelgrund:

This PR adds an alias for the missing -L flag in ternary. 

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/ternary.py
+++ b/pygmt/src/ternary.py
@@ -11,6 +11,7 @@ from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, us
     C="cmap",
     G="fill",
     JX="width",
+    L="label",
     R="region",
     S="style",
     U="timestamp",
@@ -53,6 +54,11 @@ def ternary(self, data, **kwargs):
         Give the min and max limits for each of the three axes **a**, **b**,
         and **c**.
     {cmap}
+    label : str
+        *a*/*b*/*c*. Set the labels for the three diagram vertices where the 
+        component is 100% [none]. These are placed at a distance of three 
+        times the MAP_LABEL_OFFSET setting from their respective corners. 
+        To skip any one of them, specify that label as -.
     {fill}
     style : str
         *symbol*\[\ *size*].


pygmt/src/ternary.py
```diff
@@ -11,6 +11,7 @@
     C="cmap",
     G="fill",
     JX="width",
+    L="label",
```
seisman, line 14: `label` is a bad name, because it's already an alias for `-l` option.

I'm wondering if we should use a single parameter `labels=["a", "b", "c"]` or three parameters `alabel="a", blabel="b", clabel="c"`.

michaelgrund: I would like the idea of having three parameters to directly adjust the desired vertices without considering a potential order when using `labels=["a", "b", "c"]`.

seisman: I also feel that three separate parameters are more Pythonic, although it needs more coding efforts.

michaelgrund: Great, I added 3 separate parameters. Do we need a unit test for this change although it's a quite simple implementation @GenericMappingTools/pygmt-maintainers ?

pygmt/src/ternary.py
```diff
@@ -54,6 +54,14 @@ def ternary(self, data, **kwargs):
         and **c**.
     {cmap}
     {fill}
+    alabel : str
+        Set the label for the *a* vertex where the component is 100%. The
+        label is placed at a distance of three times the MAP_LABEL_OFFSET
+        setting from the corner.
```
seisman, line 60:
```suggestion
        label is placed at a distance of three times the 
        :gmt-term:`MAP_LABEL_OFFSET` setting from the corner.
```

pygmt/src/ternary.py
```diff
@@ -68,6 +76,18 @@ def ternary(self, data, **kwargs):
...
+        if alabel is None:
+            alabel = "-"
+
+        if blabel is None:
+            blabel = "-"
+
+        if clabel is None:
+            clabel = "-"
+
+        kwargs["L"] = "/".join([alabel, blabel, clabel])
```
seisman, line 89: It's better to put these lines out of the `with` statement.

pygmt/src/ternary.py
```diff
@@ -66,8 +74,21 @@ def ternary(self, data, **kwargs):
...
+    if alabel is None:
+        alabel = "-"
+
+    if blabel is None:
+        blabel = "-"
+
+    if clabel is None:
+        clabel = "-"
+
+    kwargs["L"] = "/".join([alabel, blabel, clabel])
```
seisman, line 87: `alabel` should be in `str` type, but users may pass other invalid types (e.g., `int`).
Do you think we should rewrite the code to:
```
alable = str(alabel) if alable is not None else "-"
```

michaelgrund: Good point! Also makes the code a little bit shorter 😉

pygmt/src/ternary.py
```diff
@@ -66,8 +74,16 @@ def ternary(self, data, **kwargs):
...
+
+    alabel = str(alabel) if alabel is not None else "-"
+    blabel = str(blabel) if blabel is not None else "-"
+    clabel = str(clabel) if clabel is not None else "-"
+
+    kwargs["L"] = "/".join([alabel, blabel, clabel])
+
     with Session() as lib:
         file_context = lib.virtualfile_from_data(check_kind="vector", data=data)
+
```
seisman, line 86:
```suggestion
```

pygmt/src/ternary.py
```diff
@@ -66,8 +74,16 @@ def ternary(self, data, **kwargs):
     {transparency}
     """
     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+
+    alabel = str(alabel) if alabel is not None else "-"
+    blabel = str(blabel) if blabel is not None else "-"
+    clabel = str(clabel) if clabel is not None else "-"
+
```
seisman, line 81:
```suggestion
```

seisman (approved):
Just removed two blank lines then this PR looks good to me.

pygmt/src/ternary.py
weiji14: Yes, I think a unit test or two would be good. Could you write a test for when all of alabel/blabel/clabel (3/3) are used, and one when only one of them (1/3) is used?

pygmt/src/ternary.py
```diff
@@ -66,6 +74,12 @@ def ternary(self, data, **kwargs):
     {transparency}
     """
     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+
+    alabel = str(alabel) if alabel is not None else "-"
+    blabel = str(blabel) if blabel is not None else "-"
+    clabel = str(clabel) if clabel is not None else "-"
+    kwargs["L"] = "/".join([alabel, blabel, clabel])
```
weiji14, line 81: If all of alabel/blabel/clabel is None, would that mean an argument like `-L-/-/-` is still passed to GMT? Probably gives the same plot, but thinking whether a if-then statement that checks if any of alabel/blabel/clabel is not None would be good.

michaelgrund: An additional if statement or replace the currently used ones @weiji14 ?

pygmt/src/ternary.py
michaelgrund: Done!

pygmt/src/ternary.py
willschlitzer: I assume this would be a new if statement, something like:
```
if alabel or blabel or clabel:
    alabel = str(alabel) if alabel is not None else "-"
    blabel = str(blabel) if blabel is not None else "-"
    clabel = str(clabel) if clabel is not None else "-"
    kwargs["L"] = "/".join([alabel, blabel, clabel])
```

weiji14 (commented):
Careful now, need to delete the PNG images commited to the `baseline/` folder.

michaelgrund:
> Careful now, need to delete the PNG images commited to the `baseline/` folder.

Done!

weiji14 (approved):
Nice, thanks @michaelgrund!

PR #2159, opened 2022-10-14
SVG format is not supported in psconvert with ghostscript 9.16+

seisman:

SVG format is not supported in recent gs versions, as explained in the GMT documentation (https://docs.generic-mapping-tools.org/dev/psconvert.html#notes):

> The ghostscript program continues to be developed and occasionally its developers make decisions that affect psconvert. As of version 9.16 the SVG device has been removed.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -145,7 +145,7 @@ class Figure:
         Convert [E]PS file(s) to other formats.
 
         Converts one or more PostScript files to other formats (BMP, EPS, JPEG,
-        PDF, PNG, PPM, SVG, TIFF) using Ghostscript.
+        PDF, PNG, PPM, TIFF) using Ghostscript.
 
         If no input files are given, will convert the current active figure
         (see :class:`pygmt.Figure`). In this case, an output name must be given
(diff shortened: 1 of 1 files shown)


willschlitzer (approved):

pygmt/figure.py
```diff
@@ -216,7 +216,7 @@ def psconvert(self, icc_gray=False, **kwargs):
             **E** means EPS with PageSize command, **f** means PDF, **F** means
             multi-page PDF, **j** means JPEG, **g** means PNG, **G** means
             transparent PNG (untouched regions are transparent), **m** means
-            PPM, **s** means SVG, and **t** means TIFF [Default is JPEG]. To
+            PPM, and **t** means TIFF [Default is JPEG]. To
```
weiji14, line 219: Should we add a link to https://docs.generic-mapping-tools.org/6.4/psconvert.html#notes and/or state in this docstring somewhere that "Fortunately, quality SVG graphics can be obtained by first converting to PDF and then install and use the package pdf2svg."?

weiji14 (commented):
Was going to comment about updating `savefig`, but it seems like SVG was only supported in `psconvert` is it?

pygmt/figure.py
seisman: I don't think we should mention SVG, because gs 9.16 was released several years ago, so SVG is never supported in PyGMT since 2017. People who want SVG format should find tools to do the conversion themselves.

weiji14: Ah right, we're at ghostscript 9.54 or something now. Then just removing the svg mention is fine.

weiji14 (approved):

Issue #2092, opened 2022-08-30
Figure.savefig() reports confusing errors when saving image into a non-existent directory


seisman:
**Description of the problem**

`Figure.savefig()` reports errors when saving an image to a subdirectory that doesn't exist. The error messages are very confusing. 

I think we should check if the parent directory of the image exists before saving an image. If the directory doesn't exist, we have two options:

1. Report that parent directories are missing and users should create these directories manually
2. Create any missing parent directories automatically before saving 

**Full code that generated the error**

```python
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[101, 108.5, 26, 29], projection="M15c", frame=True)
fig.savefig("anypath/map.png")
```

**Full error message**

```
GPL Ghostscript 9.55.0: **** Could not open the file 'anypath/map.png'.
Error: /invalidfileaccess in --showpage--
Operand stack:
   1   true
Execution stack:
   %interp_exit   .runexec2   --nostringval--   showpage   --nostringval--   2   %stopped_push   --nostringval--   showpage   showpage   false   1   %stopped_push   1974   1   3   %oparray_pop   1973   1   3   %oparray_pop   1961   1   3   %oparray_pop   1817   1   3   %oparray_pop   --nostringval--   %errorexec_pop   .runexec2   --nostringval--   showpage   --nostringval--   2   %stopped_push   --nostringval--   1824   0   4   %oparray_pop   showpage   showpage
Dictionary stack:
   --dict:776/1123(ro)(G)--   --dict:0/20(G)--   --dict:75/200(L)--   --dict:158/250(L)--
Current allocation mode is local
Last OS error: No such file or directory
Current file position is 24898
GPL Ghostscript 9.55.0: Unrecoverable error, exit code 1
psconvert [ERROR]: System call [gs -q -dNOPAUSE -dBATCH -dNOSAFER -dSCANCONVERTERTYPE=2 -dMaxBitmap=2147483647 -dUseFastColor=true -dGraphicsAlphaBits=2 -dTextAlphaBits=2 -sDEVICE=png16m  -g1904x872 -r300 -sOutputFile='anypath/map.png' '/home/seisman/.gmt/sessions/gmt_session.90482/psconvert_90482d.eps'] returned error 256.
---------------------------------------------------------------------------
GMTCLibError                              Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 fig.savefig("anypath/map.png")

[...]

PR #2160, opened 2022-10-14
Figure.savefig: Raise a FileNotFoundError if the parent directory doesn't exist

seisman:

Fixes https://github.com/GenericMappingTools/pygmt/issues/2092.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -244,8 +244,9 @@ class Figure:
 
         # check if the parent directory exists
         if kwargs.get("F") and not Path(kwargs.get("F")).parent.exists():
+            directory = Path(kwargs.get("F")).parent
             raise FileNotFoundError(
-                f"No such directory: '{Path(kwargs.get('F')).parent}'"
+                f"No such directory: '{directory}', please create it first."
             )
 
         # Manually handle prefix -F argument so spaces aren't converted to \040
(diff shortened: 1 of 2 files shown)


pygmt/figure.py
```diff
@@ -241,6 +242,12 @@ def psconvert(self, icc_gray=False, **kwargs):
             else:
                 kwargs["N"] += "+i"
 
+        # check if the parent directory exists
+        if kwargs.get("F") and not Path(kwargs.get("F")).parent.exists():
+            raise FileNotFoundError(
+                f"No such directory: '{Path(kwargs.get('F')).parent}'"
```
weiji14, line 248:
```suggestion
            directory = Path(kwargs.get("F")).parent
            raise FileNotFoundError(
                f"No such directory: '{directory}', please "
                f"create it first, e.g. using `os.makedirs(name='{directory}')`"
```

seisman: Here is the error message from matplotlib:
```python
FileNotFoundError: [Errno 2] No such file or directory: 'anypathh/image.png'
```
I think we can follow mpl and don't need to say how to create the directory.

weiji14: Prefer having an error message that not only lets the user know what is the issue (missing a directory) but also how to resolve it (create the directory). We could shorten the message though like so:
```suggestion
            directory = Path(kwargs.get("F")).parent
            raise FileNotFoundError(
                f"No such directory: '{directory}', please create it first."
```
and assume that the user knows how to create the folder using `os.makedirs` or just manually in their file manager.

weiji14 (approved):

Issue #2016, opened 2022-07-19
beachball offsetting doesn't work for pandas/dict/ndarray input


MsDannie76:
I am trying to plot focal mechanisms on a map with an offset. I’ve tried to follow the tips given here:
[ [PyGMT: How to offset focal mechanism beachballs?](https://forum.generic-mapping-tools.org/t/pygmt-how-to-offset-focal-mechanism-beachballs/1846)]
but still cannot get it to work, the two beachballs in the southeast are still plotted on top of each other, although there is no error message.
![FM_reg1_ex](https://user-images.githubusercontent.com/29919447/179743527-ae39fee9-2f69-475a-a438-77dad73213d2.png)

My file contains the columns
longitude latitude depth strike dip rake magnitude plot_longitude plot_latitude
and columns are separated by tabs:

|longitude|latitude|depth|strike|dip|rake|magnitude|plot_longitude|plot_latitude|
|1.4421|51.2819|15.5|47.28|32.05|112.09|4.5|1.4421|52|
|1.3884|51.3337|17|69.84|25.88|151.52|4.5|0.75|51.3337|
|-0.055|53.648|19.3|21.33|84.92|-19.37|4.1|-0.055|53.648|
|-1.314|54.594|4.6|102.1|84.1|-162|3.1|-1.314|54.594|
[FM_ISC_exampleplot.txt](https://github.com/GenericMappingTools/pygmt/files/9140749/FM_ISC_exampleplot.txt)

I reported that issue on the GMT community forum
[[Pygmt psmeca beachball - no offset despite following tips from previous post](https://forum.generic-mapping-tools.org/t/pygmt-psmeca-beachball-no-offset-despite-following-tips-from-previous-post/3132)]

**Full code that generated the error**

```python

import pygmt
import pandas as pd

FM_ISCex=pd.read_csv("FM_ISC_exampleplot.txt",delimiter="\t")

fig = pygmt.Figure()
fig.coast(
region="-2/2/51/55",
projection="M15c",
land="lightgray",
water="white",
borders="1/0.5p",
shorelines="1/0.5p",
frame="ag",
)

fig.meca(
spec=FM_ISCex,
convention="aki",
scale="2c",
G="red",
offset=True
)

fig.show()
```

**Full error message**

```
No error message, but offset given as plot_longitude and plot_latitude is not used during plotting.
```

**System information**

Please paste the output of `python -c "import pygmt; pygmt.show_versions()"`:

```
[...]

PR #1784, opened 2022-03-04
Rewrite the meca function to support offsetting and labeling beachballs

seisman:
**Description of proposed changes**

In this PR, the core part of the `Figure.meca()` method is fully rewritten, because the original `Figure.meca()` implementation is too messy to read. The new implementation takes some inspiration from the original implementation and the changes in PR #1613, especially this comment about passing `pandas.DataFrame` directly https://github.com/GenericMappingTools/pygmt/pull/1613#discussion_r744349790.

**Main changes in the new implementation:**

- Use the universal `virtualfile_from_data` function to pass data (address #949)
- Pass `pandas.DataFrame` to the `virtualfile_from_data` directly, because `pandas.DataFrame` allows columns with different data types (proposed in https://github.com/GenericMappingTools/pygmt/pull/1613#discussion_r744349790 and works well)
- Support offsetting beachballs with the `plot_longitude` and `plot_latitude` columns/parameters (fixes #1129)
- Support labeling beachballs with the `event_name` column/parameter (address https://github.com/GenericMappingTools/pygmt/pull/516#issuecomment-657313059). [Here I use `event_name` rather than `label`, because `label` is usually aliased to GMT's `-l` option, and there is a plan to add `-l` to `meca` xref https://github.com/GenericMappingTools/gmt/issues/1973)
- Fix the inaccruate parameter names as reported in https://github.com/GenericMappingTools/pygmt/issues/1612

**BREAKING CHANGES**

1. Current fix for #1612 introduces a breaking change, but it's possible to make it backward-compatible. If backward-compatible is required, I can do it in a separate PR, to make this PR small.
2. When parameters like `longitude` are passed as the function parameter **and** are also given in dict/pd.DataFrame (e.g., `spec["longitude"]`), the behavior changes. In the old implementation, values in ``spec`` has high priority; but in the new implementation, values explicitly given as the function parameter override the values in ``spec``. I think the new implementation is more intuitive. The changes are backward-incompatible. 

**TODO in future PRs:**

There are some other `meca` related issues. I think it would be better to address them in separate PRs because the PR is already too large:

- [ ] #1466
- [ ] more checks (for example, checking `longitude` and `latitude` have the same length)
- [ ] More tests

PR #2202, opened 2022-11-22
Figure.meca: Fix beachball offsetting with dict/pandas inputs

seisman:

Address https://github.com/GenericMappingTools/pygmt/issues/2016.

Due to how the GMT API works, the "plot_longitude", "plot_latitude", "event_name" columns must be in string type before passing to the GMT API.

The original code doesn't convert the columns to string type when these parameters are stored in dict/pandas inputs.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -189,11 +189,11 @@ def meca(
         Depth(s) of event location in kilometers. Must be the same length as
         the number of events. Will override the ``depth`` values in ``spec``
         if ``spec`` is a dict or pd.DataFrame.
-    plot_longitude: int, float, list, or 1d numpy array
+    plot_longitude: int, float, str, list, or 1d numpy array
         Longitude(s) at which to place beachball. Must be the same length as
         the number of events. Will override the ``plot_longitude`` values in
         ``spec`` if ``spec`` is a dict or pd.DataFrame.
-    plot_latitude: int, float, list, or 1d numpy array
+    plot_latitude: int, float, str, list, or 1d numpy array
         Latitude(s) at which to place beachball. List must be the same length
         as the number of events. Will override the ``plot_latitude`` values in
         ``spec`` if ``spec`` is a dict or pd.DataFrame.
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -272,9 +272,9 @@ def meca(
             spec["latitude"] = np.atleast_1d(latitude)
         if depth is not None:
             spec["depth"] = np.atleast_1d(depth)
-        if plot_longitude is not None:  # must be in string type
+        if plot_longitude is not None:
             spec["plot_longitude"] = np.atleast_1d(plot_longitude)
-        if plot_latitude is not None:  # must be in string type
+        if plot_latitude is not None:
             spec["plot_latitude"] = np.atleast_1d(plot_latitude)
         if event_name is not None:
             spec["event_name"] = np.atleast_1d(event_name).astype(str)
--- a/pygmt/src/meca.py
+++ b/pygmt/src/meca.py
@@ -273,9 +273,9 @@ def meca(
         if depth is not None:
             spec["depth"] = np.atleast_1d(depth)
         if plot_longitude is not None:  # must be in string type
-            spec["plot_longitude"] = np.atleast_1d(plot_longitude).astype(str)
+            spec["plot_longitude"] = np.atleast_1d(plot_longitude)
         if plot_latitude is not None:  # must be in string type
-            spec["plot_latitude"] = np.atleast_1d(plot_latitude).astype(str)
+            spec["plot_latitude"] = np.atleast_1d(plot_latitude)
         if event_name is not None:
             spec["event_name"] = np.atleast_1d(event_name).astype(str)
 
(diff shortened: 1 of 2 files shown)


weiji14:
> Due to how the GMT API works, the "plot_longitude", "plot_latitude", "event_name" columns must be in string type before passing to the GMT API.

Which line is it in https://github.com/GenericMappingTools/gmt/blob/6.4.0/src/seis/pscoupe.c?

seisman:
> > Due to how the GMT API works, the "plot_longitude", "plot_latitude", "event_name" columns must be in string type before passing to the GMT API.
> 
> Which line is it in [GenericMappingTools/gmt@`6.4.0`/src/seis/pscoupe.c](https://github.com/GenericMappingTools/gmt/blob/6.4.0/src/seis/pscoupe.c?rgh-link-date=2022-11-22T21%3A03%3A40Z)?

https://github.com/GenericMappingTools/gmt/blob/43bccb384663d97ee84f8110e4c50858483ec80f/src/seis/psmeca.c#L836

psmeca and pscoupe read "newX", "newY" and "title" from `S->text[row]`, which is the trailing text after numeric columns.

pygmt/src/meca.py
```diff
@@ -273,9 +273,9 @@ def meca(
         if depth is not None:
             spec["depth"] = np.atleast_1d(depth)
         if plot_longitude is not None:  # must be in string type
```
weiji14, line 275: Remove the comment here?
```suggestion
        if plot_longitude is not None:
```

pygmt/src/meca.py
```diff
@@ -273,9 +273,9 @@ def meca(
...
-            spec["plot_longitude"] = np.atleast_1d(plot_longitude).astype(str)
+            spec["plot_longitude"] = np.atleast_1d(plot_longitude)
         if plot_latitude is not None:  # must be in string type
```
weiji14, line 277:
```suggestion
        if plot_latitude is not None:
```

pygmt/src/meca.py
```diff
@@ -293,9 +293,13 @@ def meca(
...
+            spec[["plot_longitude", "plot_latitude"]] = spec[
+                ["plot_longitude", "plot_latitude"]
+            ].astype(str)
             kwargs["A"] = True
         if "event_name" in spec.columns:
             newcols += ["event_name"]
+            spec["event_name"] = spec["event_name"].astype(str)
         # reorder columns in DataFrame
         spec = spec.reindex(newcols, axis=1)
     elif isinstance(spec, np.ndarray) and spec.ndim == 1:
```
weiji14, line 305: Hmm, I was re-reading #1129 again and was gonna say, it's a little surprising that a 2D numpy array passed in with `plot_longitude` and `plot_latitude` columns (stored in float) works (since the `test_meca_spec_2d_array` unit test is passing fine). I guess the GMT C code is smart enough with handling data from `virtualfile_from_matrix` vs `virtualfile_from_vectors`.

Nothing to do here, just a random comment.

pygmt/src/meca.py
```diff
@@ -293,9 +293,13 @@ def meca(
         newcols = ["longitude", "latitude", "depth"] + param_conventions[convention]
         if "plot_longitude" in spec.columns and "plot_latitude" in spec.columns:
             newcols += ["plot_longitude", "plot_latitude"]
+            spec[["plot_longitude", "plot_latitude"]] = spec[
+                ["plot_longitude", "plot_latitude"]
+            ].astype(str)
```
weiji14, line 298: Should we add `str` to the docstring at L192 and L196? Also thinking about DMS coordinates like `23:30:30E` per #647.

pygmt/src/meca.py
seisman: > it's a little surprising that a 2D numpy array passed in with `plot_longitude` and `plot_latitude` columns (stored in float) works (since the `test_meca_spec_2d_array` unit test is passing fine).

That's a good point. I think it shouldn't work. The `test_meca_spec_2d_array` test passes because both `plot_longitude` and `plot_latitude` are 0 (which means no offsetting due to backward compatibility issues).

I'll find some time and see if offsetting using a 2D numpy array works, and if not, will  try to find a fix.

weiji14: Ah yes, so we should probably update the test to use a non-zero offset. I actually wanted to parse all the user inputs into a `pandas.DataFrame` (which can hold different dtypes like int, str, float) in https://github.com/GenericMappingTools/pygmt/pull/1613#discussion_r744349790 which meant everything goes through `virtualfile_from_vectors`, but that was superseded by #1784 that maintained two separate paths going via `virtualfile_from_vectors` for `dict`/`pd.DataFrame` or `virtualfile_from_matrix` for 2D `np.ndarray` inputs. Just as an idea :slightly_smiling_face:

If this ends up getting too complicated, we can merge in this PR first to fix offsets for dict/pandas inputs, and have a separate one to handle offsets for 2D numpy array inputs.

pygmt/src/meca.py
seisman: Done in 42083088.

pygmt/src/meca.py
seisman: > If this ends up getting too complicated, we can merge in this PR first to fix offsets for dict/pandas inputs, and have a separate one to handle offsets for 2D numpy array inputs.

I prefer to work on it in a separate PR.

weiji14 (approved):
Ok to merge, but don't close #2016 yet since offsets aren't working properly for 2d numpy array inputs.

PR #2135, opened 2022-09-24
Recommend Figure.timestamp and remove timestamp (U) alias from all plotting methods

seisman:
**Description of proposed changes**

Similar to https://github.com/GenericMappingTools/pygmt/pull/2071.

The purpose of this PR is to remove timestamp (U) alias and recommend use `Figure.timestamp` instead.

**NOTE**: `Figure.timestamp` is not implemented yet.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2208, opened 2022-11-24
Add Figure.timestamp to plot the GMT timestamp logo

seisman:

In this PR, I add the `Figure.timestamp` method to plot the GMT timestamp logo, just like what GMT's `-U` option does. The final goal is to remove the `timestamp` alias from all the plotting methods (c.f. https://github.com/GenericMappingTools/pygmt/pull/2135 and https://github.com/GenericMappingTools/pygmt/issues/924#issuecomment-780963272).

Here are some notes:

1. The full syntax of the `-U` option is: `-U[<label>][+c][+j<just>][+o<dx>[/<dy>]]`. `-U+c` can be used to append the raw GMT command (e.g., `gmt coast -R... -J...`) after the timestamp logo. It's not implemented because (1) It's rarely used. A figure is usually produced by multiple commands, so it makes no sense to show the command string of a single command; (2) It makes little sense to show the raw GMT command string for a figure produced by a Python script;  (3) It's also impossible to implement it in the `Figure.timestamp()` function.
2. The `Figure.timestamp` method is implemented by calling `gmt plot -T -U...`
3. There are four GMT defaults (`FONT_LOGO`, `MAP_LOGO`, `MAP_LOGO_POS`, and `FORMAT_TIME_STAMP`) that can affect the timestamp logo. 
   - `MAP_LOGO` controls if a timestamp logo should be plotted. It makes no sense here, so ignored.
   - `MAP_LOGO_POS` defaults to `BL/-54p/-54p`. It's equivalent to setting the `+j` and `+o` modifiers. So it's ignored.
   - `FONT_LOGO` can control the font ID of the text strings. It's implemented by passing `--FONT_LOGO=xxx` to the GMT C API.
   - `FORMAT_TIME_STAMP` can control the format of the timestamp. It's implemented by passing `--FORMAT_TIME_STAMP` to the GMT C API.
4. An upstream bug of the `+o` modifier for GMT <=6.4.0 (https://github.com/GenericMappingTools/gmt/issues/7107).
5. For GMT<=6.4.0, `FONT_LOGO` can only change the font ID, but since GMT 6.5.0, it can also change both the font ID and color (but not the font size). See https://github.com/GenericMappingTools/gmt/issues/7125
6. In GMT 6.5.0, a new modifier `+t` will be added. See https://github.com/GenericMappingTools/gmt/issues/7125.
7. It's difficult to test this function, because the image has the timestamp string, which changes every second. The workaround here is passing a constant format string to `FORMAT_TIME_STAMP` (e.g., `--FORMAT_TIME_STAMP="9999-99-99T99:99:99"`).

**Examples:**
```
>>> # Plot the GMT timestamp logo.
>>> import pygmt
>>> fig = pygmt.Figure()
>>> fig.timestamp()
>>> fig.show()

>>> # Plot the GMT timestamp logo with a custom label.
>>> fig = pygmt.Figure()
>>> fig.timestamp(label="Powered by PyGMT")
>>> fig.show()
```

**Documentation preview:** https://pygmt-dev--2208.org.readthedocs.build/en/2208/api/generated/pygmt.Figure.timestamp.html#pygmt.Figure.timestamp

**Reminders**

**Slash Commands**


pygmt/src/timestamp.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
+    kwargs["U"] = ""
+    if label:
+        kwargs["U"] = label
+    if justification:
+        kwargs["U"] += f"+j{justification}"
+    if offsets:
+        if "/" not in offsets:  # only offset_x is given
+            kwargs["U"] += f"+o{offsets}/{offsets}"  # upstream bug
```
seisman, line 22:
```suggestion
            # Workaround for an upstream bug which will be fixed in GMT 6.5.0
            # See https://github.com/GenericMappingTools/gmt/issues/7107
            kwargs["U"] += f"+o{offsets}/{offsets}"
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,70 @@
...
+    kwdict = {}
+    kwdict["T"] = True  # pass the -T option to the plot module
+    kwdict["U"] = True
+
+    # build the argument for the -U option
+    label = "" if label is None else f"{label}"
+    kwdict["U"] = f"{label}+j{justification}"
+
+    if is_nonstr_iter(offset):  # given a list
+        kwdict["U"] += "+o" + "/".join(f"{item}" for item in offset)
```
seisman, line 57: Please note that we can't use decorator `@kwargs_to_strings(offset="sequence")` here, because the decorator can only deal with keyword arguments and doesn't work for the default values. I.e., it doesn't work for `Figure.timestamp()` but works for `Figure.timestamp(offset=["-54p", "-54p"])`, although `offset` defaults to `["-54p", "-54p"]`.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for comments on the current implementation of the `Figure.timestamp()` function and ideas about how to test it.

seisman:
The [libfaketime](https://github.com/wolfcw/libfaketime) package can be used to temporarily change the system date/time. It's also available on [conda-forge](https://anaconda.org/conda-forge/libfaketime), but it's a CLI program.

There are Python wrapper (https://github.com/simon-weber/python-libfaketime) and pytest plugin (https://github.com/touilleMan/pytest-libfaketime), but both are not actively maintained.

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,100 @@
...
+    if text is not None:
+        if Version(gmt_version) <= Version("6.4.0"):
+            raise GMTInvalidInput("The parameter 'text' requires GMT>=6.5.0.")
+        kwdict["U"] += f"+t{text}"
+
+    with Session() as lib:
+        lib.call_module(
+            module="plot",
+            args=build_arg_string(kwdict)
+            + f" --FONT_LOGO={font} --FORMAT_TIME_STAMP={timefmt}",
```
seisman, line 99: Two thoughts when implementing this function:

1. Add a new parameter (e.g., `confdict`) to the `build_arg_string` function, so that we can pass a dict of GMT defaults to the `call_module` function, e.g.,
   ```
   build_arg_string(kwdict, confdict={"FONT_LOGO": font, "FORMAT_TIME_STAMP": timefmt}
   ```

2. The four parameters `FONT_LOGO`, `MAP_LOGO`, `MAP_LOGO_POS`, and `FORMAT_TIME_STAMP` won't be used after the `Figure.timestamp` function is used, so they can be removed from the `pygmt.config` keyword list:
 https://github.com/GenericMappingTools/pygmt/blob/9253af320ffbe90cc675acba33695a0325bea600/pygmt/src/config.py#L26-L28

weiji14:
> 7\. It's difficult to write tests for this module, because the image has the timestamp string, which changes every second. We can't test it unless we can find a way to set the current time to a fixed value.

Maybe use `@check_figures_equal` following https://github.com/GenericMappingTools/pygmt/blob/v0.8.0/doc/contributing.md#using-check_figures_equal, and output only the date and not the precise time down to the second?

pygmt/src/timestamp.py
weiji14: Hmm, point 1 seems ok, but I'm not sure about point 2. I'd prefer to keep the three parameters in the list, because even if you remove it, you're only removing the autocompletion, but users can still manually set those parameters.

pygmt/src/timestamp.py
seisman: Related issue report at https://github.com/GenericMappingTools/pygmt/issues/2361.

pygmt/src/timestamp.py
seisman: Point 1 was implemented in PR #2324

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    r"""
+    Plot the GMT timestamp logo.
+
+    Parameters
+    ----------
+    text : None or str
+        If ``None``, the current UNIX time stamp is shown in the GMT timestamp
+        logo. Setting this parameter to replace the UNIX time stamp with a
+        custom text string instead. The text must be less than 64 characters.
+        *Requires GMT>=6.5.0*.
```
seisman, line 30: I'm still unsure what's the best way to document features that require new GMT versions.

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    label : None or str
+        The text string shown after the GMT timestamp logo.
```
yvonnefroehlich, line 32:
```suggestion
        The text string shown after the GMT time stamp logo.
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    justification : str
+        Justification of the timestamp. *justification* is a two-character code
```
yvonnefroehlich, line 34:
```suggestion
        Justification of the time stamp. *justification* is a two-character code
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+        that is a combination of a horizontal (**L**\ (eft), **C**\ (enter), or
+        **R**\ (ight)) and a vertical (**T**\ (op), **M**\ (iddle), or
+        **B**\ (ottom)) code.
+    offset : str or tuple
+        *offset* or (*offset_x*, *offset_y*).
+        Offset the anchor point of the timestamp by *offset_x* and *offset_y*.
```
yvonnefroehlich, line 40:
```suggestion
        Offset the anchor point of the time stamp by *offset_x* and *offset_y*.
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+        If a single value *offset* is given, *offset_y* = *offset_x* =
+        *offset*.
+    font : str
+        Font of the timestamp and the optional label.
```
yvonnefroehlich, line 44:
```suggestion
        Font of the time stamp and the optional label.
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+        For GMT<=6.4.0, the parameter can only change the font style.
+        For GMT>=6.5.0, the parameter can change the font style and font color.
+    timefmt : str
+        Format of the time information in the UNIX time stamp. This format is
+        parsed by the C function ``strftime``, so that virtually any text can
+        be used (even not containing any time information).
+
+    Examples
+    --------
+    >>> # Plot the GMT timestamp logo.
```
yvonnefroehlich, line 54:
```suggestion
    >>> # Plot the GMT time stamp logo.
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    >>> import pygmt
+    >>> fig = pygmt.Figure()
+    >>> fig.timestamp()
+    >>> fig.show()
+    <IPython.core.display.Image object>
+
+    >>> # Plot the GMT timestamp logo with a custom label.
```
yvonnefroehlich, line 61:
```suggestion
    >>> # Plot the GMT time stamp logo with a custom label.
```

pygmt/src/timestamp.py
yvonnefroehlich: Hm. Is there a similar situation for the Julia wrapper, which can serve as an orientation?

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    self,
+    text=None,
+    label=None,
+    justification="BL",
+    offset=("-54p", "-54p"),
+    font="Helvetica",
+    timefmt="%Y %b %d %H:%M:%S",
+):
+    r"""
+    Plot the GMT timestamp logo.
```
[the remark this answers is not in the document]
seisman: Wikipedia says "timestamp" (https://en.wikipedia.org/wiki/Timestamp).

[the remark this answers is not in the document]
michaelgrund: I think both are possible, but would also go with `timestamp` as is.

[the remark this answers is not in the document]
yvonnefroehlich: I am also fine with "timestamp" (no space).

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+        *offset* or (*offset_x*, *offset_y*).
+        Offset the anchor point of the timestamp by *offset_x* and *offset_y*.
+        If a single value *offset* is given, *offset_y* = *offset_x* =
+        *offset*.
+    font : str
+        Font of the timestamp and the optional label.
+        For GMT<=6.4.0, the parameter can only change the font style.
+        For GMT>=6.5.0, the parameter can change the font style and font color.
+    timefmt : str
+        Format of the time information in the UNIX time stamp. This format is
```
yvonnefroehlich, line 48:
```suggestion
        Format of the time information in the UNIX timestamp. This format is
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    timefmt="%Y %b %d %H:%M:%S",
+):
+    r"""
+    Plot the GMT timestamp logo.
+
+    Parameters
+    ----------
+    text : None or str
+        If ``None``, the current UNIX time stamp is shown in the GMT timestamp
+        logo. Set this parameter to replace the UNIX time stamp with a
```
yvonnefroehlich, line 28:
```suggestion
        logo. Set this parameter to replace the UNIX timestamp with a
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    font="Helvetica",
+    timefmt="%Y %b %d %H:%M:%S",
+):
+    r"""
+    Plot the GMT timestamp logo.
+
+    Parameters
+    ----------
+    text : None or str
+        If ``None``, the current UNIX time stamp is shown in the GMT timestamp
```
yvonnefroehlich, line 27:
```suggestion
        If ``None``, the current UNIX timestamp is shown in the GMT timestamp
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,106 @@
...
+    ----------
+    text : None or str
+        If ``None``, the current UNIX timestamp is shown in the GMT timestamp
+        logo. Set this parameter to replace the UNIX timestamp with a
+        custom text string instead. The text must be less than 64 characters.
+        *Requires GMT>=6.5.0*.
+    label : None or str
+        The text string shown after the GMT timestamp logo.
+    justification : str
+        Justification of the timestamp. *justification* is a two-character code
```
michaelgrund, line 34:
```suggestion
        Justification of the timestamp. The *justification* is a two-character code
```

pygmt/tests/test_timestamp.py
```diff
@@ -0,0 +1,91 @@
...
+from pygmt import Figure, clib
+
+with clib.Session() as _lib:
+    gmt_version = Version(_lib.info["version"])
+
+
+@pytest.fixture(scope="module", name="faketime")
+def fixture_faketime():
+    """
+    Fake datetime that will be passed to the "timefmt" paramater, so that the
```
michaelgrund, line 15:
```suggestion
    Fake datetime that will be passed to the "timefmt" parameter, so that the
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers for reviews.

pygmt/tests/test_timestamp.py
```diff
@@ -0,0 +1,88 @@
...
+    Check if the "justification" parameter works.
+
+    Only a subset of justification codes are tested to avoid overlapping
+    timestamps.
+    """
+    fig = Figure()
+    fig.basemap(projection="X10c/5c", region=[0, 10, 0, 5], frame=0)
+    for just in ["BL", "BR", "TL", "TR"]:
+        fig.timestamp(justification=just, timefmt=just)
+    return fig
```
weiji14, line 50: The output of the justification image doesn't look correct?



All of the timestamps are on the bottom left (BL). Shouldn't they be at the four different corners?

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,102 @@
...
+            # See https://github.com/GenericMappingTools/gmt/issues/7107.
+            kwdict["U"] += f"+o{offset}/{offset}"
+        else:
+            kwdict["U"] += f"+o{offset}"
+
+    # The +t modifier was added in GMT 6.5.0.
+    # See https://github.com/GenericMappingTools/gmt/pull/7127.
+    if text is not None:
+        if Version(__gmt_version__) <= Version("6.4.0"):
+            raise GMTInvalidInput("The parameter 'text' requires GMT>=6.5.0.")
```
weiji14, line 89: Not that we have a GMT 6.4.1, but this is more precise given the second line.

```suggestion
        if Version(__gmt_version__) < Version("6.5.0"):
            raise GMTInvalidInput("The parameter 'text' requires GMT>=6.5.0.")
```

pygmt/src/timestamp.py
```diff
@@ -0,0 +1,102 @@
...
+        **C**\ (enter), or **R**\ (ight)) and a vertical (**T**\ (op),
+        **M**\ (iddle), or **B**\ (ottom)) code.
+    offset : str or tuple
+        *offset* or (*offset_x*, *offset_y*).
+        Offset the anchor point of the timestamp by *offset_x* and *offset_y*.
+        If a single value *offset* is given, *offset_y* = *offset_x* =
+        *offset*.
+    font : str
+        Font of the timestamp and the optional label. The parameter can't
+        change the font color for GMT<=6.4.0.
```
weiji14, line 46: Should a GMTInvalidInput error be raised when users pass `font` as a parameter with GMT 6.3.0? Or is the error clear enough from GMT?

seisman: The `font` parameter works for any GMT versions. 

For GMT>6.4.0, it can change both the **font ID** and **font color**, but for GMT<=6.4.0, the font color is always black so it can only change the font ID.

pygmt/tests/test_timestamp.py
seisman: Unlike other embellishments like inset, legend or direction rose, the timestamp logo are always located at the lower left corner of the map. I believe this is GMT's design.

See https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-MAP_LOGO_POS

> Sets the justification and the position of the logo/timestamp box **relative to the current plot’s lower left corner** (i.e., map origin)

pygmt/src/timestamp.py
weiji14: Ah ok, maybe reword it into something like this?
```suggestion
        Font of the timestamp and the optional label. The parameter can't
        change the font color for GMT<=6.4.0, only the font label.
```

seisman: I'm afraid "font label" is incorrect here. It should be "font ID" or "font name".

pygmt/tests/test_timestamp.py
weiji14: Interesting design choice :sweat_smile: Actually, I just realized I was thinking of `position` rather than `justification`. Maybe we should document the default `offset="-54p/-54p"` under the offset docstring?

pygmt/src/timestamp.py
weiji14: Owh, you mean like font style (e.g. Helvetica)? Feel free to edit the suggested changes.

pygmt/tests/test_timestamp.py
seisman: The default values are clear from the function definition:

`Figure.timestamp((text=None, label=None, justification='BL', offset=('-54p', '-54p'), font='Helvetica,black', timefmt='%Y %b %d %H:%M:%S')`

I'm hesitate to document the defaults in the docstring.

weiji14: Ah ok, was thinking of adding `[Default is ('-54p', '-54p')]`, but it's not really needed I guess.

weiji14 (approved):
Only a very minor suggestion as mentioned in https://github.com/GenericMappingTools/pygmt/pull/2208/files#r1122590265 (up to you to apply), but otherwise wait for second reviewer to approve.

pygmt/src/timestamp.py
seisman, line 46:
```suggestion
        change the font color for GMT<=6.4.0, only the font ID.
```

seisman: Added "only the font ID" in 007964375e091a93c8e7c63c840fb3affc2a8e01.

michaelgrund:
Would you also like to add a gallery example or should we do this in a separate PR @seisman ?

seisman:
> Would you also like to add a gallery example or should we do this in a separate PR @seisman ?

It's not on my todo list, so better to wait for someone else to work on the gallery example in another PR.

PR #2213, opened 2022-11-26
pygmt.config: Support tab auto-completion for all GMT defaults

seisman:

Let `pygmt.config()` support tab auto-completion by manually setting the `__signature__` attribute of the `pygmt.config()` class, inspired by the implementation in https://github.com/GenericMappingTools/pygmt/pull/1282.

Screenshot:

image

Fixes #

**Reminders**

**Slash Commands**


weiji14 (commented):
The tab autocomplete is nice! Just need to decide how to list the 135 (if I didn't miscount) parameters in https://docs.generic-mapping-tools.org/6.4/gmt.conf.html :sweat_smile:

pygmt/src/config.py
```diff
@@ -20,6 +22,15 @@ class config:  # pylint: disable=invalid-name
     Full GMT defaults list at :gmt-docs:`gmt.conf.html`
     """
 
+    # Manually set the __signature__ attribute to enable tab autocompletion
+    _keywords = ["FONT_LABEL", "FONT_TAG", "FONT_TITLE"]
```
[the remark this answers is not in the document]
seisman: `gmt defaults` can output all the keywords and their current values. So at least we can call `gmt defaults` and extract the keywords from the output.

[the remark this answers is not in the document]
seisman: Calling `gmt defaults` is not a perfect idea, because the code will be executed when importing the `pygmt` package. Perhaps manually maintaining the the long list of keywords is not a bad idea because the list of GMT keywords don't change much between versions.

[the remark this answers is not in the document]
weiji14: Ok I guess. Even if the list gets outbof date because GMT adds new configs, people can still use it, just without tab completion.

seisman:
> The tab autocomplete is nice! Just need to decide how to list the 135 (if I didn't miscount) parameters in [docs.generic-mapping-tools.org/6.4/gmt.conf.html](https://docs.generic-mapping-tools.org/6.4/gmt.conf.html) 😅

I've manually added the long list of keywords to `config.py` in 2fa17a533f7d346162e9db81dbd557621c865e05.

The long list contains 128 normal keywords as listed by the following command:
```
gmt defaults | awk -F"=" '/^[^#]/ {print $1}' | sed 's/,[ \t]\?/,/g' | awk '{printf("\"%s\",\n", $1)}'
```
plus 8 special keywords like `FONT` and `FONT_ANNOT`, so the total number is 136.

pygmt/src/config.py
```diff
@@ -20,6 +22,154 @@ class config:  # pylint: disable=invalid-name
...
+    ] + [  # special keywords
+        "FONT",
+        "FONT_ANNOT",
+        "FORMAT_TIME_MAP",
+        "MAP_ANNOT_OFFSET",
+        "MAP_GRID_CROSS_SIZE",
+        "MAP_GRID_PEN",
+        "MAP_TICK_LENGTH",
+        "MAP_TICK_PEN",
+    ]
```
seisman, line 164: I think it makes more sense to also put the `self.special_params` attribute out of the `__init__` function.

weiji14: Yeah, I was thinking about that too. Maybe create a `_special_params` variable to hold that dictionary.

seisman: Done in 318501bf8d870278a5199649b93bd31a18d84f71, and the variable name is `_special_keywords` instead of `_special_params`.

weiji14: Nice work, just double checked one by one and it looks good!

weiji14 (approved):

PR #2219, opened 2022-12-03
Figure.meca: Refactor the tests for pass a single focal mechanism to spec

seisman:

The `Figure.meca` tests are a little messy and difficult to maintain because each test uses different focal mechanisms and event information. I plan to refactor the tests and this PR is the first step. 

This PR focuses on testing passing a single focal mechanism to the `spec` parameter using different data types (dict, file, 1darray and DataFrame). Note that all tests now single a single baseline image `test_meca_spec_dict.png`.

Changes in this PR:

- Rename the test test_meca_spec_dictionary to test_meca_spec_dict
- Improve the test_meca_spec_dict test with a better baseline image
- Rewrite test_meca_spec_1d_array test to use the existing baseline image
- Rewrite the test_meca_spec_file test to reuse the existing baseline image
- Rename the test test_meca_spec_dataframe to test_meca_spec_two_dataframe
- Add a new test test_meca_spec_dataframe to test passing a dataframe to meca

### DIFF
--- a/pygmt/tests/test_meca.py
+++ b/pygmt/tests/test_meca.py
@@ -24,7 +24,7 @@ def test_meca_spec_dict():
         scale="2.5c",
         region=[-1, 1, 4, 6],
         projection="M14c",
-        frame=2,
+        frame=True,
     )
     return fig
 
--- a/pygmt/tests/test_meca.py
+++ b/pygmt/tests/test_meca.py
@@ -29,6 +29,26 @@ def test_meca_spec_dict():
     return fig
 
 
+@pytest.mark.mpl_image_compare(filename="test_meca_spec_dict.png")
+def test_meca_spec_1darray():
+    """
+    Test supplying a 1D numpy array containing a single focal mechanism to the
+    spec parameter.
+    """
+    fig = Figure()
+    fig.meca(
+        # The columns are:
+        # longitude, latitude, depth, strike, dip, rake, magnitude
+        spec=np.array([0, 5, 0, 0, 90, 0, 5]),
+        convention="aki",
+        region=[-1, 1, 4, 6],
+        scale="2.5c",
+        projection="M14c",
+        frame=True,
+    )
+    return fig
+
+
 @pytest.mark.mpl_image_compare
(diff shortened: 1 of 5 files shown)


pygmt/tests/test_meca.py
```diff
@@ -54,7 +118,7 @@ def test_meca_spec_dict_list():
 
 
 @pytest.mark.mpl_image_compare
-def test_meca_spec_dataframe():
+def test_meca_spec_two_dataframe():
```
willschlitzer, line 121: Why is this being renamed? As far as I can tell there is just a single DataFrame in this function.

seisman: I was planning to have two groups of tests:

1. In tests `test_meca_spec_*`, a single focal mechanism in different data types (dict, DataFrame, 1darray and plaintext file) is passed to the `spec` parameter
2. In tests `test_meca_spec_two_*`, two focal mechanisms in different data types (dict with list as keys, DataFrame, 2D list, 2D array and plaintext file) are passed to the `spec` parameter.

That's why I added the `_two` suffix to the tests, but I agree that's not a good name for these tests. Do you have better ideas?

pygmt/tests/test_meca.py
```diff
@@ -54,7 +118,7 @@ def test_meca_spec_dict_list():
...
     """
     Test supplying a pandas.DataFrame containing focal mechanisms and locations
     to the spec parameter.
```
willschlitzer, line 124:
```suggestion
    Test supplying two focal mechanisms and their locations by passing a pandas.DataFrame 
    to the spec parameter.
```

I think this makes it more clear why "two" is in the function name.

pygmt/tests/test_meca.py
willschlitzer: That makes sense. I think it will be hard to come up with a descriptive test name without making it too wordy. I think it might just be easier to update the docstring (I made my suggestion below) to make it clear that it's passing two focal mechanisms.

seisman:
I'm closing the issue and will open another PR to reorganize the meca tests later.

PR #2226, opened 2022-12-05
Figure.meca: Fix line and circle of offset parameter for dict/pandas input 

yvonnefroehlich:

The line and the small circle are not plotted or modified via the `offset` parameter when passing a dictionary to the `spec` parameter of `pygmt.Figure.meca`.

Fixes / For context see: https://github.com/GenericMappingTools/pygmt/issues/2016#issuecomment-1336356855

**Reminders**

**Slash Commands**


pygmt/src/meca.py
```diff
@@ -296,7 +296,6 @@ def meca(
             spec[["plot_longitude", "plot_latitude"]] = spec[
                 ["plot_longitude", "plot_latitude"]
             ].astype(str)
-            kwargs["A"] = True
```
seisman, line 299: The original idea of this line is: if `plot_longitude` and `plot_latitude` are specified, then we assume users want to offset the beachballs, so we set the `kwargs["A"]` parameter.

I think it makes more sense to change this line to:
```
if kwargs["A"] is None:
    kwargs["A"] = True
```

yvonnefroehlich: This change is done in  commit https://github.com/GenericMappingTools/pygmt/pull/2226/commits/ab68ccc5b9d639c6fa9e9d35172c6f0553c72230.  However, it (the upper line) is changed in commit https://github.com/GenericMappingTools/pygmt/pull/2226/commits/37102c5b30e8b59a44c0c776b7aa2d98da3ca9ee to:
```python
if kwargs.get("A") is None:
    kwargs["A"] = True
```

seisman (approved):

michaelgrund (approved):

PR #2223, opened 2022-12-05
Adjust data checks in datasets sample tests

michaelgrund:
**Description of proposed changes**

As proposed in #2211 this PR adjusts the data checks in the datasets sample tests for the min() and max() values in the corresponding tables/dfs.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2211, opened 2022-11-25
Add ternary sample dataset

michaelgrund:
Related to #2138.

**Description of proposed changes**

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2138, opened 2022-09-25
Add gallery example for Figure.ternary method

michaelgrund:
**Description of proposed changes**

This PR adds a gallery example showing the usage of Figure.ternary(). Example is adopted from the corresponding [GMT docs](https://docs.generic-mapping-tools.org/dev/ternary.html#examples).

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2227, opened 2022-12-05
Add usage of z value in ternary dataset

michaelgrund:

Just noticed that @yvonnefroehlich made an overlooked comment right before PR #2211 was merged in. 

> Maybe I am missing here some context.
> 
> Do I understand it correctly that this dataset should be used for the `ternary` gallery example in PR #2138 which is based on the corresponding GMT upstream example https://docs.generic-mapping-tools.org/dev/ternary.html#examples? In this case, the fourth column of the dataset is need for the color-coding, i.e., the fill color of the symbols, currently circles (**-C** or`cmap=True`).
> 
> This dataset seems to be taken from this MATLAB file exchange record https://de.mathworks.com/matlabcentral/fileexchange/7210-ternary-plots?s_tid=mwa_osa_a. There, the fourth column is listed as the "effective dielectric permittivity" (following the header of `termain.m`).
> 
> At the moment I am wondering if the column names are correct: We read 0->water, 1->air, 2->limestone. In the MATLAB file, they say 1->solid/limestone 2->water 3->gas/air (following the header of `termain.m`)?

I double checked this point and indeed the convention was not correct. Additonally, the z value of the table (permittivity) is used in the gallery example in PR #2138 for coloring. Also a point which we should change in upstream GMT although it's only an example dataset to show the usage of ternary plots.

#2223 needs to be merged first before adding the ``permittivity`` column to tests.

**Reminders**

**Slash Commands**


seisman:
I can't access the Matlab file so can't verify it.

> Also a point which we should change in upstream GMT although it's only an example dataset to show the usage of ternary plots.

Please submit an upstream fix first.

pygmt/datasets/samples.py
```diff
@@ -369,16 +369,15 @@ def _load_rock_sample_compositions():
...
     """
 
     fname = which("@ternary.txt", download="c")
     return pd.read_csv(
         fname,
         delim_whitespace=True,
         header=None,
-        names=["water", "air", "limestone"],
-        usecols=(0, 1, 2),
+        names=["limestone", "water", "air", "permittivity"],
```
seisman, line 380: Also need to update the test.

pygmt/datasets/samples.py
```diff
@@ -369,16 +369,15 @@ def _load_rock_sample_compositions():
     Returns
     -------
     data : pandas.DataFrame
-        The data table with columns "water", "air", and "limestone".
+        The data table with columns "limestone", "water", "air" and "permittivity".
```
yvonnefroehlich, line 372:
```suggestion
        The data table with columns "limestone", "water", "air",
        and "permittivity".
```

pygmt/tests/test_datasets_samples.py
```diff
@@ -186,9 +186,11 @@ def test_rock_sample_compositions():
     """
     data = load_sample_data(name="rock_compositions")
     assert data.shape == (1000, 3)
+    assert data["limestone"].min() == 0
+    assert data["limestone"].max() == 1
```
yvonnefroehlich, line 190: After adding the `permittivity` there are four columns.

```suggestion
    data = load_sample_data(name="rock_compositions")
    assert data.shape == (1000, 4)
    assert data["limestone"].min() == 0
    assert data["limestone"].max() == 1
```

seisman (approved):

willschlitzer (approved):

PR #2260, opened 2022-12-20
Rename tutorial files to prevent duplicate file names

### LINKED ISSUE #2250: Incorrect build documentation due to identical file names of examples
It seems like it is not allowed by Sphinx-Gallery that two examples have identical filenames, even when they are saved in different folders. Currently, this is the case for:
- Intro to PyGMT - **Create a contour map**: https://www.pygmt.org/dev/get_started/contour_map.html
- Tutorials - **Creating a map with contour lines**: https://www.pygmt.org/dev/tutorials/advanced/contour_map.html

When running `make html` to build the documentation locally, the following warning occurs:
```
WARNING: Duplicate example file name(s) found. Having duplicate file names will break some links.
List of files: ['../examples/get_started/contour_map.py']
```

According to this the documentation for the tutorial is wrongly build:
- correct - v0.7.0: https://www.pygmt.org/latest/tutorials/advanced/contour_map.html
- wrong - dev after adding "Create a contour map": https://www.pygmt.org/dev/tutorials/advanced/contour_map.html

Thus, we have to rename one of these two example files.

willschlitzer:
This adds "tutorial_" to the start of `first_figure.py` and `contour_map.py` to avoid naming conflicts when building documentation.

Fixes #2250 

**Reminders**

**Slash Commands**


michaelgrund (approved):

seisman:
I don't think I like the new names. The `tutorial_` prefix makes the URL much longer and doesn't provide more information. What about simply renaming `contour_map` to `contours`?

willschlitzer:
> I don't think I like the new names. The `tutorial_` prefix makes the URL much longer and doesn't provide more information. What about simply renaming `contour_map` to `contours`?

I don't feel that strongly about it. I figure that adding `tutorial_` to the front groups the tutorials together, but can also rename it to `contours` if that would make things easier.

@GenericMappingTools/gmt-maintainers Thoughts?

michaelgrund:
> > I don't think I like the new names. The `tutorial_` prefix makes the URL much longer and doesn't provide more information. What about simply renaming `contour_map` to `contours`?
> 
> I don't feel that strongly about it. I figure that adding `tutorial_` to the front groups the tutorials together, but can also rename it to `contours` if that would make things easier.
> 
> @GenericMappingTools/gmt-maintainers Thoughts?

I would prefer @willschlitzer's solution.

seisman:
> > > > I don't think I like the new names. The `tutorial_` prefix makes the URL much longer and doesn't provide more information. What about simply renaming `contour_map` to `contours`?
> > > 
> > > 
> > > I don't feel that strongly about it. I figure that adding `tutorial_` to the front groups the tutorials together, but can also rename it to `contours` if that would make things easier.
> > > @GenericMappingTools/gmt-maintainers Thoughts?
> > 
> > 
> > I would prefer @willschlitzer's solution.
> 
> I like the grouping-effect due to adding `tutorial_` at the beginning of the filenames.

All these files are in the `get_started` directory, so they're already grouped.

yvonnefroehlich:
Ping @weiji14 and @maxrjones for their thoughts and preferences on this, please.

weiji14:
I almost prefer a third option - prefixing with the sequence number of the tutorial (i.e. `01_first_figure`, `02_contour_map`) since they are numbered in https://www.pygmt.org/dev/get_started/index.html. How about we vote for this?

:+1: Prefix with `tutorial_`
:tada: Rename `contour_map` to `contours` and so on
:eyes: Prefix with number (e.g. `01_`, `02_`)

seisman:
> Prefix with number (e.g. `01_`, `02_`)

Prefixing with numbers sounds like a good idea, but the contour tutorial may not be the 2nd tutorial when we add more tutorials in the future, so we may have to rename it in the future.

yvonnefroehlich:
I think I am OK with numbers as prefix.

willschlitzer:
Changed to use numbers.

weiji14 (approved):

yvonnefroehlich (approved):

seisman (approved):

Issue #2302, opened 2023-01-04
Remove the deprecated load_xx functions in v0.9.0 (deprecated since v0.6.0)


seisman:
In PR https://github.com/GenericMappingTools/pygmt/pull/1685, we added a single `load_sample_data` function to load sample data used in tutorials and examples, and deprecated `load_xx` functions like `load_japan_quakes` in v0.6.0. These deprecated functions will be removed in v0.9.0.

Since our next targeted release is v0.9.0, now it's time to remove these deprecated functions. IMHO, we should remove these deprecated functions **as early as possible**, rather than removing them a few days before we make the v0.9.0 release. In this way, we can have more time (3 months vs a few days) to know whether removing these functions would cause any unexpected problems.

This issue is created to track the progress of removing these functions.

- [x] Remove `load_japan_quakes`: https://github.com/GenericMappingTools/pygmt/pull/2301
- [x] Remove `load_sample_bathymetry` https://github.com/GenericMappingTools/pygmt/pull/2305
- [x] Remove `load_fractures_compilation` https://github.com/GenericMappingTools/pygmt/pull/2303
- [x] Remove `load_hotspots` https://github.com/GenericMappingTools/pygmt/pull/2309
- [x] Remove `load_mars_shape` #2304 
- [x] Remove `load_ocean_ridge_points` https://github.com/GenericMappingTools/pygmt/pull/2308
- [x] Remove `load_usgs_quakes` #2306 

I've made the PR https://github.com/GenericMappingTools/pygmt/pull/2301 to remove the deprecated `load_japan_quakes` function. This PR could be used as a reference if anyone wants to work on other deprecated functions.

willschlitzer:
@michaelgrund Since we're working at the same time, I'm holding off on `load_hotspots` and assuming you're working on it. Let me know if you want me to add a PR to remove it!

yvonnefroehlich:
In this context, I feel it would be helpful to also work on issue #1774 and the related PR #1814 by @maxrjones for adding a list or an overview of the available sample datasets 🙂. Currently, users do not know directly from the documentation which sample datasets are available via `datasets.load_sample_data` and have to run `datasets.list_sample_data`.

PR #2301, opened 2023-01-04
Remove the deprecated load_japan_quakes function (deprecated since v0.6.0)

seisman:
**Description of proposed changes**

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1386, opened 2021-07-16
Add function to import hotspot dataset

willschlitzer:
This pull request adds a function to import the `hotspots.txt` dataset used in some of the GMT code examples. Unfortunately, due to inconsistent spacing between the columns in the data, I was not able to use `pd.read_csv()` to read the data into a DataFrame. It would be great to get any feedback to shorten this code, as it currently appears messy.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

Issue #967, opened 2021-02-25
ObsPy integration


liamtoney:
[ObsPy](https://github.com/obspy/obspy/wiki) is the *de facto* standard tool for processing seismological data using Python. GMT is already heavily used by seismologists. It seems natural to make PyGMT integrate better with ObsPy. This issue can serve as a platform for broad discussion of how we might accomplish this.

On the seismology side in general, we already have [`Figure.meca()`](https://www.pygmt.org/latest/api/generated/pygmt.Figure.meca.html#pygmt.Figure.meca) wrapped. Some GMT plotting commands (namely, [`sac`](https://docs.generic-mapping-tools.org/dev/supplements/seis/sac.html)) could be wrapped by PyGMT and integrated with standard ObsPy data objects, such as [`Trace`](https://docs.obspy.org/packages/autogen/obspy.core.trace.Trace.html) and [`Stream`](https://docs.obspy.org/packages/autogen/obspy.core.stream.Stream.html) objects. We could even configure [`Figure.plot()`](https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot) to take [`Inventory`](https://docs.obspy.org/packages/autogen/obspy.core.inventory.inventory.Inventory.html) objects.

For waveform plotting, imagine:
```python
import pygmt
import obspy

tr = obspy.read()[0]

fig = pygmt.Figure()
fig.coast(...)  # Make your map
fig.plot(...)  # Plot your stations / event / etc.
fig.sac(trace=tr, ...)  # Plot your waveform
fig.show()
```
(Of course, perhaps we'd want to rename `Figure.sac()` for this example...)

We'd want to make ObsPy an *optional* dependency if we went down this route. I think that most of the work would be related to I/O, type-checking, etc. — as well as **surveying the seismology community on what'd be useful.**

Do folks have other ideas? Pinging @megies and @krischer in case you have any thoughts on this (thanks).

core-man:
> Some GMT plotting commands (namely, [`sac`](https://docs.generic-mapping-tools.org/dev/supplements/seis/sac.html)) could be wrapped by PyGMT

As a seismologist, I almost use `gmt sac` in every project. Currently, I call `lib.call_module` to use `gmt sac`. See a simple example:
```python
import pygmt
from pygmt.clib import Session

fig = pygmt.Figure()
fig.basemap(region=[-50, 100, -1, 3], 
            projection="X10c/5c",
[...]

PR #2306, opened 2023-01-04
Remove the deprecated load_usgs_quakes function (deprecated since v0.6.0)

willschlitzer:
This replaces `load_usgs_quakes` with the internal function `_load_usgs_quakes`. It also adds additional checks to `test_usgs_quakes`.

Addresses: #2302

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -219,8 +219,6 @@ def _load_usgs_quakes():
     """
     Load a table of global earthquakes from the USGS as a pandas.DataFrame.
 
-    This is the ``@usgs_quakes_22.txt`` dataset used in the GMT tutorials.
-
     Returns
     -------
     data : pandas.DataFrame
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -72,7 +72,6 @@ def load_sample_data(name):
     load_func_old = {
         "hotspots": load_hotspots,
         "mars_shape": load_mars_shape,
-        "usgs_quakes": load_usgs_quakes,
     }
 
     # Dictionary of private load functions
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -76,7 +76,6 @@ def load_sample_data(name):
         "japan_quakes": load_japan_quakes,
         "mars_shape": load_mars_shape,
         "ocean_ridge_points": load_ocean_ridge_points,
-        "usgs_quakes": load_usgs_quakes,
     }
 
     # Dictionary of private load functions
@@ -85,6 +84,7 @@ def load_sample_data(name):
         "earth_relief_holes": _load_earth_relief_holes,
         "maunaloa_co2": _load_maunaloa_co2,
         "notre_dame_topography": _load_notre_dame_topography,
+        "usgs_quakes": _load_usgs_quakes,
     }
 
     if name in load_func_old:
@@ -215,20 +215,11 @@ def load_sample_bathymetry(**kwargs):
     return data
 
 
-def load_usgs_quakes(**kwargs):
+def _load_usgs_quakes():
     """
-    (Deprecated) Load a table of global earthquakes from the USGS as a
-    pandas.DataFrame.
-
-    .. warning:: Deprecated since v0.6.0. This function has been replaced with
-       ``load_sample_data(name="usgs_quakes")`` and will be removed in
-       v0.9.0.
+    Load a table of global earthquakes from the USGS as a pandas.DataFrame.
 
     This is the ``@usgs_quakes_22.txt`` dataset used in the GMT tutorials.
-
-    The data are downloaded to a cache directory (usually ``~/.gmt/cache``) the
-    first time you invoke this function. Afterwards, it will load the data from
-    the cache. So you'll need an internet connection the first time around.
 
     Returns
     -------
@@ -236,15 +227,6 @@ def load_usgs_quakes(**kwargs):
         The data table. Use ``print(data.describe())`` to see the available
         columns.
     """
-
-    if "suppress_warning" not in kwargs:
-        warnings.warn(
-            "This function has been deprecated since v0.6.0 and will be "
-            "removed in v0.9.0. Please use "
-            "load_sample_data(name='usgs_quakes') instead.",
-            category=FutureWarning,
-            stacklevel=2,
-        )
     fname = which("@usgs_quakes_22.txt", download="c")
     data = pd.read_csv(fname)
     return data
--- a/pygmt/datasets/samples.py
+++ b/pygmt/datasets/samples.py
@@ -73,7 +73,6 @@ def load_sample_data(name):
         "bathymetry": load_sample_bathymetry,
         "hotspots": load_hotspots,
         "mars_shape": load_mars_shape,
-        "ocean_ridge_points": load_ocean_ridge_points,
         "usgs_quakes": load_usgs_quakes,
     }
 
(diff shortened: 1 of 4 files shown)


pygmt/datasets/samples.py
```diff
@@ -215,36 +215,18 @@ def load_sample_bathymetry(**kwargs):
...
     """
-    (Deprecated) Load a table of global earthquakes from the USGS as a
-    pandas.DataFrame.
-
-    .. warning:: Deprecated since v0.6.0. This function has been replaced with
-       ``load_sample_data(name="usgs_quakes")`` and will be removed in
-       v0.9.0.
+    Load a table of global earthquakes from the USGS as a pandas.DataFrame.
 
     This is the ``@usgs_quakes_22.txt`` dataset used in the GMT tutorials.
```
yvonnefroehlich, line 222: Please see the comments at https://github.com/GenericMappingTools/pygmt/pull/2301#discussion_r1062048779.
```suggestion
```

pygmt/tests/test_datasets_samples.py
```diff
@@ -72,10 +71,56 @@ def test_usgs_quakes():
     """
     Check that the dataset loads without errors.
```
yvonnefroehlich, line 72:
```suggestion
    Check that the @usgs_quakes_22.txt dataset loads without errors.
```

pygmt/datasets/samples.py
```diff
@@ -198,36 +198,16 @@ def load_sample_bathymetry(**kwargs):
...
-        warnings.warn(
-            "This function has been deprecated since v0.6.0 and will be "
-            "removed in v0.9.0. Please use "
-            "load_sample_data(name='usgs_quakes') instead.",
-            category=FutureWarning,
-            stacklevel=2,
-        )
     fname = which("@usgs_quakes_22.txt", download="c")
     data = pd.read_csv(fname)
     return data
```
seisman, line 213:
```suggestion
    return pd.read_csv(fname)
```

seisman:
I just realized that this sample dataset is not used in any examples. Should we remove it?

yvonnefroehlich:
> I just realized that this sample dataset is not used in any examples. Should we remove it?

Hm. Not sure, as there seem to be other sample datasets which are currently also not used in any example, e.g., `maunaloa_co2`, `hotspots`, or `mars_shape`.

seisman:
> > I just realized that this sample dataset is not used in any examples. Should we remove it?
> 
> Hm. Not sure, as there seem to be other sample datasets which are currently also not used in any example, e.g., `maunaloa_co2`, `hotspots`, or `mars_shape`.

- `mars_shape` is used in `pygmt/src/sphinterpolate.py`, 
- `maunaloa_co2` is used in `pygmt/tests/test_filter1d.py`
- `hotspots` was added in https://github.com/GenericMappingTools/pygmt/pull/1386, and it seems there was a plan to use it in a sphdistance example.

willschlitzer:
I'm fine with removing `load_usgs_quakes`. I don't remember why (or when) I added it.

weiji14:
A quick search on GitHub shows it being used before at https://github.com/GenericMappingTools/try-gmt-python/blob/master/try-gmt-python.ipynb. But it's 4 years old now.

How much of a burden is it to keep the function maintained? Maybe we could just keep it in case?

willschlitzer:
It's not too much of an issue to keep it maintained. I just assumed that @seisman suggested removing it to avoid the precedent that all GMT remote files have an associated PyGMT function to load them.

seisman:
> It's not too much of an issue to keep it maintained. I just assumed that @seisman suggested removing it to avoid the precedent that all GMT remote files have an associated PyGMT function to load them.

I'm OK with maintaining this function, and I believe this dataset for global earthquakes can be used in examples in the future.

My major concern is that the returned data table contains too many columns and most of them are  text strings and are unlikely used in any PyGMT examples. I suggest only keeping the "time", "latitude", "longitude", "depth" and "mag", similar to the data table returned by `_load_japan_quakes`.

willschlitzer:
> > It's not too much of an issue to keep it maintained. I just assumed that @seisman suggested removing it to avoid the precedent that all GMT remote files have an associated PyGMT function to load them.
> 
> I'm OK with maintaining this function, and I believe this dataset for global earthquakes can be used in examples in the future.
> 
> My major concern is that the returned data table contains too many columns and most of them are text strings and are unlikely used in any PyGMT examples. I suggest only keeping the "time", "latitude", "longitude", "depth" and "mag", similar to the data table returned by `_load_japan_quakes`.

I've never studied seismology, so I'm not sure how important information that pertains to the errors is to someone hoping to use this dataset. But I don't see the need to load this dataset with a separate function if we're going to remove quite a few of the columns. If it's the same columns as `_load_japan_quakes`, we might as well just use the Japan quakes dataset for any example/test functions. 

Looking at the [USGS Earthquake API](https://earthquake.usgs.gov/fdsnws/event/1/), we could probably create an example that incorporates getting a JSON file from the API and using it to create a plot/perform operations. I think that would be a better example of a real application of earthquake data instead of  `usgs_quakes_22.txt`.

pygmt/datasets/samples.py
```diff
@@ -73,6 +73,7 @@ def load_sample_data(name):
         "bathymetry": load_sample_bathymetry,
         "hotspots": load_hotspots,
         "mars_shape": load_mars_shape,
+        "ocean_ridge_points": load_ocean_ridge_points,
```
seisman, line 76:
```suggestion
```

seisman:
> > > It's not too much of an issue to keep it maintained. I just assumed that @seisman suggested removing it to avoid the precedent that all GMT remote files have an associated PyGMT function to load them.
> > 
> > 
> > I'm OK with maintaining this function, and I believe this dataset for global earthquakes can be used in examples in the future.
> > My major concern is that the returned data table contains too many columns and most of them are text strings and are unlikely used in any PyGMT examples. I suggest only keeping the "time", "latitude", "longitude", "depth" and "mag", similar to the data table returned by `_load_japan_quakes`.
> 
> I've never studied seismology, so I'm not sure how important information that pertains to the errors is to someone hoping to use this dataset. But I don't see the need to load this dataset with a separate function if we're going to remove quite a few of the columns. If it's the same columns as `_load_japan_quakes`, we might as well just use the Japan quakes dataset for any example/test functions. I don't feel strongly about removing this function, as someone may already be using it in their examples, but I don't wish to pare down the information in the returned dataset.

OK, I think it's fine to keep this function unchanged.

> Looking at the [USGS Earthquake API](https://earthquake.usgs.gov/fdsnws/event/1/), we could probably create an example that incorporates getting a JSON file from the API and using it to create a plot/perform operations. I think that would be a better example of a real application of earthquake data instead of `usgs_quakes_22.txt`.

Seismologists usually use [ObsPy's get_events() function](https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_events.html#obspy.clients.fdsn.client.Client.get_events) to get earthquakes from USGS or other seismic data centers. The function returns a [`Catalog` object](https://docs.obspy.org/packages/autogen/obspy.core.event.Catalog.html) which contains event information. This is related to the ObsPy integration in https://github.com/GenericMappingTools/pygmt/issues/967.

pygmt/datasets/samples.py
```diff
@@ -85,6 +85,7 @@ def load_sample_data(name):
         "notre_dame_topography": _load_notre_dame_topography,
         "ocean_ridge_points": _load_ocean_ridge_points,
         "rock_compositions": _load_rock_sample_compositions,
+        "usgs_quakes": _load_usgs_quakes,
```
seisman, line 88: Need to remove `"usgs_quakes": load_usgs_quakes` at line 75.

seisman (approved):

michaelgrund (approved):

PR #2329, opened 2023-01-17
Add a gallery example showing how to use patterns via the "fill" parameter (or similiar parameters)

### LINKED ISSUE #2323: Gallery example for histogram/plot showing bit and hachure patterns

Some functions like [`histogram`](https://www.pygmt.org/v0.8.0/api/generated/pygmt.Figure.histogram.html) and [`plot`](https://www.pygmt.org/v0.8.0/api/generated/pygmt.Figure.plot.html) have a `-Gp` or `-GP` option (in PyGMT, `-G` is aliased to `fill` after #1617) that sets bit and hachure patterns like so:

Predefined Bit and Hachure Patterns in GMT

@JingHuiTong suggested during the AGU 2022 Fall Meeting that there could be a gallery example for this, showing how to create e.g. histogram bars with these fill patterns, to be placed under https://www.pygmt.org/v0.8.0/gallery/index.html#histograms

Extends #1156.

References:
- https://docs.generic-mapping-tools.org/6.4/cookbook/features.html#specifying-area-fill-attributes
- https://docs.generic-mapping-tools.org/6.4/cookbook/predefined-patterns.html

Maybe

yvonnefroehlich:

This PR aims to add a gallery example showing how to use patterns via the `fill` parameter or similar parameters:
- [x] Add example code
- [x] Add documentation

**Preview**: https://pygmt-dev--2329.org.readthedocs.build/en/2329/gallery/symbols/patterns.html

References:
- https://docs.generic-mapping-tools.org/6.4/cookbook/features.html#specifying-area-fill-attributes
- https://docs.generic-mapping-tools.org/6.4/cookbook/predefined-patterns.html
 
Fixes #2323 (partly)

The example is currently placed in the "Symbols and markers" section.

**Reminders**

**Slash Commands**


examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,76 @@
...
+or similar parameters:
+
+- Symbols and polygons: :meth:`pygmt.Figure.plot` via ``fill``
+- Histogram bars or sectors: :meth:`pygmt.Figure.histogram` or
+  :meth:`pygmt.Figure.rose` via ``fill``
+- Anomalies: :meth:`pygmt.Figure.wiggle` via ``fillpositive``
+  and ``fillnegative``
+- Land and water masses: :meth:`pygmt.Figure.coast` via ``land``
+  and ``water``
+- Focal mechanisms: :meth:`pygmt.Figure.meca` via ``G`` and ``E``
```
seisman, line 15: We need to add aliases meca's `-G` and `-E` options.

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,76 @@
...
+
+The required argument has the following form:
+
+**P**\ |**p**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*]
```
seisman, line 19: I wonder if we should define a class for GMT patterns, e.g.,
```
class GMTPattern:
    def __init__(self, id, reversed=False, bgcolor="white", fgcolor="black", dpi=1200):
        ...
```

seisman (approved):

examples/gallery/symbols/patterns.py
yvonnefroehlich: Additional to the list at the beginning of this example, there is also the possibility to add a pattern via the `dcw` and `frame` parameters. The syntax seems to be difficult, as it seems to be something like **+g** **p|P** _resolution_ / _pattern_. The functionality seems to be limited, as it seems that no foreground and background colors could be given (please correct me if I am wrong, so far I did not figure out how to set colors here). 

_Example code_
```python
import pygmt

fig = pygmt.Figure()

with fig.subplot(nrows=1, ncols=2, figsize=("10c", "5c")):

    with fig.set_panel(panel=0):
        fig.basemap(
            region=[-12, 50, 30, 70],
            projection="M?",
            frame="afg",
        )
        fig.coast(
            land="lightbrown",
            water="lightblue",
            shorelines="1/0.5p,black",
            borders="1/0.25p,red",
            dcw="=EU+gp200/8",  # here resolution/pattern 
        )        

    with fig.set_panel(panel=1):
        fig.basemap(
            region=[0, 10, 0, 10],
            projection="X?",
            frame=["afg", "+gp400/8"],  # here resolution/pattern
        )

fig.show()
# fig.savefig(fname="pattern_dcw_frame.png")
```

_Output figure_
pattern_dcw_frame

seisman: > The syntax seems to be difficult, as it seems to be something like **+g** **p|P** _resolution_ / _pattern_.

This is the old syntax of patterns for GMT 4. The full old syntax is **-Gp**_dpi_/_pattern_[**:B**_color_[**F**_color_]]. I think the old syntax is still supported in GMT5 and GMT6 (backward compatibility), but it's no longer documented.

> The functionality seems to be limited, as it seems that no foreground and background colors could be given.

As I understand it, it's impossible to specify foreground/background colors and dpi using the new syntax in `coast -E`.

Ping @PaulWessel for comments on this.

PaulWessel:
This is surely a case where a fairly recent enhancement (specifying the fill of country via **-E** in **coast**), is stuck with a lot of other modifiers (e.g., **+r**) that elsewhere in **-G** for fills might mean different things.  So here, **+g** is meant to set **-G** but we cannot easily do the full list of modifiers, unless we do some coding changes.  If we just wanted to follow the pattern specs for **-G** then I guess we would want to use a 400 dpi red and blue pattern number 8 for Germany with

`-EDE+gp8+bred+fblue+r400`

but that only gives

```
coast [ERROR]: Option -E: Unrecognized modifier +bred.
coast [ERROR]: Option -E: Unrecognized modifier +fblue.
-R-10/30/40/60
```

for the CLI.  Of course, the trouble is in the C parsing since it is looking at **+r** to mean something with rounding of the board coordinates, not dpi of a pattern.  Looking at the docs, the **+f**, **+b** are otherwise not used by **-E** for other things, so it may be possible to finesse something but having two different meanings of **+r** in the same syntax is not good.  One option is to declare a bug since the **+r**|**R** main modifier prohibits us to pass a general pattern string that contains **+r.** We could decide that **+d**_dpi_ is the required modifier and allow **+f**, **+b**, and **+d** if found after **+g** to be parsed as a pattern.  We would then have to do the same to **-G** main pattern option and backwards introduce **+d** instead of **+r** there.

You can let me know if you think this is a good solution or not. I have not looked to see how easy this is to implement (well, doing **+d** in **-G** would be simple I think), but extracting the full pattern string given to **+g** might need some special checking.

examples/gallery/symbols/patterns.py
yvonnefroehlich: Should I do this first in a separate PR?
Are there any suggestions or preferences for the aliases? Maybe `fillcompression` for **-G** and `fill_extension` for **-E** (similar to `fillpositive` and `fillpositive` of `pygmt.Figure.wiggle`)? However, they are inconsistent, as one would be with hyphen and one without. So, maybe the other way around is better, i.e., `compressionfill` and `extensionfill` (similar to `uncertaintyfill` of `pygmt.Figure.velo`).

yvonnefroehlich:
Thanks a lot @seisman and @PaulWessel  for these fast and detailed comments on this!
I have never used GMT 4, and thus I know little to no about the GMT4 syntax. I apologize for bringing up this discussion. I did not expect this. It was a spontaneous idea to use patterns with the `dcw` (**-E**) as well as the `frame` (**-B**) parameters, as they both have the **+g** modifier. While searching for the right syntax, I orientated myself on the GMT example at https://docs.generic-mapping-tools.org/dev/gallery/ex34.html. Maybe/hopefully I can look at these thoughts in more detail on the weekend, but I can not grant for this yet. I apologize for the delay.

examples/gallery/symbols/patterns.py
michaelgrund: Would prefer the second option to have consistent namings.

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,76 @@
...
+
+fig = pygmt.Figure()
+fig.basemap(
+    region=[0, 10, 0, 12],
+    projection="X10c",
+    frame="rlbt+glightgray+tBit and Hachure Patterns",
+)
+
+# To use a pattern as fill append "p" and the number of the desired
+# pattern. By default, the pattern is plotted in black and white
```
michaelgrund, line 40:
```suggestion
# patterns. By default, the pattern is plotted in black and white
```

yvonnefroehlich: Hm. Currently I am not sure about changing this to plural.

michaelgrund: You're right, just ignore the change :wink:

yvonnefroehlich: No worries @michaelgrund!

examples/gallery/symbols/patterns.py
yvonnefroehlich: I started working on this in PR #2345.

yvonnefroehlich: I also updated the list at the beginning of this example with the aliases for **E** and **G** in commit https://github.com/GenericMappingTools/pygmt/pull/2329/commits/25c37771728cd0c41aa04d76fcb20fd2b0b35726.

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+- :meth:`pygmt.Figure.velo`: Uncertainty wedges and velocity error
+  ellipses via ``uncertaintyfill``
+- :meth:`pygmt.Figure.wiggle`: Anomalies via ``fillpositive``
+  and ``fillnegative``
+
+The required argument has the following form:
+
+**P**\ |**p**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*]
+
+*pattern* can either be a number in the range 1â€“90 or the name of a
```
michaelgrund, line 26:
```suggestion
*pattern* can either be a number in the range 1-90 or the name of a
```

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+1-, 8-, or 24-bit image raster file. The former will result in one of the 90
+predefined 64 x 64 bit-patterns provided with GMT; an overview can by found at
```
michaelgrund, line 28:
```suggestion
predefined 64 x 64 bit-patterns provided by GMT; an overview can by found at
```

yvonnefroehlich: From my side, this change is fine. I orientated myself on the formulation in the upstream GMT documentation at https://docs.generic-mapping-tools.org/dev/cookbook/features.html#specifying-area-fill-attributes.

examples/gallery/symbols/patterns.py
yvonnefroehlich: > I wonder if we should define a class for GMT patterns, e.g.,
> 
> ```
> class GMTPattern:
>     def __init__(self, id, reversed=False, bgcolor="white", fgcolor="black", dpi=1200):
>         ...
> ```

Maybe we should we open a separate issue for this?

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+- :meth:`pygmt.Figure.solar`: Day-light terminators via ``fill``
+- :meth:`pygmt.Figure.ternary`: Symbols via ``fill``
+- :meth:`pygmt.Figure.velo`: Uncertainty wedges and velocity error
+  ellipses via ``uncertaintyfill``
+- :meth:`pygmt.Figure.wiggle`: Anomalies via ``fillpositive``
+  and ``fillnegative``
+
+The required argument has the following form:
+
+**P**\ |**p**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*]
```
seisman, line 24:
```suggestion
**P**\ \|**p**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*]
```

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+and/or **+f**\ *color*) that will replace the default white and black pixels,
+respectively. Excluding *color* from a fore- or background specification yields
+a transparent image where only the back- or foreground pixels will be painted.
+The **+r**\ *dpi* modifier sets the resolution in dpi.
+"""
+
+import pygmt
+
+y = 13
+
```
seisman, line 46:
```suggestion
y = 11

```

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+    "p19+bred3+flightbrown",
+    # Invert the pattern by using a capitalized "P"
+    "P19+bred3+flightbrown",
+    # Change the resolution ("+r") to 100 dpi
+    "p19+bred3+flightbrown+r100",
+    # Make the background transparent by not giving a color after "+b";
+    # works analogous for the foreground
+    "p19+b+flightbrown+r100",
+]:
+    y -= 2
```
seisman, line 73:
```suggestion
```

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,91 @@
...
+        fill=pattern,
+    )
+    # Add a description of the pattern
+    fig.text(
+        x=4,
+        y=y,
+        text=pattern,
+        font="Courier-Bold",
+        justify="ML",  # justification of the text is Middle Left
+    )
```
seisman, line 89:
```suggestion
    )
    y -= 2
```

examples/gallery/symbols/patterns.py
yvonnefroehlich: Thanks for spotting this!

examples/gallery/symbols/patterns.py
```diff
@@ -0,0 +1,92 @@
...
+- :meth:`pygmt.Figure.ternary`: Symbols via ``fill``
+- :meth:`pygmt.Figure.velo`: Uncertainty wedges and velocity error
+  ellipses via ``uncertaintyfill``
+- :meth:`pygmt.Figure.wiggle`: Anomalies via ``fillpositive``
+  and ``fillnegative``
+
+The required argument has the following form:
+
+**p**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*] or
+**P**\ *pattern*\ [**+b**\ *color*\ ][**+f**\ *color*\ ][**+r**\ *dpi*]
```
seisman, line 25: This one works and is less than 80 characters:
```suggestion
**P**\|\ **p**\ *pattern*\ [**+b**\ *color*][**+f**\ *color*][**+r**\ *dpi*]
```

It's tested on https://www.tutorialspoint.com/online_restructure_editor.php

seisman (approved):

michaelgrund (approved):

PR #2345, opened 2023-02-01
Figure.meca: Add aliases for C, E, G, and W

yvonnefroehlich:
This PR aims to add aliases for `pygmt.Figure.meca`:
- [ ] add aliases
  - [x] **C** -> `cmap`
  - [x] **E** -> `extensionfill`
  - [x] **G** -> `compressionfill`
  - [x] **W** -> `pen`
- ~[ ] add unit test~

Related to comments https://github.com/GenericMappingTools/pygmt/pull/2329#discussion_r1090185604, https://github.com/GenericMappingTools/pygmt/pull/2329#discussion_r1092539819, and https://github.com/GenericMappingTools/pygmt/pull/2329#discussion_r1092849785.

**Reminders**

**Slash Commands**


pygmt/src/meca.py
```diff
@@ -211,6 +213,12 @@ def meca(
         **+s**\ *size* to set the diameter of the circle [Default is
         no circle]. Use **+p**\ *pen* to set the line pen attributes [Default
         is 0.25p].
+    compressionfill : str
+        Select color or pattern for filling the compressional quadrants
```
michaelgrund, line 217:
```suggestion
        Set fill shade, color, or pattern for the compressional quadrants
```

pygmt/src/meca.py
```diff
@@ -211,6 +213,12 @@ def meca(
...
+        [Default is black].
+    extensionfill : str
+        Select color or pattern for filling the extensive quadrants
```
michaelgrund, line 220:
```suggestion
        Set fill shade, color, or pattern for the extensive quadrants
```
To be consistent with e.g. the formulation in Figure.wiggle.

michaelgrund: We should also add at least one unit test to check if the fill works as expected.

yvonnefroehlich: I recognized already yesterday that the docstrings for `fill` are not consistent across the documentation.
So far I orientated myself on the one of `pygmt.Figure.plot`, as I was not sure about the meaning of the term "shade" and whether it is available for filling quadrants of beach balls (I had not time to check it until now - I apologize for this :slightly_frowning_face:).

However, I agree, that there should be a consistent formulation across the different plotting methods (with adjustments in case the available functionality is partly different), e.g.,:

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot
> fill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Select color or pattern for filling of symbols or polygons [Default is no fill]. fill can be a 1-D array, but it is only valid if using x/y and cmap=True is also required.

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.plot3d.html#pygmt.Figure.plot3d
> fill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Select color or pattern for filling of symbols or polygons [Default is no fill]. fill can be a 1-D array, but it is only valid if using x/y and cmap=True is also required.

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.velo.html
> uncertaintyfill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Sets the color or shade used for filling uncertainty wedges (spec="w") or velocity error ellipses (spec="e" or spec="r"). If uncertaintyfill is not specified, the uncertainty regions will be transparent. Note: Using cmap and zvalue="+e" will update the uncertainty fill color based on the selected measure in zvalue [magnitude error]. More details at https://docs.generic-mapping-tools.org/latest/cookbook/features.html#gfill-attrib.

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.wiggle.html#pygmt.Figure.wiggle
> fillpositive ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Set fill shade, color, or pattern for positive wiggles [Default is no fill].
> fillnegative ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Set fill shade, color, or pattern for negative wiggles [Default is no fill].

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.solar.html#pygmt.Figure.solar
> fill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Color or pattern for filling of terminators.

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.histogram.html#pygmt.Figure.histogram
> fill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Select color or pattern for filling of symbols or polygons [Default is no fill].

- https://www.pygmt.org/latest/api/generated/pygmt.Figure.rose.html#pygmt.Figure.rose
> fill ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Selects shade, color or pattern for filling the sectors [Default is no fill].

yvonnefroehlich: Based on the comparison of the docstrings for ``fill``, we have to decide regarding consistency:
- **verb**: "set" or "select"
- **grammar**: third person "s" or not (this is not consistent across parameters as well as functions/methods in the docs)
- **formulation structure**: "fill" or "for filling"
- **content**: mention "shade" explicitly or not (I guess it simply refers to gray shade i.e. gray color (?))

@GenericMappingTools/pygmt-maintainers should we go with the version of @michaelgrund's code suggestion above, or what are the preferences here? In case this change is welcome, I can update the (other) docstrings for ``fill`` in a separate PR in the next days.

P.S.: @michaelgrund I apologize for my probably a bit harsh comment above!

seisman: > as I was not sure about the meaning of the term "shade".

I also have no idea about "shade". Based on my understanding about this page (https://www.w3schools.com/colors/colors_shades.asp), I think "shades" means colors (not only gray) with different lights, so it's already covered by "colors".

So I would prefer not to mention "shades".

"Set fill color or pattern" or "Set color or pattern filling" (preferred) both sound good to me.

yvonnefroehlich: > > as I was not sure about the meaning of the term "shade".
> 
> I also have no idea about "shade". Based on my understanding about this page (https://www.w3schools.com/colors/colors_shades.asp), I think "shades" means colors (not only gray) with different lights, so it's already covered by "colors".

Yep, something in this direction I was also thinking. But how do I change or vary the "light of the color" via ``fill``?
The upstream GMT documentation at https://docs.generic-mapping-tools.org/6.4/cookbook/features.html#specifying-area-fill-attributes makes a difference between gray shade and RGB/HSV/CMYK color:
> In the first case we may specify a gray shade (0–255), RGB color (r/g/b all in the 0–255 range or in hexadecimal #rrggbb), HSV color (hue-saturation-value in the 0–360, 0–1, 0–1 range), CMYK color (cyan/magenta/yellow/black, each ranging from 0–100%), or a valid color name; in that respect it is similar to specifying the pen color settings (see pen color discussion under Section [Specifying pen attributes](https://docs.generic-mapping-tools.org/6.4/cookbook/features.html#specifying-pen-attributes)).

seisman: > But how do I change or vary the "light of the color" via `fill`?

Specifying a gray shade is as simple as giving a number from 0 to 255. For other colors, there is no easy to specify their shade. If you want a blue color with a different light, then you need to use other tools to get the name or RGB values of the color, and pass it to GMT.

xref: 

1. https://en.wikipedia.org/wiki/List_of_colors_by_shade. 
2. https://mdigi.tools/color-shades/#0000ff

yvonnefroehlich: Thanks! That is similar to my understanding.
I feel that mention "shade" explicitly can cause some confusions, as shade is only available for gray. Thus, I also prefer not to mention "shade" in the docs and state only "color" and "pattern".

yvonnefroehlich: I started updating the (other) docstrings for `fill` in PR #2351.

pygmt/src/meca.py
```diff
@@ -170,30 +172,30 @@ def meca(
...
+    component : str
         The component of the seismic moment tensor to plot.
 
         - ``"full"``: the full seismic moment tensor
         - ``"dc"``: the closest double couple defined from the moment tensor
           (zero trace and zero determinant)
         - ``"deviatoric"``: deviatoric part of the moment tensor (zero trace)
-    longitude: int, float, list, or 1-D numpy array
+    longitude : int, float, list, or 1-D numpy array
         Longitude(s) of event location. Must be the same length as the
```
michaelgrund, line 183:
```suggestion
        Longitude(s) of event location(s). Must be the same length as the
```

pygmt/src/meca.py
```diff
@@ -170,30 +172,30 @@ def meca(
...
         number of events. Will override the ``longitude`` values
         in ``spec`` if ``spec`` is a dict or pd.DataFrame.
-    latitude: int, float, list, or 1-D numpy array
+    latitude : int, float, list, or 1-D numpy array
         Latitude(s) of event location. Must be the same length as the
```
michaelgrund, line 187:
```suggestion
        Latitude(s) of event location(s). Must be the same length as the
```

pygmt/src/meca.py
```diff
@@ -170,30 +172,30 @@ def meca(
...
         number of events. Will override the ``latitude`` values
         in ``spec`` if ``spec`` is a dict or pd.DataFrame.
-    depth: int, float, list, or 1-D numpy array
+    depth : int, float, list, or 1-D numpy array
         Depth(s) of event location in kilometers. Must be the same length as
```
michaelgrund, line 191:
```suggestion
        Depth(s) of event location(s) in kilometers. Must be the same length as
```

yvonnefroehlich: Thanks! I updated some more occurrences in commit https://github.com/GenericMappingTools/pygmt/commit/e2b1a6df3a5adf3d8ca25a0fb29d12aa5367da75.

However, I am wondering why this commit does not show up in this PR. For me I looks like I committed to the right branch (please see https://github.com/GenericMappingTools/pygmt/compare/add-meca-aliases-E-G). Currently I am unfortunately not sure what I did wrong.

seisman:
@yvonnefroehlich Is this PR finished? I don't think a unit test is necessary, because this PR only adds aliases, and we're confident that aliases work.

seisman:
BTW, could you please split this PR into two separate PRs: one for the docstring changes and another one for the new aliases.

yvonnefroehlich:
> BTW, could you please split this PR into two separate PRs: one for the docstring changes and another one for the new aliases.

I submitted a separate PR #2360 for the docstring changes.

> @yvonnefroehlich Is this PR finished?

Yes, this PR should be finished. 
But I think, I like to add also aliases for **W** -> `pen` and **C** -> `cmap`. But I can do this in a separate PR 🙂.

> I don't think a unit test is necessary, because this PR only adds aliases, and we're confident that aliases work.

This is fine with me, but I have experience with testing code... Adding a unit test was suggested in https://github.com/GenericMappingTools/pygmt/pull/2345#discussion_r1094247635.

seisman:
> Yes, this PR should be finished.
> But I think, I like to add also aliases for **W** -> `pen` and **C** -> `cmap`. But I can do this in a separate PR .

It depends on you. You can keep working in this PR if you want to add more aliases.

pygmt/src/meca.py
```diff
@@ -211,6 +215,20 @@ def meca(
...
+        [Default is black].
+    extensionfill : str
+        Set color or pattern for filling extensive quadrants
+        [Default is white].
+    pen : str
+        Set pen attributes for outline of beachball
+        [Default is ``"0.25p,black,solid"``].
+    cmap : str
+        File name of a CPT file or a series of comma-separated colors (e.g.,
+        color1,color2,color3) to build a linear continuous CPT from those
```
seisman, line 229:
```suggestion
        *color1,color2,color3*) to build a linear continuous CPT from those
```

pygmt/src/meca.py
```diff
@@ -211,6 +215,20 @@ def meca(
...
+        Set color or pattern for filling extensive quadrants
+        [Default is white].
+    pen : str
+        Set pen attributes for outline of beachball
+        [Default is ``"0.25p,black,solid"``].
+    cmap : str
+        File name of a CPT file or a series of comma-separated colors (e.g.,
+        *color1,color2,color3*) to build a linear continuous CPT from those
+        colors automatically. The color of the compressive quadrants is
+        determined by the z-value (e.g., event depth) in the third column.
```
seisman, line 231: I wonder if the following changes are better.
```suggestion
        determined by the z-value (i.e., event depth or the third column for 
        an input file).
```

seisman: For an input file or an array, z-value is the third columns.
For dict and pd.DataFrame, z-value is the event depth.

seisman (approved):

pygmt/src/meca.py
```diff
@@ -211,6 +215,21 @@ def meca(
...
+        Set pen attributes for outline of beachball
+        [Default is ``"0.25p,black,solid"``].
+    cmap : str
+        File name of a CPT file or a series of comma-separated colors (e.g.,
+        *color1,color2,color3*) to build a linear continuous CPT from those
+        colors automatically. The color of the compressive quadrants is
+        determined by the z-value (i.e., event depth or the third column for
+        an input file).
     no_clip : bool
         Does NOT skip symbols that fall outside frame boundary specified by
```
michaelgrund, line 234:
```suggestion
        Do NOT skip symbols that fall outside frame boundary specified by
```

yvonnefroehlich: I agree with this change by @michaelgrund!
Maybe we should decide, whether we want to use either the general verb form or the third-person form in the docstrings? This is currently not consistent across the documentation, even not within a single function / method, e.g., here for `compressionfill`, `extensionfill`, and `pen` it is `set`, but for `offset` and `no_clip` it is (was) `offsets` and `does`.

michaelgrund (approved):

pygmt/src/meca.py
seisman: I think GMT decided to use the general verb form some years ago (see changes in https://github.com/GenericMappingTools/gmt/commit/a57fd6c2f5b16c7fcaaaaa4b806233d19f0cd75b and https://github.com/GenericMappingTools/gmt/commit/8256073836925692451f5ece92e2cb81ec7fad26)

yvonnefroehlich: Thanks for pointing out these upstream PRs. I started working on this in PR #2376.

PR #2366, opened 2023-02-14
Add a package-level variable __gmt_version__ for development use

### LINKED ISSUE #2362: Add a package-level variable __gmt_version__ for development use

As we're supporting multiple GMT versions, sometimes we need to check the current GMT version and see if it has a specific bug or provides a new feature.

To check the GMT version, we need to write codes like:
```
from pygmt import clib

with clib.Session() as _lib:                                                    
	gmt_version = Version(_lib.info["version"])

if gmt_version >= Version("6.4.0"):
    ...
```

The same code is being used in many places, so it's better to write a function for it. The function will be like this:
```python
from pygmt import clib

def get_gmt_version():
	with clib.Session() as _lib:                                                    
	    return Version(_lib.info["version"])
```

Yes, but the details need more discussions.

seisman:

Fixes #2362

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -22,7 +22,7 @@ import atexit as _atexit
 from importlib.metadata import version
 
 # Import modules to make the high-level GMT Python API
-from pygmt import datasets
+from pygmt import clib, datasets
 from pygmt.accessors import GMTDataArrayAccessor
 from pygmt.figure import Figure, set_display
 from pygmt.io import load_dataarray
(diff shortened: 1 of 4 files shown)


pygmt/__init__.py
```diff
@@ -76,6 +76,21 @@
 _atexit.register(_end)
 
 
+def _get_gmt_version():
```
seisman, line 79: I've added a utility function `_get_gmt_version` in the `pygmt/__init__.py` file. I wonder if it's better to define a package-level variable `__gmt_version__` instead.

weiji14: Sounds good. This would be similar to `__gdal_version__` in `rasterio`, see https://github.com/rasterio/rasterio/blob/6bf045cc5e02c9f28e78d48b97b308490207d27d/rasterio/__init__.py#L84-L86

pygmt/__init__.py
```diff
@@ -21,8 +21,15 @@
 import atexit as _atexit
 from importlib.metadata import version
 
+from pygmt import clib
+
+# Get semantic version through setuptools-scm
+__version__ = f'v{version("pygmt")}'  # e.g. v0.1.2.dev3+g0ab3cd78
+__commit__ = __version__.split("+g")[-1] if "+g" in __version__ else ""  # 0ab3cd78
+with clib.Session() as lib:
+    __gmt_version__ = lib.info["version"]
```
seisman, line 30: The package-level variable `__gmt_version__` is defined here.

I have to put the definition of `__gmt_version__` at the top otherwise there will be circular import errors (see https://github.com/GenericMappingTools/pygmt/actions/runs/4179854302/jobs/7240260379).

pygmt/__init__.py
```diff
@@ -21,8 +21,15 @@
...
+
 # Import modules to make the high-level GMT Python API
-from pygmt import datasets
```
seisman, line 25: I removed `from pygmt import datasets` because it's not needed.

pygmt/__init__.py
michaelgrund: Sounds good!

michaelgrund (approved):

Issue #1412, opened 2021-08-04
Automatic labels and titles to plots based on xarray metadata


weiji14:
**Description of the desired feature**

Originally posted by @leouieda as a comment about `xarray` integration at https://hackmd.io/@pygmt/SkVmX_20d:

> Any thoughts on better integration with xarray? Dataset objects can have rich metadata (coordinate names, units, projections, etc) that could be leveraged to automatically add labels and titles to plots. Their plotting backend does that with matplotlib and it's very convenient.

My interpretation of this 'feature request' is that `grdimage` could automatically add x-axis and y-axis labels/annotations to the plot, and include a title about the z-variable name (without the user setting it explicitly). Something based on https://github.com/pydata/xarray/pull/519.

![xarray automatic labels and titles](https://cloud.githubusercontent.com/assets/5356122/9125954/6c30206a-3c5b-11e5-8fd7-98c7ecd2ef0e.png)

The auto-labels/titles currently comes for free in `xarray` through the [`xarray.DataArray.plot.imshow()`](http://xarray.pydata.org/en/v0.19.0/generated/xarray.DataArray.plot.imshow.html) function. We could inspect the source code at https://github.com/pydata/xarray/blob/v0.19.0/xarray/plot/plot.py#L244 to see how this can be implemented in PyGMT.

**Are you willing to help implement and maintain this feature?** Long term stuff for PyGMT > v0.6.0

seisman:
> My interpretation of this 'feature request' is that `grdimage` could automatically add x-axis and y-axis labels/annotations to the plot, and include a title about the z-variable name (without the user setting it explicitly).

The idea sounds interesting, but I think it's challenging to implement it in PyGMT. 

The main problem is, PyGMT is "_stateless_". When calling `grdimage`, we don't know if users have already added labels/annotations/titles by calling `basemap`. Even if PyGMT is "_stateful_", we also don't know if users want to add labels/annotations/titles themselves later.

However, I find the automatic labels/annotations/titles feature is still very appealing when users want to quickly view a DataArray. 

In https://github.com/GenericMappingTools/pygmt/pull/2372, I've added the `imshow()` method to the GMTDataArrayAccessor class. With the new method, now it's possible to view a DataArray using a single line of code:

Here is an example using the feature implemented in PR #2372:
```
from pygmt.datasets import load_earth_relief
grid = load_earth_relief()
grid.gmt.imshow()
```
[...]

PR #2372, opened 2023-02-17
Add DataArray.gmt.imshow() function to show 2D DataArray easily

seisman:

Fixes #

**Reminders**

**Slash Commands**


pygmt/accessors.py
```diff
@@ -95,3 +96,19 @@ def gtype(self, value):
...
+        --------
+
+        >>> from pygmt.datasets import load_earth_relief
+        >>> grid = load_earth_relief()
+        >>> grid.gmt.imshow()
+        """
+        fig = Figure()
+        fig.grdimage(self._obj, frame=True)
+        fig.colorbar()
+        fig.show()
```
weiji14, line 114: Debating on whether we should add `return fig`, so that users could potentially add extra stuff to the figure like so:

```python
fig = grid.gmt.imshow()
fig.colorbar()
fig.show()
```

or would that get a bit confusing? I'm thinking a bit about how matplotlib does it where users can add extra elements to the figure, but unsure if we want to mimic that in PyGMT.

pygmt/accessors.py
```diff
@@ -95,3 +96,19 @@ def gtype(self, value):
...
+        Examples
+        --------
+
+        >>> from pygmt.datasets import load_earth_relief
+        >>> grid = load_earth_relief()
+        >>> grid.gmt.imshow()
+        """
+        fig = Figure()
+        fig.grdimage(self._obj, frame=True)
+        fig.colorbar()
```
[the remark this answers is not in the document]
seisman: > I'm a little on the fence about adding a colorbar. Sometimes the default colorbar looks very bad (e.g. too many black lines).

That's because the original CPT file has too many z-slices. `fig.colorbar(frame=True)` should always be good.

pygmt/accessors.py
seisman: See my comment in https://github.com/GenericMappingTools/pygmt/issues/1412#issuecomment-1434341967.

Initially, I propose to add the `imshow()` function as a quick DataArray viewer, so I expect to view the grid using a simple single command.

Similar to the `ax` parameter in [`xarray.DataArray.plot.imshow`](https://xarray.pydata.org/en/v0.19.0/generated/xarray.DataArray.plot.imshow.html), I think the `imshow` function can at least accept a `fig` parameter.

If `fig=None` (the default behavior), `grid.gmt.imshow()` should plot the grid and show the image.

If a `Figure` instance is given, then we know that users want to plot the grid in an existing figure, then we should not show the image.

Then the `imshow()` will be like:
```python
    def imshow(self, fig=None):
        if fig is None:
            fig = Figure()

        fig.grdimage(self._obj, frame=True)
        fig.colorbar()

        if fig is None:
            fig.show()
        else:
            return fig
```

Issue #2390, opened 2023-03-03
Add a gallery example for the Figure.timestamp method


seisman:
The `Figure.timestamp()` method was added in https://github.com/GenericMappingTools/pygmt/pull/2208, and a gallery example is needed (xref https://github.com/GenericMappingTools/pygmt/pull/2208#issuecomment-1451522619).

PR #2391, opened 2023-03-03
Add a gallery example for the Figure.timestamp method

michaelgrund:

Addresses #2390.

**Preview**: https://pygmt-dev--2391.org.readthedocs.build/en/2391/gallery/embellishments/timestamp.html

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -2,7 +2,7 @@
 Timestamp
 ---------
 The :meth:`pygmt.Figure.timestamp` method can draw the
-GMT time stamp logo on the map/plot. A custom label 
+GMT timestamp logo on the map/plot. A custom label 
 can be added via the ``label`` parameter.
 """
 
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -1,6 +1,14 @@
+"""
+Timestamp
+---------
+The :meth:`pygmt.Figure.timestamp` method can draw the
+GMT time stamp logo on the map/plot. A custom label 
+can be added via the ``label`` parameter.
+"""
+
 import pygmt
 
-# define static format string
+# Define static format string
 pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")
 
 fig = pygmt.Figure()
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -8,7 +8,7 @@ can be added via the ``label`` parameter.
 
 import pygmt
 
-# Define static format string
+# Define static format string.
 pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")
 
 fig = pygmt.Figure()
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -1,0 +1,10 @@
+import pygmt
+
+fig = pygmt.Figure()
+fig.timestamp()
+fig.show()
+
+# Plot the GMT timestamp logo with a custom label.
+fig = pygmt.Figure()
+fig.timestamp(label="Powered by PyGMT")
+fig.show()
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -11,12 +11,17 @@ of the figure. By default, the ``offset`` and
 (bottom-left), respectively. 
 """
 
+# sphinx_gallery_thumbnail_number = 1
+
 import pygmt
 
 fig = pygmt.Figure()
 fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
 fig.timestamp()
 fig.show()
+
+###############################################################################
+# Maybe some short description of the ``font`` and ``timefmt`` parameters.
 
 fig = pygmt.Figure()
 fig.coast(region="d", projection="H10c", land="black", water="cornsilk", frame="afg")
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -1,5 +1,8 @@
 import pygmt
 
+# define static format string
+pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")
+
 fig = pygmt.Figure()
 fig.timestamp()
 fig.show()
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -2,10 +2,13 @@
 Timestamp
 ---------
 The :meth:`pygmt.Figure.timestamp` method can draw the
-GMT time stamp logo on the map/plot. A custom label
+GMT timestamp logo on the figure. A custom label
 can be added via the ``label`` parameter. The timestamp
-will always be shown in the bottom-left corner of the
-figure.
+will always be shown relative to the bottom-left corner
+of the figure. By default, the ``offset`` and
+``justification`` parameters are set to
+``("-54p", "-54p")`` (x, y directions) and ``"BL"``
+(bottom-left), respectively. 
 """
 
 import pygmt
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -2,17 +2,17 @@
 Timestamp
 ---------
 The :meth:`pygmt.Figure.timestamp` method can draw the
-GMT timestamp logo on the map/plot. A custom label
-can be added via the ``label`` parameter.
+GMT time stamp logo on the map/plot. A custom label 
+can be added via the ``label`` parameter. The timestamp 
+will always be shown in the bottom-left corner of the 
+figure.
 """
 
 import pygmt
 
 fig = pygmt.Figure()
+fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
 fig.timestamp()
-fig.show()
-
 # Plot the GMT timestamp logo with a custom label.
-fig = pygmt.Figure()
-fig.timestamp(label="Powered by PyGMT")
+fig.timestamp(label="Powered by PyGMT", justification="TL")
 fig.show()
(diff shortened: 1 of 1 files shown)


michaelgrund:
Adding a static time format string via e.g. ``pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")`` does actually not work. Any ideas why @GenericMappingTools/pygmt-maintainers ?

seisman:
> Adding a static time format string via e.g. `pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")` does actually not work. Any ideas why @GenericMappingTools/pygmt-maintainers ?

It's because the `timefmt` parameter overrides the `FORMAT_TIME_STAMP` setting. See the relevant codes at: https://github.com/GenericMappingTools/pygmt/blob/8d165c14d369188eb13597a918243b22776b59d8/pygmt/src/timestamp.py#L100

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,21 @@
...
+---------
+The :meth:`pygmt.Figure.timestamp` method can draw the
+GMT timestamp logo on the map/plot. A custom label
+can be added via the ``label`` parameter.
+"""
+
+import pygmt
+
+# Define static format string.
+pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")
```
seisman, line 12: We don't need to have a static UNIX timestamp string here, because the gallery examples are not tested in the CI.

yvonnefroehlich:
> Adding a static time format string via e.g. `pygmt.config(FORMAT_TIME_STAMP="2023-03-01T20:45:15")` does actually not work. Any ideas why @GenericMappingTools/pygmt-maintainers ?

Hm. Probably you may want to use the `text` parameter of `pygmt.Figure.timestamp`  to modify  the date (please see https://www.pygmt.org/dev/api/generated/pygmt.Figure.timestamp.html#pygmt.Figure.timestamp). Please note, that using the `text` parameter requires GMT >=6.5.0. As I have not installed the dev version of GMT, I can not test this at the moment 🙁.
> text (None or [str](https://docs.python.org/3/library/stdtypes.html#str)) – If None, the current UNIX timestamp is shown in the GMT timestamp logo. Set this parameter to replace the UNIX timestamp with a custom text string instead. The text must be less than 64 characters. Requires GMT>=6.5.0.

From the upstream GMT documentation I understand that `FORMAT_TIME_STAMP`  does affect the format, not the date or text itself (please see the upstream Documentation https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-FORMAT_TIME_STAMP).
> Defines the format of the time information in the UNIX time stamp. This format is parsed by the C function strftime, so that virtually any text can be used (even not containing any time information) [default is %Y %b %d %H:%M:%S].

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,18 @@
...
+Timestamp
+---------
+The :meth:`pygmt.Figure.timestamp` method can draw the
+GMT timestamp logo on the map/plot. A custom label
+can be added via the ``label`` parameter.
+"""
+
+import pygmt
+
+fig = pygmt.Figure()
```
seisman, line 11: It's better to also plot a map via `basemap` or `coast`, so that users know that the timestamp will be shown on the bottom-left corner of their figures.

yvonnefroehlich:
> As already suggested by @weiji14 in #2208, in my opinion the justification shortcuts are not correct:
> I would say the labels must be switched across the diagonal axis. 

Hm. I feel they are correct. Maybe the confusion is coming from the difference between `position` and `justification`?
For the timestamp logo, only `justification` can be changed; the default is `BL`. `position` is fixed to `BL`.
For example, for an anchor point set to `BR` _I_ expect everyting be left above this point, same for the other three.
Maybe this rough sketch helps:
timestamp_anchorpoint01

michaelgrund:
> > As already suggested by @weiji14 in #2208, in my opinion the justification shortcuts are not correct:
> > I would say the labels must be switched across the diagonal axis.
> 
> Hm. I feel they are correct. Maybe the confusion is coming from the difference between `position` and `justification`? For the timestamp logo, only `justification` can be changed; the default is `BL`. `position` is fixed to `BL`. For example, for an anchor point set to `BR` _I_ expect everything to be left above this point, similar for the other three. Maybe this rough sketch helps: timestamp_anchorpoint01

Ok now I got it, thanks @yvonnefroehlich.

michaelgrund:
> > > As already suggested by @weiji14 in #2208, in my opinion the justification shortcuts are not correct:
> > > I would say the labels must be switched across the diagonal axis.
> > 
> > 
> > Hm. I feel they are correct. Maybe the confusion is coming from the difference between `position` and `justification`? For the timestamp logo, only `justification` can be changed; the default is `BL`. `position` is fixed to `BL`. For example, for an anchor point set to `BR` _I_ expect everything to be left above this point, similar for the other three.
> 
> Ok now I got it, thanks @yvonnefroehlich.

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,18 @@
+"""
+Timestamp
+---------
+The :meth:`pygmt.Figure.timestamp` method can draw the
+GMT time stamp logo on the map/plot. A custom label
```
yvonnefroehlich, line 5:
```suggestion
The :meth:`pygmt.Figure.timestamp` method can draw the
GMT timestamp logo on the map/plot. A custom label
```

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,18 @@
...
+can be added via the ``label`` parameter. The timestamp
+will always be shown in the bottom-left corner of the
+figure.
```
yvonnefroehlich, line 8: I would be a bit more detailed here. Feel free, to improve my formulation.

```suggestion
can be added via the ``label`` parameter. The timestamp
will always be shown relative to the bottom-left corner
of the figure. By default, the ``offset`` and
``justification`` parameters are set to
``("-54p", "-54p")`` (x, y directions) and ``BL``
(bottom-left), respectively. 
```

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+import pygmt
+
+fig = pygmt.Figure()
+fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
+fig.timestamp()
+# Plot the GMT timestamp logo with a custom label.
```
seisman, line 16: I prefer to have only one timestamp on one figure, so better to plot two separate figures in the example. For the second figure, maybe use a geographic projection like `-JH`?

michaelgrund: Done, thanks!

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,25 @@
...
+import pygmt
+
+fig = pygmt.Figure()
+fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
+fig.timestamp()
+fig.show()
+
+fig = pygmt.Figure()
+fig.coast(region="d", projection="H10c", land="black", water="cornsilk", frame="afg")
+# Plot the GMT timestamp logo with a custom label.
```
seisman, line 23: I think it would be better to also add `font="Times-Bold"` (or any other fonts) to show that users can control the label font and also the `timefmt="%Y-%m-%dT%H:%M"` to show how to control the timestamp format.

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,25 @@
...
+
+import pygmt
+
+fig = pygmt.Figure()
+fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
+fig.timestamp()
+fig.show()
+
+fig = pygmt.Figure()
+fig.coast(region="d", projection="H10c", land="black", water="cornsilk", frame="afg")
```
yvonnefroehlich, line 22: I feel separating the plots within this gallery example similar to, e.g., [Multi-parameter symbols](https://www.pygmt.org/dev/gallery/symbols/multi_parameter_symbols.html) would make the example look overall nicer. 

```suggestion

# sphinx_gallery_thumbnail_number = 1

import pygmt

fig = pygmt.Figure()
fig.basemap(region=[20, 30, -10, 10], projection="X10c/5c")
fig.timestamp()
fig.show()

###############################################################################
# Maybe some short description of the ``font`` and ``timefmt`` parameters.

fig = pygmt.Figure()
fig.coast(region="d", projection="H10c", land="black", water="cornsilk", frame="afg")
```

examples/gallery/embellishments/timestamp.py
```diff
@@ -0,0 +1,35 @@
+"""
+Timestamp
+---------
+The :meth:`pygmt.Figure.timestamp` method can draw the
+GMT timestamp logo on the figure. The timestamp
+will always be shown relative to the bottom-left corner
+of the figure. By default, the ``offset`` and
+``justification`` parameters are set to
+``("-54p", "-54p")`` (x, y directions) and ``"BL"``
+(bottom-left), respectively.
```
seisman, line 10:
```suggestion
(bottom-left), respectively.
```

seisman (approved):
This PR looks good to me except the minor fix above.

yvonnefroehlich (commented):
Looks good to me! Thanks @michaelgrund for considering all the suggestions 🙂.

examples/gallery/embellishments/timestamp.py
yvonnefroehlich: Hm. Based on @michaelgrund's comment https://github.com/GenericMappingTools/pygmt/pull/2395#discussion_r1131261305, I am wondering whether this formulation is strictly speaking also not precise / correct and "figure" should be "map" (or "plot")?

seisman: Perhaps open a separate issue so that we can have more discussions.

yvonnefroehlich: Please see issue #2437.

PR #2125, opened 2022-09-19
Add function to load raster tile maps using contextily

weiji14:
**Description of proposed changes**

New dataset function to load XYZ tiles! Uses [contextily](https://github.com/geopandas/contextily) to retrieve the tiles based on a bounding box.

**Preview** at https://pygmt-dev--2125.org.readthedocs.build/en/2125/api/generated/pygmt.datasets.load_tile_map.html

This is the first part of #2115 that does the tile loading and georeferencing to an `xarray.DataArray`. A follow up PR will be needed to take the output `xarray.DataArray` from `pygmt.datasets.load_tile_map` and plot it using `fig.grdimage` (to be discussed in #2115).

Usage:

```python
import contextily
from pygmt.datasets import load_tile_map

raster = load_tile_map(
    region=[103.60, 104.06, 1.22, 1.49],  # West, East, South, North
    source=contextily.providers.Stamen.TerrainBackground,
    lonlat=True,  # bounding box coordinates are longitude/latitude
)

print(raster)
# <xarray.DataArray (band: 3, y: 1024, x: 1536)>
# array([[[208, 208, 205, ..., 210, 213, 213],
#         [208, 208, 205, ..., 213, 213, 213],
#         [205, 205, 204, ..., 213, 213, 213],
#         ...,
#         [153, 153, 153, ..., 153, 153, 153],
#         [153, 153, 153, ..., 153, 153, 153],
#         [153, 153, 153, ..., 153, 153, 153]],
# 
#        [[214, 214, 213, ..., 217, 218, 218],
#         [214, 214, 213, ..., 219, 220, 219],
#         [213, 213, 212, ..., 219, 220, 219],
#         ...,
#         [179, 179, 179, ..., 179, 179, 179],
#         [179, 179, 179, ..., 179, 179, 179],
#         [179, 179, 179, ..., 179, 179, 179]],
# 
#        [[175, 175, 174, ..., 186, 185, 185],
#         [175, 175, 174, ..., 188, 188, 188],
#         [174, 174, 171, ..., 188, 188, 188],
#         ...,
#         [204, 204, 204, ..., 204, 204, 204],
#         [204, 204, 204, ..., 204, 204, 204],
#         [204, 204, 204, ..., 204, 204, 204]]], dtype=uint8)
# Coordinates:
#   * band     (band) int64 0 1 2
#   * y        (y) float64 1.49 1.49 1.489 1.489 1.489 ... 1.221 1.221 1.22 1.22
[...]

PR #2393, opened 2023-03-03
Fix bug with different band dtype from load_tile_map by casting to uint8

weiji14:

The band list `[0, 1, 2]` was being converted to int32 on Windows + NumPy 1.24, but is int64 on other platforms. Casting the band coordinate explicitly to `uint8` for consistency.

Patches bug in #2125 as reported in https://github.com/GenericMappingTools/pygmt/pull/2125#issuecomment-1453006453

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -94,17 +94,18 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
     >>> import contextily
     >>> from pygmt.datasets import load_tile_map
     >>> raster = load_tile_map(
-    ...     region=[103.60, 104.06, 1.22, 1.49],  # West, East, South, North
+    ...     region=[-180.0, 180.0, -90.0, 0.0],  # West, East, South, North
+    ...     zoom=1,  # less detailed zoom level
     ...     source=contextily.providers.Stamen.TerrainBackground,
     ...     lonlat=True,  # bounding box coordinates are longitude/latitude
     ... )
     >>> raster.sizes
-    Frozen({'band': 3, 'y': 1024, 'x': 1536})
+    Frozen({'band': 3, 'y': 256, 'x': 512})
     >>> raster.coords
     Coordinates:
       * band     (band) uint8 0 1 2
-      * y        (y) float64 1.663e+05 1.663e+05 1.663e+05 ... 1.272e+05 ...
-      * x        (x) float64 1.153e+07 1.153e+07 1.153e+07 ... 1.158e+07 ...
+      * y        (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 -2.004e+07
+      * x        (x) float64 -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
     """
     # pylint: disable=too-many-locals
     if contextily is None:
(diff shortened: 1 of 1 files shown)


pygmt/datasets/tile_map.py
```diff
@@ -102,7 +102,7 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
     Frozen({'band': 3, 'y': 1024, 'x': 1536})
     >>> raster.coords
     Coordinates:
-      * band     (band) int64 0 1 2
+      * band     (band) uint8 0 1 2
       * y        (y) float64 1.663e+05 1.663e+05 1.663e+05 ... 1.272e+05 ...
       * x        (x) float64 1.153e+07 1.153e+07 1.153e+07 ... 1.158e+07 ...
```
weiji14, line 107: From https://github.com/GenericMappingTools/pygmt/pull/2125#issuecomment-1454181058, I remembered why I didn't print out the full y and x coordinates here. It's because that would make the docstring too long (>79 characters), so I added `...`  which is supported by https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS.
```suggestion
      * y        (y) float64 1.663e+05 1.663e+05 1.663e+05 ... 1.272e+05 1.272e+05
      * x        (x) float64 1.153e+07 1.153e+07 1.153e+07 ... 1.158e+07 1.158e+07
```

weiji14:
All the tests pass now, but I just noticed at https://github.com/GenericMappingTools/pygmt/actions/runs/4327743401/jobs/7556724608#step:11:875 that the doctest is taking 6.76 seconds (the 2nd slowest test). Should we update the region to return a smaller tile instead?

seisman:
> All the tests pass now, but I just noticed at https://github.com/GenericMappingTools/pygmt/actions/runs/4327743401/jobs/7556724608#step:11:875 that the doctest is taking 6.76 seconds (the 2nd slowest test). Should we update the region and/or set a zoom level to return a smaller tile instead?

Yes, we need a smaller tile.

pygmt/datasets/tile_map.py
weiji14: > Perhaps you need to add the `# doctest: +ELLIPSIS` directive after line 103.

I thought so too, but using `...` by itself seems to work? Like, the tests still all pass. I think it was just the int32/int64 diff that was the problem, and pytest still reports the full diff.

Anyways, I've changed the bounding box region in ea9e7dceef00ef1b2c68630a8d59190a33c1037a and the y/x coordinates fit under 79 characters now, so doesn't really matter.

weiji14:
> > All the tests pass now, but I just noticed at https://github.com/GenericMappingTools/pygmt/actions/runs/4327743401/jobs/7556724608#step:11:875 that the doctest is taking 6.76 seconds (the 2nd slowest test). Should we update the region and/or set a zoom level to return a smaller tile instead?
> 
> Yes, we need a smaller tile.

Ok, I've modified the bounding box region and set things to zoom level 1 in ea9e7dceef00ef1b2c68630a8d59190a33c1037a, which should return only 2 tiles of size 256x256. The doctest now takes 0.60 seconds according to https://github.com/GenericMappingTools/pygmt/actions/runs/4330035373/jobs/7560981253#step:11:1116

seisman (approved):

Issue #153, opened 2018-03-30
Tutorial notebook for showcasing projections and plot region arguments


leouieda:
The `first-steps` tutorial is a good length introduction to the basics of the `gmt.Figure` class and what is different from classic GMT.

It would be great to have a separate tutorial notebook with an in-depth look at how GMT handles the region argument (showing the ISO country codes) and how projections work (with coastline plots showing each one). Perhaps including a short table of all projections.

This notebook would live in `doc/tutorials` and it's run automatically by the documentation build process.

willschlitzer:
@leouieda Are you still hoping to get a tutorial like this? I think documenting the projections in a tutorial may be a little redundant now that there is a projection gallery.

seisman:
> I think documenting the projections in a tutorial may be a little redundant now that there is a projection gallery.

I agree.

But I think we still need a tutorial showing how to specify a plot region. Some possible options are:

- Give either `g` or `d` for global region
- A list `[xmin, xmax, ymin, ymax]`
- A string `xmin/xmax/ymin/ymax`
- ISO country code (e.g., `US`) and some modifiers (e.g., `US+r2`)
- and more in the cookbook (https://docs.generic-mapping-tools.org/latest/cookbook/options.html#data-domain-or-map-region-the-r-option).

PR #2394, opened 2023-03-04
Add Figure.tilemap to plot XYZ tile maps

### LINKED ISSUE #2115: Integration with contextily to plot xyz basemaps

To allow for a greater variety of basemaps from e.g. OpenStreetMap, Stamen, MapBox, etc, it would be nice to integrate with [`contextily`](https://github.com/geopandas/contextily) (part of the geopandas organization). This is motivated by https://forum.generic-mapping-tools.org/t/pygmt-vs-plotly/3250/10.

image

On the implementation, here are 2 ideas on how to integrate `contextily` basemaps with `pygmt`.

1. Have an additional option under `fig.basemap` which allows selection of the basemap (e.g. OpenStreetMap, Stamen, Mapbox, etc) (#2394, to be called `fig.tilemap`)
2. Create a new `load_basemap` function under [`pygmt/datasets`](https://github.com/GenericMappingTools/pygmt/tree/v0.7.0/pygmt/datasets) that would load the image (`np.ndarray`) and georeference it (to an `xarray.DataArray`), and then the user plots it using `fig.grdimage`. (done at #2125, called `load_tile_map`)
3. Maybe there's another more Pythonic way? Comment down below!

Either way, these would require:

1. Wrapping around the `contextily.bounds2img` function at https://contextily.readthedocs.io/en/latest/reference.html#contextily.bounds2img. This returns a `numpy.ndarray` and a geographic extent.
2. The `numpy.ndarray` needs to be georeferenced into an `xarray.DataArray`. I have some recent code at https://zen3geo.readthedocs.io/en/v0.4.0/object-detection-boxes.html#georeference-image-using-rioxarray to do that
4. The `xarray.DataArray` will need to be plotted on the PyGMT map

**Are you willing to help implement and maintain this feature?** Yes, but discuss about the implementation first

weiji14:

New `tilemap` function for plotting XYZ tile maps! This is essentially a convenience wrapper around `pygmt.datasets.load_tile_map` and `pygmt.Figure.grdimage`.

**Preview** at https://pygmt-dev--2394.org.readthedocs.build/en/2394/api/generated/pygmt.Figure.tilemap.html

This is the second part of #2115 that takes the output xarray.DataArray from `pygmt.datasets.load_tile_map` (added in #2125) and plots it using `fig.grdimage`.

Usage:

```python
import contextily
import pygmt

fig = pygmt.Figure()
fig.tilemap(
    region=[-157.6, -157.1, 1.68, 2.08],  # West, East, South, North
    source=contextily.providers.Stamen.Watercolor,
    frame=True,
)
fig.show()
```
produces this watercolor map of Kiribati:

Kiribati_watercolor_map

Notes:
- `rioxarray` will be added as an optional dependency, as it is used for the `xarray.DataArray` to GeoTIFF conversion
- ~The implementation of `tilemap` is a little tricky due to cyclic import issues, please suggest if there are better ways~ fixed with https://github.com/GenericMappingTools/pygmt/pull/2394#discussion_r1125417904
- By default, a given `region` with lonlat coordinates will produce a map with lonlat (OGC:WGS84) coordinates that are distorted from the original Spherical Mercator (EPSG:3857) tiles, see https://github.com/GenericMappingTools/pygmt/pull/2394#issuecomment-1475939603
- The `region` parameter only accepts a list like `[lonmin, lonmax, latmin, latmax]`, which is different from other GMT/PyGMT functions that accept more types of `region` values, see https://github.com/GenericMappingTools/pygmt/pull/2394#discussion_r1125203009

Fixes #2115

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -62,6 +62,7 @@ Plotting raster data
     Figure.grdimage
     Figure.grdview
     Figure.image
+    Figure.tilemap
 
 Configuring layout
 ~~~~~~~~~~~~~~~~~~
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -108,6 +108,7 @@ The following are optional dependencies:
 * `IPython <https://ipython.org>`__: For embedding the figures in Jupyter notebooks (recommended).
 * `Contextily <https://contextily.readthedocs.io>`__: For retrieving tile maps from the internet.
 * `GeoPandas <https://geopandas.org>`__: For using and plotting GeoDataFrame objects.
+* `RioXarray <https://https://corteva.github.io/rioxarray>`__: For saving multi-band rasters to GeoTIFFs.
 
 Installing GMT and other dependencies
 -------------------------------------
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -108,7 +108,7 @@ The following are optional dependencies:
 * `IPython <https://ipython.org>`__: For embedding the figures in Jupyter notebooks (recommended).
 * `Contextily <https://contextily.readthedocs.io>`__: For retrieving tile maps from the internet.
 * `GeoPandas <https://geopandas.org>`__: For using and plotting GeoDataFrame objects.
-* `RioXarray <https://https://corteva.github.io/rioxarray>`__: For saving multi-band rasters to GeoTIFFs.
+* `RioXarray <https://corteva.github.io/rioxarray>`__: For saving multi-band rasters to GeoTIFFs.
 
 Installing GMT and other dependencies
 -------------------------------------
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -113,7 +113,7 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
         raise ModuleNotFoundError(
             "Package `contextily` is required to be installed to use this function. "
             "Please use `pip install contextily` or "
-            "`conda install -c conda-forge contextily` "
+            "`mamba install -c conda-forge contextily` "
             "to install the package."
         )
 
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -103,9 +103,10 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
     Frozen({'band': 3, 'y': 256, 'x': 512})
     >>> raster.coords
     Coordinates:
-      * band     (band) uint8 0 1 2
-      * y        (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 -2.004e+07
-      * x        (x) float64 -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
+      * band         (band) uint8 0 1 2
+      * y            (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 ...
+      * x            (x) float64 -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
+        spatial_ref  int64 0
     """
     # pylint: disable=too-many-locals
     if contextily is None:
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -106,7 +106,6 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
       * band         (band) uint8 0 1 2
       * y            (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 ...
       * x            (x) float64 -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
-        spatial_ref  int64 0
     """
     # pylint: disable=too-many-locals
     if contextily is None:
@@ -148,6 +147,6 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
 
     # If rioxarray is installed, set the coordinate reference system
     if hasattr(dataarray, "rio"):
-        dataarray = dataarray.rio.write_crs(input_crs="EPSG:3857")
+        dataarray = dataarray.rio.set_crs(input_crs="EPSG:3857")
 
     return dataarray
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -120,17 +120,16 @@ def tilemap(
             "to install the package."
         )
 
+    raster = load_tile_map(
+        region=region,
+        zoom=zoom,
+        source=source,
+        lonlat=lonlat,
+        wait=wait,
+        max_retries=max_retries,
+    )
     with GMTTempFile(suffix=".tif") as tmpfile:
-        raster = load_tile_map(
-            region=region,
-            zoom=zoom,
-            source=source,
-            lonlat=lonlat,
-            wait=wait,
-            max_retries=max_retries,
-        )
         raster.rio.to_raster(raster_path=tmpfile.name)
-
         with Session() as lib:
             lib.call_module(
                 module="grdimage", args=build_arg_string(kwargs, infile=tmpfile.name)
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -142,9 +142,9 @@ def tilemap(
         raster = raster.rio.reproject(dst_crs="OGC:CRS84")
         raster.gmt.gtype = 1  # set to geographic type
 
-    # Only set region if no_clip is False, so that plot is clipped to exact
-    # bounding box region
-    if not kwargs.get("N"):
+    # Only set region if no_clip is None or False, so that plot is clipped to
+    # exact bounding box region
+    if kwargs.get("N") in [None, False]:
         kwargs["R"] = "/".join(str(coordinate) for coordinate in region)
 
     with GMTTempFile(suffix=".tif") as tmpfile:
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -115,7 +115,7 @@ def tilemap(
         raise ModuleNotFoundError(
             "Package `rioxarray` is required to be installed to use this function. "
             "Please use `pip install rioxarray` or "
-            "`conda install -c conda-forge rioxarray` "
+            "`mamba install -c conda-forge rioxarray` "
             "to install the package."
         )
 
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -43,12 +43,12 @@ def tilemap(
     :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
     the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
 
-    **Note**: By default, standard web map tiles served in a Web Mercator
+    **Note**: By default, standard web map tiles served in a Spherical Mercator
     (EPSG:3857) Cartesian format will be reprojected to a geographic coordinate
     reference system (OGC:WGS84) and plotted with longitude/latitude bounds
     when ``lonlat=True``. If reprojection is not desired, please set
-    ``lonlat=False`` and provide Web Mercator (EPSG:3857) coordinates to the
-    ``region`` parameter.
+    ``lonlat=False`` and provide Spherical Mercator (EPSG:3857) coordinates to
+    the ``region`` parameter.
 
     {aliases}
 
@@ -136,8 +136,8 @@ def tilemap(
         max_retries=max_retries,
     )
 
-    # Reproject raster from Web Mercator (EPSG:3857) to lonlat (OGC:CRS84) if
-    # bounding box region was provided in lonlat
+    # Reproject raster from Spherical Mercator (EPSG:3857) to
+    # lonlat (OGC:CRS84) if bounding box region was provided in lonlat
     if lonlat and raster.rio.crs == "EPSG:3857":
         raster = raster.rio.reproject(dst_crs="OGC:CRS84")
         raster.gmt.gtype = 1  # set to geographic type
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -43,6 +43,13 @@ def tilemap(
     :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
     the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
 
+    **Note**: By default, standard web map tiles served in a Web Mercator
+    (EPSG:3857) Cartesian format will be reprojected to a geographic coordinate
+    reference system (OGC:WGS84) and plotted with longitude/latitude bounds
+    when ``lonlat=True``. If reprojection is not desired, please set
+    ``lonlat=False`` and provide Web Mercator (EPSG:3857) coordinates to the
+    ``region`` parameter.
+
     {aliases}
 
     Parameters
@@ -128,6 +135,13 @@ def tilemap(
         wait=wait,
         max_retries=max_retries,
     )
+
+    # Reproject raster from Web Mercator (EPSG:3857) to lonlat (OGC:CRS84) if
+    # bounding box region was provided in lonlat
+    if lonlat and raster.rio.crs == "EPSG:3857":
+        raster = raster.rio.reproject(dst_crs="OGC:CRS84")
+        raster.gmt.gtype = 1  # set to geographic type
+
     with GMTTempFile(suffix=".tif") as tmpfile:
         raster.rio.to_raster(raster_path=tmpfile.name)
         with Session() as lib:
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -1,8 +1,8 @@
 """
 tilemap - Plot XYZ tile maps.
 """
-
 from pygmt.clib import Session
+from pygmt.datasets.tile_map import load_tile_map
 from pygmt.helpers import (
     GMTTempFile,
     build_arg_string,
@@ -116,8 +116,6 @@ def tilemap(
         :doc:`install instructions for rioxarray <rioxarray:installation>`,
         (e.g. via ``pip install rioxarray``) before using this function.
     """
-    from pygmt.datasets import load_tile_map  # pylint: disable=import-outside-toplevel
-
     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
 
     if rioxarray is None:
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -112,7 +112,7 @@ def tilemap(
 
     Raises
     ------
-    ModuleNotFoundError
+    ImportError
         If ``rioxarray`` is not installed. Follow
         :doc:`install instructions for rioxarray <rioxarray:installation>`,
         (e.g. via ``pip install rioxarray``) before using this function.
@@ -120,7 +120,7 @@ def tilemap(
     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
 
     if rioxarray is None:
-        raise ModuleNotFoundError(
+        raise ImportError(
             "Package `rioxarray` is required to be installed to use this function. "
             "Please use `pip install rioxarray` or "
             "`mamba install -c conda-forge rioxarray` "
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -39,7 +39,7 @@ def tilemap(
     r"""
     Plots an XYZ tile map.
 
-    This function loads XYZ tile maps from a tile server or local file using
+    This method loads XYZ tile maps from a tile server or local file using
     :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
     the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
 
--- a/pygmt/src/tilemap.py
+++ b/pygmt/src/tilemap.py
@@ -1,0 +1,145 @@
+"""
+tilemap - Plot XYZ tile maps.
+"""
+
+from pygmt.clib import Session
+from pygmt.helpers import (
+    GMTTempFile,
+    build_arg_string,
+    fmt_docstring,
+    kwargs_to_strings,
+    use_alias,
+)
+
+try:
+    import rioxarray
+except ImportError:
+    rioxarray = None
+
+
+@fmt_docstring
+@use_alias(
+    A="img_out",
+    B="frame",
+    C="cmap",
+    D="img_in",
+    E="dpi",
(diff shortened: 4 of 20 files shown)


doc/install.rst
```diff
@@ -108,6 +108,7 @@ The following are optional dependencies:
 * `IPython <https://ipython.org>`__: For embedding the figures in Jupyter notebooks (recommended).
 * `Contextily <https://contextily.readthedocs.io>`__: For retrieving tile maps from the internet.
 * `GeoPandas <https://geopandas.org>`__: For using and plotting GeoDataFrame objects.
+* `RioXarray <https://https://corteva.github.io/rioxarray>`__: For saving multi-band rasters to GeoTIFFs.
```
weiji14, line 111: Probably should reword this description?

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,145 @@
...
+    c="panel",
+    f="coltypes",
+    p="perspective",
+    t="transparency",
+    x="cores",
+)
+@kwargs_to_strings(c="sequence_comma", p="sequence")  # R="sequence",
+def tilemap(
+    self, region, zoom="auto", source=None, lonlat=True, wait=0, max_retries=2, **kwargs
+):
```
weiji14, line 45: Currently `region` needs to be a list to work with `load_tile_map`, but I'm debating on whether to allow strings like `0/1/2/3` or other styles supported by GMT `-R` as mentioned in https://github.com/GenericMappingTools/pygmt/issues/153#issuecomment-742896187 and https://docs.generic-mapping-tools.org/6.4/cookbook/options.html#data-domain-or-map-region-the-r-option

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,145 @@
...
+        Extra keyword arguments to pass to :meth:`pygmt.Figure.grdimage`.
+
+    Raises
+    ------
+    ModuleNotFoundError
+        If ``rioxarray`` is not installed. Follow
+        :doc:`install instructions for rioxarray <rioxarray:installation>`,
+        (e.g. via ``pip install rioxarray``) before using this function.
+    """
+    from pygmt.datasets import load_tile_map  # pylint: disable=import-outside-toplevel
```
weiji14, line 119: Workaround for cyclic import issue. Maybe there's a better way to do this?

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,145 @@
...
+            lonlat=lonlat,
+            wait=wait,
+            max_retries=max_retries,
+        )
+        raster.rio.to_raster(raster_path=tmpfile.name)
+
+        with Session() as lib:
+            lib.call_module(
+                module="grdimage", args=build_arg_string(kwargs, infile=tmpfile.name)
+            )
```
weiji14, line 145: Note that this is not the full implementation of PyGMT's `fig.grdimage`, which is actually a bit more complicated to support shading (`-I`) with an `xarray.DataArray` grid

https://github.com/GenericMappingTools/pygmt/blob/8f31706fdfdc5d3cf1b15e77fc64db304031b53e/pygmt/src/grdimage.py#L179-L192

Question is: should we support shading of the XYZ tiles with `xarray.DataArray` grids? Or is there a way to use the `grdimage` implementation above directly without code duplication?

doc/install.rst
seisman, line 111:
```suggestion
* `RioXarray <https://corteva.github.io/rioxarray>`__: For saving multi-band rasters to GeoTIFFs.
```

pygmt/src/tilemap.py
seisman: > other styles supported by GMT `-R` as mentioned in [#153 (comment)](https://github.com/GenericMappingTools/pygmt/issues/153#issuecomment-742896187) and https://docs.generic-mapping-tools.org/6.4/cookbook/options.html#data-domain-or-map-region-the-r-option

Other styles are more tricky to support. For example, how can we know the boundingbox for `region="US"`?

pygmt/src/tilemap.py
seisman: Using `from pygmt.datasets.tile_map import load_tile_map` at the top may be working.

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,145 @@
...
+    Q="nan_transparent",
+    # R="region",
+    V="verbose",
+    n="interpolation",
+    c="panel",
+    f="coltypes",
+    p="perspective",
+    t="transparency",
+    x="cores",
+)
```
seisman, line 41: I'm unsure about these aliases. I would like to keep a few parameters (e.g., `frame`, `projection`, `verbose`, `panel`, `transparency`. Does `cmap` work for images?) in the initial implementation and add more later.

pygmt/src/tilemap.py
weiji14: Yes, that works! Done in c51d9c692e522b3eae3e1f3eaa0bb13069871d96

doc/install.rst
weiji14: Good catch, typo fixed in fa8641b12ff77fed86653426e78b75157dcdf223.

pygmt/src/tilemap.py
weiji14: > Other styles are more tricky to support.

Yeah, I was really just thinking about `0/1/2/3`, and maybe `np.array([0, 1, 2, 3])`. Usually users would set a `region` variable near the top of their code, and they might find it strange why their `region` that works for `fig.basemap` doesn't work for `fig.tilemap`.

> For example, how can we know the boundingbox for `region="US"`?

If only there was an API in GMT to return the bounding box for a given region :wink: [Global regions](https://www.pygmt.org/v0.8.0/tutorials/basics/regions.html#global-regions) like `region="g"` and `region="d"` might be easy to support, but ISO code regions (https://www.pygmt.org/v0.8.0/tutorials/basics/regions.html#iso-code) like `region="JP"` and `region=JP+r3" will be a bit trickier.

pygmt/src/tilemap.py
weiji14: > I'm unsure about these aliases. I would like to keep a few parameters (e.g., `frame`, `projection`, `verbose`, `panel`, `transparency`, ...) in the initial implementation and add more later.

Fair enough, let me take a look at what's actually relevant to RGB images and trim down the list of aliases.

> Does `cmap` work for images?) 

I just tried using `fig.tilemap(region=[-180.0, 180.0, -90, 90], zoom=0, frame="afg", cmap="turbo")`, and it doesn't error, but the CPT isn't used either. Maybe best to remove `cmap` then.

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,143 @@
...
+    C="cmap",
+    D="img_in",
+    E="dpi",
+    G="bit_color",
+    I="shading",
+    J="projection",
+    M="monochrome",
+    N="no_clip",
+    Q="nan_transparent",
+    # R="region",
```
weiji14, line 33: Was playing with the `zoom` parameter just now, and it seems that setting lower zoom levels (e.g. 0) with actually cause the area covered by `contextily.bounds2img` to be bigger than the original `region`. This might come as a surprise for users, so maybe we should set `no_clip` to True by default? This might require a bit of extra work to get correct.

Alternatively, we could also do the clipping in `pygmt.datasets.load_tile_map`.

doc/api/index.rst
```diff
@@ -62,6 +62,7 @@ Plotting raster data
     Figure.grdimage
     Figure.grdview
     Figure.image
+    Figure.tilemap
```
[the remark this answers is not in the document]
yvonnefroehlich: I prefer leaving it as is in the "Plotting raster data" category.

pygmt/src/tilemap.py
michaelgrund: Currently I would say no to shading.

pygmt/src/tilemap.py
weiji14: Ok, I've removed some aliases in c3156097bb512386b7f861b9e7f2b45a185831db.

pyproject.toml
```diff
@@ -44,7 +44,8 @@ dynamic = ["version"]
 all = [
     "contextily",
     "geopandas",
-    "ipython"
+    "ipython",
+    "rioxarray"
```
seisman, line 48: Always add a trailing comma following PEP8 (https://peps.python.org/pep-0008/#when-to-use-trailing-commas), although it's not a Python file.
```suggestion
    "rioxarray",
```

pygmt/src/tilemap.py
seisman: > If only there was an API in GMT to return the bounding box for a given region  [Global regions](https://www.pygmt.org/v0.8.0/tutorials/basics/regions.html#global-regions) like `region="g"` and `region="d"` might be easy to support, but ISO code regions (https://www.pygmt.org/v0.8.0/tutorials/basics/regions.html#iso-code) like `region="JP"` and `region=JP+r3" will be a bit trickier.

Open a upstream feature request in the GMT repository?

weiji14: Hmm, or do you think we can use `GMT_Extract_Region` that's wrapped here:

https://github.com/GenericMappingTools/pygmt/blob/0b5a69b06481335bd4aa7a09f44b779e01e1ce43/pygmt/clib/session.py#L1500-L1571

Downside is that it requires a figure instance to be set up already.

seisman: I don't think it will work. Actually, a figure instance is not enough. We have to plot something to generate the hidden PostScript file before calling `extract_region`.

```
In [1]: import pygmt

In [2]: fig = pygmt.Figure()

In [3]: fig.region
pygmt-session [ERROR]: No hidden PS file found
GMTCLibError                              Traceback (most recent call last)
Cell In[3], line 1
----> 1 fig.region

File ~/OSS/gmt/pygmt/pygmt/figure.py:126, in Figure.region(self)
    124 self._activate_figure()
    125 with Session() as lib:
--> 126     wesn = lib.extract_region()
    127 return wesn

File ~/OSS/gmt/pygmt/pygmt/clib/session.py:1570, in Session.extract_region(self)
   1568 status = c_extract_region(self.session_pointer, None, wesn_pointer)
   1569 if status != 0:
-> 1570     raise GMTCLibError("Failed to extract region from current figure.")
   1571 return wesn
```

pyproject.toml
seisman, line 48: Adding a trailing comma following [PEP 8](https://peps.python.org/pep-0008/#when-to-use-trailing-commas), although it's TOML not Python.
```suggestion
    "rioxarray",
```

seisman:
> Usage:
> 
> ```python
> import contextily
> import pygmt
> 
> fig = pygmt.Figure()
> fig.tilemap(
>     region=[-157.6, -157.1, 1.68, 2.08],  # West, East, South, North
>     source=contextily.providers.Stamen.Watercolor,
>     frame=True,
> )
> fig.show()
> ```
> 
> produces this watercolor map of Kiribati:
> 
> Kiribati_watercolor_map

Do we need to convert the map tiles from EPSG:3857 to longitude/latitude?

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,136 @@
...
+    with GMTTempFile(suffix=".tif") as tmpfile:
+        raster = load_tile_map(
+            region=region,
+            zoom=zoom,
+            source=source,
+            lonlat=lonlat,
+            wait=wait,
+            max_retries=max_retries,
+        )
+        raster.rio.to_raster(raster_path=tmpfile.name)
```
seisman, line 131: Move `load_tile_map` out of the `GMTTempFile` context:
```suggestion
    raster = load_tile_map(
        region=region,
        zoom=zoom,
        source=source,
        lonlat=lonlat,
        wait=wait,
        max_retries=max_retries,
    )

    with GMTTempFile(suffix=".tif") as tmpfile:
        raster.rio.to_raster(raster_path=tmpfile.name)
```

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,136 @@
...
+)
+@kwargs_to_strings(c="sequence_comma", p="sequence")  # R="sequence",
+def tilemap(
+    self, region, zoom="auto", source=None, lonlat=True, wait=0, max_retries=2, **kwargs
+):
+    r"""
+    Plots an XYZ tile map.
+
+    This is a wrapper around :func:`pygmt.datasets.load_tile_map` and
+    :meth:`pygmt.Figure.grdimage`.
```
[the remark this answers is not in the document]
weiji14: Reworded in 7efdbfda7db75c4705b95585698a5c571753317b, let me know if it could be improved a bit more!

weiji14:
> Do we need to convert the map tiles from EPSG:3857 to longitude/latitude?

Yes, if `lonlat=True` (default), then I think the output map should be in lon/lat as well to make it intuitive. I've implemented this in fde94ad448bf61bb4e0d7f0f1a03f71fd8967fd1. The issue with this is that the pixels will be reprojected and distorted by default, which might not be a problem for tile maps like Stamen Terrain, but can be problematic when there are words on the tile map (e.g. for OpenStreetMap).

pygmt/src/tilemap.py
weiji14: Added commit 17d07341eea751bc392cad3225c8a72b8ee87559 to ensure that:
- The plot is clipped to the bounding box region when `no_clip=False` (default)
- The plot extends to areas outside of the bounding box region when `no_clip=True`

pygmt/src/tilemap.py
weiji14: > I don't think it will work. Actually, a figure instance is not enough. We have to plot something to generate the hidden PostScript file before calling `extract_region`.

Hmm yeah, we would need to maybe do a `fig.basemap` and then `fig.grdimage`, but that's a bit of a hacky solution.

Let's just keep this PR simple and support only a list of `[xmin, xmax, ymin, ymax]` coordinates for now. There can be a follow up PR to support other `region` arguments (most likely after PyGMT v0.9.0) if there's interest in this 'enhancement'.

pygmt/tests/test_tilemap.py
```diff
@@ -0,0 +1,58 @@
...
+    Create a tilemap plot using longitude/latitude coordinates (OGC:WGS84),
+    centred on the international date line.
+    """
+    fig = Figure()
+    fig.tilemap(
+        region=[-180.0, 180.0, -90, 90],
+        zoom=0,
+        frame="afg",
+        projection="R180/5c",
+        verbose=True,
```
seisman, line 38: Remove 'verbose=True'?
```suggestion
```

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+    self, region, zoom="auto", source=None, lonlat=True, wait=0, max_retries=2, **kwargs
+):
+    r"""
+    Plots an XYZ tile map.
+
+    This function loads XYZ tile maps from a tile server or local file using
+    :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
+    the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
+
+    **Note**: By default, standard web map tiles served in a Web Mercator
```
michaelgrund, line 46:
```suggestion
    **Note**: By default, standard web map tiles served in a Spherical Mercator
```
Although the terms are meaning the same, I think we should only use one definition throughout the whole documentation to avoid confusion. I would prefer Spherical Mercator instead of Web Mercator.

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+    (EPSG:3857) Cartesian format will be reprojected to a geographic coordinate
+    reference system (OGC:WGS84) and plotted with longitude/latitude bounds
+    when ``lonlat=True``. If reprojection is not desired, please set
+    ``lonlat=False`` and provide Web Mercator (EPSG:3857) coordinates to the
```
michaelgrund, line 50:
```suggestion
    ``lonlat=False`` and provide Spherical Mercator (EPSG:3857) coordinates to the
```

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+    raster = load_tile_map(
+        region=region,
+        zoom=zoom,
+        source=source,
+        lonlat=lonlat,
+        wait=wait,
+        max_retries=max_retries,
+    )
+
+    # Reproject raster from Web Mercator (EPSG:3857) to lonlat (OGC:CRS84) if
```
michaelgrund, line 139:
```suggestion
    # Reproject raster from Spherical Mercator (EPSG:3857) to lonlat (OGC:CRS84) if
```

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+    # bounding box region was provided in lonlat
+    if lonlat and raster.rio.crs == "EPSG:3857":
+        raster = raster.rio.reproject(dst_crs="OGC:CRS84")
+        raster.gmt.gtype = 1  # set to geographic type
```
seisman, line 143: Setting `raster.gmt.gtype = 1` is useless, because `raster.rio.to_raster` doesn't understand it, right?

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Only set region if no_clip is False, so that plot is clipped to exact
+    # bounding box region
+    if not kwargs.get("N"):
+        kwargs["R"] = "/".join(str(coordinate) for coordinate in region)
```
seisman, line 148: When `no_clip` is not given, it defaults to `None` and `not kwargs.get("N")` is True, which means the plot is clipped. I find the code is a little difficult to understand. Do you think the following changes make it better?
```suggestion
    # Only set region if no_clip is False or None, so that plot is clipped to exact
    # bounding box region
    if kwargs.get("N") is not True:
        kwargs["R"] = "/".join(str(coordinate) for coordinate in region)
```

pygmt/src/tilemap.py
weiji14: Agree, and https://epsg.io/3857 uses Spherical Mercator too.

pygmt/src/tilemap.py
weiji14: Yeah, double negatives are confusing. I went with `kwargs.get("N") in [None, False]` at 125574ee52f91624cb406769ee8b45297ebdb5f5.

seisman:
@michaelgrund and @yvonnefroehlich It would be great if you can give this PR a final review before we merge it.

michaelgrund (approved):
Looks good! Thanks again for working on that @weiji14 😉

pygmt/src/tilemap.py
```diff
@@ -0,0 +1,155 @@
...
+@kwargs_to_strings(c="sequence_comma", p="sequence")  # R="sequence",
+def tilemap(
+    self, region, zoom="auto", source=None, lonlat=True, wait=0, max_retries=2, **kwargs
+):
+    r"""
+    Plots an XYZ tile map.
+
+    This function loads XYZ tile maps from a tile server or local file using
+    :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
+    the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
```
yvonnefroehlich, line 44: I am wondering whether this is strictly speaking not a "function", but a "method" since it seems to be part of the `pygmt.Figure` class?

```suggestion
    This method loads XYZ tile maps from a tile server or local file using
    :func:`pygmt.datasets.load_tile_map` into a georeferenced form, and plots
    the tiles as a basemap or overlay using :meth:`pygmt.Figure.grdimage`.
```

yvonnefroehlich (approved):
I am afriad I do not have the knowledge to review this PR reliable, but what I understand make sense to me.

seisman:
@weiji14 OK to merge?

weiji14:
> @weiji14 OK to merge?

Yes I'll merge it today. Let me incorporate https://github.com/GenericMappingTools/pygmt/pull/2394#discussion_r1148435150 first.

pygmt/src/tilemap.py
weiji14: Good catch, thanks!

weiji14:
Hold on, let me fix this dtype error on Windows / Python 3.11 + NumPy 1.24 at https://github.com/GenericMappingTools/pygmt/actions/runs/4527632070/jobs/7973665357?pr=2394#step:11:737. Looks similar to https://github.com/GenericMappingTools/pygmt/pull/2393:

```python-traceback
_______________ [doctest] pygmt.datasets.tile_map.load_tile_map _______________
095     >>> from pygmt.datasets import load_tile_map
096     >>> raster = load_tile_map(
097     ...     region=[-180.0, 180.0, -90.0, 0.0],  # West, East, South, North
098     ...     zoom=1,  # less detailed zoom level
099     ...     source=contextily.providers.Stamen.TerrainBackground,
100     ...     lonlat=True,  # bounding box coordinates are longitude/latitude
101     ... )
102     >>> raster.sizes
103     Frozen({'band': 3, 'y': 256, 'x': 512})
104     >>> raster.coords
Differences (unified diff with -expected +actual):
    @@ -1,5 +1,5 @@
     Coordinates:
       * band         (band) uint8 0 1 2
    -  * y            (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 ...
    +  * y            (y) float64 -7.081e-10 -7.858e+04 ... -1.996e+07 -2.004e+07
       * x            (x) float64 -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
    -    spatial_ref  int64 0
    +    spatial_ref  int32 0

D:\a\pygmt\pygmt\pygmt\datasets\tile_map.py:104: DocTestFailure
============================== warnings summary ===============================
```

pygmt/datasets/tile_map.py
```diff
@@ -147,6 +147,6 @@ def load_tile_map(region, zoom="auto", source=None, lonlat=True, wait=0, max_ret
 
     # If rioxarray is installed, set the coordinate reference system
     if hasattr(dataarray, "rio"):
-        dataarray = dataarray.rio.write_crs(input_crs="EPSG:3857")
+        dataarray = dataarray.rio.set_crs(input_crs="EPSG:3857")
```
weiji14, line 150: Changed from using [`.rio.write_crs`](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.rioxarray.XRasterBase.write_crs) to [`.rio.set_crs`](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.rioxarray.XRasterBase.set_crs) to prevent the CF coordinate `spatial_ref` from being written to the output `xarray.DataArray`. This fixes the int32 on Windows bu int64 on Linux/macOS issue mentioned in https://github.com/GenericMappingTools/pygmt/pull/2394#issuecomment-1484348012.

If this looks ok (and the tests pass), I'll proceed to merge in this PR.

seisman: Looks good.

Issue #1819, opened 2022-03-15
Passing xarray.DataArray grids through data processing functions should preserve dimension names


weiji14:
**Description of the desired feature**

Currently, the dimension names of the `xarray.DataArray` grid outputs appear to be hardcoded to 'x', 'y' and 'z', instead of using the original dimension names of the input `xarray.DataArray`. To illustrate

```python
grid = xr.DataArray(
    data=[
        [0, 1, 2],
        [3, 4, 5],
        [6, 7, 8],
    ],
    coords=dict(height=[1.5, 2.5, 3.5], width=[4.5, 5.5, 6.5]),
    dims=["height", "width"],
)
print(grid)
# <xarray.DataArray (height: 3, width: 3)>
# array([[0, 1, 2],
#        [3, 4, 5],
#        [6, 7, 8]])
# Coordinates:
#   * height   (height) float64 1.5 2.5 3.5
#   * width    (width) float64 4.5 5.5 6.5
```

Run it through a data processing function like [`grdhisteq.equalize_grid`](https://www.pygmt.org/v0.6.0/api/generated/pygmt.grdhisteq.equalize_grid.html)

```python
newgrid = pygmt.grdhisteq.equalize_grid(grid=grid)
print(newgrid)
# <xarray.DataArray 'z' (y: 3, x: 3)>
# array([[ 0.,  4.,  6.],
#        [ 8.,  8., 11.],
#        [13., 15., 15.]], dtype=float32)
# Coordinates:
#   * x        (x) float64 4.5 5.5 6.5
#   * y        (y) float64 1.5 2.5 3.5
# Attributes:
#     long_name:     z
#     actual_range:  [ 0. 15.]
```

Ideally, the `newgrid.dims` should be `('height', 'width')` instead of `('y', 'x')`.

**Are you willing to help implement and maintain this feature?** Yes, but happy to let someone take this up too!

PR #2235, opened 2022-12-07
Add function to load Blue Marble dataset

willschlitzer:
This PR (hopefully) adds a function to load the Blue Marble image dataset.

- **Official documentation**: https://www.generic-mapping-tools.org/remote-datasets/earth-daynight.html
- **Recipe**: https://github.com/GenericMappingTools/gmtserver-admin/blob/master/recipes/earth_day.recipe
- **Preview**: https://pygmt-dev--2235.org.readthedocs.build/en/2235/api/generated/pygmt.datasets.load_blue_marble.html

Addresses #1442 

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [x] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2398, opened 2023-03-05
Wrap GMT's standard data type GMT_GRID for grids

seisman:

This PR wraps GMT's standard data type `GMT_GRID` for grids, so that we can store the output grid in GMT virtual files and don't have to write the output grid into a temporary file and load it as xarray.DataArray.

## Expected enhancements/features

Here is a list of enhancements/features related to the `GMT_GRID` data type:

- [x] Wrap GMT's data structure `GMT_GRID_HEADER` for grid metadata [#3127]
- [x] Wrap GMT's standard data type `GMT_GRID` for grids 
- [x] Add the `Session.virtualfile_to_raster` method for handling output to an actual file or a virtual file [Similar to the `Session.virtualfile_to_dataset` in PR #3083] 
- [x] Refactor wrappers that write grids: (1) Simplify the codes; (2) Get rid of temporary files; (3) Improve performance 
- [ ] Passing `GMT_GRID` directly to GMT API. Currently, for an xarray.DataArray grid, we first convert it to a matrix, then pass the matrix with some metadata (like inc, registration and gtype) to GMT. This way has proven to cause troubles (e.g., #390), mainly due to grid paddings. It would be better if we could internally convert xarray.DataArray to GMT_GRID and then pass GMT_GRID to GMT API. It may help us fix many long-term issues. [Will do in future PRs]

## The `GMT_GRID`/`GMT_GRID_HEADER` data structure

The `GMT_GRID` data structure is defined at [gmt/src/gmt_resources.h](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_resources.h#L467):
```c
struct GMT_GRID {	/* To hold a GMT gmt_grdfloat grid and its header in one container */
	struct GMT_GRID_HEADER *header;	/* Pointer to full GMT header for the grid */
	gmt_grdfloat *data;             /* Pointer to the gmt_grdfloat grid */
	double *x, *y;                  /* Vector of coordinates */
	void *hidden;                   /* Row-by-row machinery information [NULL] */
};
```
it contains the x/y array, the data array and the grid header. The grid header `GMT_GRID_HEADER` is defined at [gmt/src/gmt_resources.h](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_resources.h#L402). It contains many important metadata information about the grid. Here are some important header variables:

- `n_columns`: Number of columns
- `n_rows`: Number of rows
- `registration`: 0 for gridline registration and 1 for pixel registration.
- `wesn`: min/max of x/y coordinates, i.e., [xmin, xmax, ymin, ymax]
- `z_min`/`z_max`: min/max of z values
- `inc`: x and y increment [should be equal to x[1] - x[0] and y[1] - y[0]]
- `z_scale_factor`/`z_add_offset`: Used for data compressing by storing floating-point data as smaller integers in a netCDF file. The relation between the actual data _z'_ and the data stored in netCDF file (_z_) is z' = z * scale_factor + offset (i.e., unpacking). Need to note that when reading a netCDF file, GMT already does the conversion, so `z_min`/`z_max` and the `data` array are already unpacked so we usually don't need to care about these two variables.
- `x_units`/`y_units`: x/y units in the form of `long_name [units]`. The `[units]` part is optional. Some examples are: `x`/`y`/`longitude [degrees_east]`/`latitude [degrees_north]`.
- `z_units`: Also in the form of `long_name [units]`.
- `pad`: Extra columns/rows on west/east/south/north sides. Paddings are mainly used for grid manipulations at grid boundary. Defaults are [2, 2, 2, 2].
- `mx`: actual columns in memory. Equal to n_columns + pad[0] + pad[1]
- `my`: actual rows in memory. Equal to n_rows + pad[2] + pad[3]

### Conversions between netCDF and `GMT_GRID`

Although netCDF is the default grid format for I/O, GMT uses the `GMT_GRID` data structure as its internal representation of grids. Thus, GMT needs to convert netCDF data/metadata to `GMT_GRID` for input and do the inverse for output. The conversions between netCDF and `GMT_GRID` are done by functions `gmt_nc_read_grd`/`gmt_nc_write_grd` in [`gmt/src/gmt_nc.c`](https://github.com/GenericMappingTools/gmt/blob/master/src/gmt_nc.c). Below Here are some technical notes:

- `x_units`/`y_units`/`z_units` has the form of `long_name [units]`, in which `[units]` is optional. `long_name` and `units` are netCDF attributes. [gmtnc_get_units](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L255)
- When reading, GMT reads the variable's `long_name` and `units` attributes and then sets the `x_units`/`y_units`/`z_units` string. If `long_name` is not defined, then use the variable's name instead. If `units` is not defined, then `[units]` is not appended. [gmtnc_get_units](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L255)
- For writing, GMT extracts the `long_name` and `units` attributes by parsing the `x_units`/`y_units`/`z_units` string, and setting the netCDF attributes if they're defined.  More attributes are added if it's a geographic coordinate. If `units="degrees_east"`, set attributes `standard_name` to `longitude` and `axis` to `X`; If `units="degrees_north"`, set attributes `standard_name` to `latitude` and `axis` to `Y`. [gmtnc_put_units](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L270)
- The default dimension names are `lon`/`time`/`x` for x axis and `lat`/`time`/`y` for y axis. [reference](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L566)
- The default name for the data is always `z` for 2-D grid or `cube` for 3-D cube, unless the variable name is given. The variable name is stored in the hidden structure `HH->varname`, but how is `HH->varname` determined?. [reference](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L591)
- When writing a netCDF file, GMT defines some global attributes, including `Conventions` (`CF-1.7`), `title`, `history`, `description`, `GMT_version`, `node_offset` (only for pixel grids). [reference](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L985)
- When writing x/y vectors, axis attributes `actual_range` (from `header-wesn`) and `axis` (`X` or `Y`) are also set. [reference](https://github.com/GenericMappingTools/gmt/blob/83f885ea4c2452e78a9d885a6c5840805781ed16/src/gmt_nc.c#L1044)
- When writing the data, some attributes are defined: `scale_factor` (if not 1.0), `add_offset` (if not 0.0), `cpt` (if defined by GMT), `_FillValue` (for NaN), `actual_range`.

## Benefits of current implementation

1. Simplify the codes of module wrappers
2. No longer to write output grid to temporary files
3. Performance improvements (https://github.com/GenericMappingTools/pygmt/pull/2398#issuecomment-1996219144)

## Limitations of current implementation

In addition to the metadata stored in the `GMT_GRID_HEADER` object, GMT also has some hidden metadata which is stored in a hidden data structure, for example, `grdtype` (Cartesian or Geographic grid?), `cpt` (the default CPT). The hidden data structure is private and may change between different GMT versions. So it's not a good practice to wrap and use the hidden data structure. But it also means we can't know the `grdtype`/`cpt` attributes when converting `GMT_GRID` to `xarray.DataArray`.

For `grdtype`, it's reasonable to say that the grid is geographic if x's unit is `degrees_east` and y's unit is `degrees_north` (see the note above and also [CF conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html)), but for `cpt`, we can't know if a grid has a default CPT or not.

Another limitation is that, `grdcut` now can't cut an image because we assume that the output is a grid. An initial fix is at #3115. But, as far as I can see, although many GMT modules can accept an image (e.g., `grdcut`, `grdimage`, `grdview`), only `grdcut` and `grdmix` can output an image. 

## References

- Function `gmtnc_put_units` in `gmt_nc.c`
- Function `gmtnc_get_units` in `gmt_nc.c`
- Function `gmtnc_grd_info` in `gmt_nc.c`
- Function `gmt_grd_init` in `gmt_grdio.c`

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1958,7 +1958,7 @@ class Session:
         outgrid: str | None = None,
     ) -> xr.DataArray | None:
         """
-        Output a raster data stored in a virtual file to an :class:`xarray.DataArray`
+        Output raster data stored in a virtual file to an :class:`xarray.DataArray`
         object.
 
         The raster data can be a grid, an image or a cube.
@@ -1966,7 +1966,7 @@ class Session:
         Parameters
         ----------
         vfname
-            The virtual file name that stores the result grid.
+            The virtual file name that stores the result grid/image/cube.
         kind
             Type of the raster data. Valid values are ``"grid"``, ``"image"``,
             ``"cube"`` or ``None``. If ``None``, will inquire the data type from the
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1947,24 +1947,34 @@ class Session:
             return result.to_numpy()
         return result  # pandas.DataFrame output
 
-    def virtualfile_to_grid(
-        self, vfname: str, outgrid: str | None
+    def virtualfile_to_raster(
+        self,
+        vfname: str,
+        kind: Literal["grid", "image", "cube", None] = None,
+        outgrid: str | None = None,
     ) -> xr.DataArray | None:
         """
-        Output a grid stored in a virtual file to an :class:`xarray.DataArray` object.
+        Output a raster data stored in a virtual file to an :class:`xarray.DataArray`
+        object.
+
+        The raster data can be a grid, an image or a cube.
 
         Parameters
         ----------
         vfname
             The virtual file name that stores the result grid.
+        kind
+            Type of the raster data. Valid values are ``"grid"``, ``"image"``,
+            ``"cube"`` or ``None``. If ``None``, will inquire the data type from the
+            virtual file name.
         outgrid
-            Name of the output grid. If specified, it means the grid was already saved
-            into an actual file and will return ``None``.
+            Name of the output grid/image/cube. If specified, it means the raster data
+            was already saved into an actual file and will return ``None``.
 
         Returns
         -------
         result
-            The result grid. If ``outgrid`` is specified, return ``None``.
+            The result grid/image/cube. If ``outgrid`` is specified, return ``None``.
 
         Examples
         --------
@@ -1977,7 +1987,7 @@ class Session:
         ...         outgrid = tmpfile.name
         ...         with lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd:
         ...             lib.call_module("read", f"@earth_relief_01d_g {voutgrd} -Tg")
-        ...             result = lib.virtualfile_to_grid(
+        ...             result = lib.virtualfile_to_raster(
         ...                 vfname=voutgrd, outgrid=outgrid
         ...             )
         ...             assert result == None
@@ -1987,12 +1997,19 @@ class Session:
         ...     outgrid = None
         ...     with lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd:
         ...         lib.call_module("read", f"@earth_relief_01d_g {voutgrd} -Tg")
-        ...         result = lib.virtualfile_to_grid(vfname=voutgrd, outgrid=outgrid)
+        ...         result = lib.virtualfile_to_raster(vfname=voutgrd, outgrid=outgrid)
         ...         assert isinstance(result, xr.DataArray)
         """
         if outgrid is not None:
             return None
-        return self.read_virtualfile(vfname, kind="grid").contents.to_dataarray()
+        if kind is None:  # Inquire the data family from the virtualfile
+            family = self.inquire_virtualfile(vfname)
+            kind = {
+                self["GMT_IS_GRID"]: "grid",
+                self["GMT_IS_IMAGE"]: "image",
+                self["GMT_IS_CUBE"]: "cube",
+            }[family]
+        return self.read_virtualfile(vfname, kind=kind).contents.to_dataarray()
 
     def extract_region(self):
         """
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -14,6 +14,7 @@ from typing import Literal
 
 import numpy as np
 import pandas as pd
+import xarray as xr
 from packaging.version import Version
 from pygmt.clib.conversion import (
     array_to_datetime,
@@ -1899,6 +1900,53 @@ class Session:
         if output_type == "numpy":  # numpy.ndarray output
             return result.to_numpy()
         return result  # pandas.DataFrame output
+
+    def virtualfile_to_grid(
+        self, vfname: str, outgrid: str | None
+    ) -> xr.DataArray | None:
+        """
+        Output a grid stored in a virtual file to an :class:`xarray.DataArray` object.
+
+        Parameters
+        ----------
+        vfname
+            The virtual file name that stores the result grid.
+        outgrid
+            Name of the output grid. If specified, it means the grid was already saved
+            into an actual file and will return ``None``.
+
+        Returns
+        -------
+        result
+            The result grid. If ``outgrid`` is specified, return ``None``.
+
+        Examples
+        --------
+        >>> from pathlib import Path
+        >>> from pygmt.clib import Session
+        >>> from pygmt.helpers import GMTTempFile
+        >>> with Session() as lib:
+        ...     # file output
+        ...     with GMTTempFile(suffix=".nc") as tmpfile:
+        ...         outgrid = tmpfile.name
+        ...         with lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd:
+        ...             lib.call_module("read", f"@earth_relief_01d_g {voutgrd} -Tg")
+        ...             result = lib.virtualfile_to_grid(
+        ...                 vfname=voutgrd, outgrid=outgrid
+        ...             )
+        ...             assert result == None
+        ...             assert Path(outgrid).stat().st_size > 0
+        ...
+        ...     # xarray.DataArray output
+        ...     outgrid = None
+        ...     with lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd:
+        ...         lib.call_module("read", f"@earth_relief_01d_g {voutgrd} -Tg")
+        ...         result = lib.virtualfile_to_grid(vfname=voutgrd, outgrid=outgrid)
+        ...         assert isinstance(result, xr.DataArray)
+        """
+        if outgrid is not None:
+            return None
+        return self.read_virtualfile(vfname, kind="grid").contents.to_dataarray()
 
     def extract_region(self):
         """
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1954,7 +1954,7 @@ class Session:
     def virtualfile_to_raster(
         self,
         vfname: str,
-        kind: Literal["grid", "image", "cube", None] = None,
+        kind: Literal["grid", "image", "cube", None] = "grid",
         outgrid: str | None = None,
     ) -> xr.DataArray | None:
         """
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1740,7 +1740,9 @@ class Session:
         return c_inquire_virtualfile(self.session_pointer, vfname.encode())
 
     def read_virtualfile(
-        self, vfname: str, kind: Literal["dataset", "grid", None] = None
+        self,
+        vfname: str,
+        kind: Literal["dataset", "grid", "image", "cube", None] = None,
     ):
         """
         Read data from a virtual file and optionally cast into a GMT data container.
@@ -1799,6 +1801,8 @@ class Session:
         # _GMT_DATASET).
         if kind is None:  # Return the ctypes void pointer
             return pointer
+        if kind in ["image", "cube"]:
+            raise NotImplementedError(f"kind={kind} is not supported yet.")
         dtype = {"dataset": _GMT_DATASET, "grid": _GMT_GRID}[kind]
         return ctp.cast(pointer, ctp.POINTER(dtype))
 
@@ -2004,7 +2008,7 @@ class Session:
             return None
         if kind is None:  # Inquire the data family from the virtualfile
             family = self.inquire_virtualfile(vfname)
-            kind = {
+            kind = {  # type: ignore[assignment]
                 self["GMT_IS_GRID"]: "grid",
                 self["GMT_IS_IMAGE"]: "image",
                 self["GMT_IS_CUBE"]: "cube",
--- a/pygmt/datatypes/grid.py
+++ b/pygmt/datatypes/grid.py
@@ -14,7 +14,7 @@ class _GMT_GRID(ctp.Structure):  # noqa: N801
     """
     GMT grid structure for holding a grid and its header.
 
-    The class is only meant for internal use and is not exposed to users. See the GMT
+    This class is only meant for internal use and is not exposed to users. See the GMT
     source code gmt_resources.h for the original C structure definitions.
 
     Examples
@@ -192,8 +192,8 @@ class _GMT_GRID(ctp.Structure):  # noqa: N801
             if grid[dim][0] > grid[dim][1]:
                 grid = grid.isel({dim: slice(None, None, -1)})
 
-        # Set the gmt accessor.
-        # Must put at the end. The information get lost after specific grid operation.
+        # Set GMT accessors.
+        # Must put at the end, otherwise info gets lost after certain grid operations.
         grid.gmt.registration = header.registration
         grid.gmt.gtype = header.gtype
         return grid
--- a/pygmt/datatypes/grid.py
+++ b/pygmt/datatypes/grid.py
@@ -3,7 +3,197 @@ Wrapper for the GMT_GRID data type.
 """
 
 import ctypes as ctp
+from typing import ClassVar
+
+import numpy as np
+import xarray as xr
+from pygmt.datatypes.header import _GMT_GRID_HEADER, gmt_grdfloat
 
 
 class _GMT_GRID(ctp.Structure):  # noqa: N801
-    pass
+    """
+    GMT grid structure for holding a grid and its header.
+
+    The class is only meant for internal use and is not exposed to users. See the GMT
+    source code gmt_resources.h for the original C structure definitions.
+
+    Examples
+    --------
+    >>> from pygmt.clib import Session
+    >>> with Session() as lib:
+    ...     with lib.virtualfile_out(kind="grid") as voutgrd:
+    ...         lib.call_module("read", f"@static_earth_relief.nc {voutgrd} -Tg")
+    ...         # Read the grid from the virtual file
+    ...         grid = lib.read_virtualfile(voutgrd, kind="grid").contents
+    ...         # The grid header
+    ...         header = grid.header.contents
+    ...         # Access the header properties
+    ...         print(header.n_rows, header.n_columns, header.registration)
+    ...         print(header.wesn[:], header.z_min, header.z_max, header.inc[:])
+    ...         print(header.z_scale_factor, header.z_add_offset)
+    ...         print(header.x_units, header.y_units, header.z_units)
+    ...         print(header.title)
+    ...         print(header.command)
+    ...         print(header.remark)
+    ...         print(header.nm, header.size, header.complex_mode)
+    ...         print(header.type, header.n_bands, header.mx, header.my)
+    ...         print(header.pad[:])
+    ...         print(header.mem_layout, header.nan_value, header.xy_off)
+    ...         # The x and y coordinates
+    ...         print(grid.x[: header.n_columns])
+    ...         print(grid.y[: header.n_rows])
+    ...         # The data array (with paddings)
+    ...         data = np.reshape(
+    ...             grid.data[: header.mx * header.my], (header.my, header.mx)
+    ...         )
+    ...         # The data array (without paddings)
+    ...         pad = header.pad[:]
+    ...         data = data[pad[2] : header.my - pad[3], pad[0] : header.mx - pad[1]]
+    ...         print(data)
+    14 8 1
+    [-55.0, -47.0, -24.0, -10.0] 190.0 981.0 [1.0, 1.0]
+    1.0 0.0
+    b'longitude [degrees_east]' b'latitude [degrees_north]' b'elevation (m)'
+    b'Produced by grdcut'
+    b'grdcut @earth_relief_01d_p -R-55/-47/-24/-10 -Gstatic_earth_relief.nc'
+    b'Reduced by Gaussian Cartesian filtering (111.2 km fullwidth) from ...'
+    112 216 0
+    18 1 12 18
+    [2, 2, 2, 2]
+    b'' nan 0.5
+    [-54.5, -53.5, -52.5, -51.5, -50.5, -49.5, -48.5, -47.5]
+    [-10.5, -11.5, -12.5, -13.5, -14.5, -15.5, ..., -22.5, -23.5]
+    [[347.5 331.5 309.  282.  190.  208.  299.5 348. ]
+    [349.  313.  325.5 247.  191.  225.  260.  452.5]
+    [345.5 320.  335.  292.  207.5 247.  325.  346.5]
+    [450.5 395.5 366.  248.  250.  354.5 550.  797.5]
+    [494.5 488.5 357.  254.5 286.  484.5 653.5 930. ]
+    [601.  526.5 535.  299.  398.5 645.  797.5 964. ]
+    [308.  595.5 555.5 556.  580.  770.  927.  920. ]
+    [521.5 682.5 796.  886.  571.5 638.5 739.5 881.5]
+    [310.  521.5 757.  570.5 538.5 524.  686.5 794. ]
+    [561.5 539.  446.5 481.5 439.5 553.  726.5 981. ]
+    [557.  435.  385.5 345.5 413.5 496.  519.5 833.5]
+    [373.  367.5 349.  352.5 419.5 428.  570.  667.5]
+    [383.  284.5 344.5 394.  491.  556.5 578.5 618.5]
+    [347.5 344.5 386.  640.5 617.  579.  646.5 671. ]]
+    """
+
+    _fields_: ClassVar = [
+        # Pointer to full GMT header for grid
+        ("header", ctp.POINTER(_GMT_GRID_HEADER)),
+        # Pointer to grid data
+        ("data", ctp.POINTER(gmt_grdfloat)),
+        # Pointer to x coordinate vector
+        ("x", ctp.POINTER(ctp.c_double)),
+        # Pointer to y coordinate vector
+        ("y", ctp.POINTER(ctp.c_double)),
+        # Low-level information for GMT use only
+        ("hidden", ctp.c_void_p),
+    ]
+
+    def to_dataarray(self) -> xr.DataArray:
+        """
+        Convert a _GMT_GRID object to a :class:`xarray.DataArray` object.
+
+        Returns
+        -------
+        dataarray
+            A :class:`xr.DataArray` object.
+
+        Examples
+        --------
+        >>> from pygmt.clib import Session
+        >>> with Session() as lib:
+        ...     with lib.virtualfile_out(kind="grid") as voutgrd:
+        ...         lib.call_module("read", f"@static_earth_relief.nc {voutgrd} -Tg")
+        ...         # Read the grid from the virtual file
+        ...         grid = lib.read_virtualfile(voutgrd, kind="grid")
+        ...         # Convert to xarray.DataArray and use it later
+        ...         da = grid.contents.to_dataarray()
+        >>> da  # doctest: +NORMALIZE_WHITESPACE, +ELLIPSIS
+        <xarray.DataArray 'z' (lat: 14, lon: 8)>...
+        array([[347.5, 344.5, 386. , 640.5, 617. , 579. , 646.5, 671. ],
+               [383. , 284.5, 344.5, 394. , 491. , 556.5, 578.5, 618.5],
+               [373. , 367.5, 349. , 352.5, 419.5, 428. , 570. , 667.5],
+               [557. , 435. , 385.5, 345.5, 413.5, 496. , 519.5, 833.5],
+               [561.5, 539. , 446.5, 481.5, 439.5, 553. , 726.5, 981. ],
+               [310. , 521.5, 757. , 570.5, 538.5, 524. , 686.5, 794. ],
+               [521.5, 682.5, 796. , 886. , 571.5, 638.5, 739.5, 881.5],
+               [308. , 595.5, 555.5, 556. , 580. , 770. , 927. , 920. ],
+               [601. , 526.5, 535. , 299. , 398.5, 645. , 797.5, 964. ],
+               [494.5, 488.5, 357. , 254.5, 286. , 484.5, 653.5, 930. ],
+               [450.5, 395.5, 366. , 248. , 250. , 354.5, 550. , 797.5],
+               [345.5, 320. , 335. , 292. , 207.5, 247. , 325. , 346.5],
+               [349. , 313. , 325.5, 247. , 191. , 225. , 260. , 452.5],
+               [347.5, 331.5, 309. , 282. , 190. , 208. , 299.5, 348. ]])
+        Coordinates:
+          * lat      (lat) float64... -23.5 -22.5 -21.5 -20.5 ... -12.5 -11.5 -10.5
+          * lon      (lon) float64... -54.5 -53.5 -52.5 -51.5 -50.5 -49.5 -48.5 -47.5
+        Attributes:
+            Conventions:   CF-1.7
+            title:         Produced by grdcut
+            history:       grdcut @earth_relief_01d_p -R-55/-47/-24/-10 -Gstatic_ea...
+            description:   Reduced by Gaussian Cartesian filtering (111.2 km fullwi...
+            long_name:     elevation (m)
+            actual_range:  [190. 981.]
+        >>> da.coords["lon"]  # doctest: +NORMALIZE_WHITESPACE, +ELLIPSIS
+        <xarray.DataArray 'lon' (lon: 8)>...
+        array([-54.5, -53.5, -52.5, -51.5, -50.5, -49.5, -48.5, -47.5])
+        Coordinates:
+          * lon      (lon) float64... -54.5 -53.5 -52.5 -51.5 -50.5 -49.5 -48.5 -47.5
+        Attributes:
+            long_name:      longitude
+            units:          degrees_east
+            standard_name:  longitude
+            axis:           X
+            actual_range:   [-55. -47.]
+        >>> da.coords["lat"]  # doctest: +NORMALIZE_WHITESPACE, +ELLIPSIS
+        <xarray.DataArray 'lat' (lat: 14)>...
+        array([-23.5, -22.5, -21.5, -20.5, -19.5, -18.5, -17.5, -16.5, -15.5, -14.5,
+            -13.5, -12.5, -11.5, -10.5])
+        Coordinates:
+          * lat      (lat) float64... -23.5 -22.5 -21.5 -20.5 ... -12.5 -11.5 -10.5
+        Attributes:
+            long_name:      latitude
+            units:          degrees_north
+            standard_name:  latitude
+            axis:           Y
+            actual_range:   [-24. -10.]
+        >>> da.gmt.registration, da.gmt.gtype
+        (1, 1)
+        """
+        # The grid header
+        header = self.header.contents
+
+        # Get dimensions and their attributes from the header.
+        dims, dim_attrs = header.dims, header.dim_attrs
+        # The coordinates, given as a tuple of the form (dims, data, attrs)
+        coords = [
+            (dims[0], self.y[: header.n_rows], dim_attrs[0]),
+            (dims[1], self.x[: header.n_columns], dim_attrs[1]),
+        ]
+
+        # The data array without paddings
+        pad = header.pad[:]
+        data = np.reshape(self.data[: header.mx * header.my], (header.my, header.mx))[
+            pad[2] : header.my - pad[3], pad[0] : header.mx - pad[1]
+        ]
+
+        # Create the xarray.DataArray object
+        grid = xr.DataArray(
+            data, coords=coords, name=header.name, attrs=header.data_attrs
+        )
+
+        # Flip the coordinates and data if necessary so that coordinates are ascending.
+        # `grid.sortby(list(grid.dims))` sometimes causes crashes.
+        # The solution comes from https://github.com/pydata/xarray/discussions/6695.
+        for dim in grid.dims:
+            if grid[dim][0] > grid[dim][1]:
+                grid = grid.isel({dim: slice(None, None, -1)})
+
+        # Set the gmt accesssor.
+        # Must put at the end. The information get lost after specific grid operation.
+        grid.gmt.registration = header.registration
+        grid.gmt.gtype = header.gtype
+        return grid
--- a/pygmt/datatypes/grid.py
+++ b/pygmt/datatypes/grid.py
@@ -192,7 +192,7 @@ class _GMT_GRID(ctp.Structure):  # noqa: N801
             if grid[dim][0] > grid[dim][1]:
                 grid = grid.isel({dim: slice(None, None, -1)})
 
-        # Set the gmt accesssor.
+        # Set the gmt accessor.
         # Must put at the end. The information get lost after specific grid operation.
         grid.gmt.registration = header.registration
         grid.gmt.gtype = header.gtype
--- a/pygmt/tests/test_datasets_load_remote_datasets.py
+++ b/pygmt/tests/test_datasets_load_remote_datasets.py
@@ -3,8 +3,6 @@ Test the _load_remote_dataset function.
 """
 
 import pytest
-from packaging.version import Version
-from pygmt.clib import __gmt_version__
 from pygmt.datasets.load_remote_dataset import _load_remote_dataset
 from pygmt.exceptions import GMTInvalidInput
 
@@ -35,8 +33,9 @@ def test_load_remote_dataset_benchmark_with_region():
     assert data.gmt.registration == 0
     assert data.shape == (11, 21)
     # The cpt attribute was added since GMT 6.4.0
-    if Version(__gmt_version__) >= Version("6.4.0"):
-        assert data.attrs["cpt"] == "@earth_age.cpt"
+    # Can't access the cpt attribute using virtual files
+    # if Version(__gmt_version__) >= Version("6.4.0"):
+    #    assert data.attrs["cpt"] == "@earth_age.cpt"
 
 
 def test_load_remote_dataset_invalid_resolutions():
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -292,6 +292,7 @@ Python objects to and from GMT virtual files:
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
     clib.Session.virtualfile_to_dataset
+    clib.Session.virtualfile_to_grid
 
 Low level access (these are mostly used by the :mod:`pygmt.clib` package):
 
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -292,7 +292,7 @@ Python objects to and from GMT virtual files:
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
     clib.Session.virtualfile_to_dataset
-    clib.Session.virtualfile_to_grid
+    clib.Session.virtualfile_to_raster
 
 Low level access (these are mostly used by the :mod:`pygmt.clib` package):
 
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -267,10 +267,12 @@ COMMON_DOCSTRINGS = {
             - ``file`` will save the result to the file specified by the ``outfile``
               parameter.""",
     "outgrid": """
-        outgrid : str or None
-            Name of the output netCDF grid file. For writing a specific grid
-            file format or applying basic data operations to the output grid,
-            see :gmt-docs:`gmt.html#grd-inout-full` for the available modifiers.""",
+        outgrid
+            Name of the output netCDF grid file. If not specified, will return an
+            :class:`xarray.DataArray` object. For writing a specific grid file format or
+            applying basic data operations to the output grid, see
+            :gmt-docs:`gmt.html#grd-inout-full` for the available modifiers.
+        """,
     "panel": r"""
         panel : bool, int, or list
             [*row,col*\|\ *index*].
--- a/pygmt/src/binstats.py
+++ b/pygmt/src/binstats.py
@@ -111,4 +111,6 @@ def binstats(data, outgrid: str | None = None, **kwargs):
             lib.call_module(
                 module="binstats", args=build_arg_string(kwargs, infile=vintbl)
             )
-            return lib.virtualfile_to_raster(vfname=voutgrd, outgrid=outgrid)
+            return lib.virtualfile_to_raster(
+                vfname=voutgrd, kind="grid", outgrid=outgrid
+            )
--- a/pygmt/src/binstats.py
+++ b/pygmt/src/binstats.py
@@ -111,4 +111,4 @@ def binstats(data, outgrid: str | None = None, **kwargs):
             lib.call_module(
                 module="binstats", args=build_arg_string(kwargs, infile=vintbl)
             )
-            return lib.virtualfile_to_grid(vfname=voutgrd, outgrid=outgrid)
+            return lib.virtualfile_to_raster(vfname=voutgrd, outgrid=outgrid)
--- a/pygmt/src/binstats.py
+++ b/pygmt/src/binstats.py
@@ -111,6 +111,4 @@ def binstats(data, outgrid: str | None = None, **kwargs):
             lib.call_module(
                 module="binstats", args=build_arg_string(kwargs, infile=vintbl)
             )
-            return lib.virtualfile_to_raster(
-                vfname=voutgrd, kind="grid", outgrid=outgrid
-            )
+            return lib.virtualfile_to_raster(vfname=voutgrd, outgrid=outgrid)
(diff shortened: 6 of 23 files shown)


pygmt/src/grdcut.py
```diff
@@ -102,14 +104,18 @@ def grdcut(grid, **kwargs):
...
-    with GMTTempFile(suffix=".nc") as tmpfile:
-        with Session() as lib:
-            file_context = lib.virtualfile_from_data(check_kind="raster", data=grid)
-            with file_context as infile:
-                if (outgrid := kwargs.get("G")) is None:
-                    kwargs["G"] = outgrid = tmpfile.name  # output to tmpfile
-                lib.call_module(
-                    module="grdcut", args=build_arg_string(kwargs, infile=infile)
-                )
+    with Session() as lib:
```
seisman, line 107: Some explanations for the new `grdcut` codes:

- Lines 108-110: Create two virtual files, one for the input `grid`, another one for the out grid
- Lines 111-114: Set `-G{outfile}` so that the `grdcut` call will store the result grid in the virtualfile `outfile`
- Lines 115: Read the content of the virtual file to the variable `gmtgrid`
- Lines 117-121: Process the result `gmtgrid` based on the value of `outgrid`.
  - if `outgrid` is not None, then write it to a grid file by calling the special `write` module
  - if `outgrid` is None, then convert the `GMT_GRID` object to a `xarray.DataArray` object.

pygmt/datatypes.py
```diff
@@ -0,0 +1,101 @@
...
+        ("y", ctp.POINTER(ctp.c_double)),  # pointer to y coordinate vector
+        ("hidden", ctp.c_void_p),  # low-level information for GMT use only
+    ]
+
+    def to_dataarray(self):
+        header = self.header.contents
+        pad = header.pad[:]
+        x = self.x[: header.n_columns]
+        y = self.y[: header.n_rows]
+        data = np.array(self.data[: header.mx * header.my]).reshape(
```
seisman, line 92: The grid has `n_columns` columns and `n_rows` rows, but usually the number of data points **in memory** are not `n_columns * n_rows` due to grid pads. For some specific reasons, GMT usually adds two more rows/columns on west, east, south, north sides. So, we should have:
```
mx = pad[0] + n_columns + pad[1]
my = pad[2] + n_rows + pad[3]
```
That's why we need to use the following codes to get the data matrix:
```
data = np.array(self.data[: header.mx * header.my]).reshape(
           (header.my, header.mx)
       )[pad[2] : header.my - pad[3], pad[0] : header.mx - pad[1]]
```

weiji14 (commented):
This is amazing, thanks for pulling this together @seisman! I'll need to take a closer look to review this properly, but it's looking very promising :star_struck:

Currently the unit tests are failing, mostly because the output `xarray.DataArray` doesn't have the same attributes/metadata as when we were using the NetCDF tempfile. Will need to find a way to copy those metadata over somehow.

seisman:
> Currently the unit tests are failing, mostly because the output `xarray.DataArray` doesn't have the same attributes/metadata as when we were using the NetCDF tempfile. Will need to find a way to copy those metadata over somehow.

It seems that all the failing tests are due to the same reason: the `actual_range` attribute is missing. 
```
>       grid.attrs.pop("actual_range")
E       KeyError: 'actual_range'
```
It's easy to add the attribute to the xarray.DataArray object, but perhaps we should check if the attribute exists before trying to remove it.
https://github.com/GenericMappingTools/pygmt/blob/301730f36bf7faac400ad730bd1a25801d125cfc/pygmt/datasets/load_remote_dataset.py#L325

pygmt/datatypes.py
```diff
@@ -0,0 +1,107 @@
...
+
+These data types are defined in the GMT source code "src/gmt_resources.h".
+"""
+import ctypes as ctp
+
+import numpy as np
+import xarray as xr
+from pygmt.clib.session import Session
+
+# Lenghts of grid header variables
```
weiji14, line 12:
```suggestion
# Lengths of grid header variables
```

weiji14 (commented):
Most of the tests on `grdview` (which uses `grdcut` for the fixture) are failing with an incorrect figure size. Is there something to be fixed with the padding?

seisman:
> Most of the tests on `grdview` (which uses `grdcut` for the fixture) are failing with an incorrect figure size. Is there something to be fixed with the padding?

No, if you compare the baseline and result images, you'll see that the differences are caused by the `gtype` property. `gtype` should be 1 for the grid used in `grdview` tests, but currently the `GMT_GRID` structure doesn't store the `gtype` information.

pygmt/clib/session.py
```diff
@@ -1479,6 +1479,64 @@ def virtualfile_from_data(
...
+
+        Parameter
+        ---------
+        grid :
+            Pointer to the grid.
+
+        Yields
+        ------
+        vfile : str
+            The name of virtual file.
```
michaelgrund, line 1518:
```suggestion
            Name of the virtual file.
```

pygmt/clib/session.py
```diff
@@ -1479,6 +1479,64 @@ def virtualfile_from_data(
...
+        """
+        family = "GMT_IS_GRID"
+        geometry = "GMT_IS_SURFACE"
+        with self.open_virtual_file(family, geometry, "GMT_IN", grid) as vfile:
+            yield vfile
+
+    @contextmanager
+    def grid_to_virtualfile(self):
+        """
+        Create a virtual file for a output grid.
```
michaelgrund, line 1528:
```suggestion
        Create a virtual file for an output grid.
```

pygmt/clib/session.py
```diff
@@ -1479,6 +1479,64 @@ def virtualfile_from_data(
...
+
+        Yields
+        ------
+        vfile : str
+            The name of virtual file.
```
michaelgrund, line 1533:
```suggestion
            Name of the virtual file.
```

seisman:
> > Most of the tests on `grdview` (which uses `grdcut` for the fixture) are failing with an incorrect figure size. Is there something to be fixed with the padding?
> 
> No, if you compare the baseline and result images, you'll see that the differences are caused by the `gtype` property. `gtype` should be 1 for the grid used in `grdview` tests, but currently the `GMT_GRID` structure doesn't store the `gtype` information.

Added a temporary workaround at 6280e39e3b074d20a90134d289818850126317e7. And now all tests pass except `test_accessor_grid_source_file_not_exist` which is expected to fail.

pygmt/datatypes.py
```diff
@@ -0,0 +1,101 @@
...
+    GMT_GRID_REMARK_LEN160 = lib["GMT_GRID_REMARK_LEN160"]
+
+# GMT uses single-precision for grids by default, but can be built to use
+# double-precision. Currently, only single-precision is supported.
+gmt_grdfloat = ctp.c_float
+
+
+class GMT_GRID_HEADER(ctp.Structure):
+    """
+    Structure for GMT grid header.
```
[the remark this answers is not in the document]
seisman: > Add a link to https://docs.generic-mapping-tools.org/6.4/devdocs/api.html#gmt-grids?

The definition of the data structure in your link is incompleted. A more complete one is available at the end of the API documentation (https://docs.generic-mapping-tools.org/6.4/devdocs/api.html#struct-grid), but it's also out-of-date (the `ProjRefEPSG` field is missing).

Perhaps we should use https://github.com/GenericMappingTools/gmt/blob/6.4.0/src/gmt_resources.h#L466?

pygmt/datatypes.py
```diff
@@ -0,0 +1,101 @@
...
+        data = np.array(self.data[: header.mx * header.my]).reshape(
+            (header.my, header.mx)
+        )[pad[2] : header.my - pad[3], pad[0] : header.mx - pad[1]]
+        if x[0] > x[1]:
+            x = list(reversed(x))
+            data = np.fliplr(data)
+        if y[0] > y[1]:
+            y = list(reversed(y))
+            data = np.flipud(data)
+        return xr.DataArray(data, coords=[y, x], dims=["lat", "lon"])
```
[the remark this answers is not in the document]
seisman: The GMT_GRID object has the dimension names in `header.x_units` and `header.y_units` (e.g., `longitude [degrees_east]` and `latitude [degrees_north]`. So we should be able to specify the dimension names based on these values.

I just did a few tests below:

1. Reading a geographic grid using xarray gives:
```python
In [1]: import xarray as xr

In [2]: grid = xr.open_dataarray("earth_relief_01d_g.grd")

In [3]: grid
Out[3]: 
<xarray.DataArray 'z' (lat: 181, lon: 361)>
[65341 values with dtype=float32]
Coordinates:
  * lon      (lon) float64 -180.0 -179.0 -178.0 -177.0 ... 178.0 179.0 180.0
  * lat      (lat) float64 -90.0 -89.0 -88.0 -87.0 -86.0 ... 87.0 88.0 89.0 90.0
Attributes:
    actual_range:  [-8600.5  5559. ]
    long_name:     elevation (m)
```

2. I also tried to generate a Cartesian grid with `xyz2grd`:
```
gmt xyz2grd -R0/1/0/1 -I1 -Gcart.nc << EOF
0 0 0
0 1 1
1 0 2
1 1 3
EOF
```
Running `gmt grdinfo cart.nc` gives:
```
cart.nc: Title: 
cart.nc: Command: gmt xyz2grd -R0/1/0/1 -I1 -Gcart.nc
cart.nc: Remark: 
cart.nc: Gridline node registration used [Cartesian grid]
cart.nc: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
cart.nc: x_min: 0 x_max: 1 x_inc: 1 name: x n_columns: 2
cart.nc: y_min: 0 y_max: 1 y_inc: 1 name: y n_rows: 2
cart.nc: v_min: 0 v_max: 3 name: z
cart.nc: scale_factor: 1 add_offset: 0
cart.nc: format: classic
cart.nc: Default CPT:
```
Loading the grid using xarray gives:
```
<xarray.DataArray 'z' (y: 2, x: 2)>
[4 values with dtype=float32]
Coordinates:
  * x        (x) float64 0.0 1.0
  * y        (y) float64 0.0 1.0
Attributes:
    long_name:     z
    actual_range:  [0. 3.]
```

3. Also tried to change the x- and y-names to other values:
```
$ gmt xyz2grd -R0/1/0/1 -I1 -D+xnamex+ynamey -Gcart.nc << EOF
0 0 0
0 1 1
1 0 2
1 1 3
EOF
$ gmt grdinfo cart.nc 
cart.nc: Title: 
cart.nc: Command: gmt xyz2grd -R0/1/0/1 -I1 -D+xnamex+ynamey -Gcart.nc
cart.nc: Remark: 
cart.nc: Gridline node registration used [Cartesian grid]
cart.nc: Grid file format: nf = GMT netCDF format (32-bit float), CF-1.7
cart.nc: x_min: 0 x_max: 1 x_inc: 1 name: namex n_columns: 2
cart.nc: y_min: 0 y_max: 1 y_inc: 1 name: namey n_rows: 2
cart.nc: v_min: 0 v_max: 3 name: z
cart.nc: scale_factor: 1 add_offset: 0
cart.nc: format: classic
cart.nc: Default CPT: 
```
Loading the grid with xarray gives:
```
<xarray.DataArray 'z' (y: 2, x: 2)>
[4 values with dtype=float32]
Coordinates:
  * x        (x) float64 0.0 1.0
  * y        (y) float64 0.0 1.0
Attributes:
    long_name:     z
    actual_range:  [0. 3.]
```

My guess is, xarray always maps dimension names "longitude/latitude" to "lon"/"lat" and other names to "x"/"y". And there may be some conventions (e.g., https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#coards-relationship)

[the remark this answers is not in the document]
weiji14: What function are you using to "load the grid with xarray"? Is it [`xr.open_dataarray`](https://docs.xarray.dev/en/v2023.04.2/generated/xarray.open_dataarray.html) or [`pygmt.load_dataarray`](https://www.pygmt.org/v0.9.0/api/generated/pygmt.load_dataarray.html)? The .nc grid from xyz2grd doesn't have dimension names, so xarray would fallback to `x`/`y`, but if you read a NetCDF file that has properly set the dimension names (e.g. those at https://github.com/pydata/xarray-data), then you should get `lat`/`lon` or some other names instead.

Anyways, I'm ok with hardcoding the dimension names to `x`/`y` for now since that's what the current behaviour is, and we can work on preserving the dimension names (i.e. fix #1819) later.

pygmt/datatypes.py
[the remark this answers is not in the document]
weiji14: > > Add a link to https://docs.generic-mapping-tools.org/6.4/devdocs/api.html#gmt-grids?
> 
> The definition of the data structure in your link is incompleted. A more complete one is available at the end of the API documentation (https://docs.generic-mapping-tools.org/6.4/devdocs/api.html#struct-grid), but it's also out-of-date (the `ProjRefEPSG` field is missing).
> 
> Perhaps we should use https://github.com/GenericMappingTools/gmt/blob/6.4.0/src/gmt_resources.h#L466?

We could put both perhaps? Just want to have an up to date link when GMT 6.5.0 comes out.

pygmt/clib/session.py
```diff
@@ -314,6 +314,22 @@ def get_libgmt_func(self, name, argtypes=None, restype=None):
             function.restype = restype
         return function
 
+    def set_allocmode(self, family, obj):
```
weiji14, line 317: Please add these methods to `doc/api/index.rst` so that they will appear on https://www.pygmt.org/latest/api/index.html#gmt-c-api:

- [ ] set_allocmode
- [ ] read_virtualfile
- [ ] virtualfile_from_gmtgrid
- [ ] virtualfile_to_gmtgrid
- [ ] virtualfile_from_xrgrid

seisman: Done in f3fb38d0ffe691ee8f810490bfc531d850c1ba24.

.github/workflows/ci_tests.yaml
```diff
@@ -73,7 +73,7 @@ jobs:
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '1.26'
-            pandas-version: ''
+            pandas-version: '=2.1.4'
```
seisman, line 76: For mysterious reasons, this PR works well for pandas v1.5 (our minimum support version) and v2.1 (see the successful job run at https://github.com/GenericMappingTools/pygmt/actions/runs/8308744195/job/22739307032?pr=2398) but crashes for pandas v2.2 (see the failures at https://github.com/GenericMappingTools/pygmt/actions/runs/8308715874/job/22739243662?pr=2398).

In https://github.com/GenericMappingTools/pygmt/actions/runs/8308715874/job/22739243662?pr=2398, the test `test_grdlandmask_outgrid` crashes, but tests of other module wrappers (e.g., `grdcut`/`grdfill`) still pass.

seisman (commented):
Ping @GenericMappingTools/pygmt-maintainers 

This is another big step towards the goal of proving a Pythonic interface (without relying on temporary files). I've squashed the  commits to make the commit history easy to read. I also updated the top post (https://github.com/GenericMappingTools/pygmt/pull/2398#issue-1610245982) with some technical notes.

There are small some issues that may need to be addressed, but I feel this PR is now ready for review.

Edit: When reviewing PR #2235, I just realized that a big issue is that these module wrappers can also output an image (e.g., `.tif`) file, which is not supported in the current implementation.

Edit: Actually, as far as I can see, `grdcut` and `grdmix` are the only modules that can output an image. So the issue is not that big. We can either find a workaround like #3115 or just keep the `grdcut` unchanged in this PR.

.github/workflows/ci_tests.yaml
seisman: Still don't know what happens, but it seems `grid.sortby()` causes the crash. Found a solution at https://github.com/pydata/xarray/discussions/6695 and apply it in 229bc1e7b6875c95babeae0f1b77c87db0f780d2. Now no crashes.

pygmt/datatypes/grid.py
```diff
@@ -3,7 +3,362 @@
...
+    parts = nameunits.split("[")
+    long_name = parts[0].strip()
+    units = parts[1].strip("]").strip() if len(parts) > 1 else None
+    return long_name, units
+
+
+class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
+    """
+    GMT grid header structure for holding a grid header.
+    """
```
weiji14, line 68: Can we separate this _GMT_GRID_HEADER struct out into a separate file (`datatypes/header.py`?), maybe in another PR? Since this will be used for the `GMT_GRID`, `GMT_IMAGE` and `GMT_CUBE` structs.

weiji14 (commented):
> Edit: When reviewing PR #2235, I just realized that a big issue is that these module wrappers can also output an image (e.g., `.tif`) file, which is not supported in the current implementation.
> 
> Edit: Actually, as far as I can see, `grdcut` and `grdmix` are the only modules that can output an image. So the issue is not that big. We can either find a workaround like #3115 or just keep the `grdcut` unchanged in this PR.

I'm interested in wrapping the `GMT_IMAGE` struct, and can probably squeeze out some time to do it. If we can get the `GMT_GRID_HEADER` struct merged in a separate PR first, then I can work on that in parallel, building on top of #3115. That might mean pushing out PyGMT v0.12.0 (originally due 1 Apr 2024) to late April or May, but should be worth it to get most of the virtualfile handling done in one go.

seisman:
> I'm interested in wrapping the `GMT_IMAGE` struct, and can probably squeeze out some time to do it.

Actually I also did some experimental work at https://github.com/GenericMappingTools/pygmt/tree/datatypes/gmtimage. It looks very promising.

weiji14:
> > I'm interested in wrapping the `GMT_IMAGE` struct, and can probably squeeze out some time to do it.
> 
> Actually I also did some experimental work at https://github.com/GenericMappingTools/pygmt/tree/datatypes/gmtimage. It looks very promising.

Oh nice, do you want to continue work on that? Or I can pick it up from your last commit at ac83b99799f88ab89e6412e7da3c6dff1a48ca7a.

pygmt/datatypes/grid.py
weiji14: > > Can we separate this _GMT_GRID_HEADER struct out into a separate file (datatypes/header.py?), maybe in another PR? Since this will be used for the GMT_GRID, GMT_IMAGE and GMT_CUBE structs.
>
> As shown in https://github.com/GenericMappingTools/pygmt/tree/datatypes/gmtimage, we can keep _GMT_GRID_HEADER in grid.py and import it like from pygmt.datatypes.grid import _GMT_GRID_HEADER. Maybe it's OK to keep it in grid.py?

Ok, we can keep it in `datatypes/grid.py`, but let's wrap that in a separate PR.

weiji14:
> > > > I'm interested in wrapping the `GMT_IMAGE` struct, and can probably squeeze out some time to do it.
> > > 
> > > Actually I also did some experimental work at https://github.com/GenericMappingTools/pygmt/tree/datatypes/gmtimage. It looks very promising.
> > 
> > Oh nice, do you want to continue work on that? Or I can pick it up from your last commit at [ac83b99](https://github.com/GenericMappingTools/pygmt/commit/ac83b99799f88ab89e6412e7da3c6dff1a48ca7a).
> 
> I don't use 3-band image in my daily research, so have no knowledge about any conventions and technical details about tiff/geotiff. So feel free to push more changes to that branch.

No problem, I've been deep into looking at TIFF/GeoTIFF files recently at https://github.com/weiji14/cog3pio, so should be able to manage 3-bands :smile: Will open a separate PR for this later.

seisman:
> > > > > I'm interested in wrapping the `GMT_IMAGE` struct, and can probably squeeze out some time to do it.
> > > > 
> > > > 
> > > > Actually I also did some experimental work at https://github.com/GenericMappingTools/pygmt/tree/datatypes/gmtimage. It looks very promising.
> > > 
> > > 
> > > Oh nice, do you want to continue work on that? Or I can pick it up from your last commit at [ac83b99](https://github.com/GenericMappingTools/pygmt/commit/ac83b99799f88ab89e6412e7da3c6dff1a48ca7a).
> > 
> > 
> > I don't use 3-band image in my daily research, so have no knowledge about any conventions and technical details about tiff/geotiff. So feel free to push more changes to that branch.
> 
> No problem, I've been deep into looking at TIFF/GeoTIFF files recently at https://github.com/weiji14/cog3pio, so should be able to manage 3-bands 😄 Will open a separate PR for this later.

I've rebased the `datatypes/gmtimage` branch and opened a PR for it (#3128).

pygmt/src/grdcut.py
```diff
@@ -99,13 +91,13 @@ def grdcut(grid, **kwargs):
...
+        with (
+            lib.virtualfile_in(check_kind="raster", data=grid) as vingrd,
+            lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd,
+        ):
+            kwargs["G"] = voutgrd
+            lib.call_module(
+                module="grdcut", args=build_arg_string(kwargs, infile=vingrd)
+            )
+            return lib.virtualfile_to_grid(vfname=voutgrd, outgrid=outgrid)
```
weiji14, line 103: How should we handle `grdcut` prouducing either a 1-band grid or 3-band image? Should we just rename `virtualfile_to_grid` to `virtualfile_to_grid_or_image`, since the outputs will only be an `xarray.DataArray` or None?

seisman: It's more complicated. For the virtualfile_out method, we need to pass the correct kind (grid or image) to create the virtualfile, so we need to know the output kind before calling grdcut. See the changes of grdcut in https://github.com/GenericMappingTools/pygmt/pull/3115

weiji14: Yep, so we do `grid` in, `grid` out, and `image` in, `image` out. Continue discussion at https://github.com/GenericMappingTools/pygmt/pull/3115#discussion_r1535227527

pygmt/datatypes/grid.py
```diff
@@ -20,6 +23,45 @@
...
+    >>> _parse_nameunits("x")
+    ('x', None)
+    >>> _parse_nameunits("y")
+    ('y', None)
+    >>>
+    """
+    parts = nameunits.split("[")
+    long_name = parts[0].strip()
+    units = parts[1].strip("]").strip() if len(parts) > 1 else None
+    return long_name, units
```
weiji14, line 62: Is the split necessary? Is there whitespace before or after the square brackets?
```suggestion
    units = parts[1].strip("]") if len(parts) > 1 else None
    return long_name, units
```

pygmt/datatypes/grid.py
```diff
@@ -102,4 +144,224 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+                    newdims[dim] = "lon"
+                elif units == "degrees_north":
+                    attrs[dim]["standard_name"] = "latitude"
+                    newdims[dim] = "lat"
+            if dim == "z":
+                attrs[dim]["actual_range"] = np.array([header.z_min, header.z_max])
+            else:
+                attrs[dim]["axis"] = dim.upper()
+                idx = 0 if dim == "x" else 2
+                attrs[dim]["actual_range"] = np.array(header.wesn[idx : idx + 2])
```
weiji14, line 338: Could reuse some of this logic (and `_parse_nameunits`) in wrapping `GMT_IMAGE` at #3128, could we make a `_get_attrs` function out of this code block, and cherry-pick this into another PR? Or just the `_parse_nameunits` function will do.

pygmt/datatypes/grid.py
```diff
@@ -20,6 +23,45 @@
...
+    >>> _parse_nameunits("latitude [degrees_north]")
+    ('latitude', 'degrees_north')
+    >>> _parse_nameunits("x")
+    ('x', None)
+    >>> _parse_nameunits("y")
+    ('y', None)
+    >>>
+    """
+    parts = nameunits.split("[")
+    long_name = parts[0].strip()
```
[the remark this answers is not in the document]
seisman: I prefer to use `strip()` to make sure that the long_name doesn't have leading/trailing whitespaces.

pygmt/datatypes/grid.py
seisman: OK, I've made the `_parse_header` function which parses the grid header to get grid metadata that will be used for creating an xarray.DataArray object. 

The PR is at https://github.com/GenericMappingTools/pygmt/pull/3134.

pygmt/clib/session.py
```diff
@@ -1987,12 +1997,19 @@ def virtualfile_to_grid(
...
         if outgrid is not None:
             return None
-        return self.read_virtualfile(vfname, kind="grid").contents.to_dataarray()
+        if kind is None:  # Inquire the data family from the virtualfile
+            family = self.inquire_virtualfile(vfname)
+            kind = {
+                self["GMT_IS_GRID"]: "grid",
+                self["GMT_IS_IMAGE"]: "image",
+                self["GMT_IS_CUBE"]: "cube",
+            }[family]
```
weiji14: Can we use a match statement here like this? (untested)
```suggestion
            match family
                case self["GMT_IS_GRID"]:
                    kind = "grid"
                case self["GMT_IS_IMAGE"]:
                    kind = "image"
                case self["GMT_IS_CUBE"]:
                    kind = "cube"
```

seisman: Dict-lookup is faster than match statement (https://stackoverflow.com/questions/68476576/python-match-case-switch-performance).

seisman:
I think this PR is now ready for review. Generally speaking, this PR does three things:

1. Wrap `GMT_GRID` data structure
2. Add `Session.virtualfile_to_raster` method which can read a virtualfile content as `GMT_GRID`/`GMT_IMAGE`/`GMT_CUBE`
3. Refactor all wrappers to the virtualfiles for grid output and fix any failing tests

I think it makes more sense to split this big PR into three smaller PRs. Maybe after this PR is reviewed and approved.

pygmt/datatypes/grid.py
```diff
@@ -3,7 +3,197 @@
...
+        )
+
+        # Flip the coordinates and data if necessary so that coordinates are ascending.
+        # `grid.sortby(list(grid.dims))` sometimes causes crashes.
+        # The solution comes from https://github.com/pydata/xarray/discussions/6695.
+        for dim in grid.dims:
+            if grid[dim][0] > grid[dim][1]:
+                grid = grid.isel({dim: slice(None, None, -1)})
+
+        # Set the gmt accesssor.
```
yvonnefroehlich, line 195:
```suggestion
        # Set the gmt accessor.
```

weiji14:
> I think it makes more sense to split this big PR into three smaller PRs. Maybe after this PR is reviewed and approved.

Can probably keep it as one PR, I think you've done a good job of cherrypicking the GMT grid header and inquire virtualfile stuff out already, and the code here is reasonable to review.

I'll give this a proper final review after Thursday-ish as I'm attending the MIGARS conference this week (have already received thanks from a couple of PyGMT users here, so I'll pass that on 😃).

pygmt/clib/session.py
```diff
@@ -1946,6 +1951,70 @@ def virtualfile_to_dataset(
...
         return result  # pandas.DataFrame output
 
+    def virtualfile_to_raster(
+        self,
+        vfname: str,
+        kind: Literal["grid", "image", "cube", None] = None,
+        outgrid: str | None = None,
+    ) -> xr.DataArray | None:
+        """
+        Output a raster data stored in a virtual file to an :class:`xarray.DataArray`
```
weiji14, line 1961:
```suggestion
        Output raster data stored in a virtual file to an :class:`xarray.DataArray`
```

pygmt/clib/session.py
```diff
@@ -1946,6 +1951,70 @@ def virtualfile_to_dataset(
...
+        object.
+
+        The raster data can be a grid, an image or a cube.
+
+        Parameters
+        ----------
+        vfname
+            The virtual file name that stores the result grid.
```
weiji14, line 1969: Not implemented for image or cube yet, but document it first so we don't forget later.

```suggestion
            The virtual file name that stores the result grid/image/cube.
```

pygmt/datatypes/grid.py
```diff
@@ -3,7 +3,197 @@
...
+import xarray as xr
+from pygmt.datatypes.header import _GMT_GRID_HEADER, gmt_grdfloat
 
 
 class _GMT_GRID(ctp.Structure):  # noqa: N801
-    pass
+    """
+    GMT grid structure for holding a grid and its header.
+
+    The class is only meant for internal use and is not exposed to users. See the GMT
```
weiji14, line 17:
```suggestion
    This class is only meant for internal use and is not exposed to users. See the GMT
```

pygmt/datatypes/grid.py
```diff
@@ -3,7 +3,197 @@
...
+        grid = xr.DataArray(
+            data, coords=coords, name=header.name, attrs=header.data_attrs
+        )
+
+        # Flip the coordinates and data if necessary so that coordinates are ascending.
+        # `grid.sortby(list(grid.dims))` sometimes causes crashes.
+        # The solution comes from https://github.com/pydata/xarray/discussions/6695.
+        for dim in grid.dims:
+            if grid[dim][0] > grid[dim][1]:
+                grid = grid.isel({dim: slice(None, None, -1)})
```
weiji14, line 193: Could you elaborate a bit on this part? Sort ascending is probably needed for the x (longitude) dimension. But for the y (latitude) dimension, it is quite common to sort in descending order (y going from North Pole +90 to South Pole -90). What is the typical sorting order returned from GMT?

pygmt/datatypes/grid.py
```diff
@@ -3,7 +3,197 @@
...
+
+        # Set the gmt accessor.
+        # Must put at the end. The information get lost after specific grid operation.
```
weiji14, line 196:
```suggestion
        # Set GMT accessors.
        # Must put at the end, otherwise info gets lost after certain grid operations.
```

pygmt/tests/test_datasets_load_remote_datasets.py
```diff
@@ -35,8 +33,9 @@ def test_load_remote_dataset_benchmark_with_region():
     assert data.gmt.registration == 0
     assert data.shape == (11, 21)
     # The cpt attribute was added since GMT 6.4.0
-    if Version(__gmt_version__) >= Version("6.4.0"):
-        assert data.attrs["cpt"] == "@earth_age.cpt"
+    # Can't access the cpt attribute using virtual files
```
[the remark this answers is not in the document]
weiji14: Ok, so auto colormap won't work after this PR is merged? We'll need to remember to document this in the PyGMT v0.12.0 changelog.

pygmt/clib/session.py
```diff
@@ -1946,6 +1951,70 @@ def virtualfile_to_dataset(
             return result.to_numpy()
         return result  # pandas.DataFrame output
 
+    def virtualfile_to_raster(
```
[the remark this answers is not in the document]
weiji14: > Most GMT modules can write grids only, so we can set `kind="grid"` to avoid calling `Session.inquire_virtualfile`.

Yes, I think we can set `kind="grid"`, unless we know if GMT will allow passing GMT_IMAGE to certain grd* modules in the future. For example, maybe `grdclip` would allow GMT_IMAGE inputs at some point, and so we might want to set `kind=None` for future-proofing.

pygmt/tests/test_datasets_load_remote_datasets.py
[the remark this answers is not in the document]
seisman: Yes, it will break. We may take different actions for input netCDF files or xarray.DataArray object. For xarray.DataArray, we can check whether it has the `cpt` attribute (makes more sense in the `gmt` accessor, xref #499); For netCDF files, we need to read the file header (header only so it should be quick) and check the `cpt` attribute. If given, then we add the same attribute to the output xarray.DataArray/netCDF. But we need to be careful, for example, `grdedit -D+c` can change the CPT, so the output grid can have different grid for the input grid.

For GMT remote datasets, we probably can add a CPT attribue to the `GMTRemoteDataset` class (https://github.com/GenericMappingTools/pygmt/blob/70d666e38d6bfac4132fc66d3c7adbeb57dc9588/pygmt/datasets/load_remote_dataset.py#L67).

pygmt/clib/session.py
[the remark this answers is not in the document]
seisman: We need to do more refactoring if `grdclip` would support GMT_IMAGE in the future, since we need to check the kind of the input data (grid in, grid out or image in, image out), similar to what we have to do to `grdcut` (xref: https://github.com/GenericMappingTools/pygmt/pull/2398#discussion_r1535234139).

pygmt/datatypes/grid.py
seisman: I think GMT requires ascending order for longitude/latitude. Similar codes are already used at (The codes below use `sortby`, but we can't use it here due to a mysterious reason (xref https://github.com/GenericMappingTools/pygmt/pull/2398#discussion_r1527437896).)

https://github.com/GenericMappingTools/pygmt/blob/70d666e38d6bfac4132fc66d3c7adbeb57dc9588/pygmt/clib/conversion.py#L122-L124

I tried to find some evidence from the GMT source codes but failed because the codes are so complicated. Perhaps the most straightforward evidence is from following lines (https://github.com/GenericMappingTools/gmt/blob/e493ccc220d9e997abb655cb461142ed2e7cc777/src/gmt_nc.c#L754):

```
		/* Determine grid step */
		header->inc[GMT_X] = gmt_M_get_inc (GMT, dummy[0], dummy[1], header->n_columns, registration);
		if (gmt_M_is_dnan(header->inc[GMT_X]) || gmt_M_is_zero (header->inc[GMT_X])) header->inc[GMT_X] = 1.0;
		if (header->n_columns == 1) registration = GMT_GRID_PIXEL_REG;	/* The only way to have a grid like that */

		/* Extend x boundaries by half if we found pixel registration */
		if (registration == GMT_GRID_NODE_REG && header->registration == GMT_GRID_PIXEL_REG)
			header->wesn[XLO] = dummy[0] - header->inc[GMT_X] / 2.0,
			header->wesn[XHI] = dummy[1] + header->inc[GMT_X] / 2.0;
		else	/* Use as is */
			header->wesn[XLO] = dummy[0], header->wesn[XHI] = dummy[1];
```
in which `dummy[0]` and `dummy[1]` is the min/max value of the x vector. For pixel registration, GMT extends the x boundaries by half a grid. The codes only make sense when `head->inc[GMT_X]` is positive.

pygmt/clib/session.py
```diff
@@ -1946,6 +1951,70 @@ def virtualfile_to_dataset(
...
+        self,
+        vfname: str,
+        kind: Literal["grid", "image", "cube", None] = None,
```
seisman, line 1957: Perhaps it makes more sense to let `kind` default to `grid` instead, since very few modules can output `image`/`cube` anyway.

weiji14: Oh yes, that's a lot smarter actually.

pygmt/datatypes/grid.py
weiji14: Ok, from what I can tell, this means GMT expects grids with coordinates sorted in ascending order. But here in the `to_dataarray()` method, we're going from GMT virtualfile to `xarray.DataArray`, so is this sorting still needed? Maybe the sort ascending logic is only needed in `virtualfile_from_grid` (which is implemented in the `dataarray_to_matrix` conversion function already)?

pygmt/clib/session.py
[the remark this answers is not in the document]
weiji14: Ok, I meant if upstream GMT were to support passing `GMT_IMAGE` into `grdclip` in the future, but not sure when that would happen. Ok to just default to `kind='grid'` for now, since that's the current behaviour I think.

pygmt/datatypes/grid.py
seisman: > But here in the `to_dataarray()` method, we're going from GMT virtualfile to `xarray.DataArray`, so is this sorting still needed or can we return the grid as is? Maybe the sort ascending logic is only needed in `virtualfile_from_grid` (which is implemented in the `dataarray_to_matrix` conversion function already)?

Hmmm, your comments make more sense. I tried to remove these sorting codes locally. Here is the output:

```python
In [1]: import xarray as xr

In [2]: from pygmt import which

In [3]: path = which("@earth_relief_01d_g", download="a")

In [4]: grid1 = xr.load_dataarray(path)

In [6]: from pygmt.clib import Session
   ...: with Session() as lib:
   ...:     with lib.virtualfile_out(kind="grid") as voutgrd:
   ...:         lib.call_module("read", f"{path} {voutgrd} -Tg")
   ...:         grid = lib.read_virtualfile(voutgrd, kind="grid")
   ...:         grid2 = grid.contents.to_dataarray()
   ...: 

In [7]: grid1
Out[7]: 
<xarray.DataArray 'z' (lat: 181, lon: 361)> Size: 523kB
array([[ 2865. ,  2865. ,  2865. , ...,  2865. ,  2865. ,  2865. ],
       [ 3088. ,  3087.5,  3087. , ...,  3088.5,  3088. ,  3088. ],
       [ 3100.5,  3100.5,  3101. , ...,  3101.5,  3101. ,  3100.5],
       ...,
       [-3745.5, -3729. , -3722.5, ..., -3734. , -3742. , -3745.5],
       [-2940. , -2945. , -2951. , ..., -2895.5, -2921.5, -2940. ],
       [-3861. , -3861. , -3861. , ..., -3861. , -3861. , -3861. ]])
Coordinates:
  * lon      (lon) float64 3kB -180.0 -179.0 -178.0 -177.0 ... 178.0 179.0 180.0
  * lat      (lat) float64 1kB -90.0 -89.0 -88.0 -87.0 ... 87.0 88.0 89.0 90.0
Attributes:
    actual_range:  [-7174.  5350.]
    long_name:     elevation (m)

In [8]: grid2
Out[8]: 
<xarray.DataArray 'z' (lat: 181, lon: 361)> Size: 523kB
array([[-3861. , -3861. , -3861. , ..., -3861. , -3861. , -3861. ],
       [-2940. , -2945. , -2951. , ..., -2895.5, -2921.5, -2940. ],
       [-3745.5, -3729. , -3722.5, ..., -3734. , -3742. , -3745.5],
       ...,
       [ 3100.5,  3100.5,  3101. , ...,  3101.5,  3101. ,  3100.5],
       [ 3088. ,  3087.5,  3087. , ...,  3088.5,  3088. ,  3088. ],
       [ 2865. ,  2865. ,  2865. , ...,  2865. ,  2865. ,  2865. ]])
Coordinates:
  * lat      (lat) float64 1kB 90.0 89.0 88.0 87.0 ... -87.0 -88.0 -89.0 -90.0
  * lon      (lon) float64 3kB -180.0 -179.0 -178.0 -177.0 ... 178.0 179.0 180.0
Attributes:
    Conventions:   CF-1.7
    title:         SRTM15 Earth Relief v2.5.5 at 01 arc degree
    history:       
    description:   Reduced by Gaussian Cartesian filtering (314.5 km fullwidt...
    long_name:     elevation (m)
    actual_range:  [-7174.  5350.]
```
It seems `xr.load_dataarray` returns an xarray.DataArray with ascending lon/lat coodinates, but now GMT returns descending latitude. I think I was trying to make `GMT_GRID.to_dataaray()` return the exactly same xarray.DataArray object as `xr.load_dataarray`. But, maybe there are some misunderstandings here.

weiji14: > It seems `xr.load_dataarray` returns an xarray.DataArray with ascending lon/lat coodinates, but now GMT returns descending latitude. I think I was trying to make `GMT_GRID.to_dataaray()` return the exactly same xarray.DataArray object as `xr.load_dataarray`. But, maybe there are some misunderstandings here.

Hmm yeah, I don't think we need to match the behaviour of `xr.load_dataarray` here. I tried with `rioxarray.open_rasterio` and it gave descending latitude:

```python
import pygmt
import rioxarray
import xarray as xr

path = pygmt.which("@earth_relief_01d_g", download="a")
grid2 = rioxarray.open_rasterio(path)
# grid2
# <xarray.DataArray 'z' (band: 1, y: 181, x: 361)> Size: 131kB
# [65341 values with dtype=int16]
# Coordinates:
#   * band         (band) int64 8B 1
#   * x            (x) float64 3kB -180.0 -179.0 -178.0 ... 178.0 179.0 180.0
#   * y            (y) float64 1kB 90.0 89.0 88.0 87.0 ... -87.0 -88.0 -89.0 -90.0
#     spatial_ref  int64 8B 0
# Attributes: (12/20)
#     lat#actual_range:   [-90.  90.]
#     lat#axis:           Y
#     lat#long_name:      latitude
#     lat#standard_name:  latitude
#     lat#units:          degrees_north
#     lon#actual_range:   [-180.  180.]
#     ...                 ...
#     title:              SRTM15 Earth Relief v2.5.5 at 01 arc degree
#     actual_range:       [-7174.  5350.]
#     long_name:          elevation (m)
#     scale_factor:       0.5
#     _FillValue:         -32768
#     add_offset:         0.0
```

so we could probably remove the sorting in `to_dataarray()`, assuming this doesn't break any of the unit tests.

seisman: > I don't think we need to match the behaviour of `xr.load_dataarray` here.

Matching the behavior of `xr.load_dataarray` makes it easier to test the output of the virtualfiles using [`xr.DataArray.indentical`](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.identical.html).

weiji14: > > I don't think we need to match the behaviour of `xr.load_dataarray` here.
> 
> Matching the behavior of `xr.load_dataarray` makes it easier to test the output of the virtualfiles using [`xr.DataArray.identical`](https://docs.xarray.dev/en/latest/generated/xarray.DataArray.identical.html).

The returned attributes are not the same, so [`xr.testing.assert_identical`](https://docs.xarray.dev/en/v2024.02.0/generated/xarray.testing.assert_identical.html) will fail anyway.

seisman: Actually, we're using `xr.testing.assert_allclose` in the tests. If we don't sort the coordinates ascendingly, then we have to update all these tests.

weiji14: Hmm yeah, I just tried running `make test` locally after commenting out the sorting code, and got 19 test failures:

```python-traceback
================================================================================== short test summary info ===================================================================================
FAILED ../pygmt/datatypes/grid.py::pygmt.datatypes.grid._GMT_GRID.to_dataarray
FAILED ../pygmt/tests/test_dimfilter.py::test_dimfilter_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdclip.py::test_grdclip_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdfill.py::test_grdfill_dataarray_out - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdfill.py::test_grdfill_asymmetric_pad - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdfilter.py::test_grdfilter_dataarray_in_dataarray_out - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdgradient.py::test_grdgradient_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdhisteq.py::test_equalize_grid_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdlandmask.py::test_grdlandmask_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdproject.py::test_grdproject_no_outgrid[M10c] - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdproject.py::test_grdproject_no_outgrid[EPSG:3395 +width=10] - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdproject.py::test_grdproject_no_outgrid[+proj=merc +ellps=WGS84 +units=m +width=10] - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_grdsample.py::test_grdsample_dataarray_out - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_surface.py::test_surface_input_file - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_surface.py::test_surface_input_data_array - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_surface.py::test_surface_input_xyz - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_triangulate.py::test_regular_grid_no_outgrid - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_xyz2grd.py::test_xyz2grd_input_array[array] - AssertionError: Left and right DataArray objects are not close
FAILED ../pygmt/tests/test_xyz2grd.py::test_xyz2grd_input_array[Dataset] - AssertionError: Left and right DataArray objects are not close
======================================================== 19 failed, 657 passed, 4 skipped, 5 xfailed, 2 xpassed, 5 warnings in 58.94s ========================================================
```
Let's just keep the ascending sort for now then.

pygmt/tests/test_datasets_load_remote_datasets.py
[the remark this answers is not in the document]
weiji14: Ok, let's keep track about this `.cpt` attribute idea at https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2058129481.

weiji14 (approved):
Thanks again @seisman for the wonderful work here, from digging into the GMT C code and handling all the grid header parsing! There are still a few limitations such in regard to the CPT attribute and grdtype detection, as you've mentioned in the top post, but we can address that in the future. The speedup improvements will be worth it :smile:

PR #2403, opened 2023-03-07
 Allow getting the parameter API_IMAGE_LAYOUT for GMT>=6.4.0

seisman:

In GMT 6.3.0, the parameter `API_IAMGE_LAYOUT` is available only if GMT is built with GDAL support (https://github.com/GenericMappingTools/gmt/blob/6.3.0/src/gmt_api.c#L13458).

After PR https://github.com/GenericMappingTools/gmt/pull/6921, GDAL becomes GMT's required dependency. Thus, `API_IMAGE_LAYOUT` is also available (https://github.com/GenericMappingTools/gmt/blob/6.4.0/src/gmt_api.c#L13663).

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -168,13 +168,17 @@ class Session:
                 "cores": self.get_default("API_CORES"),
                 "grid layout": self.get_default("API_GRID_LAYOUT"),
             }
-            # API_BIN_VERSION is new in GMT 6.4.0.
             # For GMT<6.4.0, API_IMAGE_LAYOUT is not defined if GMT is not
             # compiled with GDAL. Since GMT 6.4.0, GDAL is a required GMT
-            # dependency.
+            # dependency. The try-except block can be refactored after we bump
+            # the minimum required GMT version to 6.4.0.
+            try:
+                self._info["image layout"] = self.get_default("API_IMAGE_LAYOUT")
+            except GMTCLibError:
+                pass
+            # API_BIN_VERSION is new in GMT 6.4.0.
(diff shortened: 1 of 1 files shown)


pygmt/clib/session.py
```diff
@@ -166,13 +166,15 @@ def info(self):
...
-                # "image layout": self.get_default("API_IMAGE_LAYOUT"),
                 "grid layout": self.get_default("API_GRID_LAYOUT"),
             }
             # API_BIN_VERSION is new in GMT 6.4.0.
+            # For GMT<6.4.0, API_IMAGE_LAYOUT is not defined if GMT is not
+            # compiled with GDAL. Since GMT 6.4.0, GDAL is a required GMT
+            # dependency.
             if Version(self._info["version"]) >= Version("6.4.0"):
                 self._info["binary version"] = self.get_default("API_BIN_VERSION")
+                self._info["image layout"] = self.get_default("API_IMAGE_LAYOUT")
```
weiji14, line 177: It sounds like `API_IMAGE_LAYOUT` would be available in GMT 6.3.0 if the GMT was built with GDAL (which is the case for the [conda-forge GMT package](https://github.com/conda-forge/gmt-feedstock)). Maybe use something like a try-except instead?

weiji14, line 177: Can you update the API_* list at L435-444 as well?

seisman: Done in 5cdabe64.

weiji14 (approved):

michaelgrund (approved):

PR #2434, opened 2023-03-17
Figure.timestamp: Improve documentation and gallery example

seisman:

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -20,7 +20,7 @@ fig.show()
 
 ###############################################################################
 # Additionally, a custom label can be added via the ``label`` parameter. The
-# label font can be defined via the ``font`` parameter and the timestamp string
+# font can be defined via the ``font`` parameter and the timestamp string
 # format via ``timefmt``.
 
 os.environ["TZ"] = "Pacific/Honolulu"  # optionally set the time zone
--- a/examples/gallery/embellishments/timestamp.py
+++ b/examples/gallery/embellishments/timestamp.py
@@ -1,16 +1,15 @@
 """
 Timestamp
 ---------
-The :meth:`pygmt.Figure.timestamp` method can draw the
-GMT timestamp logo on the figure. The timestamp
-will always be shown relative to the bottom-left corner
-of the figure. By default, the ``offset`` and
-``justification`` parameters are set to
-``("-54p", "-54p")`` (x, y directions) and ``"BL"``
-(bottom-left), respectively.
+The :meth:`pygmt.Figure.timestamp` method can draw the GMT timestamp logo on
+the plot. The timestamp will always be shown relative to the bottom-left corner
+of the plot. By default, the ``offset`` and ``justification`` parameters are
+set to ``("-54p", "-54p")`` (x, y directions) and ``"BL"`` (bottom-left),
+respectively.
 """
-
 # sphinx_gallery_thumbnail_number = 2
+
+import os
 
 import pygmt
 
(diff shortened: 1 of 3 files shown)


pygmt/src/timestamp.py
```diff
@@ -23,27 +23,38 @@ def timestamp(
     r"""
     Plot the GMT timestamp logo.
 
+    Add the GMT timestamp logo with an optional label at the bottom-left corner
+    of a plot. The timestamp will be in the locale set by the environment
+    variable **TZ** (generally local time) and its foramt is controlled by the
```
michaelgrund, line 28:
```suggestion
    variable **TZ** (generally local time) and its format is controlled by the
```

pygmt/src/timestamp.py
```diff
@@ -23,27 +23,38 @@ def timestamp(
     r"""
     Plot the GMT timestamp logo.
 
+    Add the GMT timestamp logo with an optional label at the bottom-left corner
+    of a plot. The timestamp will be in the locale set by the environment
+    variable **TZ** (generally local time) and its format is controlled by the
```
michaelgrund, line 28:
```suggestion
    variable **TZ** via ``os.environ["TZ"]`` (generally local time) and its format is controlled by the
```
As is it's not clear how to change that variable (via os).

michaelgrund (approved):
Except the suggestion it's looks good.

pygmt/src/timestamp.py
seisman: I'm unsure about this change, because it's also possible to set TZ using `export TZ=xxx` or other ways.

michaelgrund: I see, however, I feel it's not fully clear without an expanded description.

seisman: OK, I've changed it to 

> The timestamp will be in the locale set by the environment variable **TZ** (generally local time but can be changed via ``os.environ["TZ"]``)

examples/gallery/embellishments/timestamp.py
```diff
@@ -24,12 +23,14 @@
 # label font can be defined via the ``font`` parameter and the timestamp string
 # format via ``timefmt``.
```
yvonnefroehlich, line 24: Strictly speaking the font set via the `font` parameter impacts on both the timestamp string and the addional label.

```suggestion
# font can be defined via the ``font`` parameter and the timestamp string
# format via ``timefmt``.
```

pygmt/src/timestamp.py
```diff
@@ -23,27 +23,39 @@ def timestamp(
     r"""
     Plot the GMT timestamp logo.
 
+    Add the GMT timestamp logo with an optional label at the bottom-left corner
+    of a plot. The timestamp will be in the locale set by the environment
+    variable **TZ** (generally local time but can be changed via
+    ``os.environ["TZ"]``) and its format is controlled by the parameter
+    ``timefmt``. It can also be replaced with any custom text string using the
+    ``text`` parameter.
+
```
yvonnefroehlich, line 32: I am wondering, whether we should mention that there is by default an offset of `("-54p", "-54p")`?
```suggestion
    Add the GMT timestamp logo with an optional label at the bottom-left corner
    of a plot with an ``offset`` of ``("-54p", "-54p")``. The timestamp will be
    in the locale set by the environmentvariable **TZ** (generally local time
    but can be changed via ``os.environ["TZ"]``) and its format is controlled
    by the parameter ``timefmt``. It can also be replaced with any custom
    text string using  the ``text`` parameter.

```

yvonnefroehlich (approved):

Issue #1467, opened 2021-08-26
Optional 'drape_input' in pygmt.Figure.grdview


NikosAlexandris:
An idea: add a  `drape_input=` option `pygmt.Figure.grdview()` -- this relates to the discussions in https://github.com/GenericMappingTools/gmt/issues/5631. Perhaps something like `drape_red=`, `drape_green=` and `drape_blue=` which should eventually be mutually exclusive with `drape_input=`.

A general comment: GMT is super-nice! Yet one thing I find hard is its non-declarative names of options and flags and else. In my humble view, fully spelling out names of options, parametes and flags and else, is best towards human-readable code.

weiji14:
Hi @NikosAlexandris, really sorry for the late reply! Just a note that PyGMT's [`fig.grdview`](https://www.pygmt.org/v0.4.1/api/generated/pygmt.Figure.grdview.html#pygmt.Figure.grdview) does have a `drapegrid` parameter which should be the same as the `drape_input` option you mentioned (specifically, `-G` in GMT grdview), made available since #330 (PyGMT v0.1.0)! This `drapegrid=` parameter should support both `xarray.DataArray` and `file` inputs (though `file` inputs might be more stable from experience).

If your :sparkles: feature request is related to supporting `drape_red=`, `drape_green=` and `drape_blue=` in PyGMT, then well, we'll need to wait for the outcome of the discussion at https://github.com/GenericMappingTools/gmt/issues/5631. Specifically, whether passing in separate RGB bands to `drapegrid`/`-G` is future-proof. Ping @meghanrjones who might know more.

In the meantime, I'm working on `grdmix` at #1437 which would be able to combine Red/Green/Blue grids (e.g. GeoTiffs) into a single image that could then be passed into `grdview`'s `drapegrid=` parameter (due for PyGMT v0.5.0). If you can describe your specific use-case (e.g. what satellite products are you using? are you combining GeoTiffs directly, or an xarray.Dataset?), that will be very helpful for us to inform the code implementation details.

> A general comment: GMT is super-nice! Yet one thing I find hard is its non-declarative names of options and flags and else. In my humble view, fully spelling out names of options, parametes and flags and else, is best towards human-readable code.

Yes, GMT is the product of ~30 years of work! Most of the functionality is there already, and I guess it's up to us to make it usable for the masses :slightly_smiling_face:

meghanrjones:
> 
[...]

PR #2493, opened 2023-04-10
Better handling of optional virtual files (e.g., shading in Figure.grdimage)

seisman:
### Background

Some GMT modules (e.g., `grdimage` and `grdview`) can accept a primary input file and an optional input file. Here, I'll take `grdimage` as an example. 

`grdimage` can accept an input grid file and an optional intensity grid for shading (`-I` option). Thus, in PyGMT, `Figure.grdimage`'s `shading` parameter can be given in many different ways:

1. `shading=None`: meaning no shading
2. `shading=True` or `shading=False`: shading or no shading
3. `shading=0.5`: a constant intensity to apply everywhere
4. `shading="+a30+nt0.8"`: derive the intensity from the primary input grid
5. `shading="intensity.nc"`: an intensity grid file
6. `shaidng="intensity.nc+a30+nt0.8"`: an intensity grid file plus more parameters
7. `shading=xrgrid` in which `xrgrid` is an xarray.DataArray object
8. more?

For cases 1-6, the `shading` parameter can be directly processed by the `build_arg_string` function, but for case 7, we need to create a virtual file for the xarray.DataArray grid.

### Motivations

Since the virtual file for shading depends on the value and data type of the `shading` parameter, thus we have to use [contextlib.ExitStack](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack), which is designed to make it easy to work with optional context manager. However, using `contextlib.ExitStack` still makes the codes complicated and also make it hard to reuse the codes:

https://github.com/GenericMappingTools/pygmt/blob/d2b0c39b075c41f55f67f1a9daa94365f3c6fce1/pygmt/src/grdimage.py#L182-L195

The main goal of this PR is to simplify the above codes to:
```python
    with Session() as lib:
        shade_context = lib.virtualfile_from_data(check_kind="raster", data=kwargs.get("I"))
        with file_context as fname, shade_context as shade_fname:
            kwargs["I"] = shade_fname
            lib.call_module(
                module="grdimage", args=build_arg_string(kwargs, infile=fname)
            )
```
which is more compact and more intuitive to read.

### Implementations

To achieve this, the `virtualfile_from_data` context manager must work for all cases. It's actually very simple. If `data` is not an xarray.DataArray object (cases 1-6), `virtualfile_from_data` should simply return a dummy context manager which does nothing but yield the argument passed to it (see the [`dummy_context`](https://github.com/GenericMappingTools/pygmt/blob/d2b0c39b075c41f55f67f1a9daa94365f3c6fce1/pygmt/helpers/utils.py#L96) function but also see https://github.com/GenericMappingTools/pygmt/pull/2491). This is done by changing the `data_kind` function to take arguments like None, True, False, 0.5, `"+a30+nt0.8"`, `"intensity.nc+a30+nt0.8"` as a "file" (see the changes in the `data_kind` function).

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -5,7 +5,6 @@ access to the API functions.
 Uses ctypes to wrap most of the core functions from the C API.
 """
 import ctypes as ctp
-import pathlib
 import sys
 from contextlib import contextmanager, nullcontext
 
@@ -1536,19 +1535,21 @@ class Session:
             data, x, y, z, required_z=required_z, optional_data=optional_data
         )
 
-        if check_kind == "raster" and kind not in ("file", "grid"):
+        if check_kind == "raster" and kind not in ("file", "grid", "null"):
             raise GMTInvalidInput(f"Unrecognized data type for grid: {type(data)}")
         if check_kind == "vector" and kind not in (
             "file",
             "matrix",
             "vectors",
             "geojson",
+            "null",
         ):
             raise GMTInvalidInput(f"Unrecognized data type for vector: {type(data)}")
 
         # Decide which virtualfile_from_ function to use
         _virtualfile_from = {
             "file": nullcontext,
+            "null": nullcontext,
             "geojson": tempfile_from_geojson,
             "grid": self.virtualfile_from_grid,
             # Note: virtualfile_from_matrix is not used because a matrix can be
@@ -1559,11 +1560,11 @@ class Session:
         }[kind]
 
         # Ensure the data is an iterable (Python list or tuple)
-        if kind in ("geojson", "grid"):
+        if kind in ("geojson", "grid", "null"):
             _data = (data,)
         elif kind == "file":
-            # Useful to handle `pathlib.Path` and string file path alike
-            _data = (str(data),) if isinstance(data, pathlib.PurePath) else (data,)
+            # Useful to handle `pathlib.PurePath` and string file path alike
+            _data = (str(data),)
         elif kind == "vectors":
             _data = [np.atleast_1d(x), np.atleast_1d(y)]
             if z is not None:
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1537,13 +1537,14 @@ class Session:
             data, x, y, z, required_z=required_z, required_data=required_data
         )
 
-        if check_kind == "raster" and kind not in ("file", "grid"):
+        if check_kind == "raster" and kind not in ("file", "grid", "arg"):
             raise GMTInvalidInput(f"Unrecognized data type for grid: {type(data)}")
         if check_kind == "vector" and kind not in (
             "file",
             "matrix",
             "vectors",
             "geojson",
+            "arg",
         ):
             raise GMTInvalidInput(f"Unrecognized data type for vector: {type(data)}")
 
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1537,10 +1537,10 @@ class Session:
             data, x, y, z, required_z=required_z, required_data=required_data
         )
 
-        if check_kind == "raster" and kind not in ("file_or_arg", "grid"):
+        if check_kind == "raster" and kind not in ("file", "grid"):
             raise GMTInvalidInput(f"Unrecognized data type for grid: {type(data)}")
         if check_kind == "vector" and kind not in (
-            "file_or_arg",
+            "file",
             "matrix",
             "vectors",
             "geojson",
@@ -1549,7 +1549,7 @@ class Session:
 
         # Decide which virtualfile_from_ function to use
         _virtualfile_from = {
-            "file_or_arg": nullcontext,
+            "file": nullcontext,
             "geojson": tempfile_from_geojson,
             "grid": self.virtualfile_from_grid,
             # Note: virtualfile_from_matrix is not used because a matrix can be
@@ -1560,7 +1560,7 @@ class Session:
         }[kind]
 
         # Ensure the data is an iterable (Python list or tuple)
-        if kind in ("geojson", "grid", "file_or_arg"):
+        if kind in ("geojson", "grid", "file"):
             _data = (data,) if not isinstance(data, pathlib.PurePath) else (str(data),)
         elif kind == "vectors":
             _data = [np.atleast_1d(x), np.atleast_1d(y)]
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1542,7 +1542,9 @@ class Session:
         elif check_kind == "vector":
             valid_kinds = ("file", "matrix", "vectors", "geojson")
         else:
-            valid_kinds = ()
+            raise GMTInvalidInput(
+                "Invalid check_kind. Should be either 'raster' or 'vector'."
+            )
         if required_data is False:
             valid_kinds += ("arg",)
         if kind not in valid_kinds:
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1475,7 +1475,7 @@ class Session:
         z=None,
         extra_arrays=None,
         required_z=False,
-        optional_data=False,
+        required_data=True,
     ):
         """
         Store any data inside a virtual file.
@@ -1500,7 +1500,7 @@ class Session:
             All of these arrays must be of the same size as the x/y/z arrays.
         required_z : bool
             State whether the 'z' column is required.
-        optional_data : bool
+        required_data : bool
             State whether the 'data' is optional.
 
         Returns
@@ -1533,7 +1533,7 @@ class Session:
         <vector memory>: N = 3 <7/9> <4/6> <1/3>
         """
         kind = data_kind(
-            data, x, y, z, required_z=required_z, optional_data=optional_data
+            data, x, y, z, required_z=required_z, required_data=required_data
         )
 
         if check_kind == "raster" and kind not in ("file_or_arg", "grid"):
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -5,6 +5,7 @@ access to the API functions.
 Uses ctypes to wrap most of the core functions from the C API.
 """
 import ctypes as ctp
+import pathlib
 import sys
 from contextlib import contextmanager, nullcontext
 
@@ -1535,21 +1536,19 @@ class Session:
             data, x, y, z, required_z=required_z, optional_data=optional_data
         )
 
-        if check_kind == "raster" and kind not in ("file", "grid", "null"):
+        if check_kind == "raster" and kind not in ("file_or_arg", "grid"):
             raise GMTInvalidInput(f"Unrecognized data type for grid: {type(data)}")
         if check_kind == "vector" and kind not in (
-            "file",
+            "file_or_arg",
             "matrix",
             "vectors",
             "geojson",
-            "null",
         ):
             raise GMTInvalidInput(f"Unrecognized data type for vector: {type(data)}")
 
         # Decide which virtualfile_from_ function to use
         _virtualfile_from = {
-            "file": nullcontext,
-            "null": nullcontext,
+            "file_or_arg": nullcontext,
             "geojson": tempfile_from_geojson,
             "grid": self.virtualfile_from_grid,
             # Note: virtualfile_from_matrix is not used because a matrix can be
@@ -1560,11 +1559,8 @@ class Session:
         }[kind]
 
         # Ensure the data is an iterable (Python list or tuple)
-        if kind in ("geojson", "grid", "null"):
-            _data = (data,)
-        elif kind == "file":
-            # Useful to handle `pathlib.PurePath` and string file path alike
-            _data = (str(data),)
+        if kind in ("geojson", "grid", "file_or_arg"):
+            _data = (data,) if not isinstance(data, pathlib.PurePath) else (str(data),)
         elif kind == "vectors":
             _data = [np.atleast_1d(x), np.atleast_1d(y)]
             if z is not None:
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1501,8 +1501,8 @@ class Session:
         required_z : bool
             State whether the 'z' column is required.
         required_data : bool
-            State whether 'data' is required (useful for dealing with optional
-            virtual files).
+            Set to True when 'data' is required, or False when dealing with
+            optional virtual files. [Default is True].
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


seisman:
Currently, all the `grdimage` tests still pass, which means the new codes work, but some other tests (e.g., most `plot` tests) fail likely caused by changes in `data_kind`. So need to more careful when changing `data_kind`.

pygmt/helpers/utils.py
```diff
@@ -14,38 +14,43 @@
...
+        {table-classes} or an option argument.
     x/y : 1-D arrays or None
         x and y columns as numpy arrays.
     z : 1-D array or None
         z column as numpy array. To be used optionally when x and y are given.
     required_z : bool
         State whether the 'z' column is required.
+    optional_data : bool
+        State whether 'data' is optional (useful for dealing with optional
+        virtual files).
```
weiji14, line 47: Currently the function has `required_z` and `optional_data` parameters. Should we use `required_z` and `required_data` instead? Or `optional_z` and `optional_data`?

pygmt/helpers/utils.py
```diff
@@ -61,23 +66,41 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     'file'
     >>> data_kind(data=pathlib.Path("my-data-file.txt"), x=None, y=None)
     'file'
+    >>> data_kind(data=None, x=None, y=None, optional_data=True)
+    'null'
+    >>> data_kind(data=2.0, x=None, y=None, optional_data=True)
+    'null'
+    >>> data_kind(data=True, x=None, y=None, optional_data=True)
+    'null'
```
weiji14, line 74: Somewhat off-topic, but thinking about whether we need a doctest for the `required_z` parameter here instead of just at https://github.com/GenericMappingTools/pygmt/blob/59a203579a2b896b0d721fa88b65d7b97a5115f3/pygmt/tests/test_clib.py#L427-L453

pygmt/helpers/utils.py
```diff
@@ -61,23 +66,41 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
...
     >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
     'grid'
     """
-    if data is None and x is None and y is None:
-        raise GMTInvalidInput("No input data provided.")
-    if data is not None and (x is not None or y is not None or z is not None):
-        raise GMTInvalidInput("Too much data. Use either data or x and y.")
-    if data is None and (x is None or y is None):
+    # pylint: disable=too-many-branches
+    # validate the combinations of data/x/y/z
```
weiji14, line 79: Kinda wishing that we can use switch statements here, but that's only Python 3.10 and above :upside_down_face: 

- https://stackoverflow.com/questions/11479816/what-is-the-python-equivalent-for-a-case-switch-statement
- https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching
- https://peps.python.org/pep-0634/

pygmt/helpers/utils.py
```diff
@@ -61,23 +66,41 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
...
+            raise GMTInvalidInput("Too much data. Use either data or x and y.")
+        if required_z and z is None:
+            raise GMTInvalidInput("Must provide x, y, and z.")
 
+    # determine the data kind
     if isinstance(data, (str, pathlib.PurePath)):
         kind = "file"
+    elif optional_data and (data is None or isinstance(data, (bool, int, float))):
+        # a nullcontext will be created for "null" kind
+        kind = "null"
```
weiji14, line 98: So for `shading="+a30+nt0.8"` (a str type argument), the returned kind will be `file` instead of `null`. Technically both `file` and `null` returns a nullcontext from `virtualfile_from_data` so it shouldn't really matter, but it makes me wonder if we should combine the `file` and `null` kind, or be a bit more explicit somehow, otherwise this `data_kind` function will get confusing to new maintainers.

weiji14 (commented):
A few comments for now, still need to wrap my head around the main nested if-then statement in the `data_kind` function.

pygmt/helpers/utils.py
seisman: Yes, we can revisit many if-else clauses after we bump to Python>=3.10.

pygmt/helpers/utils.py
seisman: Actually, I'm thinking about if we should remove `required_z` and use `required_ncols` instead. For example, in `Figure.plot()`, sometimes we need more than three columns (e.g., x, y, size, z_for_color).

seisman:
> still need to wrap my head around the main nested if-then statement in the `data_kind` function.

I'm quite satisfied with the changes in `grdimage.py` and `grdview.py`, but I also find that codes in `data_kind` is not that easy to understand. So better ideas about `data_kind` are welcomed.

pygmt/helpers/utils.py
weiji14: Right, for #1076/#1117 you mean? Should probably do that in a separate PR.

pygmt/helpers/utils.py
seisman: > So for `shading="+a30+nt0.8"` (a str type argument), the returned kind will be `file` instead of `null`.

Yes, maybe we can rename `file` to `file_or_arg`?

pygmt/helpers/utils.py
seisman: What are the valid values in the list? `color`, `size`, `symbol` and more? I think it will make the codes more complicated.

pygmt/src/grdview.py
```diff
@@ -154,24 +152,19 @@ def grdview(self, grid, **kwargs):
...
+            f"Unrecognized data type for drapegrid: {type(kwargs['G'])}"
+        )
+
+    with Session() as lib:
+        with lib.virtualfile_from_data(
+            check_kind="raster", data=grid
+        ) as fname, lib.virtualfile_from_data(
+            check_kind="raster", data=kwargs.get("G"), optional_data=True
+        ) as drapegrid:
+            kwargs["G"] = drapegrid
```
weiji14, line 167: Slightly relevant, how would we handle RGB xarray.DataArray inputs to drapegrid `-G` as mentioned at #1467? Currently [GMT 6.4.0](https://docs.generic-mapping-tools.org/6.4/grdview.html#g) requires passing it in as `-Ggrd_r -Ggrd_g -Ggrd_b`, but I'm wondering if we should only support just a single `xarray.DataArray` of shape (3, Y, X) instead.

pygmt/helpers/utils.py
weiji14: Naming is hard :upside_down_face: Yeah, let's go with `file_or_arg`.

pygmt/src/grdview.py
seisman: > we should only support just a single `xarray.DataArray` of shape (3, Y, X) instead.

Sounds good to me.

pygmt/helpers/utils.py
seisman: Done in 9a43dc8edab2b5b0f907b84e0ff740cda2c4c80e.

pygmt/helpers/utils.py
```diff
@@ -58,26 +63,43 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
...
     >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
     'grid'
     """
-    if data is None and x is None and y is None:
-        raise GMTInvalidInput("No input data provided.")
-    if data is not None and (x is not None or y is not None or z is not None):
-        raise GMTInvalidInput("Too much data. Use either data or x and y.")
-    if data is None and (x is None or y is None):
+    # pylint: disable=too-many-branches
+    # validate the combinations of data/x/y/z
```
weiji14, line 79: > but I also find that codes in data_kind is not that easy to understand. So better ideas about data_kind are welcomed.

How about we split the function into two? A `_validate_data_input` for the first half that checks that there's no missing or invalid data, and then the second half can stay in `data_kind` and that checks the instance type of the dataset.

seisman: Sounds a good idea. Will do later.

pygmt/helpers/utils.py
seisman: > Currently the function has `required_z` and `optional_data` parameters. Should we use `required_z` and `required_data` instead? Or `optional_z` and `optional_data`?

Renamed `optional_data` to `required_data` in 932d07d5abea788b3ca4aa0a526fe124d01f3020.

weiji14 (commented):
Some documentation fixes

pygmt/helpers/utils.py
```diff
@@ -61,11 +62,11 @@ def _validate_data_input(
     """
     if data is None:  # data is None
         if x is None and y is None:  # both x and y are None
-            raise GMTInvalidInput("No input data provided.")
+            if required_data:  # data is not optional
+                raise GMTInvalidInput("No input data provided.")
```
seisman, line 66: The logical here is still wrong, leading to ~50 failures.

pygmt/helpers/utils.py
seisman: > So for `shading="+a30+nt0.8"` (a str type argument), the returned kind will be `file` instead of `null`.

Technically speaking, `"+a30+nt0.8"` is a valid file name, so a `file` kind still makes sense. 

The `data_kind` function is also being used in many module wrappers. For example, in `legend`,

https://github.com/GenericMappingTools/pygmt/blob/6764554c84e9b4c328b3b41d60c8337fa0f18373/pygmt/src/legend.py#L78-L83

`specfile` can be either `None` or a "file" (string type), but can't be in bool/int/float type. Thus, instead of having a single `file_or_arg` kind, we may need to have more kinds like `file` (str or PurePath), `null` (None), `bool` (True or False), `value` (int/float).

weiji14: Ok, maybe just keeping `"file"` kind is good rather than `"file_or_arg"`, and we can add the `"null"` kind back? Not sure if adding separate kinds for `bool` or `value` is needed, since all of these would just return a nullcontext.

pygmt/helpers/utils.py
seisman: Done in 73c31e214

pygmt/helpers/utils.py
seisman: Have two different kinds `file` and `arg` in commit 5faa05dc27b458e7ba675e02704faefcf05d9086

seisman:
This PR is now completed and ready for review.

pygmt/clib/session.py
```diff
@@ -1474,6 +1475,7 @@ def virtualfile_from_data(
         z=None,
         extra_arrays=None,
         required_z=False,
+        required_data=True,
```
yvonnefroehlich, line 1478: Following the [Code Style](https://www.pygmt.org/dev/contributing.html#code-style), no underscore should be used for words bridged only by consonants. However, already `required_z` and `check_kind` contain underscores...
```suggestion
        required_z=False,
        requireddata=True,
```

seisman: These are internal functions for developers/maintainers. So maybe we don't have to strictly follow the code style?

weiji14: That code style is meant for GMT aliases, so we can use `required_data` here.

pygmt/clib/session.py
```diff
@@ -1528,21 +1533,29 @@ def virtualfile_from_data(
...
+        )
+
+        if check_kind == "raster":
+            valid_kinds = ("file", "grid")
+        elif check_kind == "vector":
+            valid_kinds = ("file", "matrix", "vectors", "geojson")
+        else:
+            raise GMTInvalidInput(
+                "Invalid check_kind. Should be either 'raster' or 'vector'."
+            )
```
weiji14, line 1547: Debating on whether to remove this else argument. Technically this `lib.virtualfile_from_data` function is for internal use only, but the function name is not marked as private with a leading underscore (`_virtualfile_from_data`). Should we assume that others are using this and might get an error with the default `check_kind=None`? Or add a test for `check_kind=None` so that there is code coverage?

pygmt/helpers/utils.py
```diff
@@ -81,38 +106,43 @@ def _validate_data_input(
...
+        {table-classes} or an option argument.
     x/y : 1-D arrays or None
         x and y columns as numpy arrays.
     z : 1-D array or None
         z column as numpy array. To be used optionally when x and y are given.
     required_z : bool
         State whether the 'z' column is required.
+    required_data : bool
+        State whether 'data' is required (useful for dealing with optional
+        virtual files).
```
weiji14, line 139: Similar suggestion to above, feel free to reword.
```suggestion
        Set to True when 'data' is required, or False when dealing with
        optional virtual files. [Default is True].
```

pygmt/clib/session.py
```diff
@@ -1528,21 +1533,29 @@ def virtualfile_from_data(
...
-        ):
-            raise GMTInvalidInput(f"Unrecognized data type for vector: {type(data)}")
+        kind = data_kind(
+            data, x, y, z, required_z=required_z, required_data=required_data
+        )
+
+        if check_kind == "raster":
+            valid_kinds = ("file", "grid")
+        elif check_kind == "vector":
+            valid_kinds = ("file", "matrix", "vectors", "geojson")
```
weiji14, line 1543: Suggested refactoring. Makes it a little clearer that raster includes "grid", and vector includes "matrix", "vectors", "geojson".

```suggestion
        valid_kinds = ("file", "arg") if required_data is False else ("file",)
        if check_kind == "raster":
            valid_kinds += ("grid",)
        elif check_kind == "vector":
            valid_kinds += ("matrix", "vectors", "geojson")
```

pygmt/clib/session.py
```diff
@@ -1528,21 +1533,29 @@ def virtualfile_from_data(
...
+        else:
+            raise GMTInvalidInput(
+                "Invalid check_kind. Should be either 'raster' or 'vector'."
+            )
+        if required_data is False:
+            valid_kinds += ("arg",)
```
weiji14, line 1549:
```suggestion

```

weiji14 (commented):
Nice work getting the if-then logic correct! Still a bit hard to follow in some parts, but I think this is the best that we can do. Just some more minor comments/suggestions and should be close to merge.

pygmt/clib/session.py
seisman: The docstring says:
```
Used to validate the type of data that can be passed in. Choose     
from 'raster', 'vector', or None. Default is None (no validation).
```
so we should not check the data type if check_kind is None. See 9b31e230fa0f4640e3a0a5bd12f5aad1c58936c0 for the changes.

weiji14: Sounds good. Could you also change the docstring for the expected type at L1489 to `check_kind : str or None` please.

seisman: Done in 0ed6248bb76bda675c5d87c51b123b48860b7213

weiji14 (approved):
Excellent, thanks @seisman! Will be easier to work on #2590 after this PR is merged :smile:

pygmt/clib/session.py
yvonnefroehlich: OK.

Issue #2401, opened 2023-03-07
Figure.shift_origin: Shifting plotting origins


seisman:
The `Figure.shift_origin` method is a high-level wrapper for GMT's common option `-X`/`-Y`. With this method, we should disallow `-X`/`-Y` option in all plotting methods. 

### The syntax of `-X`/`-Y`

The full syntax of GMT's `-X` and `-Y` options is (https://docs.generic-mapping-tools.org/dev/gmt.html#xy-full): 

`-X[a|c|f|r][xshift]`

As mentioned in the upstream issue https://github.com/GenericMappingTools/gmt/issues/7296, `-Xc` and `-Xf` make no sense in modern mode. So, we only focus on `-Xr` (the same as `-X`) and `-Xa` in PyGMT.

- `-Xr`: move the origin relative to its current location.                      
- `-Xa`: move the origin relative to its current location, and move the origin back to the original position after plotting

The following GMT bash script shows the behavior of `-Xr` and `-Xa`:
```
gmt begin                                                                       
for mode in "r" "a"; do                                                         
    echo "Shift plotting origin with -X${mode}10c"                              
    gmt figure shift-X${mode}10c png                                            
    gmt basemap -R0/10/0/10 -JX10c -Baf -B+tFrame1                              
    gmt basemap -R0/10/0/10 -JX10c -Baf -B+tFrame2 -X${mode}10c                 
    gmt basemap -R0/5/0/5 -JX5c -Baf -B+tFrame3                                 
done                                                                            
gmt end show 
```
| `-Xr` | `-Xa` |
|---|---|
|![image](https://user-images.githubusercontent.com/3974108/223334577-5041fc70-8c3f-46c6-a898-557b03adf828.png) |![image](https://user-images.githubusercontent.com/3974108/223334618-37f4ed28-d464-4f0a-8ed9-41c2f467b28e.png)|

### Implementation of `Figure.shift_origin`

The `Figure.shift_origin` method should support both `-Xr` (shifting origins permanently) and `-Xa` (shiftiing origins temporarily). So, the most intuitive syntax is:
```
# Permanently shift the plotting origin
Figure.shift_origin(xshift="1c")  

# Temporarily shift the plotting origin
with Figure.shift_origin(xshift="1c"):
    ...
    ...
```

### Issues to track
 
- [X] Implement the permanent shift: `Figure.shift_origin()` #289
[...]

PR #2500, opened 2023-04-18
clib.Session: Wrap the GMT_Get_Common API function

seisman:

Wrap the `GMT_Get_Common` API function. 

- API documentation: https://docs.generic-mapping-tools.org/dev/devdocs/api.html#inquiring-about-the-gmt-common-options
- C source code: https://github.com/GenericMappingTools/gmt/blob/35be2deca809ba32c6fef89a0833f6b1f5ddebbc/src/gmt_api.c#L13703

Wrapping this function is a necessary step toward addressing other issues/enhancements:

- [ ] https://github.com/GenericMappingTools/pygmt/issues/2401 and 
- [ ] Let `Figure.tilemap` support syntax like `region="JP"` https://github.com/GenericMappingTools/pygmt/pull/2394#discussion_r1125417383

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -516,7 +516,7 @@ class Session:
             the return value depends on the choice of the option.
 
             - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
-              and ``"s"``: return ``True``
+              and ``"s"``: return ``True`` if set, else ``False`` (bool)
             - ``"I"``: 2-element array for the increments (float)
             - ``"R"``: 4-element array for the region (float)
             - ``"V"``: the verbose level (int)
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -535,16 +535,24 @@ class Session:
         Examples
         --------
         >>> with Session() as lib:
-        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf")
+        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf -Ve")
         ...     region = lib.get_common("R")
         ...     projection = lib.get_common("J")
         ...     timestamp = lib.get_common("U")
+        ...     verbose = lib.get_common("V")
         ...     lib.call_module("plot", "-T -Xw+1i -Yh-1i")
         ...     xshift = lib.get_common("X")  # xshift/yshift are in inches
         ...     yshift = lib.get_common("Y")
         ...
-        >>> print(region, projection, timestamp, xshift, yshift)
-        [ 0. 10. 10. 15.] True False 6.0 1.5
+        >>> print(region, projection, timestamp, verbose, xshift, yshift)
+        [ 0. 10. 10. 15.] True False 3 6.0 1.5
+        >>> with Session() as lib:
+        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf")
+        ...     lib.get_common("A")
+        ...
+        Traceback (most recent call last):
+        ...
+        pygmt.exceptions.GMTInvalidInput: Unknown GMT option flag 'A'.
         """
         if option not in "BIJRUVXYabfghinoprst:":
             raise GMTInvalidInput(f"Unknown GMT option flag '{option}'.")
@@ -573,7 +581,7 @@ class Session:
             return value[:status]
         if option in "XY":  # only one valid element in the array
             return value[0]
-        # option is set and the option value (in interger type) is returned via
+        # option is set and the option value (in integer type) is returned via
         # the function return value (i.e., 'status')
         return status
 
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -552,7 +552,7 @@ class Session:
         ...
         Traceback (most recent call last):
         ...
-        pygmt.exceptions.GMTInvalidInput: Unknown GMT option flag 'A'.
+        pygmt.exceptions.GMTInvalidInput: Unknown GMT common option flag 'A'.
         """
         if option not in "BIJRUVXYabfghinoprst:":
             raise GMTInvalidInput(f"Unknown GMT common option flag '{option}'.")
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -494,6 +494,89 @@ class Session:
 
         return value.value.decode()
 
+    def get_common(self, option):
+        """
+        Inquire if a GMT common option has been set and return its current
+        value if possible.
+
+        Parameters
+        ----------
+        option : str
+            The GMT common option to check. Valid options are ``"B"``, ``"I"``,
+            ``"J"``, ``"R"``, ``"U"``, ``"V"``, ``"X"``, ``"Y"``, ``"a"``,
+            ``"b"``, ``"f"``, ``"g"``, ``"h"``, ``"i"``, ``"n"``, ``"o"``,
+            ``"p"``, ``"r"``, ``"s"``, ``"t"``, and ``":"``.
+
+        Returns
+        -------
+        value : bool or int, float or numpy.ndarray
+            Whether the option was set or its value.
+
+            If the option was not set, return ``False``. Otherwise,
+            value depends on the choice of the option.
+
+            - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
+              and ``"s"``: return ``True``
+            - ``"I"``: 2-element array for the increments (float)
+            - ``"R"``: 4-element array for the region (float)
+            - ``"V"``: the verbose level (int)
+            - ``"X"``: the xshift (float)
+            - ``"Y"``: the yshift (float)
+            - ``"a"``: geometry of the dataset (int)
+            - ``"b"``: the option was set for input or output (int)
+            - ``"f"``: the option was set for input or output (int)
+            - ``"h"``: whether to delete existing header records (int)
+            - ``"i"``: number of input columns (int)
+            - ``"o"``: number of output columns (int)
+            - ``"r"``: registration type (int)
+            - ``"t"``: 2-element array for the transparency (float)
+            - ``":"``: the option was set for input or output (int)
+
+        Examples
+        --------
+        >>> with Session() as lib:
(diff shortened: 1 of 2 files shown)


pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+        ----------
+        option : str
+            The GMT common option to check. Valid options are ``"B"``, ``"I"``,
+            ``"J"``, ``"R"``, ``"U"``, ``"V"``, ``"X"``, ``"Y"``, ``"a"``,
+            ``"b"``, ``"f"``, ``"g"``, ``"h"``, ``"i"``, ``"n"``, ``"o"``,
+            ``"p"``, ``"r"``, ``"s"``, ``"t"``, and ``":"``.
+
+        Returns
+        -------
+        value : bool or int, float or numpy.ndarray
```
seisman, line 512:
```suggestion
        value : bool, int, float, or numpy.ndarray
```

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+            Whether the option was set or its value.
+
+            If the option was not set, return ``False``. Otherwise,
+            value depends on the choice of the option.
```
seisman, line 516:
```suggestion
            the return value depends on the choice of the option.
```

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+        >>> with Session() as lib:
+        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf")
+        ...     lib.get_common("A")
+        ...
+        Traceback (most recent call last):
+        ...
+        pygmt.exceptions.GMTInvalidInput: Unknown GMT option flag 'A'.
+        """
+        if option not in "BIJRUVXYabfghinoprst:":
+            raise GMTInvalidInput(f"Unknown GMT option flag '{option}'.")
```
seisman, line 558:
```suggestion
            raise GMTInvalidInput(f"Unknown GMT common option flag '{option}'.")
```

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+        ...
+        >>> print(region, projection, timestamp, verbose, xshift, yshift)
+        [ 0. 10. 10. 15.] True False 3 6.0 1.5
+        >>> with Session() as lib:
+        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf")
+        ...     lib.get_common("A")
+        ...
+        Traceback (most recent call last):
+        ...
+        pygmt.exceptions.GMTInvalidInput: Unknown GMT option flag 'A'.
```
seisman, line 555:
```suggestion
        pygmt.exceptions.GMTInvalidInput: Unknown GMT common option flag 'A'.
```

michaelgrund (approved):

seisman:
Hopefully, this PR can get another review before I merge it @GenericMappingTools/pygmt-maintainers

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+            - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
+              and ``"s"``: return ``True``
+            - ``"I"``: 2-element array for the increments (float)
+            - ``"R"``: 4-element array for the region (float)
+            - ``"V"``: the verbose level (int)
+            - ``"X"``: the xshift (float)
+            - ``"Y"``: the yshift (float)
+            - ``"a"``: geometry of the dataset (int)
+            - ``"b"``: the option was set for input or output (int)
+            - ``"f"``: the option was set for input or output (int)
```
weiji14, line 527: Actually, are b and f the same? binary (b) and coltypes (f)?
```suggestion
            - ``"b"``: the option that was set for input or output (int)
            - ``"f"``: the option that was set for input or output (int)
```

seisman: No, they're not the same.

Initially, I thought the function `GMT_Get_Common` should return the value of the option. For example, after running a command like `gmt plot input.txt -R0/10/0/10 -JX10c/10c -Baf -X2c -Y2c -bi1f -fi0x,1y`, `GMT_Get_Common` should return `X10c/10c` for `J` option and return `i0x,1y` for `f` option, but it doesn't. 

That's because the `GMT_Get_Common` function can only return a status (the function return value) or a double array (function parameter passed by reference). Thus, for options like `-b` and `-f`, the function can only tell if the option was set for input or output (i.e., if `-bi`/`-fi` or `-bo`/`-fo` was specified).

seisman, line 527:
```suggestion
            - ``"b"``: return 0 if `-bi` was set and 1 if `-bo` was set (int)
            - ``"f"``: return 0 if `-fi` was set and 1 if `-fo` was set (int)
```

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+            - ``"h"``: whether to delete existing header records (int)
+            - ``"i"``: number of input columns (int)
+            - ``"o"``: number of output columns (int)
+            - ``"r"``: registration type (int)
+            - ``"t"``: 2-element array for the transparency (float)
+            - ``":"``: the option was set for input or output (int)
```
seisman, line 533:
```suggestion
            - ``":"``: return 0 if `-:i` was set and 1 if `-:o` was set (int)
```

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+        Examples
+        --------
+        >>> with Session() as lib:
+        ...     lib.call_module("basemap", "-R0/10/10/15 -JX5i/2.5i -Baf -Ve")
+        ...     region = lib.get_common("R")
+        ...     projection = lib.get_common("J")
+        ...     timestamp = lib.get_common("U")
+        ...     verbose = lib.get_common("V")
+        ...     lib.call_module("plot", "-T -Xw+1i -Yh-1i")
+        ...     xshift = lib.get_common("X")  # xshift/yshift are in inches
```
weiji14, line 544: Does GMT_Get_Common return numbers in inches only? I.e. even if we do `-JX5c/2.5c`, the output value is still converted from centimeters to inches?

pygmt/clib/session.py
weiji14: Ah ok, thanks for the explanation. I think your changes in https://github.com/GenericMappingTools/pygmt/pull/2500#discussion_r1173233018 look better.

pygmt/clib/session.py
seisman: > Does GMT_Get_Common return numbers in inches only? I.e. even if we do `-Xw+1c`, the output value is still converted from centimeters to inches?

Yes, I believe internally GMT stores all dimensions in inches or points, most likely because PostScript uses points as its default unit (https://en.wikipedia.org/wiki/PostScript#Units_of_length), and 72 points are exactly 1 inch, but 72/2.54 points are 1 cm.

seisman: > > Does GMT_Get_Common return numbers in inches only? I.e. even if we do `-Xw+1c`, the output value is still converted from centimeters to inches?
> 
> Yes, I believe internally GMT stores all dimensions in inches or points, most likely because PostScript uses points as its default unit (https://en.wikipedia.org/wiki/PostScript#Units_of_length), and 72 points are exactly 1 inch, but 72/2.54 points are 1 cm.

That's also why I use `-Xw+1i` in the docstrings, which makes the doctest easier to read.

seisman: https://github.com/GenericMappingTools/gmt/blob/8f11a6f2ebd665f825e50564e8286433be439f9f/src/gmt_init.c#L1168

This line shows that GMT always converts *xshift* to inches.

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+              and ``"s"``: return ``True``
+            - ``"I"``: 2-element array for the increments (float)
+            - ``"R"``: 4-element array for the region (float)
+            - ``"V"``: the verbose level (int)
+            - ``"X"``: the xshift (float)
+            - ``"Y"``: the yshift (float)
+            - ``"a"``: geometry of the dataset (int)
+            - ``"b"``: return 0 if `-bi` was set and 1 if `-bo` was set (int)
+            - ``"f"``: return 0 if `-fi` was set and 1 if `-fo` was set (int)
+            - ``"h"``: whether to delete existing header records (int)
```
weiji14, line 528: Looking at https://docs.generic-mapping-tools.org/6.4/gmt.html#h-full, is this just checking whether `-h+d` was set?

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
 
         return value.value.decode()
 
+    def get_common(self, option):
```
weiji14, line 497: Please add this function to `doc/api/index.rst`.

pygmt/clib/session.py
```diff
@@ -494,6 +494,97 @@ def get_default(self, name):
...
+        Returns
+        -------
+        value : bool, int, float, or numpy.ndarray
+            Whether the option was set or its value.
+
+            If the option was not set, return ``False``. Otherwise,
+            the return value depends on the choice of the option.
+
+            - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
+              and ``"s"``: return ``True``
```
weiji14, line 519: Should we mention `bool` like this?
```suggestion
            - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
              and ``"s"``: return ``True`` if set, else ``False`` (bool)
```

seisman, line 519:
```suggestion
            - options ``"B"``, ``"J"``, ``"U"``, ``"g"``, ``"n"``, ``"p"``,
              and ``"s"``: return ``True`` (bool)
```

pygmt/clib/session.py
seisman: Done in 9fb950db28755243a3785bd67d464f747579bc34.

pygmt/clib/session.py
seisman: Yes.

weiji14 (approved):

PR #2533, opened 2023-05-08
Figure.meca: Refactor tests for plotting a single focal mechanism

seisman:

The `Figure.meca` tests are a little messy and difficult to maintain because each test uses different focal mechanisms and event information. 

I plan to refactor the tests. This PR focuses on testing passing a single focal mechanism to the spec parameter using different data types (dict, file, 1darray and DataFrame). Note that all tests now single a single baseline image test_meca_spec_dict.png.

Changes in this PR:

- Rename test_meca_spec_dictionary to test_meca_spec_single_focalmecha
- Test different input types for a single focal mechanism
- Remove test_meca_spec_file and add test_meca_spec_single_focalmecha_file
- Remove two tests test_meca_gcmt_convention and test_meca_spec_1d_array

This PR is a rework of PR https://github.com/GenericMappingTools/pygmt/pull/2219.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/dvc-diff.yml
+++ b/.github/workflows/dvc-diff.yml
@@ -58,13 +58,6 @@ jobs:
         awk 'NF==5 && NR>=7 && $2=="added" {print $4}' report.md > added_files.txt
         awk 'NF==5 && NR>=7 && $2=="modified" {print $4}' report.md > modified_files.txt
 
-        # Backup the new images in the "baseline-new" directory
-        mkdir pygmt/tests/baseline-new
-        cp pygmt/tests/baseline/*.png pygmt/tests/baseline-new
-        # Pull images in the main branch from cloud storage
-        git checkout main
-        dvc pull --remote upstream
-
         # Append each image to the markdown report
         echo -e "## Image diff(s)\n" >> report.md
         echo -e "<details>\n" >> report.md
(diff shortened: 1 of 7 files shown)


pygmt/tests/test_meca.py
```diff
@@ -8,24 +8,73 @@
...
 
 
-@pytest.mark.mpl_image_compare
-def test_meca_spec_dictionary():
+@pytest.mark.mpl_image_compare(filename="test_meca_spec_single_focalmecha.png")
+@pytest.mark.parametrize("inputtype", ["dict_mecha", "dict_full", "array1d", "pandas"])
+def test_meca_spec_single_focalmecha(inputtype):
+    """
+    Test passing a single focal mechanism to the spec parameter.
+    """
```
seisman, line 16: Different `inputtype` are different ways to passing a single focal mechanism to the `spec` parameter.

.github/workflows/dvc-diff.yml
```diff
@@ -66,35 +73,23 @@ jobs:
         echo -e "### Added images\n" >> report.md
         while IFS= read -r line; do
           echo -e "- $line \n" >> report.md
-          cml-publish --title $line --md "$line" >> report.md < /dev/null
+          echo -e "![](${line/baseline/baseline-new})" >> report.md
```
weiji14, line 76: The generated markdown for this is `` which doesn't show up since it points to a local path. Need to update this somehow?

weiji14 (commented):
The updated unit test `test_meca_spec_single_focalmecha` looks good, but the dvc-diff.yml workflow seems broken. Do you think it would be better to fix the dvc workflow in a separate PR?

seisman:
> but the dvc-diff.yml workflow seems broken. Do you think it would be better to fix the dvc workflow in a separate PR?

The changes to the dvc-diff workflow file was reverted in commit abe334ecff66e82a096484e252477d48fc794869. I'll open a separate PR to fix the dvc-diff workflow.

.github/workflows/dvc-diff.yml
weiji14: There's this deprecation warning on https://github.com/GenericMappingTools/pygmt/actions/runs/5107768299/jobs/9180964772#step:6:55:

```
Warning: "cml publish" is deprecated since "cml comment" now supports "![inline](./asset.png)"
```

Looks like we should be using `cml comment --publish=true` instead now, xref https://cml.dev/doc/ref/comment?

weiji14 (approved):

.github/workflows/dvc-diff.yml
weiji14: Ah wait, just noticed you did try to use `cml comment` in 0c6bc53f1087395ebd11688f0e9bd95d7b7f0e9a. Nevermind then.

Issue #2416, opened 2023-03-13
The command to  extract GMT-equivalent commands needs revision for GMT >= 6.4.0


seisman:
In the [contributing guides](https://www.pygmt.org/v0.8.0/contributing.html#reporting-a-bug), currently we use this command to get the GMT-equivalent commands:
```
python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
```

The command doesn't work as expected after upstream PR https://github.com/GenericMappingTools/gmt/pull/6502 (first appeared in GMT 6.4.0), because new GMT versions always add the "gmt" prefix.

So, the new command for GMT>=6.4.0 should be:
```
python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
```

PR #2561, opened 2023-06-01
Fix the command to extract GMT CLI for GMT>=6.4.0

seisman:

See #2416 for context. In short, for GMT 6.3, we need to use:
```
python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
```
but for GMT>=6.4, we need to use 
```
python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
```

It's possible to have a more complicated command which works for both GMT 6.3 and GMT>=6.4, but I don't think it's necessary. I prefer a simple command, even though it works for GMT>=6.4 only,

Closes https://github.com/GenericMappingTools/pygmt/issues/2416.

### DIFF
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -62,7 +62,7 @@ the problem:
   filter the messages to include only the GMT-equivalent commands using a command
   such as:
 
-      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
+	  python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
 
   where `<test>` is the name of your test script.
 * If the bug is produced when passing an in-memory data object (e.g., a
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -64,7 +64,7 @@ the problem:
 
       python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
 
-  where `<test>` is the name of your test script.
+  where `<test>` is the name of your test script. Note that this script works with GMT>=6.4
 * If the bug is produced when passing an in-memory data object (e.g., a
   pandas.DataFrame or xarray.DataArray) to a PyGMT function, try writing the
   data to a file (e.g., a NetCDF or ASCII txt file) and passing the data file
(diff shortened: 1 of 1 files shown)


doc/contributing.md
```diff
@@ -62,7 +62,7 @@ the problem:
   filter the messages to include only the GMT-equivalent commands using a command
   such as:
 
-      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
+	  python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
```
seisman, line 65:
```suggestion
      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
```

doc/contributing.md
```diff
@@ -62,7 +62,7 @@ the problem:
   filter the messages to include only the GMT-equivalent commands using a command
   such as:
 
-      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
+      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
 
   where `<test>` is the name of your test script.
```
weiji14, line 67: Add a note that this only works for GMT6.4 and above?
```suggestion
  where `<test>` is the name of your test script. Note that this script works with GMT>=6.4
```

weiji14 (approved):

doc/contributing.md
```diff
@@ -62,9 +62,9 @@ the problem:
   filter the messages to include only the GMT-equivalent commands using a command
   such as:
 
-      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print "gmt", $3}'
+      python <test>.py 2>&1 | awk -F': ' '$2=="GMT_Call_Command string" {print $3}'
 
-  where `<test>` is the name of your test script.
+  where `<test>` is the name of your test script. Note that this script works with GMT>=6.4
```
michaelgrund, line 67:
```suggestion
  where `<test>` is the name of your test script. Note that this script works only with GMT>=6.4
```

michaelgrund (approved):

PR #2570, opened 2023-06-11
Set rtol=1e-5 to fix broken tests on macOS due to floating point precision

seisman:

See https://github.com/GenericMappingTools/pygmt/actions/runs/5217144872/jobs/9416664148. Some tests fail on macOS. For example:

```
________________________ test_earth_age_01d_with_region ________________________

    def test_earth_age_01d_with_region():
        """
        Test loading low-resolution earth age with 'region'.
        """
        data = load_earth_age(resolution="01d", region=[-10, 10, -5, 5])
        assert data.shape == (11, 21)
        npt.assert_allclose(data.lat, np.arange(-5, 6, 1))
        npt.assert_allclose(data.lon, np.arange(-10, 11, 1))
>       npt.assert_allclose(data.min(), 11.293945)

../pygmt/tests/test_datasets_earth_age.py:56: 

args = (<function assert_allclose.<locals>.compare at 0x150579bc0>, array(11.293947, dtype=float32), array(11.293945))
kwds = {'equal_nan': True, 'err_msg': '', 'header': 'Not equal to tolerance rtol=1e-07, atol=0', 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 2.21984863e-06
E           Max relative difference: 1.96552102e-07
E            x: array(11.293947, dtype=float32)
E            y: array(11.293945)
```

These tests fail because the default `rtol` value of `npt.assert_allclose` is 1e-7. This PR changes `rtol` to 1e-5.

### DIFF
--- a/pygmt/tests/test_datasets_earth_age.py
+++ b/pygmt/tests/test_datasets_earth_age.py
@@ -53,8 +53,8 @@ def test_earth_age_01d_with_region():
     assert data.gmt.registration == 0
     npt.assert_allclose(data.lat, np.arange(-5, 6, 1))
     npt.assert_allclose(data.lon, np.arange(-10, 11, 1))
-    npt.assert_allclose(data.min(), 11.293945)
-    npt.assert_allclose(data.max(), 125.1189)
+    npt.assert_allclose(data.min(), 11.293945, rtol=1e-5)
+    npt.assert_allclose(data.max(), 125.1189, rtol=1e-5)
 
 
 def test_earth_age_01m_without_region():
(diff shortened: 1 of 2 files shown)


willschlitzer (approved):
"But 'rtol=1e-7' worked on my machine!"

seisman:
> "But 'rtol=1e-7' worked on my machine!"

It works on Linux but fails on macOS with the GMT dev version.

michaelgrund (approved):

Issue #2204, opened 2022-11-23
Support non-ASCII characters in PyGMT arguments and text in Figure.text


seisman:
## Problems

Due to the limitation of the PostScript language, GMT can only work with ASCII characters and a small set of non-ASCII characters. See https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html for the full list of characters that PostScript/GMT/PyGMT can accept.

These non-ASCII characters must be specified using their [octal codes](https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html) or [character escape sequence](https://docs.generic-mapping-tools.org/latest/cookbook/features.html#character-escape-sequences). A few non-ASCII characters (e.g., ü, Î) are allowed and GMT can substitute these non-ASCII characters with the correct PostScript octal codes.

Users who don't know the limitations may pass non-ASCII characters directly in the arguments. For example:
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 5], projection="x1c", frame="WSen+tTime (s) vs Distance (°)")
fig.show()
```
The above script produces this "surprising" figure:

![non-ascii](https://user-images.githubusercontent.com/3974108/203463340-bad4d200-252d-477e-8428-0a451fe0c7b0.png)

So, if users want to add a non-ASCII character to a plot, they must know the limitations and have to go to this page https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html, look for the character in the four tables, and figure out the corresponding octal code (`\260` for the symbol `°`), which is tedious and not easy.

After finding the octal code, users may think changing `°` to `\260` should work: 
```
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 5], projection="x1c", frame="WSen+tTime (s) vs Distance (\260)")
fig.show()
```
but it still produces the same "surprising" figure, because the Python interpreter recognizes `\260` first, and converts it to `°` before passing it to the GMT API. So, users have to use double backslashes or raw strings:
```
frame="WSen+tTime (s) vs Distance (\\260)"
```
or
```
frame=r"WSen+tTime (s) vs Distance (\260)"
```

## Solutions

Since Python works well with non-ASCII characters (acutally it works with any unicode characters), it's possible to pass `°` in Python, and PyGMT should substitute the non-ASCII characters with the corresponding octal codes.

Here are some tests in Python:

```python
# Python support non-ASCII characters
[...]

Issue #2000, opened 2022-07-10
Add a tutorial explaining the usage of octal codes


seisman:
Inspired by the question in https://github.com/GenericMappingTools/pygmt/issues/1994, we need to add a tutorial explaining how to correctly use [octal codes](https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html) in PyGMT.

seisman:
Related to https://github.com/GenericMappingTools/pygmt/issues/2204.

JingHuiTong:
Hi @seisman, 
So far, I want to add the tutorial to examples/tutorials/basics/text.py 
I have some ideas to show 
1) incorrect v.s. correct usage 
2) correct usage with setting PS_CHAR_ENCODING to Standard+ or ISOLatin1+ 
Do you have other idea to show the example ?

seisman:
> * incorrect v.s. correct usage
> * correct usage with setting PS_CHAR_ENCODING to Standard+ or ISOLatin1+
>   Do you have other idea to show the example ?

Sounds good. Also need to teach users how to read the big octal tables.

PR #2584, opened 2023-06-24
Support non-ASCII characters in function arguments

seisman:

Support non-ASCII characters in PyGMT arguments, so that users don't have to use octal codes. 
See https://github.com/GenericMappingTools/pygmt/issues/2204 for context.

**Working example**

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 5], projection="x1c", frame="WSEN+tABC¥±°″≥×≈∇⇔∑DEF")
fig.show()
```
it produces:

map

**Notes to maintainers**
The symbols and ZapfDingbats charset can be obtained from https://unicode.org/Public/MAPPINGS/VENDORS/ADOBE/zdingbat.txt and https://unicode.org/Public/MAPPINGS/VENDORS/ADOBE/symbol.txt. The following script is used to generate the list of characters, but manually editing are necessary.
```
with open("zdingbat.txt", "r") as fin:
    for line in fin:
        if line.startswith("#"):
            continue
        hexval = line.split()[0].strip()
        print(chr(int(hexval, base=16)), end="")
        i += 1
        if i % 16 == 0:
            print()
```

**TODO**

- [x] Support all non-ASCII characters in the ISOLatin1+ character set
- [x] Support non-ASCII characters in the Symbols character set
- [x] Support non-ASCII characters in the ZapfDingbats character set
- [ ] Check `PS_CHAR_ENCODING` so it also works with Standard+ character set?
- [ ] Check arguments that are not processed by `build_arg_string` (e.g., `text` in `Figure.text`)
- [ ] Add a gallery example or a tutorial?
- [x] update any existing examples

Closes https://github.com/GenericMappingTools/pygmt/issues/2204 Closes https://github.com/GenericMappingTools/pygmt/issues/2000

**Reminders**

**Slash Commands**


pygmt/helpers/utils.py
```diff
@@ -91,6 +91,41 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
     return kind
 
 
+def non_ascii_to_octal(argstr):
+    r"""
+    Translate non-ASCII characters to their corresponding octal codes.
+
+    Currently, only ISOLatin1+ character set is supported.
```
yvonnefroehlich, line 98:
```suggestion
    Currently, only the ISOLatin1+ character set is supported.
```

seisman:
This new feature is almost done and is ready for more testings. Please let me know what you think.

weiji14:
Looking good! Could you also change this line in the colorbar gallery example to use degrees perhaps?

https://github.com/GenericMappingTools/pygmt/blob/dde23f11f309ecafa51262149ecb13c083812728/examples/gallery/embellishments/colorbar.py#L43

Also, do we need to test different [PS_CHAR_ENCODING](https://docs.generic-mapping-tools.org/6.4/gmt.conf.html#term-PS_CHAR_ENCODING) settings like Standard, Standard+, ISOLatin1, ISOLatin1+, and ISO-8859-x?

seisman:
> Looking good! Could you also change this line in the colorbar gallery example to use degrees perhaps?
> 
> https://github.com/GenericMappingTools/pygmt/blob/dde23f11f309ecafa51262149ecb13c083812728/examples/gallery/embellishments/colorbar.py#L43
> 

Done in e1b43b2eadd6923064d8faf5b1b4ad7b58929c81.

> Also, do we need to test different [PS_CHAR_ENCODING](https://docs.generic-mapping-tools.org/6.4/gmt.conf.html#term-PS_CHAR_ENCODING) settings like Standard, Standard+, ISOLatin1, ISOLatin1+, and ISO-8859-x? Currently the mapping seems to assume ISOLatin1, but if a user sets PS_CHAR_ENCODING to another encoding, the output might be different.

ISOLatin1+ is an extension of ISOLatin1, which means ISOLatin1+ supports more characters than ISOLatin (https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html). Thus, it does no harm to support ISOLatin1+ only characters even if users set ISOLatin1 encoding.

For Standard+ and other ISO-8859-x encodings, yes, we can definitely support them. But it also means we have to carefully check their differences and maintain the corresponding mapping dictionary from character to octal codes.

pygmt/helpers/utils.py
```diff
@@ -91,6 +92,84 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
...
+        {
+            c: "\\" + format(i, "o")
+            for c, i in zip(
+                "•…™—–ﬁžšŒ†‡Ł⁄‹Š›œŸŽł‰„“”ı`´ˆ˜¯˘˙¨‚˚¸'˝˛ˇ",
+                [*range(25, 32), *range(127, 160)],
+            )
+        }
+    )
+    # ISOLatin1+ charset: \240-\377
+    mapping.update({chr(i): "\\" + format(i, "o") for i in range(160, 256)})
```
seisman, line 166: Actually, Python supports many different encodings, thus it's possible to implement this feature without manually maintaining the big dictionary, just like what I already do at line 166 for ISOLatin1+ characters \240 to \377.

pygmt/helpers/utils.py
```diff
@@ -91,6 +92,84 @@ def data_kind(data, x=None, y=None, z=None, required_z=False):
...
+    -------
+    str
+        The translated string.
+
+    Examples
+    --------
+    >>> non_ascii_to_octal("•‰“”±°ÿ")
+    '\\31\\214\\216\\217\\261\\260\\377'
+    >>> non_ascii_to_octal("αζ∆Ω∑∏∇")
+    '@~\\141@~@~\\172@~@~\\104@~Ω@~\\345@~@~\\325@~@~\\321@~
```
seisman, line 122:
```suggestion
    '@~\\141@~@~\\172@~@~\\104@~Ω@~\\345@~@~\\325@~@~\\321@~'
```

pygmt/helpers/utils.py
seisman: **Update:**

Can be done using:
```python
for i in range(160, 256):
    print(i, chr(i), chr(i).encode("iso-8859-1").decode("iso-8859-5"))
```

seisman:
> For Standard+ and other ISO-8859-x encodings, yes, we can definitely support them. But it also means we have to carefully check their differences and maintain the corresponding mapping dictionary from character to octal codes.

For other ISO-8859-x encodings, it's possible to generate the big mapping dictionary using the following codes:

```
for i in range(160, 256):
    print(i, chr(i), chr(i).encode("iso-8859-1").decode("iso-8859-5"))
```

examples/gallery/embellishments/colorbar.py
```diff
@@ -40,7 +40,7 @@
     # with a length/width (+w) of 4 cm by 0.5 cm, and plotted horizontally (+h)
     position="g0.3/8.7+w4c/0.5c+h",
     box=True,
-    frame=["x+lTemperature", r"y+l\260C"],
+    frame=["x+lTemperature", r"y+l°C"],
```
yvonnefroehlich, line 43: I feel the leading `r` is not anymore needed 🙂, or?
```suggestion
    frame=["x+lTemperature", "y+l°C"],
```

pygmt/helpers/utils.py
```diff
@@ -145,6 +146,119 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False):
...
+                + "≅ΑΒΧΔΕΦΓΗΙϑΚΛΜΝΟ"  # \10x-11x
+                + "ΠΘΡΣΤΥςΩΞΨΖ[∴]⊥_"  # \12x-13x
+                + "αβχδεφγηιϕκλμνο"  # \14x-15x
+                + "πθρστυϖωξψζ{|}∼"  # \16x-17x. \177 is undefined
+                + "€ϒ′≤⁄∞ƒ♣♦♥♠↔←↑→↓"  # \24x-\25x
+                + "°±″≥×∝∂•÷≠≡≈…↵"  # \26x-27x
+                + "ℵℑℜ℘⊗⊕∅∩∪⊃⊇⊄⊂⊆∈∉"  # \30x-31x
+                + "∠∇®©™∏√⋅¬∧∨⇔⇐⇑⇒⇓"  # \32x-33x
+                + "◊〈®©™∑"  # \34x-35x
+                + "〉∫⌠⌡",  # \36x-37x. \360 and \377 are undefined
```
seisman, line 206: Symbols \140, \275 and \276 don't exist in the unicode table, so they're incorrectly shown.

Some symbols of \36x and \37x are shown as boxes in the GitHub web GUI, but they're shown correctly in Vim, so these symbols should be good.

Screenshot from 2023-08-02 15-08-53

seisman:
> Check PS_CHAR_ENCODING so it also works with Standard+ character set?

Already answered in https://github.com/GenericMappingTools/pygmt/pull/2584#issuecomment-1616411916.

> For other ISO-8859-x encodings, it's possible to generate the big mapping dictionary automatically using the following code:
> ```
> {chr(i).encode("iso-8859-1").decode("iso-8859-5"): "\\" + format(i, "o") for i in range(160, 256)}
> ```
> but we have to check the PS_CHAR_ENCODING parameter to decide which dictionary to use. However, adding `from pygmt.clib.session import Session` to `pygmt.helpers.utils` results in circular imports.
> 

> Check arguments that are not processed by build_arg_string (e.g., `text` in Figure.text)

Yes, we can also translate the strings passed by the `text` parameter, but we can do very little to plaintext files (we can, but it means we have to open and read the whole plaintext file, which is not efficient). So not sure what we should do here.

> Add a gallery example or a tutorial?

Better to do it in a separate PR.

seisman:
@GenericMappingTools/pygmt-maintainers Any comments and suggestions?

michaelgrund (approved):
Looks good to me!

pygmt/helpers/utils.py
```diff
@@ -196,6 +197,119 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
...
+
+    Parameters
+    ----------
+    argstr : str
+        The string to be translated.
+
+    Returns
+    -------
+    str
+        The translated string.
```
weiji14, line 215:
```suggestion
    translated_argstr : str
        The translated string.
```

weiji14 (approved):
Haven't checked every individual symbol, but I trust that they are correct :slightly_smiling_face:

> > Check arguments that are not processed by build_arg_string (e.g., `text` in Figure.text)
> 
> Yes, we can also translate the strings passed by the `text` parameter, but we can do very little to plaintext files (we can, but it means we have to open and read the whole plaintext file, which is not efficient). So not sure what we should do here.

Ok, to handle ASCII characters in `fig.text` in a separate PR.

pygmt/helpers/utils.py
seisman: Here is the improved version:
```
for code in [*range(0o040, 0o200), *range(0o240, 0o400)]:
    char = codecs.decode(bytes([code]), f"iso8859-{i}", errors="replace")
```

PR #2585, opened 2023-06-26
Create gallery example to show usage of tile maps

michaelgrund:

So far we don't promote the usage of tilemaps. Here's a gallery example to show its application using two different sources.

**Preview:** https://pygmt-dev--2585.org.readthedocs.build/en/2585/gallery/maps/tilemaps.html

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -27,7 +27,8 @@ fig.show()
 ###############################################################################
 # It's also possible to use tiles provided via the
 # `contextily <https://github.com/geopandas/contextily>`__
-# library.
+# library. See :doc:`Contextily providers <contextily:providers_deepdive>`
+# for a list of possible tilemap options.
 
 fig = pygmt.Figure()
 fig.tilemap(
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -21,7 +21,8 @@ fig.show()
 
 ###############################################################################
 # It's also possible to use tiles provided via the
-# [contextily](https://github.com/geopandas/contextily) library.
+# `contextily <https://github.com/geopandas/contextily>`__ 
+# library.
 
 fig = pygmt.Figure()
 fig.tilemap(
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -1,0 +1,32 @@
+"""
+Tilemaps
+--------------------
+The :meth:`pygmt.Figure.tilemap` method allows to plot
+tiles from a tile server or local file as a basemap or overlay. 
+"""
+import pygmt 
+
+fig = pygmt.Figure()
+fig.tilemap(region=[-157.84, -157.8, 21.255, 21.285],
+            projection="M12c",
+            zoom=14,
+            # Use tiles from OpenStreetMap tile server
+            source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
+            frame="afg")
+
+fig.show()
+
+###############################################################################
+# It's also possible to use tiles provided via the 
+# [contextily](https://github.com/geopandas/contextily) library.
+
+import contextily
+
+fig = pygmt.Figure()
+fig.tilemap(region=[-157.84, -157.8, 21.255, 21.285],
+            projection="M12c",
+            # Use the Stamen.Watercolor option from contextily
+            source=contextily.providers.Stamen.Watercolor,
+            frame="afg")
+
+fig.show()
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -1,6 +1,6 @@
 """
 Tilemaps
---------------------
+--------
 The :meth:`pygmt.Figure.tilemap` method allows to plot
 tiles from a tile server or local file as a basemap or overlay. 
 """
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -4,29 +4,33 @@ Tilemaps
 The :meth:`pygmt.Figure.tilemap` method allows to plot
 tiles from a tile server or local file as a basemap or overlay. 
 """
-import pygmt 
+import pygmt
 
 fig = pygmt.Figure()
-fig.tilemap(region=[-157.84, -157.8, 21.255, 21.285],
-            projection="M12c",
-            zoom=14,
-            # Use tiles from OpenStreetMap tile server
-            source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
-            frame="afg")
+fig.tilemap(
+    region=[-157.84, -157.8, 21.255, 21.285],
+    projection="M12c",
+    zoom=14,
+    # Use tiles from OpenStreetMap tile server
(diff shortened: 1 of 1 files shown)


examples/gallery/maps/tilemaps.py
```diff
@@ -0,0 +1,36 @@
...
+tiles from a tile server or local file as a basemap or overlay.
+"""
+import contextily
+import pygmt
+
+fig = pygmt.Figure()
+fig.tilemap(
+    region=[-157.84, -157.8, 21.255, 21.285],
+    projection="M12c",
+    zoom=14,
```
yvonnefroehlich, line 14: I feel it would be helpful to add some explanation of the `zoom` parameter either within the code or above. Feel free to adjust or shorten the formulation :slightly_smiling_face:.
```suggestion
    # Set level of details (0-22)
    # Higher levels mean a zoom level closer to the Earth's
    # surface with more tiles covering a smaller
    # geographic area and thus more details and vice versa
    # Please note, not all zoom levels are always available
    zoom=14,
```

examples/gallery/maps/tilemaps.py
```diff
@@ -0,0 +1,41 @@
...
+    source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
+    frame="afg",
+)
+
+fig.show()
+
+###############################################################################
+# It's also possible to use tiles provided via the
+# `contextily <https://github.com/geopandas/contextily>`__
+# library.
```
weiji14, line 30: Add link to list of contextily providers at https://contextily.readthedocs.io/en/stable/providers_deepdive.html#What-is-this-%E2%80%9Cprovider%E2%80%9D-object-?

```suggestion
###############################################################################
# It's also possible to use tiles provided via the
# `contextily <https://github.com/geopandas/contextily>`__
# library. See :doc:`Contextily providers <contextily:providers_deepdive>`
# for a list of possible tilemap options.
```

weiji14 (commented):
Nice one, thanks @michaelgrund for adding this example! Just one comment for now.

seisman:
Nice example! Do you think switching the order of the two examples makes more sense? I.e., the first example uses the tile providers available from contextily and the second is a more advanced example for people who wants to use custom tile servers that contextily doesn't provide (`source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"` as an example although contextily supports OpenStreetMap).

michaelgrund:
> Nice example! Do you think switching the order of the two examples makes more sense? I.e., the first example uses the tile providers available from contextily and the second is a more advanced example for people who wants to use custom tile servers that contextily doesn't provide (`source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"` as an example although contextily supports OpenStreetMap).

Not sure, since contextily needs to be loaded additionally as a separate library, compared to the "direct" link to the custom tile servers. What do others think?

yvonnefroehlich:
> > Nice example! Do you think switching the order of the two examples makes more sense? I.e., the first example uses the tile providers available from contextily and the second is a more advanced example for people who wants to use custom tile servers that contextily doesn't provide (`source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"` as an example although contextily supports OpenStreetMap).
> 
> Not sure, since contextily needs to be loaded additionally as a separate library, compared to the "direct" link to the custom tile servers. What do others think?

I have no clear answer here, and I feel I am fine with both orders. Here are some thoughts or ideas, which are maybe helpful regarding a decision:
Considering that `contextily` needs to be loaded additionally and that the `Stamen.Watercolor` option is a  more special visualization than `OpenStreetMap`, the current Figure order makes sense to me. However, by default (i.e., the `source` parameter is not specified), the `xyzservices.TileProvider` object `xyzservices.providers.Stamen.Terrain` is used. So, maybe expand the example and start without specifying `source`. Then introduce `source`  based on using not the default `Stamen.Terrain` web tiles, but the `Stamen.Watercolor` option from `contextily`. Create a second figure to show that also a web tile provider in the form of a URL can be used and introduce the `zoom` parameter.

_Code example_: (comments need definitly some improvement)
```python
import contextily
import pygmt

fig = pygmt.Figure()

# Default xyzservices.providers.Stamen.Terrain
fig.tilemap(
    region=[-157.84, -157.8, 21.255, 21.285],
    projection="M12c",
    frame="af+e",
)

fig.shift_origin(xshift="13c")

fig.tilemap(
    region=[-157.84, -157.8, 21.255, 21.285],
    projection="M12c",
    # Instead of Stamen.Terrain web tiles use the Stamen.Watercolor option
    # from contextily
    source=contextily.providers.Stamen.Watercolor,
    frame=["wSne", "af+e"],
)

fig.show()

###############################################################################
# Instead of a web tile provider in the form of a xyzservices.TileProvider
# object use one in the form of a URL

fig = pygmt.Figure()

fig.tilemap(
    region=[-157.84, -157.8, 21.255, 21.285],
    projection="M12c",
    # Set level of details (0-22)
    # Higher levels mean a zoom level closer to the Earth's
    # surface with more tiles covering a smaller
    # geographic area and thus more details and vice versa
    # Please note, not all zoom levels are always available
    zoom=14,
    # Use tiles from OpenStreetMap tile server
    source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
    frame="afg+e",
)

fig.show()
```
_Output figures_:
tilemaps_xyzservices

tilemaps_url

seisman:
> maybe expand the example and start without specifying `source`. Then introduce `source` based on using not the default `Stamen.Terrain` web tiles, but the `Stamen.Watercolor` option from `contextily`. Create a second figure to show that also a web tile provider in the form of a URL can be used and introduce the `zoom` parameter.

Good idea!

weiji14:
> > maybe expand the example and start without specifying `source`. Then introduce `source` based on using not the default `Stamen.Terrain` web tiles, but the `Stamen.Watercolor` option from `contextily`. Create a second figure to show that also a web tile provider in the form of a URL can be used and introduce the `zoom` parameter.
> 
> Good idea!

This is starting to sound like a tutorial instead of a gallery example :sweat_smile: Let's maybe keep to just 2 maps at a maximum instead of 3 for this gallery example. We can still show one using `Stamen`, and another figure using OpenStreetmap's tile server.

Note that the latter OpenStreetMap example is actually the same as using `contextily.OpenStreetMap.Mapnik`, but it's still good to show how people can use a tile server url like `https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png`. Also, either way, `contextily` will have to be used with or without `import contextily`, so the order doesn't matter too much.

seisman:
> This is starting to sound like a tutorial instead of a gallery example

Why not extend it as a tutorial?

seisman (approved):
OK. Approving this PR.

yvonnefroehlich (approved):
I am fine with having both gallery example and tutorial, as suggested by @michaelgrund. In this case, I feel @weiji14 is right, and the gallery example should be short. For the tutorial, I suggest considering also the `lonlat` parameter.

weiji14 (approved):
Yep, let's open up a separate PR for a longer tutorial.

yvonnefroehlich:
> Yep, let's open up a separate PR for a longer tutorial.

@michaelgrund do you also plan to add this tutorial, or @weiji14 do you want to work on this?

PR #370, opened 2019-11-06
Allow passing in Red, Green, Blue grids into grdimage

weiji14:
**Description of proposed changes**

Allow for multiple grid inputs into `grdimage` via a Python list. ~~Using the sequence_space converter (created in #325) to do that.~~ Edit: Using if-then parsing instead to handle a list of xarray.DataArray grids better. Useful when users want to make an RGB image plot e.g. from Landsat images (Bands 4,3,2).

Below is a bit of a convoluted example based on this [tutorial](https://gmt-tutorials.org/en/map_elements.html) (in Chinese) which has preprocessed grd files scaled to 0-255 range, all courtesy of @whyjz (Hi :wave:!).

```python
import pygmt

for band in [4, 3, 2]:
    pygmt.which(
        f"https://gmt-tutorials.org/en/_downloads/LC08_L1TP_160045_20170408_20170414_01_T1_B{band}_s.grd",
        download="l",
    )

fig = pygmt.Figure()
fig.grdimage(
    grid=[
        "LC08_L1TP_160045_20170408_20170414_01_T1_B4_s.grd",
        "LC08_L1TP_160045_20170408_20170414_01_T1_B3_s.grd",
        "LC08_L1TP_160045_20170408_20170414_01_T1_B2_s.grd",
    ],
    frame=["af", "wSnE"],
)
fig.show()
```
Rub’ al Khali desert image

Fixes #

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

Issue #1555, opened 2021-10-01
Integrate with rioxarray and implement GMT_IMAGE to read RGB GeoTIFFs


weiji14:
**Description of the desired feature**

This is a parallel issue to #578 and #1442. While researching how to read GeoTIFFs into an `xarray.DataArray`, specifically GeoTIFFs with RGB bands stored in a single band but with a color lookup table, I discover that there wasn't really a straightforward solution :sweat_smile: So this issue is mainly to jot down some notes.

### Method A: [`rioxarray.open_rasterio`](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray-open-rasterio)

Pros: Reads in data with proper dtype (e.g. uint8) and proper spatial_ref
Cons: [`rioxarray`](https://github.com/corteva/rioxarray) is a bit of a heavy dependency as it requires `scipy`

```python
import pygmt
import rioxarray

filename_or_obj = pygmt.which(fname="@earth_day_01d", download="a")

with rioxarray.open_rasterio(filename_or_obj) as dataarray:
    print(dataarray)
```
gives
```python
<xarray.DataArray (band: 1, y: 180, x: 360)>
array([[[251, 251, ..., 251, 251],
        [251, 252, ..., 251, 252],
        ...,
        [ 68,  16, ...,  68, 217],
        [149, 149, ..., 149,  16]]], dtype=uint8)
Coordinates:
  * band         (band) int64 1
  * x            (x) float64 -179.5 -178.5 -177.5 -176.5 ... 177.5 178.5 179.5
  * y            (y) float64 89.5 88.5 87.5 86.5 ... -86.5 -87.5 -88.5 -89.5
    spatial_ref  int64 0
Attributes:
    scale_factor:  1.0
    add_offset:    0.0
```

### Method B: [`xr.open_dataarray(..., engine="rasterio")`](https://xarray.pydata.org/en/stable/generated/xarray.open_rasterio.html)

Pros: Requires only [`rasterio`](https://github.com/mapbox/rasterio) (a Pythonic GDAL wrapper) to be installed, which keeps things lightweight
Cons: Seems to read in data as float32 by default

Note  [`xr.open_rasterio`](https://xarray.pydata.org/en/stable/generated/xarray.open_rasterio.html) is pending deprecation, see https://github.com/pydata/xarray/issues/4697.

```python
with xr.open_dataarray(filename_or_obj) as dataarray:
    print(dataarray)
```
produces
```python
[...]

PR #2590, opened 2023-07-03
Allow passing RGB xarray.DataArray images into grdimage

weiji14:

Enables `grdimage` to plot 3-band RGB images stored in an `xarray.DataArray` with a shape like (3, Y, X).

Example usage based on https://corteva.github.io/rioxarray/stable/examples/COG.html:

```python
import pygmt
import rioxarray

image = rioxarray.open_rasterio(
    "https://oin-hotosm.s3.amazonaws.com/5d7dad0becaf880008a9bc88/0/5d7dad0becaf880008a9bc89.tif",
    masked=True,
    overview_level=4,
)
assert image.sizes == {"band": 3, "y": 312, "x": 688}

fig = pygmt.Figure()
fig.grdimage(grid=image, frame=True)
fig.show()
```
produces

rgb_image

Implementation currently saves the 3-band `xarray.DataArray` image to a temporary GeoTIFF before passing it to `grdimage`, similar to what `tilemap` did in #2394 https://github.com/GenericMappingTools/pygmt/blob/e98efe16c22fb7eb3f20e2004248bc71c6f68097/pygmt/src/tilemap.py#L151-L156

Notes:
- Conversion of `xarray.DataArray` to temporary GeoTIFF file requires `rioxarray`.
- Only certain dtypes like uint8 and float32 are supported for plotting, so need to either warn users if other dtypes are passed in. Best scenario is if histogram equalization via `grdhisteq` #1433 or normalization to 0-255 via `grdmix` (#1437) is applied
- This implementation will need to be refactored when `GMT_IMAGE` is wrapped in PyGMT so that temporary files can be avoided, see #1555. The unit tests in this PR should still be valid though.

TODO:
- [x] Initial implementation using tempfile
- [x] Raise error on missing `rioxarray` dependency
- [x] Raise warning when unsupported dtypes are passed in
- [ ] Maybe implement same logic for `grdview` too?

Closes #370, partially addresses #1555.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -25,7 +25,12 @@ from pygmt.exceptions import (
     GMTInvalidInput,
     GMTVersionError,
 )
-from pygmt.helpers import data_kind, fmt_docstring, tempfile_from_geojson
+from pygmt.helpers import (
+    data_kind,
+    fmt_docstring,
+    tempfile_from_geojson,
+    tempfile_from_image,
+)
 
 FAMILIES = [
     "GMT_IS_DATASET",  # Entity is a data table
@@ -1530,7 +1535,7 @@ class Session:
         """
         kind = data_kind(data, x, y, z, required_z=required_z)
 
-        if check_kind == "raster" and kind not in ("file", "grid"):
+        if check_kind == "raster" and kind not in ("file", "grid", "image"):
             raise GMTInvalidInput(f"Unrecognized data type for grid: {type(data)}")
         if check_kind == "vector" and kind not in (
             "file",
@@ -1545,6 +1550,7 @@ class Session:
             "file": nullcontext,
             "geojson": tempfile_from_geojson,
             "grid": self.virtualfile_from_grid,
+            "image": tempfile_from_image,
             # Note: virtualfile_from_matrix is not used because a matrix can be
             # converted to vectors instead, and using vectors allows for better
             # handling of string type inputs (e.g. for datetime data types)
@@ -1553,7 +1559,7 @@ class Session:
         }[kind]
 
         # Ensure the data is an iterable (Python list or tuple)
-        if kind in ("geojson", "grid"):
+        if kind in ("geojson", "grid", "image"):
             _data = (data,)
         elif kind == "file":
             # Useful to handle `pathlib.Path` and string file path alike
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -6,6 +6,7 @@ Uses ctypes to wrap most of the core functions from the C API.
 """
 import ctypes as ctp
 import sys
+import warnings
 from contextlib import contextmanager, nullcontext
 
 import numpy as np
@@ -1560,6 +1561,15 @@ class Session:
 
         # Ensure the data is an iterable (Python list or tuple)
         if kind in ("geojson", "grid", "image"):
+            if kind == "image" and data.dtype != "uint8":
+                msg = (
+                    f"Input image has dtype: {data.dtype} which is unsupported, "
+                    "and may result in an incorrect output. Please recast image "
+                    "to a uint8 dtype and/or scale to 0-255 range, e.g. "
+                    "using a histogram equalization function like "
+                    "skimage.exposure.equalize_hist."
+                )
+                warnings.warn(message=msg, category=RuntimeWarning, stacklevel=2)
             _data = (data,)
         elif kind == "file":
             # Useful to handle `pathlib.Path` and string file path alike
--- a/pygmt/helpers/tempfile.py
+++ b/pygmt/helpers/tempfile.py
@@ -147,3 +147,34 @@ def tempfile_from_geojson(geojson):
                     geoseries.to_file(**ogrgmt_kwargs)
 
         yield tmpfile.name
+
+
+@contextmanager
+def tempfile_from_image(image):
+    """
+    Saves a 3-band `xarray.DataArray` to a temporary GeoTIFF file via
+    rioxarray.
+
+    Parameters
+    ----------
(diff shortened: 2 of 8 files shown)


pygmt/tests/test_grdimage_image.py
```diff
@@ -0,0 +1,63 @@
...
+            rda.data = red
+            da_red = rda.astype(dtype=np.uint8).copy()
+            rda.data = green
+            da_green = rda.astype(dtype=np.uint8).copy()
+            rda.data = blue
+            da_blue = rda.astype(dtype=np.uint8).copy()
+
+            xr_image = xr.concat(objs=[da_red, da_green, da_blue], dim="band")
+        assert xr_image.sizes == {"band": 3, "y": 180, "x": 360}
+        return xr_image
```
weiji14, line 43: Should replace this with the `load_blue_marble` function from https://github.com/GenericMappingTools/pygmt/pull/2235 once done. Ideally, the `rioxarray.open_rasterio` function would be able to load this 3-band GeoTIFF from colorinterp directly without the hacky logic here.

pygmt/src/grdimage.py
```diff
@@ -179,17 +180,25 @@ def grdimage(self, grid, **kwargs):
...
-            fname = stack.enter_context(file_context)
-            lib.call_module(
-                module="grdimage", args=build_arg_string(kwargs, infile=fname)
-            )
+    with GMTTempFile(suffix=".tif") as tmpfile:
+        if hasattr(grid, "dims") and len(grid.dims) == 3:
+            grid.rio.to_raster(raster_path=tmpfile.name)
+            _grid = tmpfile.name
+        else:
+            _grid = grid
```
weiji14, line 189: Should maybe put this logic in `lib.virtualfile_from_data`, since we'll need to reuse it for `grdview`, `grdcut`, and possibly other GMT modules that work with GMT_IMAGE.

weiji14: Ok, created a new `tempfile_from_image` function at 89da91603705b817d70bf60f08464ba817f40314 (similar to `tempfile_from_geojson`). This can be replaced with `lib.virtualfile_from_image` once that is implemented.

pygmt/clib/session.py
```diff
@@ -1553,7 +1560,16 @@ def virtualfile_from_data(
...
+        if kind in ("geojson", "grid", "image"):
+            if kind == "image" and data.dtype != "uint8":
+                msg = (
+                    f"Input image has dtype: {data.dtype} which is unsupported, "
+                    "and may result in an incorrect output. Please recast image "
+                    "to a uint8 dtype and/or scale to 0-255 range, e.g. "
+                    "using a histogram equalization function like "
+                    "skimage.exposure.equalize_hist."
+                )
+                warnings.warn(message=msg, category=RuntimeWarning, stacklevel=2)
```
weiji14, line 1572: Took a look at [`grdhisteq`](https://docs.generic-mapping-tools.org/6.4/grdhisteq.html), but it doesn't appear to support 3-band image inputs, only 1-band grids, so suggesting to use [`skimage.exposure.equalize_hist`](https://scikit-image.org/docs/0.21.x/api/skimage.exposure.html#skimage.exposure.equalize_hist) instead. Could probably raise a feature request to upstream GMT to let grdhisteq support this too?

pygmt/helpers/utils.py
```diff
@@ -112,7 +112,7 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False):
     Returns
     -------
     kind : str
-        One of: ``'file'``, ``'grid'``, ``'matrix'``, ``'vectors'``.
+        One of: ``'file'``, ``'grid'``, ``image``, ``'matrix'``, ``'vectors'``.
```
weiji14, line 115: TODO handle merge conflicts after #2493 is merged.

pygmt/helpers/utils.py
```diff
@@ -130,11 +130,16 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False):
...
+    'image'
     """
     if isinstance(data, (str, pathlib.PurePath)):
         kind = "file"
     elif isinstance(data, xr.DataArray):
-        kind = "grid"
+        if len(data.dims) == 3:
+            kind = "image"
+        else:
+            kind = "grid"
```
seisman, line 142: Much shorter version:
```suggestion
        kind = "image" if len(data.dims) == 3 else "grid"
```

pygmt/helpers/tempfile.py
```diff
@@ -147,3 +147,34 @@ def tempfile_from_geojson(geojson):
                     geoseries.to_file(**ogrgmt_kwargs)
 
         yield tmpfile.name
+
+
+@contextmanager
+def tempfile_from_image(image):
+    """
+    Saves a 3-band `xarray.DataArray` to a temporary GeoTIFF file via
```
seisman, line 155:
```suggestion
    Saves a 3-band :class:`xarray.DataArray` to a temporary GeoTIFF file via
```

weiji14:
> This PR looks good to me. Are there any more changes you want to make into this PR?

Was planning to wait for https://github.com/GenericMappingTools/pygmt/pull/2493#discussion_r1254073865 to be finalized so I can merge those changes on top of here, but I can merge this PR first too. Wanted to add some extra tests for `grdview`, which I'll try to do this weekend.

weiji14:
On second thought, let's do the unit tests for `grdview` in a separate PR, because this one is already quite big and there are I'd like to sync the `grdview` docstring with upstream GMT too. I've solved the merge conflicts with #2493 now @seisman, so maybe take another look before I merge this in.

seisman (approved):

PR #2595, opened 2023-07-05
Add private function _validate_data_input to validate input data

seisman:

Includes validation logic from #1523, #1478 and #93.

Address comment https://github.com/GenericMappingTools/pygmt/pull/2493#discussion_r1251647656.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -62,11 +62,11 @@ def _validate_data_input(
     if data is None:  # data is None
         if x is None and y is None:  # both x and y are None
             raise GMTInvalidInput("No input data provided.")
-        elif x is None or y is None:  # either x or y is None
+        if x is None or y is None:  # either x or y is None
             raise GMTInvalidInput("Must provide both x and y.")
-        else:  # both x and y are not None
-            if required_z and z is None:
-                raise GMTInvalidInput("Must provide x, y, and z.")
+        # both x and y are not None, now checks z
+        if required_z and z is None:
+            raise GMTInvalidInput("Must provide x, y, and z.")
     else:  # data is not None
         if x is not None or y is not None or z is not None:
             raise GMTInvalidInput("Too much data. Use either data or x/y/z.")
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -64,21 +64,20 @@ def _validate_data_input(
             raise GMTInvalidInput("No input data provided.")
         if x is None or y is None:  # either x or y is None
             raise GMTInvalidInput("Must provide both x and y.")
-        # both x and y are not None, now checks z
+        # both x and y are not None, now check z
         if required_z and z is None:
             raise GMTInvalidInput("Must provide x, y, and z.")
     else:  # data is not None
         if x is not None or y is not None or z is not None:
             raise GMTInvalidInput("Too much data. Use either data or x/y/z.")
+        # For 'matrix' kind, check if data has the required z column
         if kind == "matrix" and required_z:
-            # np.ndarray or pd.DataFrame
-            if hasattr(data, "shape"):
+            if hasattr(data, "shape"):  # np.ndarray or pd.DataFrame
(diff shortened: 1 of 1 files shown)


seisman:
It makes more sense to also move the following lines to the `_validate_data_input` function, but it's not easy.
https://github.com/GenericMappingTools/pygmt/blob/cca17c74d2b36e6cee0098e81b0c75d3d2a8f201/pygmt/helpers/utils.py#L83-L87

pygmt/helpers/utils.py
```diff
@@ -14,7 +14,63 @@
...
+        if x is None and y is None:  # both x and y are None
+            raise GMTInvalidInput("No input data provided.")
+        elif x is None or y is None:  # either x or y is None
+            raise GMTInvalidInput("Must provide both x and y.")
+        else:  # both x and y are not None
+            if required_z and z is None:
+                raise GMTInvalidInput("Must provide x, y, and z.")
+    else:  # data is not None
+        if x is not None or y is not None or z is not None:
+            raise GMTInvalidInput("Too much data. Use either data or x/y/z.")
```
weiji14, line 70: > It makes more sense to also move the following lines to the `_validate_data_input` function, but it's not easy.
> 
> https://github.com/GenericMappingTools/pygmt/blob/cca17c74d2b36e6cee0098e81b0c75d3d2a8f201/pygmt/helpers/utils.py#L83-L87

Hmm, can't we just add it here?

```suggestion
    else:  # data is not None
        if x is not None or y is not None or z is not None:
            raise GMTInvalidInput("Too much data. Use either data or x/y/z.")
        if required_z and (
            getattr(data, "shape", (3, 3))[1] < 3  # np.array, pd.DataFrame
            or len(getattr(data, "data_vars", (0, 1, 2))) < 3  # xr.Dataset
        ):
            raise GMTInvalidInput("data must provide x, y, and z columns.")
```

seisman: > But maybe we can do `hasattr(data, "shape") or hasattr(data, "data_vars")` first?

Sounds promising.

weiji14: Haven't tested this, but maybe something like this:

```suggestion
    else:  # data is not None
        if x is not None or y is not None or z is not None:
            raise GMTInvalidInput("Too much data. Use either data or x/y/z.")
        if (
            hasattr(data, "shape")  # np.array, pd.DataFrame
            and getattr(data, "shape", (3, 3))[1] < 3
        ) or (
            hasattr(data, "data_vars")  # xr.Dataset
            and len(getattr(data, "data_vars", (0, 1, 2))) < 3
        ):
            raise GMTInvalidInput("data must provide x, y, and z columns.")
```

Not sure why I added a default value with `getattr` at https://github.com/GenericMappingTools/pygmt/pull/1523, might be able to remove it if not needed, but test first.

seisman: > Can't remember why I added a default value with `getattr` at #1523,

`getattr(data, "shape", (3, 3))` returns `(3,3)` if `data` doesn't has the `shape` attribute, so it won't raise the exception.

With the above changes, `test_geopandas_plot3d_default_cube` will fail, because in this test: 
```
>>> multipoint = shapely.geometry.MultiPoint([(0.5, 0.5, 0.5), (1.5, 1.5, 1.5)])
>>> gdf = gpd.GeoDataFrame(geometry=[multipoint])
>>> gdf.shape
(1, 1)
```
`gdf` has a shape `(1,1)`.

weiji14: Add a `not hasattr(data, "__geo_interface__")` check perhaps?

seisman: In ca2d49f2dbfd1daafb492f7032d45a04c791bba4, I changed the order of some codes to determine the data kind first and then check if the data input is valid. It works well.

weiji14: Nice, that works too. Can you fix the pylint error? Should be good after that.

weiji14 (approved):

PR #2615, opened 2023-07-25
Figure.image: Add alias "bitcolor" for "G"

yvonnefroehlich:

This PR aims to add the alias `bitcolor` for **G** of the method `pygmt.Figure.image`.
Following the Code Style, no underscore is used (please see https://www.pygmt.org/dev/contributing.html#contributing-code). However, the GMTjl documentation uses an underscore, i.e., `bit_color` (please see https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/image/index.html#imag).

**Upstream GMT documentation**: https://docs.generic-mapping-tools.org/dev/image.html#g

**Preview**: https://pygmt-dev--2615.org.readthedocs.build/en/2615/api/generated/pygmt.Figure.image.html

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/image.py
+++ b/pygmt/src/image.py
@@ -11,6 +11,7 @@ from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, us
     J="projection",
     D="position",
     F="box",
+    G="bit_color",
     M="monochrome",
     V="verbose",
     c="panel",
--- a/pygmt/src/image.py
+++ b/pygmt/src/image.py
@@ -52,6 +52,12 @@ def image(self, imagefile, **kwargs):
         using :gmt-term:`MAP_FRAME_PEN`.
     bit_color : str
         [*color*][**+b**\|\ **f**\|\ **t**]
+        Change certain pixel values to another color or make them transparent.
+        For 1-bit images you can specify an alternate *color* for the
+        background (**+b**) or the foreground (**+f**) pixels, or give no color
+        to make those pixels transparent. Alternatively, for color images you
+        can select a single *color* that should be made transparent instead
+        (**+t**). This option may be repeated with different settings.
     monochrome : bool
         Convert color image to monochrome grayshades using the (television)
         YIQ-transformation.
--- a/pygmt/src/image.py
+++ b/pygmt/src/image.py
@@ -55,9 +55,9 @@ def image(self, imagefile, **kwargs):
         Change certain pixel values to another color or make them transparent.
         For 1-bit images you can specify an alternate *color* for the
         background (**+b**) or the foreground (**+f**) pixels, or give no color
-        to make those pixels transparent. Alternatively, for color images you
-        can select a single *color* that should be made transparent instead
-        (**+t**). Can be repeated with different settings.
+        to make those pixels transparent. Can be repeated with different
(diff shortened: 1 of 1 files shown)


pygmt/src/image.py
```diff
@@ -49,6 +50,14 @@ def image(self, imagefile, **kwargs):
         [**+p**\ [*pen*]][**+r**\ [*radius*]][**+s**\ [[*dx*/*dy*/][*shade*]]].
         If set to ``True``, draw a rectangular border around the image
         using :gmt-term:`MAP_FRAME_PEN`.
+    bitcolor : str or list
```
weiji14, line 53: Is a list allowed?

yvonnefroehlich: A list can be used to set both foreground and background colors of 1-bit images:
```python
import pygmt

fig = pygmt.Figure()

for i_G in [
        None,
        "blue+b",
        ["blue+b", "red+f"],
]:
    fig.image(imagefile="@vader1.png", G=i_G)  
    fig.shift_origin(xshift="1c")

fig.show()
# fig.savefig(fname="image_Q_list.png")
```
_Output figure_:
image_Q_list

weiji14: Ah I see, thanks for the example! So the list would translate to `-Gblue+b -Gred+f`. Makes sense.

weiji14 (approved):

pygmt/src/image.py
yvonnefroehlich: Exactly! Maybe moving the sentence "Can be repeated with different settings." upwards makes this a bit clearer (please see commit https://github.com/GenericMappingTools/pygmt/pull/2615/commits/951a11072cef0a72bda75b2f4f1ca76dac3eb967)?

michaelgrund (approved):

yvonnefroehlich:
@GenericMappingTools/pygmt-maintainers I just realized that for [`pygmt.Figure.grdimage`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.grdimage.html#pygmt.Figure.grdimage) the parameter `bit_color` has an underscore. In this PR for the similar parameter of [`pygmt.Figure.image`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.image.html#pygmt.Figure.image), I used no underscore because of the Code Style. However, I feel it is confusing to have different versions across different functions or methods. Does it make sense, that I change `bitcolor` to `bit_color` for `pygmt.Figure.image` (even we break the Code Style)? This would cause no issue with backwards compatibility, as this new alias goes into v0.10.0, but changing it the other way around for `pygmt.Figure.grdimage` would.

seisman:
> Does it make sense, that I change `bitcolor` to `bit_color` for `pygmt.Figure.image` (even we break the Code Style)? This would cause no issue with backwards compatibility, as this new alias goes into v0.10.0, but changing it the other way around for `pygmt.Figure.grdimage` would.

I think it makes more sense to change `bit_color` to `bitcolor`, but we need to keep backward compatibility for at least two more minor versions.

yvonnefroehlich:
> [`pygmt.Figure.image`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.image.html#pygmt.Figure.image)

Please see PR #2635 for this change.

Issue #2629, opened 2023-08-15
datasets.load_earth_mask: Unexpected display of remote dataset "earth_mask"


yvonnefroehlich:
### Description of the problem

When plotting the dataset `earth_mask` via `pygmt.Figure.grdimage` there is a difference between using `grid = "@earth_mask_01d_g"` and `grid = pygmt.datasets.load_earth_mask()`. This is not the case for the other remote datasets, e.g. `earth_relief`.

![datasets_mask_size](https://github.com/GenericMappingTools/pygmt/assets/94163266/82b65834-b3a9-4d55-8ee3-1d6ac56424e1)

### Minimal Complete Verifiable Example

```python
import pygmt

fig = pygmt.Figure()

with fig.subplot(nrows=2, ncols=2, figsize=("20c", "14c")):

    for i_grid in range(2):
                
        if i_grid == 0:
            grid_at = "@earth_relief_01d_g"
            grid_load = pygmt.datasets.load_earth_relief()
        elif i_grid == 1:
            grid_at = "@earth_mask_01d_g"
            grid_load = pygmt.datasets.load_earth_mask()
        
        # Left column
        fig.grdimage(grid=grid_at, panel=True, frame="+t@" + grid_at)
        
        # Right column
        fig.grdimage(grid=grid_load, panel=True, frame="+tdatasets.load_" + grid_at[1:11])
 
fig.show()
# fig.savefig("datasets_mask_size.png")
```

### Full error message

```python-traceback
No error message occurs. The two maps related to the dataset `earth_mask` are expected to be identical.
```

### System information

```bash
PyGMT information:
  version: v0.9.1.dev120
System information:
  python: 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 17:59:51) [MSC v.1935 64 bit (AMD64)]
  executable: C:\ProgramData\Anaconda3\envs\pygmt_env_dev\python.exe
  machine: Windows-10-10.0.19045-SP0
Dependency information:
  numpy: 1.24.3
  pandas: 2.0.2
  xarray: 2023.1.1.dev17
  netCDF4: 1.6.2
  packaging: 23.1
  contextily: 1.3.0
  geopandas: 0.13.2
  IPython: 8.14.0
  rioxarray: 0.14.1
  ghostscript: 9.54.0
GMT library information:
  binary version: 6.4.0
  cores: 4
  grid layout: rows
  image layout: 
[...]

PR #2632, opened 2023-08-16
load_earth_mask: Keep data's encoding to correctly infer data's registration and gtype information

seisman:

See https://github.com/GenericMappingTools/pygmt/issues/2629 for the bug report.

The bug is because:

1. In the GMT accessor, we rely on `grid.encoding["source"]` to determine the grid's registration and gtype
2. For the earth_mask dataset, we use `return grid.astype("int8")`. It copies the old xarray object to a new one, converts the new grid's data array to `int8` type and returns the new xarary object. (ref: https://docs.xarray.dev/en/stable/generated/xarray.DataArray.astype.html)
3. During the "int8" conversion, the `grid.encoding` information is lost.

This PR fixes the bug by directly working on the grid's data array. I also updated the test to check if the returned grid has the correct gtype.

Closes https://github.com/GenericMappingTools/pygmt/issues/2629.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_datasets_earth_mask.py
+++ b/pygmt/tests/test_datasets_earth_mask.py
@@ -54,6 +54,7 @@ def test_earth_mask_01d_with_region():
     assert data.shape == (7, 12)
     assert data.gmt.registration == 0
     assert data.gmt.gtype == 1
+    assert data.dtype == "int8"
     npt.assert_allclose(data.lat, np.arange(13, 20, 1))
     npt.assert_allclose(data.lon, np.arange(-7, 5, 1))
     npt.assert_allclose(data[1, 5], 1)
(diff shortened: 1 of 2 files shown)


seisman:
BTW, the upstream discussions in https://github.com/pydata/xarray/issues/1297 and https://github.com/pydata/xarray/issues/1614 provide more background information.

pygmt/tests/test_datasets_earth_mask.py
```diff
@@ -52,6 +53,7 @@ def test_earth_mask_01d_with_region():
     data = load_earth_mask(resolution="01d", region=[-7, 4, 13, 19])
     assert data.shape == (7, 12)
     assert data.gmt.registration == 0
+    assert data.gmt.gtype == 1
```
weiji14, line 56: Check the dtype here too.
```suggestion
    assert data.gmt.gtype == 1
    assert data.dtype == "int8"
```

weiji14 (approved):
Looks good, just one minor suggested check on the dtype for `test_earth_mask_01d_with_region`

michaelgrund (approved):

PR #2636, opened 2023-08-21
Fix tests for Aug 2023 updated remote datasets

seisman:
**Description of proposed changes**

Some tests related to the GMT remote datasets are failing due to the recent updates of these datasets (see https://www.generic-mapping-tools.org/remote-datasets/changes.html for a list of these changes). The min/max values of these datasets change slightly in the new versions.

This PR fixes these failing tests by updating the min/max values.

**Notes:**

1. In the tests, we use statements like `npt.assert_allclose(data.min(), -180.40002, rtol=1e-5)` to compare the grid value with the expected value. All datasets have a specific precision (e.g., the `earth_geoid` dataset has a precision of 0.01), thus it makes no sense to compare `earth_geoid` value with a number like `-180.40002` and a relative tolerance of 1.0e-5.  Instead, it makes more sense to compare the value with `-180.4` and an absolute tolerance of 0.01 (the precision of the earth_geoid dataset). I've changed `rtol` to `atol` in all these tests.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2638, opened 2023-08-21
Figure.text: Support non-ASCII characters in the 'text' parameter

seisman:

Address https://github.com/GenericMappingTools/pygmt/issues/2204 following #2584.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -69,7 +69,7 @@ def text_(
     The text strings passed via the ``text`` parameter can contain ASCII
     characters and non-ASCII characters defined in the ISOLatin1+ encoding
     (i.e., IEC_8859-1), and the Symbol and ZapfDingbats character sets.
-    Ses :gmt-docs:`cookbook/octal-codes.html` for the full list of supported
+    See :gmt-docs:`cookbook/octal-codes.html` for the full list of supported
     non-ASCII characters.
 
     Full option list at :gmt-docs:`text.html`
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -10,6 +10,7 @@ from pygmt.helpers import (
     fmt_docstring,
     is_nonstr_iter,
     kwargs_to_strings,
+    non_ascii_to_octal,
     use_alias,
 )
 
@@ -223,7 +224,9 @@ def text_(
 
     # Append text at last column. Text must be passed in as str type.
     if kind == "vectors":
-        extra_arrays.append(np.atleast_1d(text).astype(str))
+        extra_arrays.append(
+            list(map(non_ascii_to_octal, np.atleast_1d(text).astype(str)))
+        )
 
     with Session() as lib:
         file_context = lib.virtualfile_from_data(
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -231,7 +231,7 @@ def text_(
     # Append text at last column. Text must be passed in as str type.
     if kind == "vectors":
         extra_arrays.append(
-            list(map(non_ascii_to_octal, np.atleast_1d(text).astype(str)))
+            np.vectorize(non_ascii_to_octal)(np.atleast_1d(text).astype(str))
         )
 
     with Session() as lib:
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -66,6 +66,12 @@ def text_(
     - ``x``/``y``, and ``text``
     - ``position`` and ``text``
 
+    The text strings passed via the ``text`` parameter can contain ASCII
+    characters and non-ASCII characters defined in the ISOLatin1+ encoding
+    (i.e., IEC_8859-1), and the Symbol and ZapfDingbats character sets.
+    Ses :gmt-docs:`cookbook/octal-codes.html` for the full list of supported
+    non-ASCII characters.
+
     Full option list at :gmt-docs:`text.html`
 
     {aliases}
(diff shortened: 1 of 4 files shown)


seisman:
Here is the baseline image generated by the new test:
test_text_nonascii

I'm having some technical difficulties when running `dvc push` so can't push it to DAGsHub.

seisman:
> I'm having some technical difficulties when running `dvc push` so can't push it to DAGsHub.

Suddenly it works for me. It may be some network issues yesterday.

pygmt/src/text.py
```diff
@@ -65,6 +66,12 @@ def text_(
     - ``x``/``y``, and ``text``
     - ``position`` and ``text``
 
+    The text strings passed via the ``text`` parameter can contain ASCII
+    characters and non-ASCII characters defined in the ISOLatin1+ encoding
+    (i.e., IEC_8859-1), and the Symbol and ZapfDingbats character sets.
+    Ses :gmt-docs:`cookbook/octal-codes.html` for the full list of supported
+    non-ASCII characters.
```
yvonnefroehlich, line 73:
```suggestion
    See :gmt-docs:`cookbook/octal-codes.html` for the full list of supported
    non-ASCII characters.
```

pygmt/src/text.py
```diff
@@ -223,7 +230,9 @@ def text_(
 
     # Append text at last column. Text must be passed in as str type.
     if kind == "vectors":
-        extra_arrays.append(np.atleast_1d(text).astype(str))
+        extra_arrays.append(
+            list(map(non_ascii_to_octal, np.atleast_1d(text).astype(str)))
```
weiji14, line 234: This would be a list of numpy arrays, instead of being just a numpy array? Is there a way to keep it as a numpy array?

pygmt/tests/test_text.py
```diff
@@ -371,3 +371,16 @@ def test_text_nonstr_text():
...
+@pytest.mark.mpl_image_compare
+def test_text_nonascii():
+    """
+    Test passing text strings with non-ascii characters.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=True)
+    fig.text(position="TL", text="position-text:°α")
+    fig.text(x=1, y=1, text="xytext:°α")
+    fig.text(x=[5, 5], y=[3, 5], text=["xytext1:°α", "xytext2:°α"])
```
weiji14, line 385: Mix in some Symbol and ZapfDingbats characters too from https://docs.generic-mapping-tools.org/latest/cookbook/octal-codes.html#octal-codes-symbol-zap?

pygmt/src/text.py
seisman: Changed to `np.vectorize` in 85a0926e457fed9a484dc55845fb87377e258c57.

pygmt/tests/test_text.py
seisman: Done in adf6eda9d4f88de10ec4169e6f662d0796a48e73.

seisman:
The new test fails, because I generated the baseline image using gs 9.56.1 but our CI is still using gs 9.54, but why can't the CI use gs 9.56.1?

weiji14:
> The new test fails, because I generated the baseline image using gs 9.56.1 but our CI is still using gs 9.54, but why can't the CI use gs 9.56.1?

Ghostscript 9.56.1 was just released 3 days ago on conda-forge - https://anaconda.org/conda-forge/ghostscript/files?version=9.56.1. I can also install 9.56.1 on my system, so not sure why CI still uses gs 9.54, let me check the dependency graph.

weiji14:
Actually, maybe we don't want gs 9.56.1 for now :sweat_smile: I just ran `make test` locally with gs 9.56.1 and got 193 failed unit tests, pretty much all due to tiny image differences...

seisman:
> Actually, maybe we don't want gs 9.56.1 for now 😅 I just ran `make test` locally with gs 9.56.1 and got 193 failed unit tests, pretty much all due to tiny image differences...

OK. I think we can use gs 9.54 now.

FYI, gs 10.02.0 is likely the next gs version that doesn't have serious bugs (https://github.com/GenericMappingTools/gmt/issues/7336#issuecomment-1690852169) and it's likely to be released soon (rc1 is already been released). I think we should only update these baseline images when gs 10.02.0 is out.

weiji14:
> > Actually, maybe we don't want gs 9.56.1 for now 😅 I just ran `make test` locally with gs 9.56.1 and got 193 failed unit tests, pretty much all due to tiny image differences...
> 
> OK. I think we can use gs 9.54 now.
> 
> FYI, gs 10.02.0 is likely the next gs version that doesn't have serious bugs ([GenericMappingTools/gmt#7336 (comment)](https://github.com/GenericMappingTools/gmt/issues/7336#issuecomment-1690852169)) and it's likely to be released soon (rc1 is already been released). I think we should only update these baseline images when gs 10.02.0 is out.

Sounds good, better to update the baseline images once for gs 10.02 rather than twice, we can track the ghostscript 10.0 release on conda-forge at https://github.com/conda-forge/ghostscript-feedstock/issues/26. Just realized you're a maintainer for the ghostscript conda-forge recipe too :laughing:

weiji14 (approved):
Looks good to me. The 16 failing tests on CI will be fixed in https://github.com/GenericMappingTools/pygmt/pull/2636, so can safely ignore.

PR #2650, opened 2023-08-31
Correct the datum of the earth_relief dataset

ZMAlt:
earth elvation and bathymetry of earth_relief are relative to ellipsoid, ie. geodetic height or ellipsoidal height.

### DIFF
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -168,9 +168,9 @@ datasets = {
     "earth_relief": GMTRemoteDataset(
         title="Earth relief",
         name="elevation",
-        long_name="Earth elevation relative to the geoid",
+        long_name="Earth elevation relative to the WGS84 reference ellipsoid",
         units="meters",
-        extra_attributes={"vertical_datum": "EGM96", "horizontal_datum": "WGS84"},
+        extra_attributes={"datum": "WGS84"},
         resolutions={
             "01d": Resolution(["gridline", "pixel"], False),
             "30m": Resolution(["gridline", "pixel"], False),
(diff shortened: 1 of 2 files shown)


pygmt/datasets/load_remote_dataset.py
```diff
@@ -168,9 +168,9 @@ class GMTRemoteDataset(NamedTuple):
     "earth_relief": GMTRemoteDataset(
         title="Earth relief",
         name="elevation",
-        long_name="Earth elevation relative to the geoid",
+        long_name="Earth elevation relative to the WGS84 reference ellipsoid",
         units="meters",
-        extra_attributes={"vertical_datum": "EGM96", "horizontal_datum": "WGS84"},
+        extra_attributes={"datum": "WGS84"},
```
weiji14, line 173: Just want to double check that this dataset is indeed referenced to the WGS84 ellipsoid instead of the EGM96 geoid? The original paper at https://doi.org/10.1029/2019EA000658 has this line:

> 4. Ensure all elevations are referenced to the geoid (i.e., elevations are set as orthometric height in meters). Both ArcticDEM and REMA elevations are referenced relative to the WGS84 ellipsoid datum. Hence, the vertical difference between the geoid and ellipsoid, as defined by the EGM96 geoid model (Lemoine et al., 1997), was determined at each grid cell and subtracted from the original elevations. The GMT functions grdsample and grdmath were used to perform this operation.

Which sounds a bit ambiguous. I'm not sure if GMT has reprocessed the grids to be relative to the ellipsoid. Should check the scripts at https://github.com/GenericMappingTools/gmtserver-admin

ZMAlt: You are right. I misremembered the datum. SRTM is indeed relative to EGM96 geoid(ie, orthometric height). 

> Both ArcticDEM and REMA elevations are referenced relative to the WGS84 ellipsoid datum

They converted ellipsoidal heights from ArcticDEM and REMA to orthometric heights.

Issue #2689, opened 2023-09-14
Support "jpeg" as file extension


seisman:
```python
In [1]: import pygmt
   ...: fig = pygmt.Figure()
   ...: fig.basemap(region=[0, 10, 0, 10], frame=True)
   ...: fig.savefig("map.jpeg")
---------------------------------------------------------------------------
GMTInvalidInput                           Traceback (most recent call last)
Cell In[1], line 4
      2 fig = pygmt.Figure()
      3 fig.basemap(region=[0, 10, 0, 10], frame=True)
----> 4 fig.savefig("map.jpeg")

File /pygmt/figure.py:315, in Figure.savefig(self, fname, transparent, crop, anti_alias, show, **kwargs)
    310     if ext == "ps":
    311         raise GMTInvalidInput(
    312             "Extension '.ps' is not supported. "
    313             "Please use '.eps' or '.pdf' instead."
    314         )
--> 315     raise GMTInvalidInput(f"Unknown extension '.{ext}'.")
    316 fmt = fmts[ext]
    317 if transparent:

GMTInvalidInput: Unknown extension '.jpeg'.
```
Currently, only `jpg` is supported, `jpeg` is not, and GMT supports both.

Issue #2683, opened 2023-09-11
Support upper-case file extensions


seisman:
### Description of the desired feature

Currently, PyGMT only allows lower-case file extensions like pdf and png, but doesn't support upper-case extensions like PDF or PNG. The Windows OS is case insensitive, so both lower-case and upper-case file extensions should be allowed.

```python
import pygmt
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], frame=True)
fig.savefig("map.PNG")
---------------------------------------------------------------------------
GMTInvalidInput                           Traceback (most recent call last)
Cell In[4], line 1
----> 1 fig.savefig("map.PNG")

File pygmt/pygmt/figure.py:315, in Figure.savefig(self, fname, transparent, crop, anti_alias, show, **kwargs)
    310     if ext == "ps":
    311         raise GMTInvalidInput(
    312             "Extension '.ps' is not supported. "
    313             "Please use '.eps' or '.pdf' instead."
    314         )
--> 315     raise GMTInvalidInput(f"Unknown extension '.{ext}'.")
    316 fmt = fmts[ext]
    317 if transparent:

GMTInvalidInput: Unknown extension '.PNG'.
```

### Are you willing to help implement and maintain this feature?

Maybe

MichaelCni:
Hy there i can see your post and i must say

Normalization:

When dealing with file extensions provided by users or retrieved from external sources, normalize them to lowercase before processing. This ensures consistency.
Validation and Handling:

Ensure that your application correctly validates and handles files with uppercase or lowercase extensions. This includes loading, displaying, and processing the files.
File Type Mapping:

Maintain a mapping of file extensions to corresponding file types or actions in your application. Make sure this mapping is case-insensitive so that both uppercase and lowercase extensions trigger the correct behavior.
User Input Handling:

If your application allows users to input file extensions (e.g., in filters or preferences), make sure that it gracefully handles both uppercase and lowercase inputs.

Thanks and regards
MichaelCni

PR #2691, opened 2023-09-15
Figure.savefig: Support .jpeg as JPEG image extension

seisman:
Support `.jpeg` as file extension. Please note that `gmt psconvert -Tj` always produces images with the `.jpg` extension. It has two downsides:

1. Users may be surprised that they give `.jpeg` but get `.jpg`
2. `show=True` won't work unless we change the codes

I feel it's better to rename the files to have the file extension that users expect. 

Closes https://github.com/GenericMappingTools/pygmt/issues/2689.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -260,7 +260,7 @@ class Figure:
         This method implements a matplotlib-like interface for
         :meth:`pygmt.Figure.psconvert`.
 
-        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg),
+        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg``),
         PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
         KML (``.kml``). The KML output generates a companion PNG file.
 
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -306,7 +306,7 @@ class Figure:
 
         fname = Path(fname)
         prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
-        ext = suffix[1:]
+        ext = suffix[1:]  # Remove the .
         # alias jpeg to jpg
         if ext == "jpeg":
             ext = "jpg"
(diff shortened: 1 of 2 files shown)


pygmt/figure.py
```diff
@@ -260,9 +260,9 @@ def savefig(
         This method implements a matplotlib-like interface for
         :meth:`pygmt.Figure.psconvert`.
 
-        Supported formats: PNG (``.png``), JPEG (``.jpg``), PDF (``.pdf``),
-        BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and KML (``.kml``).
-        The KML output generates a companion PNG file.
+        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg),
+        PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
+        KML (``.kml``). The KML output generates a companion PNG file.
```
yvonnefroehlich, line 265:
```suggestion
        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg``),
        PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
        KML (``.kml``). The KML output generates a companion PNG file.
```

weiji14 (approved):

seisman:
I've revised the codes so that it's much easier to support upper-case extensions (#2683) with very few code changes. Please review again.

pygmt/figure.py
```diff
@@ -304,8 +304,13 @@ def savefig(
             "kml": "g",
         }
 
-        prefix, ext = os.path.splitext(fname)
-        ext = ext[1:]  # Remove the .
+        fname = Path(fname)
+        prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
+        ext = suffix[1:]
```
weiji14, line 309:
```suggestion
        ext = suffix[1:]  # Remove the .
```

weiji14 (approved):
Just adding back the inline comment from `ext = ext[1:]  # Remove the .` before.

PR #2697, opened 2023-09-20
Figure.savefig: Support uppercase file extensions (e.g., PNG, PDF)

### LINKED ISSUE #2683: Support upper-case file extensions

seisman:

Closes #2683

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -306,7 +306,7 @@ class Figure:
 
         fname = Path(fname)
         prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
-        ext = suffix[1:]  # Remove the .
+        ext = suffix[1:].lower()  # Remove the . and normalize to lowercase
         # alias jpeg to jpg
         if ext == "jpeg":
             ext = "jpg"
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -86,8 +86,10 @@ def test_figure_savefig_exists():
     for fmt in "png pdf jpg jpeg bmp eps tif PNG JPG JPEG Png".split():
         fname = ".".join([prefix, fmt])
         fig.savefig(fname)
-        assert os.path.exists(fname)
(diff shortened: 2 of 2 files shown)


pygmt/figure.py
```diff
@@ -306,7 +306,7 @@ def savefig(
 
         fname = Path(fname)
         prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
-        ext = suffix[1:]  # Remove the .
+        ext = suffix[1:].lower()  # Remove the . and normalize to lowercase
```
seisman, line 309: I don't update the docstrings to document this new enhancement because I don't want to encourage users to use uppercase file extensions.

yvonnefroehlich: I agree that the use of uppercase file extensions should be avoided. Thus, I feel it is fine to not explicitly mention this possibility in the documentation.

yvonnefroehlich (commented):
For me it works as expected!

Issue #2658, opened 2023-09-03
Figure.psconvert: Add support for GeoTIFF output


EJFielding:
### Description of the desired feature

In classic GMT, I have long used this command to generate a GeoTIFF file from an EPS file:
```
gmt psconvert -A -P -Tt -W+g $out
```

I see there is a `pygmt.Figure.psconvert` call that wraps the GMT `psconvert` but it does not seem to have the `-W+g` option to make a GeoTIFF.

The GeoTIFF output is extremely helpful for taking a GMT or PyGMT map and loading it into Google Earth or QGIS to view it with other geographic information. I attach an example of a map that I made in GMT that I converted to GeoTIFF and then opened as an image overlay in Google Earth.

![Google_Earth-slip-map](https://github.com/GenericMappingTools/pygmt/assets/7253576/ae8c937c-288a-45e9-88a8-e1bdd85bdb05)

### Are you willing to help implement and maintain this feature?

Maybe

seisman:
@EJFielding 
This feature seems easy to implement in PyGMT. 

After reading the documentation, my impressions are:

- **-W** creates a world file which can be used by other software
- **-W+g** will use this world file and convert PS to GeoTIFF
- **-W+c**: the doc says "This option automatically sets -A -P. Append +c to not crop the image.", so it also affect the generated GeoTIFF
- All other **-W** modifiers are related to creating KML files instead?

Do I understand it correctly?

EJFielding:
I have only used the `-W+g` modifier for my purposes. The GeoTIFF is good for loading in QGIS, ArcGIS, and other GIS systems in addition to Google Earth or Google Maps. I have not used the KML file generation, but I imagine that others use that, so it would be good to support as well.

seisman:
I've done a few tests to understand GMT's behavior about GeoTIFF support. 

The test PS file is generated using the following command:
```
gmt coast -R0/10/0/10 -JM10c -Baf -W1p -ps map
```
I've tried the following commands:
1. `gmt psconvert map.ps -Tt` produces `map.tif` (a normal TIFF file)
2. `gmt psconvert map.ps -W` produces `map.jpw` (the plaintext world file) and `map.jpg` (default image format)
3. `gmt psconvert map.ps -W+g` produces `map.pgw` (the plaintext world file) and `map.tiff` (GeoTIFF)
[...]

PR #2698, opened 2023-09-22
Figure.savefig: Support generating GeoTIFF file (with extension '.tiff')

seisman:
`.tif` means a standard TIFF file, while `.tiff` means a GeoTIFF file.

```python
import pygmt
fig = pygmt.Figure()
fig.coast(region=[0, 10, 0, 10], projection="M10c", frame=True, shorelines="1p")
fig.savefig("map.tiff")
```

Closes https://github.com/GenericMappingTools/pygmt/issues/2658.

**Preview**: https://pygmt-dev--2698.org.readthedocs.build/en/2698/api/generated/pygmt.Figure.savefig.html

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -308,6 +308,7 @@ class Figure:
             "bmp": "b",
             "eps": "e",
             "tif": "t",
+            "tiff": None,  # GeoTIFF doesn't need the -T option
             "kml": "g",
         }
 
@@ -317,10 +318,9 @@ class Figure:
 
         if ext == "jpeg":  # Alias jpeg to jpg
             ext = "jpg"
-        elif ext == "tiff":  # Alias tif to tiff and make it a GeoTIFF
-            ext = "tif"
+        elif ext == "tiff":  # GeoTIFF
             kwargs["W"] = "+g"
-        elif ext == "kml":
+        elif ext == "kml":  # KML
             kwargs["W"] = "+k"
 
         if ext not in fmts:
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -260,9 +260,9 @@ class Figure:
         This method implements a matplotlib-like interface for
         :meth:`pygmt.Figure.psconvert`.
 
-        Supported formats: PNG (``.png``), JPEG (``.jpg``), PDF (``.pdf``),
-        BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and KML (``.kml``).
-        The KML output generates a companion PNG file.
+        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg``),
+        PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
+        KML (``.kml``). The KML output generates a companion PNG file.
 
         You can pass in any keyword arguments that
         :meth:`pygmt.Figure.psconvert` accepts.
@@ -304,8 +304,13 @@ class Figure:
             "kml": "g",
         }
 
-        prefix, ext = os.path.splitext(fname)
-        ext = ext[1:]  # Remove the .
+        fname = Path(fname)
+        prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
+        ext = suffix[1:]  # Remove the .
+        # alias jpeg to jpg
+        if ext == "jpeg":
+            ext = "jpg"
+
         if ext not in fmts:
             if ext == "ps":
                 raise GMTInvalidInput(
@@ -327,6 +332,11 @@ class Figure:
             kwargs["W"] = "+k"
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
+
+        # Rename if file extension doesn't match the input file suffix
+        if ext != suffix[1:]:
+            fname.with_suffix("." + ext).rename(fname)
+
         if show:
             launch_external_viewer(fname)
 
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -257,12 +257,19 @@ class Figure:
         """
         Save the figure to a file.
 
-        This method implements a matplotlib-like interface for
-        :meth:`pygmt.Figure.psconvert`.
+        Supported file formats and their extensions:
 
-        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg``),
-        PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
-        KML (``.kml``). The KML output generates a companion PNG file.
+        - PNG (``.png``)
+        - JPEG (``.jpg`` or ``.jpeg``)
+        - PDF (``.pdf``)
+        - BMP (``.bmp``)
+        - TIFF (``.tif`` or `.tiff`)
+        - EPS (``.eps``)
+        - KML (``.kml``)
+
+        For TIFF format, ``.tiff`` generates a GeoTIFF file with embedded
+        georeferencing information and a companion world file. For KML format,
+        a companion PNG file is also generated.
 
         You can pass in any keyword arguments that
         :meth:`pygmt.Figure.psconvert` accepts.
@@ -279,8 +286,8 @@ class Figure:
             If ``True``, will crop the figure canvas (page) to the plot area.
         anti_alias: bool
             If ``True``, will use anti-aliasing when creating raster images
-            (PNG, JPG, TIFF). More specifically, it passes arguments ``t2``
-            and ``g2`` to the ``anti_aliasing`` parameter of
+            (BMP, PNG, JPEG and TIFF). More specifically, it passes arguments
+            ``t2`` and ``g2`` to the ``anti_aliasing`` parameter of
             :meth:`pygmt.Figure.psconvert`. Ignored if creating vector
             graphics.
         show: bool
@@ -307,9 +314,14 @@ class Figure:
         fname = Path(fname)
         prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
         ext = suffix[1:].lower()  # Remove the . and normalize to lowercase
-        # alias jpeg to jpg
-        if ext == "jpeg":
+
+        if ext == "jpeg":  # Alias jpeg to jpg
             ext = "jpg"
+        elif ext == "tiff":  # Alias tif to tiff and make it a GeoTIFF
+            ext = "tif"
+            kwargs["W"] = "+g"
+        elif ext == "kml":
+            kwargs["W"] = "+k"
 
         if ext not in fmts:
             if ext == "ps":
@@ -328,8 +340,6 @@ class Figure:
         if anti_alias:
             kwargs["Qt"] = 2
             kwargs["Qg"] = 2
-        if ext == "kml":
-            kwargs["W"] = "+k"
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
 
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -263,7 +263,8 @@ class Figure:
         - JPEG (``.jpg`` or ``.jpeg``)
         - PDF (``.pdf``)
         - BMP (``.bmp``)
-        - TIFF (``.tif`` or `.tiff`)
+        - TIFF (``.tif``)
+        - GeoTIFF (``.tiff``)
         - EPS (``.eps``)
         - KML (``.kml``)
 
@@ -286,10 +287,10 @@ class Figure:
             If ``True``, will crop the figure canvas (page) to the plot area.
         anti_alias: bool
             If ``True``, will use anti-aliasing when creating raster images
-            (BMP, PNG, JPEG and TIFF). More specifically, it passes arguments
-            ``t2`` and ``g2`` to the ``anti_aliasing`` parameter of
-            :meth:`pygmt.Figure.psconvert`. Ignored if creating vector
-            graphics.
+            (BMP, PNG, JPEG, and TIFF). More specifically, it passes the
+            arguments ``"t2"`` and ``"g2"`` to the ``anti_aliasing``
+            parameter of :meth:`pygmt.Figure.psconvert`. Ignored if
+            creating vector graphics.
         show: bool
             If ``True``, will open the figure in an external viewer.
         dpi : int
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -306,7 +306,7 @@ class Figure:
 
         fname = Path(fname)
         prefix, suffix = fname.with_suffix("").as_posix(), fname.suffix
-        ext = suffix[1:]  # Remove the .
+        ext = suffix[1:].lower()  # Remove the . and normalize to lowercase
         # alias jpeg to jpg
         if ext == "jpeg":
             ext = "jpg"
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -268,9 +268,7 @@ class Figure:
         - EPS (``.eps``)
         - KML (``.kml``)
 
-        For TIFF format, ``.tiff`` generates a GeoTIFF file with embedded
-        georeferencing information and a companion world file. For KML format,
-        a companion PNG file is also generated.
+        For KML format, a companion PNG file is also generated.
 
         You can pass in any keyword arguments that
         :meth:`pygmt.Figure.psconvert` accepts.
@@ -287,10 +285,10 @@ class Figure:
             If ``True``, will crop the figure canvas (page) to the plot area.
         anti_alias: bool
             If ``True``, will use anti-aliasing when creating raster images
-            (BMP, PNG, JPEG, and TIFF). More specifically, it passes the
-            arguments ``"t2"`` and ``"g2"`` to the ``anti_aliasing``
-            parameter of :meth:`pygmt.Figure.psconvert`. Ignored if
-            creating vector graphics.
+            (BMP, PNG, JPEG, TIFF, and GeoTIFF). More specifically, it passes
+            the arguments ``"t2"`` and ``"g2"`` to the ``anti_aliasing``
+            parameter of :meth:`pygmt.Figure.psconvert`. Ignored if creating
+            vector graphics.
         show: bool
             If ``True``, will open the figure in an external viewer.
         dpi : int
@@ -343,6 +341,11 @@ class Figure:
             kwargs["Qg"] = 2
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
+
+        # Remove the .pgw world file if exists
+        # See upstream fix https://github.com/GenericMappingTools/gmt/pull/7865
+        if ext == "tiff" and fname.with_suffix(".pgw").exists():
+            fname.with_suffix(".pgw").unlink()
 
         # Rename if file extension doesn't match the input file suffix
         if ext != suffix[1:]:
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -342,7 +342,8 @@ class Figure:
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
 
-        # Remove the .pgw world file if exists
+        # Remove the .pgw world file if exists 
+        # Not necessary after GMT 6.5.0.
         # See upstream fix https://github.com/GenericMappingTools/gmt/pull/7865
         if ext == "tiff" and fname.with_suffix(".pgw").exists():
             fname.with_suffix(".pgw").unlink()
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -417,20 +417,16 @@ class Figure:
 
         if method not in ["external", "notebook", "none"]:
             raise GMTInvalidInput(
-                (
-                    f"Invalid display method '{method}', "
-                    "should be either 'notebook', 'external', or 'none'."
-                )
+                f"Invalid display method '{method}', "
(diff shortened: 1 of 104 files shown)


pygmt/figure.py
```diff
@@ -257,12 +257,19 @@ def savefig(
...
+        Supported file formats and their extensions:
 
-        Supported formats: PNG (``.png``), JPEG (``.jpg`` or ``.jpeg``),
-        PDF (``.pdf``), BMP (``.bmp``), TIFF (``.tif``), EPS (``.eps``), and
-        KML (``.kml``). The KML output generates a companion PNG file.
+        - PNG (``.png``)
+        - JPEG (``.jpg`` or ``.jpeg``)
+        - PDF (``.pdf``)
+        - BMP (``.bmp``)
+        - TIFF (``.tif`` or `.tiff`)
```
yvonnefroehlich, line 266:
```suggestion
        - TIFF (``.tif`` or ``.tiff``)
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
 
+def test_figure_savefig_geotiff():
+    """
+    Make sure .tif generates a normal TIFF file and .tiff generates a GeoTIFF
+    file.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 10, 0, 10], projection="M10c", frame=True)
+
+    # save as GeoTIFF
```
yvonnefroehlich, line 103:
```suggestion
    # Save as GeoTIFF
```

pygmt/figure.py
```diff
@@ -279,8 +286,8 @@ def savefig(
             If ``True``, will crop the figure canvas (page) to the plot area.
         anti_alias: bool
             If ``True``, will use anti-aliasing when creating raster images
-            (PNG, JPG, TIFF). More specifically, it passes arguments ``t2``
-            and ``g2`` to the ``anti_aliasing`` parameter of
+            (BMP, PNG, JPEG and TIFF). More specifically, it passes arguments
+            ``t2`` and ``g2`` to the ``anti_aliasing`` parameter of
             :meth:`pygmt.Figure.psconvert`. Ignored if creating vector
             graphics.
```
yvonnefroehlich, line 292:
```suggestion
            (BMP, PNG, JPEG, and TIFF). More specifically, it passes the
            arguments ``"t2"`` and ``"g2"`` to the ``anti_aliasing``
            parameter of :meth:`pygmt.Figure.psconvert`. Ignored if
            creating vector graphics.
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
+    fig.savefig(geofname)
+    assert geofname.exists()
+    assert geofname.with_suffix(".pgw").exists()  # The companion world file
+
+    # Save as TIFF
+    fname = Path("test_figure_savefig_geotiff.tif")
+    fig.savefig(fname)
+    assert fname.exists()
+
+    # Check is a TIFF is georeferenced or not
```
yvonnefroehlich, line 114:
```suggestion
    # Check if a TIFF is georeferenced or not
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
+    fig.basemap(region=[0, 10, 0, 10], projection="M10c", frame=True)
+
+    # Save as GeoTIFF
+    geofname = Path("test_figure_savefig_geotiff.tiff")
+    fig.savefig(geofname)
+    assert geofname.exists()
+    assert geofname.with_suffix(".pgw").exists()  # The companion world file
+
+    # Save as TIFF
+    fname = Path("test_figure_savefig_geotiff.tif")
```
weiji14, line 110:
```suggestion
    fname = Path("test_figure_savefig_tiff.tif")
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
+    file.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 10, 0, 10], projection="M10c", frame=True)
+
+    # Save as GeoTIFF
+    geofname = Path("test_figure_savefig_geotiff.tiff")
+    fig.savefig(geofname)
+    assert geofname.exists()
+    assert geofname.with_suffix(".pgw").exists()  # The companion world file
```
weiji14, line 107: Strange, why is a pgw file generated instead of a tfw file? See https://en.wikipedia.org/wiki/World_file#Filename_extension. A .tif or .tiff file should be linked to .tfw no?

pygmt/figure.py
```diff
@@ -257,12 +257,19 @@ def savefig(
...
+        - PNG (``.png``)
+        - JPEG (``.jpg`` or ``.jpeg``)
+        - PDF (``.pdf``)
+        - BMP (``.bmp``)
+        - TIFF (``.tif`` or ``.tiff``)
+        - EPS (``.eps``)
+        - KML (``.kml``)
+
+        For TIFF format, ``.tiff`` generates a GeoTIFF file with embedded
+        georeferencing information and a companion world file. For KML format,
```
weiji14, line 271: Maybe clearer to split into two lines?

```suggestion
        - TIFF (``.tif``)
        - GeoTIFF (``.tiff``)
        - EPS (``.eps``)
        - KML (``.kml``)

        For TIFF format, ``.tiff`` generates a GeoTIFF file with embedded
        georeferencing information and a companion world file. For KML format,
```

pygmt/tests/test_figure.py
seisman: See https://github.com/GenericMappingTools/pygmt/issues/2658#issuecomment-1712392325 for some tests.

Here, we call `gmt psconvert -A -P -W+g` without specifying the `-T` option. 

Looking at the `psconvert` source code (https://github.com/GenericMappingTools/gmt/blob/master/src/psconvert.c#L1008-L1010), 
```
	if (!Ctrl->T.active) {	/* Set default output device if none is specified */
		Ctrl->T.device = (Ctrl->W.warp) ? GS_DEV_PNG : GS_DEV_JPG;	/* Lossless PNG if we are making a geotiff in the end */
	}
```
The default format is PNG is `-W+g` is used. I think GMT fist converts PS to lossless PNG, then write the world file, and calls `gdal_translate` to combine the PNG and world file into a GeoTIFF file. The extension of the world file is determined from the extension of the raster file (in this case PNG), thus we have a `pgw` file. This may be an upstream feature or bug.

Actually, since we already have a `GeoTIFF` file, the world_file is no longer needed. Maybe we should remove it?

weiji14: Ah ok, thanks for the explanation. You're right that we wouldn't need the `pgw` file or any world file anyway since the GeoTIFF should already have the metadata embedded inside.

> Maybe we should remove it?

Maybe best to add the code to remove the `pgw` file upstream in GMT when a GeoTIFF is created, or at least rename `pgw` to `tfw`? On the PyGMT side, we could match the upstream implementation with GMT <= 6.4.0.

seisman: > Maybe best to add the code to remove the `pgw` file upstream in GMT when a GeoTIFF is created, or at least rename `pgw` to `tfw`?

Removing `pgw` makes more sense to me. I'll open an issue report and see what the GMT team thinks about it.

> On the PyGMT side, we could match the upstream implementation with GMT <= 6.4.0.

Do you mean we should keep the `pgw` file? I'm thinking about adding a new parameter `worldfile=True` to `Figure.savefig()` which generates a companion world file.

weiji14: > > On the PyGMT side, we could match the upstream implementation with GMT <= 6.4.0.
> 
> Do you mean we should keep the `pgw` file? I'm thinking about adding a new parameter `worldfile=True` to `Figure.savefig()` which generates a companion world file.

I mean we should match whatever GMT does with the `pgw` world file later in GMT 6.5.0, but open up that issue first. The `worldfile=True` parameter sounds nice actually, this would be similar to GDAL's `TFW=YES/NO` option when creating GeoTIFFs - https://gdal.org/drivers/raster/gtiff.html#creation-options

seisman:
Ping @GenericMappingTools/pygmt-maintainers for reviews

pygmt/figure.py
```diff
@@ -328,11 +339,14 @@ def savefig(
...
             kwargs["Qg"] = 2
-        if ext == "kml":
-            kwargs["W"] = "+k"
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
 
+        # Remove the .pgw world file if exists
+        # See upstream fix https://github.com/GenericMappingTools/gmt/pull/7865
+        if ext == "tiff" and fname.with_suffix(".pgw").exists():
+            fname.with_suffix(".pgw").unlink()
```
weiji14, line 348: Not sure how many years into the future we'll set GMT 6.5.0 as the minimum required version, but let's add a note as a reminder. Feel free to reword before commiting.

```suggestion
        # Remove the .pgw world file if exists (remove when PyGMT sets minimum
        # required GMT version to 6.5.0).
        # See upstream fix https://github.com/GenericMappingTools/gmt/pull/7865
        if ext == "tiff" and fname.with_suffix(".pgw").exists():
            fname.with_suffix(".pgw").unlink()
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
+
+    # Check if a TIFF is georeferenced or not
+    try:
+        # pylint: disable=import-outside-toplevel
+        import rioxarray
+        from rasterio.errors import NotGeoreferencedWarning
+
+        # GeoTIFF
+        with rioxarray.open_rasterio(geofname) as xds:
+            assert xds.rio.crs is not None
```
weiji14, line 123: The output CRS from this 10cm Mercator projection map is:

```
PROJCS["unknown",GEOGCS["unknown",DATUM["unknown",SPHEROID["unknown",6378137,298.257220143428]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]]],PROJECTION["Mercator_1SP"],PARAMETER["central_meridian",5],PARAMETER["scale_factor",1],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Easting",EAST],AXIS["Northing",NORTH]]
```

I tried dragging and dropping the `test_figure_savefig_geotiff.tiff` into QGIS, and the projection showed up as 'unknown' (same as rioxarray above), but it defaulted to plotting on longitude 0-10, latitude 0-10:

image

However, the blue gridlines (actual longitude/latitude lines from QGIS) don't match up with the GeoTIFF (from psconvert), see the top left corner around 10°N, where the black/white zebra grid markings don't line up with the blue lines.

image

Not sure what the default projection is that is output from psconvert, but I think we should make sure there isn't a bug somewhere. Maybe we should try a few other projections besides `M10c` and see if the output is ok?

@EJFielding, could you share with us the projection (`-J`) parameter you used to generate the overlay plot over Mexico at https://github.com/GenericMappingTools/pygmt/issues/2658#issue-1879147130? Just wanted to know how you're aligning the grids properly.

pygmt/tests/test_figure.py
weiji14: Note that https://github.com/GenericMappingTools/gmt/pull/7865 was merged, and the upcoming GMT 6.5.0 should delete the `pgw` world file if a GeoTIFF output is requested.

pygmt/tests/test_figure.py
weiji14: Oh wait nevermind, I think I had a cached `tiff.aux.xml` file from trying a few other projections like `-JEPSG:3857`, and it was rendering the GeoTIFF in the wrong projection. Deleting all the files and resaving them seems to work better now:

image

There's still a 900m gap between the 9°N blue longitude line and the black/white zebra grid border, but I think that's just because of the default dpi resolution being a bit coarse.

weiji14, line 123: Add some more assert statements on the lon/lat bounds and shape of the raster.

```suggestion
            assert xds.rio.crs is not None
npt.assert_allclose(
                actual=xds.rio.bounds(),
                desired=(
                    -667728.5869475466,
                    6356999.491390044,
                    594269.0244906943,
                    8381540.2498699855,
                ),
            )
            assert xds.rio.shape == (2149, 1340)
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,46 @@ def test_figure_savefig_exists():
...
+        # TIFF
+        with pytest.warns(expected_warning=NotGeoreferencedWarning) as record:
+            with rioxarray.open_rasterio(fname) as xds:
+                assert xds.rio.crs is None
```
weiji14, line 127: More bounding box and shape tests for the ungeoreferenced TIFF image.
```suggestion
                assert xds.rio.crs is None
                npt.assert_allclose(
                    actual=xds.rio.bounds(), desired=(0.0, 0.0, 1331.0, 1257.0)
                )
                assert xds.rio.shape == (1257, 1331)
```

EJFielding:
In case it helps, here is the GMT6 classic mode script that I used to make the GeoTIFF map (sorry for the ugly csh scripting):
```
# chose file to plot
set file = $1
set scale = $2
# CSI psvelo output in cm, not m
set GPSscale = 0.2
set Fault = 'Acapulco'
set topoDir = ../../topo
set vecDir = $WORK_DIR/vectors
set GPSdir = $WORK_DIR/modeling/CSI/GPS
set GPSfile = 'Acapulco_v1.psvelo'
set GPSstations = 'Acapulco_v1_sites.txt'

set LLreg = -R-101.0/16.0/-98.5/18.0r
set LLproj = -Ju14/1:2000000

mapproject Acapulco-corn-dd.xy -R-102/-98/10/20 -Ju14/1:1 -F -C > Acapulco-corn-UTM14.xy
# these map units seem to be 100 times greater, probably cm vs. m
set proj = -Jx0.5e-4
set reg = "-R285980/552932/1769965/1990256"
set projkm = -Jx0.5e-1
set regkm = "-R285.980/552.932/1769.965/1990.256"

# new map view with vectors and error ellipses
set mapOut = $Fault-slip-map5.eps

set slipd = ${Fault}.slipdir
set fault = $Fault.fault

set out = $file-Acapulco-map5.eps
set in_cpt = polar
set cpt = co_unw-red_blue-m.cpt
set col_scale = "-T-$scale/$scale/0.01"

set resol = ""

# set for tight bounding box
gmt set PS_MEDIA letter
gmt set FORMAT_GEO_OUT Ddd.x

gmt makecpt -C$in_cpt $col_scale -Z -D > $cpt
gmt psbasemap $LLproj $LLreg  -Bxy0.2ag --MAP_FRAME_TYPE=inside -K -V -P > $out

# fault slip rectangles colored by their -Z values
gmt psxy $file.dat $LLproj $LLreg -C$cpt -L -W0.01c -O -K >> $out

# better to have no fill for GeoTIFF
gmt pscoast $LLproj $LLreg -N1/0/0/0 -I1/0/0/255 -Wthick,black -Df -O -K >> $out


# GPS stations
gmt psxy $GPSdir/$GPSstations $LLproj $LLreg -fi2x,1y -h1 -St0.4 -Wthick,yellow -Gyellow -V -O -K >> $out
gmt psxy $GPSdir/$GPSstations $LLproj $LLreg -fi2x,1y -h1 -St0.3 -Wthin,black -Gblack -V -O -K >> $out

gmt psvelo GPS_data_transformed_Acapulco.psvelo $LLproj $LLreg -Se$GPSscale/0.95 -A0.07/0.20/0.15 -G0/0/0  -P -K -O -V >> $out
gmt psvelo GPS_synthetic_Acapulco.psvelo $LLproj $LLreg -Se$GPSscale/0.95+f0p -A0.07/0.20/0.15 -Ggreen  -P -K -O -V >> $out

# fault rectangle on top--model fault already in UTM km
gmt psxy $vecDir/M7_1_SSN_MX_hypocenter_early_verificado.llz $LLproj $LLreg -Sa0.45 -W2/0/0/0 -Ggreen -O -K >> $out

gmt psscale -Dx0.2c/1.5c+w2.5c/0.4+e -C$cpt -Bxaf+1"Slip (m)" -O -V  >> $out

gv $out

gmt psconvert -A -P -Tt -W+g $out
```

pygmt/tests/test_figure.py
seisman: Does the frame make trouble here? The **psconvert -W** (https://docs.generic-mapping-tools.org/latest/psconvert.html) says:

> Be aware, however, that different results are obtained depending on the image contents and if the [-B](https://docs.generic-mapping-tools.org/latest/psconvert.html#b) option has been used or not. The trouble with the [-B](https://docs.generic-mapping-tools.org/latest/psconvert.html#b) option is that it creates a frame and very likely its annotations. That introduces pixels outside the map data extent, and therefore the map extents estimation will be wrong. To avoid this problem use --MAP_FRAME_TYPE=inside option which plots all annotations and ticks inside the image and therefore does not compromise the coordinate computations. Pay attention also to the cases when the plot has any of the sides with whites only because than the algorithm will fail miserably as those whites will be eaten by the Ghostscript. In that case you really must use [-B](https://docs.generic-mapping-tools.org/latest/psconvert.html#b) or use a slightly off-white color.

weiji14: Assuming that the `-B` means `-B` from `grdimage/basemap/etc` and not `psconvert`... I tried:

```python
with pygmt.config(MAP_FRAME_TYPE="inside"):
        fig.basemap(region=[0, 10, 0, 10], projection="M10c", frame="afg")
```

And it does look a bit better on the latitude (y-axis) side. The blue line and black lines align well on the latitude grid.

image

But on the longitude (x-axis) side, there still seems to be some offset (~900m) with the blue line not aligning with the black line:

image

seisman: Could you please try to produce a figure using `Figure.coast` or `Figure.grdimage` without the `frame` parameter? The frame still has a thickness, which may affect the boundingbox or something.

weiji14: Still having an offset without the frame:

```python
fig = Figure()
with pygmt.config(MAP_FRAME_TYPE="inside"):
    fig.coast(shorelines=True, region=[0, 10, 0, 10], projection="M10c")
# Save as GeoTIFF
geofname = Path("test_figure_savefig_geotiff.tiff")
fig.savefig(geofname)
```

Top-left corner

image

Bottom-right corner

image

I measured the size of one of the gray pixels and it was ~900+m, same as the offset. So it might be a case of gridline vs pixel registration (xref #487)? Though I'd think the offset would be half a pixel instead (~450m).

seisman: > So it might be a case of gridline vs pixel registration (xref #487)?

But this is not a grid, so can't be the gridline/pixel registration issue. Anyway, if something doesn't make sense, then it must be an upstream bug.

EJFielding: I found that the GeoTIFF images I was making with GMT classic looked better with the inside frame. I don't remember if having the frame outside made a difference in the geolocation.

I am not sure what the conventions are for "world" files, but I know that GDAL has the reference location for the image as the top-left corner of the top-left pixel, unless you change the registration. This is the same as "pixel" registration in GMT.

EJFielding: The other thing to realize for checking image geolocation is that the coastline database used in GMT is something like 40 years old, and not very accurate. I would not rely on the GMT coastline to check the geolocations.

seisman: @weiji14 When you have time, it would be better to submit an upstream issue for more in-depth discussions about the possible misalignments. I believe we can do nothing on the PyGMT side.

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,60 @@ def test_figure_savefig_exists():
...
+            npt.assert_allclose(
+                actual=xds.rio.bounds(),
+                desired=(
+                    -661136.0621116752,
+                    -54631.82709660966,
+                    592385.4459661598,
+                    1129371.7360144067,
+                ),
+            )
+            assert xds.rio.shape == (1257, 1331)
```
weiji14, line 133:
```suggestion
            assert xds.rio.shape == (1257, 1331)
            assert xds.rio.transform() == Affine(
                a=941.789262267344,
                b=0.0,
                c=-661136.0621116752,
                d=0.0,
                e=-941.92805338983,
                f=1129371.7360144067,
            )
```

pygmt/tests/test_figure.py
```diff
@@ -92,6 +92,60 @@ def test_figure_savefig_exists():
...
+        # TIFF
+        with pytest.warns(expected_warning=NotGeoreferencedWarning) as record:
+            with rioxarray.open_rasterio(fname) as xds:
+                assert xds.rio.crs is None
+                npt.assert_allclose(
+                    actual=xds.rio.bounds(), desired=(0.0, 0.0, 1331.0, 1257.0)
+                )
+                assert xds.rio.shape == (1257, 1331)
```
weiji14, line 141:
```suggestion
                assert xds.rio.shape == (1257, 1331)
                assert xds.rio.transform() == Affine(
                    a=1.0, b=0.0, c=0.0, d=0.0, e=1.0, f=0.0
                )
```

weiji14 (approved):
An extra check for the Affine transformation, otherwise ok.

EJFielding:
Thanks to @seisman and @weiji14 for adding this GeoTIFF function to PyGMT! I am now using it all the time, and it is working great.

Issue #2712, opened 2023-10-03
pylint 3.0.0 reports cyclic-import errors


seisman:
First reported in https://github.com/GenericMappingTools/pygmt/pull/2709#issuecomment-1743777415.

With the latest pylint 3.0.0 (https://github.com/pylint-dev/pylint/releases/tag/v3.0.0), the style checks workflow now reports the following errors: 
```
Run make lint
pylint pygmt doc/conf.py
************* Module conf
doc/conf.py:1:0: R0401: Cyclic import (pygmt -> pygmt.src -> pygmt.src.timestamp) (cyclic-import)
doc/conf.py:1:0: R0401: Cyclic import (pygmt -> pygmt.src -> pygmt.src.ternary) (cyclic-import)
doc/conf.py:1:0: R0401: Cyclic import (pygmt -> pygmt.figure -> pygmt.src.timestamp) (cyclic-import)
doc/conf.py:1:0: R0401: Cyclic import (pygmt -> pygmt.figure -> pygmt.src.ternary) (cyclic-import)

-----------------------------------
Your code has been rated at 9.99/10

make: *** [Makefile:79: lint] Error 8
Error: Process completed with exit code 2.
```
The `cyclic-import` errors are caused by the `__gmt_version__` variable added in #2366.

PR #2713, opened 2023-10-03
Move variable __gmt_version__ to pygmt.clib to avoid cyclic-import errors

seisman:

Defining `__gmt_version__` in `pygmt/__init__.py` causes cyclic-import errors.
This PR fixes the issue by moving `__gmt_version__` from `pygmt/__init__.py`
to `pygmt/clib/__init__.py`.

Closes https://github.com/GenericMappingTools/pygmt/issues/2712.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -21,13 +21,9 @@ import atexit as _atexit
 import sys
 from importlib.metadata import version
 
-from pygmt import clib
-
 # Get semantic version through setuptools-scm
 __version__ = f'v{version("pygmt")}'  # e.g. v0.1.2.dev3+g0ab3cd78
 __commit__ = __version__.split("+g")[-1] if "+g" in __version__ else ""  # 0ab3cd78
-with clib.Session() as lib:
-    __gmt_version__ = lib.info["version"]
 
 # Import modules to make the high-level GMT Python API
 from pygmt import datasets
--- a/pygmt/clib/__init__.py
+++ b/pygmt/clib/__init__.py
@@ -6,3 +6,6 @@ Pythonic interface. Access to the C library is done through ctypes.
 """
 
 from pygmt.clib.session import Session
+
+with Session() as lib:
+    __gmt_version__ = lib.info["version"]
--- a/pygmt/src/ternary.py
+++ b/pygmt/src/ternary.py
@@ -3,8 +3,7 @@ ternary - Plot data on ternary diagrams.
 """
 import pandas as pd
 from packaging.version import Version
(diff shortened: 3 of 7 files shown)


pygmt/__init__.py
```diff
@@ -21,13 +21,9 @@
 import sys
 from importlib.metadata import version
 
-from pygmt import clib
-
 # Get semantic version through setuptools-scm
 __version__ = f'v{version("pygmt")}'  # e.g. v0.1.2.dev3+g0ab3cd78
 __commit__ = __version__.split("+g")[-1] if "+g" in __version__ else ""  # 0ab3cd78
-with clib.Session() as lib:
-    __gmt_version__ = lib.info["version"]
```
weiji14, line 30: Looking at how `rasterio` handles `__gdal_version__` at https://github.com/rasterio/rasterio/blob/6bf045cc5e02c9f28e78d48b97b308490207d27d/rasterio/__init__.py#L84-L86, they do:

```python
from rasterio._version import gdal_version

__gdal_version__ = gdal_version()
```

They have a [`_version.pyx`](https://github.com/rasterio/rasterio/blob/6bf045cc5e02c9f28e78d48b97b308490207d27d/rasterio/_version.pyx) and a [`_version.pxd`](https://github.com/rasterio/rasterio/blob/6bf045cc5e02c9f28e78d48b97b308490207d27d/rasterio/_version.pxd) file that seems to be reading from the `gdal.h` header file. Do you think we could implement something similar for PyGMT/GMT by looking at `gmt_version.h`?

seisman: The way we get the GMT version string doesn't matter. What matters is that the `__gmt_version__` variable is defined in `pygmt/__init__.py` and is internally used in other files. 

Take `pygmt/src/timestamp.py` as an example. It imports the `__gmt_version__` variable from `pygmt/__init__.py`. However, `pygmt/__init__.py` imports `Figure` from `pygmt/figure.py` and `pygmt/figure.py` imports `timestamp` from `pygmt/src/timestamp.py`. Thus, there are always cyclic import issues if we define `__gmt_version__` in `pygmt/__init__.py`. 

The only solution I can see is defining `__gmt_version__` in another file, like `pygmt/clib/__init__.py` in this PR, or a new file like `pygmt/_version.py`.

weiji14: Ok, let's keep things in `pygmt/clib/__init__.py` for now since it's all internal variables, to get the Style Checks passing. We can look into moving things later to `_version.py` if needed.

weiji14 (approved):

Issue #2361, opened 2023-02-12
Limitations of the kwargs_to_strings decorator not handling positional arguments


seisman:
Here are some limitations of the `kwargs_to_strings` decorator:

1. Doesn't work for the default values of parameters:

https://github.com/GenericMappingTools/pygmt/pull/2208#discussion_r1038274212

> Please note that we can't use decorator `@kwargs_to_strings(offset="sequence")` here, because the decorator can only deal with keyword arguments and doesn't work for the default values. I.e., it doesn't work for `Figure.timestamp()` but works for `Figure.timestamp(offset=["-54p", "-54p"])`, although offset defaults to `["-54p", "-54p"]`.

```python
@kwargs_to_strings(offset="sequence")
def timestamp(
    self,
    text=None,
    label=None,
    justification="BL",
    offset=("-54p", "-54p"),
    font="Helvetica",
    timefmt="%Y %b %d %H:%M:%S",
):
```

2. Doesn't work if the parameter name is not given
```python
>>> from pygmt import which
>>> # doesn't work
>>> which(["@hotspots.txt", "@earth_relief_01d_g"])
gmtwhich [ERROR]: File ['@hotspots.txt', not found!
gmtwhich [ERROR]: File '@earth_relief_01d_g'] not found!
>>> # works
>>> which(fname=["@hotspots.txt", "@earth_relief_01d_g"])
```

weiji14:
This issue would also apply to `fig.text`'s `textfiles` parameter that uses `sequence_space` added in https://github.com/GenericMappingTools/pygmt/pull/325.

https://github.com/GenericMappingTools/pygmt/blob/c9d6147bcf383903ffc75fb94f9787b6b3a1d231/pygmt/src/text.py#L41-L50

Should we remove `sequence_space` entirely since it is only used in `pygmt.which` and `fig.text`? And use the `if is_nonstr_iter(fname): fname = " ".join(fname)` patch in #2726?

seisman:
> And use the `if is_nonstr_iter(fname): fname = " ".join(fname)` patch in #2726?

No, it won't work for `Figure.text`, because `textfiles` can be a 2-D array, e.g., `textfiles=[[1, 2, "A"], [3, 4, "B"]]`.

weiji14:
> > And use the `if is_nonstr_iter(fname): fname = " ".join(fname)` patch in #2726?
> 
[...]

PR #2726, opened 2023-10-09
pygmt.which: Fix the bug when passing multiple files

seisman:

Issue first reported in https://github.com/GenericMappingTools/pygmt/issues/2361#issue-1581189232. See that issue for context.

To reproduce the issue:
```python
>>> from pygmt import which
>>> # doesn't work
>>> which(["@hotspots.txt", "@earth_relief_01d_g"])
gmtwhich [ERROR]: File ['@hotspots.txt', not found!
gmtwhich [ERROR]: File '@earth_relief_01d_g'] not found!
>>> # works
>>> which(fname=["@hotspots.txt", "@earth_relief_01d_g"])
```

Address #2361.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/which.py
+++ b/pygmt/src/which.py
@@ -6,14 +6,13 @@ from pygmt.helpers import (
     GMTTempFile,
     build_arg_string,
     fmt_docstring,
-    kwargs_to_strings,
+    is_nonstr_iter,
     use_alias,
 )
 
 
 @fmt_docstring
 @use_alias(G="download", V="verbose")
-@kwargs_to_strings(fname="sequence_space")
 def which(fname, **kwargs):
     r"""
     Find the full path to specified files.
@@ -63,6 +62,9 @@ def which(fname, **kwargs):
     FileNotFoundError
         If the file is not found.
     """
+    if is_nonstr_iter(fname):  # Got a list of files
+        fname = " ".join(fname)
+
     with GMTTempFile() as tmpfile:
         with Session() as lib:
             lib.call_module(
--- a/pygmt/tests/test_which.py
+++ b/pygmt/tests/test_which.py
@@ -1,7 +1,7 @@
 """
 Test pygmt.which.
 """
-import os
+from pathlib import Path
 
 import pytest
 from pygmt import which
@@ -14,8 +14,8 @@ def test_which():
     """
     for fname in ["tut_quakes.ngdc", "tut_bathy.nc"]:
         cached_file = which(f"@{fname}", download="c")
-        assert os.path.exists(cached_file)
-        assert os.path.basename(cached_file) == fname
+        assert Path(cached_file).exists()
+        assert Path(cached_file).name == fname
 
 
 def test_which_multiple():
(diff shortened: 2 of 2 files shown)


pygmt/tests/test_which.py
```diff
@@ -14,19 +14,19 @@ def test_which():
...
+        assert Path(cached_file).name == fname
 
 
 def test_which_multiple():
     """
     Make sure `which` returns file paths for multiple @files correctly.
     """
     filenames = ["ridge.txt", "tut_ship.xyz"]
-    cached_files = which(fname=[f"@{fname}" for fname in filenames], download="c")
+    cached_files = which([f"@{fname}" for fname in filenames], download="c")
```
seisman, line 26: Now we can pass a list of files without having to use `fname=`

pygmt/tests/test_which.py
```diff
@@ -14,19 +14,19 @@ def test_which():
     """
     for fname in ["tut_quakes.ngdc", "tut_bathy.nc"]:
         cached_file = which(f"@{fname}", download="c")
```
weiji14, line 16:
```suggestion
        cached_file = which(fname=f"@{fname}", download="c")
```

Ensure that the `fname` keyword parameter still works.

pygmt/src/which.py
```diff
@@ -6,14 +6,13 @@
...
-    kwargs_to_strings,
+    is_nonstr_iter,
     use_alias,
 )
 
 
 @fmt_docstring
 @use_alias(G="download", V="verbose")
-@kwargs_to_strings(fname="sequence_space")
 def which(fname, **kwargs):
```
weiji14, line 16: Reading up on https://peps.python.org/pep-0570/#positional-or-keyword-arguments and https://stackoverflow.com/questions/9450656/positional-argument-vs-keyword-argument (see also Max's previous comment at https://github.com/GenericMappingTools/pygmt/pull/731#discussion_r825477816), should we explicitly mark `fname` as positional_or_keyword like this (unsure if syntax is ok)?

```suggestion
def which(/, fname, *, **kwargs):
```

Or just stick with the current one (which works also I think).

seisman: Better to open a separate issue so that we can have more discussions about the behavior and also have consistent function definitions in the whole project.

weiji14: On second thought,, maybe we should enforce using keyword-only arguments like so:

```suggestion
def which(*, fname, **kwargs):
```

Then the `gmtwhich [ERROR]` you mentioned at https://github.com/GenericMappingTools/pygmt/issues/2361#issue-1581189232 would turn into a slightly more useful error:

```
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 cached_files = which([f"@{fname}" for fname in filenames], download="c")

File ~/pygmt/pygmt/helpers/decorators.py:598, in use_alias.<locals>.alias_decorator.<locals>.new_module(*args, **kwargs)
    591     msg = (
    592         "Parameters 'Y' and 'yshift' are deprecated since v0.8.0. "
    593         "and will be removed in v0.12.0. "
    594         "Use Figure.shift_origin(yshift=...) instead."
    595     )
    596     warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
--> 598 return module_func(*args, **kwargs)

TypeError: which() takes 0 positional arguments but 1 was given
```
Well, actually not that useful (it doesn't say that the `fname` keyword should be used. But something to consider.

seisman: Instead, I prefer to `def which(fname, *, **kwargs)`. It allows `which("file.txt")` which is much simpler than `which(fname="file.txt")`.

weiji14: Ok, let's go with that. I'll approve this PR, but you can apply the change afterwards.

weiji14 (approved):

pygmt/src/which.py
seisman:
```
def which(fname, *, **kwargs)
```
gives a syntax error:
```
'named arguments must follow bare *
```
See https://stackoverflow.com/questions/14301967/what-does-a-bare-asterisk-do-in-a-parameter-list-what-are-keyword-only-parame.

I'll leave the function definition as it is now.

PR #2729, opened 2023-10-09
Wrap GMT's standard data type GMT_DATASET for table inputs

seisman:

Similar to https://github.com/GenericMappingTools/pygmt/pull/2398, this PR contains some experimental codes to support GMT's standard data type `GMT_DATASET` for table inputs.

Here are some technical background. In GMT, table inputs are stored in a `GMT_DATASET` object. A `GMT_DATASET` object can contain several table files (stored in `GMT_DATATABLE`) and each table can contain several segments (stored in `GMT_DATASEGMENT`), and each data segment have *M* rows and *N* columns.

**References**: https://docs.generic-mapping-tools.org/latest/devdocs/api.html#struct-dataset

The ultimate goals are:

- [ ] Avoid using `GMTTempfile` for modules that output a table [tracked in #2730]
- [ ] Have a Pythonic representation for the `GMT_DATASET` object and pass it as input to some modules

---

Currently, I'm doing too much work in this PR, for example:

1. ~~Wrapping the GMT C API `GMT_Read_Virtualfile`~~ Merged to main in #2993
2. ~~Provide a new session method `virtualfile_to_data` to allow GMT modules write output into a virtual file.~~ Merged to main in #3057
3. Wrapping the GMT_DATASET data type and related `GMT_DATATABLE` and `GMT_DATASEGMENT`
5. Refactor many modules (e.g., `grd2xyz`, `grdtrack`) to show how this PR will change the way we handle table-like outputs.

Thus, we need to split this PR into a few smaller PRs if this PR is proven to be working.


pygmt/src/grd2xyz.py
```diff
@@ -172,25 +167,37 @@ def grd2xyz(grid, output_type="pandas", outfile=None, **kwargs):
...
-                )
-
-        # Read temporary csv output to a pandas table
-        if outfile == tmpfile.name:  # if user did not set outfile, return pd.DataFrame
-            result = pd.read_csv(
-                tmpfile.name, sep="\t", names=dataframe_header, comment=">"
+    with Session() as lib:
+        with lib.virtualfile_from_data(
+            check_kind="raster", data=grid
+        ) as invfile, lib.virtualfile_to_gmtdataset() as outvfile:
```
seisman, line 173: After wrapping `GMT_DATASET` and providing some data-conversion functions, now it's possible to get rid of temporary files from GMT modules that write table files. `grd2xyz` is used as an example.

`grd2xyz` (and other similar functions) supports three different output types: "file", "numpy" and "pandas". The simplest strategy is

1. Always write the table files into a `GMT_DATASET` object
2. If `output_type="file"`, then we can call the special `write` module to write it into a file
3. Otherwise, convert `GMT_DATASET` to a list of vectors and then convert vectors to numpy or pandas.

This strategy is implemented as option 1 in the codes. It's simple, so it's possible to wrap the if-tests (lines 180-186) in a separate function and reuse it in many places.

The downside is, that for `output_type="file"`, we need to one more GMT C API call to write the `GMT_DATASET` object to a file.

So, the option 2 is also implemented in the codes. If `output_type="file"`, we write the output directly into a file. It's more efficient, but the codes is less clear than those of option 1.

pygmt/clib/session.py
```diff
@@ -1680,3 +1681,56 @@ def extract_region(self):
...
+        vectors = []
+        for itble in range(ds.n_tables):
+            dtbl = ds.table[itble].contents
+            for iseg in range(dtbl.n_segments):
+                dseg = dtbl.segment[iseg].contents
+                for icol in range(dseg.n_columns):
+                    vectors.append(
+                        np.ctypeslib.as_array(dseg.data[icol], shape=(dseg.n_rows,))
+                    )
+        return vectors
```
seisman, line 1736: Not sure what's the Pythonic way to store multiple tables or segments in NumPy arrays.
For example, if I have a GMT multi-segment file
```
> 
1 2
3 4
> 
5 6
7 8
```
Currently, the code incorrectly returns 
```
[[1, 3], [2, 4], [5, 6], [7, 8]]
```
At least it should return
```
[[1, 3, 5, 7], [2, 4, 6, 8]]
```
but it's still NOT a multi-segment "array".

pygmt/src/grdtrack.py
```diff
@@ -292,29 +289,30 @@ def grdtrack(grid, points=None, newcolname=None, outfile=None, **kwargs):
...
+        ) as grdfile, lib.virtualfile_from_data(
+            check_kind="vector", data=points, required_data=False
+        ) as csvfile, lib.virtualfile_to_gmtdataset() as outvfile:
+            kwargs["G"] = grdfile
+            lib.call_module(
+                module="grdtrack",
+                args=build_arg_string(kwargs, infile=csvfile, outfile=outvfile),
+            )
+        if outfile is not None:
+            # if output_type == "file":
```
seisman, line 304: For grdtrack and other similar modules, we need to add the `output_type` parameter.

pygmt/src/grd2xyz.py
seisman: Problem solved in commit f5a9d447d1b003f93207949ecf940b0ab8bceef4.

pygmt/clib/session.py
```diff
@@ -1678,3 +1680,74 @@ def extract_region(self):
         if status != 0:
             raise GMTCLibError("Failed to extract region from current figure.")
         return wesn
+
+    def read_virtualfile(self, vfname, kind=None):
```
weiji14, line 1684: Probably need to add some unit tests to `test_clib...` for this. Since that `test_clib.py` file is getting long, should we split the virtualfile related tests into a `test_clib_virtualfile.py`? I was thinking of doing this before continuing work on #2774.

weiji14 (commented):
Will try to give this a proper review this weekend :crossed_fingers:, but just one comment for now on where to put the unit tests.

pygmt/clib/session.py
seisman: > Probably need to add some unit tests to `test_clib...` for this. Since that `test_clib.py` file is getting long, should we split the virtualfile related tests into a `test_clib_virtualfile.py`? I was thinking of doing this before continuing work on #2774.

Sounds good to me.

seisman: I'll open a PR for it.

weiji14: Done at #2784.

pygmt/datatypes.py
```diff
@@ -0,0 +1,221 @@
+"""
+Wrappers for GMT data types.
+"""
+import ctypes as ctp
+
+import numpy as np
+
+
+class GMT_DATASET(ctp.Structure):  # pylint: disable=invalid-name,too-few-public-methods
```
seisman, line 9: The `GMT_DATASET` object is difficult to use (see the docstrings and the `to_vectors` method) and should not be exposed to our end users. So maybe we should rename it to something like `_GMT_DATASET`?

On the other hand, we may want to provide a Pythonic data structure to represent GMT's `GMT_DATASET` object, like what GMT.jl does (https://www.generic-mapping-tools.org/GMT.jl/dev/types/#Dataset-type). The new Pythonic data structure is necessary to let PyGMT support multi-segment datasets but it can be done in future PRs after more discussions.

pygmt/datatypes.py
```diff
@@ -0,0 +1,221 @@
...
+        # deal with trailing text column
+        textvector = []
+        for itbl in range(self.n_tables):
+            dtbl = self.table[itbl].contents
+            for iseg in range(dtbl.n_segments):
+                dseg = dtbl.segment[iseg].contents
+                if dseg.text:
+                    textvector.extend(dseg.text[: dseg.n_rows])
+        if textvector:
+            vectors.append(np.char.decode(textvector))
```
seisman, line 219: Currently, all trailing columns are stored in a single vector.

pygmt/helpers/utils.py
```diff
@@ -550,3 +551,37 @@ def args_in_kwargs(args, kwargs):
...
+    Parameters
+    ----------
+    session : :class:`pygmt.clib.Session`
+        The current session.
+    output_type : str
+        The output type. Can be ``"pandas"``, ``"numpy"``, or ``"file"``.
+    vfile : str
+        The virtual file name.
+    colnames : list of str
+        The column names for the :class:`pandas.DataFrame` output.
```
seisman, line 569: Not sure if `colnames` is a good name. In the module wrappers, we're inconsistently using `column_names` or `dataframe_header`.

pygmt/helpers/utils.py
```diff
@@ -550,3 +551,37 @@ def args_in_kwargs(args, kwargs):
...
+        The output table. If ``output_type`` is ``"file"``, returns ``None``.
+    """
+    if output_type == "file":  # Already written to file, so return None
+        return None
+    # Read the virtual file as a GMT dataset and convert to vectors
+    vectors = session.read_virtualfile(vfile, kind="dataset").contents.to_vectors()
+    # pandas.DataFrame output
+    if colnames is None:
+        colnames = pd.RangeIndex(0, len(vectors))
+    result = pd.DataFrame.from_dict(dict(zip(colnames, vectors)))
```
seisman, line 583: `vectors` is a list of vectors and each vector represents a column in the data. I've tried different solutions to create a pd.DataFrame object, but most are not ideal.

Here is an example:
```
>>> import numpy as np
>>> import pandas as pd
>>> vectors = [np.array([1, 2, 3]), [4.0, 5.0, 6.0], [7.0, 8.0, 9.0], ["AA", "BB", "CC"]]
>>> colnames = ["col1", "col2", "col3", "col4"]
```

```
# Each vector is taken as a row, not a column
>>> df = pd.DataFrame(vectors)
>>> df
     0    1    2
0    1    2    3
1  4.0  5.0  6.0
2  7.0  8.0  9.0
3   AA   BB   CC

# The dtype information is lost 
>>> df = pd.DataFrame(np.array(vectors).T)
>>> df
   0    1    2   3
0  1  4.0  7.0  AA
1  2  5.0  8.0  BB
2  3  6.0  9.0  CC
>>> df.dtypes
0    object
1    object
2    object
3    object
dtype: object

# It's weird to set column names via `index` and again dtype is lost
>>> df = pd.DataFrame(vectors, index=colnames).T
>>> df
  col1 col2 col3 col4
0    1  4.0  7.0   AA
1    2  5.0  8.0   BB
2    3  6.0  9.0   CC
>>> df.dtypes
col1    object
col2    object
col3    object
col4    object
dtype: object

# Current best solution
>>> df = pd.DataFrame.from_dict(dict(zip(colnames, vectors)))
>>> df
   col1  col2  col3 col4
0     1   4.0   7.0   AA
1     2   5.0   8.0   BB
2     3   6.0   9.0   CC
>>> df.dtypes
col1      int64
col2    float64
col3    float64
col4     object
dtype: object
```

pygmt/src/grdhisteq.py
```diff
@@ -63,7 +63,7 @@ class grdhisteq:  # pylint: disable=invalid-name
         h="header",
     )
     @kwargs_to_strings(R="sequence")
-    def _grdhisteq(grid, output_type, **kwargs):
+    def _grdhisteq(caller, grid, output_type, **kwargs):
```
seisman, line 66: This one is a little complicated because it can output either a table or a grid.

pygmt/src/grdhisteq.py
```diff
@@ -108,34 +108,36 @@ def _grdhisteq(grid, output_type, **kwargs):
...
+                output_type=output_type,
+                vfile=vouttbl,
+                colnames=["start", "stop", "bin_id"],
+            )
+            if output_type == "pandas":
+                result = result.astype(
+                    {"start": np.float32, "stop": np.float32, "bin_id": np.uint32}
+                )
+                return result.set_index("bin_id")
+            return result
```
seisman, line 140: For this module, we also set the dtypes for the three columns. Maybe this should also be done in the `return_table` function, i.e., change the `return_table` function definition to
```
return_table(session, output_type, vfile, colnames, dtypes=None, index_col=None)
```
then we can call it this way:
```
result = return_table(
    session=lib,
    output_type=output_type,
    vfile=vouttbl,
    colnames=["start", "stop", "bin_id"],
    dtype=[np.float32, np.float32, np.uint32], # dtype=None means the default float64
    index_col="bin_id",  # similar to pd.read_csv's `index_col` parameter
)
```

pygmt/src/grdtrack.py
```diff
@@ -43,7 +45,9 @@
     w="wrap",
 )
 @kwargs_to_strings(R="sequence", S="sequence", i="sequence_comma", o="sequence_comma")
-def grdtrack(grid, points=None, newcolname=None, outfile=None, **kwargs):
+def grdtrack(
+    grid, points=None, output_type="pandas", outfile=None, newcolname=None, **kwargs
```
seisman, line 49: Why don't `newcolname` has a default value like `z`?

pygmt/tests/test_triangulate.py
```diff
@@ -43,7 +43,8 @@ def fixture_expected_dataframe():
             [4, 6, 1],
             [3, 4, 2],
             [9, 3, 8],
-        ]
+        ],
+        dtype=float,
```
seisman, line 47: When writing to a file, GMT uses a format specifier like `%g` to write float-pointing data. For example, the output of the GMT CLI is:
```
> echo 1.0 1.0 | gmt select -R0/10/0/10
1   1
```
When we read this file via `pd.read_csv`, the columns will have a `int64` type. However, all data are stored as double-precision numbers in the `GMT_DATASET` object, thus column dtype will be `float64` by default. This behavior is backward incompatible, but I think the new behavior is better.

weiji14: Are there different format specifiers besides `%g`? Not too sure about storing numeric columns as float64 by default, I'd prefer a way to automatically return the lightest dtype if possible (something close to pandas' [`convert_dtypes`](https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.DataFrame.convert_dtypes.html)).

pygmt/datatypes.py
weiji14: > The `GMT_DATASET` object is difficult to use (see the docstrings and the `to_vectors` method) and should not be exposed to our end users. So maybe we should rename it to something like `_GMT_DATASET`?

Instead of naming the class `_GMT_DATASET`, you could also name the file `_datatypes.py` to mark it private.

> On the other hand, we may want to provide a Pythonic data structure to represent GMT's `GMT_DATASET` object, like what GMT.jl does (https://www.generic-mapping-tools.org/GMT.jl/dev/types/#Dataset-type). The new Pythonic data structure is necessary to let PyGMT support multi-segment datasets but it can be done in future PRs after more discussions.

Let's discuss this in a separate issue. I'd much prefer that we don't reinvent another `pandas.DataFrame` or `geopandas.GeoDataFrame` data structure for Python users (but having an internal `GMT_DATASET` class is ok). Also we should be looking at upcoming standards like the [Python dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html) and [GeoArrow](https://github.com/geoarrow/geoarrow) if we want a data structure that is reusable across the Python ecosystem.

weiji14 (commented):
Thanks @seisman, I know you've spent a lot of time on this already, so apologies for the late review! Comments below are mainly focusing on the `GMT_DATASET.to_vectors` and `pygmt.utils.return_table` function for now, and responding to some of your inline comments.

pygmt/helpers/utils.py
seisman: OK, I've updated it to `column_names` in 3d5147d422763ef24b3a9b415933e00d535f8e92.

pygmt/datatypes.py
seisman: > > The `GMT_DATASET` object is difficult to use (see the docstrings and the `to_vectors` method) and should not be exposed to our end users. So maybe we should rename it to something like `_GMT_DATASET`?
> 
> Instead of naming the class `_GMT_DATASET`, you could also name the file `_datatypes.py` to mark it private.

`_datatypes.py` sounds good. Will do it later.

> > On the other hand, we may want to provide a Pythonic data structure to represent GMT's `GMT_DATASET` object, like what GMT.jl does (https://www.generic-mapping-tools.org/GMT.jl/dev/types/#Dataset-type). The new Pythonic data structure is necessary to let PyGMT support multi-segment datasets but it can be done in future PRs after more discussions.
> 
> Let's discuss this in a separate issue. I'd much prefer that we don't reinvent another `pandas.DataFrame` or `geopandas.GeoDataFrame` data structure for Python users (but having an internal `GMT_DATASET` class is ok). Also we should be looking at upcoming standards like the [Python dataframe interchange protocol](https://data-apis.org/dataframe-protocol/latest/index.html) and [GeoArrow](https://github.com/geoarrow/geoarrow) if we want a data structure that is reusable across the Python ecosystem.

Yes, definitely needs more discussions to support multi-segment data files.

pygmt/datatypes.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+
+    class GMT_DATATABLE(ctp.Structure):
+        # pylint: disable=invalid-name,too-few-public-methods
+        """
+        GMT datatable structure for holding a single table with multiple
+        segments.
+        """
+
+        class GMT_DATASEGMENT(ctp.Structure):
```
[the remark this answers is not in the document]
seisman: We don't have to. Actually, I initially defined the three classes separately and decided to use nested classes in https://github.com/GenericMappingTools/pygmt/pull/2729/commits/4707b872451f0848ee90c183ac6c5346f10c415a, although I don't have a strong reason to do that.

pygmt/tests/test_triangulate.py
seisman: > Are there different format specifiers besides `%g`? 

The format of these outputs is controlled by GMT's `FROAMT_FLOAT_OUT` parameter and its default value is `%.12g`.

> Not too sure about storing numeric columns as float64 by default, I'd prefer a way to automatically return the lightest dtype if possible (something close to pandas' [`convert_dtypes`](https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.DataFrame.convert_dtypes.html)).

Sounds good to me.

pygmt/helpers/utils.py
seisman: > If we want to turn that string column's object dtype to a proper `string` dtype, we can use pandas's [`convert_dtypes`](https://pandas.pydata.org/pandas-docs/version/2.1/reference/api/pandas.DataFrame.convert_dtypes.html):

It's likely `convert_dtypes` will cause more trouble for us:
```python
>>> df = pd.DataFrame.from_dict(dict(zip(colnames, vectors)))
>>> df["col1"].to_numpy()
array([1, 2, 3])
>>> df["col1"].to_numpy().dtype
dtype('int64')

>>> df = df.convert_dtypes(dtype_backend="numpy_nullable")
>>> df["col1"].to_numpy()
array([1, 2, 3], dtype=object)
```
As you can see, the `to_numpy` function now returns a `object` dtype instead of a `int64` type.

pygmt/datatypes.py
seisman: I believe most (if not all) GMT modules don't write text strings to output, unless the input file has trailing texts, for example,
```
echo 0 0 ABC DEF | gmt select -R-10/10/-10/10
0	0	ABC DEF
```

pygmt/helpers/utils.py
weiji14: We could also just convert string dtypes, and avoid converting integer and float into nullable dtypes like so:

```python
df = pd.DataFrame.from_dict(dict(zip(colnames, vectors)))
df = df.convert_dtypes(
    convert_integer=False,
    convert_floating=False,
    convert_string=True,
    dtype_backend="numpy_nullable",
)
df.dtypes
# col1             int64
# col2           float64
# col3           float64
# col4    string[python]
# dtype: object
df["col1"].to_numpy().dtype
# dtype('int64')
```

pygmt/datatypes.py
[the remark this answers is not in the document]
weiji14: Would anyone ever want to access one of the nested classes in `GMT_DATASET` -> `GMT_DATATABLE` -> `GMT_DATASEGMENT` directly? I don't have a strong opinion on the nesting, just that I haven't seen three levels of nested classes in Python before :sweat_smile:

pygmt/datatypes.py
weiji14: Ok, I thought so that there aren't many GMT modules writing trailing text. We can probably keep the trailing columns as a single column vector for now.

pygmt/datatypes.py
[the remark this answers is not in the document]
seisman: > Would anyone ever want to access one of the nested classes in `GMT_DATASET` -> `GMT_DATATABLE` -> `GMT_DATASEGMENT` directly?

I bet no one, even the PyGMT maintainers, will access these classes directly.

pygmt/helpers/utils.py
seisman: Done in b9454ce0c29e83b194650906704fa1aa26f398f8. 

[`convert_dtypes`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.convert_dtypes.html)'s `convert_floating` parameter is only for pandas >= 1.2.0. Do we need to add `pandas>=1.2.0` in `pyproject.toml`?

pygmt/src/grdtrack.py
seisman: OK, we can discuss if `newcolname` should have a default value when more module wrappers have the same parameter.

pygmt/datatypes.py
```diff
@@ -0,0 +1,221 @@
...
+            vectors.append(np.concatenate(colvector))
+
+        # deal with trailing text column
+        textvector = []
+        for itbl in range(self.n_tables):
+            dtbl = self.table[itbl].contents
+            for iseg in range(dtbl.n_segments):
+                dseg = dtbl.segment[iseg].contents
+                if dseg.text:
+                    textvector.extend(dseg.text[: dseg.n_rows])
```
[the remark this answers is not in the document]
seisman: Great suggestion! 

In commit 87424677f37158be872c42c344ed02c0aa68a261, I modified the old `to_vectors()` method to the new `to_dataframe()` method, which returns a pd.DataFrame object instead. The `pd.DataFrame` object is built using [`pd.concat`](https://pandas.pydata.org/docs/reference/api/pandas.concat.html), rather than `pd.DataFrame.from_dict(dict(zip(colnames, vectors)))`, which is more elegant.

pygmt/helpers/utils.py
weiji14: [Pandas 1.2.0 was released on 26 Dec 2020](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.2.0.html), or about 3 years ago. I tried to create a conda environment with pandas 1.1.x

```
mamba create -n temp 'pandas~=1.1.0' 'numpy==1.22' 'python==3.9'
```

and it seemed to be possible

```
Looking for: ['pandas~=1.1.0', 'numpy==1.23', 'python==3.9']

conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache
Transaction

  Prefix: /home/user/mambaforge/envs/temp

  Updating specs:

   - pandas ~=1.1.0
   - numpy==1.23
   - python==3.9


  Package                  Version  Build                Channel                    Size
──────────────────────────────────────────────────────────────────────────────────────────
  Install:
──────────────────────────────────────────────────────────────────────────────────────────

  + _libgcc_mutex              0.1  conda_forge          conda-forge/linux-64     Cached
  + _openmp_mutex              4.5  2_gnu                conda-forge/linux-64     Cached
  + ca-certificates      2023.7.22  hbcca054_0           conda-forge/linux-64     Cached
  + ld_impl_linux-64          2.40  h41732ed_0           conda-forge/linux-64     Cached
  + libblas                  3.9.0  19_linux64_openblas  conda-forge/linux-64     Cached
  + libcblas                 3.9.0  19_linux64_openblas  conda-forge/linux-64     Cached
  + libffi                     3.3  h58526e2_2           conda-forge/linux-64       53kB
  + libgcc-ng               13.2.0  h807b86a_3           conda-forge/linux-64     Cached
  + libgfortran-ng          13.2.0  h69a702a_3           conda-forge/linux-64     Cached
  + libgfortran5            13.2.0  ha4646dd_3           conda-forge/linux-64     Cached
  + libgomp                 13.2.0  h807b86a_3           conda-forge/linux-64     Cached
  + liblapack                3.9.0  19_linux64_openblas  conda-forge/linux-64     Cached
  + libopenblas             0.3.24  pthreads_h413a1c8_0  conda-forge/linux-64     Cached
  + libsqlite               3.44.0  h2797004_0           conda-forge/linux-64     Cached
  + libstdcxx-ng            13.2.0  h7e041cc_3           conda-forge/linux-64     Cached
  + libzlib                 1.2.13  hd590300_5           conda-forge/linux-64     Cached
  + ncurses                    6.4  h59595ed_2           conda-forge/linux-64     Cached
  + numpy                   1.23.0  py39hba7629e_0       conda-forge/linux-64        7MB
  + openssl                 1.1.1w  hd590300_0           conda-forge/linux-64        2MB
  + pandas                   1.1.5  py39hde0f152_0       conda-forge/linux-64       12MB
  + pip                     23.3.1  pyhd8ed1ab_0         conda-forge/noarch       Cached
  + python                   3.9.0  hffdb5ce_5_cpython   conda-forge/linux-64       30MB
  + python-dateutil          2.8.2  pyhd8ed1ab_0         conda-forge/noarch       Cached
  + python_abi                 3.9  4_cp39               conda-forge/linux-64        6kB
  + pytz              2023.3.post1  pyhd8ed1ab_0         conda-forge/noarch       Cached
  + readline                   8.2  h8228510_1           conda-forge/linux-64     Cached
  + setuptools              59.8.0  py39hf3d152e_1       conda-forge/linux-64        1MB
  + six                     1.16.0  pyh6c4a22f_0         conda-forge/noarch       Cached
  + sqlite                  3.44.0  h2c6b66d_0           conda-forge/linux-64     Cached
  + tk                      8.6.13  noxft_h4845f30_101   conda-forge/linux-64     Cached
  + tzdata                   2023c  h71feb2d_0           conda-forge/noarch       Cached
  + wheel                   0.41.3  pyhd8ed1ab_0         conda-forge/noarch       Cached
  + xz                       5.2.6  h166bdaf_0           conda-forge/linux-64     Cached
  + zlib                    1.2.13  hd590300_5           conda-forge/linux-64     Cached

  Summary:

  Install: 34 packages

  Total download: 53MB
```

So yeah, we might need to set a lower pin on pandas>=1.2.0 if using `convert_floating=False`. I don't think there's a way to make this backwards compatible with pandas 1.1.x since `convert_floating` is `True` by default.

seisman: > we might need to set a lower pin on pandas>=1.2.0 if using `convert_floating=False`.

Done in b7e082347de9a009ef314596a199207681c6b154.

seisman:
In commit a21b2df8d149ee4af172e8bd6d14dbebc017179a, I've reverted all changes that I made to the module wrappers so that we can focus on wrapping `GMT_DATASET` in this PR.

A backup of the reverted changes are available at this branch https://github.com/GenericMappingTools/pygmt/tree/backup/gmtdataset.

seisman:
> ## [CodSpeed Performance Report](https://codspeed.io/GenericMappingTools/pygmt/branches/datatypes/gmtdataset)
> ### Merging #2729 will **degrade performances by 90.44%**
> Comparing `datatypes/gmtdataset` ([76a09f0](https://github.com/GenericMappingTools/pygmt/commit/76a09f08d2132e767dcb4110f812527359a7e876)) with `main` ([ab39616](https://github.com/GenericMappingTools/pygmt/commit/ab396164c2c600471ad26352e4aeec28f29ae9f0))
> 
> ### Summary
> `⚡ 1` improvements `❌ 12` regressions `✅ 85` untouched benchmarks
> 
> > ⚠️ _Please fix the performance issues or [acknowledge them on CodSpeed](https://codspeed.io/GenericMappingTools/pygmt/branches/datatypes/gmtdataset)._
> 
> ### Benchmarks breakdown
> Benchmark	`main`	`datatypes/gmtdataset`	Change
> ⚡	`test_blockmean_input_xyz`	412.7 ms	366.8 ms	+12.54%
> ❌	`test_contour_matrix[DataFrame]`	90.8 ms	281.6 ms	-67.76%
> ❌	`test_contour_matrix[Dataset]`	103.9 ms	294.7 ms	-64.74%
> ❌	`test_contour_matrix[array]`	88.6 ms	279.3 ms	-68.3%
> ❌	`test_filter1d_format`	199.6 ms	518.8 ms	-61.52%
> ❌	`test_grd2xyz`	43.1 ms	229.9 ms	-81.28%
> ❌	`test_grdtrack_input_dataframe_and_dataarray`	43.9 ms	233.8 ms	-81.22%
> ❌	`test_grdvolume_no_outgrid`	41.7 ms	294.9 ms	-85.86%
> ❌	`test_project_input_matrix[DataFrame]`	41.6 ms	422.7 ms	-90.16%
> ❌	`test_project_input_matrix[Dataset]`	50.7 ms	431.8 ms	-88.25%
> ❌	`test_project_input_matrix[array]`	40.3 ms	421.3 ms	-90.44%
> ❌	`test_select_input_dataframe`	184.5 ms	371.8 ms	-50.37%
> ❌	`test_delaunay_triples_input_xyz`	169.3 ms	357.8 ms	-52.67%

This benchmark report looks scary, but it doesn't match my test in https://github.com/GenericMappingTools/pygmt/issues/2730#issuecomment-1863177081. I think it's likely because codspeed doesn't count the I/O time (via pd.read_csv).

weiji14 (approved):
Super big thank you for pushing this through @seisman, especially in cherry-picking many of the changes into separate PRs to make this easier to review! I'm still a little unsure about the triple-nested classes as mentioned at https://github.com/GenericMappingTools/pygmt/pull/2729#discussion_r1390150748, but we can always un-nest them later if needed (since they're marked as private classes).

Won't hold this PR up any longer, excited to have PyGMT not store intermediate CSV files in the next release :rocket:

PR #2734, opened 2023-10-10
CI: List installed Python packages for easier debugging

seisman:

Run `pip list` to show the installed packages so that we can easily find the versions of lining tools the CI is using.


weiji14 (approved):
Thanks! Actually wanted this for a while now :smile:

PR #2739, opened 2023-10-11
Figure.timestamp: Minor improvement to the docs

seisman:

**Preview**: https://pygmt-dev--2739.org.readthedocs.build/en/2739/api/generated/pygmt.Figure.timestamp.html

### DIFF
--- a/pygmt/src/timestamp.py
+++ b/pygmt/src/timestamp.py
@@ -62,13 +62,15 @@ def timestamp(
 
     Examples
     --------
-    >>> # Plot the GMT timestamp logo.
+    Plot the GMT timestamp logo.
+
     >>> import pygmt
     >>> fig = pygmt.Figure()
     >>> fig.timestamp()
     >>> fig.show()
 
-    >>> # Plot the GMT timestamp logo with a custom label.
+    Plot the GMT timestamp logo with a custom label.
+
     >>> fig = pygmt.Figure()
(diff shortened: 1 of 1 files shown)


pygmt/src/timestamp.py
```diff
@@ -62,13 +62,15 @@ def timestamp(
...
+
     >>> import pygmt
     >>> fig = pygmt.Figure()
     >>> fig.timestamp()
     >>> fig.show()
 
-    >>> # Plot the GMT timestamp logo with a custom label.
+    Plot the GMT timestamp logo with a custom label.
+
     >>> fig = pygmt.Figure()
```
yvonnefroehlich, line 74: I am wondering whether it is better to repeat `import pygmt` for the second example, as the two code examples are in two separate "cells"? When users copy and past the second example it will not work directly.

```suggestion
    Plot the GMT timestamp logo with a custom label.

    >>> import pygmt
    >>> fig = pygmt.Figure()
```

seisman: I prefer not to repeat `import pygmt`. `import pygmt` initialize a new GMT session so it's time-consuming. Even new users should know that they need to `import pygmt` so it's not a big issue.

yvonnefroehlich: I agree, it's only a smaller aspect and I am OK with leaving it as is.

yvonnefroehlich (approved):

PR #2740, opened 2023-10-11
Inline examples: Use "longitude" and "latitude" instead of "x" and "y" and add units in comments

yvonnefroehlich:

This PR aims to make comments of the inline examples more precise, by using `longitude` and `latitude` in stead of `x` and `y` for geographic grids and adding units.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -83,8 +83,8 @@ def grdclip(grid, **kwargs):
     Example
     -------
     >>> import pygmt
-    >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
-    >>> # and a y-range of 15 to 25
+    >>> # Load a grid of @earth_relief_30m data, with a longitude range of
+    >>> # 10° E to 30° E, and a latitude range of 15° N to 25° N
     >>> grid = pygmt.datasets.load_earth_relief(
     ...     resolution="30m", region=[10, 30, 15, 25]
     ... )
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -84,7 +84,7 @@ def grdclip(grid, **kwargs):
     -------
     >>> import pygmt
     >>> # Load a grid of @earth_relief_30m data, with a longitude range of
-    >>> # 10 E to 30 E, and a latitude range of 15 N to 25 N
+    >>> # 10° E to 30° E, and a latitude range of 15° N to 25° N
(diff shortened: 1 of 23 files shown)


pygmt/src/grdclip.py
```diff
@@ -83,8 +83,8 @@ def grdclip(grid, **kwargs):
     Example
     -------
     >>> import pygmt
-    >>> # Load a grid of @earth_relief_30m data, with an x-range of 10 to 30,
-    >>> # and a y-range of 15 to 25
+    >>> # Load a grid of @earth_relief_30m data, with a longitude range of
```
seisman, line 86: Actually, I'm not sure if we should have the `°` symbol in the examples. It's a non-ASCII symbol.

seisman: This is how the `grdhisteq` docstrings look like if opened with "GB 2312" encoding. It looks pretty bad.
Screenshot from 2023-10-12 16-10-14

GB 2312 is the default encoding for the Windows Chinese version, but I'm not sure about the default encoding when VSCode opens a plaintext file.

yvonnefroehlich: Yeah, that looks not good, and we should change this to either not using the `°` symbol or writing `deg`.

seisman: I prefer to "10 E", instead of "10 deg E".

yvonnefroehlich: I have updated this in commit https://github.com/GenericMappingTools/pygmt/pull/2740/commits/88d4b751ea011945b44c966f71e0e4ba45351605.

seisman: Sorry, after thinking twice, I think I've changed my mind. These non-ASCII characters are rendered correctly on the PyGMT site, so they won't affect most users.

Users who open PyGMT source codes in their text editor should know that UTF-8 is always the best and default encoding for Python codes (actually for any codes).

I feel we should revert the changes in 88d4b751ea011945b44c966f71e0e4ba45351605.

yvonnefroehlich: No worries, that's fine with me. I reverted commit https://github.com/GenericMappingTools/pygmt/commit/88d4b751ea011945b44c966f71e0e4ba45351605 in commit https://github.com/GenericMappingTools/pygmt/pull/2740/commits/df55aeebff1fff2bb0f028d3ce00f55b7d9cff9b.

seisman (approved):

Issue #2731, opened 2023-10-09
Consistent table-like input for PyGMT functions/methods


seisman:
xref https://github.com/GenericMappingTools/pygmt/issues/1318 and https://github.com/GenericMappingTools/pygmt/pull/2722.

When working on the tutorial for table inputs in PyGMT, I realized that the behavior of table inputs are still inconsistent. 

Generally speaking, there are two different ways to pass tables into PyGMT: via `data` or via `x`/`y`/`z`. For `data`, it can be a file, a pandas.DataFrame, a 2-D NumPy array. Currently, some functions/methods don't support x/y/z. We need to check all methods/functions to make sure they have consistent behavior for table inputs.

This issue is a central place to track the progress. Here is the list of methods/functions that we may need to check:

- [ ] `binstats`
- [ ] `blockm*`
- [ ] `contour`
- [ ] `filter1d`
- [ ] `histogram`
- [ ] `info`
- [ ] `legend`
- [ ] `meca`
- [ ] `nearneighbor`
- [ ] `plot`
- [ ] `plot3d`
- [ ] `project`
- [ ] `rose`
- [ ] `select`
- [ ] `sph2grd`
- [ ] `sphdistance`
- [ ] `sphinterpolate`
- [ ] `surface`
- [ ] `ternary`
- [ ] `text`
- [ ] `triangulate`
- [ ] `velo`
- [ ] `wiggle`
- [ ] `xyz2grd`
- [ ] `x2sys_cross`

PR #2744, opened 2023-10-14
Refactor the virtualfile_in function to accept more 1-D arrays 

seisman:

Here are the current definitions of the `virtualfile_from_data` method and the `data_kind` function:
https://github.com/GenericMappingTools/pygmt/blob/c9d6147bcf383903ffc75fb94f9787b6b3a1d231/pygmt/clib/session.py#L1473-L1483

https://github.com/GenericMappingTools/pygmt/blob/c9d6147bcf383903ffc75fb94f9787b6b3a1d231/pygmt/helpers/utils.py#L110

When I started issue https://github.com/GenericMappingTools/pygmt/issues/2731, I realized the current function definitions have some limitations:

1. For some modules, the number of input columns can vary, depending on the given options. For example, `binstats` usually requires 3 columns (x/y/z), but only requires 2 columns (x/y) if `-Cn` is used, and requires 4 columns (x/y/z/w) if `-W` is used. I don't think we want to add `w=None` and `required_w=False` to these functions. Also, we don't check if the input table has the required number of columns.
2. The `data_kind` function does three things: (1) determines the `kind` of the input data, and (2) checks if the data/x/y/z combinations are valid; and (3) checks if the `matrix`-type data has 3 columns. The `data_kind` function is called inside `virtualfile_from_data`, but sometimes we need to know in data kind when wrapping GMT modules, for example, in `Figure.plot` and `Figure.plot3d`. It means the `data_kind` function is called twice, which is not necessary. 

**Solutions:**

1. Refactor the `virtualfile_from_data` function like `data=None, vectors=None, names=["x", "y"]`. `vectors` is a list of vectors (e.g., `vectors=[x, y])` and `names` is a list of column names. The wrappers are responsible for preparing the list of 1-D arrays (`vectors`) and counting the column names (`names`).
2. Let the `data_kind` focus on determining the data kind and have another separate function to check if the input data/vectors are valid.


pygmt/helpers/utils.py
```diff
@@ -15,127 +15,133 @@
 from pygmt.exceptions import GMTInvalidInput
 
 
-def _validate_data_input(
-    data=None, x=None, y=None, z=None, required_z=False, required_data=True, kind=None
+def validate_data_input(
```
seisman, line 18: I think it's more useful to pass the list of column names instead, i.e., replacing `ncols=2` with `names=["x", "y"]`.

So, for most modules, `vectors=["x", "y"]` and `names=["x", "y"]` or `vectors=[x, y, z]` and `names=["x", "y", "z"]`.

For more complicated modules like `plot` or `plot3d`, the names can be 
`names=["x", "y", "direction_arg1", "direction_arg2", "fill", "size", "symbol", "transparency"]`.

The column names will be very useful when the GMTInvalidInput exception is raised.
For example, instead of `"Column 5 can't be None."`, we can say `"Column 5 ('size') can't be None."`. Instead of `"data must have at least 8 columns."`, we can say
```
data must have at least 8 columns:
x y direction_arg1 direction_arg2 fill size symbol transparency
```

seisman: Done in f37413b2aa03c24de6c9ea1e94053d7607b4d9b2

pygmt/clib/session.py
```diff
@@ -1495,13 +1493,12 @@ def virtualfile_from_data(
...
-            Optional. A list of numpy arrays in addition to x, y, and z.
-            All of these arrays must be of the same size as the x/y/z arrays.
-        required_z : bool
-            State whether the 'z' column is required.
+        vectors : list of 1-D arrays or None
+            A list of 1-D arrays. Each array will be a column in the table.
+            All of these arrays must be of the same size.
+        names : list of str
+            A list of names for each of the columns. Must be of the same size
+            as the number of vectors. Default is ``["x", "y"]``.
```
seisman, line 1501:
```suggestion
        names : str or list of str
            A list of names for each of the columns. Must be of the same size
            as the number of vectors. Default is ``"xy"``.
```

pygmt/helpers/utils.py
```diff
@@ -15,127 +15,154 @@
...
+        Pass in either a file name or :class:`pathlib.Path` to an ASCII data
+        table, an :class:`xarray.DataArray`, a 1-D/2-D
+        {table-classes} or an option argument.
+    vectors : list of 1-D arrays
+        A list of 1-D arrays with the data columns.
+    names : list of str
+        List of column names.
+    required_data : bool
+        Set to True when 'data' is required, or False when dealing with
+        optional virtual files. [Default is True].
```
michaelgrund, line 36:
```suggestion
        optional virtual files [Default is True].
```

pygmt/helpers/utils.py
```diff
@@ -15,127 +15,154 @@
...
-    * a pathlib.PurePath object provided as 'data'
-    * an xarray.DataArray object provided as 'data'
-    * a 2-D matrix provided as 'data'
-    * 1-D arrays x and y (and z, optionally)
-    * an optional argument (None, bool, int or float) provided as 'data'
+    Possible data kinds:
 
-    Arguments should be ``None`` if not used. If doesn't fit any of these
-    categories (or fits more than one), will raise an exception.
+    - ``'file'``: a file name or a pathlib.PurePath object providfed as 'data'
```
michaelgrund, line 148:
```suggestion
    - ``'file'``: a file name or a pathlib.PurePath object provided as 'data'
```

michaelgrund (approved):

pygmt/helpers/utils.py
```diff
@@ -15,127 +15,154 @@
 from pygmt.exceptions import GMTInvalidInput
 
 
-def _validate_data_input(
-    data=None, x=None, y=None, z=None, required_z=False, required_data=True, kind=None
+def validate_data_input(
```
weiji14, line 18: Any reason to remove the leading underscore from `_validate_data_input`? Do we want to make this function more public?

pygmt/helpers/utils.py
```diff
@@ -15,127 +15,154 @@
...
+    if kind is None:
+        kind = data_kind(data=data, required=required_data)
+
+    if kind == "vectors":  # From data_kind, we know that data is None
+        if vectors is None:
+            raise GMTInvalidInput("No input data provided.")
+        if len(vectors) < len(names):
+            raise GMTInvalidInput(
+                f"Requires {len(names)} 1-D arrays but got {len(vectors)}."
+            )
```
weiji14, line 119: Missing unit test for this if-condition.

pygmt/helpers/utils.py
```diff
@@ -15,127 +15,154 @@
...
+            msg = f"data must have at least {len(names)} columns.\n" + " ".join(names)
             if hasattr(data, "shape"):  # np.ndarray or pd.DataFrame
-                if len(data.shape) == 1 and data.shape[0] < 3:
-                    raise GMTInvalidInput("data must provide x, y, and z columns.")
-                if len(data.shape) > 1 and data.shape[1] < 3:
-                    raise GMTInvalidInput("data must provide x, y, and z columns.")
-            if hasattr(data, "data_vars") and len(data.data_vars) < 3:  # xr.Dataset
-                raise GMTInvalidInput("data must provide x, y, and z columns.")
+                if len(data.shape) == 1 and data.shape[0] < len(names):
+                    raise GMTInvalidInput(msg)
```
weiji14, line 130: Missing unit test for this if-condition.

pygmt/src/project.py
```diff
@@ -222,13 +222,18 @@ def project(data=None, x=None, y=None, z=None, outfile=None, **kwargs):
             "The `convention` parameter is not allowed with `generate`."
         )
 
+    # z is optional
+    vectors, names = [x, y], "xy"
+    if z is not None:
+        vectors.append(z)
```
weiji14, line 228: Need to append 'z' to `names` here? Also, need a unit test for this if-condition.

pygmt/tests/test_helpers.py
```diff
@@ -31,25 +29,6 @@ def test_load_static_earth_relief():
...
-        (None, np.array([1, 2]), None),
-        (None, None, np.array([4, 5])),
-    ],
-)
-def test_data_kind_fails(data, x, y):
-    """
-    Make sure data_kind raises exceptions when it should.
-    """
-    with pytest.raises(GMTInvalidInput):
-        data_kind(data=data, x=x, y=y)
```
weiji14, line 50: Recycle these parametrized tests for `validate_data_inputs`? Or are they all implemented in the doctests already?

pygmt/helpers/utils.py
seisman: Because this function is now called in `pygmt/clib/sessions.py`.

pygmt/clib/session.py
```diff
@@ -1530,8 +1527,13 @@ def virtualfile_from_data(  # noqa: PLR0912
...
-        kind = data_kind(
-            data, x, y, z, required_z=required_z, required_data=required_data
+        kind = data_kind(data, required=required_data)
+        validate_data_input(
+            data=data,
+            vectors=vectors,
+            names=names,
+            required_data=required_data,
+            kind=kind,
         )
```
weiji14, line 1537: The validation checks have been moved from within `data_kind` to `virtualfile_from_data` here. But in `plot.py`, we actually use `data_kind` on its own here:

https://github.com/GenericMappingTools/pygmt/blob/3076ddcff5d21c6d5e06c9e588b5a785b6c3eec6/pygmt/src/plot.py#L217

Are we ok with raising GMTInvalidInput much later here in `virtualfile_from_data` (after all the keyword argument parsing), rather than early on in `data_kind`?

pygmt/src/project.py
seisman: Actually no. The problem is that `project` only requires two columns, but three or more columns are required. Currently, the variable `names` are used for two purposes: (1) names of passed columns; (2) the number of columns. So, if we append `z` to `names` here, the calling `pygmt.project(data=data)` will fail if `data` has only two columns. I think we still need to maintain a variable for the number of required columns.

seisman:
Closing this PR since it will be superseded by #3369.

PR #2694, opened 2023-09-17
Bump ghostscript to 10.02.1

seisman:
Bump ghostscript from 9.54.0 to 10.02.1.

Will update baseline images in later PRs.

PR #2754, opened 2023-10-18
CI: Pin ghostscript to 9.54.0 in Tests and Dev Tests workflows

seisman:

The latest gs version is 10.02.0 but our baseline images were generated with gs 9.54.0, thus we will have lots of failures if we use gs 10.02.0 (see PR #2694).

We should update our gs to 10.02.0 and update all baseline images, but we decided to wait for GMT 6.5 first (https://github.com/GenericMappingTools/pygmt/pull/2694#issuecomment-1724749593) so that we don't have to update baseline images twice in a short period.

In this PR, Ghostscript is pinned to 9.54.0 for the Tests and Dev Tests workflow, so that we don't have failures due to gs 10.02.0. Other workflows are not changed because they don't care about the gs version.

We will pin Ghostscript to 10.02.0 in PR https://github.com/GenericMappingTools/pygmt/pull/2694.

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -106,6 +106,7 @@ jobs:
           create-args: >-
             python=${{ matrix.python-version }}${{ matrix.optional-packages }}
             gmt=6.4.0
+            ghostscript=9.54.0
             numpy=${{ matrix.numpy-version }}
             pandas
             xarray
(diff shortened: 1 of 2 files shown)


yvonnefroehlich (commented):
One issue with gs version 10.02.0 is that the (semi-)transparency is (again) broken.

yvonnefroehlich (approved):

seisman:
> One issue with gs version 10.02.0 is that the (semi-)transparency is (again) broken.

yes, gs 10.02.0 works well with GMT 6.5.0

Issue #2599, opened 2023-07-11
Figure.grdimage: "img_out" ("A") does not work


yvonnefroehlich:
_Related to this post in the GMT Forum_: https://forum.generic-mapping-tools.org/t/how-to-create-a-geotiff-with-pygmt/4113

The `img_out` (**A**) parameter of `Figure.grdimage` does not allow to save a grid to a raster format (.bmp, .gif, .jpg, .png, or .tif) as proposed by the documentation https://www.pygmt.org/dev/api/generated/pygmt.Figure.grdimage.html:

```python
import pygmt

fig = pygmt.Figure()
fig.grdimage(grid="@earth_age_01d_g", img_out="test_pygmt_grdimage.tif")
```

The relatd GMT code works (see https://docs.generic-mapping-tools.org/dev/grdimage.html#a):
```
gmt begin
    gmt grdimage @earth_age_01d_g -Atest_gmt_grdimage.tif
gmt end
```

In PyGMT, a figure can be saved as a .bmp, .gif, .jpg, .png, or .tif file via `Figure.savefig` (https://www.pygmt.org/dev/api/generated/pygmt.Figure.savefig.html). But I am not sure whether this is identical to using `img_out` of `pygmt.Figure.grdimage`.

```python
import pygmt

fig = pygmt.Figure()
fig.grdimage(grid="@earth_age_01d_g")
fig.savefig(fname="test_pygmt_savefig.tif")  
```

**Error message**
```
GMTCLibError: Module 'grdimage' failed with status code 72:
grdimage [ERROR]: Option -A: No output argument allowed
grdimage [ERROR]: Option -A: Must provide an output filename for image 
```

**System information**
```
PyGMT information:
  version: v0.9.1.dev98
System information:
  python: 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 17:59:51) [MSC v.1935 64 bit (AMD64)]
  executable: C:\ProgramData\Anaconda3\envs\pygmt_env_dev\python.exe
  machine: Windows-10-10.0.19045-SP0
Dependency information:
  numpy: 1.24.3
  pandas: 2.0.2
  xarray: 2023.1.1.dev17
  netCDF4: 1.6.2
  packaging: 23.1
  contextily: 1.3.0
  geopandas: 0.13.2
  IPython: 8.14.0
  rioxarray: 0.14.1
  ghostscript: 9.54.0
GMT library information:
  binary version: 6.4.0
  cores: 4
[...]

PR #2765, opened 2023-10-24
Figure.grdimage: Let the img_out parameter work as documented

seisman:

grdimage `-A` option works diffenrently for GMT CLI and external wrappers.

For GMT CLI, the syntax is like `gmt grdimage ingrd.nc -Aimg_out.xxx`.
For external wrappers, the syntax is like `gmt grdimage ingrid.nc -A > img_out.xxx`.

This PR makes `img_out` parameter working as documented, i.e., `img_out="out.tiff"` works.

Address https://github.com/GenericMappingTools/pygmt/issues/2599.


seisman:
This PR works well for me locally on macOS, but I have no idea why it fails on CI.

weiji14:
> This PR works well for me locally on macOS, but I have no idea why it fails on CI.

Is this only for GMT 6.5? Doesn't work locally for me with GMT 6.4.0 on Linux.

pygmt/src/grdimage.py
```diff
@@ -175,6 +175,14 @@ def grdimage(self, grid, **kwargs):
     """
     kwargs = self._preprocess(**kwargs)  # pylint: disable=protected-access
 
+    # Special handling of -A option.
+    # For -A option, the syntax is different for GMT CLI and external wrappers.
+    # For GMT CLI, "gmt grdimage ingrid.nc -Aimg_out;xxx".
```
yvonnefroehlich, line 180:
```suggestion
    # For GMT CLI, "gmt grdimage ingrid.nc -Aimg_out.xxx".
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers 

Please see #2599 for the related discussions, especially comments like https://github.com/GenericMappingTools/pygmt/issues/2599#issuecomment-1751795014 and https://github.com/GenericMappingTools/pygmt/issues/2599#issuecomment-1754938986.

The grdimage module usually writes PostScript codes to a hidden PostScript file, but with `-A` option, it produces a raster/vector image without writing any PS codes. So, it's a totally different workflow.

As mentioned in https://github.com/GenericMappingTools/pygmt/issues/2599#issuecomment-1754938986, `grdimage -A` works like `gdal_translate`, but allows grid projection, shading and maybe more. So, the feature is still useful.

So, I propose to:

1. Do NOT support `-A` option in the `grdimage` method
2. Fully remove `img_out` parameter from the `grdimage` method
3. If such a feature is requested by users, then we can add a new function (e.g., `pygmt.grd2img`) which wraps the `grdimage -A` option.

Thoughts?

weiji14:
> Fully remove `img_out` parameter from the `grdimage` method

Sound good. So we're removing the alias without a deprecation cycle right? Since it didn't really work before?

seisman:
> > Fully remove `img_out` parameter from the `grdimage` method
> 
> Sound good. So we're removing the alias without a deprecation cycle right? Since it didn't really work before?

Yes, but I think it's better to raise a warning/error if `-A`/`img_out` is used.

seisman:
Superseded by #2907. Closing.

PR #2766, opened 2023-10-24
Figure.savefig: Add the 'worldfile' parameter to write a companion world file for raster images

seisman:

Address https://github.com/GenericMappingTools/pygmt/pull/2698#discussion_r1335316532.

Fixes #

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -296,13 +296,13 @@ class Figure:
             the arguments ``"t2"`` and ``"g2"`` to the ``anti_aliasing``
             parameter of :meth:`pygmt.Figure.psconvert`. Ignored if creating
             vector graphics.
-        show: bool
+        show : bool
             If ``True``, will open the figure in an external viewer.
-        worldfile: bool
+        worldfile : bool
             If ``True``, will create a companion
             `world file <https://en.wikipedia.org/wiki/World_file>`__ for the
             figure. The world file will have the same name as the figure file
-            but with different extension. See
+            but with different extension (e.g. tfw for tif). See
             https://en.wikipedia.org/wiki/World_file#Filename_extension
             for the convention of world file extensions.
             This parameter does not work for KML and GeoTIFF formats.
@@ -359,7 +359,7 @@ class Figure:
         if worldfile:
             if ext in ["kml", "tiff"]:
                 raise GMTInvalidInput(
-                    f"World file is not supported for '{ext}' format."
+                    f"Saving a world file is not supported for '{ext}' format."
                 )
             kwargs["W"] = True
 
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -304,8 +304,8 @@ class Figure:
             figure. The world file will have the same name as the figure file
             but with different extension (e.g. tfw for tif). See
             https://en.wikipedia.org/wiki/World_file#Filename_extension
-            for the convention of world file extensions.
-            This parameter does not work for KML and GeoTIFF formats.
+            for the convention of world file extensions. This parameter only
+            works for raster image formats.
         dpi : int
             Set raster resolution in dpi [Default is ``720`` for PDF, ``300``
             for others].
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -314,6 +314,7 @@ class Figure:
             :meth:`pygmt.Figure.psconvert`. Valid parameters are ``gs_path``,
             ``gs_option``, ``resize``, ``bb_style``, and ``verbose``.
         """
+        # pylint: disable=too-many-branches
         # All supported formats
         fmts = {
             "png": "g",
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -252,7 +252,14 @@ class Figure:
             )
 
     def savefig(
-        self, fname, transparent=False, crop=True, anti_alias=True, show=False, **kwargs
+        self,
+        fname,
+        transparent=False,
+        crop=True,
+        anti_alias=True,
+        show=False,
+        worldfile=False,
+        **kwargs,
     ):
         """
         Save the figure to a file.
@@ -291,6 +298,14 @@ class Figure:
             vector graphics.
         show: bool
             If ``True``, will open the figure in an external viewer.
+        worldfile: bool
+            If ``True``, will create a companion
+            `world file <https://en.wikipedia.org/wiki/World_file>`__ for the
+            figure. The world file will have the same name as the figure file
+            but with different extension. See
+            https://en.wikipedia.org/wiki/World_file#Filename_extension
+            for the convention of world file extensions.
+            This parameter does not work for KML and GeoTIFF formats.
         dpi : int
             Set raster resolution in dpi [Default is ``720`` for PDF, ``300``
             for others].
@@ -339,6 +354,13 @@ class Figure:
         if anti_alias:
             kwargs["Qt"] = 2
             kwargs["Qg"] = 2
+
+        if worldfile:
+            if ext in ["kml", "tiff"]:
+                raise GMTInvalidInput(
+                    f"World file is not supported for '{ext}' format."
+                )
+            kwargs["W"] = True
 
         self.psconvert(prefix=prefix, fmt=fmt, crop=crop, **kwargs)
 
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_figure.py
```diff
@@ -292,6 +292,33 @@ def mock_psconvert(*args, **kwargs):  # pylint: disable=unused-argument
...
+        assert Path(imgfile.name).stat().st_size > 0
+        worldfile_suffix = "." + fmt[1] + fmt[3] + "w"
+        assert Path(imgfile.name).with_suffix(worldfile_suffix).stat().st_size > 0
+
+
+@pytest.mark.parametrize("fmt", [".tiff", ".kml"])
+def test_figure_savefig_worldfile_unsupported_format(fmt):
+    """
+    Figure.savefig should raise an error when worldfile is requested for an
+    unsupported format.
```
yvonnefroehlich, line 313:
```suggestion
    Figure.savefig should raise an error when a world file is requested for
    an unsupported format.
```

pygmt/tests/test_figure.py
```diff
@@ -292,6 +292,33 @@ def mock_psconvert(*args, **kwargs):  # pylint: disable=unused-argument
     }
 
 
+@pytest.mark.parametrize("fmt", [".png", ".pdf", ".jpg", ".tif"])
+def test_figure_savefig_worldfile(fmt):
+    """
+    Check if worldfile is created when requested.
```
yvonnefroehlich, line 298:
```suggestion
    Check if a world file is created when requested.
```

pygmt/figure.py
```diff
@@ -291,6 +298,14 @@ def savefig(
             vector graphics.
         show: bool
             If ``True``, will open the figure in an external viewer.
+        worldfile: bool
```
weiji14, line 301:
```suggestion
        show : bool
            If ``True``, will open the figure in an external viewer.
        worldfile : bool
```

pygmt/figure.py
```diff
@@ -291,6 +298,14 @@ def savefig(
...
+            If ``True``, will create a companion
+            `world file <https://en.wikipedia.org/wiki/World_file>`__ for the
+            figure. The world file will have the same name as the figure file
+            but with different extension. See
```
weiji14, line 305: Add an example?
```suggestion
            but with different extension (e.g. tfw for tif). See
```

pygmt/figure.py
```diff
@@ -340,6 +356,13 @@ def savefig(
             kwargs["Qt"] = 2
             kwargs["Qg"] = 2
 
+        if worldfile:
+            if ext in ["kml", "tiff"]:
```
weiji14, line 360: Are worldfiles possible for other formats like pdf/bmp/eps?

pygmt/figure.py
```diff
@@ -340,6 +356,13 @@ def savefig(
...
+                raise GMTInvalidInput(
+                    f"World file is not supported for '{ext}' format."
```
weiji14, line 362:
```suggestion
                    f"Saving a world file is not supported for '{ext}' format."
```

pygmt/figure.py
seisman: Yes for PDF and BMP. 

Aslo yes for EPS, but for unknown reasons, the current codes doesn't produce a `.esw` file correctly. Need some time to find out why.

weiji14: Don't think I've ever heard of people create world files for PDF (pfw files?!), but ok then if it works. Have added another suggestion below to test that bmp/bpw files are created.

pygmt/tests/test_figure.py
```diff
@@ -292,6 +292,33 @@ def mock_psconvert(*args, **kwargs):  # pylint: disable=unused-argument
     }
 
 
+@pytest.mark.parametrize("fmt", [".png", ".pdf", ".jpg", ".tif"])
```
weiji14, line 295: Test `bmp`, and sort alphabetically. Edit to add `eps` when you figure out if an `esw` file can be created.
```suggestion
@pytest.mark.parametrize("fmt", ["bmp", ".jpg", ".pdf", ".png", ".tif"])
```

pygmt/figure.py
seisman: > Are worldfiles possible for other formats like pdf/bmp/eps?

For EPS, see an upstream issue report https://github.com/GenericMappingTools/gmt/issues/7976

seisman: See https://github.com/GenericMappingTools/gmt/issues/7976, a world file is useless for vector images like EPS and PDF.

pygmt/tests/test_figure.py
```diff
@@ -292,6 +292,27 @@ def mock_psconvert(*args, **kwargs):  # pylint: disable=unused-argument
...
 
+def test_figure_savefig_worldfile():
+    """
+    Check if a world file is created for supported formats and raise an error
+    for unsupported formats.
+    """
+    fig = Figure()
+    fig.basemap(region=[0, 1, 0, 1], projection="X1c/1c", frame=True)
+    # supported formats
+    for fmt in [".bmp", ".jpg", ".png", ".tif"]:
```
seisman, line 303: Need to add `.ppm` if #2771 is merged:
```suggestion
    for fmt in [".bmp", ".jpg", ".png", ".ppm", ".tif"]:
```

pygmt/tests/test_figure.py
```diff
@@ -292,6 +292,27 @@ def mock_psconvert(*args, **kwargs):  # pylint: disable=unused-argument
     }
 
 
+def test_figure_savefig_worldfile():
```
weiji14, line 295: Was there a reason to switch from the parametrized tests to a single test in a863132fb58629aa8efb1541ba88cbf7b13fcca4? It'll be harder to see which formats/extensions fail, if e.g. the first one in a list like ['.bmp', '.jpg', ...] raises an AssertionError, because the others in the list won't be tested.

pygmt/figure.py
```diff
@@ -297,6 +304,14 @@ def savefig(
...
         show: bool
             If ``True``, will open the figure in an external viewer.
+        worldfile : bool
+            If ``True``, will create a companion
+            `world file <https://en.wikipedia.org/wiki/World_file>`__ for the
+            figure. The world file will have the same name as the figure file
+            but with different extension (e.g. tfw for tif). See
+            https://en.wikipedia.org/wiki/World_file#Filename_extension
+            for the convention of world file extensions. This parameter only
+            works for raster image formats.
```
weiji14, line 314: (Geo)TIFF is a raster file format, but a worldfile won't be produced. Also, maybe best to just list out the supported extensions.

```suggestion
            works for non-georeferenced raster image formats (e.g. '.bmp',
            '.jpg', '.png', '.ppm', '.tif')
```

pygmt/tests/test_figure.py
seisman: The main idea is to reducing the number of tests and also testing times. Comparing https://github.com/GenericMappingTools/pygmt/actions/runs/6677102075/job/18146781350?pr=2766 and https://github.com/GenericMappingTools/pygmt/actions/runs/6648648090/job/18065964154?pr=2766, the old way (using `parametrize`) has 7 tests for worldfile, and the testing time is 0.5 * 5 + 0.25 * 2 = 3 seconds, while the new way (a single test) costs 1.5 s only.

weiji14: Ah ok, I see that we avoid having to call `fig.basemap` multiple times, so it will be faster. All good then.

pygmt/figure.py
seisman: In PR https://github.com/GenericMappingTools/pygmt/pull/2771 (commit 0c8184710ea93a59ce6e3b2027114edb53e63421), we have revised the function description to separate the list of raster and vector formats (preview: https://www.pygmt.org/dev/api/generated/pygmt.Figure.savefig.html#pygmt.Figure.savefig).

So I prefer to not list the formats again and prefer to revise the docstring to 
```
works for raster image formats (except GeoTIFF).
```

weiji14: Oh right, totally missed that. Your shorter suggestion sounds good.

seisman, line 314:
```suggestion
            works for raster image formats (except GeoTIFF).
```

weiji14 (approved):

PR #2772, opened 2023-10-27
Add the validate_output_type function to check the output_type parameter

seisman:

Fixes #

**Reminders**

**Slash Commands**


pygmt/helpers/validators.py
```diff
@@ -0,0 +1,41 @@
...
+Functions to check if given arguments are valid.
+"""
+import warnings
+
+from pygmt.exceptions import GMTInvalidInput
+
+
+def validate_output_type(output_type, outfile=None):
+    """
+    Check if the 'output_type' and 'outfile' parameters are valid.
```
yvonnefroehlich, line 11:
```suggestion
    Check if the ``output_type`` and ``outfile`` parameters are valid.
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers for review.

pygmt/src/grdhisteq.py
```diff
@@ -321,23 +321,11 @@ def compute_bins(
...
         """
-        # Return a pandas.DataFrame if ``outfile`` is not set
-        if output_type not in ["numpy", "pandas", "file"]:
-            raise GMTInvalidInput(
-                "Must specify 'output_type' either as 'numpy', 'pandas' or 'file'."
-            )
+        output_type = validate_output_type(output_type, outfile=outfile)
 
         if header is not None and output_type != "file":
             raise GMTInvalidInput("'header' is only allowed with output_type='file'.")
```
weiji14, line 327: Thinking if the order should be reversed here?
```suggestion
        if header is not None and output_type != "file":
            raise GMTInvalidInput("'header' is only allowed with output_type='file'.")

        output_type = validate_output_type(output_type, outfile=outfile)
```

pygmt/src/grdhisteq.py
```diff
@@ -13,6 +12,7 @@
     fmt_docstring,
     kwargs_to_strings,
     use_alias,
+    validate_output_type,
```
weiji14, line 15: Just on the naming, should we specify that this applies to tabular outputs? E.g. use something like `validate_output_table_type`? There are some functions like `pygmt.grdhisteq.equalize_grid` that has code looking like this:

```python
with GMTTempFile(suffix=".nc") as tmpfile:
    if isinstance(outgrid, str):
        output_type = "file"
    elif outgrid is None:
        output_type = "xarray"
        outgrid = tmpfile.name
    else:
        raise GMTInvalidInput("Must specify 'outgrid' as a string or None.")
```

If we ever decide to have more output_types for grids (e.g. NetCDF or TIFF or xarray or ???), then there might be a `validate_output_grid_type`.

seisman: I was thinking about adding a new parameter `kind` (valid values are `dataset`/`grid`) to make the function working for both tables and grids.

weiji14: There would a lot of if-thens, and the error/warning messages would be different between tables and grids though.

seisman: Ah right, tables use `outfile` but grids use `outgrid`.

seisman: > Just on the naming, should we specify that this applies to tabular outputs? E.g. use something like `validate_output_table_type`?

Renamed in becbb739b8c78d28684583bf7686e597917dd7b0.

weiji14 (approved):

PR #756, opened 2020-12-21
Add a CI job testing GMT master branch on Windows

seisman:
**Description of proposed changes**

This PR tests PyGMT with GMT **master** branch on **Windows**, 
using the GMT dev versions from conda-forge.

Note that, on Linux and macOS, we test GMT master branch by building the lastest
GMT source codes. However, on Windows, the GMT dev versions from conda-forge
are one or a few weeks behind the GMT master branch.

Fixes #755.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Notes**

- You can write `/format` in the first line of a comment to lint the code automatically

PR #2773, opened 2023-10-27
CI: Test GMT dev version on Windows by building from source

seisman:

In the "GMT Dev Tests" workflow, the way we get the GMT dev version is different on Linux/macOS and Windows. We build the GMT master branch source code on Linux/macOS, but on Windows, we install the dev version from the conda-forge/dev channel. The conda-forge/dev channel is updated every few weeks to months, so it's not always update-to-date. We choose to use the conda-forge/dev channel for Windows because building GMT on Windows is difficult.

However, it turns out that building GMT on Windows is much easier than I initially though, as long as we install all GMT dependencies via conda.

In this PR, I update the workflow to build GMT from source code on Windows, so that we're always using the same GMT dev version on all three platforms.

Supersedes https://github.com/GenericMappingTools/pygmt/pull/756

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -14,7 +14,7 @@ name: GMT Dev Tests
 
 on:
   # push:
-  #  branches: [ main ]
+  #   branches: [ main ]
   pull_request:
     types: [ready_for_review]
     paths-ignore:
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -130,7 +130,12 @@ jobs:
           mkdir build
           cd build
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
-          cmake -G Ninja .. -DCMAKE_INSTALL_PREFIX=${{ env.GMT_INSTALL_DIR }} -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${{ env.MAMBA_ROOT_PREFIX }}\envs\pygmt\Library
+          cmake -G Ninja .. ^
+            -DCMAKE_INSTALL_PREFIX=${{ env.GMT_INSTALL_DIR }} ^
+            -DCMAKE_BUILD_TYPE=Release ^
+            -DCMAKE_PREFIX_PATH=${{ env.MAMBA_ROOT_PREFIX }}\envs\pygmt\Library ^
+            -DGMT_ENABLE_OPENMP=TRUE ^
+            -DGMT_USE_THREADS=TRUE
           cmake --build .
           cmake --build . --target install
         env:
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -3,10 +3,8 @@
 # This workflow runs regular PyGMT tests with the GMT dev version, and also
 # pre-release versions of several dependencies like NumPy, Pandas, Xarray, etc.
 # If any tests fail, it also uploads the diff images as workflow artifacts.
-# On Linux/macOS, GMT dev version is installed by fetching the latest source
-# codes from the GMT master branch and compiling.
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_tests_dev.yaml
```diff
@@ -116,6 +114,33 @@ jobs:
...
+
+      - name: Install GMT ${{ matrix.gmt_git_ref }} branch (Windows)
+        shell: cmd
+        run: |
+          git clone --depth=1 --single-branch --branch ${{ env.GMT_GIT_REF }} https://github.com/GenericMappingTools/gmt
+          cd gmt/
+          mkdir build
+          cd build
+          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          cmake -G Ninja .. -DCMAKE_INSTALL_PREFIX=${{ env.GMT_INSTALL_DIR }} -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${{ env.MAMBA_ROOT_PREFIX }}\envs\pygmt\Library
```
weiji14, line 133: We're not using many of the build flags in https://github.com/conda-forge/gmt-feedstock/blob/8cf880960420215125ee43b1cbd24b3fc6d03291/recipe/bld.bat#L10-L28. Is that ok?

weiji14 (approved):
Nice! The compilation on Windows worked at https://github.com/GenericMappingTools/pygmt/actions/runs/6680586107/job/18153844748#step:8:405. I'll approve this since the build works, but I have one comment about setting more build flags (which can be done separately if needed).

.github/workflows/ci_tests_dev.yaml
seisman: We set the [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html) flag to tell CMake where to find these libraries, so these flags are not necessary:
```
      -D FFTW3_ROOT=%LIBRARY_PREFIX% ^
      -D GDAL_ROOT=%LIBRARY_PREFIX% ^
      -D NETCDF_ROOT=%LIBRARY_PREFIX% ^
      -D PCRE_ROOT=%LIBRARY_PREFIX% ^
      -D ZLIB_ROOT=%LIBRARY_PREFIX% ^
      -D CURL_ROOT=%LIBRARY_PREFIX% ^
```

Parallelization support. Can be enable if necessary.
```
      -D GMT_ENABLE_OPENMP=TRUE ^
      -D GMT_USE_THREADS=TRUE ^
```
Not necessary:
```
      -D GMT_LIBDIR=lib ^
```

GSHHG and DCW data are not installed, but they can be automatically downloaded when needed.
```
      -D DCW_ROOT=%DCW_DIR% ^
      -D GSHHG_ROOT=%GSHHG_DIR% ^
      -D COPY_GSHHG=TRUE ^
      -D COPY_DCW=TRUE ^
```

Unnecessary flags for our "GMT Dev Tests": 
```
      -D GMT_INSTALL_TRADITIONAL_FOLDERNAMES=FALSE ^
      -D GMT_INSTALL_MODULE_LINKS=FALSE ^
```

seisman:
It's unclear to me why we have a dvc error in this branch, which lead to ~200 failures
```
ERROR: unexpected error - failed to load directory ('4d', 'd0ad31844cb0b0b451648cda314e2a.dir'): [Errno 2] No such file or directory: 'D:\\a\\pygmt\\pygmt\\.dvc\\cache\\files\\md5\\4d\\d0ad31844cb0b0b451648cda314e2a.dir'
```

The scheduled runs two days ago don't have the issue https://github.com/GenericMappingTools/pygmt/actions/runs/6661401871/job/18104207392.

.github/workflows/ci_tests_dev.yaml
weiji14: > Parallelization support. Can be enable if necessary.

Adding this would be good!

seisman: Done in c6b1baacb601fec3de62389a05667a7bbee13a1e.

PR #2662, opened 2023-09-04
Update code block separators to "# %%" in examples

yvonnefroehlich:
**Description of proposed changes**

This PR aims to replace the code block separator `###############################################################################` by `# %%` and also add this separator at thte beginning of the code:
- [x] Gallery examples
- [x] Tutorials
- [x] Intros
- [x] Projections
 
Fixes #2660

**Preview**:  https://pygmt-dev--2662.org.readthedocs.build/en/2662/tutorials/basics/coastlines.html

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2805, opened 2023-11-11
Gallery example "choropleth map": Use code block separator "# %%"

yvonnefroehlich:

Related to PR #2662, this PR aims to use the code block separator `# %%` in the gallery example "Choropleth map".

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/gallery/maps/choropleth_map.py
+++ b/examples/gallery/maps/choropleth_map.py
@@ -14,6 +14,7 @@ To fill the polygons based on a corresponding column you need to set
 parameter as shown in the example below.
 """
 
+# %%
 import geopandas as gpd
 import pygmt
 


seisman:
perhaps we should add a check to make sure that examples must have at least one # %% separators

seisman (approved):

yvonnefroehlich:
> perhaps we should add a check to make sure that examples must have at least one # %% separators

If this is something that can be added easily, I feel we should add such a check, as it's likely that we'll forget adding `#%%` also in upcoming examples. But if this is more complicated, I am not sure whether the effort is worth it.

weiji14:
> > perhaps we should add a check to make sure that examples must have at least one # %% separators
> 
> If this is something that can be added easily, I feel we should add such a check, as it's likely that we'll forget adding `#%%` also in upcoming examples. But if this is more complicated, I am not sure whether the effort is worth it.

Could try to find a regex statement that checks for the presence of at least one `# %%` in the examples/**/*.py files, and add it to https://github.com/GenericMappingTools/pygmt/blob/v0.10.0/.github/workflows/style_checks.yaml?

seisman:
> Could try to find a regex statement that checks for the presence of at least one `# %%` in the examples/**/*.py files, and add it to https://github.com/GenericMappingTools/pygmt/blob/v0.10.0/.github/workflows/style_checks.yaml?

This one-line command can be used to detect example files that don't have `# %%`:
```
grep -L '# %%' $(find examples -name '*.py')
```

PR #2808, opened 2023-11-13
Use Mypy for static type checking

seisman:
**Description of proposed changes**

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

Issue #2794, opened 2023-11-02
Support Type Hints


seisman:
*This issue is still in draft mode. Type hints are something new to me, so I need to learn more and will keep improving the issue description.*

**References:**

- https://docs.python.org/3/library/typing.html
- https://typing.readthedocs.io/en/latest/source/libraries.html
- https://typing.readthedocs.io/en/latest/source/best_practices.html

**Why we should support type hints**

1. Better auto-completion for parameter values #2793

weiji14:
Personally I do use type hints (and tried to add it to `pygmt.grdtrack` in 2019 apparently - https://github.com/GenericMappingTools/pygmt/pull/308#discussion_r287089810), but it can be a lot of work to add them across the whole project (and a bit tough for new contributors to learn).

Maybe we can start small with just `typing.Literal` since that's a practical one for when there's a choice of str type arguments, e.g. `load_earth_relief(...,  data_source: Literal["igpp", "gebco", "gebcosi", "synbath"], ...)` in #2793? Other types like bool/int/float don't offer any autocomplete so may be less useful for now.

seisman:
Yes, it's definitely a lot of work and a long-term goal. We need to split it into multiple smaller jobs so that people can help.

seisman:
What are the best type hints for parameters that expect a fixed number of values?

In GMT, many parameters accept a sequence with a fixed number of elements. For example, the `panel` parameter can take a sequence of two values, such as `panel=(1, 2)`.

There are two common options for type hints:

1. `tuple[int, int]`: It clearly indicates that the parameter expects exactly two integers. However, it restricts the argument to a tuple, excluding lists or NumPy arrays.
2. `Sequence[int]`: It accepts any sequence-like object (e.g., tuple, list, array), but does not explicitly specify the expected number of elements. [We're doing the checking at run time in the AliasSystem class.]

Currently, both styles are used, but I prefer option 2 for flexibility and plan to update the code for consistency. 

@GenericMappingTools/pygmt-maintainers Comments?

PR #2809, opened 2023-11-14
optional imports

seisman:
- Use a bool variable to indicate whether an optional module is available
- Use upper-case for constant name (Fix C0103)
- Use importlib.util.find_spec for modules that are not used (ruff F401)

Fixes #

**Reminders**

**Slash Commands**


pygmt/tests/test_figure.py
```diff
@@ -3,22 +3,19 @@
...
 Doesn't include the plotting commands which have their own test files.
 """
+import importlib
 import os
 from pathlib import Path
 
-try:
-    import IPython
-except ImportError:
-    IPython = None  # pylint: disable=invalid-name
```
weiji14, line 12: Another option looking at https://stackoverflow.com/questions/61384752/how-to-type-hint-with-an-optional-import/77341843#77341843 seems to be something like:

```python
IPython: Any = None
try:
    import IPython
except ImportError:
    pass
```

seisman: It's unclear what the recommended way is for importing an optional module. It seems sometimes pandas and xarray use the `has_xxx` bool variable:

https://github.com/pandas-dev/pandas/blob/feca574525b2c37fac8d1d26bf82d428b95630c8/pandas/compat/compressors.py#L11

https://github.com/pandas-dev/pandas/blob/feca574525b2c37fac8d1d26bf82d428b95630c8/pandas/io/formats/style.py#L84

https://github.com/pydata/xarray/blob/49bd63a8332c1930a866724a2968b2d880dae25e/xarray/core/ops.py#L15

but sometimes they also use the old way, which mypy should complain, e.g., 
https://github.com/pydata/xarray/blob/49bd63a8332c1930a866724a2968b2d880dae25e/xarray/coding/times.py#L31

weiji14: Looking at mypy (e.g. https://github.com/python/mypy/issues/1297#issuecomment-508593494), they also seem to recommend a try-except-else pattern?

seisman: I don't think it's officially recommended and it's also more difficult to understand than `IPython = None` and `_HAS_IPYTHON = False`.

weiji14: Ok, let's go with `_HAS_IPython` unless there's a better recommended way :slightly_smiling_face:

That said, we should probably decide first if adding `mypy` is something we want (https://github.com/GenericMappingTools/pygmt/pull/2808), otherwise the changes here aren't really needed.

seisman: Since we're going to add type hints (#2794, #2812), we definitely need a static type checker. There are many static type checkers, like [mypy](https://github.com/python/mypy), [pyright](https://github.com/microsoft/pyright）, and [pytype](https://github.com/google/pytype), in which mypy is the official one, so it's a good option to start.

pygmt/src/tilemap.py
```diff
@@ -6,9 +6,11 @@
 from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
 
 try:
-    import rioxarray
+    import rioxarray  # noqa: F401  # pylint: disable=unused-import
```
seisman, line 9:
```suggestion
    import rioxarray  # noqa: F401
```

PR #2813, opened 2023-11-15
TYP: Improve the doc style for type hints

seisman:
In commit cc6c4614eefcbcbe0c0de8a7e7ae547cba15d639, I added type hints to the `region`, `registration` and `data_source` parameters and also remove the parameter types from docstrings (following this [guide](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html)). Other parameters like `resolution` and `use_srtm` are not changed. So now this function has "partial" type hints support.

This PR tries to find out the best style for API documentations.

**Preview**: https://pygmt-dev--2813.org.readthedocs.build/en/2813/api/generated/pygmt.datasets.load_earth_relief.html

|No| | Signature | Docstrings | Comment |
|---|---|---|---|---|
| 1 | Default | image| image| Function signature is too complicated to read |
|2| 5c5864c6 |image|image| Function signature is easy to read, but the link-style types (e.g., `str`) are a little distractive |
|3|9ceacb9c|image|image| Types are still clickable but not as destructive as style 2|
|4|c7be7b1|image|image| Similar to style 3, but it also automatically show the default value|

I feel the current doc style (style 4 in the above table) is the best one. What do you think @GenericMappingTools/pygmt-maintainers?


pygmt/datasets/earth_relief.py
```diff
@@ -14,9 +17,9 @@
 @kwargs_to_strings(region="sequence")
 def load_earth_relief(
     resolution="01d",
-    region=None,
-    registration=None,
-    data_source="igpp",
+    region: Union[str, Sequence, None] = None,
+    registration: Literal["gridline", "pixel", None] = None,
+    data_source: Literal["igpp", "gebco", "gebcosi", "synbath"] = "igpp",
```
yvonnefroehlich, line 22: It is probably not possible to get:
- double quotation marks (`" "`) instead of single quotation marks (`' '`)
- an upper-case `Default`

283153676-d904cd92-67f1-4306-a73a-775a79229b4a_cut

pygmt/datasets/earth_relief.py
```diff
@@ -14,9 +17,9 @@
 @kwargs_to_strings(region="sequence")
 def load_earth_relief(
     resolution="01d",
-    region=None,
-    registration=None,
-    data_source="igpp",
+    region: Union[str, Sequence, None] = None,
+    registration: Literal["gridline", "pixel", None] = None,
```
yvonnefroehlich, line 21: > I feel the current doc style (style 4 in the above table) is the best one.

I agree that the signature of No 1 is too complicated. The highlighting of No 3 and No 4 is good, as it is consistent with the rest of the documentation.
Is it required to always give a default value? I am wondering if something like this can be a bit misleading (even though it is also given like this in the signature):
283153676-d904cd92-67f1-4306-a73a-775a79229b4a_cut_default

pygmt/datasets/earth_relief.py
seisman: I don't think it's possible.

pygmt/datasets/earth_relief.py
seisman: > Is it required to always give a default value? 

We don't have to give a default value, but I feel a default value in the function strings is good so that users don't have to go back to the function signature to know the default.



> I am wondering if something like this can be a bit misleading (even though it is also given like this in the signature): 283153676-d904cd92-67f1-4306-a73a-775a79229b4a_cut_default

Yes, it's misleading. We can rephrase it to something like: 

> Default is `None`, which means `"gridline"` for all resolutions except ...

yvonnefroehlich: > We don't have to give a default value, but I feel a default value in the function strings is good so that users don't have to go back to the function signature to know the default.

I also think it's good to have the default value in the function string. Just thought about not stating it for cases like above, but rephrasing the string should also prevent confusion.

yvonnefroehlich (approved):

PR #2827, opened 2023-11-23
Tutorial "Plotting data points": Change pen color for better contrast

yvonnefroehlich:

This PR suggests to change the `pen` color to improve the contrast in the tutorial "Plotting data points":

| `pen="black"` | `pen="white"`
| --- | --- |
| black_outline | white_outline |
**Reminders**

**Slash Commands**

### DIFF
--- a/examples/tutorials/basics/plot.py
+++ b/examples/tutorials/basics/plot.py
@@ -88,7 +88,7 @@ fig.plot(
     fill=data.depth_km,
     cmap=True,
     style="cc",
-    pen="black",
+    pen="white",
 )
 fig.colorbar(frame="af+lDepth (km)")
 fig.show()
--- a/examples/tutorials/basics/plot.py
+++ b/examples/tutorials/basics/plot.py
@@ -75,7 +75,6 @@ fig.show()
 # the earthquakes using :func:`pygmt.makecpt`, then set ``cmap=True`` in
 # :meth:`pygmt.Figure.plot` to use the colormap. At the end of the plot, we
 # also plot a colorbar showing the colormap used in the plot.
-#
 
 fig = pygmt.Figure()
 fig.basemap(region=region, projection="M15c", frame=True)


seisman:
> change the `pen` color to improve the contrast in the tutorial "Plotting data points":

But the other two images in the same gallery example all use a "black" pen.

yvonnefroehlich:
> > change the `pen` color to improve the contrast in the tutorial "Plotting data points":
> 
> But the other two images in the same gallery example all use a "black" pen.

Yes, but there the circles are not filled with a dark color, but with white color. So, the black outline of the circles is clearly visible.

michaelgrund:
Not sure if the white outline is a better solution across the whole plot since for the bright colors (right upper corner) the black outline works much better.

seisman:
> > > change the `pen` color to improve the contrast in the tutorial "Plotting data points":
> > 
> > 
> > But the other two images in the same gallery example all use a "black" pen.
> 
> Yes, but there the circles are not filled with a dark color, but with white color. So, the black outline of the circles is clearly visible.

The 3rd image is an improved version of the 2nd image (by setting `fill=data.depth_km` and `cmap=True`), so it would be better to keep the code changes as minimal as possible. Otherwise, we need to at least add one sentence explaining why we change the pen color to `white`.

seisman:
> Not sure if the white outline is a better solution across the whole plot since for the bright colors (right upper corner) the black outline works much better.

Agree with this point.

yvonnefroehlich:
I am sorry for the delay in my answer here 🙁. 

> > Not sure if the white outline is a better solution across the whole plot since for the bright colors (right upper corner) the black outline works much better.
> 
> Agree with this point.

Yes, for these three data points, a black outline works better. But the majority of data points has a dark fill color, for which a white outline color works better.

>>>> change the pen color to improve the contrast in the tutorial "Plotting data points":

>>> But the other two images in the same gallery example all use a "black" pen.

>> Yes, but there the circles are not filled with a dark color, but with white color. So, the black outline of the circles is clearly visible.

> The 3rd image is an improved version of the 2nd image (by setting `fill=data.depth_km` and `cmap=True`), so it would be better to keep the code changes as minimal as possible. Otherwise, we need to at least add one sentence explaining why we change the pen color to `white`.

I am fine with adding a comment explaining the change from a black to a white outline (`pen`) color.
But as this PR brought up more discussion than I expected, it is maybe easier to leave the tutorial as is, as the outline color is only a detail.

seisman:
I'm inclined to close this PR.

PR #2832, opened 2023-11-26
Figure.text: Improve the documentation to clarify the order of columns in the input files

seisman:
**Preview**: https://pygmt-dev--2832.org.readthedocs.build/en/2832/api/generated/pygmt.Figure.text.html#pygmt.Figure.text

Address https://github.com/GenericMappingTools/pygmt/issues/479#issuecomment-1751714146

Closes https://github.com/GenericMappingTools/pygmt/issues/479.

### DIFF
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -79,12 +79,12 @@ def text_(  # noqa: PLR0912
         A file name or a list of file names containing one or more records.
         Each record has the following columns:
 
-        * *x* : X coordinate or longitude
-        * *y* : Y coordinate or latitude
-        * *angle* : Angle in degrees counter-clockwise from horizontal
-        * *font* : Text size, font, and color
-        * *justify* : Two-character justification code
-        * *text* : The text string to typeset
+        * *x*: X coordinate or longitude
+        * *y*: Y coordinate or latitude
+        * *angle*: Angle in degrees counter-clockwise from horizontal
+        * *font*: Text size, font, and color
+        * *justify*: Two-character justification code
+        * *text*: The text string to typeset
 
         The *angle*, *font*, and *justify* columns are optional and can be set
         by using the ``angle``, ``font``, and ``justify`` parameters,
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -87,11 +87,10 @@ def text_(  # noqa: PLR0912
         * *text* : The text string to typeset
 
         The *angle*, *font*, and *justify* columns are optional and can be set
-        by using the ``angle``, ``font``, and ``justify`` parameters, respectively.
-        If these parameters are set to ``True``, then the corresponding columns
-        must be present in the input file(s) and the columns must be in the
-        order mentioned above.
-
+        by using the ``angle``, ``font``, and ``justify`` parameters,
+        respectively. If these parameters are set to ``True``, then the
+        corresponding columns must be present in the input file(s) and the
+        columns must be in the order mentioned above.
     x/y : float or 1-D arrays
         The x and y coordinates, or an array of x and y coordinates to plot
         the text.
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -77,7 +77,7 @@ def text_(  # noqa: PLR0912
     ----------
     textfiles : str or list
         A file name or a list of file names containing one or more records.
-        Each record has following columns:
+        Each record has the following columns:
 
         * *x* : X coordinate or longitude
         * *y* : Y coordinate or latitude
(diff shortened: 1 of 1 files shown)


pygmt/src/text.py
```diff
@@ -76,8 +76,22 @@ def text_(  # noqa: PLR0912
     Parameters
     ----------
     textfiles : str or list
-        A text data file name, or a list of file names containing 1 or more
-        records with (x, y[, angle, font, justify], text).
+        A file name or a list of file names containing one or more records.
+        Each record has following columns:
```
yvonnefroehlich, line 80:
```suggestion
        Each record has the following columns:
```

pygmt/src/text.py
```diff
@@ -76,8 +76,22 @@ def text_(  # noqa: PLR0912
...
+        * *angle* : Angle in degrees counter-clockwise from horizontal
+        * *font* : Text font
+        * *justify* : Two-character justification code
+        * *text* : The text string to typeset
+
+        The *angle*, *font*, and *justify* columns are optional and can be set
+        by using ``angle``, ``font``, and ``justify`` parameters, respectively.
+        If thse parameters are set to ``True``, then the corresponding columns
+        must be present in the input file(s) and the columns must be in the
+        order mentioned above.
```
yvonnefroehlich, line 93:
```suggestion
        The *angle*, *font*, and *justify* columns are optional and can be set
        by using the ``angle``, ``font``, and ``justify`` parameters, respectively.
        If these parameters are set to ``True``, then the corresponding columns
        must be present in the input file(s) and the columns must be in the
        order mentioned above.
```

pygmt/src/text.py
```diff
@@ -76,8 +76,22 @@ def text_(  # noqa: PLR0912
...
     textfiles : str or list
-        A text data file name, or a list of file names containing 1 or more
-        records with (x, y[, angle, font, justify], text).
+        A file name or a list of file names containing one or more records.
+        Each record has following columns:
+
+        * *x* : X coordinate or longitude
+        * *y* : Y coordinate or latitude
+        * *angle* : Angle in degrees counter-clockwise from horizontal
+        * *font* : Text font
```
yvonnefroehlich, line 85: Maybe we should mention that `font` allows not only to set the font itself but also the size and color?

pygmt/src/text.py
```diff
@@ -76,8 +76,21 @@ def text_(  # noqa: PLR0912
...
-        records with (x, y[, angle, font, justify], text).
+        A file name or a list of file names containing one or more records.
+        Each record has the following columns:
+
+        * *x* : X coordinate or longitude
+        * *y* : Y coordinate or latitude
+        * *angle* : Angle in degrees counter-clockwise from horizontal
+        * *font* : Text size, font, and color
+        * *justify* : Two-character justification code
+        * *text* : The text string to typeset
```
yvonnefroehlich, line 87:
```suggestion
        * *x*: X coordinate or longitude
        * *y*: Y coordinate or latitude
        * *angle*: Angle in degrees counter-clockwise from horizontal
        * *font*: Text size, font, and color
        * *justify*: Two-character justification code
        * *text*: The text string to typeset
```

pygmt/src/text.py
```diff
@@ -76,8 +76,21 @@ def text_(  # noqa: PLR0912
...
-        A text data file name, or a list of file names containing 1 or more
-        records with (x, y[, angle, font, justify], text).
+        A file name or a list of file names containing one or more records.
+        Each record has the following columns:
+
+        * *x*: X coordinate or longitude
+        * *y*: Y coordinate or latitude
+        * *angle*: Angle in degrees counter-clockwise from horizontal
+        * *font*: Text size, font, and color
+        * *justify*: Two-character justification code
```
yvonnefroehlich, line 86: Here we write currently `two-character justification code`. But we write
- below for the `justify` parameter `two-letter combination`.
- in the [colorbar example](https://www.pygmt.org/dev/gallery/embellishments/colorbar.html) `2-character combination`
- in the [text tutorial](https://www.pygmt.org/dev/tutorials/basics/text.html) `two-letter (order independent) code`
- etc.

I think it would be nice to formulate this consistently. This is a bit related to issue #2834. I will try to find time to look at this in more detail; currently busy with other things.

seisman: Yes, we can simplify these descriptions if issue #2834 is addressed.

yvonnefroehlich: Then maybe we leave these formulations as they are for now, and make it consistent via addressing #2834 🙂.

yvonnefroehlich (approved):

Issue #2800, opened 2023-11-07
Support pyarrow arrays and dataframes


weiji14:
### Description of the desired feature

[Apache Arrow](https://arrow.apache.org/docs/index.html) is an in-memory format that is starting to become a common exchange format between different libraries in Python and other programming languages. For example:

- Pandas 3.0 will use `pyarrow.string` instead of `object` dtype for strings, see [PDEP10](https://pandas.pydata.org/pdeps/0010-required-pyarrow-dependency.html), so we will eventually need to support PyArrow
- [Polars](https://github.com/pola-rs/polars) DataFrames can be zero-copy converted to Pyarrow via the `__dataframe__` protocol, see https://arrow.apache.org/docs/python/interchange_protocol.html

This issue is to track support of different [PyArrow data types](https://arrow.apache.org/docs/python/api/datatypes.html) in PyGMT:

- [ ] Numerical datatypes (uint/int/float) #2774
- [ ] String
  - Requires modifying the [`put_strings`](https://github.com/GenericMappingTools/pygmt/blob/v0.10.0/pygmt/clib/session.py#L917) method that currently uses `np.char.encode`
- [ ] Datetime
  - Requires modifying [`array_to_datetime`](https://github.com/GenericMappingTools/pygmt/blob/v0.10.0/pygmt/clib/conversion.py#L252) that expects Python datetime or numpy-backed arrays
  - Xref #242

Simplest way of integrating would be to just handle PyArrow-backed Pandas dataframes. But we can also discuss about using PyArrow as the internal array representation since it may allow better interoperability across other Python libraries using Arrow, and this might be relevant for #1318 and #2731.

Further reading:
- https://voltrondata.com/codex/standards-over-silos#1-2-standardizing-on-arrow

### Are you willing to help implement and maintain this feature?

Yes

seisman:
Before supporting pyarrow-backed pandas objects like what you're doing in PR #2774 and #2845, maybe we should check/support passing pyarrow arrays directly to PyGMT? If all/most pyarrow dtypes work, then we can go on with pyarrow-backed pandas objects. Then if interested, we may support polars.

weiji14:
> maybe we should check/support passing pyarrow arrays directly to PyGMT? If all/most pyarrow dtypes work, then we can go on with pyarrow-backed pandas objects. Then if interested, we may support polars.

[...]

PR #2884, opened 2023-12-16
Support timedelta64 dtype as input

### LINKED ISSUE #2848: Full support of numpy dtypes
Before starting to support PyArrow arrays (#2800), I think we should first ensure that PyGMT has complete support for NumPy dtypes. Currently, the following NumPy dtypes are supported:

https://github.com/GenericMappingTools/pygmt/blob/e5ecee98e316da0ec9176286f5d1bfb5623bcbc7/pygmt/clib/session.py#L85-L102

Here is a simple way to list all the available NumPy dtypes (xref: https://numpy.org/doc/stable/reference/arrays.scalars.html):
```python
>>> import numpy as np
>>> set(np.sctypeDict.values())
```

- [X] `numpy.int8`: `GMT_CHAR`
- [X] `numpy.int16`: `GMT_SHORT`
- [X] `numpy.int32`: `GMT_INT`
- [X] `numpy.int64`: `GMT_LONG`
- [X] `numpy.uint8`: `GMT_UCHAR`
- [X] `numpy.uint16`: `GMT_USHORT`
- [X] `numpy.uint32`: `GMT_UINT`
- [X] `numpy.uint64`: `GMT_ULONG`
- [ ] `numpy.float16`
- [X] `numpy.float32`: `GMT_FLOAT`
- [X] `numpy.float64`: `GMT_DOUBLE`
- [x] `numpy.longlong`: `GMT_LONG` in #3566
- [x] `numpy.ulonglong`: `GMT_ULONG` in #3566
- [ ] `numpy.longdouble`: alias for np.float96 or np.float128 (xref: https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.longdouble)
- [ ] ~~`numpy.complex64`~~: GMT doesn't support complex floating point data types.
- [ ] ~~`numpy.complex128`~~: GMT doesn't support complex floating point data types.
- [ ] ~~`numpy.clongdouble`~~: GMT doesn't support complex floating point data types.
- [ ] `numpy.bool`
- [ ] ~~`numpy.bytes_`~~: For bytestring. Makes no sense in GMT.
- [x] `numpy.str_`: `GMT_TEXT`
- [ ] `numpy.datetime64`: `GMT_DATETIME`. But still need to check all time units (https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units)
- [x] `numpy.timedelta64`: `GMT_LONG` #2884
- [ ] `numpy.object_`: Any Python objects. We should try to convert it to np.datetime64 or np.str_.
- [ ] ~~`numpy.void`~~: For creating a new structured or unstructured void scalar. Make no sense in GMT.

As you can see, most NumPy types are already supported in PyGMT. It's clear that GMT and PyGMT can't support float128 and any complex floating-point types, but at least it's possible to support `np.float16` (`np.half`), `np.bool_` and `np.timedelta`.

- For `np.float16`: maybe we have to cast the array into `np.float32`?
- For `np.bool_`: maybe convert to 0 and 1 (e.g., `array.astype("int8")`
- For `np.timedelta`: Not sure about this yet.

weiji14:

Map [`np.timedelta64`](https://numpy.org/doc/1.26/reference/arrays.scalars.html#numpy.timedelta64) to `GMT_LONG` in clib/session.py's DTYPES dictionary. Added a unit test in test_clib_put_vectors.py to test passing a numpy array with timedelta64 dtypes of various time units (year to microsecond).

TODO:
- [x] Allow passing in `np.timedelta64` via `put_vector`
- [ ] Handle `np.timedelta64` min/max values in the `region` parameter

Addresses #2848, extends #464 and #562.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -5,7 +5,6 @@ Apply them to functions wrapping GMT modules to automate: alias generation for
 arguments, insert common text into docstrings, transform arguments to strings,
 etc.
 """
-import datetime
 import functools
 import textwrap
 import warnings
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -673,19 +673,14 @@ def kwargs_to_strings(**conversions):
     >>> module(123, bla=(1, 2, 3), foo=True, A=False, i=(5, 6))
     {'A': False, 'bla': (1, 2, 3), 'foo': True, 'i': '5,6'}
     args: 123
-
-    >>> # Test that region accepts arguments with datetime or timedelta type
     >>> import datetime
     >>> module(
     ...     R=[
     ...         np.datetime64("2010-01-01T16:00:00"),
     ...         datetime.datetime(2020, 1, 1, 12, 23, 45),
-    ...         np.timedelta64(0, "h"),
-    ...         np.timedelta64(24, "h"),
     ...     ]
     ... )
-    {'R': '2010-01-01T16:00:00/2020-01-01T12:23:45.000000/0/24'}
-
+    {'R': '2010-01-01T16:00:00/2020-01-01T12:23:45.000000'}
     >>> import pandas as pd
     >>> import xarray as xr
     >>> module(
@@ -695,7 +690,6 @@ def kwargs_to_strings(**conversions):
     ...     ]
     ... )
     {'R': '2005-01-01T08:00:00.000000000/2015-01-01T12:00:00.123456'}
-
     >>> # Here is a more realistic example
     >>> # See https://github.com/GenericMappingTools/pygmt/issues/2361
     >>> @kwargs_to_strings(
@@ -766,23 +760,14 @@ def kwargs_to_strings(**conversions):
                 if fmt in separators and is_nonstr_iter(value):
                     for index, item in enumerate(value):
                         if " " in str(item):
-                            # Check if there is a space " " in the item, which
-                            # is typically present in objects such as
-                            # np.timedelta64, pd.Timestamp, or xr.DataArray.
-                            # If so, convert the item to a numerical or string
-                            # type that is understood by GMT as follows:
-                            if getattr(
-                                getattr(item, "dtype", ""), "name", ""
-                            ).startswith("timedelta"):
-                                # A np.timedelta64 item is cast to integer
-                                value[index] = item.astype("int")
-                            else:
-                                # A pandas.Timestamp/xr.DataArray containing
-                                # a datetime-like object is cast to ISO 8601
-                                # string format like YYYY-MM-DDThh:mm:ss.ffffff
-                                value[index] = np.datetime_as_string(
-                                    np.asarray(item, dtype=np.datetime64)
-                                )
+                            # Check if there is a space " " when converting
+                            # a pandas.Timestamp/xr.DataArray to a string.
+                            # If so, use np.datetime_as_string instead.
+                            # Convert datetime-like item to ISO 8601
+                            # string format like YYYY-MM-DDThh:mm:ss.ffffff.
+                            value[index] = np.datetime_as_string(
+                                np.asarray(item, dtype=np.datetime64)
+                            )
                     newvalue = separators[fmt].join(f"{item}" for item in value)
                     # Changes in bound.arguments will reflect in bound.args
                     # and bound.kwargs.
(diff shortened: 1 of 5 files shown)


pygmt/tests/test_plot.py
```diff
@@ -452,6 +452,25 @@ def test_plot_datetime():
...
 
+@pytest.mark.mpl_image_compare
+def test_plot_timedelta64():
+    """
+    Test plotting numpy.timedelta64 input data.
+    """
+    fig = Figure()
+    fig.basemap(
+        projection="X8c/5c",
+        region=[0, 8, 0, 10],
```
weiji14, line 463: Ideally, we could pass in `np.timedelta64` into the region argument as mentioned at https://github.com/GenericMappingTools/pygmt/issues/2848#issuecomment-1858698289.

```suggestion
        region=[np.timedelta64(0, "D"), np.timedelta64(8, "D"), 0, 10],
```

weiji14: Ah ok, looks like I modified `kwargs_to_strings` 3 years ago at #562 to handle `datetime` objects in `region`, so could probably do something similar here for `timedelta64`.

pygmt/helpers/decorators.py
```diff
@@ -760,14 +767,23 @@ def new_module(*args, **kwargs):
...
+                            # Check if there is a space " " in the item, which
+                            # is typically present in objects such as
+                            # np.timedelta64, pd.Timestamp, or xr.DataArray.
+                            # If so, convert the item to a numerical or string
+                            # type that is understood by GMT as follows:
+                            if getattr(
+                                getattr(item, "dtype", ""), "name", ""
+                            ).startswith("timedelta"):
+                                # A np.timedelta64 item is cast to integer
+                                value[index] = item.astype("int")
```
weiji14, line 779: This double getattr to obtain `item.dtype.name` isn't very nice, wondering if there's a good alternative.

pygmt/helpers/decorators.py
```diff
@@ -673,14 +674,19 @@ def kwargs_to_strings(**conversions):
...
     >>> module(
     ...     R=[
     ...         np.datetime64("2010-01-01T16:00:00"),
     ...         datetime.datetime(2020, 1, 1, 12, 23, 45),
+    ...         np.timedelta64(0, "h"),
+    ...         np.timedelta64(24, "h"),
     ...     ]
     ... )
-    {'R': '2010-01-01T16:00:00/2020-01-01T12:23:45.000000'}
+    {'R': '2010-01-01T16:00:00/2020-01-01T12:23:45.000000/0/24'}
```
weiji14, line 688: Apparently Python has a built-in [`datetime.timedelta`](https://docs.python.org/3/library/datetime.html#timedelta-objects) objects, but it's super hard to cast it to an integer type while preserving the orignal unit (e.g. day, hour, minute, etc). E.g.

```python
import datetime

td = datetime.timedelta(hours=24)

print(np.timedelta64(td))
# numpy.timedelta64(86400000000,'us')
print(np.timedelta64(td).astype(int))
# 86400000000
```

Should we still support `datetime.timedelta`? Or only `np.timedelta64`?

pygmt/helpers/decorators.py
```diff
@@ -760,14 +766,23 @@ def new_module(*args, **kwargs):
...
+                            ).startswith("timedelta"):
+                                # A np.timedelta64 item is cast to integer
+                                value[index] = item.astype("int")
+                            else:
+                                # A pandas.Timestamp/xr.DataArray containing
+                                # a datetime-like object is cast to ISO 8601
+                                # string format like YYYY-MM-DDThh:mm:ss.ffffff
+                                value[index] = np.datetime_as_string(
+                                    np.asarray(item, dtype=np.datetime64)
+                                )
```
weiji14, line 785: The if-then logic here might get a lot more complicated if we also decide to support passing PyArrow time/date types (xref #2800) into `region`/`-R`. At such, it might be good to isolate this logic into a separate function instead, so I'm thinking of maybe reverting commit d5bedc2ac71a337f2b2952843778d2e8c78c29a5 for now, but cherry-pick that change into a separate PR where we create a dedicated `_cast_datetime_to_str` function.

seisman: Sounds good.

weiji14: Ok, commit reverted in 849feda91c37a4ca2b449309b518c0a9310e5820. Should be ready to review now.

seisman (approved):
Looks good, but maybe wait for #2694?

weiji14:
Ok, re-generated the baseline image using GMT 6.5.0 and ghostscript 10.02.1 at 4426d951a753247746af5327408714aaf4d6f943, and test_plot_timedelta64 passes at 4426d951a753247746af5327408714aaf4d6f943, so merging.

PR #2890, opened 2023-12-17
Figure.timestamp: Add type hints to parameters

seisman:
**Preview**: https://pygmt-dev--2890.org.readthedocs.build/en/2890/api/generated/pygmt.Figure.timestamp.html

Address #2794.


seisman:
I've self-approved this PR so that I can continue with other more interesting coding.

PR #2894, opened 2023-12-18
Enable ruff's flake8-print (T20) rules

seisman:
Changes:

- Enable ruff's [flake8-print (T20)](https://docs.astral.sh/ruff/rules/#flake8-print-t20) rules. 
- Fix a few violations
- Ignore the rules in all examples

### DIFF
--- a/pygmt/tests/test_clib_virtualfiles.py
+++ b/pygmt/tests/test_clib_virtualfiles.py
@@ -80,7 +80,7 @@ def test_virtual_file_fails():
     with clib.Session() as lib, mock(lib, "GMT_Open_VirtualFile", returns=1):
         with pytest.raises(GMTCLibError):
             with lib.open_virtual_file(*vfargs):
-                print("Should not get to this code")
+                pass
 
     # Test the status check when closing the virtual file
     # Mock the opening to return 0 (success) so that we don't open a file that
@@ -91,7 +91,6 @@ def test_virtual_file_fails():
         with pytest.raises(GMTCLibError):
             with lib.open_virtual_file(*vfargs):
                 pass
-            print("Shouldn't get to this code either")
 
 
 def test_virtual_file_bad_direction():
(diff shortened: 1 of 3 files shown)


pygmt/tests/test_clib_virtualfiles.py
```diff
@@ -80,7 +80,7 @@ def test_virtual_file_fails():
     with clib.Session() as lib, mock(lib, "GMT_Open_VirtualFile", returns=1):
         with pytest.raises(GMTCLibError):
             with lib.open_virtual_file(*vfargs):
-                print("Should not get to this code")
+                pass
```
weiji14, line 83: Should we do like a `raise ValueError("Should not get to this code")` instead of silently passing? Not that the code should ever get to it, but just in case?

seisman: It's unnecessary. The test will fill if `lib.open_virtual_file` doesn't raise the `GMTCLibError` error.

Here is a minimal test to verify it:
```
import pytest

def test_zero_division():
    with pytest.raises(ZeroDivisionError):
        1 / 0

def test_zero_division_pass():
    with pytest.raises(ZeroDivisionError):
        pass
```
The output is:
```
========================================================================================== FAILURES ===========================================================================================
___________________________________________________________________________________ test_zero_division_pass ___________________________________________________________________________________

    def test_zero_division_pass():
>       with pytest.raises(ZeroDivisionError):
E       Failed: DID NOT RAISE <class 'ZeroDivisionError'>

test.py:8: Failed
=================================================================================== short test summary info ===================================================================================
FAILED test.py::test_zero_division_pass - Failed: DID NOT RAISE <class 'ZeroDivisionError'>
```

weiji14: Ok then!

weiji14 (approved):

PR #2897, opened 2023-12-19
Enable ruff's flake8-return (RET) rules and ignore RET504

seisman:
Enable ruff's [flake8-return](https://docs.astral.sh/ruff/rules/#flake8-return-ret) rules and ignore RET504.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,7 @@ select = [
     "PGH",  # pygrep-hooks
     "PIE",  # flake8-pie
     "PL",   # pylint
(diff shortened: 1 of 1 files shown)


weiji14:
Not sure about applying RET504 unnecessary-assign. I feel like it can be more readable to name what is being returned?

seisman:
> Not sure about applying RET504 unnecessary-assign. I feel like it can sometimes be more readable to name what is being returned?

I agree. RET504 is ignored.

weiji14 (approved):

PR #2905, opened 2023-12-20
CI: Use Python shell to make the download step more readable

seisman:
Reference: https://github.com/mamba-org/setup-micromamba#custom-micromamba-shell-wrapper


seisman:
Doesn't work as expected. Closing.

PR #2907, opened 2023-12-21
Figure.grdimage: Remove the unsupported 'img_out'/'A' parameter

seisman:
As proposed in https://github.com/GenericMappingTools/pygmt/pull/2765#issuecomment-1864131592, we will NOT support the `img_out`/`A` parameter in the `Figure.grdimage` method. The related feature will be added to a new function/method if requested.

Supersede #2765.

### DIFF
--- a/pygmt/src/grdimage.py
+++ b/pygmt/src/grdimage.py
@@ -2,6 +2,7 @@
 grdimage - Plot grids or images.
 """
 from pygmt.clib import Session
+from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import (
     build_arg_string,
     deprecate_parameter,
@@ -16,7 +17,6 @@ __doctest_skip__ = ["grdimage"]
 @fmt_docstring
 @deprecate_parameter("bit_color", "bitcolor", "v0.10.0", remove_version="v0.12.0")
 @use_alias(
-    A="img_out",
     B="frame",
     C="cmap",
     D="img_in",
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_grdimage.py
```diff
@@ -241,3 +241,14 @@ def test_grdimage_central_meridians_and_standard_parallels(grid, proj_type, lon0
     )
     fig_test.grdimage(grid, projection=f"{proj_type}{lon0}/{lat0}/15c", cmap="geo")
     return fig_ref, fig_test
+
+
+def test_grdimage_imgout(grid):
```
weiji14, line 246:
```suggestion
def test_grdimage_imgout_fails(grid):
```

pygmt/src/grdimage.py
```diff
@@ -172,6 +157,13 @@ def grdimage(self, grid, **kwargs):
     """
     kwargs = self._preprocess(**kwargs)
 
+    # Do not support -A option
+    if any(kwargs.get(arg) is not None for arg in ["A", "img_out"]):
```
weiji14, line 161:
```suggestion
    if any(kwargs.get(arg) for arg in ["A", "img_out"]):
```

Should be ok to remove `is not None`, e.g.:

```python
kwargs = {"img_out": "temp.nc"}
if any(kwargs.get(arg) for arg in ["A", "img_out"]):
    raise Exception
```
should raise an Exception.

weiji14 (approved):
A couple of minor suggestions, otherwise ok.

pygmt/src/grdimage.py
seisman: Won't raise an exception if an empty string is given:
```
kwargs = {"img_out": ""}
if any(kwargs.get(arg) for arg in ["A", "img_out"]):
    raise Exception
```

weiji14: Ah yes, keep it as is then.

Issue #2730, opened 2023-10-09
Get rid of temporary files from pygmt functions and plotting methods


seisman:
One of the project goals is to interface with GMT C API directly using ctypes without system calls. It lets us work on data in memory, which is more efficient than working on files on disk.

However, currently PyGMT still writes tables/grids/CPTs into temporary files. With PRs #2729 and #2398 (not merged yet), we can write data into memory and then working on these data-in-memory directly. Thus, it's possible to get rid of temporary files completely.

This issue report is the central place to track all the functions/methods that need to be refactored.

**Modules writting tables**

- [x] `blockm*` #3103
- [x] `filter1d` #3085
- [ ] `grdinfo` [Will be tracked in #593]
- [x] `grdtrack` #3106
- [x] `triangulate` #3107
- [x] `grdvolume` #3102
- [x] `select` #3108
- [x] `which` #3148
- [x] `grdhisteq` #3109
- [ ] `x2sys_cross` [Will be tracked in #3160]
- [ ] `info` [Will be tracked in #3159]
- [x] `project` #3110
- [x] `grd2xyz` #3097

**Modules writting grids**

Most modules are refactored in #2398, except `grdcut`.

- [x] `dimfilter`
- [x] `grdclip`
- [x] `sphinterpolate`
- [x] `grdsample`
- [x] `grdfilter`
- [x] `grdproject`
- [x] `grdgradient`
- [x] `grdfill`
- [x] `triangulate`
- [x] `sphdistance`
- [x] `xyz2grd`
- [ ] `grdcut` [#3115]
- [x] `grdhisteq`
- [x] `nearneighbor`
- [x] `sph2grd`
- [x] `grdlandmask`
- [x] `binstats`
- [x] `surface`

seisman:
Just tried to see if getting rid of temporary files can make PyGMT faster. The result are quit promising. For `grd2xyz`, PR #2729 is ~35% faster (130 ms in PR #2729 vs 200 ms in the main branch):

For the main branch:
```python
In [1]: import pygmt

In [2]: %timeit -n 10 -r 5 xyz = pygmt.grd2xyz("@earth_relief_01d_g", output_type="pandas")
200 ms ± 3.61 ms per loop (mean ± std. dev. of 5 runs, 10 loops each)

In [3]: %timeit -n 10 -r 5 xyz = pygmt.grd2xyz("@earth_relief_01d_g", output_type="pandas")
[...]

PR #2908, opened 2023-12-23
Setup Continuous Benchmarking workflow with pytest-codspeed

weiji14:

Measuring the execution speed of tests to track performance of PyGMT functions over time.

Using [`pytest-codspeed`](https://github.com/CodSpeedHQ/pytest-codspeed) to do the benchmarking, with the help of https://github.com/CodSpeedHQ/action. Decorated `test_basemap` with `@pytest.mark.benchmark` to see if the benchmarking works.

Note: Running the benchmarks on Python 3.12 to enable flame graph generation, available with [`pytest-codspeed>=2.0.0`](https://github.com/CodSpeedHQ/pytest-codspeed/releases/tag/v2.0.0) - see https://docs.codspeed.io/features/trace-generation.

References:
- https://codspeed.io/blog/introducing-codspeed
- https://docs.codspeed.io/benchmarks/python#running-the-benchmarks-in-your-ci

Relates to https://github.com/GenericMappingTools/pygmt/issues/2730#issuecomment-1863359297, addresses #2910.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -51,19 +51,17 @@ jobs:
 
       # Show installed pkg information for postmortem diagnostic
       - name: List installed packages
-        run: |
-          micromamba list
-          which gmt
-          echo ${{ github.workspace }}
+        run: micromamba list
 
       # Run the benchmark tests
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v2.0.2
         with:
           run: |
-            python -m pip install numpy pandas xarray netCDF4 packaging \
+            python -m pip install -U pip
+            python -m pip install -U numpy pandas xarray netCDF4 packaging \
                                   pytest pytest-benchmark pytest-codspeed \
-                                  setuptools
+                                  'setuptools>64'
             python -m pip list
             make install
             make test PYTEST_EXTRA="-r P --codspeed"
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -50,16 +50,16 @@ jobs:
           echo $CONDA
           conda install --solver=libmamba gmt=6.4.0 \
                         numpy pandas xarray netCDF4 packaging \
-                        pytest pytest-benchmark setuptools
-          $CONDA/bin/python -m pip install pytest-codspeed
+                        pytest pytest-benchmarks
+          $CONDA/bin/python -m pip install pytest-codspeed setuptools
 
       # Install dependencies from PyPI and run the benchmark tests
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v2.0.2
         with:
           run: |
-            ls -lh /usr/share/miniconda/envs/pygmt/lib/
+            ls -lh /usr/share/miniconda/lib/
             make install
             make test PYTEST_EXTRA="-r P --codspeed"
         env:
-          GMT_LIBRARY_PATH: /usr/share/miniconda/envs/pygmt/lib/
+          GMT_LIBRARY_PATH: /usr/share/miniconda/lib/
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -53,14 +53,16 @@ jobs:
                         pytest pytest-benchmark
           $CONDA/bin/python -m pip install -U pytest-codspeed setuptools
 
+      # Install the package that we want to test
+      - name: Install the package
+        run: $CONDA/bin/python -m pip install --no-deps -e .
+
       # Install dependencies from PyPI and run the benchmark tests
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v2.0.2
         with:
           run: |
-            ls -lh /usr/share/miniconda/lib/
-            $CONDA/bin/python -m pip install --no-deps -e .
             $CONDA/bin/python -c "import pygmt; pygmt.show_versions()"
-            PYGMT_USE_EXTERNAL_DISPLAY="false" pytest -r P --codspeed
+            PYGMT_USE_EXTERNAL_DISPLAY="false" $CONDA/bin/python -m pytest -r P --codspeed
         env:
           GMT_LIBRARY_PATH: /usr/share/miniconda/lib/
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -46,18 +46,16 @@ jobs:
       # Install GMT and dependencies from conda-forge
       - name: Install dependencies
         run: |
-          which conda
-          echo $CONDA
           conda install --solver=libmamba gmt=6.4.0 \
                         numpy pandas xarray netCDF4 packaging \
-                        pytest pytest-benchmark
+                        pytest pytest-benchmark pytest-mpl
           $CONDA/bin/python -m pip install -U pytest-codspeed setuptools
 
       # Install the package that we want to test
       - name: Install the package
         run: $CONDA/bin/python -m pip install --no-deps -e .
 
-      # Install dependencies from PyPI and run the benchmark tests
+      # Run the benchmark tests
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v2.0.2
         with:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -34,26 +34,22 @@ jobs:
           # fetch all history so that setuptools-scm works
           fetch-depth: 0
 
-      # Install Micromamba with conda-forge dependencies
-      - name: Setup Micromamba
-        uses: mamba-org/setup-micromamba@v1.7.3
+      # Install Miniconda with conda-forge dependencies
+      - name: Setup Miniconda
+        uses: conda-incubator/setup-miniconda@v3.0.1
         with:
-          environment-name: pygmt
-          condarc: |
-            channels:
-              - conda-forge
-              - nodefaults
-          cache-downloads: true
-          cache-environment: true
-          create-args: >-
-            python=3.12
-            gmt=6.4.0
+          activate-environment: pygmt
+          python-version: '3.12'
+          channels: conda-forge,nodefaults
+          channel-priority: strict
 
-      # Show installed pkg information for postmortem diagnostic
-      - name: List installed packages
-        run: micromamba list
+      # Install GMT from conda-forge
+      - name: Install GMT
+        run: |
+          conda install --solver=libmamba gmt=6.4.0
+          which gmt
 
-      # Run the benchmark tests
+      # Install dependencies from PyPI and run the benchmark tests
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v2.0.2
         with:
@@ -63,7 +59,6 @@ jobs:
                                   setuptools
             python -m pip list
             make install
-            which gmt
             make test PYTEST_EXTRA="-r P --codspeed"
         env:
-          GMT_LIBRARY_PATH: /home/runner/micromamba/envs/pygmt/lib/
+          GMT_LIBRARY_PATH: /home/runner/miniconda/envs/pygmt/lib/
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,73 @@
+# Run performance benchmarks
+#
+# Continuous benchmarking using pytest-codspeed. Measures the execution speed
+# of tests marked with @pytest.mark.benchmark decorator.
+
+name: Benchmarks
+
+on:
+  # Run on pushes to the main branch
+  push:
+    branches: [ main ]
+  # Uncomment the 'pull_request' line below to trigger the workflow in PR
+  pull_request:
+  # `workflow_dispatch` allows CodSpeed to trigger backtest
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  benchmarks:
+    runs-on: ubuntu-22.04
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+      # Checkout current git repository
+      - name: Checkout
+        uses: actions/checkout@v4.1.1
+
+      # Install Micromamba with conda-forge dependencies
+      - name: Setup Micromamba
+        uses: mamba-org/setup-micromamba@v1.7.3
+        with:
+          environment-name: pygmt
+          condarc: |
+            channels:
+              - conda-forge
+              - nodefaults
+          cache-downloads: true
+          cache-environment: true
+          create-args: >-
+            python=3.12
+            gmt=6.4.0
+            numpy=${{ matrix.numpy-version }}
+            pandas
+            xarray
+            netCDF4
+            packaging
+            pytest
+            pytest-benchmark
+
+      # Install dependencies from PyPI
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install the package
+        run: $CONDA/bin/python -m pip install --no-deps -e .
+
+      # Run the benchmark tests
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@v2
+        with:
+          run: |
+            $CONDA/bin/python -c "import pygmt; pygmt.show_versions()"
+            PYGMT_USE_EXTERNAL_DISPLAY="false" $CONDA/bin/python -m pytest -r P --pyargs pygmt --codspeed
```
weiji14, line 64: Hmm, the benchmark tests seems to be running fine on CI, but the upload step doesn't work. E.g. at https://github.com/GenericMappingTools/pygmt/actions/runs/7312581712/job/19923300422?pr=2908#step:6:121:

```python-traceback
   ============================= test session starts ==============================
  platform linux -- Python 3.12.1, pytest-7.4.3, pluggy-1.3.0 -- /usr/share/miniconda/bin/python
  codspeed: 2.2.0 (callgraph: enabled)
  CodSpeed had to disable the following plugins: pytest-benchmark
  cachedir: .pytest_cache
  benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
  Matplotlib: 3.8.2
  Freetype: 2.12.1
  rootdir: /home/runner/work/pygmt/pygmt
  configfile: pyproject.toml
  plugins: benchmark-4.0.0, codspeed-2.2.0, mpl-0.16.1
  collecting ... collected 679 items / 677 deselected / 8 skipped / 2 selected
  
  pygmt/tests/test_blockm.py::test_blockmean_input_dataframe PASSED        [ 50%]
  pygmt/tests/test_grd2xyz.py::test_grd2xyz PASSED                         [100%]
  ================================ 2 benchmarked =================================
  
  =============================== warnings summary ===============================
  ../../../../../usr/share/miniconda/lib/python3.12/site-packages/dateutil/tz/tz.py:37
    /usr/share/miniconda/lib/python3.12/site-packages/dateutil/tz/tz.py:37: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
      EPOCH = datetime.datetime.utcfromtimestamp(0)
  
  -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
  ==================================== PASSES ====================================
  ============================== slowest durations ===============================
  4.42s call     pygmt/tests/test_blockm.py::test_blockmean_input_dataframe
  3.04s setup    pygmt/tests/test_blockm.py::test_blockmean_input_dataframe
  1.85s call     pygmt/tests/test_grd2xyz.py::test_grd2xyz
  1.16s setup    pygmt/tests/test_grd2xyz.py::test_grd2xyz
  
  (2 durations < 0.2s hidden.  Use -vv to show these durations.)
  =========== 2 passed, 8 skipped, 677 deselected, 1 warning in 21.84s ===========
Upload the results
  CodSpeed Run Hash: "8332511d76e690d4a1ef10b5df14108240ab0ccb8be79e5059092401c7a5ec31"
  Preparing upload...
  Error: error decoding response body: missing field `status` at line 1 column 58
  Error: Process completed with exit code 1.
```

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,66 @@
...
+          channels: conda-forge,nodefaults
+          channel-priority: strict
+
+      # Install GMT and dependencies from conda-forge
+      - name: Install dependencies
+        run: |
+          conda install --solver=libmamba gmt=6.4.0 python=3.12 \
+                        numpy pandas xarray netCDF4 packaging \
+                        pytest pytest-benchmark pytest-mpl
+          $CONDA/bin/python -m pip install -U pytest-codspeed setuptools
```
weiji14, line 52: The CodSpeedHQ/action doesn't work with `setup-micromamba` yet because it doesn't seem to use the bash shell or something (see https://github.com/CodSpeedHQ/action/issues/65). So we'll need to manually use GitHub Actions' built-in conda to install GMT, and install all the dependencies into the `base` environment manually.

weiji14 (commented):
Finally got the benchmark code to run... But, the upload to codspeed isn't working yet. Need to investigate what is happening.

.github/workflows/benchmarks.yml
weiji14: Ah, it might be because we need to install it first (https://docs.codspeed.io/ci/github-actions#1-allow-the-action-to-upload-on-codspeed). @seisman, could you approve the installation request please?

image

seisman: Approved.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      # Install the package that we want to test
+      - name: Install the package
+        run: $CONDA/bin/python -m pip install --no-deps -e .
+
+      # Run the benchmark tests
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@v2
+        with:
```
seisman, line 61: Need a token here?
```suggestion
        with:
          token: ${{ secrets.CODSPEED_TOKEN }}
```

weiji14: No, I don't think a token is needed for public repos.

weiji14:
Closing and reopening to re-trigger tests.

.github/workflows/benchmarks.yml
weiji14: Hmm, I don't think I have admin permissions to enable the PyGMT repo on https://codspeed.io/settings/organizations/GenericMappingTools/general. Could you do it on your side @seisman?

seisman: Done. FYI, I also added the `CODSPEED_TOKEN` token following the instructions.

weiji14: Thanks! Let me try re-running the tests first without the token.

.github/workflows/benchmarks.yml
weiji14: Yep, didn't need to set a token, just have to install the GitHub App properly, see https://github.com/GenericMappingTools/pygmt/pull/2908#issuecomment-1868476695!

weiji14:
Ok, we've got CodSpeed to work :tada: So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?

seisman:
> Ok, we've got CodSpeed to work 🎉 So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?

I see the comment https://github.com/GenericMappingTools/pygmt/pull/2908#issuecomment-1868476695, but the workflow still fails.

weiji14:
> > Ok, we've got CodSpeed to work 🎉 So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?
> 
> I see the comment [#2908 (comment)](https://github.com/GenericMappingTools/pygmt/pull/2908#issuecomment-1868476695), but the workflow still fails.

Oh, I think it's because I re-ran the second lat commit... Let me re-run that one again.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,66 @@
...
+# of tests marked with @pytest.mark.benchmark decorator.
+
+name: Benchmarks
+
+on:
+  # Run on pushes to the main branch
+  push:
+    branches: [ main ]
+  # Uncomment the 'pull_request' line below to trigger the workflow in PR
+  pull_request:
```
seisman, line 13: Comment out this line before merging:
```suggestion
  # pull_request:
```

.github/workflows/benchmarks.yml
seisman, line 52: Add `$CONDA/bin` to PATH, so we don't have to use `$CONDA/bin/python`?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,66 @@
...
+                        pytest pytest-benchmark pytest-mpl
+          $CONDA/bin/python -m pip install -U pytest-codspeed setuptools
+
+      # Install the package that we want to test
+      - name: Install the package
+        run: $CONDA/bin/python -m pip install --no-deps -e .
+
+      # Run the benchmark tests
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@v2
```
seisman, line 60: Use a specific version like all other actions?
```suggestion
        uses: CodSpeedHQ/action@v2.0.2
```

seisman:
> Edit: Yep, all working now, with or without the token.

Great. I'll remove the token from GitHub later.

seisman:
Need to mention the new workflow in:

- `doc/maintenance.md`
- `.github/ISSUE_TEMPLATE/release_checklist.md` (maybe unnecessary in this file)
- `.github/ISSUE_TEMPLATE/bump_gmt_checklist.md`

seisman:
> So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?

I think we should focus on benchmarking low-level functions rather than modules' wrappers, since the low-level functions are heavily used in everywhere and most wrappers have very simple and similar code structures. 

For example, most plotting modules have very simple code like the one in `Figure.basemap`
```
    with Session() as lib:                                                      
        lib.call_module(module="basemap", args=build_arg_string(kwargs))
```
so we just need to benchmark one `basemap` test (e.g., `test_basemap()`) and don't need to benchmark other `basemap`'s tests and other plotting methods (e.g., `Figure.coast`). Of course, there are few exceptions, for example, `Figure.meca`, `Figure.plot`, `Figure.plot3d`, `Figure.text` are among the complicated wrappers and should be benchmarked.

Similarly, for table-processing and grid-processing functions, benchmarking `pygmt.select` and `pygmt.grdfill` should be enough.

.github/workflows/benchmarks.yml
weiji14: Done at 31eb3da573c85c35d4429e5debd3f48c1714c23d.

weiji14:
> Need to mention the new workflow in:
> 
>     * `doc/maintenance.md`
>     * `.github/ISSUE_TEMPLATE/bump_gmt_checklist.md`

Ok, done at c92fcb27597d4a4a803de2c9fb1fe08d75491699 and 7bf09b9f9c7bf20ebace952542ed8e339bd3986f

>     * `.github/ISSUE_TEMPLATE/release_checklist.md` (maybe unnecessary in this file)

Probably not needed in the checklist, but I added a trigger at c8d19656a4d5396b44ab48103eab2bbb52f02b91 so that the benchmarks will be run when a release is published.

> > So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?
> 
> I think we should focus on benchmarking low-level functions rather than modules' wrappers, since the low-level functions are heavily used in everywhere and most wrappers have very simple and similar code structures.

Let me open up a separate issue to discuss this, and also to track which unit tests we should benchmark.

.github/workflows/benchmarks.yml
weiji14: We need to make sure that conda's python (`/usr/share/miniconda/bin/python`) is used though instead of the system python (`/usr/bin/python`). Is there a way to prepend to $PATH instead of append?

seisman: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
prepending is the default.

weiji14: Ah nice, I see this being used at https://github.com/actions/starter-workflows/blob/c31fe3d5d44d7cb4c912f4c3213f7b4610f13ea2/ci/python-package-conda.yml#L19-L20 too, but assumed it was appending. Done at f45b727ce0c389d022398314a84464b4977cb330, but I might make another commit to use the `make ...` commands instead of `python -m ...`.

.github/workflows/benchmarks.yml
weiji14: Actually, how about we set [`paths:`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) so that the benchmark runs when any code in clib/datasets/helpers/src is modified:
```suggestion
  pull_request:
    paths:
      - 'pygmt/clib/**'
      - 'pygmt/datasets/**'
      - 'pygmt/helpers/**'
      - 'pygmt/src/**'
```

seisman: Why not just use `- 'pygmt/**/*.py'` to include all files?

weiji14: That won't work right if we want to include other files, e.g. `environment.yml` when bumping to new versions of GMT for example.

weiji14: Just re-read https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths, specifically this part:

> The order that you define paths patterns matters:
>
>  - A matching negative pattern (prefixed with !) after a positive match will exclude the path.
>  - A matching positive pattern after a negative match will include the path again.

So it looks like we can add more paths after `!pygmt/tests/**`.

weiji14: Ok, made a commit in dedfa7af7250762f473a81928a609b04198d11b1. Have applied these `paths:` restrictions to both the `push` and `pull_request` triggers, see if that looks ok.

seisman (approved):
Looks good to me.

weiji14:
Thanks for reviewing @seisman! I'll merge this now and open a follow-up PR for marking more unit tests with `@pytest.mark.benchmark`. We can discuss more in #2910 too.

seisman:
The Tests workflow (https://github.com/GenericMappingTools/pygmt/actions/runs/7319763038/job/19937998567) now has the following warnings:
```
  /home/runner/work/pygmt/pygmt/pygmt/tests/test_basemap.py:8: PytestUnknownMarkWarning: Unknown pytest.mark.benchmark - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.benchmark
```
I believe we need to silent these warnings.

weiji14:
> The Tests workflow (https://github.com/GenericMappingTools/pygmt/actions/runs/7319763038/job/19937998567) now has the following warnings because `pytest-benchmark` is not installed in these workflows:

Ok, opened #2912 for this.

weiji14:
By the way, can you change the setting for CodSpeed to only make PR comments when there is a performance improvement or regression (still don't have permissions)? This should reduce noise like https://github.com/GenericMappingTools/pygmt/pull/2907#issuecomment-1868838318 when there's no noticeable change (we can still see the report in the 'Checks' tab if needed).

image

Also, maybe reduce the regression threshold to 5%?

seisman:
> By the way, can you change the [CodSpeed settings](https://codspeed.io/GenericMappingTools/pygmt/settings) to only make PR comments when there is a performance improvement or regression (still don't have permissions)? This should reduce noise like [#2907 (comment)](https://github.com/GenericMappingTools/pygmt/pull/2907#issuecomment-1868838318) when there's no noticeable change (we can still see the report in the 'Checks' tab if needed).

Done.

PR #2911, opened 2023-12-25
Mark unit tests with @pytest.mark.benchmark part 1

### LINKED ISSUE #2910: Benchmark performance of PyGMT functions

We are attempting some big refactoring steps in PyGMT to avoid the use of temporary intermediate files #2730, and will also be performing some updates to GMT 6.5.0 soon, so it would be good to track any performance improvements or regressions in terms of execution speed. Since #835, we have actually measured the execution time of the slowest tests (>0.2s) on CI for #584, but those execution times are not really tracked over time, and only hidden in the CI log files.

So, to better track performance over time, we are setting up a Continuous Benchmarking workflow in #2908. This is using [`pytest-codspeed`](https://github.com/CodSpeedHQ/pytest-codspeed), with the results logged to https://codspeed.io/GenericMappingTools/pygmt. The benchmarks is selective however, and will only be run on unit tests marked with [`@pytest.mark.benchmark`](https://github.com/ionelmc/pytest-benchmark).

Main question: Which tests do we decide to benchmark?

> > So which tests should we benchmark (and should we mark all those tests with `@pytest.mark.benchmark` in this PR or a follow-up one)?
> 
> I think we should focus on benchmarking low-level functions rather than modules' wrappers, since the low-level functions are heavily used in everywhere and most wrappers have very simple and similar code structures. 
> 
> For example, most plotting modules have very simple code like the one in `Figure.basemap`
> ```
>     kwargs = self._preprocess(**kwargs)                                         
>     with Session() as lib:                                                      
>         lib.call_module(module="basemap", args=build_arg_string(kwargs))
> ```
> so we just need to benchmark one `basemap` test (e.g., `test_basemap()`) and don't need to benchmark other `basemap`'s tests and other plotting methods (e.g., `Figure.coast`). Of course, there are few exceptions, for example, `Figure.meca`, `Figure.plot`, `Figure.plot3d`, `Figure.text` are among the complicated wrappers and should be benchmarked.
> 
> Similarly, for table-processing and grid-processing functions, benchmarking `pygmt.select` and `pygmt.grdfill` should be enough.

_Originally posted by @seisman in https://github.com/GenericMappingTools/pygmt/issues/2908#issuecomment-1868514665_
            
Other questions: When do we want to run the benchmarks? Should this be enabled on every Pull Request, or just Pull Requests that modify certain files perhaps? Edit: Answered in https://github.com/GenericMappingTools/pygmt/pull/2908#discussion_r1435894337

Yes

weiji14:

Follow-up to #2908, this is where we mark the unit tests for any PyGMT functions we want to benchmark with the [`@pytest.mark.benchmark`](https://github.com/ionelmc/pytest-benchmark) decorator.

See https://github.com/GenericMappingTools/pygmt/pull/2911#issuecomment-1868800856 below on which tests have been marked for benchmarking. Will first prioritize PyGMT functions that use temporary files, or are commonly used ones. General discussion on which tests to mark can go in #2910.

Note: We have purposely not installed `pytest-mpl` in `benchmarks.yml`, so the benchmarks are ran without any image comparison. This allows for faster benchmark runs, and it also means we don't need to setup `dvc` in the workflow.

Part 1 of addressing #2910. Covers test functions from `test_geopandas` to `test_xyz2grd` according to the list in https://github.com/GenericMappingTools/pygmt/issues/2910#issuecomment-1869852471.

**Reminders**

**Slash Commands**

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -61,7 +61,7 @@ jobs:
           echo $CONDA/bin >> $GITHUB_PATH
           conda install --solver=libmamba gmt=6.4.0 python=3.12 \
                         numpy pandas xarray netCDF4 packaging \
-                        pytest pytest-benchmark pytest-mpl
+                        geopandas pytest pytest-benchmark pytest-mpl
           python -m pip install -U pytest-codspeed setuptools
 
       # Install the package that we want to test
--- a/pygmt/tests/test_geopandas.py
+++ b/pygmt/tests/test_geopandas.py
@@ -61,6 +61,7 @@ def fixture_gdf_ridge():
     return gdf
 
 
+@pytest.mark.benchmark
 def test_geopandas_info_geodataframe(gdf):
     """
     Check that info can return the bounding box region from a
--- a/pygmt/tests/test_grd2cpt.py
+++ b/pygmt/tests/test_grd2cpt.py
@@ -18,6 +18,7 @@ def fixture_grid():
     return load_static_earth_relief()
 
 
+@pytest.mark.benchmark
 @pytest.mark.mpl_image_compare
 def test_grd2cpt(grid):
     """
--- a/pygmt/tests/test_grd2xyz.py
+++ b/pygmt/tests/test_grd2xyz.py
@@ -20,6 +20,7 @@ def fixture_grid():
     return load_static_earth_relief()
 
 
+@pytest.mark.benchmark
 def test_grd2xyz(grid):
     """
     Make sure grd2xyz works as expected.
--- a/pygmt/tests/test_grdclip.py
+++ b/pygmt/tests/test_grdclip.py
@@ -55,6 +55,7 @@ def test_grdclip_outgrid(grid, expected_grid):
         xr.testing.assert_allclose(a=temp_grid, b=expected_grid)
 
 
+@pytest.mark.benchmark
 def test_grdclip_no_outgrid(grid, expected_grid):
     """
     Test the below and above parameters for grdclip with no set outgrid.
--- a/pygmt/tests/test_grdcontour.py
+++ b/pygmt/tests/test_grdcontour.py
@@ -33,6 +33,7 @@ def test_grdcontour(grid):
     return fig
 
 
+@pytest.mark.benchmark
 @pytest.mark.mpl_image_compare
 def test_grdcontour_labels(grid):
     """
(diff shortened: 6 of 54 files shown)


seisman:
Can we disable the comments like https://github.com/GenericMappingTools/pygmt/pull/2911#issuecomment-1868800856? I feel it's annoying since most PRs have no effects on the performances.

weiji14:
> Can we disable the comments like https://github.com/GenericMappingTools/pygmt/pull/2911#issuecomment-1868800856? I feel it's annoying since most PRs have no effects on the performances.

Yep, you just did it at https://github.com/GenericMappingTools/pygmt/pull/2908#issuecomment-1868843827!

seisman (approved):

weiji14:
Ok, have added `@pytest.mark.benchmark` to the second half of the list mentioned in https://github.com/GenericMappingTools/pygmt/issues/2910#issuecomment-1869852471, except these ones:

- [ ] test_helpers.py - too many random functions, unsure which one to benchmark
- [ ] test_image.py - requires pulling test_logo.py from DVC
- [ ] test_init.py - only tests `pygmt.show_versions()`, should we benchmark it since it's not really something that needs to have fast performance?
- [ ] test_sphinx_gallery.py - requires sphinx_gallery to be installed, plus `PyGMTScraper` isn't really user facing code.

weiji14:
Ok, I'm seeing some slow benchmark tests taking >1min because we didn't download the cache, and there are lots of PytestReturnNotNoneWarnings because `pytest-mpl` isn't installed (see https://github.com/pytest-dev/pytest/issues/10427). Gonna merge this first though and improve the benchmarks.yml CI workflow in a separate PR.

seisman:
> * [ ]  test_helpers.py - too many random functions, unsure which ones need to be benchmarked

None of them are important, so no need to benchmark now.

> * [ ]  test_image.py - requires pulling test_logo.py from DVC

Yes, it's very annoying. In the `ci_tests_legacy.yaml` workflow, we also have to setup dvc due to this test. Perhaps we should use an image file in the GMT cache instead.

> * [ ]  test_init.py - only tests `pygmt.show_versions()`, should we benchmark it since it's not really something that needs to have fast performance?

OK to ignore it.

> * [ ]  test_sphinx_gallery.py - requires sphinx_gallery to be installed, plus `PyGMTScraper` isn't really user facing code.

OK to ignore it.

weiji14:
> > * test_image.py - requires pulling test_logo.py from DVC
> 
> Yes, it's very annoying. In the `ci_tests_legacy.yaml` workflow, we also have to setup dvc due to this test. Perhaps we should use an image file in the GMT cache instead.

Yeah, it seems like we can use a remote file, e.g. `fig.image("@circuit.png", ...)`. Let me open a PR.

PR #2917, opened 2023-12-26
Refactor the internal _load_remote_dataset function to simplify datasets' definitions

seisman:
For each GMT remote dataset, we need to define it like:
```python
    "earth_age": GMTRemoteDataset(
        title="seafloor age",
        name="seafloor_age",
        long_name="age of seafloor crust",
        units="Myr",
        extra_attributes={"horizontal_datum": "WGS84"},
        resolutions={
            "01d": Resolution(["gridline", "pixel"], False),
            "30m": Resolution(["gridline", "pixel"], False),
            "20m": Resolution(["gridline", "pixel"], False),
            "15m": Resolution(["gridline", "pixel"], False),
            "10m": Resolution(["gridline", "pixel"], False),
            "06m": Resolution(["gridline", "pixel"], False),
            "05m": Resolution(["gridline", "pixel"], True),
            "04m": Resolution(["gridline", "pixel"], True),
            "03m": Resolution(["gridline", "pixel"], True),
            "02m": Resolution(["gridline", "pixel"], True),
            "01m": Resolution(["gridline"], True),
        },
    ),
```
The `resolutions` property is a dictionary of available resolutions and the corresponding registrations and tile information. As you can see, `["gridline", "pixel"]` is duplicated multiple times. I feel maintaining such a dictionary is tedious.

Since most resolutions support both "gridline" and "pixel" registrations, it makes sense to let `Resolution.registrations` have a default value `["gridline", "pixel"]`. Similarly, `Resolution.tiled` can have a default value `False`.

Then, the dataset definition can be written as:
```python
    "earth_age": GMTRemoteDataset(
        title="seafloor age",
        name="seafloor_age",
        long_name="age of seafloor crust",
        units="Myr",
        extra_attributes={"horizontal_datum": "WGS84"},
        resolutions={
            "01d": Resolution(),
            "30m": Resolution(),
            "20m": Resolution(),
            "15m": Resolution(),
            "10m": Resolution(),
            "06m": Resolution(),
            "05m": Resolution(tiled=True),
            "04m": Resolution(tiled=True),
            "03m": Resolution(tiled=True),
            "02m": Resolution(tiled=True),
            "01m": Resolution(registrations=["gridline"], tiled=True),
        },
    ),
```
The new function definition looks more compact, but an entry like `"01d": Resolution()` looks weird, since it may be unclear what `Resolution()` means.

Thus, I add the new `code` property to the `Resolution` class, then a `Resolution` object can be defined like
```
Resolution(code="01d", registrations=["gridline", "pixel"], tiled=False)
```
or the shortest version:
```
Resolution("01d")
```
After the above changes, entries like `"01d": Resolution("01d")` are still weird since the resolution code (e.g., `01d`) must be duplicated twice. Thus, I changed `resolutions` from a dict to a list.

Here is the final version of the dataset definition:
```python
    "earth_age": GMTRemoteDataset(
        title="seafloor age",
        name="seafloor_age",
        long_name="age of seafloor crust",
        units="Myr",
        extra_attributes={"horizontal_datum": "WGS84"},
        resolutions=[
            Resolution("01d"),
            Resolution("30m"),
            Resolution("20m"),
            Resolution("15m"),
            Resolution("10m"),
            Resolution("06m"),
            Resolution("05m", tiled=True),
            Resolution("04m", tiled=True),
            Resolution("03m", tiled=True),
            Resolution("02m", tiled=True),
            Resolution("01m", registrations=["gridline"], tiled=True),
        ]
    )
```
which I think is more compact and more readable.


.github/workflows/benchmarks.yml
```diff
@@ -13,7 +13,7 @@ on:
       - 'pygmt/**/*.py'
       - '.github/workflows/benchmarks.yml'
   # Uncomment the 'pull_request' line below to trigger the workflow in PR
-  # pull_request:
+  pull_request:
```
seisman, line 16:
```suggestion
  # pull_request:
```

pygmt/datasets/load_remote_dataset.py
```diff
@@ -3,32 +3,35 @@
...
 class Resolution(NamedTuple):
     """
-    The available grid registrations for a given resolution and whether it is a tiled
-    grid.
+    Resolution code, the available grid registrations and whether it is tiled.
 
     Attributes
     ----------
+    code : str
+        The resolution code.
```
weiji14, line 24: Resolution 'code' sounds a bit weird to me, but ok since it is used at https://www.generic-mapping-tools.org/remote-datasets/earth-gebco.html#usage. Just add some examples to make it clear.

```suggestion
        The resolution code. E.g. '01d', '30m', '01s'.
```

pygmt/datasets/load_remote_dataset.py
```diff
@@ -269,43 +266,44 @@ def _load_remote_dataset(
...
 
     # check resolution
-    if resolution not in dataset.resolutions:
+    for res in dataset.resolutions:
+        if res.code == resolution:
+            valid_registrations = res.registrations
+            is_tiled = res.tiled
+            break
+    else:
         raise GMTInvalidInput(f"Invalid resolution '{resolution}'.")
```
weiji14, line 281: This for-loop seems a lot more inefficient compared to the previous `if resolution not in dataset.resolutions` dictionary-based lookup. I almost think we should go with the intermediate `{"01d": Resolution(), ...}` dictionary you mentioned at https://github.com/GenericMappingTools/pygmt/pull/2917#issue-2056550622, or come up with a better data structure than a list of `Resolution` NamedTuples. I almost feel like that `datasets` variable could be a nested JSON or a multi-level pandas.DataFrame object.

weiji14: Alternatively, we could just keep the `datasets` variable as is, but bring in all the type hint stuff. The remote datasets aren't really updated that often, though I know you're working on those Moon/Venus/Mars PRs.

seisman: > This for-loop seems a lot more inefficient compared to the previous `if resolution not in dataset.resolutions` dictionary-based lookup. 

Yes, it's almost 10-times slower.

> I almost think we should go with the intermediate `{"01d": Resolution(), ...}` dictionary you mentioned at [#2917 (comment)](https://github.com/GenericMappingTools/pygmt/pull/2917#issue-2056550622), or come up with a better data structure than a list of `Resolution` NamedTuples. I almost feel like that `datasets` variable could be a nested JSON or a multi-level pandas.DataFrame object (but maybe that's overkill).

What about a dictionary:
```
{
   "01d": Resolution("01d"),
}
```
It's more clear than:
```
{
    "01d": Resolution(),
}
```

weiji14: > What about a dictionary:\n\n{\n   \"01d\": Resolution(\"01d\"),\n}

Yes, that could work.

PR #2925, opened 2023-12-27
Enable ruff's pydocstyle (D) rules and remove docformatter

### LINKED ISSUE #962: Change the line length limit of docstrings to 88 characters
We use the [Black](https://github.com/psf/black) tool to format the code, and the default line length of Black is [88 characters](https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length). The 88-char line length limit is only applied to code, not docstrings. For docstrings, the line length limit is 79 characters, and we use several tools, [blackdoc](https://github.com/keewis/blackdoc), [docformatter](https://github.com/myint/docformatter) and [flake8](https://flake8.pycqa.org/en/latest/) to check and format the docstrings.

Then, the question is, why do we use different line length limits for code and docstrings? 

We started to adopt the 79-char line length limit for docstrings in PR #384. The PR description also explains why we choose 79-char length (also see some external discussions in https://github.com/fatiando/verde/issues/177)

> Jupyter, IPython, and other IDEs don't usually render the rst in the
docstrings and just show them literally. The problem is that many of
these break lines at 79 characters. So docstrings with more characters
look terrible in these settings and sometimes almost unreadable. Wrap
all docstrings to 79 characters instead of Black's 88. 

---

Is the reason still valid?

I just changed one line of the docstrings of `Figure.basemap()` to ~120 chars, and checked what the docstrings look like in Jupyter, IPython, PyCharm, and VS Code. Here are the results:

In Jupyter Notebook, the long docstrings look good with the `help()` function:

image

Also look good in the Notebook Contextual Help page:

image

In a Python or IPython console:

image

In VS Code:

image

In PyCharm:

image

----

From the above screenshots, it's clear that these consoles, editors, and IDEs don't wrap docstrings at 80 characters and work well with long docstrings.

Now the questions are:

1. Are they any other common editors or IDEs that force wrapping docstrings at 80 characters?
2. Should we change the line length limit of docstrings to 88 characters?

seisman:
**Better to review the commits one by one to understand the changes.**

In b9c4d42899c662fff6efcf78a9ecab8df87b87df, we enable ruff's [pydocstyle (D) rules](https://docs.astral.sh/ruff/rules/#pydocstyle-d) and set the docstring convention to "numpy".

After setting the "numpy" convention, we have to fix/ignore some violations (done in fa93c3799d0b9602ad083b047cb040618305b8a1, 092075a1ddfcbe539ab84309ca1996c1ec4c4881, 608f38da9c88497f3d175d089107c7bc91f1dbde, 0f9cb0b2de65de2914f3d440ada7b76b30b623d7, d9116bd240a93c10f5a27be33cd66e9b21a18f6c, f22bfa0a0a3ba43c3e982ab37a5045c7965a9b02, 1280a1c4869669aeb9bdb161b6a4e5b0474099b4).

After the above fixes, we still have hundreds of violations against 6 rules, as shown below:
```
$ ruff check pygmt doc/conf.py examples --statistics
645	D200	[*] One-line docstring should fit on one line
312	D205	[ ] 1 blank line required between summary line and description
123	D400	[ ] First line should end with a period
 30	D401	[ ] First line of docstring should be in imperative mood: "A mock GMT API function that always returns a given value."
 30	D412	[*] No blank lines allowed between a section header and its content ("Examples")
 12	D202	[*] No blank lines allowed after function docstring (found 1)
```
- [`D200`](https://docs.astral.sh/ruff/rules/fits-on-one-line/): This is against our current style, so it must be disabled.
- [D205](https://docs.astral.sh/ruff/rules/blank-line-after-summary/): This rule makes sense. It seems [PEP 257](https://peps.python.org/pep-0257/) recommends that the summary line should be one line, but we actually have some multi-line summary, which causes lots of violations. We have to disable the rule now and may enable it in the future.
- [D400](https://docs.astral.sh/ruff/rules/ends-in-period/): Makes sense but too many violations. Disable it now to avoid too many changes in this PR.
- [D401](https://docs.astral.sh/ruff/rules/non-imperative-mood/): Not always make sense, so disabled.
- [D412](https://docs.astral.sh/ruff/rules/blank-lines-between-header-and-content/): Makes sense, but have 30 violations. Disabled to make the changes in this PR small, but I'm also OK with enabling it.
- [D202](https://docs.astral.sh/ruff/rules/no-blank-line-after-function/): Makes sense but have 12 violations. Disabled to make the changes in this PR small, but I'm also OK with enabling it.

In commit 31704013648147e55220c816c6f9c36329fc1e04, these 6 rules are disabled.

As the documentation says (https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings):

> The NumPy convention includes all D errors apart from: [D107](https://docs.astral.sh/ruff/rules/undocumented-public-init/), [D203](https://docs.astral.sh/ruff/rules/one-blank-line-before-class/), [D212](https://docs.astral.sh/ruff/rules/multi-line-summary-first-line/), [D213](https://docs.astral.sh/ruff/rules/multi-line-summary-second-line/), [D402](https://docs.astral.sh/ruff/rules/no-signature/), [D413](https://docs.astral.sh/ruff/rules/no-blank-line-after-section/), [D415](https://docs.astral.sh/ruff/rules/ends-in-punctuation/), [D416](https://docs.astral.sh/ruff/rules/section-name-ends-in-colon/), and [D417](https://docs.astral.sh/ruff/rules/undocumented-param/).

Among these rules, `D213` and `D410` make sense to the project and are enabled in commit 5d983e633bffe46b324fc55f2c4ebf7880f7d3d2.

Then docformatter can be fully replaced by ruff and is removed in commit 435290f8f3d1e3234c41b0f1d4e7e9741e126e91. The contributing guides are updated in 6e04543ebdfed6122e8e454c73f0941aa9c9abca.

Please also note that, in the old settings, docformatter wraps docstrings at 79 characters. If we change the docformatting settings to 88, then the summary lines can be automatically rewrapped to 88 characters, as mentioned in https://github.com/GenericMappingTools/pygmt/issues/962#issuecomment-1868849496. However, ruff's pydocstyle rules can't wrap long lines and only raise warnings. So, ruff is not as powerful as docformatter. In other words, we may want to change docformatter line-length settings to 88 (as done in https://github.com/GenericMappingTools/pygmt/issues/962#issuecomment-1868849496) and let docformatter rewrap lines to 88 characters before completely removing it (Done in #2926).

Closes #962.


pygmt/helpers/decorators.py
```diff
@@ -445,7 +445,7 @@ def fmt_docstring(module_func):
     - J = projection
     - R = region
     <BLANKLINE>
-    """
+    """  # noqa: D410,D411
```
weiji14, line 448: Since [D410](https://docs.astral.sh/ruff/rules/no-blank-line-after-section) (a blank line after section headings) is enabled in 5d983e633bffe46b324fc55f2c4ebf7880f7d3d2, this `noqa: D410, D411` can be removed (and you'll need to run formatting after).

pygmt/src/config.py
```diff
@@ -203,10 +203,15 @@ def __init__(self, **kwargs):
...
     def __enter__(self):
+        """
+        Do nothing but return the object.
+        """
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        # revert to initial values
+        """
+        Revert the configurations to initial values.
```
weiji14, line 213:
```suggestion
        Revert GMT configurations to initial values.
```

pygmt/helpers/decorators.py
seisman: Actually, this line is required. If removing `noqa: D410, D411`, the codes will be formatted with the following diff:
```diff
diff --git a/pygmt/helpers/decorators.py b/pygmt/helpers/decorators.py
index 57bd0f900..b44dd7fec 100644
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -425,6 +425,7 @@ def fmt_docstring(module_func):
     <BLANKLINE>
     My nice module.
     <BLANKLINE>
+
     Parameters
     ----------
     data : str, numpy.ndarray, pandas.DataFrame, xarray.Dataset, or geo...
@@ -445,7 +446,7 @@ def fmt_docstring(module_func):
     - J = projection
     - R = region
     <BLANKLINE>
-    """  # noqa: D410,D411
+    """
     filler_text = {}

     if hasattr(module_func, "aliases"):
```
Then the doctest fails, because only one blank line is expected, but two blank lines are given (one is `<BLANKLINE>`, another one is the newly added blank line).

weiji14: We could just remove the `<BLANKLINE>` from L427 no?

weiji14: Yeah, looks like we'll need to keep it, since doctest takes an empty line to mean the end of the expected output according to https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized

weiji14 (approved):
Thanks again @seisman for all the work simplifying the code formatters to just `ruff`! Now `make format` runs in less than a second :rocket:

Issue #217, opened 2018-07-25
Multiprocessing


norfordb:
**-- TEMPLATE FOR FEATURE REQUESTS AND FEEDBACK --**

## Feedback or description of feature requested

*Your text goes here. Please be as specific and detailed as possible.*
When attempting to call a script using gmt-python by way of the python multiprocessing library (multiprocessing.pool), an error is returned, as shown below.

**Are you willing to help implement and maintain this feature?** No

**-- TEMPLATE FOR BUG REPORTS --** If you're not filing a bug, please delete this part.

## Full code that generated the error

```python
import gmt
import multiprocessing as mp

def gmt_fun(n): #a very simple function that uses gmt-python
	fig = gmt.Figure()
	fig.coast(R="-20/20/-20/20", J="M10i", G='gray', S="lightblue", B=True) 
	fig.savefig("{}.pdf".format(n), show=False)

#gmt_fun(5) #calls the function without multiprocessing - uncomment as desired to see that the function works on its own

with mp.Pool(2) as p: #calls the process with multiprocessing, using 2 cores
	a = p.map(gmt_fun,[x for x in range(0,2)])
```

The code does work with the following change:
```
import multiprocessing as mp

def gmt_fun(n): #a very simple function that uses gmt-python
	import gmt
	fig = gmt.Figure()
	fig.coast(R="-20/20/-20/20", J="M10i", G='gray', S="lightblue", B=True) 
	fig.savefig("{}.pdf".format(n), show=False)

#gmt_fun(5) #calls the function without multiprocessing - uncomment as desired to see that the function works on its own

with mp.Pool(2) as p: #calls the process with multiprocessing, using 2 cores
	a = p.map(gmt_fun,[x for x in range(0,2)])
```

## Full error message

```
(gmt-python) mac97214:models norfordb$ python tweaker.py
gmt-python-session [ERROR]: Not available in classic mode
gmt-python-session [ERROR]: Not available in classic mode
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/Users/norfordb/anaconda2/envs/gmt-python/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/Users/norfordb/anaconda2/envs/gmt-python/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "tweaker.py", line 6, in gmt_fun
[...]

PR #2924, opened 2023-12-27
Mark unit tests with @pytest.mark.benchmark part 2

weiji14:
**Description of proposed changes**

Follow-up to https://github.com/GenericMappingTools/pygmt/pull/2908, this is where we mark the unit tests for any PyGMT functions we want to benchmark with the [@pytest.mark.benchmark](https://github.com/ionelmc/pytest-benchmark) decorator.

Benchmark report at https://codspeed.io/GenericMappingTools/pygmt/branches/pytest/mark-benchmark-part2

See https://github.com/GenericMappingTools/pygmt/pull/2924#issuecomment-1870085928 below on which new tests have been marked for benchmarking. Will prioritize PyGMT functions that use temporary files, and low-level clib functions that should be benchmarked. General discussion on which tests to mark can go in https://github.com/GenericMappingTools/pygmt/issues/2910.

Part 2 of addressing https://github.com/GenericMappingTools/pygmt/issues/2910. Covers test functions from test_binstats to test_filter1d according to the list in https://github.com/GenericMappingTools/pygmt/issues/2910#issuecomment-1869852471.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

Issue #1582, opened 2021-10-13
Strange behaviour when using Session.call_module().


timrlaw:
**Description of the problem**

Hi all. I'm having issues related to sessions, and the order in which seemingly unrelated commands are issued affecting the final plot in strange ways. I think this is related to #1242 and #733.

Essentially what I'm trying to do is set the dimensions of an inset figure to match the dimensions of the inset map. Perhaps there is a better way to do this, but the way I've found is to use GMT's `mapproject` to calculate the height given the width. PyGMT doesn't expose `mapproject` so I have to use `Session.call_module()`, see the code below.

The problem is that doing this results in strange behaviour, dependent on where exactly the `mapproject` call is issued. I have found a way to arrange the code to get the figure I want, but it is very fragile, and the inset options do not work correctly.

**Full code that generated the error**

This is the "working" sample:

```python
import pygmt
from pygmt.clib import Session
from pygmt.helpers import GMTTempFile


def mapproject_height(region, projection):
    with GMTTempFile() as tmp:
        args = f"-R{region} -J{projection} -Wh ->{tmp.name}"
        with Session() as lib:
            lib.call_module("mapproject", args)
        h = tmp.read().strip()

    return h


region = [-51, -48, -24, -23]
inset_region = [-80, -28, -43, 0]
inset_w = 2

fig = pygmt.Figure()

grid = pygmt.datasets.load_earth_relief(resolution="03s", region=region)
fig.grdimage(grid=grid, projection="M15c", frame=True, cmap="world")

fig.colorbar(frame=["x", "y+lm"], position="JBC+o0c/0.6c+w4.8c")

inset_h = mapproject_height("80W/28W/43S/0S", "M2c")
with fig.inset(position=f"jBL+w{inset_w}c/{inset_h}c+o0.1c", box="+pblack"):
    fig.coast(region=inset_region, projection=f"M{inset_w}c",
            land="gray", shorelines="thin", water="white")

fig.savefig("test.png")
```

Issues arise when trying to:

- Plot the colorbar after the inset---it is positioned relative to the inset rather than the main figure despite being outside the `with` statement.
- Move the inset to another corner (e.g. TL instead of BL). This results in the inset not being positioned correctly.
- Moving the `mapproject_height` call to the beginning, causes the outer figure to use the region boundaries from the inset for some reason.

[...]

Issue #1242, opened 2021-04-27
How to resolve flaky tests resulting from using a single GMT session


weiji14:
**Description of the problem**

There's been instances of [flaky](https://docs.pytest.org/en/stable/flaky.html) tests in PyGMT's test suite reported in https://github.com/GenericMappingTools/pygmt/issues/1217#issuecomment-825561365. This likely stems from the fact that PyGMT uses a single GMT session (initiated during `import pygmt`) instead of separate GMT sessions for each figure (see https://github.com/GenericMappingTools/pygmt/pull/327#issuecomment-541782890).

@meghanrjones asked about whether we should stick with using a single GMT session in https://github.com/GenericMappingTools/pygmt/issues/1217#issuecomment-827230998, or use independent sessions per figure

> I understand the original logic behind a single GMT session for all tests in https://github.com/GenericMappingTools/pygmt/pull/327#issuecomment-541782890. Still, I don't expect that users will be attempting to use the entire PyGMT library in a single session, which is the goal of the test suite. So I think it would be worth revisiting this decision. Could it be possible to periodically test the examples/tutorials against baseline images to ensure that producing multiple plots in a single session is consistent and have the unit tests each use individual sessions?

**Full code that generated the error**

Flaky tests are hard to reproduce (that is their definition actually), but in PyGMT's case, can be found e.g. when a single test passing on `pytest pygmt/tests/test_somemodule.py` fails when ran using `make test`, or vice versa.

E.g. as reported by @meghanrjones in https://github.com/GenericMappingTools/pygmt/issues/1217#issuecomment-825753479

> edit: I have not yet been able to figure out a solution. The two makecpt tests fail if there is a docstring example that imports pygmt and instantiates a figure (e.g., `extract_region()` in `pygmt/clib/session.py` and `pygmt/src/grdfilter.py`) and is tested before `pygmt/tests/test_makecpt.py`.

Related issues affected by having a single GMT session:
- https://github.com/GenericMappingTools/pygmt/issues/372#issuecomment-551855032
- #733

**System information**

Please paste the output of `python -c "import pygmt; pygmt.show_versions()"`:

```
PyGMT information:
  version: v0.3.2.dev117+g7466dc31
System information:
  python: 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 16:07:37)  [GCC 9.3.0]
  executable: ~/username/miniconda3/envs/pygmt/bin/python
[...]

PR #2930, opened 2023-12-28
Improve performance by avoiding loading the GMT library repeatedly

### LINKED ISSUE #867: Improve how PyGMT loads the GMT library
A few thoughts on the design of loading the library as a whole (might be out of scope for this PR but worth mentioning):

* I initially made the `Session` class load the library when it's used to avoid doing this work at import time. Mostly so we wouldn't have `GMTLibNotFoundError` at import.
* That clearly didn't work because we need to call `begin` at import for this whole thing to work. 
* So it might be time to rethink loading libgmt in the class instead of as a global instead to avoid searching for it every time. 
* Or we figure out a way to not call `begin` at import time.

_Originally posted by @leouieda in https://github.com/GenericMappingTools/pygmt/issues/702#issuecomment-737314991_

seisman:

The [`load_libgmt`](https://github.com/GenericMappingTools/pygmt/blob/bb8234530af41ebcfa15ba4d516754ec2c633b45/pygmt/clib/loading.py#L18) function can search for the GMT library and load it so that we can call GMT API functions in PyGMT. The function is called in the [`Session` class](https://github.com/GenericMappingTools/pygmt/blob/bb8234530af41ebcfa15ba4d516754ec2c633b45/pygmt/clib/session.py#L91)'s [`get_libgmt_func`](https://github.com/GenericMappingTools/pygmt/blob/bb8234530af41ebcfa15ba4d516754ec2c633b45/pygmt/clib/session.py#L275) function.

Although the `get_libgmt_func` function is called multiple times in `Session`'s different methods, the GMT library is only loaded once for each session, as shown below:
https://github.com/GenericMappingTools/pygmt/blob/bb8234530af41ebcfa15ba4d516754ec2c633b45/pygmt/clib/session.py#L310-L311

However, when wrapping GMT modules, we always create a new session and then call the GMT module like this
```
with Session() as lib:
    lib.call_module(...)
```
It means the GMT library is searched and loaded multiple times, as reported in #867.

To verify if the GMT library is loaded multiple times, we just need to add a simple print statement like `print("Loading GMT library")` at the start of the `load_libgmt` function, and then call PyGMT. Here is what you'll see. It's clear that the GMT library are loaded multiple times:
```python
In [1]: import pygmt
Loading GMT library
Loading GMT library

In [2]: fig = pygmt.Figure()
Loading GMT library

In [3]: fig.basemap(region=[0, 10, 0, 10], frame=True, projection="X10c")
Loading GMT library
Loading GMT library

In [4]: fig.savefig("map.png")
Loading GMT library
Loading GMT library
```

It's unnecessary to search for and load the GMT library multiple times, which is actually time-consuming (120 ms for me on macOS).
```
In [1]: from pygmt.clib.loading import load_libgmt

In [2]: %timeit load_libgmt()
118 ms ± 5.75 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

This PR fixes the issue by moving the `load_libgmt` call outside the `Session` class, so that the GMT library is only searched for and loaded once when importing pygmt.

This PR addresses the following comment in PR #867 so it fixes #867:

> So it might be time to rethink loading libgmt in the class instead of as a global instead to avoid searching for it every time.

#867 also proposed another alternative solution:

> Or we figure out a way to not call begin at import time.

This solution may help address issue #217, but it's unclear what exactly we should do and it definitely needs big refactors, so we can explore the solution later. 

It's worth noting that the small changes in this PR **speed up our Tests significantly**! For example, the "Run tests" step now only takes **85 seconds** (on Linux) compared to **250 seconds** in the main branch.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -87,7 +87,7 @@ DTYPES = {
     np.datetime64: "GMT_DATETIME",
 }
 
-# Load the GMT library outside the Session class to avoid repeating loading.
+# Load the GMT library outside the Session class to avoid repeated loading.
 _libgmt = load_libgmt()
 
 
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -87,6 +87,9 @@ DTYPES = {
     np.datetime64: "GMT_DATETIME",
 }
 
+# load the GMT library outside the Session class so it's loaded once.
+_libgmt = load_libgmt()
+
 
 class Session:
     """
@@ -308,7 +311,7 @@ class Session:
         <class 'ctypes.CDLL.__init__.<locals>._FuncPtr'>
         """
         if not hasattr(self, "_libgmt"):
-            self._libgmt = load_libgmt()
+            self._libgmt = _libgmt
         function = getattr(self._libgmt, name)
         if argtypes is not None:
             function.argtypes = argtypes
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -87,7 +87,7 @@ DTYPES = {
     np.datetime64: "GMT_DATETIME",
 }
 
-# load the GMT library outside the Session class so it's loaded once.
+# Load the GMT library outside the Session class to avoid repeating loading.
 _libgmt = load_libgmt()
 
 
--- a/pygmt/tests/test_clib_loading.py
+++ b/pygmt/tests/test_clib_loading.py
@@ -11,6 +11,7 @@ from pathlib import PurePath
 
 import pytest
 from pygmt.clib.loading import check_libgmt, clib_full_names, clib_names, load_libgmt
+from pygmt.clib.session import Session
 from pygmt.exceptions import GMTCLibError, GMTCLibNotFoundError, GMTOSError
 
 
@@ -206,6 +207,45 @@ class TestLibgmtBrokenLibs:
         """
         lib_fullnames = [self.faked_libgmt1, self.faked_libgmt1, self.loaded_libgmt]
         assert check_libgmt(load_libgmt(lib_fullnames=lib_fullnames)) is None
+
+
+class TestLibgmtCount:
+    """
+    Test that the GMT library is not repeatly loaded in every session.
+    """
+
+    loaded_libgmt = load_libgmt()  # Load the GMT library and reuse it when necessary
+    counter = 0  # Count how many times ctypes.CDLL is called
+
+    def _mock_ctypes_cdll_return(self, libname):  # noqa: ARG002
+        """
+        Mock ctypes.CDLL to count how many times the function is called.
+
+        If ctypes.CDLL is called, the counter increases by one.
+        """
+        self.counter += 1  # Increase the counter
+        return self.loaded_libgmt
+
+    @pytest.fixture()
+    def _mock_ctypes(self, monkeypatch):
+        monkeypatch.setattr(ctypes, "CDLL", self._mock_ctypes_cdll_return)
+
+    @pytest.mark.usefixtures("_mock_ctypes")
+    def test_libgmt_load_counter(self):
+        """
+        Make sure that the GMT library is not loaded in every session.
+        """
+        with Session() as lib:
+            _ = lib
+        with Session() as lib:
+            _ = lib
+        assert self.counter == 0  # ctypes.CDLL is not called after two sessions.
+
+        # Explicitly calling load_libgmt to make sure the mock function is correct
+        load_libgmt()
+        assert self.counter == 1
+        load_libgmt()
+        assert self.counter == 2
 
 
 ###############################################################################
--- a/pygmt/tests/test_clib_loading.py
+++ b/pygmt/tests/test_clib_loading.py
@@ -215,7 +215,7 @@ class TestLibgmtCount:
     """
 
     loaded_libgmt = load_libgmt()  # Load the GMT library and reuse it when necessary
-    counter = 0  # Count how many times ctypes.CDLL is called
+    counter = 0  # Global counter for how many times ctypes.CDLL is called
 
     def _mock_ctypes_cdll_return(self, libname):  # noqa: ARG002
         """
@@ -226,15 +226,14 @@ class TestLibgmtCount:
         self.counter += 1  # Increase the counter
         return self.loaded_libgmt
 
-    @pytest.fixture()
-    def _mock_ctypes(self, monkeypatch):
-        monkeypatch.setattr(ctypes, "CDLL", self._mock_ctypes_cdll_return)
-
-    @pytest.mark.usefixtures("_mock_ctypes")
-    def test_libgmt_load_counter(self):
+    def test_libgmt_load_counter(self, monkeypatch):
         """
         Make sure that the GMT library is not loaded in every session.
         """
+        # Monkeypatch the ctypes.CDLL function
+        monkeypatch.setattr(ctypes, "CDLL", self._mock_ctypes_cdll_return)
+
+        # Create two sessions and check the global counter
         with Session() as lib:
             _ = lib
         with Session() as lib:
--- a/pygmt/tests/test_clib_loading.py
+++ b/pygmt/tests/test_clib_loading.py
@@ -211,7 +211,7 @@ class TestLibgmtBrokenLibs:
 
 class TestLibgmtCount:
     """
-    Test that the GMT library is not repeatly loaded in every session.
+    Test that the GMT library is not repeatedly loaded in every session.
     """
 
     loaded_libgmt = load_libgmt()  # Load the GMT library and reuse it when necessary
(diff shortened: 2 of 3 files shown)


seisman:
I'm a liitle confused why we don't see some  improvements in the benchmark report above.

weiji14:
If you click into the link - https://codspeed.io/GenericMappingTools/pygmt/branches/clib/load-libgmt, it shows that most tests improve by 1-2%. Since most unit tests only call a few GMT commands, they shouldn't show too much of a gain, but still, I think it's worth it.

Maybe wait for #2924? There are a few clib/session related tests in that PR which might show a clearer improvement.

seisman:
> If you click into the link - https://codspeed.io/GenericMappingTools/pygmt/branches/clib/load-libgmt, it shows that most tests improve by 1-2%. Since most unit tests only call a few GMT commands, they shouldn't show too much of a gain, but still, I think it's worth it.
> 
> Maybe wait for #2924? There are a few clib/session related tests in that PR which might show a clearer improvement.

Just tried to benchmark locally. 

In the main branch:
```
In [1]: import pygmt

In [2]: fig = pygmt.Figure()

In [3]: %timeit fig.basemap(region=[0, 10, 0, 10], frame=True, projection="X10c")
237 ms ± 9.54 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
And in this branch:
```
In [1]: import pygmt

In [2]: fig = pygmt.Figure()

In [3]: %timeit fig.basemap(region=[0, 10, 0, 10], frame=True, projection="X10c")
6.12 ms ± 246 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
So, the `Figure.basemap` call is 6.12 ms vs 237 ms. The reduced time (~230 ms) is just the time used to load the GMT library twice. So I expect to see that even the simplest test should have a time reduction of at least 200 ms. But it doesn't match the CodSpeed reports.

seisman:
I've added a unit test in commit f589ff8b687c9fc35d516175ae731dcf1d64d490. The unit test passes in this branch and fails if in the main branch:

```
_______________________________ TestLibgmtCount.test_libgmt_load_counter _________________________________

self = <pygmt.tests.test_clib_loading.TestLibgmtCount object at 0x7f212fc958b0>

    @pytest.mark.usefixtures("_mock_ctypes")
    def test_libgmt_load_counter(self):
        """
        Make sure that the GMT library is not loaded in every session.
        """
        with Session() as lib:
            _ = lib
        with Session() as lib:
            _ = lib
>       assert self.counter == 0  # ctypes.CDLL is not called after two sessions.
E       assert 2 == 0
E        +  where 2 = <pygmt.tests.test_clib_loading.TestLibgmtCount object at 0x7f212fc958b0>.counter

pygmt/tests/test_clib_loading.py:242: AssertionError

```

pygmt/clib/session.py
```diff
@@ -87,6 +87,9 @@
     np.datetime64: "GMT_DATETIME",
 }
 
+# Load the GMT library outside the Session class to avoid repeating loading.
```
weiji14, line 90:
```suggestion
# Load the GMT library outside the Session class to avoid repeated loading.
```

pygmt/tests/test_clib_loading.py
```diff
@@ -208,6 +209,45 @@ def test_brokenlib_brokenlib_workinglib(self):
         assert check_libgmt(load_libgmt(lib_fullnames=lib_fullnames)) is None
 
 
+class TestLibgmtCount:
```
weiji14, line 212: Maybe better to put this unit test in `test_session_management.py`?

weiji14 (commented):
> This solution may help address issue https://github.com/GenericMappingTools/pygmt/issues/217, but it's unclear what exactly we should do and it definitely needs big refactors, so we can explore the solution later.

I can see the value of not loading GMT repeatedly in terms of speed, but I'm still a little unsure about making this a global variable. There are also issues like #1242 and #1582 where there has been flakiness with using a single GMT session, and we'll need to come up with some really good unit tests to make sure that there isn't any side effects from this change.

Could we make a unit test based on https://github.com/GenericMappingTools/pygmt/issues/217#issue-344245022, and test to see if the change in this PR allows for this multiprocessing/parallel code to work:

```python
import pygmt
import multiprocessing as mp

def gmt_func(n):
    fig = pygmt.Figure()
    fig.coast(...) 
    fig.show()

# gmt_func(n=5) #calls the function without multiprocessing - uncomment as desired to see that the function works on its own

with mp.Pool(2) as p: #calls the process with multiprocessing, using 2 cores
    a = p.map(gmt_func, [x for x in range(0,2)])
```

seisman:
> Could we make a unit test based on [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issue-344245022), and test to see if the change in this PR allows for this multiprocessing/parallel code to work:

No, it definitely won't work. To allow multiprocessing for PyGMT, we need to call `gmt begin` in each process. Currently, `gmt begin` is only called once when importing pygmt. That's why the trick in https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-754774875 works.

weiji14:
> > Could we make a unit test based on [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issue-344245022), and test to see if the change in this PR allows for this multiprocessing/parallel code to work:
> 
> No, it definitely won't work. To allow multiprocessing for PyGMT, we need to call `gmt begin` in each process. Currently, `gmt begin` is only called once when importing pygmt. That's why the trick in [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-754774875) works.

Hmm you're right. Is it ok for each process to have its own 'global' libgmt though? I suppose the changes here are only affecting things on the initial 'load GMT library' library, and not the 'GMT session' level, so it's probably ok.

pygmt/tests/test_clib_loading.py
```diff
@@ -208,6 +209,45 @@ def test_brokenlib_brokenlib_workinglib(self):
...
+    """
+    Test that the GMT library is not repeatly loaded in every session.
```
seisman, line 214:
```suggestion
    Test that the GMT library is not repeatedly loaded in every session.
```

seisman:
> Could we make a unit test based on [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issue-344245022), and test to see if the change in this PR allows for this multiprocessing/parallel code to work:
> 
> ```python
> import pygmt
> import multiprocessing as mp
> 
> def gmt_func(n):
>     fig = pygmt.Figure()
>     fig.coast(...) 
>     fig.show()
> 
> # gmt_func(n=5) #calls the function without multiprocessing - uncomment as desired to see that the function works on its own
> 
> with mp.Pool(2) as p: #calls the process with multiprocessing, using 2 cores
>     a = p.map(gmt_func, [x for x in range(0,2)])
> ```

In 48c00aee65d9a66ab94770ea6fb7bca6a1ece4d5, I've added a test to make sure that the workaround in #217 still works after this change. But, as mentioned in https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-762649874, the workaround doesn't work for Windows, so it's marked as `xfail` on Windows.

seisman:
> > Could we make a unit test based on [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issue-344245022), and test to see if the change in this PR allows for this multiprocessing/parallel code to work:
> > ```python
> > import pygmt
> > import multiprocessing as mp
> > 
> > def gmt_func(n):
> >     fig = pygmt.Figure()
> >     fig.coast(...) 
> >     fig.show()
> > 
> > # gmt_func(n=5) #calls the function without multiprocessing - uncomment as desired to see that the function works on its own
> > 
> > with mp.Pool(2) as p: #calls the process with multiprocessing, using 2 cores
> >     a = p.map(gmt_func, [x for x in range(0,2)])
> > ```
> 
> In [48c00ae](https://github.com/GenericMappingTools/pygmt/commit/48c00aee65d9a66ab94770ea6fb7bca6a1ece4d5), I've added a test to make sure that the workaround in #217 still works after this change. But, as mentioned in [#217 (comment)](https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-762649874), the workaround doesn't work for Windows, so it's marked as `xfail` on Windows.

I've decided to remove this multiprocessing test from this PR, and will add it in PR #2938 and also fix the Windows issue.

pygmt/tests/test_clib_loading.py
seisman: Prefer to keep it in `test_clib_loading.py` because this unit test is actually not related to session management.

weiji14: Ok, I kinda wanted to move the test since `test_clib_loading.py` has 300+ lines of code, while `test_session_management.py` has <100 lines, and we're kinda checking that `Session()` doesn't reload libgmt here, but up to you :slightly_smiling_face:

weiji14 (approved):

PR #2931, opened 2023-12-28
CI: Avoid pulling GMT baseline images in the "GMT Dev Tests" workflow

seisman:
The "GMT Dev Tests" workflow started to fail recently after PR #2913. See https://github.com/GenericMappingTools/pygmt/actions/runs/7333803702/job/19969730316 for the latest run.

It fails because the `dvc pull` command tries to pull the baseline images in both the PyGMT repository and the GMT repository, mainly because GMT is installed `${{ github.workspace }}/gmt-install-dir` (i.e., in the `pygmt` root directory). 

This PR fixes the issue by installing GMT to a temporary directory `${{ runner.temp }}/gmt-install-dir` and also delete the GMT source codes after installing GMT.

Now it works. See https://github.com/GenericMappingTools/pygmt/actions/runs/7349827937/job/20010428424?pr=2931

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -120,7 +120,7 @@ jobs:
         run: curl https://raw.githubusercontent.com/GenericMappingTools/gmt/master/ci/build-gmt.sh | bash
         env:
           GMT_GIT_REF: ${{ matrix.gmt_git_ref }}
-          GMT_INSTALL_DIR: ${{ github.workspace }}/gmt-install-dir
+          GMT_INSTALL_DIR: ${{ runner.temp }}/gmt-install-dir
         if: runner.os != 'Windows'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests_dev.yaml
```diff
@@ -139,13 +139,15 @@ jobs:
             -DGMT_USE_THREADS=TRUE
           cmake --build .
           cmake --build . --target install
+          cd ..
+          rm -rf gmt/
```
weiji14, line 143: Shouldn't be necessary to delete the gmt/ directory anymore right since we install to a temporary directory? Or is there something different on Windows?

seisman: it's not necessary but deleting it does no harm.

weiji14 (approved):

PR #2933, opened 2023-12-30
pyarrow: Check compatibility of pyarrow.array with string type

weiji14:

Check that [`pyarrow.array`](https://arrow.apache.org/docs/13.0/python/generated/pyarrow.array.html) objects of [string type](https://arrow.apache.org/docs/13.0/python/generated/pyarrow.string.html) can be passed to PyGMT modules/functions.

This PR also adds type hints to the docstring of the `pygmt.Figure.text`'s `text` parameter.

**Preview** at https://pygmt-dev--2933.org.readthedocs.build/en/2933/api/generated/pygmt.Figure.text.html

TODO:
- [x] Initial compatibility check of `pyarrow.array` by adding parametrized tests to `pygmt.text`
- [x] Add parametrized test to `test_clib_virtualfile_from_vectors.py`
- [x] Check that `pa.StringArray` can be passed into `pygmt.Figure.text`

Addresses https://github.com/GenericMappingTools/pygmt/issues/2800#issuecomment-1844400954, part of #2800.

**Reminders**

**Slash Commands**


pygmt/clib/conversion.py
```diff
@@ -286,7 +291,12 @@ def strings_to_ctypes_array(strings: Sequence[str]) -> ctp.Array:
     >>> [s.decode() for s in ctypes_array]
     ['first', 'second', 'third']
     """
-    return (ctp.c_char_p * len(strings))(*[s.encode() for s in strings])
+    try:
+        bytes_string_list = [s.encode() for s in strings]
+    except AttributeError:  # 'pyarrow.StringScalar' object has no attribute 'encode'
+        # Convert pyarrow.StringArray to Python list first
+        bytes_string_list = [s.encode() for s in strings.to_pylist()]
+    return (ctp.c_char_p * len(strings))(*bytes_string_list)
```
weiji14, line 299: Key part to review is the changes on the `strings_to_ctypes_array` function here (see also #3137). I'm converting the `pyarrow.StringArray` -> `list` -> `ctypes`, but wondering if there's a more efficient way to convert pyarrow to ctypes.

pygmt/clib/conversion.py
```diff
@@ -263,14 +267,15 @@ def sequence_to_ctypes_array(
     return (ctype * size)(*sequence)
 
 
-def strings_to_ctypes_array(strings: Sequence[str]) -> ctp.Array:
+def strings_to_ctypes_array(strings: Sequence[str] | pa.StringArray) -> ctp.Array:
```
weiji14, line 270: Need to figure out how to fix the `mypy` type errors for an optional import :thinking:

```
pygmt/clib/conversion.py:298: error: Item "Sequence[str]" of "Sequence[str] | Any" has no attribute "to_pylist"  [union-attr]
pygmt/clib/session.py:1004: error: Item "Sequence[str]" of "Sequence[str] | Any" has no attribute "dtype"  [union-attr]
```

pygmt/clib/conversion.py
```diff
@@ -9,6 +9,15 @@
 import numpy as np
 from pygmt.exceptions import GMTInvalidInput
 
+StringArrayTypes = Sequence[str]
```
seisman, line 12: I guess it's time to add the `pygmt/_typing.py` module and define `StringArrayTypes` in it, like what I'm doing in https://github.com/GenericMappingTools/pygmt/pull/3459/files

weiji14: Done at 17c1e9ca28acbb9f35223b0bb770a8dc8bb75843

pygmt/clib/conversion.py
```diff
@@ -304,7 +306,12 @@ def strings_to_ctypes_array(strings: Sequence[str]) -> ctp.Array:
     >>> [s.decode() for s in ctypes_array]
     ['first', 'second', 'third']
     """
-    return (ctp.c_char_p * len(strings))(*[s.encode() for s in strings])
+    try:
+        bytes_string_list = [s.encode() for s in strings]
+    except AttributeError:  # 'pyarrow.StringScalar' object has no attribute 'encode'
+        # Convert pyarrow.StringArray to Python list first
+        bytes_string_list = [s.encode() for s in strings.to_pylist()]  # type: ignore[union-attr]
+    return (ctp.c_char_p * len(strings))(*bytes_string_list)
```
seisman, line 314: What about using `np.asarray` to convert `strings` into a numpy array first?
```suggestion
    return (ctp.c_char_p * len(strings))(*[s.encode() for s in np.asarray(strings)])
```

Here are benchmarks:
```python
In [1]: import pyarrow as pa

In [2]: import numpy as np

In [3]: stringlist = ["hello", "world", "test"]

In [4]: stringarray = pa.array(stringlist)

In [5]: %timeit [s.encode() for s in stringlist]
196 ns ± 0.879 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit [s.encode() for s in stringarray.to_pylist()]
2.56 μs ± 14.8 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [7]: %timeit [s.encode() for s in np.asarray(stringlist)]
2.13 μs ± 22.3 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [8]: %timeit [s.encode() for s in np.asarray(stringarray)]
3.43 μs ± 19.9 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
For a much larger string array:
```python
In [19]: stringlist = ["hello", "world", "test"] * 1000

In [20]: stringarray = pa.array(stringlist)

In [21]: %timeit [s.encode() for s in stringlist]
115 μs ± 484 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [22]: %timeit [s.encode() for s in stringarray.to_pylist()]
2.17 ms ± 17.6 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [23]: %timeit [s.encode() for s in np.asarray(stringlist)]
1.07 ms ± 4.88 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [24]: %timeit [s.encode() for s in np.asarray(stringarray)]
226 μs ± 1.26 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
- **Pros:** consistent way to handle a sequence, a numpy array and pyarrow.StringDtype.
- **Cons**: 10x slower for string list, but I guess it's OK since it's very fast even for a large array

Also need to note that, currently `strings_to_ctypes_array` is only used in four functions, `put_vector`/`put_strings`/`virtualfile_from_stringio`/`call_module`.  `put_vector` and `put_string` are called in `virtualfile_from_vectors` only and before calling these two functions, the *array* has been converted to a numpy array by the `vectors_to_arrays` function at https://github.com/GenericMappingTools/pygmt/blob/96b12634d95a0fd2a262bcf5245a942cb4077e69/pygmt/clib/session.py#L1384.

So the performance reduction should be negligible:
```
In [1]: import numpy as np

In [2]: stringlist = ["hello", "world", "test"] * 1000

In [3]: stringarray = np.asarray(stringlist)

In [4]: %timeit [s.encode() for s in stringarray]
671 μs ± 13.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [5]: %timeit [s.encode() for s in np.asarray(stringarray)]
688 μs ± 3.3 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

weiji14: Thanks for the benchmarks. I re-ran them on my computer (plus tried `np.char.encode` mentioned in https://github.com/GenericMappingTools/pygmt/pull/3137#issue-2203772747), and think we should go with `[s.encode() for s in np.asarray(strings)]` as you suggest:

```python
%timeit [s.encode() for s in stringlist]
65.1 μs ± 579 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit [s.encode() for s in stringarray.to_pylist()]
1.35 ms ± 5.1 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

%timeit [s.encode() for s in np.asarray(stringlist)]
588 μs ± 6.61 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

%timeit [s.encode() for s in np.asarray(stringarray)]
129 μs ± 599 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit np.char.encode(stringlist)
784 μs ± 7.09 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

%timeit np.char.encode(np.asarray(stringarray, dtype=str))
841 μs ± 11 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

.github/workflows/ci_docs.yml
```diff
@@ -103,6 +103,7 @@ jobs:
             contextily
             geopandas<1.0
             ipython
+            pyarrow
```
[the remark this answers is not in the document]
weiji14: Done at 4bea288e34aee5cab520fa267e6ef9e9828cfcbe.

pygmt/clib/session.py
```diff
@@ -930,39 +931,43 @@ def put_vector(self, dataset, column, vector):
                 f"in column {column} of dataset."
             )
 
-    def put_strings(self, dataset, family, strings):
+    def put_strings(
+        self,
+        dataset: ctp.c_void_p,
+        family: Literal["GMT_IS_VECTOR", "GMT_IS_MATRIX"],
```
seisman, line 937: We can't use Literal here, since family can be `GMT_IS_VECTOR|GMT_IS_DUPLICATE` or similar
```suggestion
        family: str,
```

pygmt/clib/session.py
```diff
@@ -930,39 +931,43 @@ def put_vector(self, dataset, column, vector):
...
+            a 2-D array, for example, you will have to make a copy. Use
+            :func:`numpy.ascontiguousarray` to make sure your vector is contiguous (it
+            won't copy if it already is).
 
         Parameters
         ----------
-        dataset : :class:`ctypes.c_void_p`
+        dataset
             The ctypes void pointer to a ``GMT_Dataset``. Create it with
             :meth:`pygmt.clib.Session.create_data`.
```
seisman, line 959: When calling `create_data`, we use `family="GMT_IS_DATASET|GMT_VIA_VECTOR"`. The function actually creates a `GMT_VECTOR` container (a simplified version of the `GMT_DATASET` container) and returns the pointer.
```suggestion
        dataset
            The ctypes void pointer to a ``GMT_VECTOR``/``GMT_MATRIX`` container. 
            Create it with :meth:`pygmt.clib.Session.create_data`.
```

pygmt/clib/conversion.py
```diff
@@ -304,7 +306,7 @@ def strings_to_ctypes_array(strings: Sequence[str]) -> ctp.Array:
     >>> [s.decode() for s in ctypes_array]
     ['first', 'second', 'third']
     """
-    return (ctp.c_char_p * len(strings))(*[s.encode() for s in strings])
+    return (ctp.c_char_p * len(strings))(*[s.encode() for s in np.asarray(strings)])
```
seisman, line 309: With PR #3507, I believe we can revert changes in this file, and only let the function support a sequence of strings and a string dtype numpy array.

pygmt/_typing.py
```diff
@@ -2,7 +2,17 @@
...
 
+import numpy as np
+
 # Anchor codes
 AnchorCode = Literal["TL", "TC", "TR", "ML", "MC", "MR", "BL", "BC", "BR"]
+
+# String array types
+StringArrayTypes = Sequence[str] | np.ndarray
+with contextlib.suppress(ImportError):
+    StringArrayTypes |= importlib.import_module(name="pyarrow").StringArray
```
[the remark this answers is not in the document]
weiji14: See https://github.com/GenericMappingTools/pygmt/pull/3448, we shouldn't have two lines nested under contextlib.suppress

seisman:
With PR #3619 and #3608 merged, pyarrow arrays with string type now can be correctly converted to numpy string array. So, we don't need to add support of pyarrow arrays in the low-level functions like `Session.put_strings`. Changes in `pygmt/_typing.py`, `pygmt/clib/conversion.py` and `pygmt/clib/session.py` can be reverted, so that this PR only focuses on adding tests for pyarrow string arrays.

pygmt/tests/test_clib_put_strings.py
```diff
@@ -8,12 +8,31 @@
 from pygmt import clib
 from pygmt.exceptions import GMTCLibError
 from pygmt.helpers import GMTTempFile
+from pygmt.helpers.testing import skip_if_no
```
seisman, line 11: Also need to revert changes in this PR, since the low-level function `Session.put_strings` only accepts `numpy.str_` and everything else should be converted to `numpy.str_` by the `_to_numpy` function before calling `put_strings`.

weiji14: Ok, just to document things, the way we are intending pyarrow.string inputs to go through high-level modules like `fig.text` is:

1. `pygmt.Figure.text` -> `clib.Session.virtualfile_in` -> `clib.Session.virtualfile_from_vectors`
2. In `clib.Session.virtualfile_from_vectors`:
   - First call `clib.conversion.vectors_to_arrays` -> `clib.conversion._to_numpy` -> `np.ascontiguous_array` to convert `pyarrow.string` to `np.str_`
   - With a `list[np.str_]`, this is then passed to `clib.Session.put_strings`

For the record, the original intention in this PR, mostly in commit d379e4667c5f8fcc907a3a56543ae734b8c605eb, was to let `clib.Session.put_strings` handle both `np.str_` and `pyarrow.string`.

pygmt/clib/conversion.py
weiji14: Reverted in a927202acfcdb9a236a765534079bb821b798189

pygmt/tests/test_clib_virtualfiles.py
```diff
@@ -137,3 +143,37 @@ def test_open_virtual_file():
         bounds = "\t".join([f"<{col.min():.0f}/{col.max():.0f}>" for col in data.T])
         expected = f"<matrix memory>: N = {shape[0]}\t{bounds}\n"
         assert output == expected
+
+
+@pytest.mark.benchmark
```
seisman, line 148: The test was moved to `test_clib_virtualfile_from_vectors` in #3512.

weiji14: Moved in 6ad6eb96d7d0a0bcbe8e01386d80059a6d70a104

pygmt/tests/test_clib_virtualfile_from_vectors.py
```diff
@@ -53,17 +53,30 @@ def test_virtualfile_from_vectors(dtypes):
...
+    ("array_func", "dtype"),
+    [
+        pytest.param(np.array, {"dtype": np.str_}, id="str"),
+        pytest.param(np.array, {"dtype": np.object_}, id="object"),
+        pytest.param(
+            getattr(pa, "array", None),
+            {},  # pa.string()
+            marks=skip_if_no(package="pyarrow"),
+            id="pyarrow",
+        ),
```
seisman, line 66: A possible solution to avoid the complicated param is:
```
try:
    import pyarrow as pa
    pa_array = pa.array
except ImportError:
    pa_array = None

@pytest.mark.parametrize(
    ("array_func", "dtype"),
    [
        pytest.param(np.array, {"dtype": np.str_}, id="str"),
        pytest.param(np.array, {"dtype": np.object_}, id="object"),
        pytest.param(pa_array, {}, marks=skip_if_no(package="pyarrow"), id="pyarrow"),
    ]
)
```

weiji14: Done at 85d2bb510650daf6685311c17f826b27c8436861

seisman:
It's likely that pytest-codspeed v3.0 breaks our Benchmark workflow. For comparison, see the CI job on the main branch yesterday (https://github.com/GenericMappingTools/pygmt/actions/runs/11830721578/job/32964716012) and this PR (https://github.com/GenericMappingTools/pygmt/actions/runs/11849193208/job/33022039738?pr=2933).

seisman (approved):

PR #2937, opened 2024-01-01
CI: Cache the earth_day_01d_p grid

seisman:

The `earth_day_01d_p.tif` file is used in `test_grdimage_image.py`.

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -12,7 +12,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  pull_request:
+  # pull_request:
   # Schedule runs on 12 noon every Sunday
   schedule:
     - cron: '0 12 * * 0'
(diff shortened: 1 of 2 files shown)


seisman:
One quick thought when working on this PR. 

Currently, the "Cache Data" workflow is disabled by default for PRs. Sometimes, when we add a new cache file, we need to manually enable the workflow in PRs to refresh the caches and also need to disable it before merging. It's tedious although it doesn't happen very often.

A better solution is:

1. Move the `download_test_data` function to a separate source file, e.g., `pygmt/helpers/caching.py`. 
2. Enable the workflow if `pygmt/helpers/caching.py` changes:
```
pull_request:
  paths:
    - 'pygmt/helpers/caching.py'
```

.github/workflows/cache_data.yaml
```diff
@@ -12,7 +12,7 @@ name: Cache data
 
 on:
   # Uncomment the 'pull_request' line below to manually re-cache data artifacts
-  # pull_request:
+  pull_request:
```
seisman, line 15:
```suggestion
  # pull_request:
```

weiji14:
> FYI, I've changed the regression threshold back to 10% to silence false warnings like the above:

Thanks, I've opened an issue (#2942) to track why this benchmark test seems to have so many false warnings.

PR #2938, opened 2024-01-01
Set GMT_SESSION_NAME to a unique name on Windows for multiprocessing support

seisman:
This PR adds the workaround in https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-754774875 as a test to make sure that the workaround for multiprocessing support (i.e., reload pygmt in each process) works as expected.

As mentioned in https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-762649874, the workaround didn't work on Windows. This PR also addresses this issue by setting `GMT_SESSION_NAME` to a unique name on Windows.

xref: https://github.com/GenericMappingTools/gmt/blob/104bcdbafd13eb96260b445910ce3587a362dbc0/src/gmt_api.c#L1036

The newly added test fails on Windows (https://github.com/GenericMappingTools/pygmt/actions/runs/7378717925/job/20074178686?pr=2938) and the commit 07ae8f1445972d1200f04384267524ba63472969 fixes it (see https://github.com/GenericMappingTools/pygmt/actions/runs/7378788411/job/20074335235?pr=2938).

### DIFF
--- a/pygmt/session_management.py
+++ b/pygmt/session_management.py
@@ -1,7 +1,11 @@
 """
 Modern mode session management modules.
 """
+import os
+import sys
+
 from pygmt.clib import Session
+from pygmt.helpers import unique_name
 
 
 def begin():
@@ -12,6 +16,10 @@ def begin():
 
     Only meant to be used once for creating the global session.
     """
+    # On Windows, need to set GMT_SESSION_NAME to a unique value
+    if sys.platform == "win32":
+        os.environ["GMT_SESSION_NAME"] = unique_name()
+
     prefix = "pygmt-session"
     with Session() as lib:
         lib.call_module(module="begin", args=prefix)
--- a/pygmt/tests/test_session_management.py
+++ b/pygmt/tests/test_session_management.py
@@ -1,7 +1,9 @@
 """
 Test the session management modules.
 """
+import multiprocessing as mp
 import os
+from importlib import reload
 from pathlib import Path
 
 import pytest
@@ -67,8 +69,6 @@ def _gmt_func_wrapper(figname):
     Currently, we have to import pygmt and reload it in each process. Workaround from
     https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-754774875.
     """
-    from importlib import reload
-
     import pygmt
 
     reload(pygmt)
@@ -81,9 +81,6 @@ def test_session_multiprocessing():
     """
     Make sure that multiprocessing is supported if pygmt is re-imported.
     """
-
-    import multiprocessing as mp
-
     prefix = "test_session_multiprocessing"
     with mp.Pool(2) as p:
         p.map(_gmt_func_wrapper, [f"{prefix}-1.png", f"{prefix}-2.png"])
--- a/pygmt/tests/test_session_management.py
+++ b/pygmt/tests/test_session_management.py
@@ -2,6 +2,7 @@
 Test the session management modules.
 """
 import os
+from pathlib import Path
 
 import pytest
 from pygmt.clib import Session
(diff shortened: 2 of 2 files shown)


pygmt/tests/test_session_management.py
```diff
@@ -57,3 +58,34 @@ def test_gmt_compat_6_is_applied(capsys):
...
+
+
+def _gmt_func_wrapper(figname):
+    """
+    A wrapper for running PyGMT scripts with multiprocessing.
+
+    Currently, we have to import pygmt and reload it in each process. Workaround from
+    https://github.com/GenericMappingTools/pygmt/issues/217#issuecomment-754774875.
+    """
+    from importlib import reload
```
weiji14, line 70: Can move this import to the top?

pygmt/tests/test_session_management.py
```diff
@@ -57,3 +58,34 @@ def test_gmt_compat_6_is_applied(capsys):
...
+    fig.basemap(region=[10, 70, -3, 8], projection="X8c/6c", frame="afg")
+    fig.savefig(figname)
+
+
+def test_session_multiprocessing():
+    """
+    Make sure that multiprocessing is supported if pygmt is re-imported.
+    """
+
+    import multiprocessing as mp
```
weiji14, line 85: Same here, can probably move this import to the top.

pygmt/session_management.py
```diff
@@ -12,6 +16,10 @@ def begin():
 
     Only meant to be used once for creating the global session.
     """
+    # On Windows, need to set GMT_SESSION_NAME to a unique value
+    if sys.platform == "win32":
+        os.environ["GMT_SESSION_NAME"] = unique_name()
```
weiji14, line 21: Hmm, I thought this should best be handled upstream in GMT, but then I saw the note from https://github.com/GenericMappingTools/gmt/blob/104bcdbafd13eb96260b445910ce3587a362dbc0/src/gmt_api.c#L1050-L1057:

```
	/* OK, the trouble is the following. On Win, if the Windows executables are run from within a bash window
	   gmtapi_get_ppid returns different values for each call, and this completely breaks the idea
	   of using the constant PPID (parent PID) to create unique file names for each session.
	   So, given that we didn't yet find a way to make this work from within MSYS (and likely Cygwin)
	   we are forcing PPID = 0 in all Windows variants unless set via GMT_SESSION_NAME. A corollary of this
	   is that Windows users running many bash windows concurrently should use GMT_SESSION_NAME in their scripts
	   to give unique values to different scripts.  */
```

Would setting a unique `GMT_SESSION_NAME` here in PyGMT result in any issues then? I couldn't run `git blame` on those lines to find out more context.

pygmt/tests/test_session_management.py
```diff
@@ -57,3 +58,34 @@ def test_gmt_compat_6_is_applied(capsys):
...
+
+    prefix = "test_session_multiprocessing"
+    with mp.Pool(2) as p:
+        p.map(_gmt_func_wrapper, [f"{prefix}-1.png", f"{prefix}-2.png"])
+    Path(f"{prefix}-1.png").unlink()
+    Path(f"{prefix}-2.png").unlink()
```
weiji14, line 91: Should probably use GMTTempFile or a try-except block to remove the files automatically, even if the multiprocessing function fails.

pygmt/tests/test_session_management.py
seisman: Done in 236f874c44daec321d5ce90e54fe1dabe4ba7071

pygmt/tests/test_session_management.py
seisman: Done in 236f874c44daec321d5ce90e54fe1dabe4ba7071.

pygmt/session_management.py
seisman: This note (https://docs.generic-mapping-tools.org/dev/begin.html#note-on-unix-shells) is also useful. So, the key point for how GMT modern mode works is having a unique and invariable value so that GMT modules know where to find the session directory and the historical information. The value can be either the constant PPID or a number manually set by `GMT_SESSION_NAME`.

> Would setting a unique `GMT_SESSION_NAME` here in PyGMT result in any issues then? 

All tests still pass, so I believe it has no side effects.

pygmt/tests/test_session_management.py
seisman: Likely not necessary here. If anything wrong happens, the two PNG files won't be generated, and `Path(...).unlink()` will raise an exception.

weiji14 (approved):

pygmt/session_management.py
weiji14: Ok, but maybe we should check if `GMT_SESSION_NAME` is set by the user first before overriding it.

seisman: Users should **NEVER** set `GMT_SESSION_NAME` as a **global** environment variable. Even if they set `GMT_SESSION_NAME`, we should override it in PyGMT.

Here are steps to understand why `GMT_SESSION_NAME` should be overridden. Open a terminal, define the variable:
```
export GMT_SESSION_NAME=123456
```
and run `gmt begin`. Then you'll see the session directory `~/.gmt/sessions/gmt_session.123456/`; all information will be saved in this directory. So, if a global `GMT_SESSION_NAME` is set, processes will write information and output to the same directory, leading to crashes.

weiji14 (approved):

Issue #2942, opened 2024-01-02
Investigate large (>10%) deviations in benchmark of test_grdlandmask_no_outgrid


weiji14:
### Description of the problem

The [`test_grdlandmask_no_outgrid`](https://github.com/GenericMappingTools/pygmt/blob/5639f24d7a511493330830585b9c2eff919a1cf0/pygmt/tests/test_grdlandmask.py#L47-L58) function's benchmark was added in #2911, but it seems like the performance has been deviating wildly (>10%) in many PRs since (e.g. https://github.com/GenericMappingTools/pygmt/pull/2937#issuecomment-1873549552, https://github.com/GenericMappingTools/pygmt/pull/2938#issuecomment-1873554070), even though there has been no evident change in the code of `grdlandmask` or any of the underlying clib functions.

See tracked performance at https://codspeed.io/GenericMappingTools/pygmt/benchmarks/pygmt/tests/test_grdlandmask.py::test_grdlandmask_no_outgrid

Sample flame graph from https://codspeed.io/GenericMappingTools/pygmt/branches/windows-multiprocessing:

![image](https://github.com/GenericMappingTools/pygmt/assets/23487320/4b7510df-6670-46f9-a548-8b11afda0b98)

Opening this issue to discuss why the variance might be so high, and if there are ways to mitigate this.

### Minimal Complete Verifiable Example

```python
make test PYTEST_EXTRA="-r P --pyargs pygmt -k test_grdlandmask_no_outgrid --codspeed"
```

### System information

```bash
PyGMT information:
    version: v0.10.1.dev187+g322f8889
  System information:
    python: 3.12.1 | packaged by conda-forge | (main, Dec 23 2023, 08:03:24) [GCC 12.3.0]
    executable: /usr/share/miniconda/bin/python
    machine: Linux-6.2.0-1018-azure-x86_64-with-glibc2.35
  Dependency information:
    numpy: 1.26.2
    pandas: 2.1.4
    xarray: 2023.12.0
    netCDF4: 1.6.5
    packaging: 23.2
    contextily: None
    geopandas: 0.14.1
    ipython: None
    rioxarray: None
    ghostscript: 10.02.1
  GMT library information:
    binary version: 6.4.0
    cores: 4
    grid layout: rows
    image layout: 
    library path: /usr/share/miniconda/lib/libgmt.so
    padding: 2
    plugin dir: /usr/share/miniconda/lib/gmt/plugins
[...]

PR #2944, opened 2024-01-02
Add common alias cores (x) for grdlandmask

weiji14:

Add alias cores (-x) for [`grdlandmask`](https://docs.generic-mapping-tools.org/6.4/grdlandmask.html?highlight=openmp), and set limit of `cores=2` for `test_grdlandmask_no_outgrid` to make benchmark execution deviate less.

**Preview** at https://pygmt-dev--2944.org.readthedocs.build/en/2944/api/generated/pygmt.grdlandmask.html

Extends #1423, may address #2942 partially.

**Reminders**

**Slash Commands**


pygmt/tests/test_grdlandmask.py
```diff
@@ -49,7 +49,7 @@ def test_grdlandmask_no_outgrid(expected_grid):
     """
     Test grdlandmask with no set outgrid.
     """
-    result = grdlandmask(spacing=1, region=[125, 130, 30, 35])
+    result = grdlandmask(spacing=1, region=[125, 130, 30, 35], cores=2)
```
weiji14, line 52: The Linux runner for GitHub Actions has 2 CPU cores according to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources, but not sure if we should set cores to 2 or 1 for consistent benchmarking.

seisman: Using two cores is faster and can help check the OpenMP feature in GMT, so I prefer to use two cores if possible.

BTW, the macOS runner for GitHub Actions has 3 cores, so instead of `cores=2`, `cores=True` will use all available cores.

seisman:
> ## [CodSpeed Performance Report](https://codspeed.io/GenericMappingTools/pygmt/branches/limit-cores)
> ### Merging #2944 will **improve performances by ×3**
> Comparing `limit-cores` ([137a04f](https://github.com/GenericMappingTools/pygmt/commit/137a04fbbd44e5493bf53d9ffe43fa8c9c5b453c)) with `main` ([88ab1ca](https://github.com/GenericMappingTools/pygmt/commit/88ab1caf0a3f0ab86da8ad5a29315895b6c03615))
> 
> ### Summary
> `⚡ 3` improvements `✅ 61` untouched benchmarks
> 
> ### Benchmarks breakdown
> Benchmark	`main`	`limit-cores`	Change
> ⚡	`test_grdlandmask_no_outgrid`	2,531.2 ms	842.1 ms	×3
> ⚡	`test_grdsample_dataarray_out`	106.3 ms	91.6 ms	+16.05%
> ⚡	`test_sph2grd_no_outgrid`	4.6 s	2.3 s	×2

The benchmark report is quite confusing. It seems that `test_sph2grd_no_outgrid` also speedup by 2 times. Does this mean the `-x` option is a session-level option (similar to GMT's configuration settings) and affect all tests?

pygmt/tests/test_grdlandmask.py
weiji14: Is only 1 core being used if we don't set `cores`, or use `cores=None`? I'm reading https://docs.generic-mapping-tools.org/6.4/gmt.html#core-full which says `[default is to use all available cores]`, but is that only with `cores=True` (i.e. `-x` in GMT)?

The `benchmark.yml` workflow is running on Linux, but I guess we could set `cores=True` since this test would also run on macOS in `ci_test.yml`. My question was more around whether multi-threading with multiple-cores (2 or more) would provide a consistent benchmark compared to using only 1 core, but I guess we need to merge this PR to find out!

weiji14:
> The benchmark report is quite confusing. It seems that `test_sph2grd_no_outgrid` also speedup by 2 times. Does this mean the `-x` option is a session-level option (similar to GMT's configuration settings) and affect all tests?

Yeah, it seems to confirm the hypothesis in https://github.com/GenericMappingTools/pygmt/issues/2942#issuecomment-1873644967 though. Does the setting affect [`GMT_MAX_CORES`](https://docs.generic-mapping-tools.org/6.4/gmt.conf.html#term-GMT_MAX_CORES) somehow?

pygmt/tests/test_grdlandmask.py
seisman: > Is only 1 core being used if we don't set `cores`, or use `cores=None`? I'm reading https://docs.generic-mapping-tools.org/6.4/gmt.html#core-full which says `[default is to use all available cores]`, but is that only with `cores=True` (i.e. `-x` in GMT)?

Yes.

seisman: > > Is only 1 core being used if we don't set `cores`, or use `cores=None`? I'm reading https://docs.generic-mapping-tools.org/6.4/gmt.html#core-full which says `[default is to use all available cores]`, but is that only with `cores=True` (i.e. `-x` in GMT)?
> 
> Yes.

Actually it uses all available cores if `-x` is not used.

```
$ gmt grdlandmask -R0/10/0/10 -Gabc.nc -I1/1 -Vi
grdlandmask [INFORMATION]: Enable all available threads (up to xxx)
```

seisman: Just tried the GMT CLI. Looking at the `real` time, I expect it to be decreasing for the increasing number of cores, but apparently it doens't:
```
$ time gmt grdlandmask -R0/180/0/90 -I1m -Gabc.nc -x1

real	0m13.474s
user	0m16.252s
sys	0m6.025s

$ time gmt grdlandmask -R0/180/0/90 -I1m -Gabc.nc -x2

real	0m10.377s
user	0m22.465s
sys	0m5.969s

$ time gmt grdlandmask -R0/180/0/90 -I1m -Gabc.nc -x3

real	0m16.906s
user	0m51.020s
sys	0m6.215s

$ time gmt grdlandmask -R0/180/0/90 -I1m -Gabc.nc -x4

real	0m15.033s
user	0m59.322s
sys	0m6.043s

$ time gmt grdlandmask -R0/180/0/90 -I1m -Gabc.nc -x5

real	0m16.387s
user	1m19.829s
sys	0m6.144s

```

seisman: Maybe set `cores=1` and see what the flame graph diff shows?

weiji14: > Maybe set `cores=1` and see what the flame graph diff shows?

The `test_grdlandmask_no_outgrid` benchmark speedup went from 3x to 9.6x :rofl: It might be that there is some overhead with using threading (2 cores) vs no threading (1 core). Here's the flame graph diff from https://codspeed.io/GenericMappingTools/pygmt/branches/limit-cores again:

image

The extra `gomp_team_barrier_wait_*` calls are still there, but also some other stuff (in blue).

seisman: This reminds me of the warnings message I saw when running the tests in random order (see https://github.com/GenericMappingTools/pygmt/actions/runs/7396374248/job/20121388367?pr=2936):

```
pygmt/tests/test_session_management.py::test_session_multiprocessing
pygmt/tests/test_session_management.py::test_session_multiprocessing
  /home/runner/micromamba/envs/pygmt/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=2433) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```
There are some related discussions at https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555 and related PR at https://github.com/python/cpython/issues/100228.

> os.fork() in a multi-threaded application is a likely source of deadlocks on many platforms. We should raise a warning when people call os.fork() from a process that we know has other threads running.

Not sure if it's related to the issue here.

weiji14: > ```
> pygmt/tests/test_session_management.py::test_session_multiprocessing
> pygmt/tests/test_session_management.py::test_session_multiprocessing
>   /home/runner/micromamba/envs/pygmt/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=2433) is multi-threaded, use of fork() may lead to deadlocks in the child.
>     self.pid = os.fork()
> ```

That [`test_session_multiprocessing`](https://github.com/GenericMappingTools/pygmt/blob/fbfed861018522225b0cae13404aa9428ba24395/pygmt/tests/test_session_management.py#L65-L88) function is only running `basemap` and `savefig` right? It shouldn't have any multi-threading going on, so not sure what's happening.

I tried running some benchmarks using cores from -1 (all cores) to 0 (no-multithreading) to 16 (all cores on my laptop). This was running `grdlandmask` with a spacing of 0.05 arc degrees. The results do vary between runs, but in general, more cores means less time:

benchmark_grdlandmask_0.05arc_degree

Code to reproduce:

```python
import time
import pandas as pd
import pygmt
import tqdm

timings = {}
for cores in tqdm.tqdm(range(-1, 16)):
    tic = time.perf_counter()
    pygmt.grdlandmask(spacing="0.05d", region=[0, 180, 0, 90], cores=cores)
    toc = time.perf_counter()
    timings[cores] = toc - tic

df = pd.DataFrame.from_dict(data=timings, orient="index", columns=["time"])

# %%
fig = pygmt.Figure()
region = pygmt.info(data=df.reset_index(), per_column=True, spacing=0.05)
fig.plot(
    x=df.index,
    y=df.time,
    region=region,
    pen="thick",
    frame=[
        "WSne+tBenchmark grdlandmask",
        "xaf+lCores",
        "yaf+lTime (s), lower is better",
    ],
)
fig.savefig("benchmark_grdlandmask.png")
fig.show()
```

If I change the spacing from 0.05 arc degrees to 1 arc degree, more cores actually take more time:

benchmark_grdlandmask_1arc_degree

For that `test_grdlandmask_no_outgrid` unit test which is using `spacing=1` (i.e. 1 arc degree), I think we should just use `cores=2`, since `cores=3` might not be much faster on such a low resolution. As long as we are benchmarking with a fixed number of core counts, it should be ok?

seisman: sounds ok

pygmt/tests/test_grdlandmask.py
```diff
@@ -49,7 +49,7 @@ def test_grdlandmask_no_outgrid(expected_grid):
     """
     Test grdlandmask with no set outgrid.
     """
-    result = grdlandmask(spacing=1, region=[125, 130, 30, 35])
+    result = grdlandmask(spacing=1, region=[125, 130, 30, 35], cores=1, verbose="i")
```
weiji14, line 52: Hmm, the debug info at https://github.com/GenericMappingTools/pygmt/actions/runs/7403072686/job/20142175894?pr=2944#step:8:111 is showing that there are 4 threads on GitHub Actions??

```
  grdlandmask [INFORMATION]: Enable 1 threads of 4 available
  grdlandmask [INFORMATION]: Selected ice front line as Antarctica coastline
  grdlandmask [INFORMATION]: GSHHG version 2.3.7
  grdlandmask [INFORMATION]: Derived from World Vector Shoreline, CIA WDB-II, and Atlas of the Cryosphere
  grdlandmask [INFORMATION]: Processed by Paul Wessel and Walter H. F. Smith, 1994-2017
  grdlandmask [INFORMATION]: Nodes in the oceans will be set to
  0
  grdlandmask [INFORMATION]: Nodes on land will be set to
  1
  grdlandmask [INFORMATION]: Nodes in lakes will be set to
  0
  grdlandmask [INFORMATION]: Nodes in islands will be set to
  1
  grdlandmask [INFORMATION]: Nodes in ponds will be set to
  0
  grdlandmask [INFORMATION]: Apply linear scale to avoid wrap-around: -Jx100id; this temporarily changes the domain
  grdlandmask [INFORMATION]: Writing grid to file /tmp/pygmt-0c0skk20.nc
  grdlandmask [INFORMATION]: Level 0 set for 33 nodes
  grdlandmask [INFORMATION]: Level 1 set for 3 nodes
  grdlandmask [INFORMATION]: Enable 1 threads of 4 available
  grdlandmask [INFORMATION]: Selected ice front line as Antarctica coastline
  grdlandmask [INFORMATION]: GSHHG version 2.3.7
  grdlandmask [INFORMATION]: Derived from World Vector Shoreline, CIA WDB-II, and Atlas of the Cryosphere
  grdlandmask [INFORMATION]: Processed by Paul Wessel and Walter H. F. Smith, 1994-2017
  grdlandmask [INFORMATION]: Nodes in the oceans will be set to
  0
  grdlandmask [INFORMATION]: Nodes on land will be set to
  1
  grdlandmask [INFORMATION]: Nodes in lakes will be set to
  0
  grdlandmask [INFORMATION]: Nodes in islands will be set to
  1
  grdlandmask [INFORMATION]: Nodes in ponds will be set to
  0
  grdlandmask [INFORMATION]: Apply linear scale to avoid wrap-around: -Jx100id; this temporarily changes the domain
  grdlandmask [INFORMATION]: Writing grid to file /tmp/pygmt-i3n6zkcg.nc
  grdlandmask [INFORMATION]: Level 0 set for 33 nodes
  grdlandmask [INFORMATION]: Level 1 set for 3 nodes
```

So the 2 CPUs mentioned at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources might have hyperthreading, and we could potentially set `cores=4`, though see https://github.com/GenericMappingTools/pygmt/pull/2944#discussion_r1441172036 on how more cores might not actually improve runtime.

pygmt/tests/test_grdlandmask.py
weiji14: Ok, back to using `cores=2` in 97637d4dffa252ce3160cec452accfe5fec57fcc.

seisman (approved):

PR #2955, opened 2024-01-05
ReadtheDocs: Cancel building pull requests for non-docs changes

seisman:
Reference: https://docs.readthedocs.io/en/stable/build-customization.html#cancel-build-based-on-a-condition

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,6 +10,22 @@ build:
   tools:
     python: "mambaforge-4.10"
   jobs:
+    post_checkout:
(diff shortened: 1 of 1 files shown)


seisman:
Actually this action is more interesting: https://github.com/dfm/rtds-action

weiji14 (approved):

weiji14:
> Actually this action is more interesting: https://github.com/dfm/rtds-action

Can look into this next time, it looks better suited if the documentation takes a long time to build. The setup doesn't look as easy, since it requires managing webhook tokens and stuff.

PR #2952, opened 2024-01-04
CI: Trigger workflows based on changed files

seisman:
Some workflow runs are unnecessary and waste CI resources. For example, PR #2950 contains no codes or docs changes, but when it was merged into the main branch, workflows "Tests" and "Docs" were still triggered (see https://github.com/GenericMappingTools/pygmt/actions?query=branch%3Amain).

This PR updates the workflow trigger conditions based on changes files, to save CI resources.

- [x] benchmarks.yml
- [x] cache_data.yaml
- [x] check-links.yml
- [x] ci_docs.yml
- [x] ci_doctests.yaml
- [x] ci_tests_dev.yaml
- [x] ci_tests_legacy.yaml
- [x] ci_tests.yaml
- [x] dvc-diff.yml
- [x] format-command.yml
- [x] publish-to-pypi.yml
- [x] release-baseline-images.yml
- [x] release-drafter.yml
- [x] slash-command-dispatch.yml
- [x] style_checks.yaml
- [x] type_checks.yml

Issue #1113, opened 2021-03-23
Discuss and document the Pull Request review process


seisman:
During the release v0.1.0 and v0.2.1, @weiji14 and I were the only active maintainers of the project. We usually submitted PRs, waited for approval from the other one, and then merged PRs immediately after approval. 

Now the team is growing quickly, and we have more maintainers and contributors comment on PRs, so we usually do **NOT** merge PRs immediately after one approval. The review process helps make the project better, but also becomes longer. I'm not sure if we have reached a consensus about _when_ a PR should be merged and _who_ should push the merge button (currently most PRs were merged by @weiji14 and me). New contributors may also be confused why their PRs are already approved but still not merged. 

So I feel we should discuss the PR review process and document it in the contributing/maintenance guides.

Here is the PR review process I think we're loosely following:

- Open an issue and discuss it
- Submit a **draft** PR and work on it 
- Add labels and milestones
- Several rounds of review and revision
- Mark the PR as Ready for Review
- Formal review process
- Someone approves the PR, and adds the "final review call" label
- Merge the PR if approved by _X_ reviewers or no further comments after X hours

Thoughts and comments on the review process?

meghanrjones:
Thanks for starting this conversation! In addition to the aspects you mentioned, I would find it helpful to get some clarity on people's preferences for if/when to request specific reviewers (rather than generally marking 'ready for review'). Only people with write access can request reviewers, so this isn't relevant for all PRs. Still, the review request process for PyGMT confused me at first regarding whether unsolicited reviews on PRs are welcome (hopefully they are). I guess my opinion is that specific reviewer requests are only needed for specialized PRs (e.g., early development on dvc, workflows) and that more general docs/code PRs don't require reviewer requests (especially since they seem to always get reviewed really fast anyways).

The self-assign functionality for issues could be a worthwhile option to mention in the maintenance guide for people with write access. At least I find this really useful for keeping track of my tasking - not sure how others feel about it.

meghanrjones:
> Merge the PR if approved by X reviewers or no further comments after X hours

I am skeptical that one rule could apply to all PRs. For example, I would prefer to see more opportunity for people to comment if #379 gets a 'final review call' (order days) since it's a outward facing change, while I would really be fine if #1110 were merged immediately after one approval.

weiji14:
> _who_ should push the merge button (currently most PRs were merged by @weiji14 and me)

[...]

PR #2958, opened 2024-01-06
CI: Run benchmarks if PR is labeled with "run/benchmark"

seisman:
Support running benchmarks in PRs with the `run/benchmark` label.

Address comment https://github.com/GenericMappingTools/pygmt/pull/2952#discussion_r1443648845.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -14,6 +14,7 @@ on:
       - '.github/workflows/benchmarks.yml'
   # Run in PRs but only if the PR has the 'run/benchmark' label
   pull_request:
+    types: [ labeled ]
   # `workflow_dispatch` allows CodSpeed to trigger backtest
   # performance analysis in order to generate initial data.
   workflow_dispatch:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -17,9 +17,6 @@ on:
   # `workflow_dispatch` allows CodSpeed to trigger backtest
   # performance analysis in order to generate initial data.
   workflow_dispatch:
-  # support slash command
-  repository_dispatch:
-    types: [benchmark-command]
   release:
     types:
       - published
@@ -40,19 +37,7 @@ jobs:
       # Checkout current git repository
       - name: Checkout
         uses: actions/checkout@v4.1.1
-        if: github.event_name != 'repository_dispatch'
         with:
-          # fetch all history so that setuptools-scm works
-          fetch-depth: 0
-
-      # Checkout the pull request branch
-      - name: Checkout
-        uses: actions/checkout@v4.1.1
-        if: github.event_name == 'repository_dispatch'
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          repository: ${{ github.event.client_payload.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.client_payload.pull_request.head.ref }}
           # fetch all history so that setuptools-scm works
           fetch-depth: 0
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   benchmarks:
     runs-on: ubuntu-22.04
-    if: github.repository == 'GenericMappingTools/pygmt' && contains(github.event.pull_request.labels.*.name, 'run/benchmark')
+    if: github.repository == 'GenericMappingTools/pygmt' && (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run/benchmark'))
     defaults:
       run:
         shell: bash -l {0}
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -14,7 +14,7 @@ on:
       - '.github/workflows/benchmarks.yml'
   # Run in PRs but only if the PR has the 'run/benchmark' label
   pull_request:
-    types: [ labeled, synchronize ]
+    types: [ opened, reopened, labeled, synchronize ]
   # `workflow_dispatch` allows CodSpeed to trigger backtest
   # performance analysis in order to generate initial data.
   workflow_dispatch:
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -12,6 +12,7 @@ on:
     paths:
       - 'pygmt/**/*.py'
       - '.github/workflows/benchmarks.yml'
+  # Run in PRs but only if the PR has the 'run/benchmark' label
   pull_request:
   # `workflow_dispatch` allows CodSpeed to trigger backtest
   # performance analysis in order to generate initial data.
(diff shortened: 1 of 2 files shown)


seisman:
The slash command won't work until the PR is merged into main branch.

weiji14:
> The slash command won't work until the PR is merged into main branch.

Actually, will CodSpeed be able to link the report to the PR branch if we trigger it via `repository_dispatch` instead of `pull_request`?

seisman:
> > The slash command won't work until the PR is merged into main branch.
> 
> Actually, will CodSpeed be able to link the benchmark report to the PR branch if we trigger it via `repository_dispatch` instead of `pull_request`?

That's a good question. I guess the answer is no.

weiji14:
> > Actually, will CodSpeed be able to link the benchmark report to the PR branch if we trigger it via `repository_dispatch` instead of `pull_request`?
> 
> That's a good question. I guess the answer is no.

How about we set up the workflow to run only when there's a `run-benchmark` label? Can try to set up an if-condition following https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

seisman:
> > > Actually, will CodSpeed be able to link the benchmark report to the PR branch if we trigger it via `repository_dispatch` instead of `pull_request`?
> > 
> > 
> > That's a good question. I guess the answer is no.
> 
> How about we set up the workflow to run only when there's a `run-benchmark` label? Can try to set up an if-condition following https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

Sounds a clever solution.

seisman:
The benchmark workflow is trigged by the "run-benchmark" label, see https://github.com/GenericMappingTools/pygmt/actions/runs/7431110786?pr=2958.

seisman:
I feel triggering a workflow based on the PR label is much better than the slash commands. Some thoughts:

1. Migrate the other two slash commands `/format` and `/test-gmt-dev` to triggering by labels
2. Instead of using labels like `run-benchmark`, maybe labels `run/benchmark`, `run/format`, `run/test-gmt-dev` are better?

seisman:
Hmm, the workflow is canceled because I added the 'maintenance' label.

weiji14:
> 1. Migrate the other two slash commands `/format` and `/test-gmt-dev` to triggering by labels

Yeah, I was thinking of using it for `/test-gmt-dev`, but maybe not for `/format` since the labels can only be added by contributors/maintainers with triage permissions or above, so new contributors won't be able to use it. 

> 2. Instead of using labels like `run-benchmark`, maybe labels `run/benchmark`, `run/format`, `run/test-gmt-dev` are better?

Sounds good.

.github/workflows/benchmarks.yml
```diff
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   benchmarks:
     runs-on: ubuntu-22.04
-    if: github.repository == 'GenericMappingTools/pygmt'
+    if: github.repository == 'GenericMappingTools/pygmt' && contains(github.event.pull_request.labels.*.name, 'run/benchmark')
```
seisman, line 31: Actually I'm unsure if the workflow will run in the main branch, because it's a push event not a pull_request event.

seisman: Tried in my own fork. It doesn't work in the main branch.

seisman: Changed the `if` condition to:
```
if: github.repository == 'GenericMappingTools/pygmt' && (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run/benchmark'))
```
Now it works in my own fork, see the changes between my fork and the pygmt repository: https://github.com/GenericMappingTools/pygmt/compare/main...seisman:pygmt:main and the workflow was triggered when pushing in the main branch (https://github.com/seisman/pygmt/actions/workflows/benchmarks.yml).

seisman:
It's unclear why I can't see the flame graphs.

.github/workflows/benchmarks.yml
```diff
@@ -12,8 +12,9 @@ on:
     paths:
       - 'pygmt/**/*.py'
       - '.github/workflows/benchmarks.yml'
-  # Uncomment the 'pull_request' line below to trigger the workflow in PR
-  # pull_request:
+  # Run in PRs but only if the PR has the 'run/benchmark' label
```
weiji14, line 15: Should we document in `.github/PULL_REQUEST_TEMPLATE.md` or somewhere for contributors to request that the `run/benchmark` label gets added to PRs that might affect performance? Or just rely on the maintainers to set the label for now?

weiji14 (commented):
> It's unclear why I can't see the flame graphs.

Hmm yeah, are you getting this `Execution profile not available` message too at https://codspeed.io/GenericMappingTools/pygmt/branches/slash-benchmark?

image

If I try another branch (e.g. https://codspeed.io/GenericMappingTools/pygmt/branches/alias/function), it looks ok, though sometimes the flame graphs can take some time to render.

.github/workflows/benchmarks.yml
seisman: I prefer to leave the jobs to the maintainers. We probably will document it in the contributors guides when addressing https://github.com/GenericMappingTools/pygmt/issues/1113.

weiji14: Ok, or we can put it in `doc/maintenance.md` somewhere. Not urgent though, since we aren't running the benchmarks that often.

weiji14 (approved):
Thanks @seisman! Have secretly wanted this label-based trigger mechanism in PyGMT for a while now to be honest :grinning:

weiji14:
Quick question: Do we want to remove the `run/benchmark` label after a PR is merged, or keep the label on?

seisman:
I'm OK with either. Maybe we should keep the label so that we know which branches have performance reports.

seisman:
> Quick question: Do we want to remove the `run/benchmark` label after a PR is merged, or keep the label on?

Actually, we should remove the 'run/benchmark' label. We usually label a PR as "needs-review", then "final-review-call" and remove the "final-review-call" label before merging. All these actions will trigger the benchmarks workflow if the `run/benchmark` label is kept.

weiji14:
> > Quick question: Do we want to remove the `run/benchmark` label after a PR is merged, or keep the label on?
> 
> Actually, we should remove the 'run/benchmark' label. We usually label a PR as "needs-review", then "final-review-call" and remove the "final-review-call" label before merging. All these actions will trigger the benchmarks workflow if the `run/benchmark` label is kept.

Owh, but we would need to remove all those labels at the same time right? Maybe better to set an if-condition so that the benchmark workflow is not re-triggered after a PR is already merged?

seisman:
> Owh, but we would need to remove all those labels at the same time right?

We need to remove `run/benchmark` label as early as possible to avoid unnecessary runs.



> Maybe better to set an if-condition so that the benchmark workflow is not re-triggered after a PR is already merged?

It may make the `if`-condition too complicated, which is already very long.

weiji14:
> > Owh, but we would need to remove all those labels at the same time right?
> 
> We need to remove `run/benchmark` label as early as possible to avoid unnecessary runs.

Ok, let's go with this. Remove the `run/benchmark` label before the other labels are removed. Same goes for the `run/test-gmt-dev` label.

Issue #2961, opened 2024-01-07
Bump to GMT 6.5.0


seisman:
:tada: [GMT 6.5.0](https://github.com/GenericMappingTools/gmt/releases/tag/6.5.0) has been released! It is installable from the [conda-forge channel](https://anaconda.org/conda-forge/gmt/files) using the following command:

`mamba install -c conda-forge gmt=6.5.0`

**To-Do for bumping the GMT version in CI**:

- [x] Bump the GMT version in CI (1 PR) #2962
  - [x] Update `environment.yml`
  - [x] Update `ci/requirements/docs.yml`
  - [x] Update `.github/workflows/benchmarks.yml`
  - [x] Update `.github/workflows/cache_data.yaml`
  - [x] Update `.github/workflows/ci_doctests.yaml`
  - [x] Update `.github/workflows/ci_docs.yml`
  - [x] Update `.github/workflows/ci_tests.yaml`
  - [x] Add the legacy GMT version to `.github/workflows/ci_tests_legacy.yaml`
  - [x] Update GMT 6.x minor version for `extlinks` urls in `doc/conf.py`
  - [x] Update GMT 6.x minor version in `.github/workflows/check-links.yml`
- [x] Fix failing tests (1 or more PRs)
- [x] Fix [xfail](https://docs.pytest.org/en/stable/skipping.html#xfail-mark-test-functions-as-expected-to-fail) pytest markers on tests that are now xpass
- [x] Fix links to GMT documentation (renaming "cookbook" to "reference"): #2990

seisman:
Probably need to see if there is GMT 6.5.1 (https://github.com/GenericMappingTools/gmt/issues/8265).

seisman:
Since there are ~200 failures, I think it makes more sense to update baseline images by modules.

weiji14:
> Since there are ~200 failures, I think it makes more sense to update baseline images by modules.

Maybe group some of them, otherwise there will be 35 PRs. This is the summary count:

Test | Count | PR
--|--|--
../pygmt/tests/test_basemap.py                                     | 10 |
../pygmt/tests/test_coast.py                                       |  4 |
../pygmt/tests/test_colorbar.py                                    |  6 |
../pygmt/tests/test_config.py                                      | 10 |
../pygmt/tests/test_contour.py                                     |  6 |
../pygmt/tests/test_datasets_earth_age.py                          |  2 |
[...]

PR #2978, opened 2024-01-09
Update min/max value of remote datasets for recent upstream changes

seisman:

See https://github.com/GenericMappingTools/pygmt/issues/2961#issuecomment-1882903268 for context.

The min/max values of remote datasets are out of our control so it makes little sense to check them in PyGMT tests. This PR remove all checks about min/max values of remote datasets, with one exception `earth_mask`, since we know that the earth_mask dataset has discrete values from 0 to 4.


michaelgrund (approved):

weiji14:
Ok with removing the min/max value checks, but we should still have some check to ensure that the pixel values returned are valid. Maybe something like a `notnull()` check, or some other statistical measure?

seisman:
> Maybe something like a `notnull()` check, 

`notnull()` checks make no sense to some datasets (e.g., `earth_age`), which  contains NaN values.

> or some other statistical measure?

It means we still need to update these statistical values when the datasets change.

seisman:
I've added the checks of min/max values back to the tests. It means more maintenance burden but at least the values can make sure that the loaded data have reasonable values.

PR #2981, opened 2024-01-10
Simplify two colorbar tests and remove all others

seisman:
See https://github.com/GenericMappingTools/pygmt/issues/2961#issuecomment-1882903268 for context.

This PR removes all existing colorbar tests and add a new, simpler one.

### DIFF
--- a/pygmt/tests/test_colorbar.py
+++ b/pygmt/tests/test_colorbar.py
@@ -7,62 +7,10 @@ from pygmt import Figure
 
 @pytest.mark.benchmark
 @pytest.mark.mpl_image_compare
-def test_colorbar_box():
+def test_colorbar():
     """
-    Create colorbar with box around it.
-    """
-    fig = Figure()
-    fig.colorbar(cmap="rainbow", box=True, position="x0c/0c+w1c/0.5c")
-    return fig
-
-
-@pytest.mark.mpl_image_compare
-def test_colorbar_box_with_fill():
-    """
-    Create colorbar with box that has a different colored fill.
+    Create a simple colorbar.
     """
     fig = Figure()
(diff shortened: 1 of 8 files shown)


pygmt/tests/test_colorbar.py
```diff
@@ -7,62 +7,10 @@
...
-
-
-@pytest.mark.mpl_image_compare
-def test_colorbar_shading_list():
-    """
-    Create colorbar and set shading by passing the high/low values as a list.
-    """
-    fig = Figure()
-    fig.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
-    fig.colorbar(cmap="geo", shading=[-0.7, 0.2], frame=True)
```
weiji14, line 67: Maybe keep this one that checks that shading works (implemented in #752/#1143). This colorbar is a little non-standard.

seisman: OK, added it back but changed the projection to `X8c` to have a smaller size.

michaelgrund (approved):

pygmt/tests/test_colorbar.py
```diff
@@ -63,6 +22,6 @@ def test_colorbar_shading_list():
     Create colorbar and set shading by passing the high/low values as a list.
     """
     fig = Figure()
-    fig.basemap(region=[0, 10, 0, 10], projection="X15c", frame="a")
+    fig.basemap(region=[0, 10, 0, 10], projection="X8c", frame="a")
```
weiji14, line 25: Can make the y-axis smaller maybe?

```suggestion
    fig.basemap(region=[0, 10, 0, 2], projection="X8c", frame="a")
```

weiji14 (approved):
A little uncomfortable removing these tests, but ok then. Just one optional suggestion, and make sure to update the title before merge.

PR #2991, opened 2024-01-12
NEP29: Set minimum required version to NumPy 1.23+

seisman:

Following NEP29 policy where NumPy 1.22 is to be dropped on or after 1 Jan 2024 (in a minor version increment, i.e. for PyGMT v0.11.0). Bumps minimum supported NumPy version to 1.23 in the pyproject.toml, requirements.txt and environment.yml files. Also update installation documentation to mention NumPy 1.22+ requirement.

This is in line with PyGMT's policy on [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) at https://www.pygmt.org/v0.10.0/maintenance.html#dependencies-policy.

Supersedes #2586

### DIFF
--- a/ci/requirements/docs.yml
+++ b/ci/requirements/docs.yml
@@ -7,7 +7,7 @@ dependencies:
     - python=3.12
     - gmt=6.5.0
     - ghostscript=10.02.1
-    - numpy>=1.22
+    - numpy
     - pandas
     - xarray
     - netCDF4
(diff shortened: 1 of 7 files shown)


ci/requirements/docs.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
     - python=3.12
     - gmt=6.5.0
     - ghostscript=10.02.1
-    - numpy>=1.22
```
seisman, line 10: Removing the numpy version information from this file, so that we have one fewer file to update next time.

weiji14: Ok, using `>=1.22` might mean the latest numpy version is pulled in anyways.

weiji14 (approved):

PR #2996, opened 2024-01-14
Rename API function Session.open_virtual_file to Session.open_virtualfile (remove in v0.15.0)

seisman:

The GMT API function name is `GMT_Open_VirtualFile`, so the Python wrapper should be `open_virtualfile` rather than `open_virtual_file`. We also use `virtualfile_from_vectors/matrix/grid` rather than `virtual_file_from_vectors/matrix/grid`.

The API function `Session.open_virtual_file` has been renamed to `Session.open_virtualfile` in a backward-compatible way and will be fully removed in v0.15.0 (four minor releases which is about one year). A warning will be raised if anyone uses the old function name.

xref: https://docs.generic-mapping-tools.org/dev/devdocs/api.html#list-of-api-functions


pygmt/clib/session.py
```diff
@@ -1190,6 +1190,45 @@ def open_virtual_file(self, family, geometry, direction, data):
...
 
+    def open_virtual_file(self, family, geometry, direction, data):
+        """
+        Open a GMT virtual file to pass data to and from a module.
+
+        The method has been renamed to ``open_virtualfile` since v0.11.0 and will be
+        fully removed in v0.15.0.
+
+        Examples
+        --------
```
weiji14, line 1201: Is a doctest actually needed for the old `open_virtual_file` method? Or can we just omit it?

weiji14 (approved):
Only a couple of minor suggestions.

pygmt/clib/session.py
seisman: Just want to make sure that the old function name still works as expected. But I'm OK with removing it, since it's a very simple wrapper.

seisman: Removed in 9fcff9c1fdc4f8b08bda3f48982cd8fa2b2302d6.

PR #3002, opened 2024-01-15
Figure.timestamp: Deprecate parameter 'justification' to 'justify' (remove in v0.13.0)

seisman:

We use the parameter `justify` in many modules (e.g., `text`, `plot`) but use `justification` in `Figure.timestamp`. This PR deprecates the `justification` parameter to `justify` to make it consistent in the project.

FYI, the GMT project also use `justify` as the long option name: https://github.com/GenericMappingTools/gmt/blob/0c3435fba382755a87ba7629cbe1db288c2a9a60/src/gmt_common_longoptions.h#L86

Since it's a rarely used module, I think it's OK to fully remove the deprecated parameter after two releases (i.e., v0.13.0).


seisman:
Although `Figure.timestamp` is rarely used, I think it's our first wrapper which

- [x] has an inline example
- [x] has a gallery example
- [x] has 100% feature completeness (see #2208 for context)
- [x] has 100% typing hints completeness
- [x] is Pythonic [arguments are simple strings without modifiers like `+t`/`+o`]

So, it's time to give this method a final review, then we can categorize it as "Ongoing maintenance".

pygmt/src/timestamp.py
```diff
@@ -44,12 +45,12 @@ def timestamp(
         The text must be no longer than 64 characters.
     label
         The text string shown after the GMT timestamp logo.
-    justification
+    justify
         Justification of the timestamp box relative to the plot's bottom-left corner
-        (i.e., the plot origin). The *justification* is a two-character code that is a
+        (i.e., the plot origin). The *justify* is a two-character code that is a
         combination of a horizontal (**L**\ (eft), **C**\ (enter), or **R**\ (ight)) and
         a vertical (**T**\ (op), **M**\ (iddle), or **B**\ (ottom)) code. For example,
```
yvonnefroehlich, line 52:
```suggestion
        (i.e., the plot origin). Give a two-character code that is a
        combination of a horizontal (**L**\ (eft), **C**\ (enter), or **R**\ (ight)) and
        a vertical (**T**\ (op), **M**\ (iddle), or **B**\ (ottom)) code. For example,
```

michaelgrund (approved):

PR #3005, opened 2024-01-16
Update baseline images for test_grdimage_image

seisman:
Address https://github.com/GenericMappingTools/pygmt/issues/2961.

PR #3006, opened 2024-01-16
Slightly increase the tolerance for test_grdimage_global_subset

seisman:
Address https://github.com/GenericMappingTools/pygmt/pull/3005#issuecomment-1893055034.


weiji14 (approved):
Should be ok once *all* of the tests pass.

PR #3009, opened 2024-01-16
Figure.makecpt: Mention exchange of background (B) and foreground (F) colors for usage of "reverse"

yvonnefroehlich:

I feel it is fair to mention in the docstring that using the `reverse` parameter of `Figure.makecpt` also exchanges the background (B) and foreground (F) colors.

Upstream GMT documentation: https://docs.generic-mapping-tools.org/6.5/makecpt.html#i

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/makecpt.py
+++ b/pygmt/src/makecpt.py
@@ -120,8 +120,8 @@ def makecpt(**kwargs):
         level :func:`pygmt.makecpt` is called.
     reverse : str
         Set this to ``True`` or **c** [Default] to reverse the sense of color
-        progression in the master CPT. Also exchanges the foreground and background
-        colors, including those specified by :gmt-term:`COLOR_BACKGROUND` and
+        progression in the master CPT. Also the foreground and background colors
+        are exchanged, including those specified by :gmt-term:`COLOR_BACKGROUND` and
         :gmt-term:`COLOR_FOREGROUND`. Set this to **z** to reverse the sign of z-values
         in the color table. Note that this change of z-direction happens before
         ``truncate`` and ``series`` values are used so the latter must be compatible
--- a/pygmt/src/makecpt.py
+++ b/pygmt/src/makecpt.py
@@ -120,10 +120,12 @@ def makecpt(**kwargs):
         level :func:`pygmt.makecpt` is called.
     reverse : str
         Set this to ``True`` or **c** [Default] to reverse the sense of color
-        progression in the master CPT. Set this to **z** to reverse the sign
-        of z-values in the color table. Note that this change of z-direction
-        happens before ``truncate`` and ``series`` values are used so the
-        latter must be compatible with the changed z-range. See also
+        progression in the master CPT. Also exchanges the foreground and background
+        colors, including those specified by :gmt-term:`COLOR_BACKGROUND` and
+        :gmt-term:`COLOR_FOREGROUND`. Set this to **z** to reverse the sign of z-values
+        in the color table. Note that this change of z-direction happens before
+        ``truncate`` and ``series`` values are used so the latter must be compatible
+        with the changed z-range. See also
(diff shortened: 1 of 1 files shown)


seisman:
I feel it's a good time to make the `reverse` parameter more Pythonic. Instead of using `reverse=True`, `reverse="c"` or `reverse="z"`, we may want to use `reverse=True`, `reverse="colors"` and `reverse="zvalues"` instead.

The long argument names `colors` and `zvalues` comes from the upstream PR https://github.com/GenericMappingTools/gmt/pull/8280/files (line 45).

pygmt/src/makecpt.py
```diff
@@ -120,10 +120,12 @@ def makecpt(**kwargs):
...
     reverse : str
         Set this to ``True`` or **c** [Default] to reverse the sense of color
-        progression in the master CPT. Set this to **z** to reverse the sign
-        of z-values in the color table. Note that this change of z-direction
-        happens before ``truncate`` and ``series`` values are used so the
-        latter must be compatible with the changed z-range. See also
+        progression in the master CPT. Also the foreground and background colors
+        are exchanged, including those specified by :gmt-term:`COLOR_BACKGROUND` and
+        :gmt-term:`COLOR_FOREGROUND`. Set this to **z** to reverse the sign of z-values
+        in the color table. Note that this change of z-direction happens before
```
michaelgrund, line 126:
```suggestion
        in the color table. Note that this change in z-direction happens before
```

pygmt/src/makecpt.py
```diff
@@ -120,10 +120,12 @@ def makecpt(**kwargs):
...
+        ``truncate`` and ``series`` values are used so the latter must be compatible
```
michaelgrund, line 127:
```suggestion
        ``truncate`` and ``series`` values are used, so the latter must be compatible
```

pygmt/src/makecpt.py
yvonnefroehlich: Hm. I think I prefer to keep "of" because the z-direction itself is changed, and not something "in" z-direction.
Maybe adding a "the" makes it sound nicer?

```suggestion
        in the color table. Note that this change of the z-direction happens before
```

yvonnefroehlich:
> I feel it's a good time to make the `reverse` parameter more Pythonic. Instead of using `reverse=True`, `reverse="c"` or `reverse="z"`, we may want to use `reverse=True`, `reverse="colors"` and `reverse="zvalues"` instead.
> 
> The long argument names `colors` and `zvalues` comes from the upstream PR https://github.com/GenericMappingTools/gmt/pull/8280/files (line 45).

Sounds fair! Should this be done in this PR or in a separate follow-up PR?

seisman:
> > I feel it's a good time to make the `reverse` parameter more Pythonic. Instead of using `reverse=True`, `reverse="c"` or `reverse="z"`, we may want to use `reverse=True`, `reverse="colors"` and `reverse="zvalues"` instead.
> > The long argument names `colors` and `zvalues` comes from the upstream PR https://github.com/GenericMappingTools/gmt/pull/8280/files (line 45).
> 
> Sounds fair! Should this be done in this PR or in a separate follow-up PR?

Better to do it in this PR, since the docstring needs rewritten for the Pythonic `reverse` parameter.

seisman:
> I feel it's a good time to make the `reverse` parameter more Pythonic. Instead of using `reverse=True`, `reverse="c"` or `reverse="z"`, we may want to use `reverse=True`, `reverse="colors"` and `reverse="zvalues"` instead.
> 
> The long argument names `colors` and `zvalues` comes from the upstream PR https://github.com/GenericMappingTools/gmt/pull/8280/files (line 45).

I prefer to have this PR merged first, and will revisit the issue above later, when we have time focusing on improving the `makecpt` wrapper.

PR #3012, opened 2024-01-17
pygmt.binstats: Let the parameter 'statistic' support descriptive arguments

seisman:
**Description of proposed changes**

Refactor using the new alias system to make the parameter `statistic` support long-form arguments like `statistic="mean"` instead  of `statistic="a"`.

Upstream documentation at https://docs.generic-mapping-tools.org/dev/gmtbinstats.html

Long-form arguments come from https://github.com/GenericMappingTools/gmt/blob/0ae77bb0ae34d6dcb331156d7ee17a747ba83c57/src/longopt/gmtbinstats_inc.h#L29

**Preview**: https://pygmt-dev--3012.org.readthedocs.build/en/3012/api/generated/pygmt.binstats.html

Address #1651.

PR #3013, opened 2024-01-18
Figure.coast/pygmt.select/pygmt.grdlandmask: Use long names ("crude"/"low"/"intermediate"/"high"/"full") for the 'resolution' parameter

seisman:
This PR updates the `resolution` parameter for `Figure.coast`/`pygmt.select`/`pygmt.grdlandmask` to use long names.


seisman:
Ping @GenericMappingTools/pygmt-maintainer for thoughts on the proposed changes.

pygmt/src/select.py
```diff
@@ -116,16 +117,13 @@ def select(
...
-        *resolution*\ [**+f**].
-        Ignored unless ``mask`` is set. Selects the resolution of the coastline
-        data set to use ((**f**)ull, (**h**)igh, (**i**)ntermediate, (**l**)ow,
-        or (**c**)rude). The resolution drops off by ~80% between data sets.
-        [Default is **l**]. Append (**+f**) to automatically select a lower
-        resolution should the one requested not be available [Default is abort
-        if not found]. Note that because the coastlines differ in details
-        it is not guaranteed that a point will remain inside [or outside] when
-        a different resolution is selected.
+    resolution
```
seisman, line 120: Copied from https://docs.generic-mapping-tools.org/dev/gmtselect.html#d, but the `+f` modifier is not mentioned.

pygmt/src/grdlandmask.py
```diff
@@ -44,17 +50,14 @@ def grdlandmask(outgrid: str | None = None, **kwargs) -> xr.DataArray | None:
...
-        ((**f**)ull, (**h**)igh, (**i**)ntermediate, (**l**)ow, or
-        (**c**)rude). The resolution drops off by ~80% between data sets.
-        [Default is **l**]. Append **+f** to automatically select a lower
-        resolution should the one requested not be available
-        [abort if not found]. Alternatively, choose (**a**)uto to automatically
-        select the best resolution given the chosen region. Note that because
-        the coastlines differ in details a node in a mask file using one
-        resolution is not guaranteed to remain inside [or outside] when a
-        different resolution is selected.
+    resolution
```
seisman, line 53: Copied from https://docs.generic-mapping-tools.org/dev/grdlandmask.html#d, but the `+f` modifier is not mentioned.

pygmt/src/coast.py
```diff
@@ -73,10 +81,12 @@ def coast(self, **kwargs):
         parameter. Optionally, specify separate fills by appending
         **+l** for lakes or **+r** for river-lakes, and passing multiple
         strings in a list.
-    resolution : str
-        **f**\|\ **h**\|\ **i**\|\ **l**\|\ **c**.
-        Select the resolution of the data set to: (**f**\ )ull, (**h**\ )igh,
-        (**i**\ )ntermediate, (**l**\ )ow, and (**c**\ )rude.
+    resolution
```
seisman, line 84: Copied from https://docs.generic-mapping-tools.org/dev/coast.html#d, but the +f modifier is not mentioned at all, since it's rarely used. It may not work as documented since in recent GMT versions, GMT can automatically download GSHHG and DCW datasets from the GMT data server.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for thoughts on the proposed changes towards a more Pythonic arguments.

seisman:
Actually, this PR is not needed if PR #3238 is implemented.

seisman:
> Actually, this PR is not needed if PR #3238 is implemented.

It's likely PR #3238 will take months to implement, so I feel it's better to have this PR merged before that PR>

pygmt/src/_common.py
```diff
@@ -251,3 +251,56 @@ def from_params(
             f"{', '.join(params)}."
         )
         raise GMTInvalidInput(msg)
+
+
+def _parse_coastline_resolution(
+    resolution: Literal["auto", "full", "high", "intermediate", "low", "crude", None],
+) -> Literal["a", "f", "h", "i", "l", "c", None]:
```
weiji14, line 258: Thoughts on making a StrEnum out of this?

seisman: What are the benefits? This private function will likely be removed after #3239 is implemented.

seisman: GMT uses short-form arguments, but we prefer long-form arguments in PyGMT. Another example is in #3012. It will be a lot of work if we define StrEnum for all these cases. #3239 proposed a more general solution for it.

pygmt/src/coast.py
```diff
@@ -20,7 +23,6 @@
     A="area_thresh",
     B="frame",
     C="lakes",
-    D="resolution",
```
seisman, line 23: After removing `D="resolution"`, this alias won't be shown in the alias list. Need to wait for #3945.

pygmt/src/_common.py
weiji14: Yeah, `StrEnum` will be a lot more work, and we'll need to bring in something like pydantic (mentioned at https://github.com/GenericMappingTools/pygmt/issues/3239#issuecomment-2126307579) to fully enable runtime checks. Ok to keep this as is for now, and wait for #3239.

weiji14 (approved):

PR #3039, opened 2024-02-05
SPEC0: Set minimum supported version to Python 3.10+

seisman:
Following [SPEC 0](https://scientific-python.org/specs/spec-0000/) policy where Python 3.9 should be dropped in 2023 quarter 4.

Changes in this PR:

- [x] Update the minimum Python version in `ci_tests.yaml`
- [x] Update the minimum Python version in `ci_tests_legacy.yaml`
- [x] Update the minimum Python version in `doc/minversions.rst`
- [x] Update the Python version in the `requires-python` field and remove the related entry from `classifiers` in `pyproject.toml`.
- [ ] Update [branch protection rules](https://github.com/GenericMappingTools/pygmt/settings/branches) 

Supersedes PR #2487.

Address #2863.

PR #3040, opened 2024-02-05
ruff: Fix Python 3.10 violations

weiji14:

Fixes the following ruff violations with Python 3.10+
- [x] B905 [*] zip() without an explicit strict= parameter - https://docs.astral.sh/ruff/rules/zip-without-explicit-strict
- [x] UP035 [*] Import from collections.abc instead: Callable - https://docs.astral.sh/ruff/rules/deprecated-import
- [x] UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)` - https://docs.astral.sh/ruff/rules/non-pep604-isinstance (should we ignore this, because the docs says `X | Y` might be slower than `(X, Y)`?)

Addresses https://github.com/GenericMappingTools/pygmt/pull/3039#issuecomment-1927252931

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -214,7 +214,11 @@ html_context = {
     "doc_path": "doc",
     "galleries": sphinx_gallery_conf["gallery_dirs"],
     "gallery_dir": dict(
-        zip(sphinx_gallery_conf["gallery_dirs"], sphinx_gallery_conf["examples_dirs"])
+        zip(
+            sphinx_gallery_conf["gallery_dirs"],
+            sphinx_gallery_conf["examples_dirs"],
+            strict=True,
+        )
     ),
     "github_repo": repository,
     "github_version": "main",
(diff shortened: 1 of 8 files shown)


weiji14:
> [ ] UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)` - https://docs.astral.sh/ruff/rules/non-pep604-isinstance (should we ignore this, because the docs says `X | Y` might be slower than `(X, Y)`?)

Do we want to ignore UP038, or change to `isinstance(..., X | Y)`? I can run some benchmarks to see if it's slower.

seisman (approved):
Looks good.

weiji14:
Wanted to double-check the flame graphs, so I manually triggered the benchmarks workflow on the main branch to get the baseline (following steps at https://github.com/GenericMappingTools/pygmt/issues/2910#issuecomment-1928722002). Will merge this after I've had a look at the timing comparison.

weiji14:
Hmm, still not seeing the flame graph, but the performance benchmark didn't report the run as noticeably slower https://github.com/GenericMappingTools/pygmt/pull/3040/checks?check_run_id=21254596529, so should be ok.

PR #3037, opened 2024-02-03
Adopt SPEC 0 policy and drop NEP 29 policy

seisman:
The [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) policy is superseded by the [SPEC 0](https://scientific-python.org/specs/spec-0000/) policy.

Address https://github.com/GenericMappingTools/pygmt/issues/2863.

PR #3043, opened 2024-02-06
SPEC 0: Set minimum supported versions to pandas>=1.5 and xarray>=2022.03

### LINKED ISSUE #2863: Migrate from NEP 029 to SPEC 0
The [NEP 029](https://numpy.org/neps/nep-0029-deprecation_policy) page has a note says:

> This NEP is superseded by the scientific python ecosystem coordination guideline [SPEC 0 — Minimum Supported Versions](https://scientific-python.org/specs/spec-0000/).

I haven't read SPEC 0 carefully, but I think we should migrate from NEP 029 to SPEC 0.

seisman:

- [x] `requirements.txt`
- [x] `environment.yml`
- [x] `pyproject.toml`
- [x] `.github/workflows/ci_tests.yaml`
- [x] `doc/minversions.rst` https://pygmt-dev--3043.org.readthedocs.build/en/3043/minversions.html

Closes #2863.

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -62,14 +62,19 @@ jobs:
             isDraft: true
           - os: windows-latest
             isDraft: true
-        # Pair Python 3.10 with NumPy 1.23 and Python 3.12 with NumPy 1.26
-        # Only install optional packages on Python 3.12/NumPy 1.26
+        # Pair Python 3.10 with the minimum supported versions of NumPy/Pandas/Xarray
+        # and Python 3.12 with the latest versions of NumPy/Pandas/Xarray
+        # Only install optional packages on Python 3.12
         include:
           - python-version: '3.10'
             numpy-version: '1.23'
+            pandas-version: '1.5'
+            xarray-version: '2022.03'
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '1.26'
+            pandas-version: '2.2'
+            xarray-version: '2024.01'
             optional-packages: ' contextily geopandas ipython pyarrow rioxarray sphinx-gallery'
 
     timeout-minutes: 30
@@ -107,8 +112,8 @@ jobs:
             gmt=6.5.0
             ghostscript=10.02.1
             numpy=${{ matrix.numpy-version }}
-            pandas
-            xarray
+            pandas=${{ matrix.pandas-version }}
+            xarray=${{ matrix.xarray-version }}
             netCDF4
             packaging
             build
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -68,13 +68,13 @@ jobs:
         include:
           - python-version: '3.10'
             numpy-version: '1.23'
-            pandas-version: '1.5'
-            xarray-version: '2022.03'
+            pandas-version: '=1.5'
+            xarray-version: '=2022.03'
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '1.26'
-            pandas-version: '2.2'
-            xarray-version: '2024.01'
+            pandas-version: ''
+            xarray-version: ''
             optional-packages: ' contextily geopandas ipython pyarrow rioxarray sphinx-gallery'
 
     timeout-minutes: 30
@@ -112,8 +112,8 @@ jobs:
             gmt=6.5.0
             ghostscript=10.02.1
             numpy=${{ matrix.numpy-version }}
-            pandas=${{ matrix.pandas-version }}
-            xarray=${{ matrix.xarray-version }}
+            pandas${{ matrix.pandas-version }}
+            xarray${{ matrix.xarray-version }}
             netCDF4
             packaging
             build
--- a/doc/minversions.rst
+++ b/doc/minversions.rst
@@ -15,77 +15,77 @@ the rest of the Scientific Python ecosystem, and therefore:
       - GMT
       - Python
       - Core package dependencies
-    * - `Dev <https://github.com/GenericMappingTools/pygmt/milestones>`_* [`Documentation <https://www.pygmt.org/dev>`__]
+    * - `Dev <https://github.com/GenericMappingTools/pygmt/milestones>`_* [`Docs <https://www.pygmt.org/dev>`__]
       - >=6.3.0
       - >=3.10
       - - numpy>=1.23
         - pandas>=1.5
         - xarray>=2022.03
-    * - `v0.11.0 <https://github.com/GenericMappingTools/pygmt/releases/tag/v0.11.0>`_ [`Documentation <https://www.pygmt.org/v0.11.0>`__]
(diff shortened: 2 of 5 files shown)


.github/workflows/ci_tests.yaml
```diff
@@ -62,14 +62,19 @@ jobs:
...
         include:
           - python-version: '3.10'
             numpy-version: '1.23'
+            pandas-version: '1.5'
+            xarray-version: '2022.03'
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '1.26'
+            pandas-version: '2.2'
+            xarray-version: '2024.01'
```
seisman, line 77: Following what we are doing with NumPy 1.26, we should also pin pandas and xarray to their latest versions here. But, it means we have to watch the releases of these core package dependencies and update these versions very often (they usually have one minor release every few months).

So, instead of tracking the latest releases, we can just pin Python to the latest version, and conda/mamba should automatically install the latest numpy/pandas/xarray versions that are compatible with the pinned Python version. Does it sounds OK?

doc/minversions.rst
```diff
@@ -15,62 +15,64 @@ the rest of the Scientific Python ecosystem, and therefore:
       - Documentation
       - GMT
       - Python
-      - NumPy
+      - Core package dependencies
```
seisman, line 18: In https://github.com/GenericMappingTools/pygmt/pull/3037#discussion_r1479035637, @weiji14 proposed to add two more columns for pandas and xarray. I think this is not good, because the table is already too wide. I prefer to list the core dependencies in a single column. See https://pygmt-dev--3043.org.readthedocs.build/en/3043/minversions.html for preview.

I also prefer to combine the "Python version" and "Documentation" columns into a single column. The new table will look like:

image

Thoughts?

.github/workflows/ci_tests.yaml
weiji14: Ok with not testing the newest versions for `pandas` and `xarray`, but we'll still need to set a minimum pin for the older version I think.

```suggestion
            pandas-version: ''
            xarray-version: ''
```

.github/workflows/ci_tests.yaml
```diff
@@ -62,14 +62,19 @@ jobs:
...
-        # Pair Python 3.10 with NumPy 1.23 and Python 3.12 with NumPy 1.26
-        # Only install optional packages on Python 3.12/NumPy 1.26
+        # Pair Python 3.10 with the minimum supported versions of NumPy/Pandas/Xarray
+        # and Python 3.12 with the latest versions of NumPy/Pandas/Xarray
+        # Only install optional packages on Python 3.12
         include:
           - python-version: '3.10'
             numpy-version: '1.23'
+            pandas-version: '1.5'
+            xarray-version: '2022.03'
```
weiji14, line 72: Try setting the equal-to sign here
```suggestion
            pandas-version: '=1.5'
            xarray-version: '=2022.03'
```

.github/workflows/ci_tests.yaml
```diff
@@ -107,8 +112,8 @@ jobs:
             gmt=6.5.0
             ghostscript=10.02.1
             numpy=${{ matrix.numpy-version }}
-            pandas
-            xarray
+            pandas=${{ matrix.pandas-version }}
+            xarray=${{ matrix.xarray-version }}
```
weiji14, line 116: Remove equal-to sign here
```suggestion
            pandas${{ matrix.pandas-version }}
            xarray${{ matrix.xarray-version }}
```

doc/minversions.rst
weiji14: > I also prefer to combine the "Python version" and "Documentation" columns into a single column.

Yep, ok with this part. We can add a note at the bottom to say that 'Dev' reflects the main branch and is for the upcoming release.

> In [#3037 (comment)](https://github.com/GenericMappingTools/pygmt/pull/3037#discussion_r1479035637), @weiji14 proposed to add two more columns for pandas and xarray. I think this is not good, because the table is already too wide. I prefer to list the core dependencies in a single column.

Once we combine the PyGMT and Docs columns, maybe there's enough space for separate numpy/pandas/xarray columns? Could you try splitting into individual columns to see how it looks?

seisman: Done in 5f22ecb0d86be78634a39de8bcf91c9378ed60cb. Not bad. See preview at https://pygmt-dev--3043.org.readthedocs.build/en/3043/minversions.html.

weiji14 (approved):
Great, thanks @seisman!

seisman:
After pinning pandas and xarray with Python 3.10, now we have many failures on macOS/Windows. See https://github.com/GenericMappingTools/pygmt/actions/runs/7946102470/job/21694725831. Need to find out why.

weiji14:
> After pinning pandas and xarray with Python 3.10, now we have many failures on macOS/Windows. See https://github.com/GenericMappingTools/pygmt/actions/runs/7946102470/job/21694725831. Need to find out why.

[`pytest-mpl=0.17.0`](https://github.com/matplotlib/pytest-mpl/releases/tag/v0.17.0) now adds a `metadata` key into our `savefig` call (added in https://github.com/matplotlib/pytest-mpl/pull/196/files I think), and this is not handled by psconvert. We'll need to ignore that `metadata` key somehow.

seisman:
> > After pinning pandas and xarray with Python 3.10, now we have many failures on macOS/Windows. See https://github.com/GenericMappingTools/pygmt/actions/runs/7946102470/job/21694725831. Need to find out why.
> 
> [`pytest-mpl=0.17.0`](https://github.com/matplotlib/pytest-mpl/releases/tag/v0.17.0) now adds a `metadata` key into our `savefig` call (added in https://github.com/matplotlib/pytest-mpl/pull/196/files I think), and this is not handled by psconvert. We'll need to ignore that `metadata` key somehow.

Thanks. Will address the issue after this PR is merged.

doc/minversions.rst
```diff
@@ -8,101 +8,123 @@ the rest of the Scientific Python ecosystem, and therefore:
...
     * - PyGMT Version
-      - Documentation
       - GMT
       - Python
       - NumPy
-    * - `Dev <https://github.com/GenericMappingTools/pygmt/milestones>`_ (upcoming release)
-      - `Dev Documentation <https://www.pygmt.org/dev>`_ (reflects `main branch <https://github.com/GenericMappingTools/pygmt>`_)
+      - Pandas
+      - Xarray
+    * - `Dev <https://github.com/GenericMappingTools/pygmt/milestones>`_ * [`Docs <https://www.pygmt.org/dev>`__]
```
yvonnefroehlich, line 21: Is there a need that the first link ends with one underscore (`_`) but the second link with two underscores (`__`)?
```suggestion
    * - `Dev <https://github.com/GenericMappingTools/pygmt/milestones>`_ * [`Docs <https://www.pygmt.org/dev>`_]
```

seisman: See https://stackoverflow.com/questions/27420317/restructured-text-rst-http-links-underscore-vs-use for explanations.

For "Docs", we have to use `__` to make the links private.

yvonnefroehlich: Ah, interesting. I did not know this. Thanks for sharing this link!

yvonnefroehlich (approved):

Issue #3033, opened 2024-02-02
Remove the deprecated parameter and alias names in v0.12.0


michaelgrund:
Several parameters and a few aliases are deprecated since v0.8.0, v0.9.0 or v0.10.0 and the deprecation warnings need to be removed before release v0.12.0.

This issue is created to track the progress of removing these deprecation warnings:

- [x] All plotting methods: Remove aliases `xshift (X)` and `yshift (Y)` (FutureWarning raised since PyGMT v0.8.0) https://github.com/GenericMappingTools/pygmt/pull/3044
- [x] All plotting methods: Remove alias `timestamp (U)` (FutureWarning raised since PyGMT v0.9.0) https://github.com/GenericMappingTools/pygmt/pull/3045
- [x]  `Figure.plot`, `Figure.plot3d`, `Figure.rose`: Remove parameter `color`, use `fill` instead (FutureWarning raised since PyGMT v0.8.0) #3032 
- [x]    `Figure.velo`:  #3034 
   - Remove parameter `uncertaintycolor`, use `uncertaintyfill` instead (FutureWarning raised since PyGMT v0.8.0) 
   - Remove parameter `color`, use `fill` instead
- [x]    `Figure.wiggle`: Remove parameter `color`, use `fillnegative` or `fillpositive` instead (FutureWarning raised since PyGMT v0.8.0) #3035
- [x]    `Figure.grdimage`: Remove parameter `bit_color`, use `bitcolor` instead (Future Warning raised since PyGMT v0.10.0) #3036

weiji14:
Thanks @michaelgrund for keeping track! Ok to open the PRs, but let's wait for a week or two (around mid-Feb) before merging, in case there's some serious bug with PyGMT v0.11.0 and we need to release v0.11.1 to patch it.

michaelgrund:
Sure @weiji14 , definitely fine for me.

seisman:
> let's wait for a week or two (around mid-Feb) before merging, in case there's some serious bug with PyGMT v0.11.0 and we need to release v0.11.1 to patch it.

We don't have to wait. We can always create a patch release from the v0.11.0 tag if there are any serious bugs.

weiji14:
> We don't have to wait. We can always create a patch release from the v0.11.0 tag if there are any serious bugs.

Sure, we can do that, just slightly more work since we need a commit on the main branch, and a backport commit on a patch branch going off from v0.11.0.

PR #3045, opened 2024-02-06
All plotting methods: Remove deprecated "timestamp" ("U") parameter, use "Figure.timestamp" instead (deprecated since v0.9.0)

yvonnefroehlich:

This PR removes the deprecated `timestampe` (`U`)  parameter from all plotting methods (deprecated since v0.9.0).
The standalone method `Figure.timestamp` should be used instead.

Fixes partly #3033 
Related to #2135

**Reminders**

**Slash Commands**


pygmt/helpers/decorators.py
```diff
@@ -567,17 +567,6 @@ def new_module(*args, **kwargs):
                     )
                     warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
 
-            # timestamp (U) is deprecated since v0.9.0.
```
seisman, line 570: After this PR, the expected behaviors are: 

- if `timestamp=True` is used, Python will report the parameter is not recognized
- if single-letter parameter `U=True` is used, PyGMT should tell users that `U` is not supported and `Figure.timestamp` should be used.

After removing lines 570-580, parameter `timestamp` is no longer recognized, but single-letter parameter `U` will still be processed. Thus, instead of removing these lines completely, we should change them to something like:

```python
            # timestamp (U)is deprecated since v0.9.0.
            if "U" in kwargs:
                msg = (
                    "Parameter timestamp/U is no longer supported since v0.12.0."
                    "Use Figure.timestamp() instead."
                )
                raise GMTInvalidInput(msg)
```

yvonnefroehlich: Ah, OK. Thanks! I have tried to change this in the commits https://github.com/GenericMappingTools/pygmt/pull/3045/commits/6977a57302681307239eafca840342a00ff07628, https://github.com/GenericMappingTools/pygmt/pull/3045/commits/1501ccc9291948b6765cb6dd140650ebf1d6e241, and https://github.com/GenericMappingTools/pygmt/pull/3045/commits/42fab3b1e99a8d0ab66d4daca8705190f68c0736. I also changed this in PR #3044 for `xshift` (`X`) and `yshift` (`Y`).

pygmt/tests/test_timestamp.py
```diff
@@ -93,38 +93,3 @@ def test_timestamp_text_truncated():
...
-    """
-    Check if the deprecated parameter 'U' works but raises a warning.
-    """
-    fig = Figure()
-    with pytest.warns(expected_warning=SyntaxWarning) as record:
-        with config(FORMAT_TIME_STAMP=faketime):
-            # plot nothing (the data is outside the region) but a timestamp
-            fig.plot(x=0, y=0, style="p", projection="X1c", region=[1, 2, 1, 2], U=True)
-        assert len(record) == 1  # check that only one warning was raised
-    return fig
```
seisman, line 130: Need to modify this test to check that the GMTInvalidInput exception is raised if `U` is used.

pygmt/tests/test_timestamp.py
```diff
@@ -93,38 +93,3 @@ def test_timestamp_text_truncated():
         fig.timestamp(text="0123456789" * 7)
         assert len(record) == 1  # check that only one warning was raised
     return fig
```
seisman, line 95:
```suggestion
    return fig
    

def test_timestamp_unsupported_u_timestamp():
    """
    Raise an exception when either U or timestamp is used.

    Parameters U and timestamp are no longer supported since v0.12.0
    """
    with pytest.raises(GMTInvalidInput):
        fig.plot(x=0, y=0, style="p", projection="X1c", region=[1, 2, 1, 2], U=True)

    with pytest.raises(GMTInvalidInput):
        fig.plot(x=0, y=0, style="p", projection="X1c", region=[1, 2, 1, 2], timestamp=True)
```

pygmt/tests/test_timestamp.py
seisman: See https://github.com/GenericMappingTools/pygmt/pull/3045/files#r1502738446.

seisman (approved):

PR #3055, opened 2024-02-19
CI: Use 'git ls-files' to list all git tracked files

seisman:
We used the `find` command to find all files in the current directory and exclude the `.git` directory. The `find` command is very long and difficult to understand/maintain. 

Instead, we can use `git ls-files` to list all git-tracked files.

ref: https://stackoverflow.com/a/41085442

Patches https://github.com/GenericMappingTools/pygmt/pull/736


weiji14 (approved):
Much more simpler, thanks!

PR #3057, opened 2024-02-20
clib: Add the virtualfile_out method for creating output virtualfile

seisman:
Currently, PyGMT has provided several methods for manipulating virtual files. These methods are:

- `open_virtualfile`: Open a virtual file and close it when done.
- `virtualfile_from_vectors`: Create a virtual file for a table file from 1-D arrays
- `virtualfile_from_matrix`: Create a virtual file for a table file from a 2-D array
- `virtualfile_from_grid`: Create a virtual file for a grid file from a xarray.DataArray object
- `virtualfile_from_data`: Convenience function which wraps `virtualfile_from_vectors`/`virtualfile_from_matrix`/`virtualfile_from_grid` and more
- `read_virtualfile`: Read data from a virtual file

All the `virtualfile_from_*` methods are for input data. This PR provides a new method `virtualfile_out` which can create a virtual file for output data. This method is necessary to get rid of temporary files for modules that write tables/grids (#2730).

The same codes are already used in PR #2729 and #2398. This PR is created to make those two PRs smaller and easier to review.

**Preview**: https://pygmt-dev--3057.org.readthedocs.build/en/3057/api/generated/pygmt.clib.Session.virtualfile_out.html

**Reminders**

**Slash Commands**


seisman:
We probably need to think more about the function name. Currently, I use `virtualfile_to_data`, similar to the `virtualfile_from_data` function. Here, `from` means input and `to` means output. `virtualfile_from_data` sounds like creating a virtual file from an *input* data, so the function name makes sense, but `virtualfile_to_data` sounds like converting a virtual file to data, which makes little sense.

Maybe `virtualfile_in`/`virtualfile_out` are better names for `virtualfile_from_data`/`virtualfile_to_data`? Or even the much shorter version `vfile_in`/`vfile_out`?

seisman:
> We probably need to think more about the function name. Currently, I use `virtualfile_to_data`, similar to the `virtualfile_from_data` function. Here, `from` means input and `to` means output. `virtualfile_from_data` sounds like creating a virtual file from an _input_ data, so the function name makes sense, but `virtualfile_to_data` sounds like converting a virtual file to data, which makes little sense.
> 
> Maybe `virtualfile_in`/`virtualfile_out` are better names for `virtualfile_from_data`/`virtualfile_to_data`? Or even the much shorter version `vfile_in`/`vfile_out`?

Renamed these functions to `virtualfile_in`/`virtualfile_out` in commits a633abf2d012378b7f6f6a6a82f3795022eaf7e9 and dfdd3eda67b8a3e781ab853f19b498c2bba15eef, but I'm still open to comments about the function names.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for a final review. Otherwise, I'll merge this PR in 24 hours so we can move forward.

pygmt/clib/session.py
```diff
@@ -1609,6 +1609,79 @@ def virtualfile_from_data(  # noqa: PLR0912
...
+        >>>
+        >>> # Write data to file without creating a virtual file
+        >>> with GMTTempFile(suffix=".nc") as tmpfile:
+        ...     with Session() as lib:
+        ...         with lib.virtualfile_out(
+        ...             kind="grid", fname=tmpfile.name
+        ...         ) as voutgrd:
+        ...             lib.call_module("read", f"@earth_relief_01d_g {voutgrd} -Tg")
+        ...             assert voutgrd == tmpfile.name
+        ...             assert Path(voutgrd).stat().st_size > 0
```
weiji14, line 1672: Should we move this into a new `test_clib_virtualfile_out.py` file? Two reasons:

1. There might be other data types like [`GMT_IMAGE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#struct-image) and [`GMT_CUBE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#gmt-cube) in the future, and the doctest would get long.
2. Might want to test the output of the virtualfiles more properly using `pygmt.info` or `pygmt.grdinfo`.

seisman: > Should we move this into a new `test_clib_virtualfile_out.py` file? Two reasons:
> 
> 1. There might be other data types like [`GMT_IMAGE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#struct-image) and [`GMT_CUBE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#gmt-cube) in the future, and the doctest would get long.

The main purpose of these doctests are to show how to use this method and the *outpu* virtual file in other modules. So I think we can only keep the `dataset` one only and remove all others.

> 2. Might want to test the output of the virtualfiles more properly using `pygmt.info` or `pygmt.grdinfo`.

Actually we can't. Because the virtual file is an output (i.e., `GMT_OUT`), it can't be passed as an *input* to `info`/`grdinfo`. I.e., `lib.call_module("info", f"{vouttbl}")` won't work.

weiji14: > So I think we can only keep the `dataset` one only and remove all others.

You can move the grid one to `test_clib_virtualfile_out.py`

> > 3\. Might want to test the output of the virtualfiles more properly using `pygmt.info` or `pygmt.grdinfo`.
> 
> Actually we can't. Because the virtual file is an output (i.e., `GMT_OUT`), it can't be passed as an _input_ to `info`/`grdinfo`. I.e., `lib.call_module("info", f"{vouttbl}")` won't work.

Ah ok, maybe not for the virtualfile then. But we should test that saving to a (non-virtual) file works, and try loading from there (i.e. roundtrip). Again, can put this in `test_clib_virtualfile_out.py`

seisman: > > Should we move this into a new `test_clib_virtualfile_out.py` file? Two reasons:
> > 
> > 1. There might be other data types like [`GMT_IMAGE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#struct-image) and [`GMT_CUBE`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#gmt-cube) in the future, and the doctest would get long.
> 
> The main purpose of these doctests are to show how to use this method and the _outpu_ virtual file in other modules. So I think we can only keep the `dataset` one only and remove all others.

In commit 9108917d5cd73bbc8897937636e8a3252ff7e50f, I've removed the doctest for `grid` and kept the doctests for `dataset` and `fname`. I also realized that these doctests are almost the same as the ones in `read_virtualfile`. This makes sense, because `read_virtualfile` was implemented to read the data from an *output* virtualfile. So, I also simplify the doctests in `read_virtualfile` using the newly added `virtualfile_out` method.

In other words, we probably need to move the doctests of `read_virtualfile` into a separate test file in the future.

seisman: > > So I think we can only keep the `dataset` one only and remove all others.
> 
> You can move the grid one to `test_clib_virtualfile_out.py`

As mentioned in the above comment, the grid one is already doctested in the `read_virtualfile` method.

> > > 3. Might want to test the output of the virtualfiles more properly using `pygmt.info` or `pygmt.grdinfo`.
> > 
> > 
> > Actually we can't. Because the virtual file is an output (i.e., `GMT_OUT`), it can't be passed as an _input_ to `info`/`grdinfo`. I.e., `lib.call_module("info", f"{vouttbl}")` won't work.
> 
> Ah ok, maybe not for the virtualfile then. But we should test that saving to a (non-virtual) file works (when `fname` is used), and try loading from there (i.e. roundtrip). Again, can put this in `test_clib_virtualfile_out.py`

Done in 888f1c9c96b4de496aa2b80368c70fab45ccc1ea.

weiji14 (approved):

PR #3072, opened 2024-02-27
Use pip instead of micromamba to list the installed packages

seisman:
The `micromamba list` command only lists packages from pypi so we can't know the exact packages that are installed. See https://github.com/mamba-org/mamba/issues/2059 for the upstream issue report and https://github.com/GenericMappingTools/pygmt/actions/runs/8042005126/job/21962067095 for the CI job run.

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -126,7 +126,7 @@ jobs:
 
       # Show installed pkg information for postmortem diagnostic
       - name: List installed packages
-        run: micromamba list
+        run: python -m pip list
 
       # Pull baseline image data from dvc remote (DAGsHub)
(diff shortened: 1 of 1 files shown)


seisman:
The `pip list` output is also not perfect.
```
pip list
Package                       Version                            Editable project location
----------------------------- ---------------------------------- ---------------------------
affine                        2.4.0
aiohttp                       3.9.1
aiohttp-retry                 2.8.3
aiosignal                     1.3.1
alabaster                     0.7.13
amqp                          5.2.0
annotated-types               0.6.0
antlr4-python3-runtime        4.9.3
```
Let's wait for the upstream fix at https://github.com/mamba-org/mamba/pull/2104.

seisman:
There is nothing we can do on our side. Closing the issue to reduce the number of opened PRs.

Issue #2185, opened 2022-11-10
No aliases lists given in docs for some functions or methods


yvonnefroehlich:
There are no aliases lists given in the documentation for the class [pygmt.triangulate](https://www.pygmt.org/dev/api/generated/pygmt.triangulate.html#pygmt.triangulate) and its methods
- [pygmt.triangulate.regular_grid](https://www.pygmt.org/dev/api/generated/pygmt.triangulate.regular_grid.html)
- [pygmt.triangulate.delaunay_triples](https://www.pygmt.org/dev/api/generated/pygmt.triangulate.delaunay_triples.html)

Is this intended?

----------------------
_Edit (see comment by @weiji14 below)_: This is simliar for the class [pygmt.grdhisteq](https://www.pygmt.org/dev/api/generated/pygmt.grdhisteq.html#pygmt.grdhisteq) and its methods [pygmt.grdhisteq.compute_bins](https://www.pygmt.org/dev/api/generated/pygmt.grdhisteq.compute_bins.html#pygmt.grdhisteq.compute_bins) and [pygmt.grdhisteq.equalize_grid](https://www.pygmt.org/dev/api/generated/pygmt.grdhisteq.equalize_grid.html#pygmt.grdhisteq.equalize_grid).

seisman:
Ping @weiji14

weiji14:
The `regular_grid` and `delaunay_triples` functions don't use the `@use_alias` decorator as seen here:

https://github.com/GenericMappingTools/pygmt/blob/e4e4c43946e43b7f03f10bfc7155d6e9f81cccfc/pygmt/src/triangulate.py#L151-L153

https://github.com/GenericMappingTools/pygmt/blob/e4e4c43946e43b7f03f10bfc7155d6e9f81cccfc/pygmt/src/triangulate.py#L277-L279

so the [`{aliases}` shortcode](https://github.com/GenericMappingTools/pygmt/blob/e4e4c43946e43b7f03f10bfc7155d6e9f81cccfc/pygmt/helpers/decorators.py#L456-L461) won't work.

We could discuss whether it's worth spending time to add the alias list, though it would be quite involved I think (also considering the eventual goal is to drop single character aliases, xref #262).

[...]

PR #3073, opened 2024-02-27
pygmt.triangulate: Refactor to use codes consistent with other wrappers

seisman:

The `triangulate` module was first wrapped in PR #731. As mentioned in PR #731:

> The complicated thing about `triangulate` is that it generates either text (table) or NetCDF (grid) outputs, and I've implemented it similar to `grdhisteq` at #1433 as so:
> 
> - `triangulate.delaunay_triples`:
>   - [x] When `output_type="numpy"`, output table as `numpy.ndarray`
>   - [x] When `output_type="pandas"` (default for table), output table as `pandas.DataFrame`
>   - [x] When `output_type="file"` and `outfile is not None`, output table as textfile 
> - `triangulate.regular_grid`:
>   - [x] When `outgrid=True` (default for grid), output grid as 'xarray.DataArray`
>   - [x] When `outgrid="test.nc"`, output grid to NetCDF file

So, in PR #731, a private function `_triangulate` was implemented and was internally called by `triangulate.delaunay_triples`/`triangulate.regular_grid`.

Such an implementation has some limitations:

- The internal `_triangulate` function is complicated because it needs to deal with both table and grid outputs.
- We have to maintain similar docstrings in three functions (although the private `_triangulate` function doesn't have to have docstrings)
- In `triangulate.delaunay_triples`/`triangulate.regular_grid`, we have to call the private function like:
```
            return triangulate._triangulate(
                data=data,
                x=x,
                y=y,
                z=z,
                output_type=output_type,
                outgrid=outgrid,
                spacing=spacing,
                projection=projection,
                region=region,
                verbose=verbose,
                binary=binary,
                nodata=nodata,
                find=find,
                coltypes=coltypes,
                header=header,
                incols=incols,
                registration=registration,
                skiprows=skiprows,
                wrap=wrap,
                **kwargs,
            )
```
- Aliases are not listed in `triangulate.delaunay_triples`/`triangulate.regular_grid` documentation (as reported in #2185)

This PR refactors the `triangulate` wrapper by removing the private `_triangulate` function. The most obvious benefits of the new implementation are that:

- The `triangulate.delaunay_triples` method have codes that look like any table-ouput wrappers (e.g., `filter1d`)
- The `triangulate.regular_grid` method has codes that look like any grid-output wrappers (e.g., `grd2xyz`)

Thus, the new implementation is much easier to maintain and also make it easier to refactor in PR #2729. 

**Preview**:

- https://pygmt-dev--3073.org.readthedocs.build/en/3073/api/generated/pygmt.triangulate.html
- https://pygmt-dev--3073.org.readthedocs.build/en/3073/api/generated/pygmt.triangulate.regular_grid.html#pygmt.triangulate.regular_grid
- https://pygmt-dev--3073.org.readthedocs.build/en/3073/api/generated/pygmt.triangulate.delaunay_triples.html#pygmt.triangulate.delaunay_triples

### DIFF
--- a/pygmt/src/triangulate.py
+++ b/pygmt/src/triangulate.py
@@ -5,7 +5,6 @@ Cartesian data.
 
 import pandas as pd
 from pygmt.clib import Session
-from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import (
     GMTTempFile,
     build_arg_string,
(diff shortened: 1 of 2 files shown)


weiji14 (approved):
Looks fine after a quick skim (a little busy this week), and the tests pass, so should be ok.

PR #3078, opened 2024-02-29
TYP: Add type hints to the clib.loading.clib_full_names/load_libgmt functions

seisman:
Add type hints and simplify codes using the walrus operator.

### DIFF
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -17,29 +17,29 @@ from pathlib import Path
 from pygmt.exceptions import GMTCLibError, GMTCLibNotFoundError, GMTOSError
 
 
-def load_libgmt(lib_fullnames=None):
+def load_libgmt(lib_fullnames: Iterator[str] | None = None) -> ctypes.CDLL:
     """
     Find and load ``libgmt`` as a :py:class:`ctypes.CDLL`.
 
     Will look for the GMT shared library in the directories determined by
-    clib_full_names().
+    ``clib_full_names()``.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


pygmt/clib/loading.py
```diff
@@ -96,20 +97,27 @@ def clib_names(os_name: str) -> list[str]:
     return libnames
 
 
-def clib_full_names(env=None):
+def clib_full_names(env: os._Environ | dict | None = None) -> Iterator[str]:
```
weiji14, line 100: Not sure about using a private `os_Environ` here...

seisman: `os.environ` returns a mapping object, so change it to `collections.abc.Mapping` in 3212ac94f3f79897a11cd90ff97a06323c752447

pygmt/clib/loading.py
```diff
@@ -96,20 +97,27 @@ def clib_names(os_name: str) -> list[str]:
     return libnames
 
 
-def clib_full_names(env=None):
+def clib_full_names(env: Mapping | dict | None = None) -> Iterator[str]:
```
weiji14, line 100: `collections.abc.Mapping` includes `dict`, see https://stackoverflow.com/questions/35690572/what-does-isinstance-with-a-dictionary-and-abc-mapping-from-collections-doing. Maybe we can remove the `dict` part?

```suggestion
def clib_full_names(env: Mapping | None = None) -> Iterator[str]:
```

weiji14 (approved):
One minor suggestion, pre-approving for now.

PR #3068, opened 2024-02-26
clib: Rename the virtualfile_from_data method to virtualfile_in

seisman:
Rename `virtualfile_from_data` to `virtualfile_in`. The old name is still supported but will be removed in the future.

Address https://github.com/GenericMappingTools/pygmt/pull/3057#issuecomment-1963629238

PR #3083, opened 2024-03-01
clib: Add virtualfile_to_dataset method for converting virtualfile to a dataset

seisman:
Add the new `Session.virtualfile_to_dataset()` method to simplify the logic of dealing with table outputs in different wrappers. Draft PR #3092 shows how the new `Session.virtualfile_to_dataset()` method can simplify other wrappers.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -294,6 +294,7 @@ conversion of Python variables to GMT virtual files:
     clib.Session.virtualfile_from_grid
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
+    clib.Session.return_table
 
 Low level access (these are mostly used by the :mod:`pygmt.clib` package):
 
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -294,7 +294,7 @@ conversion of Python variables to GMT virtual files:
     clib.Session.virtualfile_from_grid
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
-    clib.Session.return_dataset
+    clib.Session.virtualfile_to_dataset
 
 Low level access (these are mostly used by the :mod:`pygmt.clib` package):
 
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -294,7 +294,7 @@ conversion of Python variables to GMT virtual files:
     clib.Session.virtualfile_from_grid
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
-    clib.Session.return_table
+    clib.Session.return_dataset
 
 Low level access (these are mostly used by the :mod:`pygmt.clib` package):
 
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -283,8 +283,8 @@ the :meth:`~pygmt.clib.Session.call_module` method:
 
 Passing memory blocks between Python data objects (e.g. :class:`numpy.ndarray`,
 :class:`pandas.Series`, :class:`xarray.DataArray`, etc) and GMT happens through
-*virtual files*. These methods are context managers that automate the
-conversion of Python variables to GMT virtual files:
+*virtual files*. These methods are context managers that automate the conversion of
+Python variables to and from GMT virtual files:
 
 .. autosummary::
     :toctree: generated
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -284,7 +284,7 @@ the :meth:`~pygmt.clib.Session.call_module` method:
 Passing memory blocks between Python data objects (e.g. :class:`numpy.ndarray`,
 :class:`pandas.Series`, :class:`xarray.DataArray`, etc) and GMT happens through
 *virtual files*. These methods are context managers that automate the conversion of
-Python variables to and from GMT virtual files:
+Python objects to and from GMT virtual files:
 
 .. autosummary::
     :toctree: generated
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1741,7 +1741,7 @@ class Session:
     def return_table(
         self,
         output_type: Literal["pandas", "numpy", "file"],
-        vfile: str | None = None,
+        vfile: str,
         column_names: list[str] | None = None,
     ) -> pd.DataFrame | np.ndarray | None:
         """
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1741,7 +1741,7 @@ class Session:
     def virtualfile_to_dataset(
         self,
         output_type: Literal["pandas", "numpy", "file"],
-        vfile: str,
+        vfname: str,
         column_names: list[str] | None = None,
     ) -> pd.DataFrame | np.ndarray | None:
         """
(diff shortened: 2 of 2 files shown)


pygmt/clib/session.py
```diff
@@ -1738,6 +1738,119 @@ def read_virtualfile(
         dtype = {"dataset": _GMT_DATASET, "grid": _GMT_GRID}[kind]
         return ctp.cast(pointer, ctp.POINTER(dtype))
 
+    def return_table(
```
seisman, line 1741: The method name `return_table` was initially proposed in https://github.com/GenericMappingTools/pygmt/issues/1318#issuecomment-861402119, but is it a good name? At line 1845, we use `kind="dataset"`, so maybe rename it to `return_dataset` or `output_dataset`? In the future, I think we will add more methods that return a grid/image/cpt/cube, so consistent method names are preferred.

seisman: I think `return_dataset` is better than `return_table`. Renamed in ce029b252a8b29dd36dcad2bfeedef36a88c8ee8.

pygmt/clib/session.py
```diff
@@ -1738,6 +1738,123 @@ def read_virtualfile(
         dtype = {"dataset": _GMT_DATASET, "grid": _GMT_GRID}[kind]
         return ctp.cast(pointer, ctp.POINTER(dtype))
 
+    def return_dataset(
+        self,
+        output_type: Literal["pandas", "numpy", "file"],
```
weiji14, line 1743: Should we set a default output type here? It looks like we're using `pandas` as the default in #3092.

```suggestion
        output_type: Literal["pandas", "numpy", "file"] = "pandas",
```

pygmt/clib/session.py
```diff
@@ -1738,6 +1738,123 @@ def read_virtualfile(
...
+        vfile: str,
+        column_names: list[str] | None = None,
+    ) -> pd.DataFrame | np.ndarray | None:
+        """
+        Output a dataset stored in a virtual file in different formats.
```
weiji14, line 1748:
```suggestion
        Output a tabular dataset stored in a virtual file to a different format.
```

pygmt/clib/session.py
weiji14: How about `virtualfile_to_dataset`, or `vfile_to_dataset` (shorter)?

pygmt/clib/session.py
seisman: It makes no differences because we always call the function with the `output_type` parameter, e.g.,:
```
        return lib.return_dataset(
            output_type=output_type,
            vfile=vouttbl,
        )
```

pygmt/clib/session.py
seisman: `virtualfile_to_dataset`/`vfile_to_dataset` sounds a good name. I like `vfile_to_dataset`, but do we want to also rename other functions for consistency?

pygmt/clib/session.py
weiji14: Yes, it doesn't make any difference in the PyGMT modules, but this is a good central location to document that `output_type="pandas"` is the default output (though in #1318, it seemed like most of us were in favour of `output_type="input"` or auto as the default).

pygmt/clib/session.py
weiji14: You mean rename the `virtualfile_from_*` methods? Let's maybe not do that (lazy to deprecate more names). We can go with `virtualfile_to_dataset` for consistency.

seisman: I meant to rename the `virtualfile_in`/`virtualfile_out` to `vfile_in`/`vfile_out`.

- `virtualfile_from_*` functions are no longer used in our PyGMT wrappers after #949, so it's OK to keep them unchanged.
- `virtualfile_out` was added in #3057, so we can still rename it without deprecations.
- `virtualfile_in` was renamed from `virtualfile_from_data` recently (PR #3068), so we can rename it to `vfile_in` without more deprecations.

pygmt/clib/session.py
seisman: `output_type="input"` or `auto` may not make sense for PyGMT, especially in cases like:

1. the input data is a file, then `auto` means outputting to a file by default, then `outfile` is required.
2. the input data is vectors (e.g., `x`/`y`/`z`) and each vector can be a list/ndarray/pd.Series. Then what's the expected format if `auto`/`input` is used?

pygmt/clib/session.py
weiji14: Let's keep with the long name (`virtualfile_to_dataset`). Renaming `virtualfile_out` -> `vfile_out` and `virtualfile_in` -> `vfile_in` will be more work and confusing (we'll need to manually update the changelog).

pygmt/clib/session.py
weiji14: Yeah, not saying that `output_type="auto"` would be easy to implement :slightly_smiling_face: I think the default `output_type="pandas"` is fine for now, and we can discuss more about what the ideal output type would be in #1318 (if there is still any debate that needs to be had).

pygmt/clib/session.py
seisman: Done in 711142c8816c91f7de51ebfb548990ce3a51731e.

doc/api/index.rst
```diff
@@ -294,6 +294,7 @@ conversion of Python variables to GMT virtual files:
     clib.Session.virtualfile_from_grid
     clib.Session.virtualfile_in
     clib.Session.virtualfile_out
+    clib.Session.virtualfile_to_dataset
```
weiji14, line 297: At L287 above, could you change the sentence to read "These methods are context managers that automate the
conversion of Python variables to **and from** GMT virtual files"? Since we can convert GMT virtualfiles to Python objects now.

seisman: Done in aee0499f7d01f81a35306c8133dc5c6cbba22615.

seisman:
In commit cd5b31d6d7f205cfdcdf7ceeb1ab0aea2da5a13f, I've changed the parameter name `vfile` to `vfname` to make it consistent with the `virtualfile_out` and `read_virtualfile` methods.

doc/api/index.rst
```diff
@@ -283,8 +283,8 @@ the :meth:`~pygmt.clib.Session.call_module` method:
 
 Passing memory blocks between Python data objects (e.g. :class:`numpy.ndarray`,
 :class:`pandas.Series`, :class:`xarray.DataArray`, etc) and GMT happens through
-*virtual files*. These methods are context managers that automate the
-conversion of Python variables to GMT virtual files:
+*virtual files*. These methods are context managers that automate the conversion of
+Python variables to and from GMT virtual files:
```
weiji14, line 287:
```suggestion
Python objects to and from GMT virtual files:
```

weiji14 (approved):
Just one more small suggestion. Ok to leave out setting `output_type="pandas"` as the default setting (https://github.com/GenericMappingTools/pygmt/pull/3083/files#r1518966573) for now.

seisman:
Thanks for all the feedbacks. Now marking the PR as "final review call" and will merge it after 24 hours.

michaelgrund (approved):

PR #3097, opened 2024-03-11
pygmt.grd2xyz: Improve performance by storing output in virtual files

seisman:
Changes in this PR:

- Refactor the `grd2xyz` function using the `virtualfile_out`/`virtualfile_to_dataset` methods (address #1318 and #2730)
- Add type hints to the `output_type`/`outfile` parameters and return value
- Remove four tests that are not directly related to `grd2xyz` (see https://github.com/GenericMappingTools/pygmt/pull/3097#discussion_r1519773887).
- Combine the `test_grd2xyz` and `test_grd2xyz_format` tests into a single one

**Preview**: https://pygmt-dev--3097.org.readthedocs.build/en/3097/api/generated/pygmt.grd2xyz.html#pygmt.grd2xyz

PR #3085, opened 2024-03-04
pygmt.filter1d: Improve performance by storing output in virtual files

seisman:
**Preview**: https://pygmt-dev--3085.org.readthedocs.build/en/3085/api/generated/pygmt.filter1d.html#pygmt.filter1d

Similar to #3097.

PR #3098, opened 2024-03-11
Add doctests for the validate_output_table_type function

seisman:
Currently, we are testing the behavior of the `output_type`/`outfile` parameters in many wrappers' tests (e.g., the 3 `grd2xyz` tests removed in #3097, `filter1d` tests in #3085 and `grdvolume` tests). Since the `output_type`/`outfile` parameters are always checked by the `validate_output_table_type` function, we should test the `validate_output_table_type` function instead. 

The benefits are:

- Few tests to run and faster CI
- We can focus on the core functionality of wrappers rather than worry about the `output_type`/`outfile` parameters

### DIFF
--- a/pygmt/helpers/validators.py
+++ b/pygmt/helpers/validators.py
@@ -3,27 +3,51 @@ Functions to check if given arguments are valid.
 """
 
 import warnings
+from typing import Literal
 
 from pygmt.exceptions import GMTInvalidInput
 
 
-def validate_output_table_type(output_type, outfile=None):
+def validate_output_table_type(
+    output_type: Literal["pandas", "numpy", "file"], outfile: str | None = None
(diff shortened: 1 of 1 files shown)


pygmt/helpers/validators.py
```diff
@@ -3,27 +3,51 @@
...
     Parameters
     ----------
-    output_type : str
-        The type for a table output. Valid values are "file", "numpy", and
-        "pandas".
-    outfile : str
-        The file name for the output table file. Required if
-        ``output_type="file"``.
+    output_type
+        Desired output type of a tabular data. Valid values are ``"pandas"``,
```
michaelgrund, line 20:
```suggestion
        Desired output type of tabular data. Valid values are ``"pandas"``,
```

michaelgrund (approved):

PR #3108, opened 2024-03-13
pygmt.select: Improve performance by storing output in virtual files

seisman:
Similar to #3102 but for `select`.

**Preview**: https://pygmt-dev--3108.org.readthedocs.build/en/3108/api/generated/pygmt.select.html#pygmt.select

### DIFF
--- a/pygmt/src/select.py
+++ b/pygmt/src/select.py
@@ -2,14 +2,17 @@
 select - Select data table subsets based on multiple spatial criteria.
 """
 
+from typing import Literal
+
+import numpy as np
 import pandas as pd
 from pygmt.clib import Session
 from pygmt.helpers import (
-    GMTTempFile,
     build_arg_string,
     fmt_docstring,
     kwargs_to_strings,
     use_alias,
+    validate_output_table_type,
 )
 
 __doctest_skip__ = ["select"]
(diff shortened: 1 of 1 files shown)


michaelgrund (approved):

pygmt/src/select.py
```diff
@@ -180,12 +188,12 @@ def select(data=None, outfile=None, **kwargs):
...
-        Return type depends on whether the ``outfile`` parameter is set:
+    ret
+        Return type depends on ``outfile`` and ``output_type``:
 
-        - :class:`pandas.DataFrame` table if ``outfile`` is not set.
-        - None if ``outfile`` is set (filtered output will be stored in file
-          set by ``outfile``).
+        - None if ``outfile`` is set (output will be stored in file set by ``outfile``)
+        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is not set
+          (depends on ``output_type``)
```
yvonnefroehlich, line 196:
```suggestion
        - ``None`` if ``outfile`` is set (output will be stored in file set by ``outfile``)
        - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is not set
          (depends on ``output_type``)
```

PR #3113, opened 2024-03-16
Fix ghostscript typo in environment.yml

weiji14:

Fix `ghotscript` -> `ghostscript`.

Patches typo in #3112.

**Reminders**

**Slash Commands**

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
     - python=3.12
     # Required dependencies
     - gmt=6.5.0
-    - ghotscript=10.03.0
+    - ghostscript=10.03.0
     - numpy>=1.23
     - pandas>=1.5
     - xarray>=2022.03


environment.yml
```diff
@@ -6,7 +6,7 @@ dependencies:
     - python=3.12
     # Required dependencies
     - gmt=6.5.0
-    - ghotscript=10.03.0
+    - ghostscript=10.03.0
```
weiji14, line 9: Do we actually need to pin ghostscript in environment.yml? This line was added in #3112, but we didn't set any pin for ghostscript before.

seisman: Better to pin ghostscript here. Otherwise, new developers may install a different gs version and the tests may fail due to minor changes in gs.

seisman (approved):

PR #3115, opened 2024-03-17
**BREAKING** pygmt.grdcut: Refactor to store output in virtualfiles for grids

seisman:
This PR applies virtualfiles in `pygmt.grdcut`, so that this function doesn't have to write to temporary files (#2730).

`grdcut` is on of the complicated module that it works for grid, image and cube. We have to know the `kind` of the input before calling the `Session.virtualfile_out` method.

It turns out it's not trivial to determine the data `kind` of an input file. We have some options:

1. Based on file extensions (e.g., `.nc` is a grid, `.tif` is a image)
2. Based on magic bytes (xref: https://en.wikipedia.org/wiki/List_of_file_signatures)
3. Try reading as a grid/image/cube. If fail, then try the 2nd one. 
4. Add a new `kind` parameter (default to `"grid"`) and users should specify the `kind` parameter.

Option 1-2 may not always work; option 3 also dosn't always work (see https://github.com/GenericMappingTools/pygmt/pull/3115#discussion_r1780439384).

This PR currently adopts option 4. Need to note option 4 meaning **BREAKING**:

- previously, `grdcut("@earth_day_01d")` works; now must use ``grdcut("@earth_day_01d", kind="image")``
- previously, grdcut works 1/3/4-band images; now it only supports 3-band images. [We can support 1/4-band images later]


pygmt/src/grdcut.py
```diff
@@ -91,10 +98,21 @@ def grdcut(grid, outgrid: str | None = None, **kwargs):
     >>> # 12° E to 15° E and a latitude range of 21° N to 24° N
     >>> new_grid = pygmt.grdcut(grid=grid, region=[12, 15, 21, 24])
     """
+    inkind = data_kind(grid)
+    outkind = "grid"
+    if inkind == "image":
+        outkind = "image"
```
weiji14, line 104: Simplify to one-line
```suggestion
    outkind = "image" if inkind == "image" else "grid"
```

pygmt/clib/session.py
```diff
@@ -1905,6 +1915,10 @@ def virtualfile_to_grid(
         """
         if outgrid is not None:
             return None
+        if vfname.split("-")[3] == "I":
+            with GMTTempFile(suffix=".tif") as tmpfile:
+                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
+                return xr.load_dataarray(tmpfile.name)
```
weiji14, line 1921: A little unsure how this part works, where are you getting the "I" from?

pygmt/src/grdcut.py
```diff
@@ -91,10 +98,21 @@ def grdcut(grid, outgrid: str | None = None, **kwargs):
...
+            outkind = "image"
+
     with Session() as lib:
         with (
             lib.virtualfile_in(check_kind="raster", data=grid) as vingrd,
-            lib.virtualfile_out(kind="grid", fname=outgrid) as voutgrd,
+            lib.virtualfile_out(kind=outkind, fname=outgrid) as voutgrd,
         ):
             kwargs["G"] = voutgrd
             lib.call_module(
```
weiji14, line 118: Ok, so if we are able to have `outkind="image"` or `outkind="grid"`, should we rename the `virtualfile_to_grid` method to something like `virtualfile_to_grid_or_image`? Since that method should be able to handle both 1-band grids and 3-band images.

pygmt/clib/session.py
seisman: The virtual file name has the form of `@GMTAPI@-S-D-F-A-G-M-######`, in which `F` stands for the file family, and `I` means the virtual file is a `GMT_IMAGE`. This is just a quick hacky way to determine the virtual file family. Ideally we should wrap the `GMT_Inquire_VirtualFile` API function (http://docs.generic-mapping-tools.org/dev/devdocs/api.html#gmt-inquire-virtualfile).

See https://github.com/GenericMappingTools/gmt/blob/e5ef7016eaccce3fac58419f81249cc59d8735d4/src/gmt_api.c#L3364 for details.

pygmt/src/grdcut.py
seisman: You're talking about the `virtualfile_to_grid` method implemented in #2398, right? The method looks like this:
```python
def virtualfile_to_grid(self, vfname, outgrid):
    if outgrid is not None:
        return None
    return self.read_virtualfile(vfname, kind="grid").contents.to_dataarray()
```
Yes, the same code can be used for grid, image and even cubes. We just need to pass the correct `kind` parameter and make sure that `GMT_GRID`/`GMT_IMAGE`/`GMT_CUBE` all have the `.to_dataarray()` method.

So we can rewrite the method to:
```
def virtualfile_to_raster(self, vfname, outgrid, kind="grid"):
    if outgrid is not None:
        return None
    return self.read_virtualfile(vfname, kind=kind).contents.to_dataarray()
```
Actually, as mentioned in https://github.com/GenericMappingTools/pygmt/pull/3115/files#r1536468908, it's possible to determine the virtual file family by calling `GMT_Inquire_Family`.

So, what we should do is:

1. Wrap the `GMT_Inquire_Family`
2. Refactor the `read_virtualfile` method. Currently, `kind=None` means no casting into a GMT_* data structure. We can change the behavior to: if `kind=None`, then we call `GMT_Inquire_Family` to determine the virtual file family and then cast it to the corresponding GMT_* data structure.
3. Then the method can be simplified to:
```
def virtualfile_to_raster(self, vfname, outgrid):
    if outgrid is not None:
        return None
    return self.read_virtualfile(vfname).contents.to_dataarray()
```

pygmt/clib/session.py
```diff
@@ -2018,6 +2028,10 @@ def virtualfile_to_raster(
                 self["GMT_IS_IMAGE"]: "image",
                 self["GMT_IS_CUBE"]: "cube",
             }[family]
+        if kind == "image":
+            with GMTTempFile(suffix=".tif") as tmpfile:
+                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
+                return xr.load_dataarray(tmpfile.name)
```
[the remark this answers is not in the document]
weiji14: Ah I see, so this is just a placeholder for now. Yes, this PR will need to wait for #3128, though I'm thinking of cherry-picking some of the clib-related changes here in #3115 to #3128, or opening a separate PR for that.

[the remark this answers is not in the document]
seisman: The original idea of this PR is to make `grdcut` support both grids (via GMT_GRID) and images (via temporary files because GMT_IMAGE is not wrapped yet). I can continue this PR without waiting for #3128. Maybe it's easier.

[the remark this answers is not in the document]
weiji14: That could work too. Finish up this PR #3115 with the placeholder code, and then we'll refactor things in #3128 to remove the tempfile path. Either way, we should get both of these more-or-less ready at the same time, so that we have both of them for PyGMT v0.13.0.

pygmt/src/grdcut.py
```diff
@@ -99,13 +98,23 @@ def grdcut(grid, **kwargs):
...
-                )
+    inkind = data_kind(grid)
+    match inkind:
+        case "image" | "grid":
+            outkind = inkind
+        case "file":
+            realpath = which(grid, download="a")
+            if isinstance(realpath, list):
+                realpath = realpath[0]
+            outkind = "image" if realpath.endswith(".tif") else "grid"
```
seisman, line 109: When a file is given, there are two different ways to determine if the file is a grid or an image:

1. Based on file extensions
2. Read the file (header only if possible) via `xarray` or `rasterio` and check the data dims.

I guess option 2 is more robust and preferred.

pygmt/src/grdcut.py
```diff
@@ -99,13 +98,32 @@ def grdcut(grid, **kwargs):
...
+            realpath = str(grid)
+            if realpath.startswith("@"):  # Is a remote file without suffix
+                realpath = which(grid, download="a")
+                if isinstance(realpath, list):
+                    realpath = realpath[0]
 
-        return load_dataarray(outgrid) if outgrid == tmpfile.name else None
+            if realpath.endswith(
+                (".jpg", ".jpeg", ".png", ".tif", ".tiff", ".bmp", ".webp")
+            ):
```
seisman, line 115: The list of image suffixes comes from https://github.com/GenericMappingTools/GMT.jl/blob/78a6f2d2ff2c2a00dd0b6760555fd6e02cf4484f/src/gmtreadwrite.jl#L10.

`jp2` is ignored here, because a `jp2` file can be either a grid or an image, which is complicated to support.

pygmt/clib/session.py
```diff
@@ -2066,6 +2079,17 @@ def virtualfile_to_raster(
                 self["GMT_IS_IMAGE"]: "image",
                 self["GMT_IS_CUBE"]: "cube",
             }[family]
+
+        if kind == "image":  # Use temporary file for images
```
seisman, line 2083: This workaround can be removed after PR #3128, then rioxarray is no longer required for writing images.

pygmt/tests/test_grdcut.py
```diff
@@ -70,3 +112,23 @@ def test_grdcut_fails():
...
+    xr.testing.assert_allclose(a=result, b=expected_image)
+
+
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
+def test_grdcut_image_dataarray(region, expected_image):
+    """
+    Test grdcut on an input xarray.DataArray object.
+    """
+    path = which("@earth_day_01d", download="a")
+    raster = rioxarray.open_rasterio(path)
```
seisman, line 132: These lines can be removed after #2235.

pygmt/clib/session.py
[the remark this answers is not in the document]
seisman: I've decided to finish this PR without waiting for PR #3128. Two tests are also added which can help test changes in PR #3128, too.

pygmt/datatypes/image.py
```diff
@@ -0,0 +1,9 @@
+"""
+Wrapper for the GMT_GRID data type.
+"""
+
+import ctypes as ctp
+
+
+class _GMT_IMAGE(ctp.Structure):  # noqa: N801
+    pass
```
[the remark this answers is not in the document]
seisman: The placeholder for `_GMT_IMAGE` wrapper is removed in this PR, because it's not needed.

pygmt/src/grdcut.py
seisman: Currently use option 1 because it's easy for coding.

pygmt/src/grdcut.py
weiji14: If someone wants to write to a `.jp2` file, would it usually be a 1-band grid or a 3-band image? Can we not set `outkind=None` to let `inquire_virtualfile` determine this automatically based on the virtualfile?

pygmt/tests/test_grdcut.py
```diff
@@ -5,11 +5,18 @@
...
 from pygmt.helpers import GMTTempFile
 from pygmt.helpers.testing import load_static_earth_relief
 
+try:
+    import rioxarray
+
+    _HAS_RIOXARRAY = True
+except ImportError:
+    _HAS_RIOXARRAY = False
+
```
weiji14, line 19: Maybe best to move the image specific tests to a `test_grdcut_image.py` file (similar to what we have for `test_grdimage_image.py`). Then it's possible to use `rioxarray = pytest.importorskip("rioxarray")`.

pygmt/clib/session.py
```diff
@@ -2066,6 +2079,18 @@ def virtualfile_to_raster(
...
+            if not _HAS_RIOXARRAY:
+                raise ImportError(
+                    "Package `rioxarray` is required to be installed to load images. "
+                    "Please use `python -m pip install rioxarray` or "
+                    "`mamba install -c conda-forge rioxarray` to install the package."
+                )
+            with GMTTempFile(suffix=".tif") as tmpfile:
+                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
+                dataarray = rioxarray.open_rasterio(tmpfile.name)  # type: ignore[return-value]
+                return dataarray.load()
```
weiji14, line 2093: Continuing from https://github.com/GenericMappingTools/pygmt/issues/3315#issuecomment-2212420503. This part is causing an error on Windows:

```python-traceback
        path = which("@earth_day_01d", download="a")
        raster = rioxarray.open_rasterio(path)
>       result = grdcut(raster, region=region)

..\pygmt\tests\test_grdcut.py:133: 
..\pygmt\helpers\decorators.py:609: in new_module
..\pygmt\helpers\decorators.py:773: in new_module
..\pygmt\src\grdcut.py:127: in grdcut
    return lib.virtualfile_to_raster(
..\pygmt\clib\session.py:2090: in virtualfile_to_raster
    with GMTTempFile(suffix=".tif") as tmpfile:
..\pygmt\helpers\tempfile.py:77: in __exit__
    Path(self.name).unlink(missing_ok=True)

self = WindowsPath('C:/Users/RUNNER~1/AppData/Local/Temp/pygmt-nwxyygd3.tif')
missing_ok = True

    def unlink(self, missing_ok=False):
        """
        Remove this file or link.
        If the path is a directory, use rmdir() instead.
        """
        try:
>           os.unlink(self)
E           PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\pygmt-nwxyygd3.tif'
```

Can you try returning the dataarray fully loaded into memory like so:

```suggestion
            with GMTTempFile(suffix=".tif") as tmpfile:
                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
                dataarray = rioxarray.open_rasterio(tmpfile.name).load()
                return dataarray
```

pygmt/clib/session.py
```diff
@@ -2066,6 +2079,18 @@ def virtualfile_to_raster(
...
+            if not _HAS_RIOXARRAY:
+                raise ImportError(
+                    "Package `rioxarray` is required to be installed to load images. "
+                    "Please use `python -m pip install rioxarray` or "
+                    "`mamba install -c conda-forge rioxarray` to install the package."
+                )
+            with GMTTempFile(suffix=".tif") as tmpfile:
+                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
+                dataarray = rioxarray.open_rasterio(tmpfile.name).load()
+                return dataarray
```
seisman, line 2093:
```suggestion
            with GMTTempFile(suffix=".tif") as tmpfile:
                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
                with rioxarray.open_rasterio(tmpfile.name) as da:
                    dataarray = da.load()
                return dataarray
```

pygmt/clib/session.py
seisman:
```
            with GMTTempFile(suffix=".tif") as tmpfile:
                self.call_module("write", f"{vfname} {tmpfile.name} -Ti")
                with rioxarray.open_rasterio(tmpfile.name) as da:
                    dataarray = da.load()
                return dataarray
```
This one works. I guess we just need to ignore the mypy errors.

pygmt/tests/test_grdcut.py
seisman: > Maybe best to move the image specific tests to a `test_grdcut_image.py` file (similar to what we have for `test_grdimage_image.py`). 

I'm OK with moving the tests to a separate file.

> Then it's possible to use `rioxarray = pytest.importorskip("rioxarray")`.

Currently yes, but after #3128 is done, rioxarray is no longer needed for the `test_grdcut_image_file` test, then we can't use `pytest.importorskip("rioxarray")` to skip all the tests.

pygmt/src/grdcut.py
seisman: > If someone wants to write to a `.jp2` file, would it usually be a 1-band grid or a 3-band image?

Are you talking about **reading** a `.jp2` file? As I understand it, the [JPEG2000](https://en.wikipedia.org/wiki/JPEG_2000) format is designed for storing images, so it's usually 3-band, but in GMT, the tiled remote grids are stored in JPEG2000 format for its smaller size. These tiled grids are loaded like below. Is it a 1-band grid or a 1-band image?
```
>>> import rioxarray
>>> rioxarray.open_rasterio("S90W180.earth_age_05m_p.jp2")
<xarray.DataArray (band: 1, y: 2160, x: 2160)> Size: 9MB
[4665600 values with dtype=int16]
Coordinates:
  * band         (band) int64 8B 1
  * x            (x) float64 17kB -180.0 -179.9 -179.8 ... -0.125 -0.04167
  * y            (y) float64 17kB 89.96 89.88 89.79 ... -89.79 -89.87 -89.96
    spatial_ref  int64 8B 0
Attributes:
    scale_factor:  1.0
    add_offset:    0.0
```

> Can we not set `outkind=None` to let `inquire_virtualfile` determine this automatically based on the virtualfile?

`outkind` can be `None` in `virtualfile_to_raster`, but can't be `None` in `virtualfile_out`.

pygmt/tests/test_grdcut_image.py
```diff
@@ -0,0 +1,78 @@
...
+        data=np.array(
+            [
+                [[90, 93, 95, 90], [91, 90, 91, 91], [91, 90, 89, 90]],
+                [[87, 88, 88, 89], [88, 87, 86, 85], [90, 90, 89, 88]],
+                [[48, 49, 49, 45], [49, 48, 47, 45], [48, 47, 48, 46]],
+            ],
+            dtype=np.uint8,
+        ),
+        coords={
+            "band": [1, 2, 3],
```
seisman, line 41: An off-topic question: 
```
In [3]: import rioxarray

In [4]: rioxarray.open_rasterio("earth_day_01d_p.tif")
Out[4]:
<xarray.DataArray (band: 3, y: 180, x: 360)> Size: 194kB
[194400 values with dtype=uint8]
Coordinates:
  * band         (band) int64 24B 1 2 3
  * x            (x) float64 3kB -179.5 -178.5 -177.5 ... 177.5 178.5 179.5
  * y            (y) float64 1kB 89.5 88.5 87.5 86.5 ... -86.5 -87.5 -88.5 -89.5
    spatial_ref  int64 8B 0
Attributes:
    AREA_OR_POINT:             Area
    STATISTICS_MAXIMUM:        255
    STATISTICS_MEAN:           49.507654320988
    STATISTICS_MINIMUM:        10
    STATISTICS_STDDEV:         65.299010755209
    STATISTICS_VALID_PERCENT:  100
    _FillValue:                0
    scale_factor:              1.0
    add_offset:                0.0
```
The `band` here is `[1, 2, 3]`, but `tile_map`, `band` is set to `[0, 1, 2]`:
 https://github.com/GenericMappingTools/pygmt/blob/41a0fe0a6a79849565348203737360855d40add8/pygmt/datasets/tile_map.py#L160

Are there any specific reasons for the differences or any conventions about `band`?

pygmt/tests/test_grdcut_image.py
```diff
@@ -0,0 +1,78 @@
...
+            "STATISTICS_MEAN": 90.91666666666667,
+            "STATISTICS_MINIMUM": 89,
+            "STATISTICS_STDDEV": 1.5523280008498,
+            "STATISTICS_VALID_PERCENT": 100,
+            "AREA_OR_POINT": "Area",
+        },
+    )
+
+
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
```
seisman, line 61: The `pytest.mark.skipif` marker can be removed after #3128 is implemented.

pygmt/tests/test_grdcut_image.py
```diff
@@ -0,0 +1,78 @@
...
+    xr.testing.assert_allclose(a=result, b=expected_image)
+
+
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
+def test_grdcut_image_dataarray(region, expected_image):
+    """
+    Test grdcut on an input xarray.DataArray object.
+    """
+    path = which("@earth_day_01d", download="a")
+    raster = rioxarray.open_rasterio(path)
```
seisman, line 76: These lines can be replaced by `load_blue_marble` after #2235.

pygmt/clib/session.py
weiji14: That `Item "list[Dataset]" of "Dataset | DataArray | list[Dataset]" has no attribute "__enter__"  [union-attr]` mypy error seems to be an upstream bug. Probably should report it, but I'm unsure whether to post it to rioxarray or xarray...

pygmt/tests/test_grdcut_image.py
weiji14: Oh right, the 123 numbering comes from GDAL/rasterio. E.g. if you do `gdalinfo ~/.gmt/server/earth/earth_day/earth_day_01d_p.tif`, you'll get:

```
Driver: GTiff/GeoTIFF
Files: /home/user/.gmt/server/earth/earth_day/earth_day_01d_p.tif
Size is 360, 180
...
Band 1 Block=360x7 Type=Byte, ColorInterp=Undefined
  Min=10.000 Max=255.000 
  Minimum=10.000, Maximum=255.000, Mean=49.508, StdDev=65.299
  NoData Value=0
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=49.507654320988
    STATISTICS_MINIMUM=10
    STATISTICS_STDDEV=65.299010755209
    STATISTICS_VALID_PERCENT=100
Band 2 Block=360x7 Type=Byte, ColorInterp=Undefined
  Min=10.000 Max=255.000 
  Minimum=10.000, Maximum=255.000, Mean=49.911, StdDev=62.129
  NoData Value=0
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=49.910941358025
    STATISTICS_MINIMUM=10
    STATISTICS_STDDEV=62.129078727248
    STATISTICS_VALID_PERCENT=100
Band 3 Block=360x7 Type=Byte, ColorInterp=Undefined
  Min=10.000 Max=255.000 
  Minimum=10.000, Maximum=255.000, Mean=65.605, StdDev=44.820
  NoData Value=0
  Metadata:
    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=65.605154320988
    STATISTICS_MINIMUM=10
    STATISTICS_STDDEV=44.819618148349
    STATISTICS_VALID_PERCENT=100
```

> The `band` here is `[1, 2, 3]`, but `tile_map`, `band` is set to `[0, 1, 2]`:

Yeah, I simply forgot about GDAL's start from 1 convention when implementing https://github.com/GenericMappingTools/pygmt/pull/2125 :sweat_smile:. I'm guessing there's a historical reason for this, and I'm not sure if this is the exact reason why GDAL counts from 1, but optical satellite bands (e.g. from Landsat) are usually numbered Band 1, Band 2, etc, so that could be why? Unsure if we should re-number things, or just leave it.

seisman: I think we should follow the GDAL convention and let band indexing start from 1.

pygmt/clib/session.py
seisman: https://corteva.github.io/rioxarray/html/_modules/rioxarray/_io.html#open_rasterio

I guess it's because `open_rasterio` is defined as a normal function rather than a context manager. It's unclear why `with rioxarray.open_rasterio(tmpfile.name) as da:` works, but as you mentioned above, the rioxarray IO is really complicated.

pygmt/src/grdcut.py
weiji14: No, I'm talking about writing a `.jp2` file. I'm assuming GMT can correctly read a `.jp2` file and infer whether it is a grid or image already.

> These tiled grids are loaded like below. Is it a 1-band grid or a 1-band image?

That `S90W180.earth_age_05m_p.jp2` file is supposed to be a 3-band image, but `rioxarray.open_rasterio` is incorrectly reading it as 1-band because rioxarray hasn't properly implemented `colormap`/`colorinterp` yet (tried at https://github.com/corteva/rioxarray/pull/414, but got a bit stuck).

> > Can we not set `outkind=None` to let `inquire_virtualfile` determine this automatically based on the virtualfile?
> 
> `outkind` can be `None` in `virtualfile_to_raster`, but can't be `None` in `virtualfile_out`.

Ah ok, so we do need to pick a kind here. From https://gdal.org/drivers/raster/jp2openjpeg.html, it says that jp2 supports arbitrary band counts, so either 3-band or 1-band is possible. I've seen both cases before in my past work, 3-band RGB images stored as JP2, and also 1-band images from Sentinel-2 (e.g. https://gis.stackexchange.com/questions/285887/converting-sentinel-2-jp2-bands-to-tiff-with-gdal), so a bit hesitant to pick one or the other :confused:

pygmt/tests/test_grdcut.py
weiji14: Ok, sounds good.

pygmt/tests/test_grdcut_image.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
+def test_grdcut_image_file(region, expected_image):
+    """
+    Test grdcut on an input image file.
+    """
+    result = grdcut("@earth_day_01d", region=region)
+    xr.testing.assert_allclose(a=result, b=expected_image)
+
+
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
```
weiji14, line 70: Run codspeed benchmark for `test_grdcut_image_dataarray`

```suggestion
@pytest.mark.benchmark
@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
```

pygmt/src/grdcut.py
seisman: Is it OK to say that a grid is equivalent to a 1-band image? Based on my local experiments, they seems to be the same. A one-band image (e.g., the `earth_day_01d_p.tif` file) can be read into either a `_GMT_GRID` container or a `_GMT_IMAGE` container.

seisman: Will do some experiments in #3318. Will let you know if it works as expected.

weiji14: > Is it OK to say that a grid is equivalent to a 1-band image?

Yes, you can say that.

> Based on my local experiments, they seem to be the same. A one-band image (e.g., the `earth_day_01d_p.tif` file) can be read into either a `_GMT_GRID` container or a `_GMT_IMAGE` container.

Just to be clear, that `earth_day_01d_p.tif` file is not a 1-band image, it should be a 3-band image... Better check the pixel values if you're able to read it into a `_GMT_GRID`.

pygmt/clib/session.py
```diff
@@ -1942,6 +1942,7 @@ def virtualfile_out(
                 "grid": ("GMT_IS_GRID", "GMT_IS_SURFACE"),
                 "image": ("GMT_IS_IMAGE", "GMT_IS_SURFACE"),
             }[kind]
+            # For unkown reasons, 'GMT_OUT' crashes for 'image' kind.
```
weiji14, line 1945:
```suggestion
            # For unknown reasons, 'GMT_OUT' crashes for 'image' kind.
```

pygmt/clib/session.py
```diff
@@ -2311,3 +2312,29 @@ def extract_region(self) -> np.ndarray:
         if status != 0:
             raise GMTCLibError("Failed to extract region from current figure.")
         return region
+
+
+def raster_kind(raster: str):
```
weiji14, line 2317: Add type hint, and make this into a private function.

```suggestion
def _raster_kind(raster: str) -> Literal["grid", "image"] | None:
```

pygmt/src/grdcut.py
```diff
@@ -4,21 +4,20 @@
 
 import xarray as xr
 from pygmt.clib import Session
+from pygmt.clib.session import raster_kind
```
weiji14, line 7:
```suggestion
from pygmt.clib.session import _raster_kind
```

pygmt/src/grdcut.py
```diff
@@ -100,13 +99,21 @@ def grdcut(grid, **kwargs) -> xr.DataArray | None:
...
-                lib.call_module(
-                    module="grdcut", args=build_arg_list(kwargs, infile=vingrd)
-                )
+    # Determine the output data kind based on the input data kind.
+    inkind = data_kind(grid)
+    match inkind:
+        case "image" | "grid":
+            outkind = inkind
+        case "file":
+            outkind = raster_kind(grid)
```
weiji14, line 108:
```suggestion
            outkind = _raster_kind(grid)
```

pygmt/clib/session.py
```diff
@@ -2311,3 +2312,29 @@ def extract_region(self) -> np.ndarray:
...
+    """
+    Determine the raster kind.
+
+    >>> raster_kind("@earth_relief_01d")
+    'grid'
+    >>> raster_kind("@static_earth_relief.nc")
+    'grid'
+    >>> raster_kind("@earth_day_01d")
+    'image'
+    >>> raster_kind("@hotspots.txt")
```
weiji14, line 2327:
```suggestion
    >>> _raster_kind("@earth_relief_01d")
    'grid'
    >>> _raster_kind("@static_earth_relief.nc")
    'grid'
    >>> _raster_kind("@earth_day_01d")
    'image'
    >>> _raster_kind("@hotspots.txt")
```

pygmt/src/grdcut.py
```diff
@@ -100,13 +99,21 @@ def grdcut(grid, **kwargs) -> xr.DataArray | None:
...
+    with Session() as lib:
+        with (
+            lib.virtualfile_in(check_kind="raster", data=grid) as vingrd,
+            lib.virtualfile_out(kind=outkind, fname=outgrid) as voutgrd,
+        ):
+            kwargs["G"] = voutgrd
+            lib.call_module(module="grdcut", args=build_arg_list(kwargs, infile=vingrd))
+            return lib.virtualfile_to_raster(
+                outgrid=outgrid, kind=outkind, vfname=voutgrd
+            )
```
weiji14, line 119: Not sure why params are reversed here.
```suggestion
            return lib.virtualfile_to_raster(
                vfname=voutgrd, kind=outkind, outgrid=outgrid
            )
```

pygmt/src/grdcut.py
weiji14, line 108: Need another unit test to test when input `grid` and output `outgrid` are both files.

pygmt/clib/session.py
```diff
@@ -2311,3 +2312,29 @@ def extract_region(self) -> np.ndarray:
...
+        try:
+            img = lib.read_data(infile=raster, kind="image", mode="GMT_CONTAINER_ONLY")
+            return "image" if img.contents.header.contents.n_bands == 3 else "grid"
+        except GMTCLibError:
+            pass
+        try:
+            _ = lib.read_data(infile=raster, kind="grid", mode="GMT_CONTAINER_ONLY")
+            return "grid"
+        except GMTCLibError:
+            pass
```
weiji14, line 2339: Just double-checking that there's not a function in GMT C already to determine grid/image type? I know there's [`GMT_Inquire_VirtualFile`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#inquire-a-virtual-file-for-family) for virtualfiles, but is there not one for just files? I saw this [`GMT_Get_Info`](https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#gmt-get-info), but unsure if it'll be usable here.

seisman: https://github.com/GenericMappingTools/gmt/blob/7809736ba32d87a4a96b15444419eb176c6a35f3/src/gmt_grdio.c#L3377 

`gmt_raster_type` is the function that GMT uses to determine the raster type, but it's not a public API function.

weiji14: Oo, magic byte parsing :laughing: If that function becomes public someday, maybe we'll use it.

pygmt/clib/session.py
```diff
@@ -2311,3 +2312,29 @@ def extract_region(self) -> np.ndarray:
         if status != 0:
             raise GMTCLibError("Failed to extract region from current figure.")
         return region
+
+
+def _raster_kind(raster: str) -> Literal["grid", "image"] | None:
```
weiji14, line 2317: Mypy is complaining about the output type hint here:

```
pygmt/src/grdcut.py:108: error: Incompatible types in assignment (expression has type "Literal['grid', 'image'] | None", variable has type "Literal['grid', 'image']")  [assignment]
```

Maybe we should just use `Literal['grid', 'image']` and raise `GMTInvalidInput` if it is not a grid/image?

seisman: Maybe falling back to `"grid"` and then let GMT itself decide if it can process it or not.

weiji14: In that case, with a binary image/grid output, maybe just make a function like `def _is_image() -> bool`?

seisman: grdcut can also cut a cube, so the `_raster_kind` function should return `grid`/`image`/`cube` if possible. Currently, the' cube' support is not yet finished (PR #3150). So, the name `_raster_kind` makes more sense.

weiji14: Ah yes, and then there would be 3 try-except statements. Now wishing we could re-use the function from GMT C...

pygmt/src/grdcut.py
seisman: Done in 3c12e2b78.

pygmt/clib/session.py
```diff
@@ -2311,3 +2312,35 @@ def extract_region(self) -> np.ndarray:
...
+    >>> _raster_kind("@static_earth_relief.nc")
+    'grid'
+    >>> _raster_kind("@earth_day_01d")
+    'image'
+    >>> _raster_kind("@hotspots.txt")
+    'grid'
+    """
+    # The logic here is because: an image can be read into a grid container, but a grid
+    # can't be read into an image container. So, try to read the file as an image first.
+    # If fails, try to read it as a grid.
```
seisman, line 2334: It turns out the logic here is not 100% correct.

- An image can be read into a grid container, but only the first band is read.
- A grid can be read into an image container, as long as `libgdal-netcdf` or `libgdal-hdf5` is installed

weiji14: > * A grid can be read into an image container, as long as another libgdal package (maybe `libgdal-netcdf` or `libgdal-hdf5`) is installed. xref:

The first try-except check is to try reading into a `GMT_IMAGE` struct, and return "image" if there are 3 bands, and "grid" if there is only 1 band. Even if grids can be read into `GMT_IMAGE`, we will still call it a grid if it has 1-band right?

pygmt/tests/test_grdcut_image.py
```diff
@@ -0,0 +1,85 @@
...
+    result = grdcut("@earth_day_01d_p", region=region)
+    xr.testing.assert_allclose(a=result, b=expected_image)
+
+
+@pytest.mark.skipif(not _HAS_RIOXARRAY, reason="rioxarray is not installed")
+def test_grdcut_image_dataarray(region, expected_image):
+    """
+    Test grdcut on an input xarray.DataArray object.
+    """
+    raster = rioxarray.open_rasterio(which("@earth_day_01d", download="a")).load()
```
weiji14, line 70: Use `load_blue_marble` or `load_black_marble` instead here?

pygmt/clib/session.py
seisman: Trying to refresh my memory about what's happening:

- Reading a grid into `GMT_GRID`: Works!
- Reading an image into `GMT_GRID`: Only the first band is read.
- Reading a grid into `GMT_IMAGE`: It depends on `libgdal-hdf5` being installed. Read as an single-band image if installed, otherwise raise errors.
- Reading an image into `GMT_IMAGE`: works.


```
import pygmt
pygmt.grdcut("@earth_relief_01d_p", region=[0, 10, 0, 10])
```
If `libgdal-hdf5` is not installed, `_raster_kind` first read it as an image, which fails and reports following errors:
```
ERROR 4: `/home/runner/.gmt/server/earth/earth_relief/earth_relief_01d_p.grd' not recognized as being in a supported file format. It could have been recognized by driver HDF5, but plugin gdal_HDF5.so is not available in your installation. You may install it with 'conda install -c conda-forge libgdal-hdf5'
Error: ession [ERROR]: Unable to open earth_relief_01d_p.grd.
Error: ession [ERROR]: ERROR reading image with gdalread.
pygmt-session (gmtapi_import_image): Could not read from file [earth_relief_01d_p.grd]
[Session pygmt-session (43)]: Error returned from GMT API: GMT_IMAGE_READ_ERROR (22)
[Session pygmt-session (43)]: Error returned from GMT API: GMT_IMAGE_READ_ERROR (22)
```
then it tries to read it as a grid, which succeeds. But we still see the error messages above. I believe we need to find a way to suppress these errors or find an alternative way to determine the data kind.

pygmt/tests/test_grdcut_image.py
seisman: Done in bb1a0b0032770c44091105d461cee2febe6cc07c.

pygmt/src/grdcut.py
```diff
@@ -100,13 +107,23 @@ def grdcut(grid, **kwargs) -> xr.DataArray | None:
...
-                )
+    # Determine the output data kind based on the input data kind.
+    match inkind := data_kind(grid):
+        case "image" | "grid":
+            outkind = inkind
+        case "file":
+            outkind = kind
+        case "_":
+            msg = f"Unsupported data type {type(grid)}."
+            raise GMTInvalidInput(msg)
```
seisman, line 118: Lines 117-118 should be caught by the `test_grdcut_fails` test.

seisman:
I think it's better to not introduce last-minute breaking changes, so bump this PR to v0.15.0.

pygmt/src/grdcut.py
```diff
@@ -100,13 +108,27 @@ def grdcut(grid, **kwargs) -> xr.DataArray | None:
...
+        msg = f"Invalid raster kind: '{kind}'. Valid values are 'grid' and 'image'."
+        raise GMTInvalidInput(msg)
+
+    # Determine the output data kind based on the input data kind.
+    match inkind := data_kind(grid):
+        case "grid" | "image":
+            outkind = inkind
+        case "file":
+            outkind = kind
+        case "_":
```
weiji14, line 121:
```suggestion
        case _:
```

weiji14 (approved):
Still think using the private `gmt_raster_type` func as mentioned at https://github.com/GenericMappingTools/pygmt/pull/3115#discussion_r1779438773 would be the best solution to auto-detect the raster kind, but failing that, the logic here to explicitly have the user pass in the type works too.

Just one typo on the match-case statement, and should be ok to merge once all tests pass without coverage misses.

PR #2706, opened 2023-09-30
Figure.contour: Adjust processing of arguments passed to the "annotation" and "levels" parameters

yvonnefroehlich:
**Description of proposed changes**

This PR aims to adjust the arguments passed to the `annotation` and `levels` parameters of `Figure.contour`:

- [x] input options
       - a fixed interval als float `100` -> `100`
       - one level as list: `[100]` -> `100,`  (mention the trailing comma)
       - multiple levels as list: `[100, 200]` -> `100,200`
       - no as string: `"n"` -> `n`  (syntax of GMT 6.5)
- [x] Update docstrings
- [x] Add tests for interval, one level, and multiple levels

Related:
* GMT docs
  * `annotation`: https://docs.generic-mapping-tools.org/dev/contour.html#a
  * `interval`: https://docs.generic-mapping-tools.org/dev/contour.html#c
* PR for `Figure.grdcontour`: #3116 

**Preview**: https://pygmt-dev--3116.org.readthedocs.build/en/2706/api/generated/pygmt.Figure.contour.html

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3116, opened 2024-03-17
Figure.grdcontour: Adjust processing of arguments passed to the "annotation" and "interval" parameters, deprecate "sequence_plus"

yvonnefroehlich:

This PR aims to adjust the arguments passed to the `annotation` and `interval` parameters of `Figure.grdcontour`:

- [x] input options
       - a fixed interval als float `100` -> `100`
       - one level as list: `[100]` -> `100,`  (mention the trailing comma)
       - multiple levels as list: `[100, 200]` -> `100,200`
       - no as string: `"n"` -> `n`  (syntax of GMT 6.5)
- [x] Deprecated `sequence_plus` for `annotation` instead use one string, e.g., `[100, "f10p", "gred"]`, now `"100+f10p+gred"`
- [x] Update docstrings
- [x] Add tests for one level and multiple levels

Related:
* GMT docs
  * `annotation`: https://docs.generic-mapping-tools.org/dev/grdcontour.html#a
  * `interval`: https://docs.generic-mapping-tools.org/dev/grdcontour.html#c
* PR for `Figure.contour`: #2706

**Preview**: https://pygmt-dev--3116.org.readthedocs.build/en/3116/api/generated/pygmt.Figure.grdcontour.html

**Reminders**

**Slash Commands**


pygmt/src/grdcontour.py
```diff
@@ -58,11 +58,11 @@ def grdcontour(self, grid, **kwargs):
...
 
         - Specify a fixed annotation interval *annot_int* or a
-          single annotation level +\ *annot_int*.
+          single annotation level **+**\ *annot_int*.
         - Disable all annotation with  **-**.
         - Optional label modifiers can be specified as a single string
-          ``"[annot_int]+e"``  or with a list of arguments
-          ``([annot_int], "e", "f10p", "gred")``.
+          ``"annot_int+e+f10p+gred"`` or with a list of arguments
+          ``[annot_int, "e", "f10p", "gred"]``.
```
yvonnefroehlich, line 65: I am wondering whether here the squared brackets are used to indicate that `annot_int` should be replaced with the desired value by the user. I find this a bit misleading.

seisman: I don't think the description is correct. In the PyGMT world, a list of arguments means the option is repeatable. For example `frame=["WSen", "xaf", "yaf"]` is translated to `-BWSen -Bxaf -Byaf`. For this parameter, it must be `annot_int+e+f10+gred`, not a list.

yvonnefroehlich: Hm. I tried the following code to understand the input expected by `annotation`. It works for passing a list and I get the same output as for passing a string.

``` python
# orientated on https://www.pygmt.org/dev/api/generated/pygmt.Figure.grdcontour.html
# last access: 2024/03/18

import pygmt

region = [-92.5, -82.5, -3, 7]

grid = pygmt.datasets.load_earth_relief(resolution="15m", region=region)

fig = pygmt.Figure()
fig.grdcontour(
    projection="M10c",
    region=region,
    grid=grid,
    # annotation="500+f10p+gred",  # string -> WORKS
    # annotation=([500], "f10p", "gred"),  # -> FAILS
    # annotation=(500, "f10p", "gred"),  # -> WORKS
    annotation=[500, "f10p", "gred"],  # list -> WORKS
    frame=True,
)
fig.show()

# fig.savefig(fname="grdcontour_A_string.png")
# fig.savefig(fname="grdcontour_A_list.png")
```

_Output figure_
grdcontour_A_list

seisman: https://github.com/GenericMappingTools/pygmt/blob/6fd4a00c29a6fe9de9b541823190697a6ec751f7/pygmt/src/grdcontour.py#L30-L32

I see. It's because `A` is defined like above, so the list of arguments is joined by `+`. 

Actually, this is the only parameter that sets to `sequence_plus`.

seisman: I think we should disallow the syntax like `annotation=[500, "f10p", "gred"]`, because:

1. From a user's perspective, it's no different from passing a string like `500+f10p+gred`
2. The syntax is not well documented and is only used for this single parameter
3. It's not Pythonic at all, and it's very likely that we won't support such syntax in the new alias system (#2949 but still WIP)

So, I think we should remove `A="sequence_plus"` and update the documentation.

yvonnefroehlich: I was actually also a bit wondering if we need this possibility, as such a list does not improve the readability significantly compared to the corresponding string. So, I think I am fine with removing `A"sequence_plus"`.

pygmt/src/grdcontour.py
```diff
@@ -53,16 +51,17 @@ def grdcontour(self, grid, **kwargs):
...
           +\ *cont_int*.
-    annotation : str, int, or list
+    annotation : int or str
         Specify or disable annotated contour levels, modifies annotated
         contours specified in ``interval``.
 
         - Specify a fixed annotation interval *annot_int* or a
-          single annotation level +\ *annot_int*.
+          single annotation level **+**\ *annot_int*.
         - Disable all annotation with  **-**.
```
seisman, line 60: Some descriptions (e.g., disable annotations with `-`) are outdated. Need to update the docstrings following the upstream documentation https://docs.generic-mapping-tools.org/6.5/grdcontour.html#a.

seisman:
> contours is annotation interval in data units; it is ignored if contour levels are given in a file via [-C](https://docs.generic-mapping-tools.org/6.5/grdcontour.html#c). [Default is no annotations]. Prepend n to disable all annotations implied by [-C](https://docs.generic-mapping-tools.org/6.5/grdcontour.html#c). To just select a few specific contours give them as a comma-separated string; if only a single contour please add a trailing comma so it is seen as a list and not a contour interval.

So, valid arguments are for the `-C` option are:

- `100` means every 100 contours
- `100,` means a single specific contour with value=100
- `100,200,300` means multiple specific contours with value=100, 200 and 300
- `n` means disable all contours

The corresponding Pythonic arguments are:

- a float value: `annotation=100`
- a list with a single value: `annotation=[100]`
- a list with multiple values: `annotation=[100, 200, 300]`
- a string: `annotation="n"`

I think we should change `A="sequence_plus"` to `A="sequence"` instead, to accept arguments like `annotation=[100, 200, 300]`.

yvonnefroehlich:
Actually there is already PR https://github.com/GenericMappingTools/pygmt/pull/2706 which is about updating / improving `annotation` and the `levels` or `interval` parameters of `Figure.contour` and `Figure.grdcontour`. Probably it makes sense to bundle this change in one of these PRs and close the other one.

seisman:
> Actually there is already PR #2706 which is about updating / improving `annotation` and the `levels` or `interval` parameters of `Figure.contour` and `Figure.grdcontour`. Probably it makes sense to bundle this change in one of these PRs and close the other one.

It seems that `grdcontour` and `contour` have identical `-A` and `-C` options. I think we can focus on improving `grdcontour` in this PR and apply the same changes to `contour` in PR #2706.

pygmt/src/grdcontour.py
```diff
@@ -27,9 +27,7 @@
     p="perspective",
     t="transparency",
 )
-@kwargs_to_strings(
-    R="sequence", L="sequence", A="sequence_plus", c="sequence_comma", p="sequence"
-)
+@kwargs_to_strings(R="sequence", L="sequence", c="sequence_comma", p="sequence")
```
seisman, line 30:
```suggestion
@kwargs_to_strings(R="sequence", L="sequence", A="squence", c="sequence_comma", p="sequence")
```

pygmt/src/grdcontour.py
```diff
@@ -53,16 +51,17 @@ def grdcontour(self, grid, **kwargs):
           angle (col 3).
         - A fixed contour interval *cont_int* or a single contour with
           +\ *cont_int*.
-    annotation : str, int, or list
+    annotation : float or str
```
seisman, line 54:
```suggestion
    annotation : float or str or list
```

yvonnefroehlich:
\format

pygmt/src/grdcontour.py
```diff
@@ -28,7 +28,7 @@
     t="transparency",
 )
 @kwargs_to_strings(
-    R="sequence", L="sequence", A="sequence_plus", c="sequence_comma", p="sequence"
+    R="sequence", L="sequence", A="sequence", c="sequence_comma", p="sequence"
```
seisman, line 31: We want to convert `annotation=[100, 200, 300]` to `A="100,200,300"`, so we should use `A="sequence_comma"` instead, but it won't work for the case of a single contour (e.g., `annotation=[100]`). It will convert `annotation=[100]` to `A="100"`, but we expect it to be `A="100,"` (not the trailing comma). I think we need to remove `A="sequence"` and add codes like:
```
if is_nonstr_iter(kwargs.get("A")):
    if len(kwargs["A"]) == 1:
        kwargs["A"] = str(kwargs["A"]) + "," 
    else:
        kwargs["A"] = ",".join(f"{item}" for item in kwargs["A"])
```

pygmt/src/grdcontour.py
```diff
@@ -28,7 +34,7 @@
...
```
seisman, line 37: Should be remove `A="sequence"` because we need to process this argument manually.
```suggestion
    R="sequence", L="sequence", c="sequence_comma", p="sequence"
```

pygmt/src/grdcontour.py
```diff
@@ -96,32 +102,45 @@ def grdcontour(self, grid, **kwargs):
...
+    ...     # Set the projection to Mercator for the 10 cm figure
     ...     projection="M10c",
     ... )
-    >>> # show the plot
+    >>> # Show the plot
     >>> fig.show()
     """
     kwargs = self._preprocess(**kwargs)
+
+    if is_nonstr_iter(kwargs.get("A")):
```
seisman, line 132: Need to add grdcontour tests for these new codes.

seisman: Need help with writing the tests?

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,45 @@ def grdcontour(self, grid, **kwargs):
...
+        if len(kwargs["A"]) == 1:
+            kwargs["A"] = str(kwargs["A"]) + ","
+        else:
+            kwargs["A"] = ",".join(f"{item}" for item in kwargs["A"])
```
seisman, line 134:
```suggestion
        if len(kwargs["A"]) == 1:
            kwargs["A"] = str(kwargs["A"][0]) + ","
        else:
            kwargs["A"] = ",".join(f"{item}" for item in kwargs["A"])
```

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,45 @@ def grdcontour(self, grid, **kwargs):
...
+
+    if is_nonstr_iter(kwargs.get("C")):
+        if len(kwargs["C"]) == 1:
+            kwargs["C"] = str(kwargs["C"]) + ","
```
seisman, line 138:
```suggestion
            kwargs["C"] = str(kwargs["C"][0]) + ","
```

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,45 @@ def grdcontour(self, grid, **kwargs):
...
+        if len(kwargs["A"]) == 1:
+            kwargs["A"] = str(kwargs["A"][0]) + ","
+        else:
+            kwargs["A"] = ",".join(f"{item}" for item in kwargs["A"])
+
+    if is_nonstr_iter(kwargs.get("C")):
+        if len(kwargs["C"]) == 1:
+            kwargs["C"] = str(kwargs["C"][0]) + ","
+        else:
+            kwargs["C"] = ",".join(f"{item}" for item in kwargs["C"])
```
seisman, line 140: These two code blocks have exactly the same code, except one for `A` and another for `C`. Please rewrite it using a for-loop instead, e.g., 
```
for arg in ["A", "C"]:
    if ...

```

yvonnefroehlich: Totally right, done in the commits https://github.com/GenericMappingTools/pygmt/pull/3116/commits/c8772bff9998b865046efb5ba416b1b1abec7962 and https://github.com/GenericMappingTools/pygmt/pull/3116/commits/9a9ae495aaacdc8d69db7944eea39177cb127bb7.

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,40 @@ def grdcontour(self, grid, **kwargs):
...
-    ...     # set the projection to Mercator for the 10 cm figure
+    ...     # Set the projection to Mercator for the 10 cm figure
     ...     projection="M10c",
     ... )
-    >>> # show the plot
+    >>> # Show the plot
     >>> fig.show()
     """
     kwargs = self._preprocess(**kwargs)
+
```
seisman, line 129: Please add a few lines of comments explaining the code block.

pygmt/tests/test_grdcontour.py
```diff
@@ -33,6 +33,32 @@ def test_grdcontour(grid):
     return fig
 
 
+def test_grdcontour_one_contour(grid):
```
seisman, line 36:
```suggestion
@pytest.mark.mpl_image_compare
def test_grdcontour_one_contour(grid):
```

pygmt/tests/test_grdcontour.py
```diff
@@ -33,6 +33,32 @@ def test_grdcontour(grid):
...
+    Plot a contour image using an xarray grid with one fixed contour.
+    """
+    fig = Figure()
+    fig.grdcontour(
+        grid=grid, interval=[50], annotation=[200], projection="M10c", frame=True
+    )
+    return fig
+
+
+def test_grdcontour_specific_contours(grid):
```
seisman, line 47:
```suggestion
@pytest.mark.mpl_image_compare
def test_grdcontour_specific_contours(grid):
```

pygmt/tests/test_grdcontour.py
yvonnefroehlich: Ops :see_no_evil:. Thanks!

seisman:
Still need to add the baseline images to dvc, and push them to the remote. See https://www.pygmt.org/dev/contributing.html#using-data-version-control-dvc-to-manage-test-images for the detailed instructions.

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,44 @@ def grdcontour(self, grid, **kwargs):
...
+    # Specify levels for contours or annotations
+    # One level is converted to a string with a trailing comma to separate it from
+    # specifying an interval
+    # Multiple levels are concatenated to a comma-separated string
```
seisman, line 133:
```suggestion
    # Specify levels for contours or annotations.
    # One level is converted to a string with a trailing comma to separate it from
    # specifying an interval.
    # Multiple levels are concatenated to a comma-separated string.
```

seisman (approved):
Looks good. Please also update the PR title to reflect the changes.

yvonnefroehlich:
> Looks good. Please also update the PR title to reflect the changes.

I updated the title (a bit long now) and the description in the initial comment of this PR.

yvonnefroehlich:
Do we actually have to mention / warn that, due to removing `sequence_plus` for `annotation` arguments like `[100, "f10p", "gred"]` are no longer supported and `"100+f10p+gred"` should be used instead?

seisman:
> due to removing `sequence_plus` for `annotation` arguments like `[100, "f10p", "gred"]` are no longer supported and `"100+f10p+gred"` should be used instead?

The syntax is awkward. So maybe no one is actually using it. Or we may have to try hard to check if a list like `[100, "f10p", "gred"]` is passed.

pygmt/src/grdcontour.py
```diff
@@ -96,32 +100,44 @@ def grdcontour(self, grid, **kwargs):
...
     ...     frame="a",
-    ...     # set the projection to Mercator for the 10 cm figure
+    ...     # Set the projection to Mercator for the 10 cm figure
     ...     projection="M10c",
     ... )
-    >>> # show the plot
+    >>> # Show the plot
     >>> fig.show()
     """
     kwargs = self._preprocess(**kwargs)
```
seisman, line 128: Add some codes for backward-compatibility:
```suggestion
    kwargs = self._preprocess(**kwargs)
    
    # Backward compatibility with the old syntax like (100, "e", "f10p", "gred").
    if is_nonstr_iter(kwargs.get("A")) and any(
        i[0] in "acdefgijlLnoprtuvwx=" for i in kwargs("A") if isinstance(i, str)
    ):
        msg = (
            "Argument of the parameter 'annotation'/'A' is using the old, deprecated "
            "syntax. Please refer to the PyGMT documentation for the new syntax. "
            "The warning will be removed in v0.14.0 and the old syntax will no longer "
            "be supported. "
        )
        warnings.warn(msg, catalog=FutureWarning, stacklevel=2)
        kwargs["A"] = "+".join(f"{item}" for item in kwargs["A"])
```

pygmt/src/grdcontour.py
```diff
@@ -96,32 +102,57 @@ def grdcontour(self, grid, **kwargs):
...
+
+    # Backward compatibility with the old syntax like (100, "e", "f10p", "gred").
+    if is_nonstr_iter(kwargs.get("A")) and any(
+        i[0] in "acdefgijlLnoprtuvwx=" for i in kwargs("A") if isinstance(i, str)
```
seisman, line 134: Typo.
```suggestion
        i[0] in "acdefgijlLnoprtuvwx=" for i in kwargs["A"] if isinstance(i, str)
```

pygmt/src/grdcontour.py
```diff
@@ -96,32 +102,57 @@ def grdcontour(self, grid, **kwargs):
...
+    ...     # Set the projection to Mercator for the 10 cm figure
     ...     projection="M10c",
     ... )
-    >>> # show the plot
+    >>> # Show the plot
     >>> fig.show()
     """
     kwargs = self._preprocess(**kwargs)
+
+    # Backward compatibility with the old syntax like (100, "e", "f10p", "gred").
```
seisman, line 132: Still need to add a test to increase the code coverage. Instead of writing a new test with a new baseline image, you can create a new test based on `test_grdcontour_one_level`.

I think the one below should produce an image that is identical to `test_grdcontour_one_level.png`:
```
    fig = Figure()
    fig.grdcontour(
        grid=grid, interval=[400], annotation=[570, "pblack"], projection="M10c", frame=True
    )
```

yvonnefroehlich: This test is added in commit https://github.com/GenericMappingTools/pygmt/pull/3116/commits/a05ec74ec721dc56bb6cf2f07d2627e60247f632.

pygmt/src/grdcontour.py
yvonnefroehlich: Corrected in commit https://github.com/GenericMappingTools/pygmt/pull/3116/commits/be21572284b1a698cf03d2ca7418b213642f8c0c.

PR #3119, opened 2024-03-18
Migrate from os.path to pathlib

### LINKED ISSUE #1834: Replace os.path with pathlib
[`pathlib`](https://docs.python.org/3.8/library/pathlib.html) is the standard and preferred way to manipulate paths. We should migrate `os.path` calls to `pathlib`.

References:

- https://docs.python.org/3.8/library/pathlib.html
- https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

seisman:
Migrate from `os.path` to `pathlib.Path` following this table: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module.

Please note that:
```
if os.path.exists(fname):
   os.unlink(fname)
```
is equivalent to `Path(fname).unlink(missing_ok=True)`.
```
assert os.path.exists(fname)
os.unlink(fname)
```
is equivalent to `Path(fname).unlink(missing_ok=False)` or just `Path(fname).unlink()`.

Closes #1834.


seisman:
Will merge the PR in 24 hours if no objections.

weiji14:
Do you want to enable ruff's PTH rules in this PR as mentioned at https://github.com/GenericMappingTools/pygmt/issues/1834#issuecomment-1862458307?

seisman:
Yes, but enabling PTH rules result in many violations against the [PTH123 rule](https://docs.astral.sh/ruff/rules/builtin-open/).

Better to do it in a separate PR and also need to discuss if we want to replace `open` with `Path().open()`.

pygmt/tests/test_figure.py
```diff
@@ -82,11 +81,8 @@ def test_figure_savefig_exists():
     fig.basemap(region="10/70/-300/800", projection="X3i/5i", frame="af")
     prefix = "test_figure_savefig_exists"
     for fmt in "bmp eps jpg jpeg pdf png ppm tif PNG JPG JPEG Png".split():
-        fname = f"{prefix}.{fmt}"
+        fname = Path(f"{prefix}.{fmt}")
         fig.savefig(fname)
-
-        fname = Path(fname)
-        assert fname.exists()
```
weiji14, line 89: Should this `.exists()` be kept?

pygmt/tests/test_figure.py
```diff
@@ -201,10 +197,9 @@ def test_figure_savefig_transparent():
         with pytest.raises(GMTInvalidInput):
             fig.savefig(fname, transparent=True)
     # png should not raise an error
-    fname = f"{prefix}.png"
+    fname = Path(f"{prefix}.png")
     fig.savefig(fname, transparent=True)
-    assert os.path.exists(fname)
```
weiji14, line 206: Need to check if file exists here?

pygmt/tests/test_psconvert.py
```diff
@@ -18,9 +18,7 @@ def test_psconvert():
     fig.basemap(region="10/70/-3/8", projection="X4i/3i", frame="a")
     prefix = "test_psconvert"
     fig.psconvert(prefix=prefix, fmt="f", crop=True)
-    fname = prefix + ".pdf"
-    assert os.path.exists(fname)
-    os.remove(fname)
+    Path(prefix + ".pdf").unlink()
```
weiji14, line 21: Add back `.exists()`?

pygmt/tests/test_psconvert.py
```diff
@@ -32,14 +30,10 @@ def test_psconvert_twice():
     prefix = "test_psconvert_twice"
     # Make a PDF
     fig.psconvert(prefix=prefix, fmt="f")
-    fname = prefix + ".pdf"
-    assert os.path.exists(fname)
-    os.remove(fname)
+    Path(prefix + ".pdf").unlink()
```
weiji14, line 33: Add back `.exists()`?

pygmt/tests/test_psconvert.py
```diff
@@ -32,14 +30,10 @@ def test_psconvert_twice():
...
     # Make a PNG
     fig.psconvert(prefix=prefix, fmt="g")
-    fname = prefix + ".png"
-    assert os.path.exists(fname)
-    os.remove(fname)
+    Path(prefix + ".png").unlink()
```
weiji14, line 36: Add back `.exists()`?

weiji14 (commented):
I'm seeing that a lot of the `assert os.path.exists()` lines have been removed, but not replaced with `Path.exists()`, possibly because there is a `Path.unlink()` directly afterwards that would raise a `FileNotFoundError` if the file was missing (see https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink), is that correct?

I think there are some cases where we do need to explicitly do `assert Path.exists()`, especially if the test name is something like `test_file_exists`. Otherwise it's not apparent that `.unlink()` actually implicitly checks if the file exists already.

weiji14:
> Yes, but enabling PTH rules result in many violations against the [PTH123 rule](https://docs.astral.sh/ruff/rules/builtin-open/).

We could also apply `PTH`, but ignore `PTH123` for now?

seisman:
> I think there are some cases where we do need to explicitly do `assert Path.exists()`, especially if the test name is something like `test_file_exists`. Otherwise it's not apparent that `.unlink()` actually implicitly checks if the file exists already.

Added back in ee0ba185a0f04f70622bffaad1fc1c92dc9ca195.



> > Yes, but enabling PTH rules result in many violations against the [PTH123 rule](https://docs.astral.sh/ruff/rules/builtin-open/).
> 
> We could also apply `PTH`, but ignore `PTH123` for now?

In some cases migrating from `open` to `Path.open` makes sense, so I prefer to do it in a separate PR.

weiji14 (approved):

PR #3120, opened 2024-03-18
Refactor the _load_remote_dataset function to load tiled and non-tiled grids in a consistent way

seisman:

The `_load_remote_dataset` function uses two different ways when loading tiled and non-tiled remote datasets.

- For non-tiled grids, it calls `which` to download the grid, get the full path and load it by calling `load_dataarray` (i.e., `xr.load_dataarray`)
- For tiled grids, it calls `grdcut`, which downloads the needed tiles, assembles them into a single grid and loads it by calling `load_dataarray`.

Currently, both tiled and non-tiled grids are loaded from a grid file on disk by `xr.load_dataarray`, so the returned xarray.DataArray are similar for tiled and non-tiled grids. The only difference is, the source file given in `grid.encoding["source"]` doesn't exist for tiled grids, because the assembled grid is a temporary file.

We're working on refactoring the `grdcut` wrapper to use virtual files in #3115. After #3115 is finished, the returned xarray.DataArray would have different attributes for tiled and non-tiled grids. So, it's better to have a consistent way for loading tiled and non-tiled grids. This can be done by calling the special `read` module. Its syntax looks like (the special `read` module is only available for external wrappers, so the command doesn't work in bash):
```
gmt read @earth_relief_01d_g vfile -Tg -R0/10/0/10
```
in which, `vfile` can be an actual file name or a virtual file name, and `-Tg` means the input is a grid.

### DIFF
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -4,12 +4,12 @@ Internal function to load GMT remote datasets.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, ClassVar, NamedTuple
+from typing import TYPE_CHECKING, ClassVar, Literal, NamedTuple
 
+from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import kwargs_to_strings
-from pygmt.io import load_dataarray
-from pygmt.src import grdcut, which
+from pygmt.helpers import build_arg_list, kwargs_to_strings
+from pygmt.src import which
 
 if TYPE_CHECKING:
     import xarray as xr
@@ -344,7 +344,7 @@ def _load_remote_dataset(
     dataset_prefix: str,
     resolution: str,
     region: str | list,
-    registration: str,
+    registration: Literal["gridline", "pixel", None],
 ) -> xr.DataArray:
     r"""
     Load GMT remote datasets.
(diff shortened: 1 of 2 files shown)


pygmt/datasets/load_remote_dataset.py
```diff
@@ -370,54 +370,68 @@ def _load_remote_dataset(
...
-        grid = load_dataarray(fname, engine="netcdf4")
-    else:
-        grid = grdcut(f"@{dataset_prefix}{resolution}{reg}", region=region)
+            grid = lib.virtualfile_to_raster(outgrid=None, vfname=voutgrd)
+
+    # Full path to the grid if not tiled grids.
+    source = which(fname, download="a") if not resinfo.tiled else None
+    # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
+    if source:
+        grid.encoding["source"] = source
```
seisman, line 434: Currently, the GMT accessors require the `grid.encoding["source"]` to work. So, for non-tiled gris, we get the full path to the grid and set `grid.encoding["source"]`. For tiled grids, `grid.encoding["source"]` is undefined.

pygmt/datasets/load_remote_dataset.py
```diff
@@ -370,54 +370,68 @@ def _load_remote_dataset(
...
-    reg = f"_{registration[0]}"
 
-    # different ways to load tiled and non-tiled grids.
-    # Known issue: tiled grids don't support slice operation
-    # See https://github.com/GenericMappingTools/pygmt/issues/524
-    if region is None:
-        if dataset.resolutions[resolution].tiled:
-            raise GMTInvalidInput(
-                f"'region' is required for {dataset.title} resolution '{resolution}'."
+    fname = f"@{dataset_prefix}{resolution}_{registration[0]}"
```
seisman, line 414: I also propose to 

- Rename `dataset_name` to `name`
- Rename `dataset_prefix` to `prefix`
- Remove the trailing `_` from `dataset_prefix`, e.g., `dataset_prefix="earth_relief_"` should be `prefix="earth_relief"`.

seisman: I feel it's better to do it in a separate PR to make this PR small for review.

pygmt/datasets/load_remote_dataset.py
```diff
@@ -370,54 +370,68 @@ def _load_remote_dataset(
...
+    if resinfo.tiled and region is None:
+        raise GMTInvalidInput(
+            f"'region' is required for {dataset.title} resolution '{resolution}'."
+        )
+
+    # Currently, only grids are supported. Will support images in the future.
```
seisman, line 420: Please note that: currently we use `-Tg` so it only works for grids. For images, we need to use `-Ti`. 

To support images, we need to know whether the remote dataset is a grid or an image. It means we need to add the data `kind` information to the `GMTRemoteDataset` class.

pygmt/datasets/load_remote_dataset.py
seisman: Issue opened at #3190.

michaelgrund (approved):

PR #3124, opened 2024-03-19
clib: Change the parameter order and set output_type to pandas in virtualfile_to_dataset

seisman:

For some module wrappers like `which`, `output_type` makes no sense because the a "file" or "numpy" output is useless. When refactoring these module wrappers, we have to use
```
lib.virtualfile_to_dataset(output_type="pandas", vfname=vouttbl)
```
which doesn't look good.

This PR move the `vfname` parameter before the `output_type` parameter and set `output_type`'s default to `"pandas"`, so that we can use a much shorter version
```
lib.virtualfile_to_dataset(vfname=vouttbl)
```

Putting `vfname` as the first parameter makes more sense because `vfname` is always a required parameter.

Patches #3083.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1740,8 +1740,8 @@ class Session:
 
     def virtualfile_to_dataset(
         self,
-        output_type: Literal["pandas", "numpy", "file"],
         vfname: str,
+        output_type: Literal["pandas", "numpy", "file"] = "pandas",
         column_names: list[str] | None = None,
     ) -> pd.DataFrame | np.ndarray | None:
         """
@@ -1751,15 +1751,15 @@ class Session:
 
         Parameters
         ----------
+        vfname
+            The virtual file name that stores the result data. Required for ``"pandas"``
+            and ``"numpy"`` output type.
         output_type
             Desired output type of the result data.
 
             - ``"pandas"`` will return a :class:`pandas.DataFrame` object.
             - ``"numpy"`` will return a :class:`numpy.ndarray` object.
             - ``"file"`` means the result was saved to a file and will return ``None``.
-        vfname
-            The virtual file name that stores the result data. Required for ``"pandas"``
-            and ``"numpy"`` output type.
         column_names
             The column names for the :class:`pandas.DataFrame` output.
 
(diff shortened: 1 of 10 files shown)


pygmt/clib/session.py
```diff
@@ -1795,7 +1795,8 @@ def virtualfile_to_dataset(
         ...             ) as vouttbl:
         ...                 lib.call_module("read", f"{tmpfile.name} {vouttbl} -Td")
         ...                 result = lib.virtualfile_to_dataset(
-        ...                     output_type="file", vfname=vouttbl
+        ...                     vfname=vouttbl,
+        ...                     output_type="file",
```
weiji14, line 1799: Should be ok to put this in a single line right?
```suggestion
        ...                     vfname=vouttbl, output_type="file"
```

pygmt/clib/session.py
```diff
@@ -1805,7 +1806,8 @@ def virtualfile_to_dataset(
         ...         with lib.virtualfile_out(kind="dataset") as vouttbl:
         ...             lib.call_module("read", f"{tmpfile.name} {vouttbl} -Td")
         ...             outnp = lib.virtualfile_to_dataset(
-        ...                 output_type="numpy", vfname=vouttbl
+        ...                 vfname=vouttbl,
+        ...                 output_type="numpy",
```
weiji14, line 1810:
```suggestion
        ...                 vfname=vouttbl, output_type="numpy"
```

pygmt/src/filter1d.py
```diff
@@ -123,4 +123,7 @@ def filter1d(
                 module="filter1d",
                 args=build_arg_string(kwargs, infile=vintbl, outfile=vouttbl),
             )
-        return lib.virtualfile_to_dataset(output_type=output_type, vfname=vouttbl)
+        return lib.virtualfile_to_dataset(
+            vfname=vouttbl,
+            output_type=output_type,
+        )
```
weiji14, line 129:
```suggestion
        return lib.virtualfile_to_dataset(vfname=vouttbl, output_type=output_type)
```

pygmt/clib/session.py
```diff
@@ -1751,15 +1751,15 @@ def virtualfile_to_dataset(
...
             Desired output type of the result data.
 
             - ``"pandas"`` will return a :class:`pandas.DataFrame` object.
             - ``"numpy"`` will return a :class:`numpy.ndarray` object.
             - ``"file"`` means the result was saved to a file and will return ``None``.
-        vfname
-            The virtual file name that stores the result data. Required for ``"pandas"``
-            and ``"numpy"`` output type.
         column_names
             The column names for the :class:`pandas.DataFrame` output.
```
weiji14, line 1764: If the proof of concept at #3117 works, these lines will need to be updated (after we add some logic to use the header column names in this `virtualfile_to_dataset` method)

weiji14 (approved):
Pre-approving, since the suggestions are just stylistic changes.

PR #3128, opened 2024-03-20
GMT_IMAGE: Implement the GMT_IMAGE.to_dataarray method for 3-band images

seisman:
This PR implements the `to_dataarray` method for `GMT_IMAGE`.

Similar to #2398 but for images.

~~Need to wait for #3446.~~


pygmt/clib/session.py
```diff
@@ -1672,12 +1674,15 @@ def virtualfile_out(
             family, geometry = {
                 "dataset": ("GMT_IS_DATASET", "GMT_IS_PLP"),
                 "grid": ("GMT_IS_GRID", "GMT_IS_SURFACE"),
+                "image": ("GMT_IS_IMAGE", "GMT_IS_SURFACE"),
             }[kind]
-            with self.open_virtualfile(family, geometry, "GMT_OUT", None) as vfile:
+            with self.open_virtualfile(
+                family, geometry, "GMT_OUT|GMT_IS_REFERENCE", None
```
seisman, line 1680: For unknown reasons (likely upstream bug), using `GMT_OUT` crashes and `GMT_OUT|GMT_IS_REFERENCE` works for `image`. So better to use `GMT_OUT` for `dataset`/`grid` and `GMT_OUT|GMT_IS_REFERENCE` for `image`.

We also need to open an upstream issue and find out why.

seisman:
xref https://github.com/GenericMappingTools/gmt/issues/3299

pygmt/datatypes/image.py
```diff
@@ -0,0 +1,102 @@
...
+        data: np.ndarray = np.reshape(
+            a=self.data[: header.n_bands * header.mx * header.my],
+            newshape=(header.my, header.mx, header.n_bands),
+        )[pad[2] : header.my - pad[3], pad[0] : header.mx - pad[1], :]
+
+        # Get x and y coordinates
+        coords: dict[str, list] = {
+            "x": self.x[: header.n_columns],
+            "y": self.y[: header.n_rows],
+            "band": np.arange(stop=3, dtype=np.uint8),
```
weiji14, line 96: Having trouble with `mypy` on this `np.arange` line. Getting an error like `pygmt/datatypes/image.py:96: error: No overload variant of "arange" matches argument types "int", "type[unsignedinteger[Any]]"  [call-overload]` (full error at https://github.com/GenericMappingTools/pygmt/actions/runs/9562556451/job/26359180290?pr=3128#step:5:13). Do you know what's the proper syntax?

seisman: Maybe we can just use `np.array([0, 1, 2], dtype=np.uint8)` instead?

weiji14: Ah yes, that works. Thanks!

pygmt/datatypes/image.py
```diff
@@ -0,0 +1,108 @@
...
+            "band": np.array([0, 1, 2], dtype=np.uint8),
+        }
+
+        # Create the xarray.DataArray object
+        image = xr.DataArray(
+            data=data,
+            coords=coords,
+            dims=("y", "x", "band"),
+            name=header.name,
+            attrs=header.data_attrs,
```
weiji14, line 105: The image `name` is currently hardcoded to `z`, is that ok for an RGB image?

https://github.com/GenericMappingTools/pygmt/blob/ac447063e95ed70c68ca8cd2eebc20c8318c63bf/pygmt/datatypes/header.py#L193-L198

The `attrs` fields might need some work. I'm getting ` 'actual_range': array([ 1.79769313e+308, -1.79769313e+308])}` when loading the `@earth_day_01d` image.

seisman: I agree they make no sense, but they're consistent with the behavior in GMT. 
```
gmt grdinfo @earth_day_01d
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Title: Grid imported via GDAL
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Command:
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Remark:
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Pixel node registration used [Geographic grid]
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Grid file format: gd = Import/export through GDAL
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: x_min: -180 x_max: 180 x_inc: 1 name: x n_columns: 360
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: y_min: -90 y_max: 90 y_inc: 1 name: y n_rows: 180
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: v_min: 1.79769313486e+308 v_max: -1.79769313486e+308 name: z
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: scale_factor: 1 add_offset: 0
/Users/seisman/.gmt/server/earth/earth_day/earth_day_01d_p.tif: Default CPT:
+proj=longlat +R=6378137 +no_defs
```
The GMT's image support was likely added by Joaquim so that you may ping him for more information.

pygmt/datatypes/header.py
```diff
@@ -203,7 +203,8 @@ def data_attrs(self) -> dict[str, Any]:
         Attributes for the data variable from the grid header.
         """
         attrs: dict[str, Any] = {}
-        attrs["Conventions"] = "CF-1.7"
+        if self.type == 18:  # Grid file format: ns = GMT netCDF format
+            attrs["Conventions"] = "CF-1.7"
```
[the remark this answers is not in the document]
seisman: The IDs are defined here https://github.com/GenericMappingTools/gmt/blob/79033c0cab34961fe18bfad8ad5d0719785485ee/src/gmt_grdio.h#L74.

pygmt/datatypes/image.py
```diff
@@ -92,3 +94,114 @@ class _GMT_IMAGE(ctp.Structure):  # noqa: N801
...
+        (1, 0)
+        """
+
+        # Get image header
+        header: _GMT_GRID_HEADER = self.header.contents
+
+        # Get DataArray without padding
+        pad = header.pad[:]
+        data: np.ndarray = np.reshape(
+            a=self.data[: header.n_bands * header.mx * header.my],
```
seisman, line 183:
```suggestion
            self.data[: header.n_bands * header.mx * header.my],
```

pygmt/datatypes/header.py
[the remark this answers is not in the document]
seisman: Addressed in PR #3463 and #3449.

seisman:
I think this PR is almost ready for review. Here are the known limitations/issues:

- Currently, only 3-band images are supported. [We can work on other images in a separate PR to make it easier to review]
- As mentioned in https://github.com/GenericMappingTools/pygmt/pull/3128#discussion_r1644203991, the name is hardcoded to `z` and the `actual_range` doesn't work. [For the hardcoded name `z`, I'm not sure if we can do better; for `actual_range`, I think we can remove it because (1) `actual_range` makes no sense for 3-band RGB images; (2) `actual_range` is the standard attribution defined by CF-1.7 convention, but it's only for netCDF files, not for images; (3) the xarray.DataArray returned by `rioxarray.open_rasterio` doesn't have the `actual_range` attribution].
- For the `y` coordinates, both GMT API and `rioxarray` return a descending `y` array, but in `_GMT_GRID.to_dataarray()` method, we flip the y coordinate to make it ascending. Do we want to do the same thing for images?

pygmt/datatypes/image.py
```diff
@@ -91,3 +95,122 @@ class _GMT_IMAGE(ctp.Structure):  # noqa: N801
...
+                [185, 187, 187, ..., 187, 186, 185],
+                [189, 191, 191, ..., 191, 191, 189]]], dtype=uint8)
+        Coordinates:
+          * y        (y) float64... 89.5 88.5 87.5 86.5 ... -86.5 -87.5 -88.5 -89.5
+          * x        (x) float64... -179.5 -178.5 -177.5 -176.5 ... 177.5 178.5 179.5
+          * band     (band) uint8... 1 2 3
+        Attributes:
+            title:
+            history:
+            description:
```
seisman, line 150: Maybe don't add these attributes if they're empty?

weiji14: These come from the header no? For GeoTIFF files, they don't make sense since GMT is not parsing the TIFF tags, but the title/history/description might show up for 3-band NetCDF files (if those can be passed into this `GMT_IMAGE` struct).

seisman: Right, `title`/`history`/`description` are recommended attributes of CF-1.7. So, for netCDF files, I guess we should always show them even if they're empty.

weiji14 (commented):
> * Currently, only 3-band images are supported. [We can work on 1-band and 4-band images in a separate PR to make it easier to review]

Yep, just do 3-band images here for now.

> * For the `y` coordinates, both GMT API and `rioxarray` return a descending `y` array, but in `_GMT_GRID.to_dataarray()` method, we flip the y coordinate to make it ascending. Do we want to do the same thing for images?

I'd prefer descending `y` to match that the output of `rioxarray.open_rasterio`. In https://github.com/GenericMappingTools/pygmt/pull/2398#discussion_r1566598948, we kept the ascending y for `GMT_GRID` for pragmatic reasons, namely to match the output of `xr.load_datarray` so that the unit tests won't break. But I think the convention for most `xarray.DataArray` objects is to have descending y and ascending x.

pygmt/datatypes/image.py
seisman: Done at 62f0ce056cb6bdd32521905529246d67faae2b8d.

pygmt/datatypes/image.py
```diff
@@ -91,3 +95,122 @@ class _GMT_IMAGE(ctp.Structure):  # noqa: N801
...
+        <xarray.DataArray 'y' (y: 180)>...
+        array([ 89.5,  88.5,  87.5,  86.5,  ...,   -87.5, -88.5, -89.5])
+        Coordinates:
+          * y        (y) float64... 89.5 88.5 87.5 86.5 ... -86.5 -87.5 -88.5 -89.5
+        Attributes:
+            long_name:     y
+            axis:          Y
+            actual_range:  [-90.  90.]
+        >>> da.gmt.registration, da.gmt.gtype
+        (1, 0)
```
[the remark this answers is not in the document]
seisman: https://github.com/GenericMappingTools/gmt/blob/7809736ba32d87a4a96b15444419eb176c6a35f3/src/gmt_grdio.c#L1360

This is the function (again, not a public API function) that GMT uses to determine the grid type. The logic here is not that complicated to duplicate in Python.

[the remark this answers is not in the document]
weiji14: How easy would it be to make that C function public actually?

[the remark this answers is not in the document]
seisman: > How easy would it be to make that C function public actually?

It doesn't seem too difficult. But we still need to implement our own functions if we want to be compatible with GMT 6.4-6.5.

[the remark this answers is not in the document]
weiji14: Yeah, it'll be some time before GMT 6.6 comes out. I just want the gtype determination to be consistent between GMT and PyGMT, and the best case is to bind to the GMT C function. But we can also reimplement it in PyGMT for now (in a separate PR), maybe in header.py?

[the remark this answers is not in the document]
seisman: Another possibility is to wrap the "hidden", private `GMT_GRID_HEADER_HIDDEN` structure https://github.com/GenericMappingTools/gmt/blob/7809736ba32d87a4a96b15444419eb176c6a35f3/src/gmt_hidden.h#L151. The structure has many members but we can only wrap the first members. The members that are most useful to us are: `grdtype` (Cartesian or geographic), `BC` (boundary condition), `varname` (the actual variable name so that we don't have to hardcode the grid name to `z`), and `cpt` (the default CPT for this grid).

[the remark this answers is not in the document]
seisman: In f715aee1d1a5497d4018c2db32d83d77661f88d8, I've updated the logic for determining the grid/image gtype based on `ProjRefPROJ4`. 

The logic of codes come from: https://github.com/GenericMappingTools/gmt/blob/7809736ba32d87a4a96b15444419eb176c6a35f3/src/gmt_grdio.c#L3778

weiji14 (approved):
Wonderful, thanks again @seisman! Excited to have direct 3-band read support in PyGMT without rioxarray! There's still a few tiny details that could be improved (around the attrs/metadata), but this looks really good already as an initial implementation. :rocket:

PR #3131, opened 2024-03-21
GMT_DATASET.to_dataframe: Return an empty DataFrame if a file contains no data

seisman:
Sometimes, GMT may output nothing to a virtual file, for example, `gmt select` may find no data points that satisfy the specified criteria. 

In such cases, `GMT_DATASET.to_dataframe` raises an exception. As shown below:
```python
In [1]: import pandas as pd

In [2]: pd.concat(objs=[], axis=1)
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 pd.concat(objs=[], axis=1)

File ~/opt/miniconda/envs/pygmt/lib/python3.12/site-packages/pandas/core/reshape/concat.py:382, in concat(objs, axis, join, ignore_index, keys, levels, names, verify_integrity, sort, copy)
    379 elif copy and using_copy_on_write():
    380     copy = False
--> 382 op = _Concatenator(
    383     objs,
    384     axis=axis,
    385     ignore_index=ignore_index,
    386     join=join,
    387     keys=keys,
    388     levels=levels,
    389     names=names,
    390     verify_integrity=verify_integrity,
    391     copy=copy,
    392     sort=sort,
    393 )
    395 return op.get_result()

File ~/opt/miniconda/envs/pygmt/lib/python3.12/site-packages/pandas/core/reshape/concat.py:445, in _Concatenator.__init__(self, objs, axis, join, keys, levels, names, ignore_index, verify_integrity, copy, sort)
    442 self.verify_integrity = verify_integrity
    443 self.copy = copy
--> 445 objs, keys = self._clean_keys_and_objs(objs, keys)
    447 # figure out what our result ndim is going to be
    448 ndims = self._get_ndims(objs)

File ~/opt/miniconda/envs/pygmt/lib/python3.12/site-packages/pandas/core/reshape/concat.py:507, in _Concatenator._clean_keys_and_objs(self, objs, keys)
    504     objs_list = list(objs)
    506 if len(objs_list) == 0:
--> 507     raise ValueError("No objects to concatenate")
    509 if keys is None:
    510     objs_list = list(com.not_none(*objs_list))

ValueError: No objects to concatenate
```
Instead of raising exceptions, `GMT_DATASET.to_dataframe` should return a reasonable value. I think an empty DataFrame (i.e., `pd.DataFrame()`) makes more sense than ``None``.

This PR updates the `GMT_DATASET.to_dataframe()` method to return an empty DataFrame in such cases and also add two tests (one for benchmark and one for testing the empty DataFrame).


pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+    df = df.convert_dtypes(
+        convert_string=True,
+        convert_integer=False,
+        convert_boolean=False,
+        convert_floating=False,
+    )
+    return df
+
+
+def dataframe_from_gmt(fname):
```
seisman, line 42: For reference, GMT provides two special/undocumented modules `read` and `write` (their source codes are `gmt/src/gmtread.c`/`gmt/src/gmtwrite.c`) that can read a file into a GMT object (e.g, reading a tabular file as GMT_DATASET, or reading a grid as GMT_GRID). Currently, we're frequently using the special `read` module in the doctest of the `pygmt.clib.session` module (similar to lines 46-50 below). We may want to make it public in the future as already done in GMT.jl (https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.gmtread-Tuple{String} and https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.gmtwrite).

pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pandas as pd
+import pytest
+from pygmt.clib import Session
+from pygmt.helpers import GMTTempFile
+
+
+def dataframe_from_pandas(filepath_or_buffer, sep=r"\s+", comment="#", header=None):
+    """
+    Read a tabular data as pandas.DataFrame object using pandas.read_csv().
```
yvonnefroehlich, line 15:
```suggestion
    Read tabular data as pandas.DataFrame object using pandas.read_csv().
```

pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+
+    The parameters have the same meaning as in ``pandas.read_csv()``.
+    """
+    try:
+        df = pd.read_csv(filepath_or_buffer, sep=sep, comment=comment, header=header)
+    except pd.errors.EmptyDataError:
+        # Return an empty DataFrame if the file has no data
```
yvonnefroehlich, line 22:
```suggestion
        # Return an empty DataFrame if the file contains no data
```

pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        return pd.DataFrame()
+
+    # By default, pandas reads text strings with whitespaces as multiple columns, but
+    # GMT contacatenates all trailing text as a single string column. Neet do find all
+    # string columns (with dtype="object") and combine them into a single string column.
```
yvonnefroehlich, line 27:
```suggestion
    # By default, pandas reads text strings with whitespaces as multiple columns, but
    # GMT concatenates all trailing text as a single string column. Need do find all
    # string columns (with dtype="object") and combine them into a single string column.
```

pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+        convert_integer=False,
+        convert_boolean=False,
+        convert_floating=False,
+    )
+    return df
+
+
+def dataframe_from_gmt(fname):
+    """
+    Read a tabular data as pandas.DataFrame using GMT virtual file.
```
yvonnefroehlich, line 44:
```suggestion
    Read tabular data as pandas.DataFrame using GMT virtual file.
```

pygmt/tests/test_datatypes_dataset.py
```diff
@@ -0,0 +1,83 @@
...
+            print("10.0 11.0 12.0 TEXT123 TEXT456789", file=fp)
+
+        df = dataframe_from_gmt(tmpfile.name)
+        expected_df = dataframe_from_pandas(tmpfile.name, comment=">")
+        pd.testing.assert_frame_equal(df, expected_df)
+
+
+def test_dataset_empty():
+    """
+    Make sure that an empty DataFrame is returned if a file has no data.
```
yvonnefroehlich, line 74:
```suggestion
    Make sure that an empty DataFrame is returned if a file contains no data.
```

pygmt/datatypes/dataset.py
```diff
@@ -211,5 +213,5 @@ def to_dataframe(self) -> pd.DataFrame:
                 pd.Series(data=np.char.decode(textvector), dtype=pd.StringDtype())
             )
 
-        df = pd.concat(objs=vectors, axis=1)
+        df = pd.concat(objs=vectors, axis=1) if vectors else pd.DataFrame()
```
weiji14, line 216: The empty `pd.DataFrame()` won't have any columns. Should there still be columns returned (even if there are no rows)? How would this work with #3117 for example?

seisman: I guess it's fine.
```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame(columns=None)

In [3]: df.empty
Out[3]: True

In [4]: df = pd.DataFrame(columns=None)

In [5]: df = pd.DataFrame(columns=["col1", "col2"])

In [6]: df.empty
Out[6]: True
```

michaelgrund (approved):

pygmt/datatypes/dataset.py
weiji14: Column names are set here like so:

https://github.com/GenericMappingTools/pygmt/blob/1eb6dec5d72a29bc00349459d1369eb0a0673a93/pygmt/clib/session.py#L1853-L1861

So we would do something like:

```python
import pandas as pd

df = pd.DataFrame(columns=None)
assert df.empty
df.columns = ["col1", "col2"]
```

But setting column names to `["col1", "col2"]` errors with:

```python-traceback
ValueError                                Traceback (most recent call last)
Cell In[8], line 1
----> 1 df.columns = ["col1", "col2"]

File ~/mambaforge/envs/pygmt/lib/python3.12/site-packages/pandas/core/generic.py:6310, in NDFrame.__setattr__(self, name, value)
   6308 try:
   6309     object.__getattribute__(self, name)
-> 6310     return object.__setattr__(self, name, value)
   6311 except AttributeError:
   6312     pass

File properties.pyx:69, in pandas._libs.properties.AxisProperty.__set__()

File ~/mambaforge/envs/pygmt/lib/python3.12/site-packages/pandas/core/generic.py:813, in NDFrame._set_axis(self, axis, labels)
    808 """
    809 This is called from the cython code when we set the `index` attribute
    810 directly, e.g. `series.index = [1, 2, 3]`.
    811 """
    812 labels = ensure_index(labels)
--> 813 self._mgr.set_axis(axis, labels)
    814 self._clear_item_cache()

File ~/mambaforge/envs/pygmt/lib/python3.12/site-packages/pandas/core/internals/managers.py:238, in BaseBlockManager.set_axis(self, axis, new_labels)
    236 def set_axis(self, axis: AxisInt, new_labels: Index) -> None:
    237     # Caller is responsible for ensuring we have an Index object.
--> 238     self._validate_set_axis(axis, new_labels)
    239     self.axes[axis] = new_labels

File ~/mambaforge/envs/pygmt/lib/python3.12/site-packages/pandas/core/internals/base.py:98, in DataManager._validate_set_axis(self, axis, new_labels)
     95     pass
     97 elif new_len != old_len:
---> 98     raise ValueError(
     99         f"Length mismatch: Expected axis has {old_len} elements, new "
    100         f"values have {new_len} elements"
    101     )

ValueError: Length mismatch: Expected axis has 0 elements, new values have 2 elements
```

seisman: I think we need to refactor `GMT_DATASET.to_dataframe()` to accept more panda-specific parameters (e.g., `column_names`, `index_col`). The the `virtualfile_to_dataset` will be called like:
```
 result = self.read_virtualfile(vfname, kind="dataset").contents.to_dataframe(columns=column_names) 
 if output_type == "numpy":  # numpy.ndarray output 
     return result.to_numpy() 
  
 return result  # pandas.DataFrame output
```

weiji14: Right, so moving these lines that handle the column names from `virtualfile_to_dataset` to `to_dataframe`:

https://github.com/GenericMappingTools/pygmt/blob/62eb5d6b9156612c89d4d0357904fd09e2dddc17/pygmt/clib/session.py#L1861-L1863

Do you want to just do that in #3117? Or have a separate PR to handle this?

seisman: Better to do it in a separate PR so that #3117 can focus on parsing the column names from header.

seisman: > Right, so moving these lines that handle the column names from `virtualfile_to_dataset` to `to_dataframe`:
> 
> https://github.com/GenericMappingTools/pygmt/blob/62eb5d6b9156612c89d4d0357904fd09e2dddc17/pygmt/clib/session.py#L1861-L1863
> 
> Do you want to just do that in #3117? Or have a separate PR to handle this?

Done in #3140, so need to refactor this PR after #3140 is merged.

pygmt/datatypes/dataset.py
```diff
@@ -226,8 +228,13 @@ def to_dataframe(
                 pd.Series(data=np.char.decode(textvector), dtype=pd.StringDtype())
             )
 
+        # Return an empty DataFrame if no columns are found.
+        if len(vectors) == 0:
+            return pd.DataFrame()
```
seisman, line 233: Currently, it returns an empty DataFrame without columns and rows, but an empty DataFrame with columns is also allowed, e.g., 
```
return pd.DataFrame(column=column_names)
```
I guess either is fine. I think we can use `return pd.DataFrame()` now and make changes if necessary.

pygmt/datatypes/dataset.py
```diff
@@ -226,11 +228,16 @@ def to_dataframe(
...
+
+        # Create a DataFrame object by concatenating multiple columns
         df = pd.concat(objs=vectors, axis="columns")
```
weiji14, line 236: Let's set the column_names, and also the dtypes below later, instead of returning an empty `pd.DataFame` directly:

```suggestion
        if len(vectors) == 0:
            df = pd.DataFrame(columns=column_names)
        else:
            # Create a DataFrame object by concatenating multiple columns
            df = pd.concat(objs=vectors, axis="columns")
```

pygmt/datatypes/dataset.py
weiji14: I think we should return the column_names, so that users who want to e.g. do `pd.concat` on multiple `pd.DataFrame` outputs from running an algorithm like `pygmt.select` can do so in a more straightforward way. Note that we should also set the dtypes of the columns properly, even if the rows are empty, otherwise the dtypes will all become `object`:

```python
df1 = pd.DataFrame(data=[[0, 1, 2]], columns=["x", "y", "z"])
print(df1.dtypes)
# x    int64
# y    int64
# z    int64
# dtype: object
df2 = pd.DataFrame(columns=["x", "y", "z"])
print(df2.dtypes)
# x    object
# y    object
# z    object
# dtype: object

pd.concat(objs=[df1, df2]).dtypes
# x    object
# y    object
# z    object
# dtype: object
```

See my other suggestion on not returning an empty `pd.DataFrame()` early, until the dtype is set.

pygmt/datatypes/dataset.py
```diff
@@ -226,11 +228,16 @@ def to_dataframe(
...
 
-        df = pd.concat(objs=vectors, axis="columns")
+        # Return an empty DataFrame if no columns are found.
+        if len(vectors) == 0:
+            df = pd.DataFrame(columns=column_names)
+        else:
+            # Create a DataFrame object by concatenating multiple columns
+            df = pd.concat(objs=vectors, axis="columns")
         if column_names is not None:  # Assign column names
             df.columns = column_names
```
seisman, line 238: Column names are already assigned if the DataFrame is empty. No need to assign again.
```suggestion
            # Create a DataFrame object by concatenating multiple columns
            df = pd.concat(objs=vectors, axis="columns")
            if column_names is not None:  # Assign column names
                df.columns = column_names
```

pygmt/datatypes/dataset.py
```diff
@@ -156,6 +157,8 @@ def to_dataframe(
         the same. The same column in all segments of all tables are concatenated. The
         trailing text column is also concatenated as a single string column.
 
+        If the object contains no data, an empty DataFrame will be returned.
```
weiji14, line 160:
```suggestion
        If the object contains no data, an empty DataFrame will be returned (with the
        column names and dtypes set if provided).
```

pygmt/datatypes/dataset.py
```diff
@@ -226,11 +229,16 @@ def to_dataframe(
                 pd.Series(data=np.char.decode(textvector), dtype=pd.StringDtype())
             )
 
-        df = pd.concat(objs=vectors, axis="columns")
-        if column_names is not None:  # Assign column names
-            df.columns = column_names
-        if dtype is not None:
+        # Return an empty DataFrame if no columns are found.
+        if len(vectors) == 0:
```
weiji14, line 233: Have the inline comment match the indent of the `else:` block below.
```suggestion
        if len(vectors) == 0:
            # Return an empty DataFrame if no columns are found.
```

weiji14 (approved):
Should probably find a way to test the case where an empty `pd.DataFrame` is returned with column names and set dtypes, but it looks a bit tricky :slightly_smiling_face:

Issue #3104, opened 2024-03-13
Figure.text sometimes doesn't work if the text contains single/double quotation marks


xumi1993:
### Description of the problem

I want to write text of `A'` on a figure with text box filled by a colour of `fill='230'`. When I use `position='CM'` to specify the position of the text, the `A' -G230` is shown in the figure. I think the attribute `fill='230'` have not been correctly parsed.
![ff6b8c3c-060b-4295-977a-684e0051dcd6](https://github.com/GenericMappingTools/pygmt/assets/7437523/4e216e03-6fa7-4690-9da7-bb74f899f44c)

### Minimal Complete Verifiable Example

```python
label = "A\'"
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X4c/4c", frame=True)
fig.text(
    text=label,
    position='CM',
    fill='230'    
)
fig.show()
```

### System information

```bash
PyGMT information:
  version: v0.11.0
System information:
  python: 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:00:33)  [Clang 13.0.1 ]
  executable: /Users/xumijian/miniconda3_arm/bin/python
  machine: macOS-14.2.1-arm64-arm-64bit
Dependency information:
  numpy: 1.24.2
  pandas: 2.2.1
  xarray: 2023.5.0
  netCDF4: 1.6.0
  packaging: 24.0
  contextily: None
  geopandas: None
  ipython: None
  rioxarray: None
  ghostscript: 10.02.1
GMT library information:
  cores: 10
  grid layout: rows
  image layout: 
  library path: /Users/xumijian/miniconda3_arm/lib/libgmt.dylib
  padding: 2
  plugin dir: /Users/xumijian/miniconda3_arm/lib/gmt/plugins
  share dir: /Users/xumijian/miniconda3_arm/share/gmt
  version: 6.3.0
```

seisman:
I can reproduce the issue. It's the single quote `'` that causes the trouble. A temporary workaround is using 
```
label = "A\\047"
```

xumi1993:
Thanks you very much for the temporary solution. Looking forward to the next version solving this issue.

seisman:
PR #3105 should fix the issue.

PR #3105, opened 2024-03-13
BREAKING: Support typesetting apostrophe (') and backtick (`)

seisman:
See #3476 for context. 

As explained in #3476, currently, we can't typeset the original apostrophes and backticks because they're mapped to `‘’` in PostScript. This PR adds the patch to support typesetting apostrophes and backticks.

Please note that this is a *breaking change*. The main motivation of this PR is to ensure what you type is what you get (WYTIWIG, similar to [WYSIWIG](https://en.wikipedia.org/wiki/WYSIWYG)). So I think it's breaking things in a good way.

See the script and images below for the changes:
```python
import pygmt

chars = "A ` ' ‘ ’ \" “ ” B"
fig = pygmt.Figure()
fig.basemap(
    projection="X4c/2c", region=[0, 4, 0, 2], frame=["WSen", f"x+l{chars}"]
)
fig.text(x=2, y=1, text=chars, font="20p")
fig.show()
```

| **Main branch** | **This PR** |
|---|---|
| main| RP-3105| 

Fixes #3476.

PR #2331, opened 2023-01-19
pygmt.config: Correctly resetting to default values that contain whitespaces

seisman:
**Description of proposed changes**

Fixes https://github.com/GenericMappingTools/pygmt/issues/2298

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3132, opened 2024-03-21
Refactor all wrappers to pass an argument list to Session.call_module

seisman:
### TL;DR

When calling a GMT module, we pass a single text string of CLI arguments (e.g., `"-R0/10/0/10 -JX10c -BWSen+tMy Title"`) to the C API. It has caused many troubles if an argument contains whitespaces (e.g., #247, #655, #2298 and more) or single/double quotation marks (e.g., #3104 and more), and we have to introduce some hacking solutions (e.g., #1487). 

Instead of passing a single text string, we can pass a list of argument string instead (e.g., `["-R0/10/0/10", "-JX10c", "-BWSen+tMy Title"]`. In this way, we no longer need any hack solutions and everything just works as expected!

### Technical backgrounds

Currently, the [`Session.call_module()`](https://github.com/GenericMappingTools/pygmt/blob/dd8e0cd87a7a0629aafb752ba8934cec4149f011/pygmt/clib/session.py#L594) method have two parameters: `module` is name of a GMT module, and `args` is the text string of CLI arguments. For example, we usually call the method like this way:
```
lib.call_module(module="basemap", args=build_arg_string(kwargs))
```
in which `build_arg_string` converts a keyword dictionary into a single text string for CLI arguments (e.g., `"-R0/10/0/10 -JX10c -BWSen+tMy Title"`).

Actually, the C API function [`GMT_Call_Module`](https://github.com/GenericMappingTools/gmt/blob/e5ef7016eaccce3fac58419f81249cc59d8735d4/src/gmt_api.c#L12790) allows passing CLI arguments in three different modes:

1. Pass a single text string for CLI arguments `"-R0/10/0/10 -JX10c -BWSen+tMy Title"`
2. Pass a list of strings and each string is an option `["-R0/10/0/10", "-JX10c", "-BWSen+tMy Title"]`
3. Pass a list of `GMT_OPTION` structure

Currently, we use mode 1. As mentioned above, it has caused many trouble if an argument has whitespace and single/double quotation marks. `GMT_Call_Module`'s `mode` parameter is actually passed to the `n_args_in` parameter in another C API function [`GMT_Create_Options`](https://github.com/GenericMappingTools/gmt/blob/e5ef7016eaccce3fac58419f81249cc59d8735d4/src/gmt_parse.c#L542). Here is the description of the `GMT_Create_Options` function:
```
	/* This function will loop over the n_args_in supplied command line arguments (in) and
	 * then creates a linked list of GMT_OPTION structures for each program option.
	 * These will in turn will be processed by the module-specific option parsers.
	 * What actually happens is controlled by n_args_in.  There are three cases:
	 * n_args_in < 0 : This means that in is already a linked list and we just duplicate and return a copy.
	 * n_args_in == 0: in is a single text string with multiple options (e.g., "-R0/2/0/5 -Jx1 -O -K > file")
	 *		   and we must first break this command string into separate words.
	 * n_args_in > 0 : in is an array of text strings (e.g., argv[]).
	 *
	 * Note: 1. If argv[0] is the calling program name, make sure to pass argc-1, args+1 instead.
	 *	 2. in == NULL is allowed only for n_args_in <= 0 (an empty list of options).
	 *
	 * The linked list returned by GMT_Create_Options should be freed by GMT_Destroy_Options().
	 */
```
So, if mode 1 is used, the `GMT_Create_Options` function first splits the single text string into an array of strings (i.e., mode 2), and then converts the array of strings to a list of GMT_OPTION structure (i.e., mode 3). As can be seen in the `GMT_Create_Options` source codes, for mode 1, GMT tries very hard to deal with whitespace and quotation marks and we're repeating the same work in the `build_arg_string` function. Mode 1 is definitely the "hardest" mode for us and mode 3 means we have to wrap the GMT_OPTION structure, so why not just use mode 2!

### Changes in this PR

- [x] Let `Session.call_module` accept a list of strings (i.e., mode 2). Passing a single string is still supported for backward-compatibility #3139
- [x] Provide a new function `build_arg_list` function which does the same thing as `build_arg_string` but returns a list. The `build_arg_list` doesn't need any hacking solutions as in `build_arg_string`. #3149
- [ ] Replace `build_arg_string` with `build_arg_list` in all module wrappers
- [ ] Fix any edge cases that are not covered by the step above
- [ ] Update all tests that use the `Session.call_module` method
- [ ] Some tests may no longer make sense and should be removed
- [ ] Need to add more tests
- [ ] If `data=["file1.txt", "file2.txt"]`, `data_kind(data)` returns `matrix`, which cause failures. I guess we need to refactor the `data_kind` function.

**To Be Done/Decided**

- [ ] ~~Raise a deprecation warning when passing a single string to `Session.call_module` and fully remove the support in v0.20.0 or even v1.0?~~ Likely no, because sometimes passing a single string is convenient but we should avoid using it in the PyGMT source codes.
- [ ] We need to keep the `build_arg_string` function for some time because we know some users write their own wrappers and use the `build_arg_string` function. A deprecation warning makes sense.

### Benifits

- No hacking solutions needed
- Natively support for passing multiple files (e.g., `Figure.plot(data=["file1.txt", "file2.txt"])`

Supersede PRs #1487, #1116, #2331, #2726.

Also fixes issue #3104 without the solution in #3105, but I'll open some more discussions about quotation marks, so this PR won't close issue #3104.


weiji14:
Seriously? After all these years and countless workarounds, and there was a simpler way?!!

seisman:
> Seriously? After all these years and countless workarounds, and there was a simpler way?!!

:smile: Should have read the C source codes earlier.

pygmt/clib/session.py
```diff
@@ -618,11 +626,31 @@ def call_module(self, module, args):
...
+            # In this way, GMT knows that each list element is an option with arguments,
+            # so it won't incorrectly parse strings with whitespaces or quotation marks.
+            # This is the preferred way to pass arguments to GMT API, because we don't
+            # have to deal with whitespaces or quotation marks on the PyGMT side.
+            mode = len(args)  # Now mode is the number of arguments.
+            if mode == 0:  # No arguments, passing a null pointer
+                argv = None
+            else:
+                argv = (ctp.c_char_p * len(args))()
+                argv[:] = np.char.encode(args)
```
seisman, line 647: Refactor after PR #3137 is merged.
```suggestion
                argv = strings_to_ctypes_array(args)
```

seisman, line 647:
```suggestion
            # Pass a null pointer if no arguments are specified
            argv = strings_to_ctypes_array(args) if mode !=0 else None
```

seisman:
I've squooshed all the changes into 5 commits. In commit 1149b3719a86465308e79d6b8b0835e3dc524137, a new function `build_arg_list` is added and in the following commits, changes are made to all module wrappers to pass a list of arguments to the `Session.call_module` method.

To make the PR easier to review, a separate PR that contains commit 1149b3719a86465308e79d6b8b0835e3dc524137 is opened in #3149.

seisman:
After PR #3139 and #3149, now it's time to refactor module wrappers to pass a list of arguments when calling modules. More than 60 files are changed in this PR but in most files, the changes are simple (just replacing `build_arg_string` with `build_arg_list`). There are a few exceptions and please pay more attention to these files:

- `pygmt/figure.py`
- `pygmt/helpers/utils.py`
- `pygmt/src/config.py`
- `pygmt/src/legend.py`
- `pygmt/src/text.py`
- `pygmt/src/which.py`

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final review call. Will merge in 48 hours if no further comments.

pygmt/src/which.py
```diff
@@ -3,12 +3,11 @@
 """
 
 from pygmt.clib import Session
-from pygmt.helpers import build_arg_string, fmt_docstring, kwargs_to_strings, use_alias
+from pygmt.helpers import build_arg_list, fmt_docstring, is_nonstr_iter, use_alias
 
 
 @fmt_docstring
 @use_alias(G="download", V="verbose")
-@kwargs_to_strings(fname="sequence_space")
```
weiji14, line 11: Don't need to do it in this PR, but should we remove `sequence_space` from decorators.py at some point?

weiji14 (approved):
Looks good!

pygmt/src/which.py
seisman: Absolutely yes.

PR #3127, opened 2024-03-20
Wrap GMT's data structure GMT_GRID_HEADER for grid/image/cube headers

seisman:
**Description of proposed changes**

Part of PR #2398 to support wrapping `GMT_GRID`/`GMT_IMAGE`/`GMT_CUBE`.

`GMT_GRID_HEADRE` is used in `GMT_GRID`/`GMT_IMAGE`/`GMT_CUBE`, so better to have it merged into the main branch then we can work on `GMT_GRID`/`GMT_IMAGE` in parallel.

Proposed by https://github.com/GenericMappingTools/pygmt/pull/2398#pullrequestreview-1947793913

No tests are added.

PR #3134, opened 2024-03-22
GMT_GRID_HEADER: Parse grid header and add grid properties

seisman:

Part of PR #2398.

This PR add codes to parse the grid header and get the dimension names, attributes, gtype and more. They're stored as properties of the header.

The codes can be reused in wrapping GMT_GRID/GMT_IMAGE/GMT_CUBE data structures.

Tests are added to the `_parse_nameunits` function but not for other functions. The function will be tested in the `GMT_GRID.to_dataarray` method.


pygmt/datatypes/grid.py
```diff
@@ -101,5 +101,44 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
 
+def _parse_nameunits(nameunits: str) -> tuple[str, str | None]:
+    """
+    Get long_name and units attributes from x_units/y_units/z_units in grid header.
+
+    In GMT grid header, the x_units/y_units/z_units are strings in the form of
+    ``long_name [units]``, in which both ``long_name`` and ``units`` and standard
+    netCDF attributes defined by CF conventions. The ``[units]`` part is optional.
+
+    This function parses the x_units/y_units/z_units string and get the ``long_name``
```
michaelgrund, line 112:
```suggestion
    This function parses the x_units/y_units/z_units string and gets the ``long_name``
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +101,44 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
     ]
 
 
+def _parse_nameunits(nameunits: str) -> tuple[str, str | None]:
+    """
+    Get long_name and units attributes from x_units/y_units/z_units in grid header.
+
+    In GMT grid header, the x_units/y_units/z_units are strings in the form of
+    ``long_name [units]``, in which both ``long_name`` and ``units`` and standard
```
michaelgrund, line 109:
```suggestion
    ``long_name [units]``, in which both ``long_name`` and ``units`` are standard
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+    dims : tuple
+        The dimension names with the last dimension for the data variable.
+    attrs : dict
+        The attributes for each dimension.
+    registration : int
+        The grid registration. 0 for gridline and 1 for pixel.
+    gtype : int
+        The grid type. 0 for Cartesian grid and 1 for geographic grid.
+    """
+    # Default dimension names. The last dimension is for the data variable.
```
seisman, line 177: The default dimension names are "x"/"y"/"z" for 2-D grids and "x"/"y"/"z"/"cube" for 3-D cubes. Currently, the code only works for 2-D grids.

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+    ):
+        gtype = 1
+    # Registration
+    registration = header.registration
+
+    # Update the attributes dictionary with new dimension names as keys
+    attrs = {newdims[dim]: attrs[dim] for dim in dims}
+    # Update the dimension names
+    dims = tuple(newdims[dim] for dim in dims)
+    return dims, attrs, registration, gtype
```
seisman, line 226: I'm debating about the order of dimension names. Currently, the returned `dims` are `("x", "y", "z")`, but when creating a xarray.DataArray, we need to use `("y", "x", "z")` instead. Which one the `_parse_header` function should return?

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
     ]
 
 
+def _parse_nameunits(nameunits: str) -> tuple[str, str | None]:
+    """
+    Get long_name and units attributes from x_units/y_units/z_units in grid header.
+
+    In GMT grid header, the x_units/y_units/z_units are strings in the form of
+    ``long_name [units]``, in which both ``long_name`` and ``units`` are standard
+    netCDF attributes defined by CF conventions. The ``[units]`` part is optional.
```
yvonnefroehlich, line 112:
```suggestion
    In the GMT grid header, the x_units/y_units/z_units are strings in the form of
    ``long_name [units]``, in which both ``long_name`` and ``units`` are standard
    netCDF attributes defined by CF conventions. The ``[units]`` part is optional.
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+
+    This function parses the x_units/y_units/z_units string and gets the ``long_name``
+    and ``units`` attributes.
```
yvonnefroehlich, line 115:
```suggestion
    This function parses the x_units/y_units/z_units strings and gets the ``long_name``
    and ``units`` attributes.
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+
+    Parameters
+    ----------
+    nameunits
+        The x_units/y_units/z_units string in grid header.
```
yvonnefroehlich, line 120:
```suggestion
        The x_units/y_units/z_units strings in the grid header.
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
...
+
+    Returns
+    -------
+    (long_name, units)
+        Tuple of netCDF attributes 'long_name' and 'units'. 'units' may be None.
```
yvonnefroehlich, line 125:
```suggestion
    (long_name, units)
        Tuple of netCDF attributes ``long_name`` and ``units``. ``units`` may be ``None``.
```

pygmt/datatypes/grid.py
```diff
@@ -101,5 +103,128 @@ class _GMT_GRID_HEADER(ctp.Structure):  # noqa: N801
     ]
 
 
+def _parse_nameunits(nameunits: str) -> tuple[str, str | None]:
+    """
+    Get long_name and units attributes from x_units/y_units/z_units in grid header.
```
yvonnefroehlich, line 108:
```suggestion
    Get the long_name and units attributes from x_units/y_units/z_units in the grid header.
```

pygmt/datatypes/grid.py
seisman: I guess we have to try wrapping the GMT_CUBE data structure to know how to adapt the `_parse_header` method for 3-D grids. Will open a PR later.

pygmt/datatypes/grid.py
```diff
@@ -5,6 +5,8 @@
 import ctypes as ctp
 from typing import ClassVar
 
+import numpy as np
+
 # Constants for lengths of grid header variables.
 #
 # Note: Ideally we should be able to get these constants from the GMT shared library
```
seisman, line 12: @weiji14 I feel you were right. We should put the `_GMT_GRID_HEADER` and some common constants/functions into `pygmt/datatypes/header.py` instead. Should we do the move now?

pygmt/datatypes/grid.py
weiji14: For multi-band images, the convention is to use `("z", "y", "x")` (band/channel, height, width), see e.g. rioxarray's example at https://corteva.github.io/rioxarray/stable/examples/COG.html. But we would then need to transpose the dims here.

pygmt/datatypes/grid.py
weiji14: Yeah, move the `_GMT_GRID_HEADER` class wrapped in #3127 and the parsing functions here to `pygmt/datatypes/header.py`.

seisman: OK, done in ab001441e490e8d1f035d89bb08a0df4158fdb2f.

pygmt/datatypes/grid.py
weiji14: Oh wait, the `z` here refers to the raster pixel value (e.g. elevation), not the band. In that case, we can probably go with `("y", "x", "z")` order.

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,228 @@
...
+        # Units in y-direction, in the form "long_name [units]"
+        ("y_units", ctp.c_char * GMT_GRID_UNIT_LEN80),
+        # Grid value units, in the form "long_name [units]"
+        ("z_units", ctp.c_char * GMT_GRID_UNIT_LEN80),
+        # Name of data set
+        ("title", ctp.c_char * GMT_GRID_TITLE_LEN80),
+        # Name of generating command
+        ("command", ctp.c_char * GMT_GRID_COMMAND_LEN320),
+        # Comments for this data set
+        ("remark", ctp.c_char * GMT_GRID_REMARK_LEN160),
```
weiji14, line 63: Store these metadata in the `attrs` dict too?

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,228 @@
...
+    attrs : dict
+        The attributes for each dimension.
+    registration : int
+        The grid registration. 0 for gridline and 1 for pixel.
+    gtype : int
+        The grid type. 0 for Cartesian grid and 1 for geographic grid.
+    """
+    # Default dimension names. The last dimension is for the data variable.
+    dims: tuple = ("x", "y", "z")
+    nameunits = (header.x_units, header.y_units, header.z_units)
```
weiji14, line 181: Reorder to yxz as mentioned at https://github.com/GenericMappingTools/pygmt/pull/3134#discussion_r1535769435
```suggestion
    dims: tuple = ("y", "x", "z")
    nameunits = (header.y_units, header.x_units, header.z_units)
```

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,228 @@
...
+    long_name = parts[0].strip()
+    units = parts[1].strip("]").strip() if len(parts) > 1 else None
+    return long_name, units
+
+
+def _parse_header(header: _GMT_GRID_HEADER) -> tuple[tuple, dict, int, int]:
+    """
+    Get dimension names, attributes, grid registration and type from the grid header.
+
+    For a 2-D grid, the dimension names are set to "x", "y", and "z" by default. The
```
weiji14, line 151:
```suggestion
    For a 2-D grid, the dimension names are set to "y", "x", and "z" by default. The
```

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Parameters
+    ----------
+    header
+        The grid header structure.
+
+    Returns
+    -------
+    dims : tuple
+        The dimension names with the last dimension for the data variable.
```
weiji14, line 171:
```suggestion
        The dimension names, with the last dimension being the data variable.
```

seisman:
I'm wondering if we should move the `_parse_nameunits`/`_parse_header` functions as the `GMT_GRID_HEADER` class methods, so that we don't have to import these private functions when wrapping `GMT_GRID`/`GMT_IMAGE`.

weiji14:
> I'm wondering if we should move the `_parse_nameunits`/`_parse_header` functions as the `GMT_GRID_HEADER` class methods, so that we don't have to import these private functions when wrapping `GMT_GRID`/`GMT_IMAGE`.

That's not a bad idea. But maybe split the implementation of `_parse_header` into getter methods like `get_attrs()`, `get_gtype()` and so on?

seisman:
> But maybe split the implementation of `_parse_header` into getter methods like `get_attrs()`, `get_gtype()` and so on?

Sounds good, but the `gtype` is determined from the `x`/`y` axis attributes. We may need to parse the headers (e.g., the current `_parse_header` function), then store the returned `attrs`, `dims`, `gtype` and `registration` into a class attributes (e.g., `GMT_GRID_HEADER.nc`) and then provide the getter methods which return the requested attributes.

pygmt/datatypes/header.py
seisman: Done in dfbc2e5237ff53e6741be2905dcb43a3e3a5877f.

seisman:
> > I'm wondering if we should move the `_parse_nameunits`/`_parse_header` functions as the `GMT_GRID_HEADER` class methods, so that we don't have to import these private functions when wrapping `GMT_GRID`/`GMT_IMAGE`.
> 
> That's not a bad idea. But maybe split the implementation of `_parse_header` into getter methods like `get_attrs()`, `get_gtype()` and so on?

Done in 2bd353abae7e2cda601a5cf386ae18e57694b063.

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,278 @@
...
+    def get_registration(self) -> int:
+        """
+        Get the grid registration from the grid header.
+
+        Returns
+        -------
+        registration
+            The grid registration. 0 for gridline and 1 for pixel.
+        """
+        return self.registration
```
weiji14, line 278: We could just remove this and call `self.registration` no?

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,278 @@
...
+        Returns
+        -------
+        attrs
+            List of attributes for each dimension.
+        """
+        if not hasattr(self, "_nc"):
+            self._parse_dimensions()
+        return self._nc["attrs"]
+
+    def get_gtype(self) -> int:
```
weiji14, line 252: How about using the `@property` decorator, similar to https://github.com/GenericMappingTools/pygmt/blob/bbf909d9ea6e1cf48af3566e490b14d10de09785/pygmt/figure.py#L121-L129

This would become:

```suggestion
    @property
    def gtype(self) -> int:
```

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,278 @@
...
+    def get_name(self) -> str:
+        """
+        Get the name of the grid from the grid header.
+
+        Returns
+        -------
+        name
+            The name of the grid.
+        """
+        return "z"
```
weiji14, line 202: Can probably remove this if it is only returning the letter `z`?

pygmt/datatypes/header.py
seisman: I was thinking about using `@property` too. But when we use `header.gtype`, it's not straightforward to know if `gtype` is a field or a property of the `GMT_GRID_HEADER` class, although technically it makes no differences.

weiji14: Yeah, either way, we're just accessing the grid header's attributes. I'd prefer adding the `@property` decorator to make it clear that we are accessing a property/attribute, but up to you.

pygmt/datatypes/header.py
seisman: The name defaults to `z` for 2-D grids and `cube` for 3-D cubes, but it can be overridden by the `varname` variable stored in the hidden `GMT_GRID_HEADER_HIDDEN` data structure. I guess we won't wrap the `GMT_GRID_HEADER_HIDDEN` data structure now because it is designed to be *hidden* and may change anytime, but maybe we have to wrap it in the future?

weiji14: Right, I see that `varname` is marked as private at https://github.com/GenericMappingTools/gmt/blob/9b93e07982233a31e97b3f8a4ed8df3a3dfa56f9/src/gmt_hidden.h#L161, but it is supposedly holding the name of a NetCDF variable? Maybe we should try to parse it instead of always returning `z` (or `cube`)?

seisman: As I understand it, `varname` is only assigned when a netCDF name like `input.nc?varname` is specified. I guess it's just rare cases.

As mentioned above, to parse `varname`, we have to wrap the `GMT_GRID_HEADER_HIDDEN` data structure, which contains more than 50 fields, and the last time it was changed is just a few months ago (https://github.com/GenericMappingTools/gmt/commit/b9d3f2bdc272e5f0c08dc93f45c622285fad9604).

weiji14: Got it. Let's just leave it unwrapped for now then.

pygmt/datatypes/header.py
seisman: Removed in 79f4c879e1cc0af2ce3f8cbd9d5eaac6a5644ad6.

pygmt/datatypes/header.py
seisman: Change them to properties in 79f4c879e1cc0af2ce3f8cbd9d5eaac6a5644ad6.

pygmt/datatypes/header.py
```diff
@@ -0,0 +1,247 @@
...
+        # Default dimension names.
+        dims = ("y", "x")
+        nameunits = (self.y_units, self.x_units)
+
+        # Dictionary for dimension attributes with the dimension name as the key.
+        attrs = {dim: {} for dim in dims}
+        # Dictionary for mapping the default dimension names to the actual names.
+        newdims = {dim: dim for dim in dims}
+        # Loop over dimensions and get the dimension name and attributes from header
+        for dim, nameunit in zip(dims, nameunits, strict=False):
```
weiji14, line 165: Should zip with `strict=True`
```suggestion
        for dim, nameunit in zip(dims, nameunits, strict=True):
```

weiji14 (approved):
Looks good. There's lots of code coverage missing, but that will be covered in #2398 and/or #3128. Might make small changes to the header parsing as we continue on wrapping `GMT_GRID`, `GMT_IMAGE` and `GMT_CUBE`.

PR #3136, opened 2024-03-23
Add sequence_to_ctypes_array to convert a sequence to a ctypes array

seisman:

The [`kwargs_to_ctypes_array`](https://github.com/GenericMappingTools/pygmt/blob/32e3cb36a0a2faaebd837ab99986431ee57e8060/pygmt/clib/conversion.py#L246) function was one of the oldest functions in PyGMT. It was first added in PR #55 (It's 2017! Seven years ago!) and has not changed since then.

The function converts a sequence (like a list or a tuple) into a ctypes array, but the sequence must be the argument of a keyword dictionary, so it's called like this:
```
dim = kwargs_to_ctypes_array("dim", kwargs, ctp.c_uint64 * 4)
```
but why not just have a more general function `sequence_to_ctypes_array`, which can be called like:
```
dim = sequence_to_ctypes_array(kwargs.get("dim"), ctp.c_uint64 * 4)
```
The new function also works if `dim` is not a keyword argument:
```
dim = sequence_to_ctypes_array(dim, ctp.c_uint64 * 4)
```

This PR removes the old `kwargs_to_ctypes_array` function and adds the new `sequence_to_ctypes_array` function. It's defined differently and should be called like:
```
dim = sequence_to_ctypes_array(dim, ctp.c_uint64, 4) 
```
which I think is more readable. 

The `Session.create_data` method is also slightly refactored because the `kwargs` parameter makes little sense in this method.

PR #3137 adds a similar function that converts a sequence of strings into a ctypes array. Better to review these two PRs back-to-back.

### DIFF
--- a/pygmt/clib/conversion.py
+++ b/pygmt/clib/conversion.py
@@ -2,7 +2,9 @@
 Functions to convert data types into ctypes friendly formats.
 """
 
+import ctypes as ctp
 import warnings
+from collections.abc import Sequence
 
 import numpy as np
 from pygmt.exceptions import GMTInvalidInput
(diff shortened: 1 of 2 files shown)


michaelgrund (approved):

pygmt/clib/conversion.py
```diff
@@ -243,41 +245,55 @@ def as_c_contiguous(array):
...
     ----------
-    argument : str
-        The name of the argument.
-    kwargs : dict
-        Dictionary of keyword arguments.
-    dtype : ctypes type
-        The ctypes array type (e.g., ``ctypes.c_double*4``)
+    sequence
+        The sequence to convert. If None, returns None. Otherwise, it must be a sequence
+        (e.g., list, tuple, numpy array).
```
yvonnefroehlich, line 260:
```suggestion
    sequence
        The sequence to convert. If ``None``, returns ``None``. Otherwise, it must be a sequence
        (e.g., list, tuple, numpy array).
```

Issue #886, opened 2021-02-14
Parameters vs Arguments?


seisman:
For function parameters like `frame`, `region`, we're inconsistently calling them "parameters", "arguments",  or "options".

I thought the correct way is calling `frame` an "argument", but after reading this FAQ (https://docs.python.org/3.9/faq/programming.html#faq-argument-vs-parameter), I think we should call `frame` a "parameter". 

Thoughts? @GenericMappingTools/python @GenericMappingTools/python-maintainers

weiji14:
A quick CTRL+F yielded >100 instances of 'argument' in the PyGMT project currently. Some are correctly called 'argument', but not all of them. It will be a big task to replace all of them, so maybe just be careful going forward and replace things on an ad-hoc basis (e.g. when refactoring stuff). I think most people won't get too hung up on whether we call it 'parameter' or 'argument', they know that it is 'something' handled by the program.

As for 'option', I think it's fine to use as a synonym for 'parameter'.

seisman:
> It will be a big task to replace all of them, so maybe just be careful going forward and replace things on an ad-hoc basis (e.g. when refactoring stuff).

Yes, I agree. It would be better to document it in the contributing guides.

meghanrjones:
In #881, I found that the another common inconsistency involved referencing to the arguments of parameters as options. Based on this discussion, it seems that these should be only referred to as 'arguments' if 'option' is a synonym for 'parameter'.

seisman:
Almost done. Closing the issue.

PR #3149, opened 2024-03-31
Add function build_arg_list for building arguments list from keyword dictionaries

seisman:
This PR contains the changes of commit 1149b3719a86465308e79d6b8b0835e3dc524137 in PR #3132.

In this PR, a new utility function `build_arg_list` function is added. The new function is modified from the `build_arg_string` function and it differs from the `build_arg_string` function in the following aspects:

1. returns a list of arguments, instead of a single argument string
2. no hacky solutions for handling whitespaces
3. Only apply the `non_ascii_to_octal` function to the single-letter option arguments, not to the configuration arguments and input/output files
4. Doctests are also simplified.

The goal is to replace `build_arg_string` with `build_arg_list` in all module wrappers and deprecate/remove `build_arg_string` in future releases.

PR #3132 contains more changes that we will apply after this PR is merged.


pygmt/helpers/utils.py
```diff
@@ -315,6 +315,108 @@ def non_ascii_to_octal(argstr):
...
+    >>> print(build_arg_list(dict(R="1/2/3/4", J="X4i", watre=True)))
+    Traceback (most recent call last):
+      ...
+    pygmt.exceptions.GMTInvalidInput: Unrecognized parameter 'watre'.
+    """
+    gmt_args = []
+    for key, value in kwdict.items():
+        if len(key) > 2:  # Raise an exception for unrecognized options
+            raise GMTInvalidInput(f"Unrecognized parameter '{key}'.")
+        if value is None or value is False:  # Exclude arguments that are None and False
```
yvonnefroehlich, line 397:
```suggestion
        if value is None or value is False:  # Exclude arguments that are None or False
```

pygmt/helpers/utils.py
```diff
@@ -315,6 +315,108 @@ def non_ascii_to_octal(argstr):
...
+) -> list[str]:
+    r"""
+    Convert keyword dictionaries and input/output files into a list of GMT arguments.
+
+    Make sure all values in ``kwdict`` have been previously converted to a string
+    representation using the ``kwargs_to_strings`` decorator. The only exceptions are
+    ``True``, ``False`` and ``None``.
+
+    Any lists or tuples left will be interpreted as multiple entries for the same
+    command line option. For example, the kwargs entry ``"B": ["xa", "yaf"]`` will be
```
weiji14, line 332: Call it 'command-line option' or 'parameter'? Xref https://github.com/GenericMappingTools/pygmt/issues/886.

```suggestion
    Any remaining lists or tuples will be interpreted as multiple entries for the same
    parameter. For example, the kwargs entry ``"B": ["xa", "yaf"]`` will be
```

pygmt/helpers/utils.py
```diff
@@ -315,6 +315,108 @@ def non_ascii_to_octal(argstr):
     return argstr.translate(str.maketrans(mapping))
 
 
+def build_arg_list(
+    kwdict: dict,
+    confdict: dict | None = None,
+    infile: str | pathlib.PurePath | list[str | pathlib.PurePath] | None = None,
```
weiji14, line 321: Either a list or tuple can be passed in, so maybe use [`Sequence`](https://docs.python.org/3/glossary.html#term-sequence)? Or `list | tuple`?
```suggestion
    infile: str | pathlib.PurePath | Sequence[str | pathlib.PurePath] | None = None,
```

pygmt/helpers/utils.py
```diff
@@ -315,6 +315,108 @@ def non_ascii_to_octal(argstr):
     return argstr.translate(str.maketrans(mapping))
 
 
+def build_arg_list(
+    kwdict: dict,
```
weiji14, line 319: Mention what's in the `dict`?
```suggestion
    kwdict: dict[str, Any],
```

pygmt/helpers/utils.py
```diff
@@ -315,6 +315,108 @@ def non_ascii_to_octal(argstr):
...
+    confdict: dict | None = None,
```
weiji14, line 320:
```suggestion
    confdict: dict[str, str] | None = None,
```

weiji14 (commented):
Code logic ok, just some suggestions related to the type hints.

weiji14 (approved):

PR #3163, opened 2024-04-10
CI: Use OIDC token for codecov uploading

seisman:
Codecov-action v4.2.0 started to support OIDC token for uploading coverage results, so that we don't need to use secret tokens any more.

Upstream PR: https://github.com/codecov/codecov-action/pull/1330
Usage: https://github.com/codecov/codecov-action#using-oidc

If it works, then we can remove the `CODECOV_TOKEN` secrets from the settings.


seisman:
After waiting for some upstream fixes (xref https://github.com/codecov/codecov-action/issues/1378), now OIDC token works.

weiji14 (approved):

PR #3165, opened 2024-04-10
Switch to official GitHub action for managing app tokens

### LINKED ISSUE #3164: Switch to official GitHub action for managing app tokens
Currently, we're using the [`tibdex/github-app-token` action](https://github.com/tibdex/github-app-token) to generate PAT tokens which are used in the `/format` slash command (initially proposed in https://github.com/GenericMappingTools/pygmt/pull/645#discussion_r502327175).

It seems that GitHub now provides an official action for this purpose (https://github.com/actions/create-github-app-token). I think we should try to switch to the official one. Looking at this PR (https://github.com/dependabot/dependabot-core/pull/9340), it seems we just need to update the action name and the action input names.

xref: 

- https://github.com/tibdex/github-app-token/issues/99
- https://github.com/actions/create-github-app-token

yvonnefroehlich:

Switch from the [tibdex/github-app-token action](https://github.com/tibdex/github-app-token) to generate PAT token which are used in the /format slash command to the official GitHub action for this purpose (https://github.com/actions/create-github-app-token).

Fixes #3164

### DIFF
--- a/.github/workflows/format-command.yml
+++ b/.github/workflows/format-command.yml
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Generate token from GenericMappingTools bot
-      - uses: tibdex/github-app-token@v2
+      - uses: actions/create-github-app-token@v1.9
         id: generate-token
         with:
-          app_id: ${{ secrets.APP_ID }}
-          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
 
       # Checkout the pull request branch
       - uses: actions/checkout@v4.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/format-command.yml
```diff
@@ -11,11 +11,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Generate token from GenericMappingTools bot
-      - uses: tibdex/github-app-token@v2
+      - uses: actions/create-github-app-token@v1.9
```
seisman, line 14:
```suggestion
      - uses: actions/create-github-app-token@v1.9.3
```

seisman:
The `/slash` command can only trigger the workflow in the main branch, so there is no easy way to test if it works in PRs or forks. To test it, you need to push the same changes into the main branch in your own fork, open a PR to test it, and also need to ask someone else to submit a PR from a fork to test it.

The whole process is tedious (I have done it several times), so I think we should just merge this PR in and if anything doesn't work as expected, we should either fix it or revert it.

seisman (approved):

Issue #3170, opened 2024-04-15
pygmt.which: Errors if downloading multiple tiled grids


seisman:
The issue was originally reported in https://github.com/GenericMappingTools/pygmt/pull/3148#issuecomment-2054814923.

Here is a minimal example to reproduce the issue. 
```
!rm -r ~/.gmt/server/earth/earth_relief/earth_relief_15s_p  # Make sure files are not downloaded
from pygmt import which
which(fname=["@N30W120.earth_relief_15s_p.nc", "@N00E000.earth_relief_15s_p.nc"], download="a")
```

The errors are:
```
File ~/OSS/gmt/pygmt/pygmt/src/which.py:67, in which(fname, **kwargs)
     62     with lib.virtualfile_out(kind="dataset") as vouttbl:
     63         lib.call_module(
     64             module="which",
     65             args=build_arg_string(kwargs, infile=fname, outfile=vouttbl),
     66         )
---> 67         paths = lib.virtualfile_to_dataset(vfname=vouttbl, output_type="strings")
     69 match paths.size:
     70     case 0:

File ~/OSS/gmt/pygmt/pygmt/clib/session.py:1940, in Session.virtualfile_to_dataset(self, vfname, output_type, column_names, dtype, index_col)
   1937 result = self.read_virtualfile(vfname, kind="dataset").contents
   1939 if output_type == "strings":  # strings output
-> 1940     return result.to_strings()
   1942 result = result.to_dataframe(
   1943     column_names=column_names, dtype=dtype, index_col=index_col
   1944 )
   1945 if output_type == "numpy":  # numpy.ndarray output

File ~/OSS/gmt/pygmt/pygmt/datatypes/dataset.py:156, in _GMT_DATASET.to_strings(self)
    154         if segment.contents.text:
    155             textvector.extend(segment.contents.text[: segment.contents.n_rows])
--> 156 return np.char.decode(textvector) if textvector else np.array([], dtype=str)

File ~/opt/miniconda/envs/pygmt/lib/python3.12/site-packages/numpy/core/defchararray.py:615, in decode(a, encoding, errors)
    572 @array_function_dispatch(_code_dispatcher)
    573 def decode(a, encoding=None, errors=None):
    574     r"""
    575     Calls ``bytes.decode`` element-wise.
    576
   (...)
    612
    613     """
    614     return _to_bytes_or_str_array(
--> 615         _vec_string(a, object_, 'decode', _clean_args(encoding, errors)))

[...]

PR #3171, opened 2024-04-15
Reorganize the list of data files for caching

### LINKED ISSUE #3144: Reorganize the list of cache files used in CI
Many remote files are cached as GitHub Actions assets to be reused in CI workflows. These files are listed in the `pygmt/helpers/caching.py` file. See below:

https://github.com/GenericMappingTools/pygmt/blob/f0d4d21f7518b4dc10f503ec26e2e5b603d8891e/pygmt/helpers/caching.py#L13-L20

Currently, the list is a little messy. Some potential improvements:

1. Put some comments (e.g., "Names like @N35E135.earth_relief_03s_g.nc are for internal use only.") at the start of the list. 
2. Remove some comments like "# Specific grid for 15s test"
3. Update the order of some entries. First categorize files into groups and sort alphabetically in each group.

Helps are welcome.

seisman:
~~Due to a likely upstream bug, we can't request multiple tiled grids in a single `which` call (xref: #3170). We have to call `which` multiple times and download a single tiled grid in each call, to make the "Cache data" workflow pass again.~~

~~I also take this opportunity to address the request in #3144.~~

Closes #3144.

Edit: With the workaround in PR #3174, now it's OK to download multiple tiled grids. So, this PR only reorganize the list of data files for caching to make it easy to maintain.


pygmt/helpers/caching.py
```diff
@@ -100,4 +89,5 @@ def cache_data():
         "@tut_ship.xyz",
         "@usgs_quakes_22.txt",
     ]
-    which(fname=datasets, download="a")
+
+    which(fname=datasets + caches + tiles, download="a")
```
weiji14, line 93: Can we continue to use just a single list?

seisman: We can. I just feel that having three separate lists makes it easier to organize the file list.

pygmt/helpers/caching.py
```diff
@@ -9,89 +9,78 @@ def cache_data():
...
+        "@earth_age_01d_g",
+        "@earth_day_01d_p",
+        "@earth_faa_01d_g",
         "@earth_gebco_01d_g",
         "@earth_gebcosi_01d_g",
         "@earth_gebcosi_15m_p",
-        "@earth_relief_01d_p",
+        "@earth_geoid_01d_g",
+        "@earth_mag4km_01d_g",
+        "@earth_mag_01d_g",
```
weiji14, line 22:
```suggestion
        "@earth_mag_01d_g",
        "@earth_mag4km_01d_g",
```

pygmt/helpers/caching.py
weiji14: Well, codecov is warning that there's missing code coverage for the extra two lists :slightly_smiling_face:

pygmt/helpers/caching.py
```diff
@@ -9,89 +9,78 @@ def cache_data():
...
+        "@N00W090.earth_relief_03m_p.nc",
+        "@N00E135.earth_relief_30s_g.nc",
+        "@N00W010.earth_relief_15s_p.nc",
+        "@N30W120.earth_relief_15s_p.nc",
+        "@N04W010.earth_relief_03s_g.nc",
+        "@N35E135.earth_relief_03s_g.nc",
+        "@N37W120.earth_relief_03s_g.nc",
+        "@S15W105.earth_synbath_30s_p.nc",
+        "@S90E000.earth_wdmam_03m_g.nc",
+        "@N00W030.earth_vgg_01m_p.nc",
```
weiji14, line 59:
```suggestion
        "@N00W030.earth_vgg_01m_p.nc",
        "@S90E000.earth_wdmam_03m_g.nc",
```

pygmt/helpers/caching.py
seisman: That's a good point to push me to merge them into a single list.

seisman: Done in 72fcaa1b18ddb922005c0dc482788d0e439098cc.

weiji14 (approved):

PR #3150, opened 2024-03-31
Wrap GMT's standard data type GMT_CUBE for cubes and support reading/writing cubes

seisman:
Similar to #3338 and #3128, but for wrapping `GMT_CUBE`. 

Main changes in this PR:

- Wrap GMT's standard data type `GMT_CUBE`
- Implement the `to_xarray` method
- Let low-level functions (e.g., `read_data`/`write_data`) support cubes
- Extend the xarray backend to support cubes (i.e., `xr.open_dataarray("xxx.cube", engine="gmt", kind="cube")`

Known issues:

- It seems we can't use `lib.call_module("read", "@cube.nc out.nc -Tu")` for cubes. The similar command works well for grids and images. It's likely an upstream bug.

PR #3181, opened 2024-04-19
Add placeholder codes for wrapping GMT_IMAGE/GMT_CUBE

seisman:
I don't think we can finish wrapping `GMT_IMAGE` and `GMT_CUBE` in the next few weeks, so it is better to push them to PyGMT v0.13.0.

This PR cherry-pick codes used in wrapping `GMT_IMAGE`/`GMT_CUBE` (#3150 and #3128) to minimize potential conflicts in #3150 and #3128.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -25,7 +25,7 @@ from pygmt.clib.conversion import (
     vectors_to_arrays,
 )
 from pygmt.clib.loading import load_libgmt
-from pygmt.datatypes import _GMT_DATASET, _GMT_GRID
+from pygmt.datatypes import _GMT_CUBE, _GMT_DATASET, _GMT_GRID, _GMT_IMAGE
 from pygmt.exceptions import (
     GMTCLibError,
     GMTCLibNoSessionError,
@@ -1649,7 +1649,9 @@ class Session:
 
     @contextlib.contextmanager
     def virtualfile_out(
-        self, kind: Literal["dataset", "grid"] = "dataset", fname: str | None = None
+        self,
+        kind: Literal["dataset", "grid", "image", "cube"] = "dataset",
+        fname: str | None = None,
     ):
         r"""
         Create a virtual file or an actual file for storing output data.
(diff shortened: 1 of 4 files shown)


pygmt/clib/session.py
```diff
@@ -1801,9 +1806,12 @@ def read_virtualfile(
...
             return pointer
-        if kind in ["image", "cube"]:
-            raise NotImplementedError(f"kind={kind} is not supported yet.")
-        dtype = {"dataset": _GMT_DATASET, "grid": _GMT_GRID}[kind]
+        dtype = {
+            "dataset": _GMT_DATASET,
+            "grid": _GMT_GRID,
+            "image": _GMT_IMAGE,
+            "cube": _GMT_CUBE,
+        }[kind]
```
weiji14, line 1814: If we merge this PR for PyGMT v0.12.0, then `NotImplementedError` won't be raised right?

seisman: Yes, but since it's a low-level API function, I guess users won't call it anyway and we maintainers should know what we're doing.

weiji14: But `kind='image'` and `king='cube'` is showing up in the API docs at https://pygmt-dev--3181.org.readthedocs.build/en/3181/api/generated/pygmt.clib.Session.virtualfile_out.html. Can we push this PR to v0.13.0? There's no real benefit to users in having this merged anyway without `GMT_IMAGE`/`GMT_CUBE` being wrapped right?

weiji14: Oh wait, that was the docs for `virtualfile_out`, I meant https://pygmt-dev--3181.org.readthedocs.build/en/3181/api/generated/pygmt.clib.Session.read_virtualfile.html

seisman: I guess we can close this PR, since it's likely we will finish wrapping `GMT_IMAGE` first and then `GMT_CUBE` later, so we only need to deal with the conflicts in PR #3150.

Issue #3160, opened 2024-04-08
x2sys_cross: Refactor to get rid of temporary files and have consistent table-like output behavior


seisman:
_Originally posted by @weiji14 in https://github.com/GenericMappingTools/pygmt/issues/2730#issuecomment-2001898646_

I had a look at refactoring `x2sys_cross` to use virtualfiles instead of temporary files, but it's a little tricky because:

- Input: Cannot pass in virtualfiles as input as mentioned at https://github.com/GenericMappingTools/pygmt/pull/546#discussion_r459366392 and https://github.com/GenericMappingTools/gmt/issues/3717, since GMT doesn't support virtualfiles to X2SYS modules
- Output: The `virtualfile_to_dataset` method from #3083 was able to produce a `pandas.DataFrame` output, **but** the column names were missing. The logic for handling `x2sys_cross`'s output is actually complicated:

    https://github.com/GenericMappingTools/pygmt/blob/bcbbcad1ffca6f73fcc8ebc2872f228aaaa2266e/pygmt/src/x2sys_cross.py#L231-L250

Important things to handle are:
1. Datetime columns need to be parsed correctly as `datetime64` dtype
2. `x2sys_cross` may output multi-segment parts (see https://docs.generic-mapping-tools.org/6.5/supplements/x2sys/x2sys_cross.html#remarks) when multiple tracks are passed in and `-Qe` (external COEs) is selected. Unsure how this is handled in GMT virtualfiles (note that we actually just merge all the multi-segments into one table when pandas.DataFrame output is selected, output to file will preserve the segments though).
3. Last two column names can either be `z_X`/`z_M` or `z_1`/`z_2` depending on whether trackvalues/`-Z` argument is set.

It should be possible to handle 1 and 3 somehow, but I'm not so sure about 2 since it will involve checking how GMT outputs virtualfiles in `x2sys_cross`. We'll need to do some careful checking to ensure the refactoring doesn't modify the output and makes it incorrect.

seisman:
> * Input: Cannot pass in virtualfiles as input as mentioned at [Wrap x2sys_init and x2sys_cross #546 (comment)](https://github.com/GenericMappingTools/pygmt/pull/546#discussion_r459366392) and [Passing in virtual files into the supplementary x2sys modules gmt#3717](https://github.com/GenericMappingTools/gmt/issues/3717), since GMT doesn't support virtualfiles to X2SYS modules

[...]

PR #3182, opened 2024-04-19
pygmt.x2sys_cross: Refactor to use virtualfiles for output tables [BREAKING CHANGE: Dummy times in 3rd and 4th columns now have np.timedelta64 type]

seisman:

This PR refactors the `pygmt.x2sys_cross` function to use virtualfiles for output. Need to note that `x2sys_cross` still uses temporary files in the `tempfile_from_dftrack` function.

Partially address #3160.

This PR introduces a breaking change: Previously, the dummy times in 3-4 columns (with column names `i_1`/`i_2`) were in `np.object` type, and now they have `np.timedelta64` type.


pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +185,56 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                # Remove the "# " from "# x" in the first column
-                table = table.rename(columns={table.columns[0]: table.columns[0][2:]})
-            elif outfile != tmpfile.name:  # if outfile is set, output in outfile only
-                table = None
 
-    return table
+                # Convert 3rd and 4th columns to datetimes.
+                # These two columns have names "t_1"/"t_2" or "i_1"/"i_2".
+                # "t_1"/"t_2" means they are datetimes and should be converted.
+                # "i_1"/"i_2" means they are dummy times (i.e., floating-point values).
```
seisman, line 235: Am I understanding the output correctly?

seisman: I've never used x2sys, but here is my understanding of the C codes and the output:

1. The 3rd and 4th columns are datetimes. They can be either absolute datetimes (e.g., `2023-01-01T01:23:45.678` or dummy datetimes (i.e., double-precision numbers), depending on whether the input tracks contain datetimes.
2. Internally, absolute datetimes are also represented as double-precision numbers in GMT. So absolute datetimes and dummy datetimes are the same internally.
3. When outputting to a file, GMT will convert double-precision numbers into absolute datetimes, since GMT know if the column has dummy datetimes or not.
5. A `GMT_DATASET` container can only contain double-precision numbers and text strings. So when outputting to a virtual file, the 3rd and 4th columns always have double-precision numbers. If the column names are `t_1`/`t_2`, then we know they're absolute datetimes and should be converted; otherwise, they are just dummy datetimes and should not be converted.

weiji14: I'm a little unsure if `i_1`/`i_2` are actually dummy datetimes. This is a sample output from `x2sys_cross`:

```
# Tag: X2SYS4ivlhlo4
# Command: x2sys_cross @tut_ship.xyz -Qi -TX2SYS4ivlhlo4 ->/tmp/lala.txt
# x	y	i_1	i_2	dist_1	dist_2	head_1	head_2	vel_1	vel_2	z_X	z_M
> @tut_ship 0 @tut_ship 0 NaN/NaN/1357.17 NaN/NaN/1357.17
251.004840022	20.000079064	18053.5647431	13446.6562433	333.339586673	229.636557499	269.996783034	270.023614846	NaN	NaN	192.232797243	-2957.22757183
251.004840022	20.000079064	18053.5647431	71783.6562433	333.339586673	1148.20975878	269.996783034	270.023614846	NaN	NaN	192.232797243	-2957.22757183
250.534946327	20.0000526811	18053.3762934	66989.0210846	332.869692978	1022.68273972	269.996783034	269.360150109	NaN	NaN	-57.6485957585	-2686.4268008
250.532033147	20.0000525175	18053.3751251	66988.9936489	332.866779797	1022.67977813	269.996783034	22.0133296951	NaN	NaN	-64.5973890802	-2682.04812157
252.068705	20.000075	13447.5	71784.5	230.700422496	1149.27362378	269.995072235	269.995072235	NaN	NaN	0	-3206.5
```

It seems like the `i_1`/`i_2` values vary between rows, but I can't quite remember what they represent... maybe an index of some sort? I might need to inspect the C code to see what's going on, can you point me to where these `i_1`/`i_2` columns are being output?

weiji14 (commented):
There are 2 main changes happening in this PR:

1. Adding the `output_type="numpy"` option
2. Handling the different dtypes of the `i_1`/`i_2` or `t_1`/`t_2` columns

We can keep this as a single PR since it's hard to separate the two things, but might need to discuss the implementation a bit more.

pygmt/src/x2sys_cross.py
seisman: Dummy times are just double-precision indexes from 0 to *n* (xref: https://github.com/GenericMappingTools/gmt/blob/b56be20bee0b8de22a682fdcd458f9b9eeb76f64/src/x2sys/x2sys.c#L533).

The column name `i_1` or `t_1` is controlled by the variable `t_or_i` in the C code (https://github.com/GenericMappingTools/gmt/blob/b56be20bee0b8de22a682fdcd458f9b9eeb76f64/src/x2sys/x2sys_cross.c#L998). From https://github.com/GenericMappingTools/gmt/blob/b56be20bee0b8de22a682fdcd458f9b9eeb76f64/src/x2sys/x2sys_cross.c#L568, it's clear that, if `got_time` is True, then the column is absolute time (`GMT_IS_ABSTIME`), otherwise it's double-precision numbers (`GMT_IS_FLOAT`). 

We can keep the dummy times as double-precision numbers or think them as seconds since  unix epoch and then convert them to absolute times.

weiji14: > We can keep the dummy times as double-precision numbers or think them as seconds since unix epoch and then convert them to absolute times.

Maybe convert the relative time to `numpy.timedelta64`? Xref https://github.com/GenericMappingTools/pygmt/issues/2848.

seisman: Sounds good. Done in 9d12ae177.

pygmt/src/x2sys_cross.py
```diff
@@ -71,7 +71,12 @@ def tempfile_from_dftrack(track, suffix):
     Z="trackvalues",
 )
 @kwargs_to_strings(R="sequence")
-def x2sys_cross(tracks=None, outfile=None, **kwargs):
+def x2sys_cross(
+    tracks=None,
+    output_type: Literal["pandas", "numpy", "file"] = "pandas",
```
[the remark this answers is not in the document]
seisman: > If there are datetime values in the columns, they will get converted to floating point (?)

You're right. Datetimes are converted to floating points by `df.to_numpy()`. Will remove the `numpy` output type.

pygmt/src/x2sys_cross.py
```diff
@@ -71,7 +71,12 @@ def tempfile_from_dftrack(track, suffix):
     Z="trackvalues",
 )
 @kwargs_to_strings(R="sequence")
-def x2sys_cross(tracks=None, outfile=None, **kwargs):
+def x2sys_cross(
+    tracks=None,
+    output_type: Literal["pandas", "file"] = "pandas",
```
weiji14, line 76: Since the only two options are `pandas` or `file`, we probably don't need an `output_type` parameter, and can revert to the previous code where a `pandas.DataFrame` output is returned when `outfile` is not set?

seisman: Done in 5f045069c76fa316d550bcf7e0cea75149045e5b.

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +182,56 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                table = table.rename(columns={table.columns[0]: table.columns[0][2:]})
-            elif outfile != tmpfile.name:  # if outfile is set, output in outfile only
-                table = None
 
-    return table
+                # Convert 3rd and 4th columns to datetime or timedelta.
+                # These two columns have names "t_1"/"t_2" or "i_1"/"i_2".
+                # "t_1"/"t_2" means they are absolute datetimes.
+                # "i_1"/"i_2" means they are dummy times relative to unix epoch.
+                if output_type == "pandas":
```
seisman, line 231: More checks and tests are needed because the conversion rely on GMT configurations like `TIME_EPOCH` and `TIME_UNIT`.

pygmt/src/x2sys_cross.py
seisman: Perhaps we should revert the changes in https://github.com/GenericMappingTools/pygmt/pull/3191 since it's not needed by this PR.

weiji14: Hmm, you're right. Sorry, should have realized that earlier :sweat_smile:

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +182,71 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                    Path(os.environ["X2SYS_HOME"], kwargs["T"], f"{kwargs['T']}.tag")
-                    .read_text(encoding="utf8")
-                    .strip()
-                    .split("\n")[-1]
-                )  # e.g. "-Dxyz -Etsv -I1/1"
+                tagfile = Path(
+                    os.environ["X2SYS_HOME"], kwargs["T"], f"{kwargs['T']}.tag"
+                )
+                # Last line is like "-Dxyz -Etsv -I1/1"
+                lastline = tagfile.read_text().splitlines()[-1]
```
weiji14, line 205: Is `encoding="utf8"` not needed anymore?
```suggestion
                lastline = tagfile.read_text(encoding="utf8").splitlines()[-1]
```

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +182,71 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                table = pd.read_csv(
-                    tmpfile.name,
-                    sep="\t",
-                    header=2,  # Column names are on 2nd row
-                    comment=">",  # Skip the 3rd row with a ">"
-                    parse_dates=[2, 3],  # Datetimes on 3rd and 4th column
-                    **date_format_kwarg,  # Parse dates in ISO8601 format on pandas>=2
+                result = lib.virtualfile_to_dataset(
+                    vfname=vouttbl, output_type=output_type, header=2
                 )
```
weiji14, line 225: Note that `x2sys_cross` can output multi-segment files (each segment is separated by `IO_SEGMENT_MARKER` which is `>` by default, see https://docs.generic-mapping-tools.org/6.5/reference/file-formats.html#optional-segment-header-records). If I'm not mistaken, the current `virtualfile_to_dataset` method does not implement multi-segment file handling yet? To be fair though, the current implementation in `x2sys_cross` simply merges all segments into one, since we skip rows starting with `>`, but we need to check that `virtualfile_to_dataset` will return all segments in a multi-segment file instead of just the first one.

seisman: > If I'm not mistaken, the current `virtualfile_to_dataset` method does not implement multi-segment file handling yet? To be fair though, the current implementation in `x2sys_cross` simply merges all segments into one, since we skip rows starting with `>`, but we need to check that `virtualfile_to_dataset` will return all segments in a multi-segment file instead of just the first one.

Yes. The main problem is that, as far as I know, there is no equivalent way to represent a multi-segment file in pandas. The multi-segment support was also mentioned in https://github.com/GenericMappingTools/pygmt/pull/2729#discussion_r1378375843. 

If we can have a general way to represent multi-segment in pandas, then it should be straightforward to output multi-segments from `_GMT_DATASET` to the desired data structure.

seisman: > To be fair though, the current implementation in `x2sys_cross` simply merges all segments into one, since we skip rows starting with `>`, but we need to check that `virtualfile_to_dataset` will return all segments in a multi-segment file instead of just the first one.

It's already tested in the `_GMT_DATASET` docstrings https://github.com/GenericMappingTools/pygmt/blob/466c8b6b45d7d6aa202077d91764bd2ab8ed7324/pygmt/datatypes/dataset.py#L215. 

For `x2sys_cross`, we also test the shape of the output pd.DataFrame.

pygmt/src/x2sys_cross.py
seisman: Added back.

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +180,75 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                table = None
 
-    return table
+            if output_type == "file":
+                return result
+
+            # Convert 3rd and 4th columns to datetime/timedelta for pandas output.
+            # These two columns have names "t_1"/"t_2" or "i_1"/"i_2".
+            # "t_1"/"t_2" means they are absolute datetimes.
+            # "i_1"/"i_2" means they are dummy times relative to unix epoch.
```
seisman, line 231:
```suggestion
            # "i_1"/"i_2" means they are dummy times.
```

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +180,75 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
-                table = None
 
-    return table
+            if output_type == "file":
+                return result
+
+            # Convert 3rd and 4th columns to datetime/timedelta for pandas output.
+            # These two columns have names "t_1"/"t_2" or "i_1"/"i_2".
+            # "t_1"/"t_2" means they are absolute datetimes.
+            # "i_1"/"i_2" means they are dummy times.
```
seisman, line 231:
```suggestion
            # "t_" means absolute datetimes and "i_" means dummy times.
```

pygmt/src/x2sys_cross.py
```diff
@@ -183,68 +180,74 @@ def x2sys_cross(tracks=None, outfile=None, **kwargs):
...
+                    unit = "s"
+                    scale = 365.2425 / 12.0 * 86400.0
+                case _:
+                    unit = time_unit.upper() if time_unit in "wd" else time_unit
+                    scale = 1.0
+
+            columns = result.columns[2:4]
+            result[columns] *= scale
+            result[columns] = result[columns].apply(pd.to_timedelta, unit=unit)
+            if result.columns[2][0] == "t":  # "t" or "i":
```
seisman, line 251:
```suggestion
            if columns[0][0] == "t":  # "t" or "i":
```

weiji14:
I'll give this a proper review over the weekend, a bit busy this week with some deadlines :melting_face:

weiji14 (approved):
Cool, thanks also for handling the output differences between macOS and Linux (xref https://github.com/GenericMappingTools/pygmt/issues/3194). Pre-approving as the main logic around timedelta conversion checks out ok. Suggestions below are mostly documentation related or minor.

PR #3184, opened 2024-04-20
Deprecate the build_arg_string function in v0.12.0 (will be removed in v0.14.0) and use build_arg_list instead

seisman:

After PR #3149 and #3132, `build_arg_string` is no longer used anywhere so can be removed. However, although it's not a public API function, we know that someone are using it in their own wrappers (e.g., https://gist.github.com/yasuit21/db1d676d784769b80365a3528d1ab547), so better to raise a warning in the next few releases.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -520,7 +520,7 @@ def build_arg_string(kwdict, confdict=None, infile=None, outfile=None):
     """
     msg = (
         "Utility function `build_arg_string()' is deprecated in v0.12.0 and will be "
-        "removed in v0.14.0. Uease `build_arg_list()' instead."
+        "removed in v0.14.0. Use `build_arg_list()' instead."
     )
     warnings.warn(msg, category=FutureWarning, stacklevel=2)
 
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -10,6 +10,7 @@ import string
 import subprocess
 import sys
 import time
+import warnings
 import webbrowser
 from collections.abc import Iterable, Sequence
 from typing import Any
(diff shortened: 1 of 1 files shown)


pygmt/helpers/utils.py
```diff
@@ -513,8 +518,13 @@ def build_arg_string(kwdict, confdict=None, infile=None, outfile=None):
     ... )
     input.txt -A0 -B -Crainbow --FORMAT_DATE_MAP="o dd" ->output.txt
     """
-    gmt_args = []
+    msg = (
+        "Utility function `build_arg_string()' is deprecated in v0.12.0 and will be "
+        "removed in v0.14.0. Uease `build_arg_list()' instead."
```
seisman, line 523: OK with removing the function in two releases?

michaelgrund, line 523:
```suggestion
        "removed in v0.14.0. Use `build_arg_list()' instead."
```

michaelgrund (approved):

pygmt/helpers/utils.py
```diff
@@ -513,8 +518,13 @@ def build_arg_string(kwdict, confdict=None, infile=None, outfile=None):
     ... )
     input.txt -A0 -B -Crainbow --FORMAT_DATE_MAP="o dd" ->output.txt
     """
-    gmt_args = []
+    msg = (
+        "Utility function `build_arg_string()' is deprecated in v0.12.0 and will be "
+        "removed in v0.14.0. Use `build_arg_list()' instead."
+    )
```
yvonnefroehlich, line 524:
```suggestion
    msg = (
        "Utility function 'build_arg_string()' is deprecated in v0.12.0 and will be "
        "removed in v0.14.0. Use 'build_arg_list()' instead."
    )
```

weiji14: Typing too much LaTeX @seisman? :laughing:

PR #3187, opened 2024-04-21
CI: Use 'gh release' to upload assets to release

seisman:

Use `gh release` to upload assets to release, so we don't need to use the `shogo82148/actions-upload-release-asset` action anymore.

Reference: https://michael-mckenna.com/how-to-upload-file-to-github-release-in-a-workflow/

The same commit was pushed to my own fork (https://github.com/seisman/pygmt). See https://github.com/seisman/pygmt/actions/runs/8772441014 for the workflow run and the testing release https://github.com/seisman/pygmt/releases/tag/v0.12.0.

### DIFF
--- a/.github/workflows/release-baseline-images.yml
+++ b/.github/workflows/release-baseline-images.yml
@@ -17,6 +17,10 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'GenericMappingTools/pygmt'
 
+    permissions:
+      # To write assets to GitHub release
+      contents: write
+
     steps:
     - name: Checkout
       uses: actions/checkout@v4.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/release-baseline-images.yml
```diff
@@ -35,7 +39,6 @@ jobs:
         shasum -a 256 baseline-images.zip
 
     - name: Upload baseline image as a release asset
-      uses: shogo82148/actions-upload-release-asset@v1.7.5
-      with:
-        upload_url: ${{ github.event.release.upload_url }}
-        asset_path: baseline-images.zip
+      run: gh release upload ${{ github.ref_name }} baseline-images.zip
```
weiji14, line 42: The blog post you linked uses `github.event.release.tag_name` instead of `github.ref_name`. I had a look at https://docs.github.com/en/actions/learn-github-actions/contexts#github-context, and I suppose they're the same for tags (e.g. `v0.12.0`), but I suppose `github.ref_name` would be safer if we need to re-run the workflow, say if the `dvc pull` step failed?

seisman: Just triggered another run of the workflow in my own fork (https://github.com/seisman/pygmt/actions/runs/8772441014/job/24079741738).
```
  gh release upload v0.12.0 baseline-images.zip
  shell: /usr/bin/bash -e {0}
  env:
    GH_TOKEN: ***
```
`github.ref_name` is `v0.12.0` for the manually triggered run. I guess `github.ref_name` is better.

weiji14 (approved):

PR #3207, opened 2024-04-28
Remove sequence_plus converter from kwargs_to_string

seisman:

After #3116, `sequence_plus` is no longer used in the project. It also makes no sense to allow a "+"-separated sequence as arguments, since it's not Pythonic.

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -633,7 +633,6 @@ def kwargs_to_strings(**conversions):
     * 'sequence': transforms a sequence (list, tuple) into a ``'/'`` separated
       string
     * 'sequence_comma': transforms a sequence into a ``','`` separated string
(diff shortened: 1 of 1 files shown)


pygmt/helpers/decorators.py
```diff
@@ -633,7 +633,6 @@ def kwargs_to_strings(**conversions):
     * 'sequence': transforms a sequence (list, tuple) into a ``'/'`` separated
       string
     * 'sequence_comma': transforms a sequence into a ``','`` separated string
```
yvonnefroehlich, line 635:
```suggestion
    * "sequence": transform a sequence (list, tuple) into a ``"/"`` separated string
    * "sequence_comma": transform a sequence into a ``","`` separated string
```

yvonnefroehlich (approved):

PR #3209, opened 2024-04-29
Figure.grdcontour: Deprecate parameter "interval" to "levels" (FutureWarning since v0.12.0, will be removed in v0.16.0)

### LINKED ISSUE #1042: Consistent (non-common) aliases between modules

The **-C** arguments (parameters) in [contour](https://docs.generic-mapping-tools.org/dev/contour.html#c) and [grdcontour](https://docs.generic-mapping-tools.org/dev/grdcontour.html#c) do essentially the same thing. But, PyGMT uses different aliases for these parameters in the two methods (``interval`` and ``level``). I think it would be better to use the same aliases for options that are very similar across the various modules. This is harder to track for non-common options, so perhaps we can also create a tool for identifying similar GMT arguments across modules.

yvonnefroehlich:

This PR deprecate the parameter `interval` to `levels` (**C**) for `Figur.grdcontour`. 
Then this alias is consistent with `Figure.contour`, the GMT documentation, and matplolib (details on the discussion at https://github.com/GenericMappingTools/pygmt/issues/1042#issuecomment-2081541546).

Fixes #1042 (partly)

**Preview**: https://pygmt-dev--3209.org.readthedocs.build/en/3209/api/generated/pygmt.Figure.grdcontour.html

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/tutorials/advanced/contour_map.py
+++ b/examples/tutorials/advanced/contour_map.py
@@ -33,16 +33,12 @@ fig.show()
 # Contour line settings
 # ---------------------
 #
-# Use the ``annotation`` and ``interval`` parameters to adjust contour line
+# Use the ``annotation`` and ``levels`` parameters to adjust contour line
 # intervals. In the example below, there are contour intervals every 250 meters
 # and annotated contour lines every 1,000 meters.
 
 fig = pygmt.Figure()
-fig.grdcontour(
-    annotation=1000,
-    interval=250,
-    grid=grid,
-)
+fig.grdcontour(annotation=1000, levels=250, grid=grid)
 fig.show()
 
 
@@ -57,7 +53,7 @@ fig.show()
 fig = pygmt.Figure()
 fig.grdcontour(
     annotation=1000,
-    interval=250,
+    levels=250,
     grid=grid,
     limit=[-4000, -2000],
 )
@@ -74,7 +70,7 @@ fig.show()
 fig = pygmt.Figure()
 fig.grdcontour(
     annotation=1000,
-    interval=250,
+    levels=250,
     grid=grid,
     limit=[-4000, -2000],
     projection="M10c",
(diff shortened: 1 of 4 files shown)


examples/tutorials/advanced/contour_map.py
```diff
@@ -33,16 +33,12 @@
...
 # intervals. In the example below, there are contour intervals every 250 meters
 # and annotated contour lines every 1,000 meters.
 
 fig = pygmt.Figure()
-fig.grdcontour(
-    annotation=1000,
-    interval=250,
-    grid=grid,
-)
+fig.grdcontour(annotation=1000, levels=250, grid=grid)
```
seisman, line 41: Can you please update this example to always put `grid` as the first parameter?
```suggestion
fig.grdcontour(grid=grid, annotation=1000, levels=250)
```

examples/tutorials/advanced/contour_map.py
```diff
@@ -57,7 +53,7 @@
 fig = pygmt.Figure()
 fig.grdcontour(
     annotation=1000,
-    interval=250,
+    levels=250,
     grid=grid,
     limit=[-4000, -2000],
 )
```
seisman, line 59: This code block and many others can be rewritten into a single line after we set the line length limit to 88 characters.

yvonnefroehlich: Yes, I should find the time to make both updates. Should this be included in this PR, or better in separate PR and have this PAR for the deprecation?

seisman (approved):

examples/tutorials/advanced/contour_map.py
seisman: OK. Let's keep this PR small.

yvonnefroehlich: Just submitted PR #3212 for working on this.

PR #3214, opened 2024-04-30
Document that a list of file names, pathlib.Path objects, URLs or remote files are supported

### LINKED ISSUE #2723: Support passing a list of file names as input
When working on https://github.com/GenericMappingTools/pygmt/pull/2722, I realized most PyGMT functions/methods don't allow a list of file names as input. For example,
```
import pygmt
fig = pygmt.Figure()
fig.plot(data=["@hotspots.txt", "@Table_5_11.txt"])
pygmt-session [WARNING]: Unable to parse 2 longitude strings
```

The main reason is that the `data_kind` function _incorrectly_ returns a `matrix` type:
```
>>> from pygmt.helpers import data_kind
>>> data_kind(data=["@hotspots.txt", "@Table_5_11.txt"])
'matrix'
```
then we call the `virtualfile_from_vectors` to pass the data to the GMT API, which is expected to fail.

Then, the question is, if a list is passed to `data`, how do we know if it's a list of files or a valid record (e.g., `data=[0, 1]`). 

To answer the question, we need to have a strict definition for "**table files**". As defined (based on my personal understanding) in https://github.com/GenericMappingTools/pygmt/pull/2722, 

> - A table is a 2D array of data, with *M* rows and *N* columns. Each column 
  represents a different variable (e.g., *x*, *y* and *z*) and each row 
  represents a different record. 

Under this definition, a table should always be a 2D array/list, even if it has only one row. Thus, `data=[[0, 1]]` is a valid table, and `data=[0, 1]` is not).

If we agree with the strict definition, then the solution is to check if `data` is 1-D or 2-D array. If 1-D, then it's a list of file names, otherwise, it's a 2-D table.

With this solution, it also means the issue report https://github.com/GenericMappingTools/pygmt/issues/1132 is no longer valid.

seisman:

See 0c17051b7bacc0f0e3efccb8b51ea67415b484d0 for the added statement. 

Also take the opportunity to rewrap the docstrings to 88 characters in 3c4294befba7afbaece2bf16addd365c2b7fa464.

Closes #2723.

### DIFF
--- a/examples/get_started/04_table_inputs.py
+++ b/examples/get_started/04_table_inputs.py
@@ -59,6 +59,9 @@ Path("input_data.dat").unlink()
 # - A file name prefixed with ``@`` (e.g., ``data="@input_data.dat"``), which
 #   is a special syntax in GMT to indicate that the file is a remote file
 #   hosted on the GMT data server.
+#
+# Additionally, PyGMT also supports a list of file names, :class:`pathlib.Path` objects,
+# URLs, or remote files, providing more flexibility in specifying input files.
 
 # %%
 # 2-D array: `list`, `numpy.ndarray`, and `pandas.DataFrame`
--- a/examples/get_started/04_table_inputs.py
+++ b/examples/get_started/04_table_inputs.py
@@ -4,20 +4,19 @@
 
 Generally, PyGMT accepts two different types of data inputs: tables and grids.
 
-- A table is a 2-D array with rows and columns. Each column represents a
-  different variable (e.g., *x*, *y* and *z*) and each row represents a
-  different record.
-- A grid is a 2-D array of data that is regularly spaced in the x and y
-  directions (or longitude and latitude).
+- A table is a 2-D array with rows and columns. Each column represents a different
+  variable (e.g., *x*, *y* and *z*) and each row represents a different record.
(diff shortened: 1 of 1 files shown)


examples/get_started/04_table_inputs.py
```diff
@@ -51,14 +50,16 @@
...
 # - A full URL. PyGMT will download the file to the current directory first.
-# - A file name prefixed with ``@`` (e.g., ``data="@input_data.dat"``), which
-#   is a special syntax in GMT to indicate that the file is a remote file
-#   hosted on the GMT data server.
+# - A file name prefixed with ``@`` (e.g., ``data="@input_data.dat"``), which is a
+#   special syntax in GMT to indicate that the file is a remote file hosted on the GMT
+#   data server.
+#
+# Additionally, PyGMT also supports a list of file names, :class:`pathlib.Path` objects,
+# URLs, or remote files, providing more flexibility in specifying input files.
```
michaelgrund, line 62:
```suggestion
# URLs, or remote files, to provide more flexibility in specifying input files.
```

michaelgrund (approved):

yvonnefroehlich:
I adjusted the title of this PR, to make clear that it modifies the Intro "04 Tabel inputs". Feel free to adjust this further.

PR #3232, opened 2024-05-07
clib: Improve Session.get_default docstring to clarify that GMT configuration parameters are supported

seisman:

Just docstring changes to clarify that `Session.get_default` can also get the **current** value of GMT configuration settings. One test is also added.

Need to note that the API name `get_default` may make people think that it returns the **default** value, but that's not true. The **current** value of a GMT configuration parameter is returned! I don't add an extra test to verify it, because the `pygmt.config` method already rely on the return value:
https://github.com/GenericMappingTools/pygmt/blob/a4d2b8e12cf1d10e31645768db0aae4253a058a3/pygmt/src/config.py#L193-L199

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -453,11 +453,12 @@ class Session:
 
         self.session_pointer = None
 
-    def get_default(self, name):
+    def get_default(self, name: str) -> str:
         """
-        Get the value of a GMT default parameter (library version, paths, etc).
+        Get the value of a GMT configuration parameter or a GMT API parameter.
 
-        Possible default parameter names include:
+        In addtional to the long list of GMT configuration parameters, following API
+        parameter names are also supported:
 
         * ``"API_VERSION"``: The GMT API version
         * ``"API_PAD"``: The grid padding setting
@@ -473,13 +474,14 @@ class Session:
 
         Parameters
         ----------
-        name : str
-            The name of the default parameter (e.g., ``"API_VERSION"``)
+        name
+            The name of the GMT configuration parameter (e.g., ``"PROJ_LENGTH_UNIT"``)
+            or a GMT API parameter (e.g., ``"API_VERSION"``).
 
         Returns
         -------
-        value : str
-            The default value for the parameter.
+        value
+            The current value for the parameter.
 
         Raises
         ------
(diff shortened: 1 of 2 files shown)


pygmt/clib/session.py
```diff
@@ -493,15 +495,11 @@ def get_default(self, name):
         )
 
         # Make a string buffer to get a return value
-        value = ctp.create_string_buffer(10000)
-
+        value = ctp.create_string_buffer(4096)
```
seisman, line 498: The `GMT_Get_Default` API function only says that `value` must have enough space to store the return value but doesn't say the exact length, but function [`gmtlib_getparameter`](https://github.com/GenericMappingTools/gmt/blob/9d48c37ef02882fd8f97673f4da35250f8c6c05f/src/gmt_init.c#L12631) (called by `GMT_Get_Default`) defines a `GMT_BUFSIZ`-long (4096) char. 

So, we can safely change 10000 to 4096 in this function.

pygmt/clib/session.py
```diff
@@ -453,11 +453,12 @@ def destroy(self):
...
         self.session_pointer = None
 
-    def get_default(self, name):
+    def get_default(self, name: str) -> str:
         """
-        Get the value of a GMT default parameter (library version, paths, etc).
+        Get the value of a GMT configuration parameter or a GMT API parameter.
 
-        Possible default parameter names include:
+        In addtional to the long list of GMT configuration parameters, following API
```
michaelgrund, line 460:
```suggestion
        In addition to the long list of GMT configuration parameters, following API
```

michaelgrund, line 460:
```suggestion
        In addtional to the long list of GMT configuration parameters, the following API
```

michaelgrund (approved):

pygmt/clib/session.py
```diff
@@ -453,11 +453,12 @@ def destroy(self):
...
         self.session_pointer = None
 
-    def get_default(self, name):
+    def get_default(self, name: str) -> str:
         """
-        Get the value of a GMT default parameter (library version, paths, etc).
+        Get the value of a GMT configuration parameter or a GMT API parameter.
 
-        Possible default parameter names include:
+        In addition to the long list of GMT configuration parameters, following API
```
seisman, line 460:
```suggestion
        In addition to the long list of GMT configuration parameters, the following API
```

Issue #3180, opened 2024-04-19
Release PyGMT v0.12.0


seisman:
**Release**: [v0.12.0](https://github.com/GenericMappingTools/pygmt/milestone/18)
**Scheduled Date**: 2024/05/01
**Pull request due date**: 2024/04/28

**Priority PRs/issues to complete prior to release**
- [x] #2760 
- [x] #3186
       #1349
- [x] #3048
       #2914
- [x] #3116 
       #2706 
- [x] #3209 
       #1042 (partly)
- [x] #3210
       #2704

**Before release**:
- [x] Check [SPEC 0](https://scientific-python.org/specs/spec-0000/) to see if we need to bump the minimum supported versions of GMT, Python and core package dependencies (NumPy/Pandas/Xarray) [@seisman]
- [x] Run `make codespell` to check common misspellings. If there are any, either fix them or add them to `ignore-words-list` in `pyproject.toml`
- [ ] Check to ensure that:
  - [ ] All tests pass in the ["GMT Legacy Tests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_tests_legacy.yaml)
  - [ ] All tests pass in the ["GMT Dev Tests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_tests_dev.yaml)
  - [x] All tests pass in the ["Doctests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_doctests.yaml)
  - [x] Deprecations and related tests are removed for this version by running `grep --include="*.py" -r 'remove_version="vX.Y.Z"' pygmt` from the base of the repository
- [x] Reserve a DOI on [Zenodo](https://zenodo.org) by clicking on "New Version" `10.5281/zenodo.11062720`
- [x] Finish up 'Changelog entry for v0.x.x' Pull Request: https://github.com/GenericMappingTools/pygmt/pull/3201
  - [x] Add a new entry in `doc/_static/version_switch.js` for documentation switcher
  - [x] Update `CITATION.cff` and BibTeX at https://github.com/GenericMappingTools/pygmt#citing-pygmt
    - [x] Update authorship list
    - [x] Update DOI (and url for BibTeX)
    - [x] Update version
    - [x] Update date released
  - [x] Add the documentation link `doc/minversions.md`
  - [x] Add minimum required version information `doc/minversions.md`
[...]

PR #3234, opened 2024-05-07
CI: Set environment cache key based on current week of year to avoid outdated cache in the "Setup Micromamba" step

seisman:
Take the [latest run](https://github.com/GenericMappingTools/pygmt/actions/runs/8979469426) of the "Tests" workflow on the main branch as an example. The "Setup Micromamba" step in the "Ubuntu + Python 3.10" job uses the cache with key `micromamba-environment--linux-64-pygmt-args-4fc7725-root-dcc80ee`, which was cached last month as shown here (https://github.com/GenericMappingTools/pygmt/actions/caches).

image

The setup-micromamba action says (https://github.com/mamba-org/setup-micromamba#caching):

> Note that the specified cache key is only the prefix of the real cache key. setup-micromamba will append a hash of the environment file and the custom-args as well as the environment name and OS to the cache key.

And GitHub says (https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy):

> GitHub will remove any cache entries that have not been accessed in over 7 days.

So, the cached environment won't be refreshed unless we make changes to the packages listed in the `create-args` argument. We're using outdated packages in our workflows, which explains why sometimes we see old packages installed in CI.

**The solution is to define the cache key based on the current date.** 

We can define the cache key like `micromamba-environment-2024-05-07`, which means the cache will expire on the next day so the first CI run on the next day will be slower since the cache is missing. I feel one day is a little too short for PyGMT, so in this PR, the cache key is defined like `micromamba-environment-2024-W19`, in which `2024-W19` means the 19th week of 2024. The cache will persist for one week.

Patches #2946.

### DIFF
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -74,6 +74,10 @@ jobs:
           # fetch all history so that setuptools-scm works
           fetch-depth: 0
 
+      - name: Get current week number of year
+        id: date
+        run: echo "date=$(date +%Y-W%W)" >> $GITHUB_OUTPUT  # e.g., 2024-W19
+
       # Install Micromamba with conda-forge dependencies
       - name: Setup Micromamba
         uses: mamba-org/setup-micromamba@v1.8.1
@@ -85,6 +89,8 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
+          # environment cache is persistent in one week.
+          cache-environment-key: micromamba-environment-${{ steps.date.outputs.date }}
           create-args: >-
             python=3.12
             gmt=6.5.0
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -89,7 +89,7 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
-          # environment cache is persistent in one week.
+          # environment cache is persistent for one week.
           cache-environment-key: micromamba-environment-${{ steps.date.outputs.date }}
           create-args: >-
             python=3.12
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -112,7 +112,7 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
-          # environment cache is persistent in one week.
+          # environment cache is persistent for one week.
           cache-environment-key: micromamba-environment-${{ steps.date.outputs.date }}
           create-args: >-
             python=${{ matrix.python-version }}${{ matrix.optional-packages }}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -106,7 +106,7 @@ jobs:
             channels:
               - conda-forge
               - nodefaults
-          cache-downloads: false
+          cache-downloads: true
           cache-environment: false
           create-args: >-
             python=${{ matrix.python-version }}${{ matrix.optional-packages }}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -107,7 +107,7 @@ jobs:
               - conda-forge
               - nodefaults
           cache-downloads: false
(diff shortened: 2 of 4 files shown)


seisman:
With the latest packages installed, now we can also see the Fiona errors reported in https://github.com/GenericMappingTools/pygmt/issues/3180#issuecomment-2081801690
```
______________ test_geopandas_info_shapely[multipolygon-desired0] ______________

geojson = <MULTIPOLYGON (((0 0, 20 0, 10 20, 0 0), (3 2, 10 16, 17 2, 3 2)), ((6 4, 14...>

    @contextmanager
    def tempfile_from_geojson(geojson):
        """
        Saves any geo-like Python object which implements ``__geo_interface__`` (e.g. a
        geopandas.GeoDataFrame or shapely.geometry) to a temporary OGR_GMT text file.
    
        Parameters
        ----------
        geojson : geopandas.GeoDataFrame
            A geopandas GeoDataFrame, or any geo-like Python object which
            implements __geo_interface__, i.e. a GeoJSON.
    
        Yields
        ------
        tmpfilename : str
            A temporary OGR_GMT format file holding the geographical data.
            E.g. '1a2b3c4d5e6.gmt'.
        """
        with GMTTempFile(suffix=".gmt") as tmpfile:
            import geopandas as gpd
    
            Path(tmpfile.name).unlink()  # Ensure file is deleted first
            ogrgmt_kwargs = {"filename": tmpfile.name, "driver": "OGR_GMT", "mode": "w"}
            try:
                # OGR_GMT only supports 32-bit integers. We need to map int/int64
                # types to int32/float types depending on if the column has an
                # 32-bit integer overflow issue. Related issues:
                # https://github.com/geopandas/geopandas/issues/967#issuecomment-84287[770](https://github.com/GenericMappingTools/pygmt/actions/runs/8981820061/job/24669172952?pr=3234#step:8:771)4
                # https://github.com/GenericMappingTools/pygmt/issues/2497
>               if geojson.index.name is None:
E               AttributeError: 'MultiPolygon' object has no attribute 'index'

../pygmt/helpers/tempfile.py:142: AttributeError

During handling of the above exception, another exception occurred:

>   ???

fiona/ogrext.pyx:136: 

>   ???
E   fiona._err.CPLE_OpenFailedError: <fiona.io.MemoryFile object at 0x7f843fe3d410>: No such file or director
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers for comments before I apply the same changes to other workflows.

.github/workflows/ci_tests_legacy.yaml
```diff
@@ -58,8 +58,6 @@ jobs:
             channels:
               - conda-forge
               - nodefaults
-          cache-downloads: false
-          cache-environment: true
```
seisman, line 62: This workflow is only scheduled to run weekly. We don't care about the speed of the workflow, so no need to cache the enviroment.

.github/workflows/ci_docs.yml
```diff
@@ -85,6 +89,8 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
+          # environment cache is persistent in one week.
```
yvonnefroehlich, line 92:
```suggestion
          # environment cache is persistent for one week.
```

.github/workflows/ci_tests.yaml
```diff
@@ -108,6 +112,8 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
+          # environment cache is persistent in one week.
```
yvonnefroehlich, line 115:
```suggestion
          # environment cache is persistent for one week.
```

.github/workflows/ci_tests_dev.yaml
```diff
@@ -63,6 +67,8 @@ jobs:
               - nodefaults
           cache-downloads: false
           cache-environment: true
+          # environment cache is persistent in one week.
```
yvonnefroehlich, line 70:
```suggestion
          # environment cache is persistent for one week.
```

Issue #3231, opened 2024-05-07
GeoPandas: Migrate the engine from fiona to pyogrio


seisman:
GeoPandas v1.0.0-alpha1 has been released (https://geopandas.org/en/latest/docs/changelog.html#version-1-0-0-alpha1-apr-13-2024). The biggest change is that the default I/O engine is changed from Fiona to pyogrio. 

So, when geopandas v1.0.0 is released, PyGMT will fail to work, because:

1. Fiona is no longer a required dependency of geopandas, so it's not installed by default
2. The pyogrio engine doesn't support `scheme` parameter so some [codes](https://github.com/GenericMappingTools/pygmt/blob/c783a79ed55068a424eb913d271fa880fa44928d/pygmt/helpers/tempfile.py#L150) will break.

I think we should start the migration now following the [instructions](https://geopandas.org/en/latest/docs/user_guide/fiona_to_pyogrio.html). 

1. Add `pyogrio` as a optional dependency since geopandas 0.x doesn't install it by default
2. Add `engine="pyogrio"` to use the pyogrio
3. Cleanup the codes and avoid using fiona

Thoughts @GenericMappingTools/pygmt-maintainers especially @weiji14?

weiji14:
Not sure if we can migrate completely away from `fiona` yet, at least in terms of writing out OGR_GMT files. Pyogrio doesn't list OGR_GMT under https://pyogrio.readthedocs.io/en/latest/supported_formats.html#full-read-and-write-support, though it seems to depend on what GDAL was compiled with. We might just need to test things out and see first.

seisman:
OGR_GMT is built-in by default in GDAL (https://gdal.org/drivers/vector/gmt.html), so we should be safe unless some package distributors decide to disable OGR_GMT explicitly.

weiji14:
Ok, opened a proof of concept PR at #3237 to switch from fiona to pyogrio. The way it works is that we get a GeoJSON representation from `__geo_interface__`, which pyogrio reads into a `geopandas.GeoDataFrame`, and that gets output to an OGR_GMT file.

Downside is that this means geopandas would become a stronger optional dependency. Before, someone could install just fiona without geopandas and plot shapely objects, but now, doing so via pyogrio would require geopandas.

seisman:
Maybe we can try geopandas first, then pyogrio, then fiona?

PR #3237, opened 2024-05-09
POC: Handle __geo_interface__ objects using pyogrio instead of fiona

weiji14:

Save objects that implement `__geo_interface__` (e.g. shapely geometries) to a temporary OGR_GMT file via pyogrio and geopandas instead of fiona.

Note that this would mean that users must have `geopandas` installed, and cannot just install `fiona`.

Fixes error seen at https://github.com/GenericMappingTools/pygmt/issues/3180#issuecomment-2081801690, Addresses #3231

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/tempfile.py
+++ b/pygmt/helpers/tempfile.py
@@ -154,14 +154,12 @@ def tempfile_from_geojson(geojson):
             # Other 'geo' formats which implement __geo_interface__
             import json
 
-            import fiona
+            import pyogrio
 
-            with fiona.Env():
-                jsontext = json.dumps(geojson.__geo_interface__)
-                # Do Input/Output via Fiona virtual memory
-                with fiona.io.MemoryFile(file_or_bytes=jsontext.encode()) as memfile:
-                    geoseries = gpd.GeoSeries.from_file(filename=memfile)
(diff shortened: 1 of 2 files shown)


pygmt/helpers/tempfile.py
```diff
@@ -154,14 +154,12 @@ def tempfile_from_geojson(geojson):
             # Other 'geo' formats which implement __geo_interface__
```
seisman, line 154: The pyogrio package doesn't support the `schema` parameter. With geopandas 1.0 released, pyogrio will be the default engine, and then the code at lines 136-150 will break.

We need to find a way to support both geopandas 0.x and 1.x.

seisman:
After PR #3247, changes in this PR are no longer needed, but we still need to address the following errors with geopandas 1.x:
```
        if schema is not None:
>           raise ValueError(
                "The 'schema' argument is not supported with the 'pyogrio' engine."
            )
E           ValueError: The 'schema' argument is not supported with the 'pyogrio' engine.
```
I guess we need to check the geopandas version, and pass `schema` or not depending on if it's geopandas 0.x and 1.x. Not sure if `pyogrio` has the issue with int64 numbers.

Issue #3242, opened 2024-05-11
pygmt.show_versions should warn if incompatible packages are installed


cuitianyu20:
### Description of the problem

When I use the `pygmt.Figure.plot` command to draw a filled shaded region with a rectangle, I find that regardless of whether I set the transparency to 80 or 100, the effect is always opaque. Transparency doesn't seem to work. (I've output files in both PNG and PDF formats). 
[test.pdf](https://github.com/GenericMappingTools/pygmt/files/15281118/test.pdf)
![test](https://github.com/GenericMappingTools/pygmt/assets/68799423/81dcd2cf-b742-4109-a18f-82bf476365f2)

My PyGMT version is 0.10.0.

### Minimal Complete Verifiable Example

```python
fig = pygmt.Figure()
fig.basemap(region=[-3,4,10,20], 
        projection="X5i/5i",
        frame=['xa1f0.5+l"Time (s)"', 'ya1f0.5+l"Distance (\260)"', "WSen"])
region_data = [[-1,10],[1,10],[1,20],[-1,20]]
fig.plot(data=region_data, fill='#66BFBF', transparency=90, close=True)
fig.savefig('test.pdf')
fig.savefig('test.png')
```

### Full error message

```python-traceback
The transparency in `pygmt.Figure.plot` command doesn't work.
```

### System information

```bash
PyGMT information:
  version: v0.10.0
System information:
  python: 3.9.15 | packaged by conda-forge | (main, Nov 22 2022, 15:55:03)  [GCC 10.4.0]
  executable: /home/tianyu/anaconda3/envs/pygmt/bin/python
  machine: Linux-5.15.0-105-generic-x86_64-with-glibc2.31
Dependency information:
  numpy: 1.23.5
  pandas: 1.5.2
  xarray: 2022.11.0
  netCDF4: 1.6.2
  packaging: 21.3
  contextily: None
  geopandas: None
  IPython: None
  rioxarray: None
  ghostscript: 10.03.0
GMT library information:
  binary version: 6.4.0
  cores: 6
  grid layout: rows
  image layout: 
  library path: /home/*/anaconda3/envs/pygmt/lib/libgmt.so
  padding: 2
  plugin dir: /home/*/anaconda3/envs/pygmt/lib/gmt/plugins
  share dir: /home/*/anaconda3/envs/pygmt/share/gmt
  version: 6.4.0
```

seisman:
[...]

PR #466, opened 2020-05-31
Add show_versions() function for printing debugging information used in issue reports

seisman:
**Description of proposed changes**

The `show_versions` function reports system information, package versions and GMT library information. The output looks like:
```
PyGMT information:
  version: v0.1.1+20.g4d11e8c.dirty
System information:
  python: 3.7.6 (default, Jan  8 2020, 13:42:34)  [Clang 4.0.1 (tags/RELEASE_401/final)]
  executable: /Users/seisman/.anaconda/bin/python
  machine: Darwin-19.4.0-x86_64-i386-64bit
Dependency information:
  numpy: 1.18.1
  pandas: 1.0.1
  xarray: 0.15.1
  netCDF4: 1.5.3
  packaging: 20.1
  ghostscript: 9.50
  gmt: 6.1.0_ddd829b_2020.05.29
GMT library information:
  binary dir: /Users/seisman/.anaconda/bin
  cores: 8
  grid layout: rows
  library path: /Users/seisman/local/GMT/lib/libgmt.dylib
  padding: 2
  plugin dir: /Users/seisman/local/GMT/lib/gmt/plugins
  share dir: /Users/seisman/local/GMT/share
  version: 6.1.0
```

TODO: 

- [x] Update the github issue template
- [x] Update the pygmt test instructions in the documentation
- [x] Replace `print_clib_info()` with `show_versions` in Makefile

Fixes #460

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [x] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

PR #3244, opened 2024-05-11
pygmt.show_versions: Warn about incompatible ghostscript versions

seisman:

1. Add a private function `_get_clib_info` to return the dictionary of GMT library information
2. Refactor `_get_ghostscript_version` using match-case statement
3. Add a private function `_check_ghostscript_version` which return warnings if the installed ghostscript version is problematic
4. Store the output in a list and print it, rather than calling `print` multiple times.

Address #3242

### DIFF
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -143,7 +143,7 @@ def show_versions(file=sys.stdout):
 
     def _get_ghostscript_version() -> str | None:
         """
-        Get ghostscript version.
+        Get Ghostscript version.
         """
         match sys.platform:
             case "linux" | "darwin":
--- a/pygmt/__init__.py
+++ b/pygmt/__init__.py
@@ -102,6 +102,9 @@ def show_versions(file=sys.stdout):
     - System information (Python version, Operating System)
     - Core dependency versions (NumPy, Pandas, Xarray, etc)
     - GMT library information
+
+    It also warns users if the installed Ghostscript version has serious bugs or is
+    incompatible with the installed GMT version.
     """
 
     import importlib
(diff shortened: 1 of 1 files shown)


pygmt/__init__.py
```diff
@@ -110,8 +113,18 @@ def show_versions(file=sys.stdout):
     import subprocess
 
     from packaging.requirements import Requirement
+    from packaging.version import Version
+
+    def _get_clib_info() -> dict:
```
seisman, line 118: The `pygmt.print_clib_info` function was initially added in #176 but has been replaced by `pygmt.show_versions` in #466 (first appear in v0.1.2).

In this PR, a new private function `_get_clib_info` is added which returns a dictionary instead of printing a long string. 

It's time to retire the `print_clib_info` function. Since it's rarely used, I think we can just remove it without a deprecation warning. Of course, we should do it in a separate PR so that at least we have a deprecation entry in the v0.13.0 changelog.

pygmt/__init__.py
```diff
@@ -147,24 +162,53 @@ def _get_ghostscript_version():
...
+                return (
+                    f"Ghostscript v{gs_version} is too old and may have serious bugs. "
+                    "Please consider upgrading your Ghostscript."
+                )
+            case v if Version("10.00") <= v < Version("10.02"):
+                return (
+                    f"Ghostscript v{gs_version} has known bugs. "
+                    "Please consider upgrading to version v10.02 or later."
+                )
+            case v if v >= Version("10.00"):
```
seisman, line 180:
```suggestion
            case v if v >= Version("10.02"):
```

pygmt/__init__.py
```diff
@@ -147,24 +162,53 @@ def _get_ghostscript_version():
...
+                from pygmt.clib import __gmt_version__
+
+                if Version(__gmt_version__) < Version("6.5.0"):
+                    return (
+                        f"GMT v{__gmt_version__} doesn't support Ghostscript "
+                        "v{gs_version}. Please consider upgrading to GMT>=6.5.0 or "
+                        "downgrade Ghostscript to v9.56."
```
seisman, line 187:
```suggestion
                        "downgrading to Ghostscript v9.56."
```

pygmt/__init__.py
```diff
@@ -128,17 +141,19 @@ def _get_module_version(modname):
         except ImportError:
             return None
 
-    def _get_ghostscript_version():
+    def _get_ghostscript_version() -> str | None:
         """
         Get ghostscript version.
```
yvonnefroehlich, line 146:
```suggestion
        Get Ghostscript version.
```

PR #3247, opened 2024-05-12
geopandas: Use io.StringIO to read geojson data and handle compatibility with geopandas v0.x and v1.x

### LINKED ISSUE #3231: GeoPandas: Migrate the engine from fiona to pyogrio

seisman:

1. For a geojson object with the `__geo_interface__`, read it via `gdf.read_file` by wrapping the geojson object into a `io.StringIO` object
2. Backward-compatible with geopandas v0.x (default engine is `fiona`) and v1.x (default engine is `pyogrio`).

Supersede #3237, so closes #3237. 

This PR should fix the errors seen at https://github.com/GenericMappingTools/pygmt/issues/3180#issuecomment-2081801690. So also closes #3231.

### DIFF
--- a/pygmt/helpers/tempfile.py
+++ b/pygmt/helpers/tempfile.py
@@ -2,6 +2,7 @@
 Utilities for dealing with temporary file management.
 """
 
+import io
 import uuid
 from contextlib import contextmanager
 from pathlib import Path
@@ -154,14 +155,8 @@ def tempfile_from_geojson(geojson):
             # Other 'geo' formats which implement __geo_interface__
             import json
 
-            import fiona
-
-            with fiona.Env():
-                jsontext = json.dumps(geojson.__geo_interface__)
-                # Do Input/Output via Fiona virtual memory
-                with fiona.io.MemoryFile(file_or_bytes=jsontext.encode()) as memfile:
-                    geoseries = gpd.GeoSeries.from_file(filename=memfile)
-                    geoseries.to_file(**ogrgmt_kwargs)
+            jsontext = json.dumps(geojson.__geo_interface__)
+            gpd.read_file(io.StringIO(jsontext)).to_file(**ogrgmt_kwargs)
 
         yield tmpfile.name
 
--- a/pygmt/helpers/tempfile.py
+++ b/pygmt/helpers/tempfile.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 import numpy as np
+from packaging.version import Version
 
 
 def unique_name():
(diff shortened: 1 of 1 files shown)


pygmt/helpers/tempfile.py
```diff
@@ -154,14 +155,8 @@ def tempfile_from_geojson(geojson):
...
-            import fiona
-
-            with fiona.Env():
-                jsontext = json.dumps(geojson.__geo_interface__)
-                # Do Input/Output via Fiona virtual memory
-                with fiona.io.MemoryFile(file_or_bytes=jsontext.encode()) as memfile:
-                    geoseries = gpd.GeoSeries.from_file(filename=memfile)
-                    geoseries.to_file(**ogrgmt_kwargs)
+            jsontext = json.dumps(geojson.__geo_interface__)
+            gpd.read_file(io.StringIO(jsontext)).to_file(**ogrgmt_kwargs)
```
weiji14, line 159: Why is codecov reporting that these two lines are not covered by tests? Also, should we temporarily set `engine='fiona'` here to prevent `ValueError: The 'schema' argument is not supported with the 'pyogrio' engine` in the GMT Dev Tests?

```suggestion
            jsontext = json.dumps(geojson.__geo_interface__)
            gpd.read_file(filename=io.StringIO(jsontext), engine="fiona").to_file(**ogrgmt_kwargs)
```

seisman: > Also, should we temporarily set `engine='fiona'` here to prevent `ValueError: The 'schema' argument is not supported with the 'pyogrio' engine` in the GMT Dev Tests?

It won't work. `fiona` is not installed in the GMT Dev Tests workflow, because it's no longer a required dependency for geopandas 1.x.

seisman: > Why is codecov reporting that these two lines are not covered by tests?

Not sure what happened, but now the coverage reports look good.

weiji14: > > Also, should we temporarily set `engine='fiona'` here to prevent `ValueError: The 'schema' argument is not supported with the 'pyogrio' engine` in the GMT Dev Tests?
> 
> It won't work. `fiona` is not installed in the GMT Dev Tests workflow, because it's no longer a required dependency for geopandas 1.x.

I meant to install `fiona` temporarily on GMT Dev Tests, so that the test pass for now with the change, but we'll need to look at how to handle this except-branch without `fiona` before PyGMT v0.13.0 is released. Also wondering if we should open an issue upstream to `geopandas` on what would be the recommended way forward if the 'schema' argument is not supported with `pyogrio`.

seisman: It turns out `pyogrio` is "smarter" than fiona. 

In https://github.com/GenericMappingTools/pygmt/issues/2497#issuecomment-1614418132, I mentioned that 

> I thought world["pop_est_int"] = world.pop_est_round.astype("int32") should work, but it doesn't.

Changing the dtype directly doesn't work with fiona, but works well with pyorgio!.

In 2fb635bbf3fb3f92d88105875c62b02507a3f958, I've added the if-test to check if it's geopandas v0.x or v1.x. For geopandas v0.x, we use the old codes which pass the updated `schema` parameter. For geopandas v1.x, we change the column dtype when necessary.

pygmt/helpers/tempfile.py
```diff
@@ -139,29 +141,33 @@ def tempfile_from_geojson(geojson):
...
+                for col, dtype in schema["properties"].items():
+                    if dtype in ("int", "int64"):
+                        overflow = geojson[col].abs().max() > 2**31 - 1
+                        schema["properties"][col] = "float" if overflow else "int32"
+                ogrgmt_kwargs["schema"] = schema
+            else:  # GeoPandas v1.x.
+                # The default engine "pyogrio" doesn't support the 'schema' parameter
+                # but we can change the dtype directly.
+                for col in geojson.columns:
+                    if geojson[col].dtype in ("int", "int64", "Int64"):
```
seisman, line 159: Have to add "Int64" here because:
```
>>> import pandas as pd
>>> pd.Int32Dtype().name
'Int32'
>>> pd.Int64Dtype().name
'Int64'
```

seisman:
- The "Tests" workflow now passes, which means that reading geojson data via io.StringIO works, and the current code works well with geopandas v0.x.
- The "GMT Dev Tests" workflow has 5/5/2 failures on Linux/macOS/Windows, respectively. All these failures are unrelated to GeoPandas/Fiona/pyogrio, so the current code also work well with geopandas v1.x

Some new codes are reported to not covered by tests because these lines of codes are only executed for geopandas v1.x, which is not installed in our "Tests" workflow.

weiji14:
Thanks @seisman for getting this to work! Sorry that I haven't been responsive with the reviews, been travelling too much for conferences in the past few weeks. I'll spend some time this week to catch up.

PR #3254, opened 2024-05-16
Refactor clib to avoid checking GMT version repeatedly and only check once when loading the GMT library

seisman:

In the `Session.__enter__` method, we check if the GMT library version is equal to or newer than the required version. This special method is called when entering the context manager, i.e., `with Session() as lib`. So it means we're repeatedly checking the GMT version.

This PR refactors the clib codes to check the GMT version only once when loading the GMT library.

I expect to see some performance improvements ~~but the benchmarks say no. So I believe the performance improvement should be very little.~~

https://codspeed.io/GenericMappingTools/pygmt/branches/clib/check-version show an average 2% performance improvement.

### DIFF
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -64,6 +64,33 @@ def load_libgmt(lib_fullnames: Iterator[str] | None = None) -> ctypes.CDLL:
     return libgmt
 
 
+def get_gmt_version(libgmt: ctypes.CDLL) -> str:
+    """
+    Get the GMT version string of the GMT shared library.
+
+    Parameters
+    ----------
+    libgmt
+        The GMT shared library.
+
+    Returns
+    -------
+    The GMT version string in *major.minor.patch* format.
+    """
+    func = libgmt.GMT_Get_Version
+    func.argtypes = (
+        ctypes.c_void_p,  # Unused parameter, so it can be None.
+        ctypes.POINTER(ctypes.c_uint),  # major
+        ctypes.POINTER(ctypes.c_uint),  # minor
+        ctypes.POINTER(ctypes.c_uint),  # patch
+    )
+    # The function return value is the current library version as a float, e.g., 6.5.
+    func.restype = ctypes.c_float
+    major, minor, patch = ctypes.c_uint(0), ctypes.c_uint(0), ctypes.c_uint(0)
+    func(None, major, minor, patch)
+    return f"{major.value}.{minor.value}.{patch.value}"
+
+
 def clib_names(os_name: str) -> list[str]:
     """
     Return the name(s) of GMT's shared library for the current operating system.
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -25,14 +25,9 @@ from pygmt.clib.conversion import (
     strings_to_ctypes_array,
     vectors_to_arrays,
 )
-from pygmt.clib.loading import load_libgmt
+from pygmt.clib.loading import get_gmt_version, load_libgmt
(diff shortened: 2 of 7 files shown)


pygmt/clib/session.py
```diff
@@ -92,8 +84,9 @@
     np.timedelta64: "GMT_LONG",
 }
 
-# Load the GMT library outside the Session class to avoid repeated loading.
+
```
seisman, line 87:
```suggestion
# Load the GMT library outside the Session class to avoid repeated loading.
```

pygmt/clib/loading.py
```diff
@@ -64,6 +64,33 @@ def load_libgmt(lib_fullnames: Iterator[str] | None = None) -> ctypes.CDLL:
...
+        ctypes.c_void_p,  # Unused parameter, so it can be None.
+        ctypes.POINTER(ctypes.c_uint),  # major
+        ctypes.POINTER(ctypes.c_uint),  # minor
+        ctypes.POINTER(ctypes.c_uint),  # patch
+    )
+    # The function return value is the current library version as a float, e.g., 6.5.
+    func.restype = ctypes.c_float
+    major, minor, patch = ctypes.c_uint(0), ctypes.c_uint(0), ctypes.c_uint(0)
+    func(None, major, minor, patch)
+    return f"{major.value}.{minor.value}.{patch.value}"
```
weiji14, line 91: Thoughts on returning an instance of [`packaging.version.Version`](https://packaging.pypa.io/en/stable/version.html#packaging.version.Version) instead of a str?

seisman: but `__gmt_version__` should be a string.

weiji14: Mm ok, could also do `str(get_gmt_version())`, but ok to just return a string itself. Just thought it could simplify some of the test code.

pygmt/tests/test_clib.py
```diff
@@ -577,27 +577,24 @@ def mock_defaults(api, name, value):  # noqa: ARG001
...
-        Return an old version.
-        """
-        if name == b"API_VERSION":
-            value.value = b"5.4.3"
-        else:
-            value.value = b"bla"
-        return 0
+    with monkeypatch.context() as mpatch:
+        # Make sure the current GMT version is 6.
+        assert clib.__gmt_version__.split(".")[0] == "6"
```
weiji14, line 588:
```suggestion
        # Make sure the current GMT major version is 6.
        assert clib.__gmt_version__.split(".")[0] == "6"
```

pygmt/tests/test_clib.py
```diff
@@ -577,27 +577,24 @@ def mock_defaults(api, name, value):  # noqa: ARG001
...
+
+        # Should raise an exception when pygmt.clib is loaded/reloaded.
         with pytest.raises(GMTVersionError):
-            with lib:
-                assert lib.info["version"] != "5.4.3"
-    # Make sure the session is closed when the exception is raised.
-    with pytest.raises(GMTCLibNoSessionError):
-        assert lib.session_pointer
+            importlib.reload(clib)
+        assert clib.__gmt_version__ == "5.4.3"  # Make sure it's old version.
```
weiji14, line 600:
```suggestion
        assert clib.__gmt_version__ == "5.4.3"  # Make sure it's still the old version
```

weiji14 (approved):
Just two tiny suggestions. Having the code spread across `clib/__init__.py`, `clib/loading.py` and `clib/session.py` is a little confusing, but seems to make sense after going through it.

Issue #2646, opened 2023-08-27
Figure.coast: No output of "+l|L" of "dcw" parameter ("E") when using an IDE


yvonnefroehlich:
### Description of the problem

_First mentioned in https://github.com/GenericMappingTools/pygmt/pull/2428#issuecomment-1536777780_.

In the docstring for the `dcw` parameter of [`pygmt.Figure.coast`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.coast.html) we say:
> Append +l|+L to =continent to only list countries in that continent; repeat if more than one continent is requested.

I tried this, but so far I am not sure how it should work. I only get an output when running the python script or code directly in a terminal. When using an IDE (e.g., Spyder, JupyterLab) there is no output neither in the console nor in the terminal.
Further, this part of the docstring is not precise, as **+l** gives the codes for the countries, but **+L** gives the codes for the states/territories.

### Minimal Complete Verifiable Example

```python
import pygmt

fig = pygmt.Figure()
fig.coast(dcw="=NA+l")
fig.coast(dcw="=NA+L")
```

### Full error message

```python-traceback
No error message occurs. 
No output occurs when running the code within an IDE (e.g., Spyder, JupyterLab).
```

### System information

```bash
PyGMT information:
  version: v0.9.1.dev132
System information:
  python: 3.11.4 | packaged by conda-forge | (main, Jun 10 2023, 17:59:51) [MSC v.1935 64 bit (AMD64)]
  executable: C:\ProgramData\Anaconda3\envs\pygmt_env_dev\python.exe
  machine: Windows-10-10.0.19045-SP0
Dependency information:
  numpy: 1.24.3
  pandas: 2.0.2
  xarray: 2023.1.1.dev17
  netCDF4: 1.6.2
  packaging: 23.1
  contextily: 1.3.0
  geopandas: 0.13.2
  IPython: 8.14.0
  rioxarray: 0.14.1
  ghostscript: 9.54.0
GMT library information:
  binary version: 6.4.0
  cores: 4
  grid layout: rows
  image layout: 
  library path: C:/ProgramData/Anaconda3/envs/pygmt_env_dev/Library/bin/gmt.dll
  padding: 2
  plugin dir: C:/ProgramData/Anaconda3/envs/pygmt_env_dev/Library/bin/gmt_plugins
  share dir: C:/Program Files (x86)/gmt6/share
  version: 6.4.0
```

seisman:
[...]

PR #3272, opened 2024-05-25
pygmt.dataset.load_*: Add type hints for the 'region' parameter

seisman:
 
In the previous version, `region` can be either a list or a string, but a string like `region="0/10/0/20"` is not Pythonic and a tuple like `region=(0, 10, 0, 20)` is also accepted.

This PR adds type hints to the `region` parameter of `load_*` functions. The `region` parameter must be a `Sequence` type (list, tuple, or array). A string type is still supported but no longer documented/recommended.

**Preview**: https://pygmt-dev--3272.org.readthedocs.build/en/3272/api/generated/pygmt.datasets.load_earth_age.html#pygmt.datasets.load_earth_age

### DIFF
--- a/pygmt/datasets/earth_age.py
+++ b/pygmt/datasets/earth_age.py
@@ -5,20 +5,19 @@ and load as :class:`xarray.DataArray`.
 The grids are available in various resolutions.
 """
 
+from collections.abc import Sequence
 from typing import Literal
 
 from pygmt.datasets.load_remote_dataset import _load_remote_dataset
-from pygmt.helpers import kwargs_to_strings
 
 __doctest_skip__ = ["load_earth_age"]
 
 
-@kwargs_to_strings(region="sequence")
 def load_earth_age(
     resolution: Literal[
         "01d", "30m", "20m", "15m", "10m", "06m", "05m", "04m", "03m", "02m", "01m"
     ] = "01d",
-    region=None,
+    region: Sequence[float] | None = None,
     registration: Literal["gridline", "pixel"] = "gridline",
 ):
     r"""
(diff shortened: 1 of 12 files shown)


seisman:
Will apply the same changes to other `load_*` functions if @GenericMappingTools/pygmt-maintainers agree with the changes.

pygmt/datasets/earth_age.py
```diff
@@ -5,20 +5,19 @@
...
+from collections.abc import Sequence
 from typing import Literal
 
 from pygmt.datasets.load_remote_dataset import _load_remote_dataset
-from pygmt.helpers import kwargs_to_strings
 
 __doctest_skip__ = ["load_earth_age"]
 
 
-@kwargs_to_strings(region="sequence")
```
seisman, line 16: This line can be removed because `region` is passed to `_load_remote_dataset` and in that function `region` can be a sequence.

michaelgrund (approved):

yvonnefroehlich (commented):
Passing a list or a tuple is definitely more Phytonic.
But I am wondering how we want to handle arguments containing ISO codes and arguments extending a region via appending `+r`, e.g., `lon_min/lon_ma/lat_min/lat_max+rincrement` or `ISO_code+rincrement`. Here, we still need a string input.

seisman:
> But I am wondering how we want to handle arguments containing ISO codes and arguments extending a region via appending `+r`, e.g., `lon_min/lon_ma/lat_min/lat_max+rincrement` or `ISO_code+rincrement`. Here, we still need a string input.

That's a good point.

> `lon_min/lon_ma/lat_min/lat_max+rincrement`

I don't think GMT's -R option has such syntax (see https://docs.generic-mapping-tools.org/dev/gmt.html#r-full). `-Rxlleft/ylleft/xuright/yuright+r` is close to this one, but it's mainly for specifying a region for oblique projections. So, I guess the syntax doesn't make sense here.

> `ISO_code+rincrement`

This one should be supported currently. We probably should implement the ideas in https://github.com/GenericMappingTools/pygmt/issues/2646 and then people should use the proposed to `dcw` module to get the region before passing to the `region` parameter.

Anyway, I'll add the string type back.

yvonnefroehlich:
> lon_min/lon_ma/lat_min/lat_max+rincrement

>> I don't think GMT's -R option has such syntax (see https://docs.generic-mapping-tools.org/dev/gmt.html#r-full). -Rxlleft/ylleft/xuright/yuright+r is close to this one, but it's mainly for specifying a region for oblique projections. So, I guess the syntax doesn't make sense here.

You are totally right; such a syntax is not available, and your guess is what I was actually thinking about. I must have gotten confused during writing - I apologize for the confusion!

Issue #3301, opened 2024-06-25
geopandas 1.0.0 or pyogrio 0.9.0 doesn't support OGR_GMT output?


seisman:
[geopandas v1.0.0](https://github.com/geopandas/geopandas/releases/tag/v1.0.0) was released a few hours ago and now our Docs workflow fails to build (https://github.com/GenericMappingTools/pygmt/actions/runs/9559533732/job/26634813440) which is not caught in our "GMT Dev Tests" workflow (why?):
```
Extension error:
Here is a summary of the problems encountered when running the examples:

Unexpected failing examples:

    ../examples/get_started/04_table_inputs.py failed leaving traceback:

    Traceback (most recent call last):
      File "/home/runner/work/pygmt/pygmt/examples/get_started/04_table_inputs.py", line 111, in <module>
        fig.plot(data=gdf, style="c0.2c", fill="purple")
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/pygmt/helpers/decorators.py", line 609, in new_module
        return module_func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/pygmt/helpers/decorators.py", line 773, in new_module
        return module_func(*bound.args, **bound.kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/pygmt/src/plot.py", line 258, in plot
        with lib.virtualfile_in(
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/contextlib.py", line 137, in __enter__
        return next(self.gen)
               ^^^^^^^^^^^^^^
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/pygmt/helpers/tempfile.py", line 164, in tempfile_from_geojson
        geojson.to_file(**ogrgmt_kwargs)
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/geopandas/geodataframe.py", line 1536, in to_file
        _to_file(self, filename, driver, schema, index, **kwargs)
      File "/home/runner/micromamba/envs/pygmt/lib/python3.12/site-packages/geopandas/io/file.py", line 686, in _to_file
[...]

PR #3305, opened 2024-06-28
CI: Unpin geopandas and rasterio after the fix in #3672

seisman:
Revert changes in #3304 so that we can track the issue #3301.

Edit on 2024/12/03: Also revert #3668


examples/gallery/lines/roads.py
```diff
@@ -14,8 +14,10 @@
 # %%
 import geopandas as gpd
 import pygmt
+import pyogrio
 
 # Read shapefile data using geopandas
+pyogrio.core._register_drivers()
 gdf = gpd.read_file(
```
weiji14, line 21: Found this issue https://github.com/geopandas/pyogrio/issues/144 that mentions an error like `pyogrio.errors.DataSourceError: Could not obtain driver: GPKG (check that it was installed correctly into GDAL)` was due to drivers not being loaded properly. This should have been fixed in https://github.com/geopandas/pyogrio/pull/145, but unsure why we're still hitting into this in the sphinx build and not in the regular tests.

examples/gallery/lines/roads.py
```diff
@@ -14,8 +14,10 @@
 # %%
 import geopandas as gpd
 import pygmt
+import pyogrio
```
seisman, line 17: Can we move it to `conf.py` so that we don't have to repeat it in multiple gallery examples?

weiji14: Just tried locally, no it doesn't work if I put it in `doc/conf.py`.

seisman: I'm hesitating to merge this PR because (1) users may be confused about `pyogrio.core._register_drivers()`; (2) Users may still use geopandas v0.x and don't have pyogrio installed.

seisman: In commit e15462f956c5ee11d0cf01efe14c4e79edb0f4bf, I use sphinx-gallery's [`sphinx_gallery_start_ignore`/`sphinx_gallery_end_ignore`](https://sphinx-gallery.github.io/stable/configuration.html#hiding-code-blocks) flags to hide the two lines of codes in the documentation.

**Preview**: https://pygmt-dev--3305.org.readthedocs.build/en/3305/gallery/lines/roads.html#sphx-glr-gallery-lines-roads-py
https://pygmt-dev--3305.org.readthedocs.build/en/3305/get_started/04_table_inputs.html#sphx-glr-get-started-04-table-inputs-py

weiji14:
@seisman, maybe we should retry this again with `rasterio=1.4.3`, xref https://github.com/GenericMappingTools/pygmt/issues/3301#issuecomment-2513054819.

weiji14 (approved):

PR #3307, opened 2024-07-02
Add pytest conftest.py file to suppress warnings with NumPy 2.0

seisman:
Workaround from https://github.com/GenericMappingTools/pygmt/issues/2628#issuecomment-1688666672.

Fixes https://github.com/GenericMappingTools/pygmt/issues/2628.

### DIFF
--- a/pygmt/conftest.py
+++ b/pygmt/conftest.py
@@ -8,4 +8,4 @@ from packaging.version import Version
 # Keep this until we require numpy to be >=2.0
 # Address https://github.com/GenericMappingTools/pygmt/issues/2628.
 if Version(np.__version__) >= Version("2.0.0.dev0+git20230726"):
-    np.set_printoptions(legacy="1.25")
+    np.set_printoptions(legacy="1.25")  # type: ignore[arg-type]
--- a/pygmt/conftest.py
+++ b/pygmt/conftest.py
@@ -1,0 +1,10 @@
+"""
+conftets.py for pytest.
+"""
+import numpy as np
+from packaging.version import Version
+
+# Keep this until we require numpy to be >=2.0
+# Address https://github.com/GenericMappingTools/pygmt/issues/2628.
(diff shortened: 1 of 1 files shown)


michaelgrund (approved):

pygmt/conftest.py
```diff
@@ -0,0 +1,11 @@
+"""
+conftets.py for pytest.
```
weiji14, line 2:
```suggestion
conftest.py for pytest.
```

pygmt/conftest.py
```diff
@@ -0,0 +1,11 @@
...
+"""
+
+import numpy as np
+from packaging.version import Version
+
+# Keep this until we require numpy to be >=2.0
+# Address https://github.com/GenericMappingTools/pygmt/issues/2628.
```
weiji14, line 9: What do we do though once we require numpy>2.0 in June 2026?

seisman: At that time, this file can be removed and we just need to update the doctests to match the numpy 2.0 behavior.

pygmt/conftest.py
```diff
@@ -0,0 +1,11 @@
...
+conftest.py for pytest.
+"""
+
+import numpy as np
+from packaging.version import Version
+
+# Keep this until we require numpy to be >=2.0
+# Address https://github.com/GenericMappingTools/pygmt/issues/2628.
+if Version(np.__version__) >= Version("2.0.0.dev0+git20230726"):
+    np.set_printoptions(legacy="1.25")
```
seisman, line 11: https://github.com/numpy/numpy/blob/1d49c7f7ff527c696fc26ab2278ad51632a66660/numpy/_core/arrayprint.py#L91
`legacy` can be `1.25` but it's not well documented at https://numpy.org/doc/stable/reference/generated/numpy.set_printoptions.html#numpy.set_printoptions, so we need to ignore the type error:

```suggestion
    np.set_printoptions(legacy="1.25")  # type: ignore[arg-type]
```

PR #3310, opened 2024-07-03
Support non-ASCII characters in ISO-8859-x charsets

seisman:
This PR enhances PyGMT to support non-ASCII characters in ISO-8859-x charsets (Refer to https://en.wikipedia.org/wiki/ISO/IEC_8859-1, https://en.wikipedia.org/wiki/ISO/IEC_8859-2 and more for available characters in each encoding). Related to #2204.

In GMT, the default encoding is `ISOLatin1+` (or `Standard+` but it's not supported in PyGMT yet). To use non-ASCII characters in ISO-8859-x, we need to use `--PS_CHAR_ENCODING=<encoding>` according to https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-PS_CHAR_ENCODING. Please note that we can only use one encoding in a single GMT module call, which means mixing characters from different encodings will cause trouble. So, the implementation is simple. We just need to check the argument string or `Figure.text`'s text string and determine the charset encoding to use, and then add `--PS_CHAR_ENCODING=<encoding>` when calling modules.

This PR does a few things:

1. Add the mapping dictionaries of all ISO-8859-x charsets to `pygmt/encodings.py`
2. Add `check_encoding` to check the encoding of a string.
3. Add the new `encoding` parameter to `non_ascii_to_otcal` so that we know the encoding to use when converting non-ascii.
4. Update to make PyGMT arguments and `Figure.text`'s `text` parameter support ISO-8859-x charsets.

**Preview** : https://pygmt-dev--3310.org.readthedocs.build/en/3310/techref/encodings.html

---
Here is an example showing how it works with non-ASCII characters in [ISO-8859-4](https://en.wikipedia.org/wiki/ISO/IEC_8859-4). The same script is added as a test.
```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=["WSEN+tAāáâãäåB"])
fig.text(position="TL", text="position-text:1ÉĘËĖ2")
fig.text(x=1, y=1, text="xytext:1éęëė2")
fig.text(x=[5, 5], y=[3, 5], text=["xytext1:ųúûüũūαζ∆❡", "xytext2:íîī∑π∇✉"])
fig.show()
```
The output is 
image

---

The script below shows all the characters in different ISO-8859-x encodings. I don't check them all carefully, but most look correct. The exceptions are [ISO-8859-6](https://en.wikipedia.org/wiki/ISO/IEC_8859-6), [ISO-8859-8](https://en.wikipedia.org/wiki/ISO/IEC_8859-8), and [ISO-8859-11](https://en.wikipedia.org/wiki/ISO/IEC_8859-11). They don't work with standard fonts (https://docs.generic-mapping-tools.org/dev/gmt.conf.html#term-PS_CHAR_ENCODING) but I think it's OK because whoever wants to use these characters should have the fonts installed. Some characters like `Ṁ` in ISO-8859-14 are missing in the figure below and they also can't be shown correctly in GMT CLI. Looking at the charset definition files in the GMT source codes, the character `Ṁ` has a PS name like `/uni1E40` (https://github.com/GenericMappingTools/gmt/blob/0584d620d27de5ed93a90e4b0dc56a2edb2d568a/src/PSL_ISO-8859-14.h#L24C43-L24C43), but it doesn't seem like a valid PS name. So, it must be upstream issues if any.

```python
import pygmt
from pygmt.encodings import charset

fig = pygmt.Figure()
with fig.subplot(ncols=5, nrows=3, subsize=("10c", "15c"), margins=0.5):

    for i in range(1, 17):
        if i == 12:
            continue
        encoding = f"ISO-8859-{i}"
        chars = "".join(charset[encoding].values())
        fig.basemap(region=[0, 16, 0, 13], projection="X?/-?", frame=[0, f"+t{encoding}"], panel=True)
        for j in range(0, len(chars), 16):
            fig.text(x=1, y=int(j/16) + 1, text=chars[j:j+16], font="20p", justify="ML")
fig.savefig("encodings.png")
```

encodings

### DIFF
--- a/doc/techref/encodings.md
+++ b/doc/techref/encodings.md
@@ -110,7 +110,7 @@ the Unicode character set.
 ## ISO/IEC 8859
 
 GMT also supports the ISO/IEC 8859 standard for 8-bit character encodings. Refer to
-https://en.wikipedia.org/wiki/ISO/IEC_8859 for descriptions of the different parts of the standard.
+<https://en.wikipedia.org/wiki/ISO/IEC_8859> for descriptions of the different parts of the standard.
 
 For a list of the characters in each part of the standard, refer to the following links:
 
--- a/doc/techref/encodings.md
+++ b/doc/techref/encodings.md
@@ -114,18 +114,18 @@ https://en.wikipedia.org/wiki/ISO/IEC_8859 for descriptions of the different par
 
 For a list of the characters in each part of the standard, refer to the following links:
 
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-1
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-2
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-3
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-4
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-5
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-6
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-7
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-8
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-9
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-10
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-11
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-13
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-14
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-15
-- https://en.wikipedia.org/wiki/ISO/IEC_8859-16
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-1>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-2>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-3>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-4>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-5>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-6>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-7>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-8>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-9>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-10>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-11>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-13>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-14>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-15>
+- <https://en.wikipedia.org/wiki/ISO/IEC_8859-16>
--- a/doc/techref/encodings.md
+++ b/doc/techref/encodings.md
@@ -1,14 +1,12 @@
 # Supported Encodings and Non-ASCII Characters
 
-GMT supports a number of encodings and each encoding contains a set of ASCII and non-ASCII
-characters. Below are some of the most common encodings and characters that are supported.
+GMT supports a number of encodings and each encoding contains a set of ASCII and
+non-ASCII characters. In PyGMT, you can use any of these ASCII and non-ASCII characters
+in arguments and text strings. When using non-ASCII characters in PyGMT, the easiest way
+is to copy and paste the character from the encoding tables below.
 
-In PyGMT, you can use any of these ASCII and non-ASCII characters in arguments and text
-strings. When using non-ASCII characters in PyGMT, the easiest way is to copy and paste
-the character from the tables below.
-
-**Note**: The special character &#xfffd; (REPLACEMENT CHARACTER) is used to indicate that
-the character is not defined in the encoding.
+**Note**: The special character &#xfffd; (REPLACEMENT CHARACTER) is used to indicate
+that the character is not defined in the encoding.
 
 ## Adobe ISOLatin1+ Encoding
 
@@ -110,7 +108,8 @@ the Unicode character set.
 ## ISO/IEC 8859
 
 GMT also supports the ISO/IEC 8859 standard for 8-bit character encodings. Refer to
-<https://en.wikipedia.org/wiki/ISO/IEC_8859> for descriptions of the different parts of the standard.
+<https://en.wikipedia.org/wiki/ISO/IEC_8859> for descriptions of the different parts of
+the standard.
 
 For a list of the characters in each part of the standard, refer to the following links:
 
--- a/doc/techref/encodings.md
+++ b/doc/techref/encodings.md
@@ -109,7 +109,7 @@ the Unicode character set.
 
 ## ISO/IEC 8859
 
-GMT also supports the ISO/IEC 8859 standars for 8-bit character encodings. Refer to
+GMT also supports the ISO/IEC 8859 standard for 8-bit character encodings. Refer to
 https://en.wikipedia.org/wiki/ISO/IEC_8859 for descriptions of the different parts of the standard.
 
 For a list of the characters in each part of the standard, refer to the following links:
--- a/doc/techref/encodings.md
+++ b/doc/techref/encodings.md
@@ -106,3 +106,26 @@ the Unicode character set.
 | **\35x** | &#x27a8; | &#x27a9; | &#x27aa; | &#x27ab; | &#x27ac; | &#x27ad; | &#x27ae; | &#x27af; |
 | **\36x** | &#xfffd; | &#x27b1; | &#x27b2; | &#x27b3; | &#x27b4; | &#x27b5; | &#x27b6; | &#x27b7; |
 | **\37x** | &#x27b8; | &#x27b9; | &#x27ba; | &#x27bb; | &#x27bc; | &#x27bd; | &#x27be; | &#xfffd; |
+
+## ISO/IEC 8859
+
+GMT also supports the ISO/IEC 8859 standars for 8-bit character encodings. Refer to
+https://en.wikipedia.org/wiki/ISO/IEC_8859 for descriptions of the different parts of the standard.
+
+For a list of the characters in each part of the standard, refer to the following links:
+
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-1
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-2
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-3
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-4
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-5
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-6
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-7
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-8
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-9
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-10
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-11
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-13
+- https://en.wikipedia.org/wiki/ISO/IEC_8859-14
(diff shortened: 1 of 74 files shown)


doc/techref/encodings.md
```diff
@@ -106,3 +106,26 @@ the Unicode character set.
 | **\35x** | &#x27a8; | &#x27a9; | &#x27aa; | &#x27ab; | &#x27ac; | &#x27ad; | &#x27ae; | &#x27af; |
 | **\36x** | &#xfffd; | &#x27b1; | &#x27b2; | &#x27b3; | &#x27b4; | &#x27b5; | &#x27b6; | &#x27b7; |
 | **\37x** | &#x27b8; | &#x27b9; | &#x27ba; | &#x27bb; | &#x27bc; | &#x27bd; | &#x27be; | &#xfffd; |
+
+## ISO/IEC 8859
+
+GMT also supports the ISO/IEC 8859 standars for 8-bit character encodings. Refer to
```
michaelgrund, line 112:
```suggestion
GMT also supports the ISO/IEC 8859 standard for 8-bit character encodings. Refer to
```

examples/gallery/images/rgb_image.py
```diff
@@ -38,6 +38,6 @@
         grid=image,
         # Use a map scale where 1 cm on the map equals 1 km on the ground
         projection="x1:100000",
-        frame=[r"WSne+tL@!a¯hain@!a¯, Hawai`i on 9 Aug 2023", "af"],
+        frame=[r"WSne+tLāhainā, Hawai`i on 9 Aug 2023", "af"],
```
seisman, line 41:
```suggestion
        frame=[r"WSne+tL@!a¯hain@!a¯, Hawai`i on 9 Aug 2023", "af"],
```

pygmt/encodings.py
```diff
@@ -129,3 +132,12 @@
...
 )
+
+# ISO-8859-x charsets and x can be 1-11, 13-16.
+for i in range(1, 17):
+    if i == 12:  # ISO-8859-2 was abandoned.
+        continue
+    charset[f"ISO-8859-{i}"] = {
+        code: codecs.decode(bytes([code]), f"iso8859_{i}", errors="replace")
+        for code in [*range(0o040, 0o200), *range(0o240, 0o400)]
+    }
```
yvonnefroehlich, line 143:
```suggestion
for i in range(1, 17):
    if i == 12:  # ISO-8859-12 was abandoned.
        continue
    charset[f"ISO-8859-{i}"] = {
        code: codecs.decode(bytes([code]), f"iso8859_{i}", errors="replace")
        for code in [*range(0o040, 0o200), *range(0o240, 0o400)]
    }
```

pygmt/encodings.py
```diff
@@ -22,8 +22,11 @@
 - Adobe Symbol: https://en.wikipedia.org/wiki/Symbol_(typeface)
 - Zapf Dingbats: https://en.wikipedia.org/wiki/Zapf_Dingbats
```
yvonnefroehlich, line 23: Hm. This wikipedia article uses a white space in the title, but everywhere else in this PR we use no white space.

```suggestion
- ZapfDingbats: https://en.wikipedia.org/wiki/Zapf_Dingbats
```

pygmt/helpers/utils.py
```diff
@@ -205,7 +205,55 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
...
 
 
-def non_ascii_to_octal(argstr: str) -> str:
+def check_encoding(argstr: str) -> str:
+    """
+    Check the charset encoding of a string.
+
+    All characters in the string must be in a single charset encoding, otherwise the
+    default ISOLatin1+ encoding is returned. Characters in the Symbol and ZapfDingbats
+    fonts are also checked because they're independent on the setting of charset.
```
yvonnefroehlich, line 214:
```suggestion
    All characters in the string must be in the same charset encoding, otherwise the
    default ISOLatin1+ encoding is returned. Characters in the Symbol and ZapfDingbats
    fonts are also checked because they're independent on the setting of the charset.
```

seisman:
In this PR, the gallery example (`examples/gallery/images/rgb_image.py`) is not updated (see https://github.com/GenericMappingTools/pygmt/issues/2204#issuecomment-1692853711 for the initial issue), mainly because backtick ` is not rendered like `‘` in ISO-8859-4 encoding, and `‘` doesn't exist in ISO-8859-4.

```
frame=[r"WSne+tLāhainā, Hawai`i on 9 Aug 2023", "af"]
```
map

```
frame=[r"WSne+tLāhainā, Hawai'i on 9 Aug 2023", "af"]
```
map2

```
frame=[r"WSne+tLāhainā, Hawai‘i on 9 Aug 2023", "af"]
```
`ā` is in ISO-8859-4 but `‘` doesn't, so the encoding reverts back to the default ISOLatin1+ encoding:
map3

pygmt/helpers/utils.py
```diff
@@ -205,7 +205,55 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
...
+        charset["ZapfDingbats"].values()
+    )
+    for encoding in ["ISOLatin1+"] + [f"ISO-8859-{i}" for i in range(1, 17)]:
+        if encoding == "ISO-8859-12":  # ISO-8859-12 was abandoned. Skip it.
+            continue
+        if all(c in (set(charset[encoding].values()) | adobe_chars) for c in argstr):
+            return encoding
+    # Return the "ISOLatin1+" encoding if the string contains characters from multiple
+    # charset encodings or contains characters that are not in any charset encoding.
+    return "ISOLatin1+"
```
seisman, line 253: The function returns `ISOLatin1+`/`ISO-8859-1`/.../`ISO-8859-16`, but Python's standard encodings are `latin_1`/`iso8859_1`/...`iso8859_16` (https://docs.python.org/3/library/codecs.html#standard-encodings). They're inconsistent, but using names like `ISOLatin1+` can greatly simplify the codes.

pygmt/helpers/utils.py
```diff
@@ -205,7 +205,55 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
...
+    'ISOLatin1+'
+    >>> check_encoding("12AB±β①②")  # Characters in ISOLatin1+
+    'ISOLatin1+'
+    >>> check_encoding("12ABāáâãäåβ①②")  # Characters in ISO-8859-4
+    'ISO-8859-4'
+    >>> check_encoding("12ABŒā")  # Mix characters in ISOLatin1+ (Œ) and ISO-8859-4 (ā)
+    'ISOLatin1+'
+    >>> check_encoding("123AB中文")  # Characters not in any charset encoding
+    'ISOLatin1+'
+    """
```
seisman, line 238: In addition to the GMT-supported encodings, I'm thinking if we should return `ascii` (the name `ascii` comes from https://docs.python.org/3/library/codecs.html#standard-encodings) if the input string only contains ASCII characters, e.g.,
```
    if all(32 <= ord(c) <= 126 for c in argstr):
        return "ascii"
```
Since in most cases, the arguments and the text string contain ASCII characters only. When `ascii` encoding is detected, we no longer need to apply `non_ascii_to_octal` to the strings, which may improve the performance for most cases.

seisman: Done in 2d01f6b8fecb3725bffe8e51531629464bbf7f90.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for reviews.

seisman:
I'll merge this PR in 24 hours if there are no further comments.

pygmt/helpers/utils.py
```diff
@@ -192,7 +192,58 @@ def data_kind(
     return kind
 
 
-def non_ascii_to_octal(argstr: str) -> str:
+def check_encoding(argstr: str) -> str:
```
weiji14, line 195: Would it be too much to have a typehint like this?

```suggestion
def check_encoding(argstr: str) -> Literal[
    "ascii",
    "ISOLatin1+",
    "ISO-8859-1",
    "ISO-8859-2",
    "ISO-8859-3",
    "ISO-8859-4",
    "ISO-8859-5",
    "ISO-8859-6",
    "ISO-8859-7",
    "ISO-8859-8",
    "ISO-8859-9",
    "ISO-8859-10",
    "ISO-8859-11",
    "ISO-8859-13",
    "ISO-8859-14",
    "ISO-8859-15",
    "ISO-8859-16",
    "ISO-8859-17",
]:
```

pygmt/helpers/__init__.py
```diff
@@ -19,6 +19,7 @@
     args_in_kwargs,
     build_arg_list,
     build_arg_string,
+    check_encoding,
```
weiji14, line 22: Thoughts on using `_check_encoding` to keep this function more private? I know we don't document `pygmt.helpers.utils` in the API docs, but want to avoid users from thinking that this function is somewhat public if there's no leading underscore.

pygmt/helpers/utils.py
seisman: I'm thinking about that too. Actually the same types can be used in typing hints the `encoding` parameter of the `non_ascii_to_octal` function. So my initial plan is to define a generic type
```
type Encodings = Literal[
    "ascii",
    "ISOLatin1+",
    "ISO-8859-1",
    "ISO-8859-2",
    "ISO-8859-3",
    "ISO-8859-4",
    "ISO-8859-5",
    "ISO-8859-6",
    "ISO-8859-7",
    "ISO-8859-8",
    "ISO-8859-9",
    "ISO-8859-10",
    "ISO-8859-11",
    "ISO-8859-13",
    "ISO-8859-14",
    "ISO-8859-15",
    "ISO-8859-16",
]
```
and then use it like:
```
def check_encoding(argstr: str) -> Encodings:
def non_ascii_to_octal(argstr: str, encoding: Encodings = "ISOLatin1+") -> str:
```
but mypy v1.10.0 complains
```
pygmt/helpers/utils.py:21: error: PEP 695 type aliases are not yet supported  [valid-type]
```
mypy v1.11.0 started to support PEP 695 but it was released just a few days ago (https://mypy-lang.blogspot.com/2024/07/mypy-111-released.html) and the PEP 695 support is still experimental.

seisman: Done in 7607c7efb0b2442b1647239d51e475ebf62284e2.

pygmt/helpers/__init__.py
seisman: Done in 6728856a90ea66a96a543c78981fbb8fad342bce.

pygmt/helpers/utils.py
```diff
@@ -192,17 +192,113 @@ def data_kind(
     return kind
 
 
-def non_ascii_to_octal(argstr: str) -> str:
+def _check_encoding(
```
weiji14, line 195: Move this private function near the top (together with `_validate_data_input`)?

pygmt/helpers/utils.py
```diff
@@ -324,11 +426,22 @@ def build_arg_list(
...
-            gmt_args.extend(non_ascii_to_octal(f"-{key}{_value}") for _value in value)
+            gmt_args.extend(f"-{key}{_value}" for _value in value)
         else:
-            gmt_args.append(non_ascii_to_octal(f"-{key}{value}"))
+            gmt_args.append(f"-{key}{value}")
+
+    # Convert non-ASCII characters (if any) in the arguments to octal codes
+    encoding = _check_encoding("".join(gmt_args))
+    if encoding != "ascii":
+        gmt_args = [non_ascii_to_octal(arg, encoding=encoding) for arg in gmt_args]
```
weiji14, line 436: This block assumes that there will only be one encoding type? To support multiple encodings in different arguments, maybe change it to something like this (untested):

```suggestion
    # Convert non-ASCII characters (if any) in the arguments to octal codes
    for i, arg in enumerate(gmt_args):
        encoding = _check_encoding("".join(gmt_args))
        if encoding != "ascii":
            gmt_args[i] = non_ascii_to_octal(arg, encoding=encoding)
```

But this can be done in a separate PR perhaps. I'm not sure how common it is to mix encodings in different arguments.

weiji14 (approved):
Thanks @seisman, haven't checked too closely, but don't want to hold this up for too long. Made one comment about mixed encoding support in different arguments, but we can discuss that #2204 and decide whether to open a separate PR for it.

pygmt/helpers/utils.py
seisman: > To support multiple encodings in different arguments, maybe change it to something like this (untested):

It's not possible in GMT. For each GMT module call, we can only pass `--PS_CHAR_ENCODING` once, so it means that all the arguments in a GMT call must be in the same encoding. 

For `Figure.text`, the `text` string and any arguments also must use the same encoding.

seisman: If people use mixed encodings in different arguments, the default `ISOLatin1+` encoding is used.

Issue #2741, opened 2023-10-12
Use ruff to improve code quality


seisman:
It seems [ruff](https://github.com/astral-sh/ruff) has been very hot recently. The README says it can replace flake8 and pylint, and more importantly, it's 10-100x faster.

Someone interested may try it and tell us how it works for the PyGMT project.

weiji14:
Opened a PR to replace `flakeheaven` and `isort` with `ruff` at #2747, so that we can move ahead with the Python 3.12 support.

It should also be possible to replace `pylint` with `ruff` (see https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-pylint), which would:
- Pro:
  - Allow us to remove `make lint` (which is slow) from the Makefile (i.e. only have `make check` and `make format`)
  - Bonus is that `ruff` can actually automatically fix some pylint rules (see those marked with 🛠️ at https://docs.astral.sh/ruff/rules/#pylint-pl)
- Con: Not all pylint rules are applied (see https://github.com/astral-sh/ruff/issues/970), but I'm not sure if we even need to check all of them...

So should we replace `pylint` with `ruff`? Vote :+1: for yes, :-1: for no. If everyone is generally in favour, we can open a PR after #2747 is merged.

seisman:
> So should we replace `pylint` with `ruff`?

Instead of replacing `pylint` with `ruff`, I think we can use both to see if there are any linting issues reported by pylint but not detected by `ruff`. We can then decide if we want to remove `pylint` before releasing v0.11.0. So I propose to have two separate PRs:

- [ ] Configure `ruff` as a linting tool
- [ ] Remove `pylint` before v0.11.0

seisman:
**Last updated on 2024/11/29**

ruff supports many groups of rules (https://docs.astral.sh/ruff/rules/). I think we should go through the available rules and apply the rules that make sense to our project.

ruff v0.16.0 (released on 2026/07) provides the following linters/rules:
```
ruff linter | sort -k1 | awk '{printf("- [ ] [`%s`](https://docs.astral.sh/ruff/rules/#%s-%s): %s\n"), $1, tolower($2), tolower($1), $2}' 
```

- [x] [`A`](https://docs.astral.sh/ruff/rules/#flake8-builtins-a): flake8-builtins #2882
- [ ] ~~[`AIR`](https://docs.astral.sh/ruff/rules/#airflow-air): Airflow~~ [unrelated to airflow]
[...]

PR #3317, opened 2024-07-08
Enable ruff's literal-membership (PLR6201) rule and fix violations

weiji14:

Enable ruff's [literal membership (PLR6201) rule](https://docs.astral.sh/ruff/rules/literal-membership/) to ensure a set literal is used when testing for membership. Note that this is a preview mode feature.

This offers a bit of a speed/performance improvement, as Python's peephole optimizer will optimize set membership tests (by converting the set to a frozenset), making it faster than tuple or list membership tests.

Example speed tests on my machine, about a 4.5x speedup using set `{}` compared to tuple `()` or list `[]`:

```
$ python -m timeit -s 'convention="t"' 'convention in {"a", "c", "m", "d", "z", "p", "x", "y", "t"}'
50000000 loops, best of 5: 9.52 nsec per loop

$ python -m timeit -s 'convention="t"' 'convention in ("a", "c", "m", "d", "z", "p", "x", "y", "t")'
5000000 loops, best of 5: 42.4 nsec per loop

$ python -m timeit -s 'convention="t"' 'convention in ["a", "c", "m", "d", "z", "p", "x", "y", "t"]'
5000000 loops, best of 5: 42.4 nsec per loop
```

References:
- https://docs.python.org/3/whatsnew/3.2.html#optimizations
- https://docs.astral.sh/ruff/rules/literal-membership/
- https://docs.astral.sh/ruff/preview/#selecting-single-preview-rules

Addresses https://github.com/GenericMappingTools/pygmt/issues/2741#issuecomment-2212728325

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_clib_loading.py
+++ b/pygmt/tests/test_clib_loading.py
@@ -131,7 +131,7 @@ class TestLibgmtBrokenLibs:
         # libname is a loaded GMT library
         return self.loaded_libgmt
 
-    @pytest.fixture()
+    @pytest.fixture
     def _mock_ctypes(self, monkeypatch):
         """
         Patch the ctypes.CDLL function.
--- a/pygmt/accessors.py
+++ b/pygmt/accessors.py
@@ -141,7 +141,7 @@ class GMTDataArrayAccessor:
 
     @registration.setter
     def registration(self, value):
-        if value not in (0, 1):
+        if value not in {0, 1}:
             raise GMTInvalidInput(
                 f"Invalid grid registration value: {value}, should be either "
                 "0 for Gridline registration or 1 for Pixel registration."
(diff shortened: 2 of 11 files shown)


pygmt/tests/test_clib_loading.py
```diff
@@ -131,7 +131,7 @@ def _mock_ctypes_cdll_return(self, libname):
         # libname is a loaded GMT library
         return self.loaded_libgmt
 
-    @pytest.fixture()
+    @pytest.fixture
```
weiji14, line 134: Note that this change is to fix a PT001 rule violation that was triggered because of the preview mode setting. In non-preview mode, `@pytest.fixture()` is valid, but in preview mode, `@pytest.fixture()` is valid. See https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/ and https://github.com/astral-sh/ruff/pull/12106.

seisman:
Three geopandas tests start to fail, but not sure if it's related to changes in this PR.

seisman (approved):

PR #3319, opened 2024-07-08
Enable ruff's unspecified-encoding (PLW1514) rule and fix violations

weiji14:

Enable ruff's [unspecified-encoding (PLW1514) rule](https://docs.astral.sh/ruff/rules/unspecified-encoding/) to check for uses of `open` and related calls without an explicit `encoding` argument. Note that this is a preview mode feature.

This lint rule tells us to consider using the `encoding` parameter to enforce a specific encoding. [PEP 597](https://peps.python.org/pep-0597/) recommends using `encoding="locale"` on Python 3.10 and later, though we could also use `encoding="utf-8"` perhaps.

For this PR, we went for setting `encoding="utf-8"`

References:
- https://docs.astral.sh/ruff/rules/unspecified-encoding/
- https://peps.python.org/pep-0597/#prepare-to-change-the-default-encoding-to-utf-8

Addresses https://github.com/GenericMappingTools/pygmt/issues/2741#issuecomment-2212728325

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -247,7 +247,7 @@ def plot(  # noqa: PLR0912
             kwargs["S"] = "s0.2c"
         elif kind == "file" and str(data).endswith(".gmt"):  # OGR_GMT file
             try:
-                with Path(which(data)).open(encoding="locale") as file:
+                with Path(which(data)).open(encoding="utf-8") as file:
                     line = file.readline()
                 if "@GMULTIPOINT" in line or "@GPOINT" in line:
                     kwargs["S"] = "s0.2c"
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -247,7 +247,7 @@ def plot(  # noqa: PLR0912
             kwargs["S"] = "s0.2c"
         elif kind == "file" and str(data).endswith(".gmt"):  # OGR_GMT file
             try:
-                with Path(which(data)).open() as file:
+                with Path(which(data)).open(encoding="locale") as file:
                     line = file.readline()
                 if "@GMULTIPOINT" in line or "@GPOINT" in line:
                     kwargs["S"] = "s0.2c"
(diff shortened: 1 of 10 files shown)


weiji14:
Setting as ready for review to test `encoding="locale"` on Windows first. Will try using `encoding="utf-8"` after all cross-platform tests complete.

pygmt/src/plot.py
```diff
@@ -247,7 +247,7 @@ def plot(  # noqa: PLR0912
             kwargs["S"] = "s0.2c"
         elif kind == "file" and str(data).endswith(".gmt"):  # OGR_GMT file
             try:
-                with Path(which(data)).open() as file:
+                with Path(which(data)).open(encoding="utf-8") as file:
```
weiji14, line 250: This assumes that `*.gmt` files are always encoded in UTF-8, is this the case, or should we choose `encoding="locale"` instead?

weiji14 (commented):
Main changes to check are in the `plot.py` and `plot3d.py` file since they are user-facing. The changes to the `tests/test_*.py` files should be ok.

pygmt/src/plot.py
seisman: `*.gmt` can be encoded in any encoding, so there is no way to make it work in all cases. 

> To be clear, we're only trying to parse the vector geometry type (Multipoint/Point) from the *.gmt file, not read the whole file, so maybe ok to assume that the header lines are utf-8 compatible?

It usually works, but will fail if the file uses a different encoding and contains comments on the first line.

Actually, we're assuming that the `@G` record is always on the first line but that's not always true.

seisman: `ogr2ogr` also produces OGR_GMT files with `# @VGMT1.0` followed by `@GP`, so it should be fine, although users can manually putting `@GP` at the second line and even add any comments at the begin of the file.

Anyway, that's just rare cases and it's safe to ignore them.

seisman (approved):

PR #3321, opened 2024-07-09
Replace rio.set_crs with rio.write_crs in load_tile_map function

weiji14:
**Description of proposed changes**

[Rioxarray 0.16.0](https://github.com/corteva/rioxarray/releases/tag/0.16.0) has deprecated the use of `set_crs` in favour of [`write_crs`](https://corteva.github.io/rioxarray/latest/rioxarray.html#rioxarray.rioxarray.XRasterBase.write_crs). Xref https://github.com/corteva/rioxarray/pull/793

Fixes `FutureWarning: It is recommended to use 'rio.write_crs()' instead. 'rio.set_crs()' will likelybe removed in a future release`

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3323, opened 2024-07-10
load_tile_map: Register the rio accessor by importing rioxarray so the returned raster has CRS

seisman:

The `rio` accessor is only registered when rioxarray is imported. Currently, we import rioxarray in `Figure.tilemap` (https://github.com/GenericMappingTools/pygmt/blob/3a32169402b992e013d5a28006cb3fa8135f7d80/pygmt/src/tilemap.py#L12) but don't do it in `load_tile_map`. Thus, the raster returned by `load_tile_map` alone may have CRS set or not, depending on if a script imports `rioxarray` or not elsewhere.

This PR fixes the issue by importing the rioxarray package in `pygmt/datasets/tile_map.py` so that the `rio` accessor is always registered when rioxarray is installed.

### DIFF
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -123,8 +123,7 @@ def load_tile_map(
       * x            (x) float64 ... -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
         spatial_ref  int64 ... 0
     >>> # CRS is set only if rioxarray is available
-    >>> import importlib
-    >>> if importlib.util.find_spec("rioxarray"):
+    >>> if hasattr(dataarray, "rio"):
     ...     raster.rio.crs
     CRS.from_epsg(3857)
     """
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -3,6 +3,7 @@ Function to load raster tile maps from XYZ tile providers, and load as
 :class:`xarray.DataArray`.
 """
 
+import contextlib
 from typing import Literal
 
 from packaging.version import Version
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -18,10 +18,8 @@ except ImportError:
 
 try:
     import rioxarray  # rioxarray is needed to register the rio accessor  # noqa: F401
-
-    _HAS_RIOXARRAY = True
 except ImportError:
-    _HAS_RIOXARRAY = False
+    pass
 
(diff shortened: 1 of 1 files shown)


weiji14 (commented):
Have manually triggered the doctests for this branch at https://github.com/GenericMappingTools/pygmt/actions/runs/9867651036/job/27248357251, looks ok, but a couple of suggestions.

pygmt/datasets/tile_map.py
```diff
@@ -16,6 +16,11 @@
     TileProvider = None
     _HAS_CONTEXTILY = False
 
+try:
```
seisman, line 19: Style check fails, and ruff recommends:
```
with contextlib.suppress(ImportError):
    import rioxarray
```
but then ruff formats it to:
```
with contextlib.suppress(ImportError):
    pass
```
which makes no sense. 

```suggestion
try:  # noqa: SIM105
```

pygmt/datasets/tile_map.py
```diff
@@ -16,6 +16,11 @@
     TileProvider = None
     _HAS_CONTEXTILY = False
 
+try:  # noqa: SIM105
+    import rioxarray  # rioxarray is needed to register the rio accessor  # noqa: F401
+except ImportError:
+    pass
```
weiji14, line 22: This seems to work for me with `ruff=0.5.1`, maybe the `noqa: F401` comment is needed? Also need to add `import contextlib` at the top.

```suggestion
with contextlib.suppress(ImportError):
    import rioxarray  # noqa: F401
```

pygmt/datasets/tile_map.py
```diff
@@ -117,6 +122,10 @@ def load_tile_map(
       * y            (y) float64 ... -7.081e-10 -7.858e+04 ... -1.996e+07 -2.004e+07
       * x            (x) float64 ... -2.004e+07 -1.996e+07 ... 1.996e+07 2.004e+07
         spatial_ref  int64 ... 0
+    >>> # CRS is set only if rioxarray is available
+    >>> if hasattr(raster, "rio"):
+    ...     raster.rio.crs
```
weiji14, line 127: To be honest, we probably don't need the `if hasattr(raster, "rio")` check, because the `spatial_ref   int64 ... 0` line above (added in #3321) already means that rioxarray is used (the doctests CI run with `rioxarray` installed). But I guess it'll be good to mention `rioxarray` somewhere in the docstring, albeit indirectly.

weiji14 (approved):

PR #3318, opened 2024-07-08
POC: Wrap GMT_Read_Data and read datasets/grids/images into GMT data container

seisman:
**Description of proposed changes**

This PR wraps the GMT API function [GMT_Read_Data](https://docs.generic-mapping-tools.org/dev/devdocs/api.html#gmt-read-data). 

Currently, this PR contains 4 commits:

- c55dd0ce: Wrap the `GMT_Read_Data` function
- 7e712257: Add tests for reading datasets/grids into GMT container `_GMT_DATASET`/`_GMT_GRID`
- 8f46ad3c: Merge PR #3128 into this branch so that we can test the behavior for `_GMT_IMAGE` (with minor fixes)
- fa37838a: Add tests for reading images

All new tests pass and the most important thing I learned is, that for an input grid/image, we can always read it into a GMT_IMAGE container. For grid, `header.n_bands=1` and for image, `header.n_bands=3`(or any other values). The `GMT_Read_Data` API function can read a grid/image in either one step or two steps. Here "two steps" means reading the header first and then reading the data. Reading the header only is very efficient (a few milliseconds) even for huge grid/image files. Thus, we can read the grid/image header in and check `header.n_bands` to determine the input data kind, which addresses the concerns in #https://github.com/GenericMappingTools/pygmt/pull/3115#discussion_r1667671439.

So, the next steps are:

- [x] Keep the first two commits only, discard the 3rd commit (already in #3128) and backup the tests in the 4th commit.
- [x] Polish the `GMT_Read_Data` wrapper and the tests for datasets/grids and finalize the wrapper #3324
- [ ] Implement the idea about reading the grid/image header to determine the input data kind, then we should be able to finish and merge #3115
- [ ] Break #3128 into two PRs. The first PR only contains the definition and fields of the `GMT_IMAGE` wrapper (line 1-71 of `pygmt/datatypes/image.py` in PR #3128, even without any doctests), and another PR with the remaining WIP codes. Then review/merge the first PR and work on the 2nd PR in the future.
- [ ] After merging the 1st PR in point 4, we should be able to add all the tests in fa37838a.  

I plan to work on the above steps in separate PRs and keep this PR unchanged so that we can trace back to this PR in the future.

PR #3324, opened 2024-07-10
Wrap the GMT API function GMT_Read_Data to read data into GMT data containers

seisman:

See #3318 for the motivations. 

This PR wraps the GMT API functions `GMT_Read_Data` for reading data file into GMT data containers. Currently, only `GMT_DATASET` and `GMT_GRID` are supported, other data containers will be supported later.

The first two commits 5a56ba8ad2c314df7a00b597a3684b1710fc0318 and 7ccf641299b4900457257741bf2571ebe4d387ca are cherry-picked from PR #3318.

**Preview** at https://pygmt-dev--3324.org.readthedocs.build/en/3324/api/generated/pygmt.clib.Session.read_data.html

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1074,7 +1074,7 @@ class Session:
         infile: str,
         kind: Literal["dataset", "grid"],
         mode: str = "GMT_READ_NORMAL",
-        wesn: Sequence[float] | None = None,
+        region: Sequence[float] | None = None,
         data=None,
     ):
         """
@@ -1093,7 +1093,7 @@ class Session:
         mode
             How the data is to be read from the file. This option varies depending on
             the given family. See the GMT API documentation for details.
-        wesn
+        region
             Subregion of the data, in the form of [xmin, xmax, ymin, ymax, zmin, zmax].
             If ``None``, the whole data is read.
         data
@@ -1132,7 +1132,7 @@ class Session:
             self["GMT_IS_FILE"],  # Reading from a file
             self[geometry],
             self[mode],
-            sequence_to_ctypes_array(wesn, ctp.c_double, 6),
+            sequence_to_ctypes_array(region, ctp.c_double, 6),
             infile.encode(),
             data,
         )
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1121,9 +1121,9 @@ class Session:
         )
 
         # Determine the family and geometry from kind
-        family, geometry = {
-            "dataset": ("GMT_IS_DATASET", "GMT_IS_PLP"),
-            "grid": ("GMT_IS_GRID", "GMT_IS_SURFACE"),
+        family, geometry, dtype = {
+            "dataset": ("GMT_IS_DATASET", "GMT_IS_PLP", _GMT_DATASET),
+            "grid": ("GMT_IS_GRID", "GMT_IS_SURFACE", _GMT_GRID),
         }[kind]
 
         family_int = self._parse_constant(family, valid=FAMILIES, valid_modifiers=VIAS)
@@ -1141,7 +1141,7 @@ class Session:
             infile.encode(),
             data,
         )
-        return data_ptr
+        return ctp.cast(data_ptr, ctp.POINTER(dtype))
 
     def write_data(self, family, geometry, mode, wesn, output, data):
         """
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1125,7 +1125,7 @@ class Session:
             restype=ctp.c_void_p,  # data_ptr
         )
 
-        # Determine the family and geometry from kind
+        # Determine the family, geometry and data container from kind
         family, geometry, dtype = {
             "dataset": ("GMT_IS_DATASET", "GMT_IS_PLP", _GMT_DATASET),
             "grid": ("GMT_IS_GRID", "GMT_IS_SURFACE", _GMT_GRID),
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1100,7 +1100,9 @@ class Session:
             data geometry from the ``kind`` parameter.
         mode
             How the data is to be read from the file. This option varies depending on
-            the given family. See the GMT API documentation for details.
+            the given family. See the
+            :gmt-docs:`GMT API documentation <devdocs/api.html#import-from-a-file-stream-or-handle>`
+            for details.
         region
             Subregion of the data, in the form of [xmin, xmax, ymin, ymax, zmin, zmax].
             If ``None``, the whole data is read.
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1071,35 +1071,31 @@ class Session:
 
     def read_data(
         self,
-        family: str,
-        geometry: str,
-        mode: str,
-        wesn: Sequence[float] | None,
         infile: str,
+        kind: Literal["dataset", "grid"],
+        mode: str | None = None,
+        wesn: Sequence[float] | None = None,
         data=None,
     ):
         """
         Read a data file into a GMT data container.
 
-        Wraps ``GMT_Read_Data`` but only allows reading from a file, so the ``method``
-        ``method`` argument is omitted.
+        Wraps ``GMT_Read_Data`` but only allows reading from a file. The function
+        definition is different from the original C API function.
 
         Parameters
         ----------
-        family
-            A valid GMT data family name (e.g., ``"GMT_IS_DATASET"``). See the
-            ``FAMILIES`` attribute for valid names.
-        geometry
-            A valid GMT data geometry name (e.g., ``"GMT_IS_POINT"``). See the
-            ``GEOMETRIES`` attribute for valid names.
+        input
+            The input file name.
+        kind
+            The data kind of the input file. Currently, valid values are ``"dataset"``
+            and ``"grid"``.
         mode
             How the data is to be read from the file. This option varies depending on
             the given family. See the GMT API documentation for details.
         wesn
             Subregion of the data, in the form of [xmin, xmax, ymin, ymax, zmin, zmax].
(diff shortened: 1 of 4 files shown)


pygmt/tests/test_clib_read_data.py
```diff
@@ -0,0 +1,129 @@
...
+        data_ptr = lib.read_data(infile, kind="grid", mode="GMT_CONTAINER_ONLY")
+        grid = data_ptr.contents
+        header = grid.header.contents
+        assert header.n_rows == 14
+        assert header.n_columns == 8
+        assert header.wesn[:] == [-55.0, -47.0, -24.0, -10.0]
+        assert header.z_min == 190.0
+        assert header.z_max == 981.0
+        assert header.n_bands == 1  # Explicitely check n_bands
+        assert not grid.data  # The data is not read yet
```
seisman, line 96: Can't do `xrgrid = data_ptr.contents.to_dataarray()` here because `grid.data` is `NULL`. Maybe we should modify the `GMT_GRID.to_dataarray()` method to fill the DataArray with `np.empty()` if `grid.data` is not read yet?

pygmt/clib/session.py
```diff
@@ -1068,6 +1069,82 @@ def put_matrix(self, dataset, matrix, pad=0):
...
+                ctp.c_uint,  # geometry
+                ctp.c_uint,  # mode
+                ctp.POINTER(ctp.c_double),  # wesn
+                ctp.c_char_p,  # infile
+                ctp.c_void_p,  # data
+            ],
+            restype=ctp.c_void_p,  # data_ptr
+        )
+
+        # Determine the family and geometry from kind
```
seisman, line 1128:
```suggestion
        # Determine the family, geometry and data container from kind
```

pygmt/clib/session.py
```diff
@@ -1067,6 +1067,96 @@ def put_matrix(self, dataset, matrix, pad=0):
...
+    ):
+        """
+        Read a data file into a GMT data container.
+
+        Wraps ``GMT_Read_Data`` but only allows reading from a file. The function
+        definition is different from the original C API function.
+
+        Parameters
+        ----------
+        input
```
weiji14, line 1088:
```suggestion
        infile
```

pygmt/tests/test_clib_read_data.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_clib_read_data_grid(expected_xrgrid):
+    """
+    Test the Session.read_data method for grids.
+    """
+    with Session() as lib:
+        grid = lib.read_data("@static_earth_relief.nc", kind="grid").contents
+        xrgrid = grid.to_dataarray()
+        xr.testing.assert_equal(xrgrid, expected_xrgrid)
+        assert grid.header.contents.n_bands == 1  # Explicitely check n_bands
```
weiji14, line 77:
```suggestion
        assert grid.header.contents.n_bands == 1  # Explicitly check n_bands
```

pygmt/tests/test_clib_read_data.py
```diff
@@ -0,0 +1,139 @@
...
+        # Read the header first
+        data_ptr = lib.read_data(infile, kind="grid", mode="GMT_CONTAINER_ONLY")
+        grid = data_ptr.contents
+        header = grid.header.contents
+        assert header.n_rows == 14
+        assert header.n_columns == 8
+        assert header.wesn[:] == [-55.0, -47.0, -24.0, -10.0]
+        assert header.z_min == 190.0
+        assert header.z_max == 981.0
+        assert header.n_bands == 1  # Explicitely check n_bands
```
weiji14, line 96:
```suggestion
        assert header.n_bands == 1  # Explicitly check n_bands
```

pygmt/clib/session.py
```diff
@@ -1067,6 +1067,96 @@ def put_matrix(self, dataset, matrix, pad=0):
...
+            A valid GMT data family name (e.g., ``"GMT_IS_DATASET"``). See the
+            ``FAMILIES`` attribute for valid names. If ``None``, will determine the data
+            family from the ``kind`` parameter.
+        geometry
+            A valid GMT data geometry name (e.g., ``"GMT_IS_POINT"``). See the
+            ``GEOMETRIES`` attribute for valid names. If ``None``, will determine the
+            data geometry from the ``kind`` parameter.
+        mode
+            How the data is to be read from the file. This option varies depending on
+            the given family. See the GMT API documentation for details.
```
weiji14, line 1103: Can you link to where in the GMT API documentation what options there are for `mode`?

pygmt/tests/test_clib_read_data.py
```diff
@@ -0,0 +1,139 @@
...
+        header = image.header.contents
+        assert header.n_rows == 180
+        assert header.n_columns == 360
+        assert header.wesn[:] == [-180.0, 180.0, -90.0, 90.0]
+        # Explicitely check n_bands. Only one band is read for 3-band images.
+        assert header.n_bands == 1
+
+        if _HAS_RIOXARRAY:  # Full check if rioxarray is installed.
+            xrimage = image.to_dataarray()
+            expected_xrimage = xr.open_dataarray(which("@earth_day_01d_p"))
```
weiji14, line 123: Maybe be explicit that this is using `engine="rasterio"`.

```suggestion
            expected_xrimage = xr.open_dataarray(which("@earth_day_01d_p"), engine="rasterio")
```

pygmt/tests/test_clib_read_data.py
```diff
@@ -0,0 +1,139 @@
...
+    with Session() as lib:
+        data_ptr = lib.read_data(
+            "@earth_day_01d_p", kind="grid", mode="GMT_CONTAINER_AND_DATA"
+        )
+        image = data_ptr.contents
+        header = image.header.contents
+        assert header.n_rows == 180
+        assert header.n_columns == 360
+        assert header.wesn[:] == [-180.0, 180.0, -90.0, 90.0]
+        # Explicitely check n_bands. Only one band is read for 3-band images.
```
weiji14, line 118:
```suggestion
        # Explicitly check n_bands. Only one band is read for 3-band images.
```

pygmt/clib/session.py
seisman: `mode` is explained at https://docs.generic-mapping-tools.org/dev/devdocs/api.html#import-from-a-file-stream-or-handle.

pygmt/clib/session.py
```diff
@@ -1067,6 +1067,98 @@ def put_matrix(self, dataset, matrix, pad=0):
         if status != 0:
             raise GMTCLibError(f"Failed to put matrix of type {matrix.dtype}.")
 
+    def read_data(
+        self,
+        infile: str,
+        kind: Literal["dataset", "grid"],
+        family: str | None = None,
+        geometry: str | None = None,
+        mode: str = "GMT_READ_NORMAL",
```
weiji14, line 1076: Thinking on whether to make this an enum or use `Literal` instead of a str. Looking at https://docs.generic-mapping-tools.org/6.5/devdocs/api.html#import-from-a-file-stream-or-handle, it seems like it's possible to pass in `mode` as one of the variants of [`GMT_enum_container`](https://github.com/GenericMappingTools/gmt/blob/3e9fcb1f54e4e93fc1e309e0be6e2288eb79f3d2/src/gmt_resources.h#L136-L143) or [`GMT_enum_gridio`](https://github.com/GenericMappingTools/gmt/blob/3e9fcb1f54e4e93fc1e309e0be6e2288eb79f3d2/src/gmt_resources.h#L363-L379)? It also looks like `mode` can be None (or unset) if reading from text or table.

seisman: I believe the documentation is incomplete, and `mode` can accept specific values plus some modifiers, making it impossible to list all of them.

weiji14: Yeah that what I thought too, there might be other options we's unaware of. That said, should we allow for `mode=None` (or whatever works when we don't want to set a `mode`)?

seisman: In the GMT C API, `mode` has an `unsigned int` type, so the `mode` argument must be an integer number and can't be `None`. The C enum `GMT_READ_NORMAL` has a value of `0` (https://github.com/GenericMappingTools/gmt/blob/3e9fcb1f54e4e93fc1e309e0be6e2288eb79f3d2/src/gmt_resources.h#L244), which is also the default read mode.

We can define `mode=None`, and add one line code like:
```
if mode == None:
   mode = "GMT_READ_NORMAL"
```
but I feel it's not as good as setting `GMT_READ_NORMAL` as the default.

weiji14: Ah ok then, then the `GMT_READ_NORMAL` default here makes sense. Maybe good to mention it in the docstring.

pygmt/clib/session.py
```diff
@@ -1067,6 +1067,98 @@ def put_matrix(self, dataset, matrix, pad=0):
...
+            family from the ``kind`` parameter.
+        geometry
+            A valid GMT data geometry name (e.g., ``"GMT_IS_POINT"``). See the
+            ``GEOMETRIES`` attribute for valid names. If ``None``, will determine the
+            data geometry from the ``kind`` parameter.
+        mode
+            How the data is to be read from the file. This option varies depending on
+            the given family. See the
+            :gmt-docs:`GMT API documentation <devdocs/api.html#import-from-a-file-stream-or-handle>`
+            for details.
```
weiji14, line 1105: Feel free to reword

```suggestion
            for details. Default is `GMT_READ_NORMAL` which corresponds to the default read mode
            value of 1 in the `GMT_enum_read` enum.
```

weiji14 (approved):
Thanks @seisman, just one additional suggestion on documenting the default `mode="GMT_READ_NORMAL"`. Also need to fix the [W505 doc-line-too-long](https://docs.astral.sh/ruff/rules/doc-line-too-long/) lint error before merging.

Issue #1078, opened 2021-03-19
Figure.text(): Support paragraph mode


yebo-liu:
**Description of the desired feature**
I was trying to place some texts on my plot. The texts are quite long and require to be separated into paragraphs. Maybe I got something wrong, but it looks like the paragraph mode of gmt hasn't been wrapped in pygmt yet?

**Are you willing to help implement and maintain this feature?** Yes/No

weiji14:
Hi @yebo-liu, could you provide an example of your PyGMT script and the text that you want to plot? Are you trying to plot the text using `fig.text`, or in `fig.legend`, or somewhere else?

yebo-liu:
Hi @weiji14 , thanks for the prompt reply. Please see below 

`fig = pygmt.Figure()`
`fig.basemap(projection="W0/12i", region="g", frame=['30g30'],transparency=70)`
`fig.text(text='long texts',x=90,y=0,font='10p')`
`fig.show()`

![image](https://user-images.githubusercontent.com/56617247/111727012-fddc0100-88a4-11eb-98a1-f84371e239fa.png)

I'm not really familiar with PyGMT nor GMT. What I'm trying to achieve is that to get the texts change lines and break it into more than one paragraph if possible. Hopefully I made myself clear.

weiji14:
Ok, I won't be able to reply for the next few hours, but maybe someone else can chip in.

There was some work on supporting long titles in `fig.basemap` at https://github.com/GenericMappingTools/gmt/pull/4562 (but that will be available in GMT 6.2.0, not the current 6.1.1). Alternatively, there might be a way to hack out a solution using `fig.legend` (see https://docs.generic-mapping-tools.org/6.1/gallery/ex22.html). Is your long paragraph meant to be a sort of description of the map? Or more of a title description? Just want to help you figure out the best solution here.

seisman:
@yebo-liu I believe it would be very trick to have a Pythonic way to support paragraph mode. Here is an PyGMT example modified from the official text manual (https://docs.generic-mapping-tools.org/dev/text.html#examples).
```python
import pygmt

with open("text.txt", "w") as fp:
    fp.write('''\
> 0 0 13p 3i j
@%5%Figure 1.@%% This illustration shows nothing useful, but it still needs
a figure caption. Highlighted in @;255/0/0;red@;; you can see the locations
of cities where it is @\_impossible@\_ to get any good Thai food; these are to be avoided.
''')

[...]

PR #3326, opened 2024-07-11
Add the Session.virtualfile_from_stringio method to allow StringIO input for certain functions/methods

### LINKED ISSUE #571: Allow Figure.legend to read from StringIO

seisman:

This PR adds support for StringIO inputs, which was initially requested in #571 (4 yeas ago!).

GMT doesn't natively support reading StringIO objects. To pass a StringIO object into GMT, we need to create a GMT_DATASET container and store the contents of the StringIO object in the GMT_DATASET container. 

Ideally, any PyGMT methods/functions should accept StringIO input, but it's technically difficult to implement. The main reason is that, when creating a GMT_DATASET container, we need to specify the number of tables, segments, columns, and rows, which means we must try very hard to parse the contents of the StringIO object. The number of columns actually is the most difficult to determine. For example, the content below should be 2 numeric columns and one trailing text column, but we may incorrectly parse it to have 0 numeric columns:
```
30.0W 20.0S TEXT1 TEXT4
35.0W 20.5S TEXT2
```
and the following content that we want to pass to `Figure.text`'s paragraph mode (#1078), should have zero numeric columns and one trailing text column, but we may incorrectly think it has two numeric columns: 
```
gmt text -R0/3/0/5 -JX3i -h1 -M -N -F+f12,Times-Roman+jLT -pdf figure << EOF
This is an unmarked header record not starting with #
> 0 -0.5 13p 3i j
20240101 010101 Event 1
20240101 010101 Event 2
EOF
```
Thus, to keep things as simple as possible, I think we should only allow StringIO input for a few modules (e.g., makes much sense for `Figure.legend`, but makes little sense to `Figure.plot`). In this case, we just need to store the StringIO contents into the trailing text of the GMT_DATASET container.

This PR implements the virtualfile_from_string method to allow StringIO objects. There are two related PRs:

- Support StringIO object in `Figure.legend` (PR #3438 addressing #571)
- Let `Figure.text` support paragraph mode (PR #XX addressing #1078)

**Reminders**

**Slash Commands**


seisman:
> Ideally, any PyGMT methods/functions should accept StringIO input, but it's technically difficult to implement. The main reason is that, when creating a GMT_DATASET container, we need to specify the number of tables, segments, columns, and rows, which means we must try very hard to parse the contents of the StringIO object. The number of columns actually is the most difficult to determine.

Actually, when calling a GMT module, we usually know the expected number of numeric columns (for example, 0 for `legend`, 2 or more for `plot`), so we could add an extra parameter `n_columns` to the `virtualfile_from_stringio` function after refactoring the `virtualfile_in` method in #2744. But I feel that passing a StringIO object to modules like `plot` makes little sense and has very low priority. Then the question is, how can we add StringIO support to specific functions only (e.g., `Figure.legend`)?

pygmt/clib/session.py
```diff
@@ -1603,6 +1610,83 @@ def virtualfile_from_grid(self, grid):
...
+        header = None
+        string_arrays = []
+        for line in stringio.getvalue().splitlines():
+            if line.startswith("#"):  # Skip comments
+                continue
+            if line.startswith(">"):  # Segment header
+                if header is not None:  # Only one segment is allowed now.
+                    raise GMTInvalidInput("Only one segment is allowed.")
+                header = line
+                continue
```
weiji14, line 1664: Need a test to check that multi-segment inputs fail.

pygmt/clib/session.py
```diff
@@ -1603,6 +1610,83 @@ def virtualfile_from_grid(self, grid):
...
+        with self.open_virtualfile(family, geometry, "GMT_IN", dataset) as vfile:
+            try:
+                yield vfile
+            finally:
+                # Must set the text to None to avoid double freeing the memory
+                seg.text = None
+
     def virtualfile_in(  # noqa: PLR0912
         self,
         check_kind=None,
```
weiji14, line 1692: > Then the question is, how can we add StringIO support to specific functions only (e.g., `Figure.legend`)?

Maybe `legend` should have a special check_kind? E.g. `if check_kind("legend"): valid_kinds += ("stringio", ...)`?

seisman: See https://github.com/GenericMappingTools/pygmt/pull/3438 for the POC PR for `Figure.legend`. Since we already checked the data kind before entering the session, we can have `check_kind=False` in `Figure.legend`.

pygmt/clib/session.py
seisman: It's likely we have to allow multi-segments in StringIO.

Here is a CLI version for typesetting a paragraph of text. It seems the first line must be a line not starting with `#` and is ignored. Not sure why it was designed like this. I'll open another POC PR for `Figure.text` and see how it works.
```
gmt text -R0/3/0/5 -JX3i -h1 -M -N -F+f12,Times-Roman+jLT -pdf figure << EOF
This is an unmarked header record not starting with #
> 0 -0.5 13p 3i j
@%5%Figure 1.@%% This illustration shows nothing useful, but it still needs
a figure caption. Highlighted in @;255/0/0;red@;; you can see the locations
of cities where it is @\_impossible@\_ to get any good Thai food; these are to be avoided.
EOF
```

seisman: Correction: I didn't see the `-h1` option in the CLI version, which skips the first line. Without `-h1`, the CLI version still works:
```
gmt text -R0/3/0/5 -JX3i -M -N -F+f12,Times-Roman+jLT -pdf figure << EOF
> 0 -0.5 13p 3i j
@%5%Figure 1.@%% This illustration shows nothing useful, but it still needs
a figure caption. Highlighted in @;255/0/0;red@;; you can see the locations
of cities where it is @\_impossible@\_ to get any good Thai food; these are to be avoided.
EOF
````

Comment (seisman) [reply pygmt/clib/session.py]: Here is a POC example to show how to support paragraph mode via StringIO:
```python
In [1]: import io

In [2]: stringio = io.StringIO(
   ...:     "> 0 -0.5 13p 3i j\n"
   ...:     "@%5%Figure 1.@%% This illustration shows nothing useful, but it still needs\n"
   ...:     "a figure caption. Highlighted in @;255/0/0;red@;; you can see the locations\n"
   ...:     "of cities where it is @_impossible@_ to get any good Thai food; these are to be avoided.\n"
   ...: )

In [3]: import pygmt

In [4]: fig = pygmt.Figure()

In [5]: from pygmt.clib import Session

In [6]: with Session() as lib:
   ...:     with lib.virtualfile_in(data=stringio) as vintbl:
   ...:         lib.call_module(module="text", args=f"{vintbl} -R0/3/0/5 -JX3i -M -N -F+f12,Times-Roman+jLT")
   ...:

In [7]: fig.show()
```

seisman: StringIO objects containing multi-segments are supported in the latest version.

pygmt/clib/session.py
```diff
@@ -1603,6 +1610,100 @@ def virtualfile_from_grid(self, grid):
...
+            family,
+            geometry,
+            mode="GMT_CONTAINER_ONLY|GMT_WITH_STRINGS",
+            dim=[n_tables, n_segments, n_rows, n_columns],
+        )
+        dataset = ctp.cast(dataset, ctp.POINTER(_GMT_DATASET))
+        table = dataset.contents.table[0].contents
+        for i, segment in enumerate(segments):
+            seg = table.segment[i].contents
+            if segment["header"] != "":
```
seisman, line 1692:
```suggestion
            if segment["header"]:
```

pygmt/clib/session.py
```diff
@@ -60,6 +61,7 @@
     "GMT_IS_PLP",  # items could be any one of POINT, LINE, or POLY
     "GMT_IS_SURFACE",  # items are 2-D grid
     "GMT_IS_VOLUME",  # items are 3-D grid
+    "GMT_IS_TEXT",  # Text strings which triggers ASCII text reading
```
weiji14, line 64: Is only ASCII supported for now? What about other encodings?

weiji14 (approved):
Thanks @seisman for adding multi-segment support! I just have one concern/thought about non-ASCII character support, but that could be handled later (if needed).

pygmt/clib/session.py
seisman: GMT only accepts ASCII text strings. Any non-ASCII characters should be written in the form of octal codes.

PR #3238, opened 2024-05-09
POC: Implement the new alias system, general parameter classes and some refactors

seisman:
See #3239 for the detailed context.

This PR implements the new alias system proposed in #3239. It also add some new features and refactor some wrappers to show how the new alias system will affect the current codes.

Here is a summary of changes in this PR:

- `pygmt/alias.py`: Implement the new alias system
- `pygmt/params/base.py`: Implement the `BaseParam` class as a base class for other parameter classes
- `pygmt/params/box.py`: Implement the `Box` class, showing how to define new classes in a simple way
- `pygmt/params/frame.py`: Impement the `Frame` class, showing that the current design works work for the most complicated GMT option
- `pygmt/src/basemap.py`: Refactor so that you can try the new `Frame` parameter in `Figure.basemap`
- `pygmt/src/dimfilter.py`: How the new alias system can affect grid-related modules
- `pygmt/src/image.py`: How the new alias system can replace the old alias system
- `pygmt/src/logo.py`: Refactor so that you can try the `Box` class
- `pygmt/src/scalebar.py`: Initial implementation of the high-level wrapper for adding a scale bar.
- `pygmt/src/timestamp.py`: `Figure.timestamp` is one of the wrappers that don't use the alias system. Using the new alias system can simplify the code.
- `pygmt/src/coast.py`/`pygmt/src/binstats.py`: How to support long-form arguments

This is just the initial implementation. So, some doctests and tests are still missing. We will break this PR into multiple smaller PRs and make the implementation more robust after people generally agree with the design of the new alias system.

**Preview**:

- https://pygmt-dev--3238.org.readthedocs.build/en/3238/api/index.html#common-parameters
- https://pygmt-dev--3238.org.readthedocs.build/en/3238/api/generated/pygmt.params.Box.html#pygmt.params.Box
- https://pygmt-dev--3238.org.readthedocs.build/en/3238/api/generated/pygmt.params.Frame.html#pygmt.params.Frame
- https://pygmt-dev--3238.org.readthedocs.build/en/3238/api/generated/pygmt.Figure.logo.html#pygmt.Figure.logo
- https://pygmt-dev--3238.org.readthedocs.build/en/3238/api/generated/pygmt.Figure.coast.html#pygmt.Figure.coast
-

PR #3334, opened 2024-07-17
pygmt.grd2cpt & pygmt.makecpt: Simplify the logic for dealing with CPT output

seisman:
Simplify:
```
        if kwargs.get("H") is None:  # if no output is set
            arg_str = build_arg_list(kwargs)
        else:  # if output is set
            outfile, kwargs["H"] = kwargs.pop("H"), True
            if not outfile or not isinstance(outfile, str):
                raise GMTInvalidInput("'output' should be a proper file name.")
            arg_str = build_arg_list(kwargs, outfile=outfile)
```
to 
```
    if (output := kwargs.pop("H", None)) is not None:
        if not isinstance(output, str) or output == "":
            raise GMTInvalidInput("'output' should be a proper file name.")
        kwargs["H"] = True
    arg_str = build_arg_list(kwargs, outfile=output)
```

It works because `outfile` in `build_arg_list` can be `None`.

### DIFF
--- a/pygmt/src/grd2cpt.py
+++ b/pygmt/src/grd2cpt.py
@@ -183,13 +183,15 @@ def grd2cpt(grid, **kwargs):
     """
     if kwargs.get("W") is not None and kwargs.get("Ww") is not None:
         raise GMTInvalidInput("Set only categorical or cyclic to True, not both.")
+
+    if (output := kwargs.pop("H", None)) is not None:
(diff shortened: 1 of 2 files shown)


weiji14 (approved):
Needed to read that if-statement a few times to get the logic, but looks ok.

seisman:
The if-statement can be further simplified after PR https://github.com/GenericMappingTools/pygmt/pull/3238

PR #3335, opened 2024-07-17
Refactor the data_kind and validate_data_input functions

seisman:

This PR is a subset of PR #2744. PR #2744 takes more time than I initially expected (Still not satisfied with the new `Session.virtualfile_in` API function), so better to refactor the `data_kind`/`validate_data_input` function in a separate PR (this PR).

As mentioned in PR #2744:

> The `data_kind` function does three things: (1) determines the kind of the input data, and (2) checks if the data/x/y/z combinations are valid; and (3) checks if the matrix-type data has 3 columns. 
> The `data_kind` function is called inside `Session.virtualfile_in`, but sometimes we need to know the input data kind when wrapping GMT modules, for example, in `Figure.plot` and `Figure.plot3d`. It means the `data_kind` function is called twice, which is not necessary.

This PR moves the `validate_data_input` function outside of the `data_kind`, and calls it in `Session.virtualfile_in` explicitly. In this way, each function only focuses on one single data. which makes the functions easier to maintain.

I also don't like the current function design of the `data_kind`/`validate_data_input` functions, but I prefer to refactor the function details in a separate PR, to make this PR smaller for reviewing.

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -130,9 +130,6 @@ def data_kind(
     * 1-D arrays x and y (and z, optionally)
     * an optional argument (None, bool, int or float) provided as 'data'
 
-    Arguments should be ``None`` if not used. If doesn't fit any of these
-    categories (or fits more than one), will raise an exception.
-
     Parameters
     ----------
     data : str, pathlib.PurePath, None, bool, xarray.DataArray or {table-like}
@@ -150,7 +147,6 @@ def data_kind(
 
     Examples
     --------
-
     >>> import numpy as np
     >>> import xarray as xr
     >>> import pathlib
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -169,7 +169,7 @@ def data_kind(
     >>> data_kind(data=xr.DataArray(np.random.rand(3, 4, 5)))
     'image'
     """
-    # determine the data kind
+    kind: Literal["arg", "file", "geojson", "grid", "image", "matrix", "vectors"]
     if isinstance(data, str | pathlib.PurePath) or (
         isinstance(data, list | tuple)
         and all(isinstance(_file, str | pathlib.PurePath) for _file in data)
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -19,7 +19,7 @@ from pygmt.encodings import charset
 from pygmt.exceptions import GMTInvalidInput
 
 
-def _validate_data_input(
+def validate_data_input(
     data=None, x=None, y=None, z=None, required_z=False, required_data=True, kind=None
 ):
     """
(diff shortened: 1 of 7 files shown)


pygmt/helpers/utils.py
```diff
@@ -12,74 +12,78 @@
...
-from typing import Any
+from typing import Any, Literal
 
 import xarray as xr
 from pygmt.encodings import charset
 from pygmt.exceptions import GMTInvalidInput
 
 
-def _validate_data_input(
+def validate_data_input(
```
seisman, line 22: Make the function public, because now it's called in `Session.virtualfile_in`.

pygmt/helpers/utils.py
```diff
@@ -193,15 +188,6 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
         kind = "matrix"
     else:
         kind = "vectors"
-    _validate_data_input(
```
seisman, line 196: Removed from here and moved to `Session.virtualfile_in`.

pygmt/helpers/utils.py
```diff
@@ -111,7 +115,9 @@ def _validate_data_input(
                 raise GMTInvalidInput("data must provide x, y, and z columns.")
 
 
-def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data=True):
+def data_kind(
```
seisman, line 118: Now this function only checks `data`, and `x`/`y`/`z` is no longer used.

pygmt/tests/test_helpers.py
```diff
@@ -32,25 +25,6 @@ def test_load_static_earth_relief():
     assert isinstance(data, xr.DataArray)
 
 
-@pytest.mark.parametrize(
```
seisman, line 35: This test is removed because:

1. `data_kind` no longer checks if the input is valid
2. All these tests are already covered by the `validate_data_input` doctests.

pygmt/helpers/utils.py
weiji14: Mentioned at https://github.com/GenericMappingTools/pygmt/pull/2744#discussion_r1435898105. I'm still unsure if we should make this public since this isn't something regular users should use. We can still import `_validate_data_input` and use it in `Session.virtualfile_in` no?

seisman: OK, changed it back to private in 7ff0d9e2698bf23339a3c271129e63b827245d4b.

pygmt/helpers/utils.py
```diff
@@ -124,64 +130,53 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
     * 1-D arrays x and y (and z, optionally)
```
weiji14, line 130: Delete this line about x/y/z?

```suggestion
```

pygmt/helpers/utils.py
```diff
@@ -124,64 +130,53 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
...
-    >>> data_kind(data=True, x=None, y=None, required_data=False)
+    >>> data_kind(data=True, required=False)
     'arg'
     >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
     'grid'
     >>> data_kind(data=xr.DataArray(np.random.rand(3, 4, 5)))
     'image'
     """
-    # determine the data kind
+    kind: Literal["arg", "file", "geojson", "grid", "image", "matrix", "vectors"]
```
weiji14, line 172: What is this line doing here?

pygmt/helpers/utils.py
```diff
@@ -111,7 +115,9 @@ def _validate_data_input(
                 raise GMTInvalidInput("data must provide x, y, and z columns.")
 
 
-def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data=True):
+def data_kind(
+    data: Any = None, required: bool = True
+) -> Literal["arg", "file", "geojson", "grid", "image", "matrix", "vectors"]:
     """
     Check what kind of data is provided to a module.
```
weiji14, line 123: Since x/y/z is no longer used, at L124-129, change the docstring to read:

```
    Possible types (provided to `data`):
    * a file name
    * a pathlib.PurePath object
    * an xarray.DataArray object
    * a 2-D matrix
```

pygmt/helpers/utils.py
```diff
@@ -124,64 +130,53 @@ def data_kind(data=None, x=None, y=None, z=None, required_z=False, required_data
     * 1-D arrays x and y (and z, optionally)
     * an optional argument (None, bool, int or float) provided as 'data'
```
weiji14, line 131:
```suggestion
    * an optional argument (None, bool, int or float)
```

pygmt/helpers/utils.py
seisman: To suppress the mypy error:
```
pygmt/helpers/utils.py:191: error: Incompatible return value type (got "str", expected "Literal['arg', 'file', 'geojson', 'grid', 'image', 'matrix', 'vectors']")  [return-value]
```

pygmt/helpers/utils.py
seisman: Updated the docstring at b575591ace1374b5bdc547f94fcf1262aacf247d. Currently, any unrecognized data type is taken as a `matrix` type. It's not ideal, so I plan to refactor the data_kind function later, and will also polish the docstring when refactoring.

pygmt/helpers/utils.py
weiji14: Ah interesting. Probably should turn this into an enum, but leave that for another day.

weiji14:
Do you know why the `test_velo_numpy_array_numeric_only` test is failing on Python 3.10/NumPy 1.24? https://github.com/GenericMappingTools/pygmt/blob/d3101f3ffa21f4aa105ee2a1fdf705e8fe23b9f2/pygmt/tests/test_velo.py#L30-L42

seisman:
> Do you know why the `test_velo_numpy_array_numeric_only` test is failing on Python 3.10/NumPy 1.24?
> 
> https://github.com/GenericMappingTools/pygmt/blob/d3101f3ffa21f4aa105ee2a1fdf705e8fe23b9f2/pygmt/tests/test_velo.py#L30-L42

It passes in the latest rerun.

weiji14 (approved):

PR #2784, opened 2023-10-31
Move virtualfile-related tests into a separate test file

seisman:
**Description of proposed changes**

- Move virtualfile-related tests into a separate test file
- Remove an invalid pylint directive

Address https://github.com/GenericMappingTools/pygmt/pull/2729#discussion_r1376684742.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3351, opened 2024-07-23
**Breaking**: data_kind: Now 'matrix' represents a 2-D numpy array and unrecognized data types fall back to 'vectors'

seisman:
~~Need to wait for #3481.~~

As can be seen in the doctests added in #3480, currently, `matrix` and `vectors` kinds are defined like below:

- `vectors` kind means `data=None` and `required=True`. It means that the input is given via a series of vectors (e.g., x/y/z). 
- `matrix` represents any data types not recognized as arg/file/stringio/grid/image/geojson. The most common data types include `pd.DataFrame`, `pd.Series`, `xr.Dataset`, `np.ndarray`, and sequence of sequences.

So, `matrix` here is an inclusive concept. However, in both Python and GMT, `matrix` usually means a **homogenous** 2-D array. When passing a "matrix" data to GMT, we usually treat the "matrix" as a series of vectors. The only exception is when the "matrix" data is a strict matrix (a 2-D numpy array) with `data.dtype.kind in "iuf"`. 

https://github.com/GenericMappingTools/pygmt/blob/d7560fa097819c7717a75dac85db1953da6ce430/pygmt/clib/session.py#L1742-L1755

I think it makes more sense to use a more strict definition for `matrix` and let all other data types be `vectors`. I propose to change the definitions of data kinds to:

- `matrix`: a 2-D homogenous numpy.ndarray 
- `vectors`: fallbacks to this kind for any unrecognized data types, including a dictionary, `pd.DataFrame`, `pd.Series`, `xr.Dataset`, nested lists/tuples, or any other unrecognized data types. 

The refactoring is mainly done in 808755d7747dcbcd93b8a0d495b11990d10deb6f.

Of course, `data=None, required=True` is still recognized as "vectors". I also think we should give this case another special kind like ``"none"``, which makes more sense and can also simplify the codes. [This is done in a separate PR at #3482, but I'm OK to merge #3482 into this PR first before merging this PR into main].

Other data kinds are unchanged. 

This is a breaking change since some previous "matrix" data are now recognized "vectors" and previously "vectors" is "none". But I guess very few users are using this function, so the breaking change should have minimal effects.


pygmt/helpers/utils.py
```diff
@@ -188,30 +188,36 @@ def _check_encoding(
 
 
 def data_kind(
-    data: Any = None, required: bool = True
```
seisman, line 191: The default value of `data` is `None`, but it makes no sense to call `data_kind()` without any arguments. Here I make `data` a required argument.

pygmt/helpers/utils.py
```diff
@@ -223,45 +230,54 @@ def data_kind(
...
     """
-    kind: Literal["arg", "file", "geojson", "grid", "image", "matrix", "vectors"]
+    # data is None, so data must be given via a series of vectors (i.e., x/y/z).
+    # The only exception is when dealing with optional virtual files.
+    if data is None:
+        return "vectors" if required else "arg"
+
+    # A file or a list of files
     if isinstance(data, str | pathlib.PurePath) or (
-        isinstance(data, list | tuple)
```
seisman, line 247: Check if `data` is a non-string iterator, instead of just checking if data is a list or a tuple.

pygmt/helpers/utils.py
```diff
@@ -223,45 +230,54 @@ def data_kind(
...
     'image'
+    >>> data_kind(data=np.arange(10).reshape((5, 2)))
+    'matrix'
+    >>> data_kind(data=[1, 2, 3])
+    'matrix'
     """
-    kind: Literal["arg", "file", "geojson", "grid", "image", "matrix", "vectors"]
+    # data is None, so data must be given via a series of vectors (i.e., x/y/z).
+    # The only exception is when dealing with optional virtual files.
+    if data is None:
```
seisman, line 256: The core part of the function is rewritten into multiple `if-return` statements to avoid too many `if-elif` branches.

pygmt/helpers/utils.py
```diff
@@ -222,46 +230,58 @@ def data_kind(
...
-    return kind
+        return "file"
+
+    # An option argument
+    if isinstance(data, bool | int | float):
+        return "arg"
+
+    # A xr.DataArray grid or image
+    if isinstance(data, xr.DataArray):
+        return "image" if len(data.dims) == 3 else "grid"
```
seisman, line 276: We probably should be more careful here, since we only check if `len(data.dims)` == 3. But I guess we can do it in the future.

pygmt/helpers/utils.py
```diff
@@ -222,46 +229,66 @@ def data_kind(
...
+
+    # Geo-like Python object that implements ``__geo_interface__`` (e.g.,
+    # geopandas.GeoDataFrame or shapely.geometry)
+    # Reference: https://gist.github.com/sgillies/2217756
+    if hasattr(data, "__geo_interface__"):
+        return "geojson"
+
+    # A 2-D numpy.ndarray
+    if isinstance(data, np.ndarray) and data.ndim == 2:
```
seisman, line 287: We probably can change this line to:
```suggestion
    if hasattr(data, "__array_interface__") and len(data.__array_interface__.shape) == 2:
```
to support more array-like objects that implements the `__array_interface__` protocol (https://numpy.org/doc/stable/reference/arrays.interface.html#object.__array_interface__), but it can be done in a future PR.

pygmt/helpers/utils.py
```diff
@@ -220,6 +229,26 @@ def data_kind(
...
+    >>> [data_kind(data=data) for data in (2, 2.0, True, False)]
+    ['arg', 'arg', 'arg', 'arg']
+    >>> data_kind(data="my-data-file.txt")
+    'file'
+    >>> data_kind(data=pathlib.Path("my-data-file.txt"))
+    'file'
+    >>> data_kind(data=["data1.txt", "data2.txt"])
+    'file'
+    >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
+    =======
```
seisman, line 251:
```suggestion
```

pygmt/src/x2sys_cross.py
```diff
@@ -195,7 +195,7 @@ def x2sys_cross(
         match data_kind(track):
             case "file":
                 file_contexts.append(contextlib.nullcontext(track))
-            case "matrix":
+            case "vectors":
```
seisman, line 198: pandas.DataFrame now is "vectors" kind.

pygmt/helpers/utils.py
```diff
@@ -312,7 +316,7 @@ def data_kind(
             # geopandas.GeoDataFrame or shapely.geometry).
             # Reference: https://gist.github.com/sgillies/2217756
             kind = "geojson"
-        case x if x is not None:  # Any not-None is considered as a matrix.
+        case np.ndarray() if data.ndim == 2:  # A 2-D numpy.ndarray object.
             kind = "matrix"
```
weiji14, line 320: Two things:

1. Should we extend `kind="matrix"` to 3-D numpy.ndarray objects?
2. I'm hesitant to match based on `np.ndarray()` class only, because while it's not widely advertised, checking for `data is not None` would have caught other array types implementing the `__array_function__` protocol (see [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html)), but checking for `isinstance(data, np.ndarray)` would exclude those. That said, I'm wondering if the fallback to `kind="vectors"` might just work, but need to test this out more extensively.

pygmt/clib/session.py
```diff
@@ -1801,29 +1798,32 @@ def virtualfile_in(  # noqa: PLR0912
...
                 _data = np.atleast_2d(np.asanyarray(data).T)
+        elif kind == "matrix":
+            # GMT can only accept a 2-D matrix which are signed integer (i), unsigned
+            # integer (u) or floating point (f) types. For other data types, we need to
+            # use virtualfile_from_vectors instead, which turns the matrix into list of
+            # vectors and allows for better handling of string type inputs (e.g. for
+            # datetime data types).
+            _data = (data,)
+            if data.dtype.kind not in "iuf":
+                _virtualfile_from = self.virtualfile_from_vectors
```
weiji14, line 1823: Missing test coverage for these lines?

pygmt/helpers/utils.py
seisman: > * Should we extend `kind="matrix"` to 3-D numpy.ndarray objects?

It's unclear if we can pass a 3-D numpy array yet. Even if we can, it means more work, since in `virtualfile_from_vectors`/`virtualfile_from_matrix`, we explicitly check if the ndim=1 or 2. So, better to revisit the 3-D numpy array support in a separate PR if necessary.

> * I'm hesitant to match based on `np.ndarray()` class only, because while it's not widely advertised, checking for `data is not None` would have caught other array types implementing the `__array_function__` protocol (see [NEP18](https://numpy.org/neps/nep-0018-array-function-protocol.html)), but checking for `isinstance(data, np.ndarray)` would exclude those. That said, I'm wondering if the fallback to `kind="vectors"` might just work, but need to test this out more extensively.

Yes, I was thinking about checking `__array_interface__` in https://github.com/GenericMappingTools/pygmt/pull/3351#discussion_r1703130372.

pygmt/clib/session.py
seisman: I can add a test for it. Before adding more tests, I'm wondering if we should split the big "test_clib_virtualfiles.py" file (with more than 500 lines) into separate smaller test files, i.e., one test file for each Session method. 

- `test_clib_virtualfile_in`
- `test_clib_virtualfile_from_vectors`
- `test_clib_virtualfile_from_matrix`
- `test_clib_open_virtualfile`

weiji14: We've split it before in #2784, so yes, ok to split it again :laughing:

seisman: > We've split it before in #2784, so yes, ok to split it again 😆

Done in #3512.

pygmt/helpers/utils.py
seisman: Done in 4a4f192d71f2754306a80858b6789c93d2fefd08, although other array-like objects are not tested.

pygmt/clib/session.py
seisman: I've added a test in cfa32ed8e5558ac3215366949b504e83ad2fa4ca to cover this line.

pygmt/tests/test_clib_virtualfile_in.py
```diff
@@ -101,3 +103,27 @@ def test_virtualfile_in_fail_non_valid_data(data):
                 z=data[:, 2],
                 data=data,
             )
+
```
seisman, line 106: This test is added to address https://github.com/GenericMappingTools/pygmt/pull/3351#discussion_r1798573532.

This new test passes a string dtype numpy array into GMT C API, which contains longitude/latitude strings. The data kind is `"matrix"`, but since the data dtype is not in `iuf`, PyGMT will use `virtualfile_from_vectors` rather than `virtualfile_from_matrix` to pass the data into GMT C API. Ideally, we should check that the `virtualfile_from_vectors` is called once and `virtualfile_from_matrix` is not called, but I find it technically complicated with `unittest.mock`, so I'll leave it untested.

The function is marked as xfail with GMT 6.5 due to a newly found upstream bug at https://github.com/GenericMappingTools/gmt/pull/8600.

weiji14: Good catch, and thanks for adding the test!

weiji14 (approved):

PR #3358, opened 2024-07-25
Add a test for plotting global grid without the redundant 360 longitude

### LINKED ISSUE #3331: grdimage does not work correctly when the redundant 360 E latitude is not included

grdimage works fine for global grids that include the redundant 0 and 360 E longitudes. However, if you create a grid that doesn't include the redundant 360 E, the returned image is all black.

I don't know if this worked before or not. For my purposes, it would be useful to use grdimage with global arrays that exclude 360 E and perhaps 90 S. The reason for this is that the arrays used for creating spherical harmonic transforms often don't require these values.

```python
import numpy as np
import xarray as xr
import pygmt

# grid that includes redundant 360 E
grid = np.random.rand(181, 361)
lat = np.arange(90, -91, -1)
lon = np.arange(0, 361, 1)

da = xr.DataArray(grid, coords=[('lat', lat), ('lon', lon)])
da.gmt.registration = 0
da.gmt.gtype = 1
                     
figure = pygmt.Figure()
figure.grdimage(da, projection='W0/10c', region='g')
figure.show()

# grid that does not include redundant 360 E
grid2 = np.random.rand(181, 360)
lat2 = np.arange(90, -91, -1)
lon2 = np.arange(0, 360, 1)

da2 = xr.DataArray(grid2, coords=[('lat', lat2), ('lon', lon2)])
da2.gmt.registration = 0
da2.gmt.gtype = 1
                     
figure2 = pygmt.Figure()
figure2.grdimage(da2, projection='W0/10c', region='g')
figure2.show()
```

### Full error message

```python-traceback
There is no error. The image is created, but it is all black.
```

### System information

```bash
PyGMT information:
  version: v0.12.0
System information:
  python: 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0]
  executable: /home/mrak/miniforge3/envs/py312/bin/python
  machine: Linux-6.9.7-200.fc40.x86_64-x86_64-with-glibc2.39
Dependency information:
  numpy: 1.26.4
  pandas: 2.2.2
  xarray: 2024.3.0
  netCDF4: 1.6.5
  packaging: 24.0
  contextily: None
  geopandas: None
  ipython: None
  rioxarray: None
  ghostscript: 10.03.0
GMT library information:
  binary version: 6.5.0
  cores: 20
  grid layout: rows
  image layout: 
  library path: /home/mrak/miniforge3/envs/py312/lib/libgmt.so
  padding: 2
  plugin dir: /home/mrak/miniforge3/envs/py312/lib/gmt/plugins
  share dir: /home/mrak/miniforge3/envs/py312/share/gmt
  version: 6.5.0
```

seisman:

This PR adds a test for the bug reported in https://github.com/GenericMappingTools/pygmt/issues/3331. See https://github.com/GenericMappingTools/pygmt/issues/3331#issuecomment-2387586029 for the comprehensive tests for when the bug can be triggered. This PR only test the central longitude=120 case. The baseline image is created by saving `da3`/`da4` into a netCDF file and passing the netCDF files directly to `Fig.grdimage`.

The new test only passes with the upstream fix in https://github.com/GenericMappingTools/gmt/pull/8554. So, the test is marked as xfail for GMT<=6.5.0.

The test passes in the "GMT Dev Tests" workflow: https://github.com/GenericMappingTools/pygmt/actions/runs/11138377051?pr=3358

Closes #3331.

### DIFF
--- a/pygmt/tests/test_grdimage.py
+++ b/pygmt/tests/test_grdimage.py
@@ -252,3 +252,37 @@ def test_grdimage_imgout_fails(grid):
         fig.grdimage(grid, img_out="out.png")
     with pytest.raises(GMTInvalidInput):
         fig.grdimage(grid, A="out.png")
+
+
+@pytest.mark.mpl_image_compare()
+def test_grdimage_grid_no_redunant_360():
+    """
+    Test that global grids with and without redundant 360/0 longitude values work.
+
+    Test for https://github.com/GenericMappingTools/pygmt/issues/3331.
+    """
+    # Global grid [-180, 180, -90, 90] with redundant longitude at 180/-180
+    da1 = load_earth_relief(region=[-180, 180, -90, 90])
+    # Global grid [0, 360, -90, 90] with redundant longitude at 0/360
+    da2 = load_earth_relief(region=[0, 360, -90, 90])
+
+    # Global grid [-180, 180, -90, 90] without redundant longitude at -180/180
+    da3 = da1[:, 0:360]
+    da3.gmt.registration, da3.gmt.gtype = 0, 1
+    assert da3.shape == (181, 360)
+    assert da3.lon.to_numpy().min() == -180.0
+    assert da3.lon.to_numpy().max() == 179.0
+
+    # Global grid [0, 360, -90, 90] without redundant longitude at 0/360
+    da4 = da2[:, 0:360]
+    da4.gmt.registration, da4.gmt.gtype = 0, 1
+    assert da4.shape == (181, 360)
+    assert da4.lon.to_numpy().min() == 0.0
+    assert da4.lon.to_numpy().max() == 359.0
+
+    fig = Figure()
+    kwdict = {"projection": "W120/10c", "region": "g", "frame": "+tlon=120"}
+    fig.grdimage(da3, **kwdict)
+    fig.shift_origin(xshift="w+2c")
+    fig.grdimage(da4, **kwdict)
+    return fig
--- a/pygmt/tests/test_grdimage.py
+++ b/pygmt/tests/test_grdimage.py
@@ -5,7 +5,9 @@ Test Figure.grdimage.
 import numpy as np
 import pytest
 import xarray as xr
+from packaging.version import Version
 from pygmt import Figure
+from pygmt.clib import __gmt_version__
 from pygmt.datasets import load_earth_relief
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers.testing import check_figures_equal
@@ -254,6 +256,10 @@ def test_grdimage_imgout_fails(grid):
         fig.grdimage(grid, A="out.png")
 
 
+@pytest.mark.xfail(
+    condition=Version(__gmt_version__) <= Version("6.5.0"),
+    reason="Upstream bug fixed in https://github.com/GenericMappingTools/gmt/pull/8554",
+)
 @pytest.mark.mpl_image_compare()
 def test_grdimage_grid_no_redunant_360():
     """
--- a/pygmt/tests/test_grdimage.py
+++ b/pygmt/tests/test_grdimage.py
@@ -269,17 +269,17 @@ def test_grdimage_grid_no_redunant_360():
     """
     # Global grid [-180, 180, -90, 90] with redundant longitude at 180/-180
     da1 = load_earth_relief(region=[-180, 180, -90, 90])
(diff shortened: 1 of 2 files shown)


seisman:
This test uses the `xrgrid` grid defined as:
```
@pytest.fixture(scope="module", name="xrgrid")
def fixture_xrgrid():
    """
    Create a sample xarray.DataArray grid for testing.
    """
    longitude = np.arange(0, 360, 1)
    latitude = np.arange(-89, 90, 1)
    x = np.sin(np.deg2rad(longitude))
    y = np.linspace(start=0, stop=1, num=179)
    data = y[:, np.newaxis] * x
    return xr.DataArray(
        data,
        coords=[
            ("latitude", latitude, {"units": "degrees_north"}),
            ("longitude", longitude, {"units": "degrees_east"}),
        ],
        attrs={"actual_range": [-1, 1]},
    )
```
It's not a real global grid, which doesn't have the 360° longitude defined and also the north/south poles are missing. I'll open a separate PR to refine the `xrgrid` into a true global grid, so mark this PR "Draft" now.

weiji14:
> It's not a real global grid, which doesn't have the 360° longitude defined and also the north/south poles are missing. I'll open a separate PR to refine the `xrgrid` into a true global grid, so mark this PR "Draft" now.

Maybe best if we keep the original `xrgrid` that has data from 0-359° longitude (no redundant data at 360°), and a new grid from 0-360° longitude (redundant data at 360°). There isn't really a correct 'real' global grid here, we just need to know GMT/PyGMT can handle both.

seisman:
> > It's not a real global grid, which doesn't have the 360° longitude defined and also the north/south poles are missing. I'll open a separate PR to refine the `xrgrid` into a true global grid, so mark this PR "Draft" now.
> 
> Maybe best if we keep the original `xrgrid` that has data from 0-359° longitude (no redundant data at 360°), and a new grid from 0-360° longitude (redundant data at 360°). There isn't really a correct 'real' global grid here, we just need to know GMT/PyGMT can handle both.

Actually the original issue is not due to the missing 360 longitude. For any grids that is not exactly 0-360, the same issue occurs for.some specific central longitude and projection. If interested, you can play with the script in https://github.com/GenericMappingTools/pygmt/issues/3331#issuecomment-2250769483.

pygmt/tests/test_grdimage.py
```diff
@@ -252,3 +254,41 @@ def test_grdimage_imgout_fails(grid):
...
+@pytest.mark.mpl_image_compare()
+def test_grdimage_grid_no_redunant_360():
+    """
+    Test that global grids with and without redundant 360/0 longitude values work.
+
+    Test for https://github.com/GenericMappingTools/pygmt/issues/3331.
+    """
+    # Global grid [-180, 180, -90, 90] with redundant longitude at 180/-180
+    da1 = load_earth_relief(region=[-180, 180, -90, 90])
+    # Global grid [0, 360, -90, 90] with redundant longitude at 0/360
```
yvonnefroehlich, line 272:
```suggestion
    # Global grid [0, 360, -90, 90] with redundant longitude at 360/0
```

pygmt/tests/test_grdimage.py
```diff
@@ -252,3 +254,41 @@ def test_grdimage_imgout_fails(grid):
...
+    da2 = load_earth_relief(region=[0, 360, -90, 90])
+
+    # Global grid [-180, 180, -90, 90] without redundant longitude at -180/180
```
yvonnefroehlich, line 275:
```suggestion
    # Global grid [-180, 179, -90, 90] without redundant longitude at 180/-180
```

pygmt/tests/test_grdimage.py
```diff
@@ -252,3 +254,41 @@ def test_grdimage_imgout_fails(grid):
...
+    da3 = da1[:, 0:360]
+    da3.gmt.registration, da3.gmt.gtype = 0, 1
+    assert da3.shape == (181, 360)
+    assert da3.lon.to_numpy().min() == -180.0
+    assert da3.lon.to_numpy().max() == 179.0
+
+    # Global grid [0, 360, -90, 90] without redundant longitude at 0/360
```
yvonnefroehlich, line 282:
```suggestion
    # Global grid [0, 359, -90, 90] without redundant longitude at 360/0
```

PR #3369, opened 2024-08-04
Session.virtualfile_in: Deprecate parameter 'required_z'. Use 'mincols' instead (will be removed in v0.20.0)

seisman:
Related to #3836.

This PR removes the `required_z` parameter and add a new parameter `ncols`. `required_z=True` means `ncols=3`.


pygmt/helpers/utils.py
```diff
@@ -115,6 +115,7 @@ def _validate_data_input(
     GMTInvalidInput
         If the data input is not valid.
     """
+    required_z = ncols >= 3
```
seisman, line 118: This is just a temporary fix. Will refactor `_validate_data_input` in #3818.

weiji14 (approved):
Debating on whether to call the parameter `mincols` instead of `ncols`, but leave it up to you to decide.

PR #3313, opened 2024-07-05
Fix typo in AUTHORS.md

yvonnefroehlich:
**Description of proposed changes**

I am wondering if the BibTeX entry (https://github.com/GenericMappingTools/pygmt?tab=readme-ov-file#citing-pygmt) also works with single curly braces (https://de.overleaf.com/learn/latex/Bibliography_management_with_bibtex#Enter_\(\mathrm{Bib\TeX}\).

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #857, opened 2021-02-09
Add affiliation and ORCID to AUTHORS.md

liamtoney:
**Description of proposed changes**

Add affiliation and ORCID to `AUTHORS.md`, in light of #726. Others can jump on this PR to add their info too.

PR #3373, opened 2024-08-05
Format the author list into a table

seisman:
Just change the list of authors into a table, which looks better IMHO.

**Preview**

- Author list (current main): https://github.com/GenericMappingTools/pygmt/blob/main/AUTHORS.md
- Author table (this PR): https://github.com/GenericMappingTools/pygmt/blob/author-table/AUTHORS.md

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -8,20 +8,22 @@ Hawaiʻi at Mānoa.
 The following people have contributed code and/or documentation to the project
 (alphabetically by name) and are considered to be "PyGMT Developers":
 
-* [Abhishek Anant](https://twitter.com/itsabhianant) | [0000-0002-5751-2010](https://orcid.org/0000-0002-5751-2010) | Unaffiliated
-* [Andre L. Belem](https://github.com/andrebelem) | [0000-0002-8865-6180](https://orcid.org/0000-0002-8865-6180) | Fluminense Federal University, Brazil
-* [Dongdong Tian](https://seisman.info/) | [0000-0001-7967-1197](https://orcid.org/0000-0001-7967-1197) | China University of Geosciences, China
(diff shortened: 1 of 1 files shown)


yvonnefroehlich (approved):
> Just change the list of authors into a table, which looks better IMHO.

Totally agree! Actually I was thinking about this when fixing the typo in #3313.

AUTHORS.md
```diff
@@ -8,20 +8,22 @@ Hawaiʻi at Mānoa.
 The following people have contributed code and/or documentation to the project
 (alphabetically by name) and are considered to be "PyGMT Developers":
```
yvonnefroehlich, line 9:
```suggestion
(alphabetically by first name) and are considered to be "PyGMT Developers":
```

weiji14: This is deliberately just 'name' (or full name) because some cultures don't have first/last names (see https://github.com/GenericMappingTools/pygmt/pull/857#discussion_r574098862).

weiji14:
I'm aware that some people on the list have new affiliations, should we update them here, or separately during the release since it will involve a corresponding change to the CITATION.cff file?

seisman:
> I'm aware that some people on the list have new affiliations, should we update them here, or separately during the release since it will involve a corresponding change to the CITATION.cff file?

Definitely in a separate PR since we need to ping them to confirm with the affiliation changes.

AUTHORS.md
yvonnefroehlich: Ah, that's a good point!
Then just leave it as is 🙂.

seisman:
> > I'm aware that some people on the list have new affiliations, should we update them here, or separately during the release since it will involve a corresponding change to the CITATION.cff file?
> 
> Definitely in a separate PR since we need to ping them to confirm with the affiliation changes.

I'm not sure we want to take the time to update their affiliations. In fact, the affiliations may be outdated but still reflect the affiliations **at the time when they contribute**. People can still update their affiliations when they come back to contribute to PyGMT again.

PR #3380, opened 2024-08-07
Determine the minimum required versions of dependencies from package metadata for docs

seisman:

It's possible to determine the minimum required versions from the package metadata using codes like below. So that we don't have to update the `doc/minversions.md` file everytime we bump the minimum required versions (e.g., #3372). 
```python
>>> import importlib
>>> from packaging.requirements import Requirement
>>> requirements = {
...:     Requirement(requirement).name: str(Requirement(requirement).specifier)
...:     for requirement in importlib.metadata.requires("pygmt")
...: }
...: requirements.update(
...:     {
...:         "python": importlib.metadata.metadata("pygmt")["Requires-Python"],
...:         #"gmt": f">={required_gmt_version}",
...:     }
...: )

>> requirements
{'numpy': '>=1.24',
 'pandas': '>=1.5',
 'xarray': '>=2022.09',
 'netCDF4': '',
 'packaging': '',
 'contextily': '',
 'geopandas': '',
 'IPython': '',
 'rioxarray': '',
 'python': '>=3.10'}
```

This PR uses the same codes to determine the minimum required versions and store in the `requirements` dictionary, which can be used in the documentation using syntax like ``{{ requires.python }}``.

**Preview**:

- https://pygmt-dev--3380.org.readthedocs.build/en/3380/minversions.html
- https://pygmt-dev--3380.org.readthedocs.build/en/3380/install.html#which-python

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -3,18 +3,25 @@ Sphinx documentation configuration file.
 """
 
 import datetime
-from importlib.metadata import metadata
+import importlib
 
-# ruff: isort: off
-from sphinx_gallery.sorting import ExplicitOrder, ExampleTitleSortKey
-from pygmt.clib import required_gmt_version
+from packaging.requirements import Requirement
 from pygmt import __commit__, __version__
+from pygmt.clib import required_gmt_version
 from pygmt.sphinx_gallery import PyGMTScraper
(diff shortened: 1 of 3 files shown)


weiji14 (approved):
So much more automation, nice!

PR #3383, opened 2024-08-07
doc: Simplify the "minimum required versions" page using MyST customized URL schemes

seisman:

MyST supports customized URLs.

References:

- https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#customising-external-url-resolution
- https://myst-parser.readthedocs.io/en/latest/configuration.html#frontmatter-local-configuration

**Preview**: https://pygmt-dev--3383.org.readthedocs.build/en/3383/minversions.html

This page doesn't change but the markdown file is much shorter!

### DIFF
--- a/doc/minversions.md
+++ b/doc/minversions.md
@@ -1,7 +1,7 @@
 ---
 myst:
   # Customized URI schemes that are converted to external links
-  # Reference:
+  # References:
   # - https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#customising-external-url-resolution
   # - https://myst-parser.readthedocs.io/en/latest/configuration.html#frontmatter-local-configuration
   url_schemes:
--- a/doc/minversions.md
+++ b/doc/minversions.md
@@ -1,3 +1,20 @@
+---
+myst:
+  # Customized URI schemes that are converted to external links
+  # Reference:
+  # - https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#customising-external-url-resolution
+  # - https://myst-parser.readthedocs.io/en/latest/configuration.html#frontmatter-local-configuration
+  url_schemes:
+    http: null
+    https: null
+    tag:
+      url: "https://github.com/GenericMappingTools/pygmt/releases/tag/{{path}}"
+      title: "{{path}}"
+    doc:
+      url: "https://www.pygmt.org/{{path}}"
+      title: "Docs"
+---
+
 # Minimum Supported Versions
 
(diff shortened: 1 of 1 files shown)


doc/minversions.md
```diff
@@ -1,3 +1,20 @@
+---
+myst:
+  # Customized URI schemes that are converted to external links
```
yvonnefroehlich, line 3: Hm. Based on how I understand the documentation so far and being consistent with `url_schemes:`, using "URL" seem to be better, here. If we use "URI" I feel it can make sense to introduce this shortcut.

```suggestion
  # Customized URL schemes that are converted to external links
```

doc/minversions.md
```diff
@@ -1,3 +1,20 @@
...
+  # Reference:
```
yvonnefroehlich, line 4:
```suggestion
  # References:
```

yvonnefroehlich, line 4:
```suggestion
  # References:
```

doc/minversions.md
michaelgrund: For me it seems URI is still the right term here since an URL is already a link (a location) while an URI is just the identifier that is then converted to an URL.

seisman: The MyST documentation uses URI: https://myst-parser.readthedocs.io/en/latest/configuration.html

For the difference between URI and URL, see https://www.godaddy.com/resources/ae/skills/uri-vs-url.

I agree with @michaelgrund that URI is the right term.

yvonnefroehlich: Hm, looks like the MyST documentation is not 100 % consistent here 🙃:

at https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#customising-external-url-resolution
> To specify a custom list of URL schemes, you can set the myst_url_schemes configuration option. By default this is set to ["http", "https", "ftp", "mailto"].

but at https://myst-parser.readthedocs.io/en/latest/configuration.html
> myst_url_schemes  |  URI schemes that are converted to external links (default: {'http': None, 'https': None, 'mailto': None, 'ftp': None})

seisman: Anyway, it's not one comment line and it doesn't hurt if it's correct or not.

yvonnefroehlich: Reading https://www.godaddy.com/resources/ae/skills/uri-vs-url
> URIs can be further divided into two categories: URLs and URNs. [...]
> The key difference is: all URLs are URIs, but not all URIs are URLs. 

So, let's keep URIs here.

yvonnefroehlich (approved):

PR #3396, opened 2024-08-13
pygmt.set_display: Fix the bug that `method=None` doesn't reset to the default display method

seisman:

`pygmt.set_display(method=None)` should revert the display method to the default value, but it doens't work. Here is a minimal example to reproduce the issue:
```python
In [1]: import pygmt
In [2]: fig = pygmt.Figure()
In [3]: fig.basemap(region=[0, 10, 0, 10], projection="X4c/4c", frame=True)
In [4]: fig.show()  # Display the image in an external viewer
In [5]: pygmt.set_display(method="none")  # Disable image preview 
In [6]: fig.show()  # Image is not displayed
In [7]: pygmt.set_display(method=None)   # Should revert to the default display method
In [8]: fig.show()  # Image is not displayed. WRONG!
```
In the last `fig.show()` call, the image should be displayed, but it doesn't. It's because `method=None` is ignored in `pygmt.set_display`. 

This PR fixes the bug and does some refactorings.

- 24092848061522523847d8cce1234cf091e07586: Add the `_get_default_display_method` function so that the same logic of codes can be called when `method=None` is used.
- 6bd742e6aa2c0bb48e2b7b9adaa056cff06d96d2: Refactor `set_display` to fix the bug.

Ideally, we should add a test to check if `SHOW_CONFIG["method"]` matches the value in `pygmt.set_display(method="XXX")`. However, `SHOW_CONFIG` is a global variable, changing its value will affect other tests. So, no test is added.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -6,6 +6,7 @@ import base64
 import os
 from pathlib import Path
 from tempfile import TemporaryDirectory
+from typing import Literal
 
 try:
     import IPython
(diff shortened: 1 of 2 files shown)


pygmt/figure.py
```diff
@@ -442,6 +463,8 @@ def show(self, dpi=300, width=500, method=None, waiting=0.5, **kwargs):
 
         # Set the display method
         if method is None:
+            if SHOW_CONFIG["method"] is None:
+                SHOW_CONFIG["method"] = _get_default_display_method()
```
seisman, line 467: `SHOW_CONFIG["method"]` is set to the default display method when `Figure.show` is called the first time.

pygmt/figure.py
```diff
@@ -29,23 +30,43 @@
...
+
+    The function checks the current environment and determine the most suitable method
+    to display preview images when calling :meth:`pygmt.Figure.show`. Valid display
+    methods are:
+
+    - ``"external"``: External PDF preview using the default PDF viewer
+    - ``"notebook"``: Inline PNG preview in the current notebook
+    - ``"none"``: Disable image preview
+
+    The default display method is ``"notebook"`` in Jupyter notebook environment, and
```
michaelgrund, line 51:
```suggestion
    The default display method is ``"notebook"`` in the Jupyter notebook environment, and
```

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final review.

pygmt/tests/test_figure.py
```diff
@@ -373,12 +373,31 @@ def test_figure_display_external():
     fig.show(method="external")
 
 
-def test_figure_set_display_invalid():
+class TestSetDisplay:
```
weiji14, line 376: We don't often use classes for the pytest unit tests, is there a reason to use it here?

weiji14 (approved):

pygmt/tests/test_figure.py
seisman: Just want to group multiple tests so that it is easier to find all tests for testing a specific function.

weiji14: Ok, it looks like we're using `TestGetDefaultDisplayMethod` below, so ok with this.

PR #3379, opened 2024-08-06
Add sphinx extension myst-nb to enable writing executable Markdown notebooks

seisman:
**Description of proposed changes**

This PR adds the Sphinx extension [`myst-nb`](https://myst-nb.readthedocs.io/en/latest/index.html) to enable executing code cells in a file and insert the output directly into the file. The file format can be (1) ReST; (2) Plain Markdown; (3) Jupyter Notebooks; (4) MyST text-based notebooks. See https://myst-nb.readthedocs.io/en/latest/authoring/basics.html for more details. 

The [text-based MyST notebook](https://myst-nb.readthedocs.io/en/latest/authoring/text-notebooks.html) is the most appealing format. To convert a plain Markdown file into a MyST notebook, we just need to do two things:

1. Add metadata at the top of a markdown file. The simplest example is:
```
---
file_format: mystnb
---
```
2. Write codes using the following syntax 
~~~
```{code-cell}
a = "This is some"
b = "Python code!"
print(f"{a} {b}")
```
~~~

This PR also updates the `install.md` file as an example showing how the extension works. In the installation guides, we provide a test script and include an external image (added in #3249) to show how the output looks like. It's not ideal since the image may be outdated when we update the test script. The `myst-nb` extension can execute the codes and directly insert the output image below the codes, so the codes and the image are always consistent.

**Preview**: https://pygmt-dev--3379.org.readthedocs.build/en/3379/install.html#testing-your-install

Notes:

1. The `myst-parser` extension is removed because it's required by `myst-nb` and is automatically enabled by `myst-nb`.
2. The `myst-nb` extension is necessary for writing documentation that needs images. For example, an image is needed to explain justification codes well.
3. This extension can't replace sphinx-gallery now, but maybe it can in the future (xref https://github.com/executablebooks/MyST-NB/issues/396)
4. `myst-nb` is one of the subprojects of [The Executable Books Project](https://github.com/executablebooks). We're also using their `sphinx-copybutton` and `sphinx-design` extensions. Maybe one day we should consider migrating to their [Jupyter Book](https://jupyterbook.org/) project, which is built on top of Sphinx and a collection of Sphinx extensions.

PR #3206, opened 2024-04-27
Refactor to improve the user experience with non-ASCII characters

seisman:
See https://github.com/GenericMappingTools/pygmt/issues/2204#issuecomment-2079661330 for background and the issue.

This PR solves the issue by:

- Generating the mapping dictionary based on the mapping tables in https://github.com/seisman/GMT-octal-codes
- Add the encoding tables to the documentation

**Preview**: 
The tables of supported encodings are at
https://pygmt-dev--3206.org.readthedocs.build/en/3206/techref/encodings.html.

This page is in the "Technical Reference" section, as discussed in #2834.

**Note to maintainers**

Below is the script used to generate the Markdown tables in the documentation
```python
from pygmt.encodings import charset

undefined = "\ufffd"
markdown = ""
for charset_name, mappings in charset.items():
    markdown += f"## Adobe {charset_name} Encoding\n\n"
    markdown += "| octal | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |\n"
    markdown += "|---|---|---|---|---|---|---|---|---|\n"
    for i in range(0o00, 0o400, 8):
        chars = [mappings.get(j, undefined) for j in range(i, i + 8)]
        if chars == [undefined] * 8:
            continue
        chars = [f"&#x{ord(char):04x};" for char in chars]
        row = f"\\{i:03o}"[:-1] + "x"
        markdown += f"| **{row}** | {' | '.join(chars)} |\n"
    markdown += "\n"

print(markdown)
```

PR #3409, opened 2024-08-22
DOC: Generate the charset tables dynamically from codes

seisman:

After adding the `myst-nb` extension in https://github.com/GenericMappingTools/pygmt/pull/3379, it's possible to generate the charset tables dynamically from codes.

**Preview**: 

- This branch: https://pygmt-dev--3409.org.readthedocs.build/en/3409/techref/encodings.html
- Main branch: https://www.pygmt.org/dev/techref/encodings.html

**Pros**:

1. No need to maintain the charset tables manually
2. Make sure that the documentation is consistent with the codes

**Cons**:

1. Maybe a little slower in building the documentation [Building docs takes the same time (105 seconds) on [the main branch](https://github.com/GenericMappingTools/pygmt/actions/runs/10510957792/job/29120803421) and [this PR](https://github.com/GenericMappingTools/pygmt/actions/runs/10510983347/job/29120886153?pr=3409), so the changes in this PR has negligible effects].
2. ruff can't format the codes in Markdown https://github.com/astral-sh/ruff/issues/8237, so we need to format the codes manually.


doc/techref/encodings.md
```diff
@@ -1,3 +1,35 @@
...
+
+```{code-cell}
+---
+tags: [remove-input]
+---
+from IPython.display import display, Markdown
+from pygmt.encodings import charset
+
+
+def get_charset_mdtable(name):
```
seisman, line 13: This function is modified from the original script at https://github.com/GenericMappingTools/pygmt/pull/3206#issue-2267029976

doc/conf.py
```diff
@@ -61,6 +61,9 @@
     "requires": requirements,
 }
 
+# MyST-NB configurations: https://myst-nb.readthedocs.io/en/latest/configuration.html
+nb_render_markdown_format = "myst"  # The format to use for text/markdown rendering
+
```
seisman, line 66: The default value is `'commonmark'`, which doesn't support tables.

doc/_static/style.css
```diff
@@ -203,3 +203,19 @@ a.copybtn {
 .sphx-glr-single-img {
     max-width: 80%!important;
 }
+
+/*
```
seisman, line 207: Here is the CSS styles from myst-nb.
```
div.cell_output tr,
div.cell_output th,
div.cell_output td {
  text-align: right;
  vertical-align: middle;
  padding: 0.5em 0.5em;
  line-height: normal;
  white-space: normal;
  max-width: none;
  border: none;
}
```
The `text-align` is set to `right`, which makes sense for displaying objects like `pandas.DataFrame` (an example at https://myst-nb.readthedocs.io/en/latest/render/format_code_cells.html#remove-stdout-or-stderr).

In MyST, the table cell alignment can be controlled using `:` characters (https://myst-parser.readthedocs.io/en/latest/syntax/tables.html), which is done by assigning `text-left`/`text-center`/`text-right` to cells. These CSS classes are not defined in Sphinx_rtd_theme, so we have to define them here.

michaelgrund (approved):

PR #3413, opened 2024-08-26
Remove the non-official GMT wrappers from README

seisman:
These non-official GMT wrappers have not been updated/maintained since the initiation of the official PyGMT project in 2017. 

It makes sense to mention these projects in 2017, but now is 2024. I prefer to remove them from README.


michaelgrund (approved):

yvonnefroehlich (commented):
Makes sense to me.

PR #3374, opened 2024-08-05
Improve the release checklist based on experience with v0.13.0 release

seisman:
This PR updates the release checklist. 

Here are the main changes in this PR:

- Add a placeholder for release DOI
- Review the "PyGMT Team" as early as possible
- Check deprecations before checking CI runs
- Simplify the task for working on the changelog entry
- Check misspellings as late as possible
- Draft the announcement as early as possible
- Rename the zip file before uploading to Zenodo
- ~~Check two options when making a release~~

PR #3283, opened 2024-06-04
Add pre-commit config with pre-commit-hooks

weiji14:
**Description of proposed changes**

Add [pre-commit](https://github.com/pre-commit/pre-commit) hooks to enforce certain rules on all files in the repo.

The following rules are applied:
- check-added-large-files
- check-yaml
- end-of-file-fixer
- trailing-whitespace
- forbid-crlf
- remove-crlf
- chmod

Note that `pre-commit run --all-files` has been added to the Makefile, so running `make format` locally or `/format` on a GitHub PR will apply the formatting rules.

TODO:
- [x] Add a `.pre-commit-config.yaml` file for standard pre-commit hooks
- ~Configure [pre-commit-ci](https://pre-commit.ci) GitHub App~ not using, doing `pre-commit` manually ourselves

References:
- https://pre-commit.com/index.html#quick-start

Fixes #1593

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3414, opened 2024-08-26
Add pre-commit.ci configurations to enable auto updates of hooks

seisman:

Run `pre-commit autoupdate` command in the Style Checks workflow, to check if there are new versions of pre-commit hooks. The workflow will fail if there are updates. Then we just need to open a PR to manually update the workflow.

Address https://github.com/GenericMappingTools/pygmt/pull/3374#discussion_r1730385091.

### DIFF
--- a/.github/workflows/style_checks.yaml
+++ b/.github/workflows/style_checks.yaml
@@ -42,6 +42,9 @@ jobs:
            make check
            pre-commit run --all-files
 
+      - name: Update pre-commit hooks
+        run: pre-commit autoupdate
+
       - name: Ensure example scripts have at least one code block separator
         run: |
           git ls-files 'examples/**/*.py' | xargs grep --files-without-match '# %%' > output.txt
--- a/.github/workflows/style_checks.yaml
+++ b/.github/workflows/style_checks.yaml
@@ -12,6 +12,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
+  workflow_dispatch:
   # Schedule daily tests
   schedule:
     - cron: '0 0 * * *'
--- a/.github/workflows/style_checks.yaml
+++ b/.github/workflows/style_checks.yaml
@@ -42,6 +42,7 @@ jobs:
         run: |
           pre-commit autoupdate
           if [[ $(git ls-files -m) ]]; then git --no-pager diff HEAD; exit 1; fi
+        if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/style_checks.yaml
```diff
@@ -37,6 +37,11 @@ jobs:
           python -m pip install ruff pre-commit
           python -m pip list
 
+      - name: Check updates of pre-commit hooks
+        run: |
+          pre-commit autoupdate
```
weiji14, line 42: Maybe run the autoupdate line only on cron jobs? So that PRs don't get slowed down by sudden Style Check changes.

seisman: OK. I've added the if-condition so that the autoupdate command only runs on schedule or workflow_dispatch events. It means we have to check https://github.com/GenericMappingTools/pygmt/actions/workflows/style_checks.yaml to know if there are new versions for hooks. Maybe we should automatically open a PR if there are updates?

weiji14: > Maybe we should automatically open a PR if there are updates?

Maybe we could just use pre-commit.ci as was tried in https://github.com/GenericMappingTools/pygmt/pull/3283? The automation would require quite a bit of code to set up. Ideally we could use dependabot but looking at https://github.com/dependabot/dependabot-core/issues/1524, it's still not working yet.

seisman: I'm OK with enabling pre-commit.ci for auto update.

weiji14: Ok, add these lines to `.pre-commit-config.yaml `:

```
# https://pre-commit.ci/#configuration
ci:
  autofix_prs: false
  autoupdate_schedule: quarterly
```

seisman: The pre-commit.ci App is also installed https://github.com/GenericMappingTools/pygmt/settings/installations.

weiji14 (approved):

seisman:
pre-commit.ci fails in the main branch https://results.pre-commit.ci/run/github/85352251/1724751409.rGgeqA4fQ22hZH-xAZG-Nw, as mentioned in https://github.com/GenericMappingTools/pygmt/pull/3283#discussion_r1625402681.

It seems we can't change how files are cloned in pre-commit.ci and also can't disable pre-commit.ci in the main branch. I guess the only option is removing the `chmod` hook from `.pre-commit-config.yaml` and add it back to the style checks workflow.

PR #3415, opened 2024-08-27
CI: Skip the chmod hook in pre-commit.ci

seisman:
Address https://github.com/GenericMappingTools/pygmt/pull/3414#issuecomment-2312120942.

Patches #3414.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,8 +13,6 @@ repos:
   hooks:
     - id: forbid-crlf
     - id: remove-crlf
-    - id: chmod
-      args: ['644']
 
 # https://pre-commit.ci/#configuration
 ci:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,8 @@ repos:
   hooks:
     - id: forbid-crlf
     - id: remove-crlf
+    - id: chmod
+      args: ['644']
 
 # https://pre-commit.ci/#configuration
 ci:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,5 +16,6 @@ repos:
 
 # https://pre-commit.ci/#configuration
 ci:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -13,8 +13,6 @@ repos:
   hooks:
     - id: forbid-crlf
     - id: remove-crlf
-    - id: chmod
-      args: ['644']
```
weiji14, line 17: You can add a `exclude: .pre-commit-config.yaml` line to exclude 644 permissions on this file if I'm not mistaken. See https://pre-commit.com/#config-exclude

weiji14: Oh wait sorry, I misread things. Let me see if there's another way...

seisman: Since we only care about if files have execute permission, so I thought `chmod -x` should work. Unfortunately the hook doesn't support it yet https://github.com/Lucas-C/pre-commit-hooks/blob/a30f0d816e5062a67d87c8de753cfe499672b959/pre_commit_hooks/chmod.py#L35-L40.

weiji14: Maybe we skip the `chmod` hook on pre-commit.ci using https://pre-commit.ci/#configuration-skip? Since we have the check on the `Style Checks' CI already.

.pre-commit-config.yaml
```diff
@@ -13,8 +13,6 @@ repos:
...
 
 # https://pre-commit.ci/#configuration
 ci:
```
weiji14, line 18:
```suggestion
ci:
    skip: [chmod]
```

.pre-commit-config.yaml
```diff
@@ -18,5 +18,6 @@ repos:
...
+    skip: [chmod]
```
seisman, line 21:
```suggestion
  skip: [chmod]
```

weiji14 (approved):

PR #3420, opened 2024-08-29
CI: Add a separate CI job on Ubuntu to test geopandas v0.x

seisman:

In #3247, we added codes to deal with compatibility with both GeoPandas v0.x and v1.x, but our CI jobs only install GeoPandas v1.x now, leaving the codes for geopandas v0.x untested (see https://app.codecov.io/gh/GenericMappingTools/pygmt/blob/main/pygmt%2Fhelpers%2Ftempfile.py#L146).

This PR adds a separate CI job in which geopandas v0.x is installed. The CI job has to use Python 3.11, otherwise other Python 3.10/3.12 jobs will be override by this one. The job also only runs on Ubuntu, to avoid using too many CI resources.

With the new CI job, the geopandas v0.x codes are covered and the code coverage increases by 0.32%.


.github/workflows/ci_tests.yaml
```diff
@@ -79,6 +79,14 @@ jobs:
             pandas-version: ''
             xarray-version: ''
             optional-packages: ' contextily geopandas ipython pyarrow rioxarray sphinx-gallery'
+          # The job below is for testing geopands v0.x on Ubuntu. Python 3.11 is used
+          # here, otherwise Python 3.10/3.12 jobs will be override by this one.
```
yvonnefroehlich, line 83:
```suggestion
          # The job below is for testing geopandas v0.x on Ubuntu. Python 3.11 is used
          # here, otherwise Python 3.10/3.12 jobs will be overwridden by this one.
```

.github/workflows/ci_tests.yaml
```diff
@@ -79,6 +79,14 @@ jobs:
             pandas-version: ''
             xarray-version: ''
             optional-packages: ' contextily geopandas ipython pyarrow rioxarray sphinx-gallery'
+          # The job below is for testing geopandas v0.x on Ubuntu. Python 3.11 is used
+          # here, otherwise Python 3.10/3.12 jobs will be overwridden by this one.
```
weiji14, line 83:
```suggestion
          # here, otherwise Python 3.10/3.12 jobs will be overridden by this one.
```

.github/workflows/ci_tests.yaml
```diff
@@ -79,6 +79,14 @@ jobs:
             pandas-version: ''
             xarray-version: ''
             optional-packages: ' contextily geopandas ipython pyarrow rioxarray sphinx-gallery'
+          # The job below is for testing geopandas v0.x on Ubuntu. Python 3.11 is used
```
weiji14, line 82: > This PR adds a separate CI job in which geopandas v0.x is installed. The CI job has to use Python 3.11, otherwise other Python 3.10/3.12 jobs will be override by this one. The job also only runs on Ubuntu, to avoid using too many CI resources.

Could we test geopandas v0.x in [`ci_tests_legacy.yaml`](https://github.com/GenericMappingTools/pygmt/blob/main/.github/workflows/ci_tests_legacy.yaml) instead? Or do we want to test it more regularly than that?

weiji14 (commented):
Thinking if we should generalize this to test minimum supported versions of all optional dependencies (contextily, geopandas, rioxarray, etc). `uv` has this `--resolution lowest-direct` flag (see [here](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy)) that "will use the lowest compatible versions for all direct dependencies, while using the latest compatible versions for all other dependencies". Would be nice if mamba had something similar.

seisman:
> Thinking if we should generalize this to test minimum supported versions of all optional dependencies (contextily, geopandas, rioxarray, etc). `uv` has this `--resolution lowest-direct` flag (see [here](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy)) that "will use the lowest compatible versions for all direct dependencies, while using the latest compatible versions for all other dependencies". Would be nice if mamba had something similar.

it sounds a good idea, but we are approaching the 10GB size limit of GitHub caches https://github.com/GenericMappingTools/pygmt/actions/caches.

.github/workflows/ci_tests.yaml
seisman: We can, but in coverage reports the geopandas v0.x would be uncovered since the Legacy workflow doesn't produce coverage reports.

seisman:
> Thinking if we should generalize this to test minimum supported versions of all optional dependencies (contextily, geopandas, rioxarray, etc).

We should note that we currently don't set the minimum required versions for the optional dependencies.

seisman:
The CI job with geopandas v0.x worked in previous runs (e.g., https://github.com/GenericMappingTools/pygmt/actions/runs/10694655883/job/29646773799?pr=3420) and suddenly failed now (e.g., https://github.com/GenericMappingTools/pygmt/actions/runs/10857583544/job/30146872146?pr=3420).

```
=================================== FAILURES ===================================
______________________ test_geopandas_plot_int64_as_float ______________________
[gw0] linux -- Python 3.11.10 /home/runner/micromamba/envs/pygmt/bin/python3.11

args = ()
kwargs = {'gdf_ridge':    PLATEID1  ...                                           geometry
0     101.0  ...  POLYGON ((-30.5584....72844, -34.56910 36.768...
2     101.0  ...  POLYGON ((-38.18994 34.45598, -38.15845 34.514...

[3 rows x 21 columns]}

    def wrapper(*args, **kwargs):
>       store.return_value[test_name] = obj(*args, **kwargs)

../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/pytest_mpl/plugin.py:125: 
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/pytest_mpl/plugin.py:125: in wrapper
    store.return_value[test_name] = obj(*args, **kwargs)
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/pytest_mpl/plugin.py:125: in wrapper
    store.return_value[test_name] = obj(*args, **kwargs)
../pygmt/tests/test_geopandas.py:234: in test_geopandas_plot_int64_as_float
    fig.plot(
../pygmt/helpers/decorators.py:609: in new_module
../pygmt/helpers/decorators.py:773: in new_module
../pygmt/src/plot.py:258: in plot
    with lib.virtualfile_in(
../../../../micromamba/envs/pygmt/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
../pygmt/helpers/tempfile.py:164: in tempfile_from_geojson
    geojson.to_file(**ogrgmt_kwargs)
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/geopandas/geodataframe.py:1249: in to_file
    _to_file(self, filename, driver, schema, index, **kwargs)
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/geopandas/io/file.py:610: in _to_file
    _to_file_fiona(df, filename, driver, schema, crs, mode, **kwargs)
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/geopandas/io/file.py:641: in _to_file_fiona
    colxn.writerecords(df.iterfeatures())
../../../../micromamba/envs/pygmt/lib/python3.11/site-packages/fiona/collection.py:541: in writerecords
    self.session.writerecs(records, self)
fiona/ogrext.pyx:1666: in fiona.ogrext.WritingSession.writerecs
    ???
fiona/ogrext.pyx:788: in fiona.ogrext.OGRFeatureBuilder.build
    ???

>   ???
E   OverflowError: value too large to convert to int

fiona/ogrext.pyx:243: OverflowError
```
The most notable difference is fiona v1.9 in old runs and v1.10 in the latest runs.

weiji14:
> > Thinking if we should generalize this to test minimum supported versions of all optional dependencies (contextily, geopandas, rioxarray, etc).
> 
> We should note that we currently don't set the minimum required versions for the optional dependencies.

Yeah, should we come up with a policy for this? E.g. extend [SPEC 0](https://scientific-python.org/specs/spec-0000/) with a recommendation like:

3. Support for _optional_ package dependencies be dropped 1 years after their initial release.

For reference, `xarray` seems to use 12 months as the default for most dependencies - https://github.com/pydata/xarray/blob/v2024.09.0/ci/min_deps_check.py#L35

seisman:
> Downside with this is more maintenance effort, since we need to update the minimum pins for optional dependencies more often. 

Yes, it's more maintenance burden to us. I actually prefer to not set any minimum pins for optional dependencies and rely on the package manager like `mamba`/`conda` to install whatever versions that are compatible with the core dependencies.

weiji14 (approved):
> > Downside with this is more maintenance effort, since we need to update the minimum pins for optional dependencies more often.
> 
> Yes, it's more maintenance burden to us. I actually prefer to not set any minimum pins for optional dependencies and rely on the package manager like `mamba`/`conda` to install whatever versions that are compatible with the core dependencies.

Opened separate issue at #3456 to discuss this. Ok to merge this in now.

PR #3431, opened 2024-09-09
CI: Pin actions/upload-artifact and peter-evans/slash-command-dispatch to the latest tag

seisman:

The [actions/upload-artifact v4.4.0](https://github.com/actions/upload-artifact/releases/tag/v4.4.0) introduced a breaking change, which excludes hidden files by default. This new version breaks the workflow in one of my repositories (here is the workflow run https://github.com/gmt-china/GMT_docs/actions/runs/10759759418 if you're interested). The new version should also affect our "Cache Data" workflow, since we upload the files in `~/.gmt`, but actually the workflow is not affected (likely because we run it under macOS).

Anyway, it's always a good idea to pin actions to a specific tag so that workflows won't break suddenly due to upstream changes.


.github/workflows/cache_data.yaml
```diff
@@ -76,9 +76,10 @@ jobs:
 
       # Upload the downloaded files as artifacts to GitHub
       - name: Upload artifacts to GitHub
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v4.4.0
         with:
           name: gmt-cache
+          include-hidden-files: true
```
seisman, line 82: In v4.4.0, `include-hidden-files` is `false` by default. Here we set it to `true`. 

The workflow runs as expected: https://github.com/GenericMappingTools/pygmt/actions/runs/10769413727?pr=3431

.github/workflows/slash-command-dispatch.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Slash Command Dispatch
-        uses: peter-evans/slash-command-dispatch@v4
+        uses: peter-evans/slash-command-dispatch@v4.4.0
```
seisman, line 17:
```suggestion
        uses: peter-evans/slash-command-dispatch@v4.0.0
```

.github/workflows/slash-command-dispatch.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Slash Command Dispatch
-        uses: peter-evans/slash-command-dispatch@v4
+        uses: peter-evans/slash-command-dispatch@v4.0.0
```
seisman, line 17: https://github.com/peter-evans/slash-command-dispatch/releases

weiji14 (approved):

PR #3432, opened 2024-09-09
Document the environment variables that can affect the behavior of PyGMT

seisman:
This PR adds another page to the "Technical Reference" section. On this page, the environmental variables that PyGMT uses are documented. There are still many other variables that are not documented since they are not explicitly used by PyGMT (e.g., `HOME`, `GMT_SHAREDIR`). We may document them later.

**Preview**:

- https://pygmt-dev--3432.org.readthedocs.build/en/3432/techref/environment_variables.html
- https://pygmt-dev--3432.org.readthedocs.build/en/3432/api/generated/pygmt.Figure.show.html
- https://pygmt-dev--3432.org.readthedocs.build/en/3432/api/generated/pygmt.Figure.timestamp.html
- https://pygmt-dev--3432.org.readthedocs.build/en/3432/api/generated/pygmt.x2sys_init.html
- https://pygmt-dev--3432.org.readthedocs.build/en/3432/api/generated/pygmt.x2sys_cross.html


doc/techref/environment_variables.md
```diff
@@ -0,0 +1,47 @@
+# Environmental Variables
+
+PyGMT's behavior can be controlled through various environment variables. These variables
+can be set either in your shell environment or within your Python script using the
+:py:data:`os.environ` dictionary.
```
yvonnefroehlich, line 5: Setting the link like this does not work correctly.
highlight_os_environ

I tried `:py:func:`, `:py:class:`, `:py:`, but with no success. Maybe we have to do it like:
```suggestion
[`os.environ`](https://docs.python.org/3/library/os.html#os.environ) dictionary.
```

seisman: Thanks for catching it. Fixed in 7ed130002bca93095e3aa89870b1848cc8c3fd95. `py:data` is correct. Just need to use the MyST Markdown syntax, not the ReST syntax.

seisman:
will merge in 24 hours if no further comments.

doc/techref/environment_variables.md
```diff
@@ -0,0 +1,47 @@
+# Environmental Variables
```
yvonnefroehlich, line 1: It seems both `environment` and `environmental` variables are used. I fell we should decide for one term.
Looking at this [StackOverflow post](https://stackoverflow.com/questions/70677153/are-they-called-environment-or-environmental-variables) `environment` variable is used more often and considered as "more correct".

```suggestion
# Environment Variables
```

doc/techref/environment_variables.md
```diff
@@ -0,0 +1,47 @@
...
+
+PyGMT's behavior can be controlled through various environment variables. These variables
+can be set either in your shell environment or within your Python script using the
+{py:data}`os.environ` dictionary.
+
+Here we list the environment variables used by PyGMT. The environment variables are
+categorized into three groups:
```
yvonnefroehlich, line 8: It's a bit shorter:
```suggestion
Here we list the environment variables used by PyGMT which are categorized into three groups:
```

PR #3437, opened 2024-09-11
Figure.legend: Refactor to simplify the logic of checking legend specification

seisman:

This PR contains a subset of changes in PR #3326. I open this PR so that PR #3326 can be smaller and easier for review.

This PR refactors the following codes: 
```python
    with Session() as lib:
        if spec is None:
            specfile = ""
        elif data_kind(spec) == "file" and not is_nonstr_iter(spec):
            # Is a file but not a list of files
            specfile = spec
        else:
            raise GMTInvalidInput(f"Unrecognized data type: {type(spec)}")
        lib.call_module(module="legend", args=build_arg_list(kwargs, infile=specfile))
```
to 
```python
    kind = data_kind(spec)
    if kind not in {"vectors", "file"}:  # kind="vectors" means spec is None
        raise GMTInvalidInput(f"Unrecognized data type: {type(spec)}")
    if kind == "file" and is_nonstr_iter(spec):
        raise GMTInvalidInput("Only one legend specification file is allowed.")

    with Session() as lib:
        lib.call_module(module="legend", args=build_arg_list(kwargs, infile=spec))
```
The pros are:

- Raise exceptions if any before entering a session
- Better error message if multiple legend files are given

### DIFF
--- a/pygmt/src/legend.py
+++ b/pygmt/src/legend.py
@@ -53,9 +53,10 @@ def legend(
     spec
         The legend specification. It can be:
 
-        - ``None`` which means using the automatically generated legend specification file.
+        - ``None`` which means using the automatically generated legend specification
+          file
         - A string or a :class:`pathlib.PurePath` object pointing to the legend
-          specification file.
+          specification file
 
         See :gmt-docs:`legend.html` for the definition of the legend specification.
     {projection}
--- a/pygmt/src/legend.py
+++ b/pygmt/src/legend.py
@@ -2,6 +2,8 @@
 legend - Plot a legend.
 """
 
+import pathlib
+
 from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
 from pygmt.helpers import (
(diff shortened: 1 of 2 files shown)


pygmt/src/legend.py
```diff
@@ -75,12 +87,11 @@ def legend(self, spec=None, position="JTR+jTR+o0.2c", box="+gwhite+p1p", **kwarg
         if kwargs.get("F") is None:
             kwargs["F"] = box
 
+    kind = data_kind(spec)
+    if kind not in {"vectors", "file"}:  # kind="vectors" means spec is None
```
seisman, line 91: This line can also be written as:
```
if spec is not None or kind != "file"
```
It's written in the current way so that this line can be extended to check "stringio" kind with minor changes (xref: https://github.com/GenericMappingTools/pygmt/pull/3326).

It's not straightforward to see why `kind="vectors"` means `spec is None`. That's also why I want to refactor the `data_kind` function in  #3351.

pygmt/src/legend.py
```diff
@@ -42,10 +50,14 @@ def legend(self, spec=None, position="JTR+jTR+o0.2c", box="+gwhite+p1p", **kwarg
...
     Parameters
     ----------
-    spec : None or str
-        Either ``None`` [Default] for using the automatically generated legend
-        specification file, or a *filename* pointing to the legend
-        specification file.
+    spec
+        The legend specification. It can be:
+
+        - ``None`` means using the automatically generated legend specification file.
```
michaelgrund, line 56:
```suggestion
        - ``None`` which means using the automatically generated legend specification file.
```

michaelgrund (approved):

PR #3448, opened 2024-09-23
SIM105: Replace try-except-pass blocks with the contextlib.suppress context manager

seisman:

Fix ruff [SIM105](https://docs.astral.sh/ruff/rules/suppressible-exception/) violations.

> ### What it does
> Checks for try-except-pass blocks that can be replaced with the contextlib.suppress context manager.
> 
> ### Why is this bad?
> Using contextlib.suppress is more concise and directly communicates the intent of the code: to suppress a given exception.
> 
> Note that contextlib.suppress is slower than using try-except-pass directly. For performance-critical code, consider retaining the try-except-pass pattern.

Not sure why ruff doesn't catch these violations, but the codes look cleaner with `contextlib.suppress`.

### DIFF
--- a/pygmt/clib/loading.py
+++ b/pygmt/clib/loading.py
@@ -5,6 +5,7 @@ The path to the shared library can be found automatically by ctypes or set throu
 environment variable :term:`GMT_LIBRARY_PATH`.
 """
 
+import contextlib
 import ctypes
 import os
 import shutil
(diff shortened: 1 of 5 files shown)


weiji14:
> Not sure why ruff doesn't catch these violations, but the codes look cleaner with `contextlib.suppress`.

Looks the the SIM105 rule won't work with >1 line under the `try:` block, see https://github.com/astral-sh/ruff/issues/8593. The idea I think is that if the 1st line fails, it's not so clear that the 2nd line wouldn't run if there is not an `except:` below it, see also rationale at https://github.com/astral-sh/ruff/issues/1947.

seisman:
> The idea I think is that if the 1st line fails, it's not so clear that the 2nd line wouldn't run if there is not an `except:` below it, see also rationale at [astral-sh/ruff#1947](https://github.com/astral-sh/ruff/issues/1947).

Makes sense. Closing the PR.

PR #3449, opened 2024-09-24
Add enums GridFormat for GMT grid format ID

seisman:

This PR is a subset of PR #3128. 

This PR adds the `pygmt.enums` module to defined enumerations that can be used in PyGMT. The source layout is inspired by the rasterio package (https://github.com/rasterio/rasterio/blob/main/rasterio/enums.py).

Currently, only enums for GMT grid format IDs are added. They're originally defined at https://github.com/GenericMappingTools/gmt/blob/7809736ba32d87a4a96b15444419eb176c6a35f3/src/gmt_grdio.h#L70. Enums for grid registrations and gtypes are likely to be added when addressing https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2272010810.

This PR also fixes the issue originally pointed by 5888e105a25063889d413ed65a6b694152458348.

**Preview**: https://pygmt-dev--3449.org.readthedocs.build/en/3449/api/generated/pygmt.enums.GridFormat.html

After ea1cc7978601f91e01e2d445fa5491a64c0edf58, the docs look like below:

image

After ac6239fa4f31e6b6049be8757e251960b89357a3, the docs look like below:
image


seisman:
Ready for review although the documentation doesn't look correct. Maybe we need to follow https://github.com/rasterio/rasterio/blob/9953b28225db3b01193c94b1442d34b828d374aa/docs/api/rasterio.enums.rst. Will explore it later.

pygmt/enums.py
```diff
@@ -0,0 +1,39 @@
...
+
+class GridID(IntEnum):
+    """
+    Enum for the GMT grid format ID.
+
+    These enums are defined in 'gmt_grdio.h'.
+    """
+
+    GMT_GRD_UNKNOWN_FMT = 0  # If grid format cannot be auto-detected
+    GMT_GRID_IS_BF = 1  # GMT native, C-binary format (32-bit float)
```
seisman, line 16: Since these enumerations are accessed like `GridID.GMT_GRID_IS_BF`, maybe we should define shorter enumerations like `GridID.BF` instead.

seisman: Done in ce3c55cd71c65919c94861790be3f3ecb1c9e98e.

pygmt/enums.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class GridID(IntEnum):
+    """
+    Enum for the GMT grid format ID.
+
+    These enums are defined in 'gmt_grdio.h'.
+    """
+
+    UNKNOWN = 0  #: Unkown grid format
```
weiji14, line 15:
```suggestion
    UNKNOWN = 0  #: Unknown grid format
```

pygmt/enums.py
```diff
@@ -0,0 +1,39 @@
+"""
+Enumerations for PyGMT.
+"""
+
+from enum import IntEnum
+
+
+class GridID(IntEnum):
```
weiji14, line 8: `GridID` doesn't sound very meaningful. How about `GridFormat`?

```suggestion
class GridFormat(IntEnum):
```

seisman:
Still, I wouldn't say I like how the enums are documented at https://pygmt-dev--3449.org.readthedocs.build/en/3449/api/generated/pygmt.enums.GridFormat.html

weiji14:
> Still, I wouldn't say I like how the enums are documented at https://pygmt-dev--3449.org.readthedocs.build/en/3449/api/generated/pygmt.enums.GridFormat.html

Do we need to have this `GridFormat` enum in the API docs actually? I don't think users will use it much, it's not like you can pass it into any function/method as an argument.

I'm also ok with keeping it as is right now, and improving the style/layout later. It doesn't look too bad to me.

pygmt/enums.py
```diff
@@ -0,0 +1,39 @@
...
+    BS = 2  #: GMT native, C-binary format (16-bit integer)
+    RB = 3  #: SUN rasterfile format (8-bit standard)
+    BB = 4  #: GMT native, C-binary format (8-bit integer)
+    BM = 5  #: GMT native, C-binary format (bit-mask)
+    SF = 6  #: Golden Software Surfer format 6 (32-bit float)
+    CB = 7  #: GMT netCDF format (8-bit integer, depreciated)
+    CS = 8  #: GMT netCDF format (16-bit integer, depreciated)
+    CI = 9  #: GMT netCDF format (32-bit integer, depreciated)
+    CF = 10  #: GMT netCDF format (32-bit float, depreciated)
+    CD = 11  #: GMT netCDF format (64-bit float, depreciated)
```
weiji14, line 26: Deprecated?
```suggestion
    CB = 7  #: GMT netCDF format (8-bit integer, deprecated)
    CS = 8  #: GMT netCDF format (16-bit integer, deprecated)
    CI = 9  #: GMT netCDF format (32-bit integer, deprecated)
    CF = 10  #: GMT netCDF format (32-bit float, deprecated)
    CD = 11  #: GMT netCDF format (64-bit float, deprecated)
```

weiji14 (approved):

seisman:
> Do we need to have this `GridFormat` enum in the API docs actually? I don't think users will use it much, it's not like you can pass it into any function/method as an argument.

Yes for `GridFormat`, but I think we need to document `GridRegistration` and `GridType` (suggested in https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2272522044). 

I'll revert the changes for docs now and explore how to make it better later.

I also prefer to splitting this PR into two smaller PRs, one for adding the GridFormat enums and one for fix the CF-Convention attribute, so that we can have two separate entries in the changelog.

pygmt/enums.py
seisman: The word "depreciated" comes from https://docs.generic-mapping-tools.org/dev/reference/features.html#grid-file-format-specifications. Reading https://stackoverflow.com/a/9208164, I think "depreciated" is still correct.

pygmt/enums.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class GridFormat(IntEnum):
+    """
+    Enum for the GMT grid format ID.
+
+    These enums are defined in 'gmt_grdio.h'.
+    """
+
+    UNKNOWN = 0  #: Unknown grid format
```
seisman, line 15: `#: ` is the special way to document attributes, following https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty.

pygmt/enums.py
weiji14: Depreciated is a monetary term (comes from the word depreciation). I still think it's a typo.

seisman: Makes sense. Applied the suggestions.

PR #3450, opened 2024-09-24
Set minimum supported version to GMT>=6.4.0

seisman:

In commit 539f3e80, I removed the codes for compatibility with GMT 6.3.0. After that, I manually triggered the "GMT Legacy Tests" workflow at https://github.com/GenericMappingTools/pygmt/actions/runs/11010826577. The CI job for GMT 6.3   no longer works because `API_BIN_VERSION` was added in GMT 6.4.0.
```
Error: ession [ERROR]: Unrecognized keyword API_BIN_VERSION
  File "/home/runner/work/pygmt/pygmt/pygmt/_show_versions.py", line [13](https://github.com/GenericMappingTools/pygmt/actions/runs/11010826577/job/30573600010#step:6:14)7, in show_versions
    lines.extend([f"  {key}: {val}" for key, val in _get_clib_info().items()])
  File "/home/runner/work/pygmt/pygmt/pygmt/_show_versions.py", line 29, in _get_clib_info
    return ses.info
  File "/home/runner/work/pygmt/pygmt/pygmt/clib/session.py", line 202, in info
    self._info["binary version"] = self.get_default("API_BIN_VERSION")
  File "/home/runner/work/pygmt/pygmt/pygmt/clib/session.py", line 500, in get_default
    raise GMTCLibError(msg)
pygmt.exceptions.GMTCLibError: Error getting value for 'API_BIN_VERSION' (error code 68).
make: *** [Makefile:35: _runtest] Error 1
```
The initial plan is to drop GMT 6.3 support in PyGMT v0.14.0, so we should remove any GMT 6.3 compatibility codes and GMT 6.3 from the CI matrix.

Then, the question is, do we want to revert changes in `pygmt/clib/session.py` so that most PyGMT functionality still works with GMT 6.3? In this case, GMT 6.3 is still allowed for future PyGMT versions, but there is no guarantee that all PyGMT functionalities work as expected. [Edit: Partially reverted in b0e6c69443f1da34f8c59223b7c4239a18d58a95].

In b0e6c69443f1da34f8c59223b7c4239a18d58a95, I updated `pygmt/clib/session.py` so that PyGMT works for GMT 6.3-6.5 (also have to temporarily revert `test_grdfill.py` in c6860b2a35c525bec022fbc13a44df478e4538f0 because it's known to crash with GMT 6.3). In the CI runs at https://github.com/GenericMappingTools/pygmt/actions/runs/11013102550, most tests still pass, except a few that are expected to fail.

In 33c1061b374e30f4cfa8bf756ea102a894762765, I removed GMT 6.3 for the testing CI matrix so GMT 6.3 is no longer tested, although PyGMT is still compatible with it. Then the question is whether we want to drop GMT 6.3 support completely. I think we have three options:

1. Bump `required_gmt_version` from `6.3.0` to `6.4.0`, which means that people who use GMT 6.3 will see the GMTVersionError exception
2. Keep `required_gmt_version="6.3.0"` and add a note like "PyGMT requires GMT>=6.3.0 but only GMT 6.4-6.5 are fully tested" on the installation page.
3. Have two variables like `required_gmt_version="6.4.0"` and `minimum_required_version="6.3.0"`, and raise the GMTVersionError for GMT<6.3.0 and raise an warning like "GMT 6.3.0 is not fully tested." for GMT=6.3.0.

Thoughts?


pygmt/clib/session.py
```diff
@@ -199,14 +198,13 @@ def info(self):
...
             }
-            # For GMT<6.4.0, API_IMAGE_LAYOUT is not defined if GMT is not
-            # compiled with GDAL. Since GMT 6.4.0, GDAL is a required GMT
-            # dependency. The code block can be refactored after we bump
-            # the minimum required GMT version to 6.4.0.
+            # API_IMAGE_LAYOUT is only defined when GMT is linked with the GDAL library.
+            # GDAL is a optional dependency in GMT 6.3.0 and is required since 6.4.0.
+            # API_BIN_VERSION is available since GMT 6.4.0.
+            # Here, we suppress the GMTCLibError exception to provide GMT 6.3 support
+            # with mininal efforts.
```
seisman, line 205:
```suggestion
            # with minimal effort.
```

seisman:
> 3\. Have two variables like `required_gmt_version="6.4.0"` and `minimum_required_version="6.3.0"`, and raise the GMTVersionError for GMT<6.3.0 and raise an warning like "GMT 6.3.0 is not fully tested." for GMT=6.3.0.

I'm inclined to option 3. What do you think @GenericMappingTools/pygmt-maintainers?

weiji14:
> > 3. Have two variables like `required_gmt_version="6.4.0"` and `minimum_required_version="6.3.0"`, and raise the GMTVersionError for GMT<6.3.0 and raise an warning like "GMT 6.3.0 is not fully tested." for GMT=6.3.0.
> 
> I'm inclined to option 3. What do you think @GenericMappingTools/pygmt-maintainers?

Ok with this in principle, but I'm thinking of the variable names. How about:
- `minimum_required_gmt_version="6.3.0"` or `minimum_supported_gmt_version="6.3.0"`?
- `recommended_gmt_version="6.4.0"`?

References:
- https://en.wikipedia.org/wiki/System_requirements#Recommended_system_requirements
- Rust has a concept called Minimum Supported Rust Version [(MSRV)](https://github.com/rust-lang/api-guidelines/discussions/231) for libraries/crates

Slightly off-topic, but I suppose we'll also set a requirement for `gmt >=6.4.0,<7.0` on PyGMT's conda-forge package here at https://github.com/conda-forge/pygmt-feedstock/blob/e34ae5af8fced8600a9c05f5298c7e224e251530/recipe/meta.yaml#L25 for the PyGMT v0.14.0 release? So users would really need to install GMT 6.3.0 manually (not through conda) to get it to work with PyGMT v0.14.0?

seisman:
> * `minimum_required_gmt_version="6.3.0"` or `minimum_supported_gmt_version="6.3.0"`?

I prefer to `minimum_required_gmt_version` since GMT 6.3.0 is no longer officially supported. 

> * `recommended_gmt_version="6.4.0"`?

It reads like we recommend 6.4.0 and don't recommend 6.5.0. What about `minimum_recommended_gmt_version`? 



> Slightly off-topic, but I suppose we'll also set a requirement for `gmt >=6.4.0,<7.0` on PyGMT's conda-forge package here at https://github.com/conda-forge/pygmt-feedstock/blob/e34ae5af8fced8600a9c05f5298c7e224e251530/recipe/meta.yaml#L25 for the PyGMT v0.14.0 release? So users would really need to install GMT 6.3.0 manually (not through conda) to get it to work with PyGMT v0.14.0?

Yes, I agree.

seisman:
> I think we have three options:
> 
> 1. Bump `required_gmt_version` from `6.3.0` to `6.4.0`, which means that people who use GMT 6.3 will see the GMTVersionError exception
> 2. Keep `required_gmt_version="6.3.0"` and add a note like "PyGMT requires GMT>=6.3.0 but only GMT 6.4-6.5 are fully tested" on the installation page.
> 3. Have two variables like `required_gmt_version="6.4.0"` and `minimum_required_version="6.3.0"`, and raise the GMTVersionError for GMT<6.3.0 and raise an warning like "GMT 6.3.0 is not fully tested." for GMT=6.3.0.

After thinking twice, maybe we should go for option 1 for now to reduce our maintenance burden and revisit option 3 when we release PyGMT v1.0.

pygmt/clib/session.py
```diff
@@ -198,16 +197,9 @@ def info(self):
                 "library path": self.get_default("API_LIBRARY"),
                 "cores": self.get_default("API_CORES"),
                 "grid layout": self.get_default("API_GRID_LAYOUT"),
+                "image layout": self.get_default("API_IMAGE_LAYOUT"),
```
weiji14, line 200: I'm getting a blank value with this as mentioned at https://github.com/GenericMappingTools/pygmt/pull/3468#discussion_r1781987591. Are you getting any output?

```python
import pygmt

with pygmt.clib.Session() as lib:
    print(lib.info["image layout"])  # API_IMAGE_LAYOUT
```

output from `lib.info` dict:

```
{'version': '6.5.0',
 'padding': '2',
 'share dir': '~/mambaforge/envs/pygmt/share/gmt',
 'plugin dir': '~/mambaforge/envs/pygmt/lib/gmt/plugins',
 'library path': '~/mambaforge/envs/pygmt/lib/libgmt.so',
 'cores': '16',
 'grid layout': 'rows',
 'image layout': '',
 'binary version': '6.5.0'}
```

weiji14 (commented):
In `doc/install.md`, change this line:

https://github.com/GenericMappingTools/pygmt/blob/f8db417d4f47d1f4d26247ecda95d38521fcdb71/doc/install.md?plain=1#L311

to "- Ghostscript 9.53-9.56 for GMT 6.4.0 (or below)"

pygmt/tests/test_datasets_load_remote_datasets.py
```diff
@@ -32,10 +32,8 @@ def test_load_remote_dataset_benchmark_with_region():
     assert data.attrs["horizontal_datum"] == "WGS84"
     assert data.gmt.registration == 0
     assert data.shape == (11, 21)
-    # The cpt attribute was added since GMT 6.4.0
     # Can't access the cpt attribute using virtual files
-    # if Version(__gmt_version__) >= Version("6.4.0"):
-    #    assert data.attrs["cpt"] == "@earth_age.cpt"
+    # assert data.attrs["cpt"] == "@earth_age.cpt"
```
weiji14, line 36: Is this something that could be obtained from the `GMT_GRID` struct somehow?

weiji14 (approved):
Thanks @seisman! There are some comments on this PR, but are more just questions that can be addressed separately.

pygmt/tests/test_datasets_load_remote_datasets.py
seisman: No, the cpt information is stored in the private, hidden structure `GMT_GRID_HEADER_HIDDEN`:

https://github.com/GenericMappingTools/gmt/blob/426abad288ef4b87c2cecf4c9633fd849866e45d/src/gmt_hidden.h#L163

We can't access it unless we decide to wrap the non-public GMT structure.

PR #3459, opened 2024-09-27
Add type hints for GMT anchor codes

seisman:

This PR adds typing hints for the GMT anchor codes like **TL**/**TR**.

The new module `pygmt._typing` is inspired by https://github.com/numpy/numpy/tree/main/numpy/_typing and https://github.com/pandas-dev/pandas/blob/main/pandas/_typing.py.

In GMT, the order of horizontal and vertical alignments doesn't matter, so both **TL** and **LT** are accepted, but here only codes like **TL** are added to avoid making the function signature to long.


pygmt/src/text.py
```diff
@@ -90,7 +93,7 @@ def text_(  # noqa: PLR0912
     x/y : float or 1-D arrays
         The x and y coordinates, or an array of x and y coordinates to plot
         the text.
-    position : str
+    position
```
weiji14, line 96: Small nit, could you reverse the order of L102 and L103 below, so that Vertical (TMD) is on top of Horizontal (LCR)?

weiji14 (approved):

PR #3460, opened 2024-09-27
SPEC 0: Set minimum supported versions to pandas>=2.0 and xarray>=2023.04

seisman:

Following [SPEC0](https://scientific-python.org/specs/spec-0000/), we should drop pandas 1.5 and xarray 2022.09 support in 2024 Q3, but now we have dependency conflicts.

```
  error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ pandas =2.0 is requested and can be installed;
      └─ xarray =2022.11 is not installable because it requires
         └─ pandas >=1.3,<2a0, which conflicts with any installable versions previously reported.
  critical libmamba Could not solve for environment specs
```

Following https://github.com/pydata/xarray/issues/7716#issuecomment-1510434058, xarray 2023.04 works with pandas 2.0. Then we must decide whether to merge this PR now (violating the SPEC0 policy) or wait until April 2025.

I prefer to wait until April 2025, so adding this PR to milestone v0.16.0.

Previous pandas version bump at #3043.

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -71,7 +71,7 @@ jobs:
         include:
           - python-version: '3.10'
             numpy-version: '1.24'
-            pandas-version: '=1.5'
+            pandas-version: '=2.0'
             xarray-version: '=2022.09'
             optional-packages: ''
           - python-version: '3.12'
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -72,7 +72,7 @@ jobs:
           - python-version: '3.10'
             numpy-version: '1.24'
             pandas-version: '=2.0'
-            xarray-version: '=2022.09'
+            xarray-version: '=2022.11'
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '2.1'
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -72,7 +72,7 @@ jobs:
           - python-version: '3.10'
             numpy-version: '1.24'
             pandas-version: '=2.0'
-            xarray-version: '=2023.03'
+            xarray-version: '=2023.04'
             optional-packages: ''
           - python-version: '3.12'
             numpy-version: '2.1'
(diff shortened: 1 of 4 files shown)


weiji14:
> Following [pydata/xarray#7716 (comment)](https://github.com/pydata/xarray/issues/7716#issuecomment-1510434058), xarray 2023.04 works with pandas 2.0. Then we must decide whether to merge this PR now (violating the SPEC0 policy) or wait until April 2025

Yeah, I think we'll need to bump the minimum required `xarray` version more than needed this time (to `2023.03`) to work with `pandas=2.0`, considering that conda-forge did a repodata-patch for this at https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/426#issuecomment-1496389758. [SPEC 0](https://scientific-python.org/specs/spec-0000) does say the following:

> ... projects may occasionally want to drop support for core package dependencies earlier than recommended by this SPEC. For instance, if a newer minimum version of a core package is needed by a project due to a critical bug fix, which is not backported to older versions. 

I would consider that xarray 2023.03 supporting pandas 2.0 is a critical bug fix :slightly_smiling_face:, so would prefer to get this into PyGMT v0.15.0.

seisman:
I'm OK with making it to v0.14.0

weiji14 (commented):
Need to update `doc/minversions.rst` too.

Can we also agree on a consistent title for SPEC 0 updates? We've been inconsistent in https://www.pygmt.org/v0.13.0/changes.html:

- SPEC 0: Bump minimum supported version to xarray 2022.09
- SPEC 0: Set minimum supported versions to Python>=3.10, pandas>=1.5 and xarray>=2022.06
- NEP29: Set minimum required version to NumPy 1.23+

I like the middle one using `pandas>=1.5`, but no strong opinions.

seisman:
> Need to update `doc/minversions.rst` too.

No need to update this file after https://github.com/GenericMappingTools/pygmt/pull/3380.



> Can we also agree on a consistent title for SPEC 0 updates? We've been inconsistent in https://www.pygmt.org/v0.13.0/changes.html:
> 
> * SPEC 0: Bump minimum supported version to xarray 2022.09
> * SPEC 0: Set minimum supported versions to Python>=3.10, pandas>=1.5 and xarray>=2022.06
> * NEP29: Set minimum required version to NumPy 1.23+
> 
> I like the middle one using `pandas>=1.5`, but no strong opinions.

Sounds good

weiji14 (approved):
> > Need to update `doc/minversions.rst` too.
> 
> No need to update this file after #3380.

Oh yes, old habits die hard. LGTM!

PR #3463, opened 2024-09-28
Set the 'Conventions' attribute to 'CF-1.7' for netCDF grids only

seisman:
`Conventions="CF-1.7"` only makes sense for netCDF grids. 

The issue was originally found and fixed in 5888e105a25063889d413ed65a6b694152458348 (PR #3128). It is further refactored after adding the `GridFormat` enums in PR #3449.

Waiting for #3449 first.


weiji14 (approved):
Approved to merge after #3449

PR #1457, opened 2021-08-19
Function to load earth night grid

srijac:
Loading the Black Marble dataset

Addresses #1442

PR #3469, opened 2024-09-30
 Add function to load Black Marble dataset

### LINKED ISSUE #1442: Function to load NASA Blue and Black marble mosaic images

Provide access to cloud-free colour images of the Earth! Specifically, the `@earth_day_` and `@earth_night_` GMT remote data files (see https://docs.generic-mapping-tools.org/6.2/datasets/remote-data.html#global-earth-day-night-images).

image

Context is that I'm using it at #1437, and it would be good to have some sample RGB images. We might as well have a proper function similar to [`pygmt.datasets.load_earth_relief`](https://www.pygmt.org/v0.4.1/api/generated/pygmt.datasets.load_earth_relief.html), but for the Blue and Black Marble images instead.

Note that there is a lot of logic from the [`load_earth_relief`](https://www.pygmt.org/v0.4.1/api/generated/pygmt.datasets.load_earth_relief.html) function that could be reused:

https://github.com/GenericMappingTools/pygmt/blob/ebbb0cc0045454f05db5abdfb3bcb5e1f18e96a7/pygmt/datasets/earth_relief.py#L92-L140

These are some ways I'm considering:

Option 1 - rename the `earth_relief.py` file to `earth_data.py` first, and then separate the re-usable chunk of code into a function called `_load_earth_data` or something and have `load_earth_relief`/`load_earth_day`/`load_earth_night` call that. This is similar to what we did for `blockmean` and `blockmedian` at #1092. Less breakages for users already using `load_earth_relief`, but maybe more work for the devs.

Option 2 - have a single function called `load_earth_data` which accepts a parameter called 'type' that accepts arguments like 'relief'/'day'/'night'. We could then pretty much re-use the existing `load_earth_relief` function though it will need to be renamed to `load_earth_data` I guess. This would involve a deprecation notice for users using `load_earth_relief`, but less work for the devs in terms of not needing to write a new function?

Option 3 - any other ideas?

**Are you willing to help implement and maintain this feature?** Best to discuss first.

weiji14:

Function to load the `@earth_night_` NASA Black Marble mosaic RGB images stored as GeoTIFF files.

**Preview** at https://pygmt-dev--3469.org.readthedocs.build/en/3469/api/generated/pygmt.datasets.load_black_marble.html

Example usage:

```python
import pygmt

image = pygmt.datasets.load_black_marble(resolution="06m")

fig = pygmt.Figure()
fig.grdimage(grid=image)
fig.show()
```
produces

earth_night

Implementation is adapted from #2235.

Fixes #1442, supersedes #1457

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/earth_night.py
+++ b/pygmt/datasets/earth_night.py
@@ -1,0 +1,97 @@
+"""
+Function to download the NASA Black Marble image datasets from the GMT data server, and
+load as :class:`xarray.DataArray`.
+
+The images are available in various resolutions.
+"""
+
+from collections.abc import Sequence
+from typing import Literal
+
+import xarray as xr
+from pygmt.datasets.load_remote_dataset import _load_remote_dataset
+
+__doctest_skip__ = ["load_black_marble"]
+
+
+def load_black_marble(
+    resolution: Literal[
+        "01d",
+        "30m",
+        "20m",
+        "15m",
+        "10m",
+        "06m",
+        "05m",
+        "04m",
+        "03m",
+        "02m",
+        "01m",
+        "30s",
+    ] = "01d",
+    region: Sequence[float] | str | None = None,
+) -> xr.DataArray:
+    r"""
+    Load NASA Black Marble images in various resolutions.
+
+    .. figure:: https://www.generic-mapping-tools.org/remote-datasets/_images/GMT_earth_daynight.jpg
+       :width: 80%
+       :align: center
+
+       Earth day/night dataset.
+
+    The images are downloaded to a user data directory (usually
+    ``~/.gmt/server/earth/earth_night/``) the first time you invoke this function.
+    Afterwards, it will load the image from the data directory. So you'll need an
+    internet connection the first time around.
+
+    These images can also be accessed by passing in the file name
+    **@earth_night**\_\ *res* to any image processing function or plotting method. *res*
+    is the image resolution (see below).
+
+    Refer to :gmt-datasets:`earth-daynight.html` for more details about available
+    datasets, including version information and references.
+
+    Parameters
+    ----------
+    resolution
+        The image resolution. The suffix ``d``, ``m``, and ``s`` stand for arc-degree,
+        arc-minute, and arc-second.
+
+    region
+        The subregion of the image to load, in the form of a sequence [*xmin*, *xmax*,
+        *ymin*, *ymax*].
+
+    Returns
+    -------
+    image
+        The NASA Black Marble image. Coordinates are latitude and longitude in degrees.
+
+    Note
+    ----
+    The registration and coordinate system type of the returned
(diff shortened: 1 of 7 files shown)


seisman (approved):

pygmt/datasets/earth_night.py
```diff
@@ -0,0 +1,97 @@
...
+    image = _load_remote_dataset(
+        name="earth_night",
+        prefix="earth_night",
+        resolution=resolution,
+        region=region,
+        registration="pixel",
+    )
+    # If rioxarray is installed, set the coordinate reference system
+    if hasattr(image, "rio"):
+        image = image.rio.write_crs(input_crs="OGC:CRS84")
```
weiji14, line 96: This line is not covered by a unit test. Should I add one, or we can ignore it?

seisman: It should be covered when rioxarray is installed, right?

seisman: Or do we have to import rioxarray to register the rio accessor

weiji14: The equivalent line in `load_blue_marble` is covered here in `test_grdimage_image.py` which has an `import rioxarray` at the top of the file:

https://github.com/GenericMappingTools/pygmt/blob/66f258aa8bb98d55b6768067ad00b60cd06c89b4/pygmt/tests/test_grdimage_image.py#L18

We don't test `load_black_marble` anywhere else except in `test_datasets_earth_night.py`, and since `rioxarray` is not imported in that file, this line isn't covered because there is no `rio` accessor.

weiji14: Maybe we can cover this line in https://github.com/GenericMappingTools/pygmt/pull/3115/files#r1780370539?

seisman: > since `rioxarray` is not imported in that file, this line isn't covered because there is no `rio` accessor.

This is not ideal. So even if rioxarray is installed, the `rio` accessor may still not exist because users don't import the rioxarray package.

I think we should add the following lines to this file (and any other files that try to access the `rio` accessor):

https://github.com/GenericMappingTools/pygmt/blob/66f258aa8bb98d55b6768067ad00b60cd06c89b4/pygmt/datasets/tile_map.py#L21-L23

seisman: xref: https://github.com/GenericMappingTools/pygmt/pull/3323

weiji14: Ok, done at commit 66183ed40fae857ca0f29b19dd69d1cf87bd5811. I've also applied this to `pygmt/datasets/earth_day.py`, hopefully you don't mind.

seisman (approved):

seisman:
Alos need to add `earth_night_01d` to `caching.py`

PR #3471, opened 2024-09-30
Fix earth_day_01d_p to earth_day_01d

seisman:
Unlike other remote datasets, the `earth_day` dataset only provides pixel-registered images. So, its full syntax should be `@earth_day_01d` rather than `@earth_day_01d_p` (ref: https://www.generic-mapping-tools.org/remote-datasets/earth-daynight.html), although the original GeoTIFF files are named like `earth_day_01d_p.tif` (https://oceania.generic-mapping-tools.org/server/earth/earth_day/).

This PR fixes `earth_day_01d_p` to `earth_day_01d`


weiji14 (approved):
Nothing user-facing I think, so added skip-changelog.

PR #2443, opened 2023-03-20
Add a new target 'doctest' to run doctests only and simplify Makefile

seisman:
**Description of proposed changes**

This is the first step to address https://github.com/GenericMappingTools/pygmt/issues/2286#issuecomment-1475300763.

Changes in this PR:

- Use Makefile's [Target-specific Variable Values](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html) to simplify the Makefile targets related to tests. Now we have the `_runtest` target to run the tests. Other targets `tests`, `fulltests`, `doctest`, `test_no_images` all depend on the `_runtest` target. For each target, different values for `PYTEST_ARGS` are passed, so that different tests are execuated.
- Add a `doctest` target which only runs the doctests, as suggested in https://github.com/GenericMappingTools/pygmt/issues/2286#issuecomment-1475300763
- Only generate coverage reports for the `test` target

To test this PR, you can run `make <target>` to see which tests run. 

To save your time, you can simply add `--collect-only` (only collect the tests and don't run them) to pytest, i.e., changing line 37 to
```
cd $(TESTDIR); PYGMT_USE_EXTERNAL_DISPLAY="false" pytest --collect-only $(PYTEST_ARGS) --pyargs $(PROJECT)
``` 

Here are the counts of tests when I run these targets:

- `make test`: 659
- `make fulltest`: 696
- `make doctest`: 68
- `make test_no_images`: 696

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

Issue #2286, opened 2022-12-29
Run full tests (including inline doctests) in workflows


seisman:
As mentioned in https://github.com/GenericMappingTools/pygmt/pull/2281#issuecomment-1366768553:

- [x] IPython is not installed in the "GMT Legacy Tests" #2283
- [x] It's likely that the "Tests" workflow doesn't run the full tests (including all doctests) on Wednesday #2288

seisman:
I'm reopening the issue, because it seems the "Tests" workflow still doesn't run the full tests on Wednesday (e.g., https://github.com/GenericMappingTools/pygmt/actions/runs/4421700978/jobs/7752774811).

seisman:
Our current strategy is to run the unit tests in daily scheduled CI jobs and only run the "full tests" (unit tests + doctests) on the Wednesday scheduled CI job. 

Actually, looking at the list of daily scheduled runs (https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_tests.yaml?query=event%3Aschedule), I find it very difficult to know which job runs on Wednesday. What's worse is, if all unit tests pass but some doctests fail, we will see six passing runs and one failing run every week. It's very likely that we will miss the failing run and don't know that we're having failing doctests.

I think the simplest solution is "always run the full tests in scheduled jobs". 

As mentioned in https://github.com/GenericMappingTools/pygmt/pull/1833#issuecomment-1073488403,

> We have PRs merging into the main branch every few days, so the unit tests are also run every few days, making the daily unit tests less necessary. So a weekly job is still acceptable to me.

If you think that running the full tests daily is a waste of CI resources, maybe we can run every two or three days or once per week.

weiji14:
Alternatively, could we revisit a solution to run only the doctest as mentioned in https://github.com/GenericMappingTools/pygmt/pull/1833#issuecomment-1073011944? I'm thinking of a workflow that does:

```
rm pygmt/tests/*
pytest --doctest-modules ...
```

That way, we have a dedicated workflow to catch errors only in the doctests.

seisman:
> Alternatively, could we revisit a solution to run only the doctests as mentioned in [#1833 (comment)](https://github.com/GenericMappingTools/pygmt/pull/1833#issuecomment-1073011944)? I'm thinking of a workflow that does:
> 
> ```
> rm pygmt/tests/*
> pytest --doctest-modules ...
> ```
> 
[...]

PR #3480, opened 2024-10-02
data_kind: Add more tests to demonstrate the data kind of various data types

seisman:

Just add more tests to help us understand the current behavior of the `data_kind` function.

As can be seen, many data types are recognized as "matrix" and `data=None` is recognized as "vectors". Sometimes, the result data kind is counterintuitive, as first mentioned in #3351, #3369 and #2744.

This PR only focuses on adding tests. Will refactor the data_kind function in separate PRs.

Please let me if there are other data types that should be tested.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -222,30 +222,72 @@ def data_kind(
 
     Examples
     --------
-    >>> import numpy as np
-    >>> import xarray as xr
-    >>> import pathlib
     >>> import io
-    >>> data_kind(data=None)
-    'vectors'
-    >>> data_kind(data=np.arange(10).reshape((5, 2)))
-    'matrix'
-    >>> data_kind(data="my-data-file.txt")
-    'file'
-    >>> data_kind(data=pathlib.Path("my-data-file.txt"))
-    'file'
+    >>> from pathlib import Path
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> import xarray as xr
+
+    The "arg" kind:
+
+    >>> [data_kind(data=data, required=False) for data in (2, 2.0, True, False)]
+    ['arg', 'arg', 'arg', 'arg']
     >>> data_kind(data=None, required=False)
     'arg'
-    >>> data_kind(data=2.0, required=False)
-    'arg'
-    >>> data_kind(data=True, required=False)
-    'arg'
-    >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
+
+    The "file" kind:
+
+    >>> [data_kind(data=data) for data in ("file.txt", ("file1.txt", "file2.txt"))]
+    ['file', 'file']
+    >>> data_kind(data=Path("file.txt"))
+    'file'
+    >>> data_kind(data=(Path("file1.txt", "file2.txt")))
+    'file'
+
+    The "grid" kind:
+
+    >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))  # 2-D xarray.DataArray
     'grid'
-    >>> data_kind(data=xr.DataArray(np.random.rand(3, 4, 5)))
+    >>> data_kind(data=xr.DataArray(np.arange(12)))  # 1-D xarray.DataArray
+    'grid'
+    >>> data_kind(data=xr.DataArray(np.random.rand(2, 3, 4, 5)))  # 4-D xarray.DataArray
+    'grid'
+
+    The "image" kind:
+
+    >>> data_kind(data=xr.DataArray(np.random.rand(3, 4, 5)))  # 3-D xarray.DataArray
     'image'
+
+    The "stringio"`` kind:
+
     >>> data_kind(data=io.StringIO("TEXT1\nTEXT23\n"))
     'stringio'
+
+    The "matrix"`` kind:
+
+    >>> data_kind(data=np.arange(10))  # 1-D numpy.ndarray
+    'matrix'
+    >>> data_kind(data=np.arange(10).reshape((5, 2)))  # 2-D numpy.ndarray
+    'matrix'
+    >>> data_kind(data=np.arange(60).reshape((3, 4, 5)))  # 3-D numpy.ndarray
+    'matrix'
+    >>> data_kind(xr.DataArray(np.arange(12), name="x").to_dataset())  # xarray.Dataset
+    'matrix'
+    >>> data_kind(data=[1, 2, 3])  # 1-D sequence
+    'matrix'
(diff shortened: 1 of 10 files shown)


pygmt/tests/test_blockm.py
```diff
@@ -12,7 +12,7 @@
 from pygmt import blockmean, blockmode
 from pygmt.datasets import load_sample_data
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import GMTTempFile, data_kind
```
seisman, line 15: These `assert`s are already covered by the `data_kind` doctests and can be removed.

pygmt/helpers/utils.py
```diff
@@ -222,30 +222,72 @@ def data_kind(
...
+    'matrix'
+    >>> data_kind(data=np.arange(60).reshape((3, 4, 5)))  # 3-D numpy.ndarray
+    'matrix'
+    >>> data_kind(xr.DataArray(np.arange(12), name="x").to_dataset())  # xarray.Dataset
+    'matrix'
+    >>> data_kind(data=[1, 2, 3])  # 1-D sequence
+    'matrix'
+    >>> data_kind(data=[[1, 2, 3], [4, 5, 6]])  # sequence of sequences
+    'matrix'
+    >>> data_kind(data={"x": [1, 2, 3], "y": [4, 5, 6]})
```
seisman, line 280:
```suggestion
    >>> data_kind(data={"x": [1, 2, 3], "y": [4, 5, 6]})  # dictionary
```

pygmt/tests/test_blockm.py
```diff
@@ -68,7 +68,6 @@ def test_blockmean_wrong_kind_of_input_table_grid(dataframe):
     Run blockmean using table input that is not a pandas.DataFrame or file but a grid.
     """
     invalid_table = dataframe.bathymetry.to_xarray()
-    assert data_kind(invalid_table) == "grid"
```
seisman, line 71: These `assert`s are already covered by the `data_kind` doctests and can be removed.

pygmt/helpers/utils.py
```diff
@@ -222,30 +220,72 @@ def data_kind(
...
-    'arg'
-    >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))
+
+    The "file" kind:
+
+    >>> [data_kind(data=data) for data in ("file.txt", ("file1.txt", "file2.txt"))]
+    ['file', 'file']
+    >>> data_kind(data=Path("file.txt"))
+    'file'
+    >>> data_kind(data=(Path("file1.txt", "file2.txt")))
```
seisman, line 242:
```suggestion
    >>> data_kind(data=(Path("file1.txt"), Path("file2.txt")))
```

pygmt/helpers/utils.py
```diff
@@ -222,30 +220,72 @@ def data_kind(
...
+    'file'
+
+    The "grid" kind:
+
+    >>> data_kind(data=xr.DataArray(np.random.rand(4, 3)))  # 2-D xarray.DataArray
+    'grid'
+    >>> data_kind(data=xr.DataArray(np.arange(12)))  # 1-D xarray.DataArray
+    'grid'
+    >>> data_kind(data=xr.DataArray(np.random.rand(2, 3, 4, 5)))  # 4-D xarray.DataArray
     'grid'
```
weiji14, line 252: Erm, do we need tests to check that 1-D and 4-D xarray.DataArray are counted as grid?

pygmt/tests/test_grdtrack.py
```diff
@@ -126,22 +126,18 @@ def test_grdtrack_profile(dataarray):
...
-
-    assert data_kind(invalid_points) == "grid"
     with pytest.raises(GMTInvalidInput):
         grdtrack(points=invalid_points, grid=dataarray, newcolname="bathymetry")
 
 
 def test_grdtrack_wrong_kind_of_grid_input(dataarray, dataframe):
     """
-    Run grdtrack using grid input that is not as xarray.DataArray (grid) or file.
+    Run grdtrack using grid input that is not as xarray.DataArray or file.
```
weiji14, line 138:
```suggestion
    Run grdtrack using grid input that is not an xarray.DataArray or file.
```

pygmt/tests/test_blockm.py
weiji14: I'd prefer to keep the asserts intact because `ci_doctests.yml` only runs weekly, not on every Pull Request, so it will be harder to check if refactors to `data_kind` broke something.

seisman: These doctests are always executed in the "Tests" workflow. See https://github.com/GenericMappingTools/pygmt/actions/runs/11145126496/job/30975174325?pr=3480

```
[gw0] [ 14%] PASSED ../pygmt/helpers/utils.py::pygmt.helpers.utils.data_kind 
```

Only those added to `__doctest_skip__` are skipped in the "Tests" workflow. See #2443 and https://github.com/GenericMappingTools/pygmt/issues/2286#issuecomment-1475300763 for how it works.

pygmt/helpers/utils.py
seisman: Not neccessary. These doctests are added to understand the current behavior of the `data_kind` function. 

I feel 1-D xarray.DataArray should be considered as a vector and 4-D xarray.DataArray should not be recognized and we should raise an exception. Will refactor the `data_kind` function in a separate PR.

pygmt/tests/test_blockm.py
weiji14: Ah yes, thanks for refreshing my memory. Then ok to remove the assertions.

pygmt/helpers/utils.py
weiji14: > I feel 1-D xarray.DataArray should be considered as a vector

We already recognize an `xr.Dataset` composed of 1-D `xarray.DataArray`s as a vector (can be treated as multiple 'columns' in a Dataframe). I don't think we need to expand this logic, since they users can call `.data` on a 1-D `xarray.DataArray` to get a numpy.ndarray that can then be passed to the `data` parameter.

> 4-D xarray.DataArray should not be recognized and we should raise an exception

Logic could be something like `if dataarray.ndim() >= 4: raise Error`, but I don't think this is really necessary since an error would be raised by the PyGMT module anyway right?

weiji14 (approved):
Pre-approving since most of this looks good. There's some discussion about whether to check 1-D/4-D `xarray.DataArray`s at https://github.com/GenericMappingTools/pygmt/pull/3480#discussion_r1785653218, but not very critical and can be handled later with the actual code refactor.

PR #3481, opened 2024-10-03
data_kind: Refactor the if-else statements into match-case statements

seisman:

Follow-up PR after PR #3480.

This PR refactors the `data_kind` function, to use `match-case` statements instead of `if-else` statements. 

This PR also updates the docstrings.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -187,7 +187,7 @@ def _check_encoding(
     return "ISOLatin1+"
 
 
-def data_kind(
+def data_kind(  # noqa: PLR0911
     data: Any = None, required: bool = True
 ) -> Literal[
     "arg", "file", "geojson", "grid", "image", "matrix", "stringio", "vectors"
@@ -195,15 +195,18 @@ def data_kind(
     r"""
     Check the kind of data that is provided to a module.
 
-    The ``data`` argument can be in any type, but only following types are supported:
+    The ``data`` argument can be in any type. Following data kinds are recognized:
 
-    - a string or a :class:`pathlib.PurePath` object or a sequence of them, representing
-      a file name or a list of file names
-    - a 2-D or 3-D :class:`xarray.DataArray` object
-    - a 2-D matrix
-    - None, bool, int or float type representing an optional arguments
-    - a geo-like Python object that implements ``__geo_interface__`` (e.g.,
-      geopandas.GeoDataFrame or shapely.geometry)
+    - ``"arg"``: data is ``None`` and ``required=False``, or bool, int, float,
+      representing an optional argument, used for dealing with optional virtual files
+    - ``"file"``: a string or a :class:`pathlib.PurePath` object or a sequence of them,
+      representing one or more file names
+    - ``"geojson"``: a geo-like Python object that implements ``__geo_interface__``
+      (e.g., geopandas.GeoDataFrame or shapely.geometry)
+    - ``"grid"``: a :class:`xarray.DataArray` object that is not 3-D
+    - ``"image"``: a 3-D :class:`xarray.DataArray` object
+    - ``"matrix"``: anything that is not None
+    - ``"vectors"``: data is ``None`` and ``required=True``
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


pygmt/helpers/utils.py
```diff
@@ -187,23 +187,26 @@ def _check_encoding(
     return "ISOLatin1+"
 
 
-def data_kind(
+def data_kind(  # noqa: PLR0911
```
seisman, line 190: PLR0911 means "too-many-return-statements".

https://docs.astral.sh/ruff/rules/too-many-return-statements/

pygmt/helpers/utils.py
```diff
@@ -287,30 +290,36 @@ def data_kind(
     >>> data_kind(data=None)
     'vectors'
     """
-    kind: Literal[
```
seisman, line 290: This type hint is no longer needed.

seisman:
Will merge in 24 hours.

pygmt/helpers/utils.py
weiji14: > This PR refactors the data_kind function, to use if-return statements instead of if-else statements. The new if-return statements are easier to read and maintain.

I don't see much of an improvement in readability/maintainability between the previous `if-elif-else` statements and the `if-return` style in this PR. Looking at https://stackoverflow.com/questions/72024229/how-to-solve-pylint-too-many-return-statements-elegantly, I'm thinking if we could either keep the `if-elif-else` syntax, or use `match-case` instead.

If using `match-case`, would it be more readable to do the `isinstance()` checks first, and then have a nested `if-elif` for `case _` for the more complex cases?

pygmt/helpers/utils.py
```diff
@@ -187,23 +187,27 @@ def _check_encoding(
...
-    - a 2-D matrix
-    - None, bool, int or float type representing an optional arguments
-    - a geo-like Python object that implements ``__geo_interface__`` (e.g.,
-      geopandas.GeoDataFrame or shapely.geometry)
+    - ``"arg"``: data is ``None`` and ``required=False``, or bool, int, float,
+      representing an optional argument, used for dealing with optional virtual files
+    - ``"file"``: a string or a :class:`pathlib.PurePath` object or a sequence of them,
+      representing one or more file names
+    - ``"geojson"``: a geo-like Python object that implements ``__geo_interface__``
+      (e.g., geopandas.GeoDataFrame or shapely.geometry)
```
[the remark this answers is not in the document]
seisman: This function is not documented in the API documentation page, so linking `shapely.geometry` or not makes little difference.

pygmt/helpers/utils.py
seisman: I feel `match-case` would make it more complicated:
```
match data:
    case x if isinstance(x, io.StringIO):
        return "stringio"
    case x if isinstance(x, xr.DataArray):
        return "image" if len(x.dims) == 3 else "grid"
    case x if isinstance(x, bool | int | float) or (x is None and not required):
        return "arg"
```

seisman: Using `match-case` also has the `too-many-return-statements` warning, and using `if-elif-else` means we need to type hints the `kind` variable like:
```
kind: Literal["arg", "file", "geojson", "grid", "image", "matrix", "stringio", "vectors"]
```

weiji14: Haven't run tests on this, but we can remove some of the `isinstance()` calls like so:

```python
match data:
        case str() | pathlib.PurePath():
            kind = "file"
        case list() | tuple() if all(
            isinstance(_file, str | pathlib.PurePath) for _file in data
        ):
            kind = "file"
        case io.StringIO():
            kind = "stringio"
        case (bool() | int() | float()) | None if not required:
            "arg"
        case xr.DataArray():
            kind = "image" if len(data.dims) == 3 else "grid"
        case x if hasattr(x, "__geo_interface__"):
            kind = "geojson"
        case x if x is not None:
            kind = "matrix"
        case _:
            kind = "vectors"
    return kind
```

See [class patterns](https://peps.python.org/pep-0634/#class-patterns) (xref https://stackoverflow.com/questions/67524641/convert-multiple-isinstance-checks-to-structural-pattern-matching/67524642#67524642)

seisman: The `match-case` statements look more compact. Done in 33051ac88eb7e32119500ac8880f329d01be6d83.

The mypy warning is also suppressed.

pygmt/helpers/utils.py
```diff
@@ -287,30 +292,31 @@ def data_kind(
...
+        case str() | pathlib.PurePath():  # One file.
+            kind = "file"
+        case list() | tuple() if all(
+            isinstance(_file, str | pathlib.PurePath) for _file in data
+        ):  # A list/tuple of files.
+            kind = "file"
+        case io.StringIO():
+            kind = "stringio"
+        case (bool() | int() | float()) | None if not required:
+            # An option argument, mainly for dealing optional virtual files.
```
weiji14, line 305:
```suggestion
            # An option argument, mainly for dealing with optional virtual files.
```

pygmt/helpers/utils.py
```diff
@@ -188,22 +188,27 @@ def _check_encoding(
...
+    - ``"arg"``: ``data`` is ``None`` and ``required=False``, or bool, int, float,
+      representing an optional argument, used for dealing with optional virtual files
+    - ``"file"``: a string or a :class:`pathlib.PurePath` object or a sequence of them,
+      representing one or more file names
+    - ``"geojson"``: a geo-like Python object that implements ``__geo_interface__``
+      (e.g., geopandas.GeoDataFrame or shapely.geometry)
+    - ``"grid"``: a :class:`xarray.DataArray` object that is not 3-D
+    - ``"image"``: a 3-D :class:`xarray.DataArray` object
+    - ``"stringio"``: a :class:`io.StringIO` object
+    - ``"matrix"``: anything that is not ``None``
```
weiji14, line 210:
```suggestion
    - ``"matrix"``: anything else that is not ``None``
```

weiji14 (approved):

PR #3482, opened 2024-10-03
**Breaking**: data_kind: data is None and required now returns the 'empty' kind

seisman:
Need to wait for #3351 or merge directly into #3351.

This PR adds a new kind ``"empty"`` for the `data=None, required=True` case. The new kind name makes more sense and can simplify the logic of codes.


pygmt/clib/session.py
```diff
@@ -1791,6 +1791,7 @@ def virtualfile_in(  # noqa: PLR0912
             "image": tempfile_from_image,
             "stringio": self.virtualfile_from_stringio,
             "matrix": self.virtualfile_from_matrix,
+            "none": self.virtualfile_from_vectors,
```
weiji14, line 1794: Can we change to another word like "empty" to reduce confusion with `None`?

seisman: I have another idea to completely eliminate the "none" kind. I'll open a POC PR later.

seisman: > I have another idea to completely eliminate the "none" kind. I'll open a POC PR later.

It turns out the idea can't eliminate the "none" kind. So let's focus on this PR again.

> Can we change to another word like "empty" to reduce confusion with `None`?

I want to emphasize that `data_kind(data=None)` returns `"none"`, so `kind == "none"` means `data is None`, which IMHO is very intuitive. That's why I choose `"none"` as the new kind name, but I'm also OK with a new `kind` name if others think `"empty"` is a better name.

weiji14: The intention of `kind == "none"` is to indicate that the data is stored in the `x`, `y`, `z` variables rather than in `data` (i.e. `data is None` means data is in xyz variables). I thought about naming it as `kind="xyz"` but since we are planning to support other 1d arrays like `intensity`, `symbol`, etc (https://github.com/GenericMappingTools/pygmt/issues/1076), maybe we should call it `kind="xyzarrays"`? Or I'm ok with using `kind="empty"` too, just don't want to use `"none"` because it is confusing with `None`.

seisman: > maybe we should call it `kind="xyzarrays"`

Although not documented, I think in our codes, when we talk about "arrays", we usually mean numpy arrays, so a better name is "xyzvectors", but I feel it's confusing with the `"vectors"` kind. So, rename `none` to `empty` in https://github.com/GenericMappingTools/pygmt/pull/3482/commits/a1e67d3432c0818fcc1cf2e3eb299a7a83e515b2.

pygmt/helpers/utils.py
```diff
@@ -242,6 +244,11 @@ def data_kind(
     >>> data_kind(data=None, required=False)
     'arg'
 
+    The "empty" kind:
+
+    >>> data_kind(data=None)
+    'empty'
```
weiji14, line 250: Be more explicit by adding `required=True` here?

```suggestion
    >>> data_kind(data=None, required=True)
    'empty'
```

weiji14 (approved):
Looks ok, but I think wait for #3351 to be merged first before merging this one.

PR #3492, opened 2024-10-08
clib.conversion: Remove the as_c_contiguous function and use np.ascontiguousarray instead

seisman:

[np.ascontiguousarray](https://numpy.org/doc/2.0/reference/generated/numpy.ascontiguousarray.html) does the same thing with the `as_c_contiguous` function, and is even slightly faster (see the benchmarks below).

This PR removes the `as_c_contiguous` function (assuming that no one is using this non-public function so the function doesn't go inot a deprecation cycle) and uses `np.ascontiguousarray` instead. 

Benchmarks:
```
In [1]: from pygmt.clib.conversion import as_c_contiguous

In [2]: import numpy as np
```
For a small array:
```python
In [3]: c_array = np.arange(20).reshape((4, 5))

In [4]: c_array.flags
Out[4]:
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

In [5]: f_array = c_array.copy(order="F")

In [6]: f_array.flags
Out[6]:
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

In [7]: %timeit np.ascontiguousarray(c_array)
55.1 ns ± 0.246 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [8]: %timeit as_c_contiguous(c_array)
82 ns ± 1.28 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: %timeit np.ascontiguousarray(f_array)
331 ns ± 1.46 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [10]: %timeit as_c_contiguous(f_array)
411 ns ± 0.539 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
For a large array:
```python
In [12]: c_array = np.arange(2000000).reshape((4000, 500))

In [13]: f_array = c_array.copy(order="F")

In [14]: %timeit np.ascontiguousarray(c_array)
54.8 ns ± 0.289 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [15]: %timeit as_c_contiguous(c_array)
82.6 ns ± 0.788 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [16]: %timeit np.ascontiguousarray(f_array)
2.76 ms ± 21.2 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [17]: %timeit as_c_contiguous(f_array)
2.75 ms ± 15.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### DIFF
--- a/pygmt/clib/conversion.py
+++ b/pygmt/clib/conversion.py
@@ -123,7 +123,7 @@ def dataarray_to_matrix(grid):
         inc = [abs(i) for i in inc]
         grid = grid.sortby(variables=list(grid.dims), ascending=True)
 
-    matrix = as_c_contiguous(grid[::-1].to_numpy())
+    matrix = np.ascontiguousarray(grid[::-1].to_numpy())
     region = [float(i) for i in region]
     inc = [float(i) for i in inc]
     return matrix, region, inc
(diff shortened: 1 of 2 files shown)


weiji14 (approved):
This `np.ascontiguousarray` was available back in NumPy 1.13 (https://numpy.org/doc/1.13/reference/generated/numpy.ascontiguousarray.html) that was released on 8 Jun 2017, just when Leo wrote this function in #100 :laughing:

Ok with not having a deprecation cycle since there is a direct replacement. Also, I think this `np.ascontiguousarray` function supports the `__array_function__` protocol which might help with interoperability with other array types.

seisman:
> This `np.ascontiguousarray` was available back in NumPy 1.13 (https://numpy.org/doc/1.13/reference/generated/numpy.ascontiguousarray.html) that was released on 8 Jun 2017 (and possibly earlier), just when Leo wrote this function in #100 😆

The function was available even in NumPy v1.0 (https://github.com/numpy/numpy/blob/v1.0/numpy/core/numeric.py#L139). I guess Leo didn't know this function at that time, just like that I didn't know it until now.

PR #3502, opened 2024-10-10
clib: Switch the order of if-conditions to improve the Session.call_module performance

seisman:

In the old version, we first check if `args` is a string, if not then check if it's a list. The codes are like:
```
if isinstance(args, str):
    ...
elif isinstance(args, list):
    ...
```
Since in most cases, `args` is a list, it's better to check against list first, then string. In this way, the first if-condition is True, so it's slightly faster, although the performance improvement can be negligible.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -629,14 +629,7 @@ class Session:
 
         # 'args' can be (1) a single string or (2) a list of strings.
         argv: bytes | ctp.Array[ctp.c_char_p] | None
-        if isinstance(args, str):
-            # 'args' is a single string that contains whitespace-separated arguments.
-            # In this way, we need to correctly handle option arguments that contain
-            # whitespaces or quotation marks. It's used in PyGMT <= v0.11.0 but is no
-            # longer recommended.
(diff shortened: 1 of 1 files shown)


michaelgrund (approved):

pygmt/clib/session.py
```diff
@@ -629,31 +629,29 @@ def call_module(self, module: str, args: str | list[str]):
...
+            # In this way, we need to correctly handle option arguments that contain
+            # whitespaces or quotation marks. It's used in PyGMT <= v0.11.0 but is no
+            # longer recommended.
+            mode = self["GMT_MODULE_CMD"]
+            argv = args.encode()
         else:
-            raise GMTInvalidInput(
-                "'args' must be either a string or a list of strings."
-            )
+            msg = "'args' must be either a string or a list of strings."
```
weiji14, line 648: Should we reword the error message to recommend `list[str]` first?

```suggestion
            msg = "'args' must either be a list of strings (recommended) or a string."
```

weiji14 (approved):

PR #2845, opened 2023-12-03
pyarrow: Support date32[day] and date64[ms] dtypes in pandas objects

weiji14:
**Description of proposed changes**

Handle date columns in pandas.DataFrame with pyarrow dtypes like `date32[day][pyarrow]` or `date64[ms][pyarrow]`.

This is implemented by modifying the [`vectors_to_arrays`](https://github.com/GenericMappingTools/pygmt/blob/v0.10.0/pygmt/clib/conversion.py#L127) conversion function to convert date32/date64 dtypes to numpy.datetime64.

An altenative implementation would be to handle pyarrow date32/date64 dtypes natively without converting, but this might require more work and discussion at #2800.

Note:
- Unlike numpy which only has datetime64, PyArrow has separate [Date types (date64/date32)](https://arrow.apache.org/docs/python/pandas.html#date-types) and [Time types (time64)](https://arrow.apache.org/docs/python/pandas.html#time-types).

References:
- https://arrow.apache.org/docs/python/pandas.html#date-types
- https://arrow.apache.org/docs/13.0/python/generated/pyarrow.date32.html
- https://arrow.apache.org/docs/13.0/python/generated/pyarrow.date64.html

Part of #2800, related to #242, extends #464

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #2774, opened 2023-10-29
pyarrow: Check compatibility of pyarrow-backed pandas objects with numeric dtypes

weiji14:
**Description of proposed changes**

Pandas is moving towards using PyArrow backed arrays (previously just `numpy`), and will [require `pyarrow` in pandas 3.0](https://pandas.pydata.org/pandas-docs/version/2.1/whatsnew/v2.1.0.html#pyarrow-will-become-a-required-dependency-with-pandas-3-0), so we should ensure that `pyarrow` dtypes work across PyGMT.

This Pull Request installs `pyarrow` as an optional dependency, and updates some unit tests to check both `numpy` and `pyarrow`-backed dtypes for `pandas.Series` and `pandas.DataFrame` objects.

TODO:
- [x] Check regular numeric dtypes (uint/int/float)
- ~[ ] Check null and bool types ?~
- ~[ ] Check string dtypes~
- ~[ ] Check datetime dtypes~

Relates to https://github.com/GenericMappingTools/pygmt/issues/1318#issuecomment-1756355245, part of #2800

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3507, opened 2024-10-11
clib.conversion: Remove the unused array_to_datetime function

seisman:
After a series of PRs (especially #3760), the `_to_numpy` function takes the responsibility to convert any array-like objects into numpy arrays with recognized dtypes. So, if `_to_numpy` returns a `np.object` dtype string, it means the original array cannot to converted to a numpy array that can be passed to GMT.

After #3760, the `array_to_datetime` function is no longer used and can be removed.


pygmt/clib/session.py
```diff
@@ -1388,7 +1378,7 @@ def virtualfile_from_vectors(self, *vectors):
         # Assumes that first 2 columns contains coordinates like longitude
         # latitude, or datetime string types.
         for col, array in enumerate(arrays[2:]):
-            if pd.api.types.is_string_dtype(array.dtype):
+            if array.dtype.type == np.str_:
```
seisman, line 1381: `pd.api.types.is_string_dtype` was added in PR https://github.com/GenericMappingTools/pygmt/pull/684. Before PR #684, we used `np.issubdtype(array.dtype, np.str_)`, which was added in #520. Any specific reason that we should use `np.issubdtype(array.dtype, np.str_)`, not `array.dtype.type == np.str_`?

pygmt/clib/conversion.py
```diff
@@ -313,6 +320,11 @@ def array_to_datetime(array: Sequence[Any]) -> np.ndarray:
 
     If the input array is not in legal datetime formats, raise a ValueError exception.
 
+    .. deprecated:: 0.14.0
```
seisman, line 323: After this PR, `array_to_datetime` is no longer used, but I still want to keep this function so that we know what kinds of datetime formats that `np.asarray(array, dtype=np.datetime64)` can support.

michaelgrund (approved):

pygmt/clib/session.py
weiji14: I think we'll need to check if this can handle [`pandas.StringDtype`](https://pandas.pydata.org/docs/reference/api/pandas.StringDtype.html) and [`pyarrow.StringArray`](https://arrow.apache.org/docs/python/generated/pyarrow.StringArray.html#pyarrow-stringarray) (xref #2933).

seisman: Both can be converted to the numpy string dtype by the `vectors_to_arrays` method, so in `virtualfile_from_vectors`, we only need to check `np.str_`:
```
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: import pyarrow as pa

In [4]: x = pd.Series(["abc", "defghi"], dtype="string")

In [5]: np.asarray(x)
Out[5]: array(['abc', 'defghi'], dtype=object)

In [6]: np.asarray(x, dtype=str)
Out[6]: array(['abc', 'defghi'], dtype='<U6')

In [7]: y = pa.array(["abc", "defghi"])

In [8]: np.asarray(y)
Out[8]: array(['abc', 'defghi'], dtype=object)

In [9]: np.asarray(y, dtype=str)
Out[9]: array(['abc', 'defghi'], dtype='<U6')
```

seisman: The main idea of this PR is to let `vectors_to_arrays` do the conversion work from any dtypes (including `pd.StringDtype` and `pa.StringArray`) into the numpy dtypes, so that the `virtualfile_from_vectors` only need to care about how to map numpy dtypes into GMT data types.

For any special dtypes that we know how to convert it to numpy dtype, we can maintain a mapping dictionary, just like what you did to support pyarrow's date32[day] and date64[ms] in #2845:

https://github.com/GenericMappingTools/pygmt/blob/c2e429c0262f4dd49a87be711cfa0883eebb408e/pygmt/clib/conversion.py#L208-L211

seisman: In 83673cf75530fe307089a30328e2e1e54055eee2, I've moved most of the doctests into a separate test file `pygmt/tests/test_clib_conversion.py`. A test `test_vectors_to_arrays_pandas_string` is added to check that the function can handle `pd.StringDtype` correctly.

seisman: > For any special dtypes that we know how to convert it to numpy dtype, we can maintain a mapping dictionary, just like what you did to support pyarrow's date32[day] and date64[ms] in #2845:
> 
> https://github.com/GenericMappingTools/pygmt/blob/c2e429c0262f4dd49a87be711cfa0883eebb408e/pygmt/clib/conversion.py#L208-L211

Based on the tests below, I think we should add the entry `"string": np.str_` to the dictionary:

```python
In [1]: import pandas as pd

In [2]: x = pd.Series(["abc", "12345"])

In [3]: x.dtype
Out[3]: dtype('O')

In [4]: str(x.dtype)
Out[4]: 'object'

In [5]: x = pd.Series(["abc", "12345"], dtype="string")

In [6]: x.dtype
Out[6]: string[python]

In [7]: str(x.dtype)
Out[7]: 'string'

In [8]: x = pd.Series(["abc", "12345"], dtype="string[pyarrow]")

In [9]: x.dtype
Out[9]: string[pyarrow]

In [10]: str(x.dtype)
Out[10]: 'string'

In [11]: import pyarrow as pa

In [12]: x = pa.array(["abc", "defghi"])

In [13]: x.type
Out[13]: DataType(string)

In [14]: str(x.type)
Out[14]: 'string'
```

seisman: In PR #2774 and #2774, we only checked if PyGMT supports pandas with the pyarrow backend, but didn't check if the original pyarrow arrays works. For example, for a pyarrow `date32` array, we need to check `array.type` rather than `array.dtype`:
```
In [1]: import datetime

In [2]: import pyarrow as pa

In [3]: x = pa.array([datetime.date(2020, 1, 1), datetime.date(2021, 12, 31)])

In [4]: str(x.type)
Out[4]: 'date32[day]'
```

seisman:
I've reverted some changes in e26afbf5fa6cb55c12f72390cc09127042f65a9c so that this PR can focus on refactoring the existing codes without introducing more enhancements.

We will revisit the issues/enhancements mentioned in https://github.com/GenericMappingTools/pygmt/pull/3507#discussion_r1799432258 later.

pygmt/clib/conversion.py
```diff
@@ -168,41 +168,7 @@ def vectors_to_arrays(vectors: Sequence[Any]) -> list[np.ndarray]:
     True
     >>> all(isinstance(i, np.ndarray) for i in arrays)
     True
-
```
seisman, line 171: Most of the doctests are moved to `pygmt/tests/test_clib_vectors_to_arrays.py`.

pygmt/clib/session.py
weiji14: > For example, for a pyarrow `date32` array, we need to check `array.type`

Yes, raw `pyarrow` arrays of `date32/date64` type are not supported yet. That's why it's marked :construction: in https://github.com/GenericMappingTools/pygmt/issues/2800 (I was planning to modify `array_to_datetime` to handle it).

pygmt/tests/test_clib_vectors_to_arrays.py
```diff
@@ -80,6 +80,19 @@ def test_vectors_to_arrays_pandas_nan():
     _check_arrays(arrays)
 
 
+def test_vectors_to_arrays_pandas_string():
+    """
+    Test the vectors_to_arrays function with pandas Series containing datetime64.
+    """
+    vectors = [
+        pd.Series(["abc", "defhig"]),
+        pd.Series(["abcdef", "123456"], dtype="string"),
```
weiji14, line 89: There are 3 'string' dtypes in pandas now according to https://pandas.pydata.org/pandas-docs/version/2.2/reference/api/pandas.StringDtype.html:

- `string` or `string[python]` (default as of Pandas 2.x)
- `string[pyarrow]`
- `string[pyarrow_numpy]`

I think the expectation is that in pandas 3.0, the default StringDtype will change from `string[python]` to `string[pyarrow]` (see https://pandas.pydata.org/pdeps/0010-required-pyarrow-dependency.html). So maybe we should explicitly test for `string[python]` here? Or maybe test for all three variants.

seisman: Done in d8777e5e8f4878a463a4524f65033c95a1352dc6.

pygmt/tests/test_clib_vectors_to_arrays.py
```diff
@@ -80,6 +81,27 @@ def test_vectors_to_arrays_pandas_nan():
...
+    "dtype",
+    [
+        "string[python]",
+        pytest.param("string[pyarrow]", marks=skip_if_no(package="pyarrow")),
+        pytest.param("string[pyarrow_numpy]", marks=skip_if_no(package="pyarrow")),
+    ],
+)
+def test_vectors_to_arrays_pandas_string(dtype):
+    """
+    Test the vectors_to_arrays function with pandas strings.
```
weiji14, line 94:
```suggestion
    Test the vectors_to_arrays function with pandas.Series objects of dtype
    string[python], string[pyarrow] and string[pyarrow_numpy].
```

pygmt/clib/conversion.py
```diff
@@ -273,12 +280,17 @@ def strings_to_ctypes_array(strings: Sequence[str] | np.ndarray) -> ctp.Array:
...
+def _array_to_datetime(array: Sequence[Any] | np.ndarray) -> np.ndarray:
     """
     Convert a 1-D datetime array from various types into numpy.datetime64.
 
     If the input array is not in legal datetime formats, raise a ValueError exception.
 
+    .. deprecated:: 0.14.0
+
+       The function is no longer used in the PyGMT project, but we keep this function
+       to docuemnt and test the supported datetime types.
```
weiji14, line 292:
```suggestion
       This function is no longer used in the PyGMT project, but we keep it to document
       and test supported datetime types.
```

weiji14 (approved):

seisman:
Will revisit changes in this PR after #3583, #3584, #3585.

pygmt/clib/session.py
```diff
@@ -934,11 +933,6 @@ def _check_dtype_and_dim(self, array: np.ndarray, ndim: int) -> int:
             msg = f"Expected a numpy {ndim}-D array, got {array.ndim}-D."
             raise GMTInvalidInput(msg)
 
-        # For 1-D arrays, try to convert unknown object type to np.datetime64.
-        if ndim == 1 and array.dtype.type is np.object_:
-            with contextlib.suppress(ValueError):
-                array = array_to_datetime(array)
```
seisman, line 940: Moved to the `_to_numpy` function in PR #3760.

pygmt/clib/conversion.py
```diff
@@ -344,81 +344,3 @@ def strings_to_ctypes_array(strings: Sequence[str] | np.ndarray) -> ctp.Array:
...
-    array
-        1-D datetime array in numpy.datetime64.
-
-    Raises
-    ------
-    ValueError
-        If the datetime string is invalid.
-
-    Examples
-    --------
```
seisman, line 378: These doctests are already covered by the tests in `test_clib_to_numpy.py`

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final reviews. I plan to merge this PR on 2025/02/08 if there are no objections.

PR #1382, opened 2021-07-10
Add support for passing pathlib.Path objects as filenames

aitorres:
**Description of proposed changes**

This PR addresses and fixes issue #1381. By adding support to `pathlib.Path` values that represent file paths to the `data_kind` helper function, it can be used interchangeably with file name strings on `pygmt` modules.

Fixes #1381

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [x] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3521, opened 2024-10-16
clib.Session.virtualfile_in: Refactor if-else statements into match-case statements

seisman:

This is another refactor about `virtualfile_in` after PR #3482 and PR #3351. This PR converts the `if-else` statements into match-case statements.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -8,7 +8,6 @@ Uses ctypes to wrap most of the core functions from the C API.
 import contextlib
 import ctypes as ctp
 import io
-import pathlib
 import sys
 import warnings
 from collections.abc import Callable, Generator, Sequence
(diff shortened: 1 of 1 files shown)


pygmt/clib/session.py
```diff
@@ -1792,43 +1791,43 @@ def virtualfile_in(  # noqa: PLR0912
...
-            # of vectors and allows for better handling of non-integer/float type inputs
-            # (e.g. for string or datetime data types).
-            _data = (data,)
-            if data.dtype.kind not in "iuf":
-                _virtualfile_from = self.virtualfile_from_vectors
-                _data = data.T
+        # Ensure the data is an iterable (Python list or tuple).
+        match kind:
+            case "arg" | "file" | "geojson" | "grid" | "image" | "stringio":
+                _data = (data,)
```
seisman, line 1797: `_data = (data,) if not isinstance(data, pathlib.PurePath) else (str(data),)` is changed to `_data = (data,)`. 

For reference, the code `str(data)` was initially added in https://github.com/GenericMappingTools/pygmt/pull/1382 to support passing pathlib.Path as filenames. It turns out it's no longer necessary because the `build_arg_list` function already handle with `pathlib.Path` objects correctly:

https://github.com/GenericMappingTools/pygmt/blob/d41922ca4b4122e031d65d82bddea33798ea2ffd/pygmt/helpers/utils.py#L516-L520.

The support of pathlib.Path objects can even be tracked back to the old, deprecated `build_arg_string` function (xref: https://github.com/GenericMappingTools/pygmt/pull/1837).

We already have a lot of tests that takes a pathlib.Path object as input (e.g., https://github.com/GenericMappingTools/pygmt/blob/d41922ca4b4122e031d65d82bddea33798ea2ffd/pygmt/tests/test_info.py#L20). So the changes should be safe if all tests pass.

weiji14 (approved):

PR #3522, opened 2024-10-16
clib.Session.virtualfile_from_vectors: Now takes a sequence of vectors as its single argument (Passing multiple arguments will be unsupported in v0.16.0)

seisman:
In the `Session.virtualfile_in` method, we must prepare for `_data` that will be passed to different `virtualfile_from_*` methods. Currently, `_data` must be an iterable (e.g., `_data = (data, )` even for the simplest case) and when pass it to `_virtualfile_from`, we need to unpack it like:
```
_virtualfile_from(*data)
```

The complexity is because, while most `_virtualfile_from_*` methods only take a single argument, the `_virtualfile_from_vectors` takes multiple arguments. The `_virtualfile_from_vectors` is defined like `_virtualfile_from_vectors(*vectors)` and must be called like `_virtualfile_from_vectors(x, y, z)`.

This PR changes its definition to `_virtualfile_from_vectors(vectors)`, and now it should be called like `_virtualfile_from_vectors((x, y, z))`.

~~With this breaking change, the `virtualfile_in` function can be simplified slightly, which I think is more readable. Then the question is, is it worth a breaking change?~~

Edit: Actually it's possible to introduce the new syntax without breaking backward compatibility. This is done in f5b0e8e99b38c9e8944dbfad05fd2fc8f1998542.


pygmt/clib/session.py
```diff
@@ -1330,7 +1330,7 @@ def open_virtual_file(self, family, geometry, direction, data):
         return self.open_virtualfile(family, geometry, direction, data)
 
     @contextlib.contextmanager
-    def virtualfile_from_vectors(self, *vectors):
+    def virtualfile_from_vectors(self, vectors, *args):
```
weiji14, line 1333: Also need to update the docstring at L1355 below.

```suggestion
    def virtualfile_from_vectors(self, vectors: tuple, *args):
```

pygmt/clib/session.py
```diff
@@ -1372,13 +1372,24 @@ def virtualfile_from_vectors(self, *vectors):
...
+        ...     with ses.virtualfile_from_vectors((x, y, z)) as fin:
         ...         # Send the output to a file so that we can read it
         ...         with GMTTempFile() as fout:
         ...             ses.call_module("info", [fin, f"->{fout.name}"])
         ...             print(fout.read().strip())
         <vector memory>: N = 3 <1/3> <4/6> <7/9>
         """
+        if len(args) > 0:
+            warnings.warn(
+                "Passing multiple arguments to Session.virtualfile_fro_vectors is "
```
weiji14, line 1384:
```suggestion
                "Passing multiple arguments to Session.virtualfile_from_vectors is "
```

pygmt/clib/session.py
```diff
@@ -1791,19 +1802,19 @@ def virtualfile_in(  # noqa: PLR0912
             "vectors": self.virtualfile_from_vectors,
         }[kind]
 
-        # Ensure the data is an iterable (Python list or tuple).
+        # "_data" is the data that will be passed to the _virtualfile_from function.
+        # "_data" defaults to "data" but should be adjusted for some cases.
+        _data = data
         match kind:
-            case "arg" | "file" | "geojson" | "grid" | "image" | "stringio":
```
weiji14, line 1796: Maybe just put `_data = data` in the `case _` below? I'd actually prefer if we make it more explicit like:

```
case "arg" | "file" | "geojson" | "grid" | "stringio":
    _data = data
```

So that if there's a new `kind` in the future, we know to add it explicitly.

pygmt/clib/session.py
seisman: I use `Sequence` in f786b8e996aeb7c8bfc1e39fe42905fc7a000720, since a list of vectors is also allowed.

pygmt/clib/session.py
seisman: > I'd actually prefer if we make it more explicit like:
> 
> ```
> case "arg" | "file" | "geojson" | "grid" | "stringio":
>     _data = data
> ```
> 
> So that if there's a new `kind` in the future, we know to add it explicitly.

We can't do it like this, because we're currently writing case statements like `case "image" if data.dtype != "uint8":`, which only catches the case when `kind="image"` and `data.dtype != "uint8"`.

We can refactor the codes to:
```
case "arg" | "file" | "geojson" | "grid" | "stringio":
    _data = data
case "image":
    _data = data
    if data.dtype != "uint8":
        # do something
case "matrix":
    _data = data
    if data.dtype.kind not in "iuf":
    # do something
```
or 
```
case "arg" | "file" | "geojson" | "grid" | "stringio":
    _data = data
case "image" if data.dtype != "uint8":
    raise ...
case "matrix" if data.dtype.kind not in "iuf":
    ...
case _:
    _data = data # this is necessary to catch cases like kind="matrix" and dtype in "iuf"  
```

I still prefer the current codes, which set `_data` to `data` by default and the match-case statements only need to deal with exceptions.

> So that if there's a new `kind` in the future, we know to add it explicitly.

If there is a new `kind` in the future, `_data=data` is still the default. If needs special handling and we forget to add it, then the related tests will fail anyway.

weiji14: Ah ok, that makes sense, let's keep `_data = data` at the top then.

pygmt/clib/session.py
```diff
@@ -1371,34 +1370,49 @@ def virtualfile_from_vectors(self, *vectors):
...
-        # virtual file is closed. The conversion is implicit in
+        # "*args" is added in v0.14.0 for backward-compatibility with the deprecated
+        # syntax of passing multiple vectors as positional arguments.
+        # Remove it in v0.16.0.
+        if len(args) > 0:
+            msg = (
+                "Passing multiple arguments to Session.virtualfile_from_vectors is "
+                "deprecated since v0.14.0 and will be unsupported in v0.16.0. "
+                "Put all vectors in a sequence (a tuple or a list) instead and pass "
+                "the sequnece as the single argument to this function. "
```
seisman, line 1388:
```suggestion
                "the sequence as the single argument to this function. "
```

pygmt/clib/session.py
```diff
@@ -1371,34 +1370,49 @@ def virtualfile_from_vectors(self, *vectors):
...
+                "e.g., use `with lib.virtualfile_from_vectors((x, y, z)) as vfile` "
```
seisman, line 1389:
```suggestion
                "E.g., use `with lib.virtualfile_from_vectors((x, y, z)) as vfile` "
```

weiji14 (approved):
Getting very close to (finally) implementing https://github.com/GenericMappingTools/pygmt/issues/1076!

I've changed the label from 'maintenance' to 'deprecation'. Should we mention this is scheduled for removal in PyGMT v0.16.0?

PR #3523, opened 2024-10-16
clib.session: Refactor to simplify the checking of currently open session

seisman:
Just some refactoring to simplify the codes. See the changes for more details.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -160,7 +160,7 @@ class Session:
             If trying to access without a currently open GMT session (i.e., outside of
             the context manager).
         """
-        if not hasattr(self, "_session_pointer") or self._session_pointer is None:
+        if getattr(self, "_session_pointer", None) is None:
             raise GMTCLibNoSessionError("No currently open GMT API session.")
         return self._session_pointer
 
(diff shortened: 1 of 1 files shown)


pygmt/clib/session.py
```diff
@@ -160,7 +160,7 @@ def session_pointer(self) -> ctp.c_void_p:
             If trying to access without a currently open GMT session (i.e., outside of
             the context manager).
         """
-        if not hasattr(self, "_session_pointer") or self._session_pointer is None:
+        if getattr(self, "_session_pointer", None) is None:
```
seisman, line 163: `if not hasattr(self, "_session_pointer") or self._session_pointer is None:` is equivalent to `if getattr(self, "_session_pointer", None) is None:`.

pygmt/clib/session.py
```diff
@@ -338,19 +338,12 @@ def create(self, name: str):
         name
             A name for this session. Doesn't really affect the outcome.
         """
```
seisman, line 340: If `self.session_pointer` is defined and is not None, then we know we have an open session. The `try..except` statement was added in 2018 (#210) and is not necessary.

michaelgrund (approved):

pygmt/clib/session.py
seisman: It's impossible that the `msg` assignment and the `raise GMTCLibError` lines can raise a `GMTCLibNoSessionError` exception, so we should be safe.

seisman: Reverted back to `try..except` in da9c5e17abdffbba49fc49bb9d82f49b2a794bc2 and improve the comments to make it easier to understand.

pygmt/clib/session.py
```diff
@@ -338,17 +338,16 @@ def create(self, name: str):
...
+        # free to create a new one. Otherwise, we will raise a GMTCLibError exception.
         try:
-            # Won't raise an exception if there is a currently open session.
             _ = self.session_pointer
-            # In this case, fail to create a new session until the old one is destroyed.
             msg = (
-                "Failed to create a GMT API session: There is a currently open session."
-                " Must destroy it first."
+                "Failed to create a GMT API session: "
+                "There is a currently open session. Must destroy it first."
```
weiji14, line 348:
```suggestion
                "There is currently an open session. Must destroy it first."
```

weiji14 (approved):

PR #3537, opened 2024-10-19
Simplify the _get_default_display_method tests using unittest.mock

seisman:
Refer to https://github.com/pytest-dev/pytest/issues/4576#issuecomment-449864333 for comparisons of pytest.monkeypatch, unittest.mock and pytest-mock.

[`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html) provides more features than `pytest.monkeypatch` which can simplify our tests. For example, with [`mock.call_count`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count), we can directly count how many times a function/method is called, rather than implementing our own counter (like what're doing in https://github.com/GenericMappingTools/pygmt/blob/d1ab97ca49fd53dcd6d8cf268d344498d39642e5/pygmt/tests/test_clib_loading.py#L224). 

Personally, I prefer `pytest-mock` which is as powerful as `unittest.mock` and can be used as a fixture, while `unittest.mock` must be used as a context manager or a decorator.

I think we have many options:

1. `pytest.monkeypatch` only, which means in some cases, we have to write more codes for tests
2. `unittest.mock` only or `pytest-mock` only
3. `pytest.monkeypatch` + `unittest.mock`
4. `pytest.monkeypatch` + `pytest-mock`


pygmt/tests/test_figure.py
```diff
@@ -436,26 +433,16 @@ def test_disable_external_display(self, monkeypatch):
         assert _get_default_display_method() == "none"
 
     @pytest.mark.skipif(not _HAS_IPYTHON, reason="Run when IPython is installed")
-    def test_notebook_display(self, monkeypatch):
+    # Mock IPython.get_ipython() to return an object with a config attribute, so PyGMT
+    # can detect that an IPython kernel is running.
+    @patch("IPython.get_ipython", return_value=Mock(config={"IPKernelApp": True}))
```
weiji14, line 438: Haven't played around with `unittest.mock` much yet, but does this `unittest.mock.path` call need to be here on top as a decorator, rather than in the test function below? I find this looks a little unintuitive (even though it is less code) compared to setting up the MockIPython class inside the unit test which makes it clear what is happening.

seisman: Here are different ways to mock IPython:

1. Define a custom class, as in the current main branch and use `pytest.monkeypatch`
```
class MockIPython:
    def __init__(self):
        self.config = {"IPKernelApp": True}

    # Mock IPython.get_ipython() to return a MockIPython instance.
    mock_ipython = MockIPython()
    monkeypatch.setattr(IPython, "get_ipython", lambda: mock_ipython)
```

2. Use `mock`/`patch` as a decorator
```
@patch("IPython.get_ipython", return_value=Mock(config={"IPKernelApp": True}))
def test_xxxx():
    ...
```

3. Use `mock`/`patch` as a context manager
```
def test_xxxx():
    with patch("IPython.get_ipython", return_value=Mock(config={"IPKernelApp": True})):
        ....
```
Maybe we should use option 3 although it means more indentations?

weiji14: Yes, option 3 sounds good.

seisman: OK, I've changed it to context manager in d08001c6e46a0a217a1e40f08e19ad2a37225249.

weiji14 (approved):

PR #3540, opened 2024-10-21
Fix badge links, use consistent badge style and update badge order

seisman:

1. Fix the pypi link from https://pypi.python.org/pypi/pygmt to https://pypi.org/project/pygmt
2. Fix the badges to the "Tests" and "GMT Dev Tests" workflows. 
3. Currently, badges have different styles. Some use the default (flat) style while others use the "flat-square" style. This PR ensures that all badges use the same style. As shown below:

**Before:**
image
**After:**
image

**After e34263518abca51049815fa423c0faf496853ade**
image

**After 6a60e1330ab9a52274c9a4259e8bc86d93e8c121**
image


michaelgrund (approved):

README.md
```diff
@@ -4,18 +4,18 @@
...
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
-[![Test coverage status](https://codecov.io/gh/GenericMappingTools/pygmt/branch/main/graph/badge.svg?token=78Fu4EWstx)](https://app.codecov.io/gh/GenericMappingTools/pygmt)
+[![Test coverage status](https://codecov.io/gh/GenericMappingTools/pygmt/graph/badge.svg?token=78Fu4EWstx)](https://app.codecov.io/gh/GenericMappingTools/pygmt)
 [![CodSpeed Performance Benchmarks](https://img.shields.io/endpoint?url=https://codspeed.io/badge.json)](https://codspeed.io/GenericMappingTools/pygmt)
-[![Compatible Python versions.](https://img.shields.io/pypi/pyversions/pygmt.svg?style=flat-square)](https://pypi.python.org/pypi/pygmt)
-[![Discourse forum](https://img.shields.io/discourse/status?label=forum&server=https%3A%2F%2Fforum.generic-mapping-tools.org%2F&style=flat-square)](https://forum.generic-mapping-tools.org)
+[![Compatible Python versions](https://img.shields.io/pypi/pyversions/pygmt)](https://pypi.org/project/pygmt)
```
yvonnefroehlich, line 14: Seems that there is no specific link to the supported Python versions. Currently we have here again the link to the PyPi website for PyGMT.
Hm, just wondering if we can maybe link here to the newly added "Minimum Supported Versions" Technical Reference page?

```suggestion
[![Compatible Python versions](https://img.shields.io/pypi/pyversions/pygmt)](https://www.pygmt.org/dev/minversions.html)
```

yvonnefroehlich (commented):
Cool, that the badges now look consistent!
(Actually I was wondering, why they are provided in different styles, but never checked if this can be adjusted by the users...)

README.md
seisman: That's a good point. Currently, we use the badge from https://shields.io/badges/py-pi-python-version, which means the supported Python versions on PyPI. 

[Compatible Python versions](https://pypi.org/project/pygmt)

So I feel linking to https://pypi.org/project/pygmt makes more sense than https://www.pygmt.org/dev/minversions.html.

I also found another badge at https://shields.io/badges/python-version-from-pep-621-toml. It can show the Python version provided in the `pyproject.toml` file, so it reflects the supported Python versions for the PyGMT dev. For this badge, I feel it makes more sense to linking it to https://www.pygmt.org/dev/minversions.html:

Python Version from PEP 621 TOML

I prefer this badge. What do you think?

yvonnefroehlich: I think I like this one, as it is shorter and linking to our version overview appears to me more helpful for users.

seisman: Ping @michaelgrund and @weiji14 for comments on adopting the new badge.

Supported Python versions

seisman: I've made the change. Now the badges look like:

image

seisman:
I also would like to take this chance to update the orders of these badges. Here is my preferred order:

image

- The first five badges (from PyPI version to DOI) provide information that normal users care
- The next three badges (from forum to CoC) provide information about the community
- The last five badges provide project information that only contributors/maintainers care

Let me know if you have other thoughts.

PR #3548, opened 2024-10-23
pygmt.set_display: Improve the test by checking if the preview image is opened in the expected way

seisman:

The `set_display` method can configure the default display method when calling `Figure.show`. Previously, our test only checks if the function changes the global variable `SHOW_CONFIG` correctly and doesn't check if `Figure.show` works in the expected way.

This PR improves the test to make sure that `Figure.show` works as expected after calling `set_display`.

### DIFF
--- a/pygmt/tests/test_figure.py
+++ b/pygmt/tests/test_figure.py
@@ -377,13 +377,45 @@ class TestSetDisplay:
 
     def test_set_display(self):
         """
-        Test if pygmt.set_display updates the SHOW_CONFIG variable correctly.
+        Test if pygmt.set_display updates the SHOW_CONFIG variable correctly and
+        Figure.show opens the preview image in the correct way.
         """
-        default_method = SHOW_CONFIG["method"]  # Current default method
+        default_method = SHOW_CONFIG["method"]  # Store the current default method.
 
-        for method in ("notebook", "external", "none"):
-            set_display(method=method)
-            assert SHOW_CONFIG["method"] == method
+        fig = Figure()
+        fig.basemap(region=[0, 3, 6, 9], projection="X1c", frame=True)
+        with (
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_figure.py
```diff
@@ -377,13 +377,50 @@ class TestSetDisplay:
...
-            set_display(method=method)
-            assert SHOW_CONFIG["method"] == method
+        default_method = SHOW_CONFIG["method"]  # Store the current default method.
+
+        fig = Figure()
+        fig.basemap(region=[0, 3, 6, 9], projection="X1c", frame=True)
+
+        # Test the "notebook" display method.
+        set_display(method="notebook")
+        assert SHOW_CONFIG["method"] == "notebook"
```
seisman, line 390: Using `unittest.mock.patch`.

Need to note that we must pacth `pygmt.figure.launch_external_viewer` rather than `pygmt.helpers.utils.launch_external_viewer`.

See https://docs.python.org/3/library/unittest.mock.html#where-to-patch for the reasons:

> The basic principle is that you patch where an object is looked up, which is not necessarily the same place as where it is defined.

pygmt/tests/test_figure.py
```diff
@@ -377,13 +377,50 @@ class TestSetDisplay:
...
+        if _HAS_IPYTHON:
+            with (
+                patch("IPython.display.display") as mock_display,
+                patch("pygmt.figure.launch_external_viewer") as mock_viewer,
+            ):
+                fig.show()
+                assert mock_viewer.call_count == 0
```
seisman, line 397: Instead of checking `mock_display.call_count == 1`, we can also use `mock_display.assert_called_once`, but I personally feel `call_count` is more flexible and readable.

PR #3549, opened 2024-10-23
clib.conversion.vectors_to_arrays: Move inline doctests into a separate test file

seisman:
This PR contains a subset of changes in PR #3507, so that we can minimize changes in PR #3507.

This PR moves most of the inline doctests into `pygmt/tests/test_clib_vectors_to_arrays.py`. We may add more tests in the future.

### DIFF
--- a/pygmt/tests/test_clib_vectors_to_arrays.py
+++ b/pygmt/tests/test_clib_vectors_to_arrays.py
@@ -75,7 +75,9 @@ def test_vectors_to_arrays_pandas_nan():
     """
     vectors = [pd.Series(data=[0, 4, pd.NA, 8, 6], dtype=pd.Int32Dtype())]
     arrays = vectors_to_arrays(vectors)
-    npt.assert_equal(arrays[0], np.array([0, 4, np.nan, 8, 6], dtype=np.float64))
+    npt.assert_equal(
+        arrays[0], np.array([0, 4, np.nan, 8, 6], dtype=np.float64), strict=True
+    )
     _check_arrays(arrays)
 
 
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_clib_vectors_to_arrays.py
```diff
@@ -0,0 +1,99 @@
...
+    vectors = [pd.Series(data=[0, 4, pd.NA, 8, 6], dtype=pd.Int32Dtype())]
+    arrays = vectors_to_arrays(vectors)
+    npt.assert_equal(arrays[0], np.array([0, 4, np.nan, 8, 6], dtype=np.float64))
+    _check_arrays(arrays)
+
+
+@pytest.mark.skipif(not _HAS_PYARROW, reason="pyarrow is not installed.")
+def test_vectors_to_arrays_pyarrow_datetime():
+    """
+    Test the vectors_to_arrays function with pyarrow arrays containing datetime64.
```
weiji14, line 85:
```suggestion
    Test the vectors_to_arrays function with pyarrow arrays containing date32/date64
    types.
```

pygmt/tests/test_clib_vectors_to_arrays.py
```diff
@@ -0,0 +1,99 @@
...
+    _check_arrays(arrays)
+
+
+def test_vectors_to_arrays_pandas_nan():
+    """
+    Test the vectors_to_arrays function with pandas Series containing NaNs.
+    """
+    vectors = [pd.Series(data=[0, 4, pd.NA, 8, 6], dtype=pd.Int32Dtype())]
+    arrays = vectors_to_arrays(vectors)
+    npt.assert_equal(arrays[0], np.array([0, 4, np.nan, 8, 6], dtype=np.float64))
```
weiji14, line 78: I feel like `vectors_to_arrays` should convert `pd.Int32Dtype()` to `np.float32` instead of `np.float64`, but ok with this for now if that is the current behaviour.

weiji14 (approved):

pygmt/tests/test_clib_vectors_to_arrays.py
seisman: The behavior is consistent with the latest pandas versions.

Issue #3357, opened 2024-07-25
Move module-specific common codes to a separate file?


seisman:
Some modules may share the same or similar codes. For example,

`Figure.plot` has several lines of codes that check if data has a geometry of Point or MultiPoint, and then set the default style.

https://github.com/GenericMappingTools/pygmt/blob/ecaab8fa60e41aca0fd17fdbaf80bad109692b51/pygmt/src/plot.py#L246-L255.

Very similar codes are also used in https://github.com/GenericMappingTools/pygmt/blob/ecaab8fa60e41aca0fd17fdbaf80bad109692b51/pygmt/src/plot3d.py#L222.

The `Figure.meca` wrapper has functions like
https://github.com/GenericMappingTools/pygmt/blob/ecaab8fa60e41aca0fd17fdbaf80bad109692b51/pygmt/src/meca.py#L12
and https://github.com/GenericMappingTools/pygmt/blob/ecaab8fa60e41aca0fd17fdbaf80bad109692b51/pygmt/src/meca.py#L103

They can be used directly in `Figure.coupe` (not wrapped yet).

So, instead of duplicating the same/similar codes in different files, we should try to reuse them as much as possible.

I think we have two options:

1. Define small function in one file and import the function in another file. (e.g., define a private function like `_determine_style` in `Figure.plot` and use `from pygmt.src.plot import _determine_style` in `Figure.plot3d`.
2. Define these functions in a common file and import the functions when necessary

I prefer option 2. Since these functions are module-specific, I think it makes more sense to put them in a file under `pygmt/src` (e.g, `pygmt/src/_common.py`). I also feel that these functions should all be private.

Thoughts @GenericMappingTools/pygmt-maintainers

Issue #2019, opened 2022-07-21
pscoupe implementation


konalindsey:
**Description of the desired feature**

I am mapping cross sections of subduction zones, and need to plot focal mechanisms. 
I tried using Session.call_module, but ran into various errors... I probably am not using it correctly, but a real pyGMT function would be nice.

**Are you willing to help implement and maintain this feature?** Yes/No

yvonnefroehlich:
Thanks @konalindsey for trying out PyGMT :smiley: !

Maybe the method [pygmt.Figure.meca](https://www.pygmt.org/latest/api/generated/pygmt.Figure.meca.html) is helpful for you?

There is also a gallery example for this method: https://www.pygmt.org/latest/gallery/seismology/meca.html#sphx-glr-gallery-seismology-meca-py

_Please note_:  This method was fully rewritten after release v0.7.0 (see https://github.com/GenericMappingTools/pygmt/issues/2016#issuecomment-1189135191). Despite this, there are still some known issues and limitations with this method (e.g. https://github.com/GenericMappingTools/pygmt/issues/2016).

konalindsey:
Yes, meca works great! But I need to make cross sections of the focal mechanisms which would correspond to pscoupe in GMT.

yvonnefroehlich:
I have no experience with `pscoupe`. So, probably the PyGMT maintainers can help you better and faster :wink:, regarding both wrapping `pscoupe` and using `pscoupe` in PyGMT via `Session.call_module`.

> I tried using Session.call_module, but ran into various errors... I probably am not using it correctly,

I think it could be helpful if you post your code or a minimal example showing your issues when using `pscoupe` via `Session.call_module`.

yvonnefroehlich:
Ping at @seisman, @weiji14, @maxrjones, @willschlitzer, and @michaelgrund for help on this, please.

seisman:
I don't think I have time to wrap the `coupe` module recently, but here is a minimal script showing how to call a module that is not wrapped in PyGMT yet.

The example data is available at https://raw.githubusercontent.com/GenericMappingTools/gmt-for-geodesy/main/5_seismology/meca.dat
```
import pygmt

spec = "meca.dat"
fig = pygmt.Figure()
with pygmt.clib.Session() as lib:
    file_context = lib.virtualfile_from_data(check_kind="vector", data=spec)
    with file_context as fname:
[...]

PR #3551, opened 2024-10-23
Figure.meca: Add the private _FocalMechanismConvention class to simplify codes

seisman:
Functions `convention_code`, `convention_name`, `convention_params` are defined in `meca.py` and they can also be used when wrapping `coupe` (#2019). So it's better to move these shared functions into `pygmt/src/_common.py` as proposed in #3357.

Instead of simply moving the functions, this PR adds the private `_FocalMechanismConvention` class which does the same thing as the three functions. With this new class, the `Figure.meca` source codes can be simplified.

It's likely the `Figure.meca` method can be further simplified, but will leave it to future PRs to make this PR small for reviewing.


pygmt/src/_common.py
```diff
@@ -39,3 +41,210 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+            self.code = self._conventions[convention]
+            if isinstance(self.code, dict):
+                self.code = self.code[component]
+        elif convention in self._codes:
+            # Convention is given as a single-letter code.
+            self.code = convention
+            self.convention = self._codes[convention]
+        else:
+            msg = f"Invalid focal mechanism convention '{convention}'."
+            raise GMTInvalidInput(msg)
```
weiji14, line 215: Wondering if some of the pattern matching on 'convention' long names (e.g. `akl`) or single-letter aliases (e.g. `a`) could be simplified using [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum), but not sure how to handle the extra logic around 'component's. Haven't looked into it too closely, but there's supposedly ways to extend Enums with extra methods - https://realpython.com/python-enum/#adding-and-tweaking-member-methods.

seisman: Yes, using `StrEnum` can simplify the code logic. The `_FocalMechanismConvention` class has been refactored in https://github.com/GenericMappingTools/pygmt/pull/3551/commits/c59b8a32d9e2219944f6034b73377f59c09307f5.

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+
+            - ``"aki"``: Aki and Richards convention.
+            - ``"gcmt"``: Global CMT (Centroid Moment Tensor) convention.
+            - ``"partial"``: Partial focal mechanism convention.
+            - ``"mt"``: Moment Tensor convention.
+            - ``"principal_axis"``: Principal axis convention.
+        component
+            The component of the seismic moment tensor to plot. Valid values are:
+
+            - ``"full"``: the full tensor seismic moment tensor
```
michaelgrund, line 184:
```suggestion
            - ``"full"``: the full seismic moment tensor
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+            - ``"dc"``: the closest double coupe defined from the moment tensor (zero
```
michaelgrund, line 185:
```suggestion
            - ``"dc"``: the closest double couple defined from the moment tensor (zero
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+    def __init__(
+        self,
+        convention: Literal["aki", "gcmt", "partial", "mt", "principal_axis"],
+        component: Literal["full", "deviatoric", "dc"] = "full",
+    ):
+        """
+        Initialize the _FocalMechanismConvention object from convention and component.
+
+        If the convention is specified via a single-letter code, the convention and
+        component are determined from the code.
```
yvonnefroehlich, line 169:
```suggestion
        If the convention is specified via a single-letter code, ``convention`` and
        ``component`` are determined from the code.
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+            - ``"principal_axis"``: Principal axis convention.
+        component
+            The component of the seismic moment tensor to plot. Valid values are:
+
+            - ``"full"``: the full seismic moment tensor
+            - ``"dc"``: the closest double couple defined from the moment tensor (zero
+              trace and zero determinant)
+            - ``"deviatoric"``: deviatoric part of the moment tensor (zero trace)
+
+            Doesn't apply to conventions ``"aki"``, ``"gcmt"``, and ``"partial"``.
```
yvonnefroehlich, line 189:
```suggestion
            Doesn't apply to the conventions ``"aki"``, ``"gcmt"``, and ``"partial"``.
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+        """
+        # TODO(Python>=3.12): Simplify to "convention in _FocalMechanismConventionCode".
+        if convention in _FocalMechanismConventionCode.__members__.values():
+            # Convention is specified via the actual single-letter convention code.
+            self.code = _FocalMechanismConventionCode(convention)
+            # Parse the convention from the convention code name.
+            self._convention = "_".join(self.code.name.split("_")[:-1]).lower()
+        else:  # Convention is specified via 'convention' and 'component'.
```
yvonnefroehlich, line 197:
```suggestion
        else:  # Convention is specified via "convention" and "component".
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+    @classmethod
+    def from_params(
+        cls,
+        params: Sequence[str],
+        component: Literal["full", "deviatoric", "dc"] = "full",
+    ) -> "_FocalMechanismConvention":
+        """
+        Create a _FocalMechanismConvention object from a sequence of parameters.
+
+        The method checks if the given parameters are a superset of a known focal
```
yvonnefroehlich, line 224:
```suggestion
        The method checks if the given parameters are a superset of a supported focal
```

pygmt/src/_common.py
```diff
@@ -39,3 +42,211 @@ def _data_geometry_is_point(data: Any, kind: str) -> bool:
...
+        ],
+    }
+
+    def __init__(
+        self,
+        convention: Literal["aki", "gcmt", "partial", "mt", "principal_axis"],
+        component: Literal["full", "deviatoric", "dc"] = "full",
+    ):
+        """
+        Initialize the _FocalMechanismConvention object from convention and component.
```
yvonnefroehlich, line 166:
```suggestion
        Initialize the ``_FocalMechanismConvention`` object from ``convention`` and ``component``.
```

yvonnefroehlich (commented):
Just some minor formulation and highlighting suggestions.

PR #3553, opened 2024-10-24
Move Figure.psconvert into a separate file

seisman:
The source code for `Figure.psconvert` has existed in the `pygmt/figure.py` file since #31. I think it makes more sense to move it into a separate file `pygmt/src/psconvert.py`.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -18,14 +18,7 @@ except ImportError:
 import numpy as np
 from pygmt.clib import Session
 from pygmt.exceptions import GMTError, GMTInvalidInput
-from pygmt.helpers import (
-    build_arg_list,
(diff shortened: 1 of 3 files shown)


pygmt/figure.py
```diff
@@ -374,6 +244,9 @@ def savefig(  # noqa: PLR0912
                 raise GMTInvalidInput(msg)
             kwargs["W"] = True
 
+        # pytest-mpl v0.17.0 added the "metadata" parameter to Figure.savefig, which
```
seisman, line 247: These lines were initially added to `Figure.psconvert` in #3054, but I feel it makes more sense to add the lines in `Figure.savefig` instead.

michaelgrund (approved):

Issue #3484, opened 2024-10-04
load_map_tiles: Allow a new parameter to control the projection of the returned DataArray


seisman:
Originally proposed in https://github.com/GenericMappingTools/pygmt/pull/2125#discussion_r1059868371

> > Perhaps we should convert the returned image to longitude/latitude because it's more commonly used than the Spherical Mercator coordinate system, following https://contextily.readthedocs.io/en/latest/warping_guide.html.
> 
> > Most tiles are originally served as Web Mercator (EPSG:3857), and reprojecting to latlon (EPSG:4326) would result in distortion. I'd prefer the reprojection to be a user controlled step (e.g. using rioxarray's .rio.to_crs), because if latlon isn't what the user wants, there would be double reprojection (EPSG:3857 -＞EPSG:4326 -＞EPSG:????) which is less accurate than single reprojection (EPSG:3857 -＞EPSG:????).

> > I'd prefer the reprojection to be a user controlled step (e.g. using rioxarray's .rio.to_crs)
> 
> What about having an option so that users can decide on the desired projection and don't have to learn the syntax of the rioxarray package?
> 

I think we can add a new parameter like `crs` to control the projection. `crs` should be default to `EPSG:3857`. If a different CRS is given, then we can use [contextily.warp_tiles](https://contextily.readthedocs.io/en/latest/reference.html#contextily.warp_tiles) to reproject the tiles.

With this new parameter, the following lines in `Figure.tilemap` can be simplified:

https://github.com/GenericMappingTools/pygmt/blob/68a17a0ab37514d4ada7af0bb7d3cd87a0dbefe0/pygmt/src/tilemap.py#L141-L145

PR #3554, opened 2024-10-25
load_tile_map: Add the new parameter 'crs' to set the CRS of the returned dataarray

seisman:

This PR adds the `crs` parameter to the `load_tile_map` function so that users can change the CRS of the returned raster image without needing to know the details. 

Address https://github.com/GenericMappingTools/pygmt/pull/2125#discussion_r1059868371. 
Closes https://github.com/GenericMappingTools/pygmt/issues/3484.

**Notes for maintainers:**

As far as I know, there are multiple different ways to reproject raster images.

1. [contextily.warp_tiles](https://contextily.readthedocs.io/en/latest/reference.html#contextily.warp_tiles): Internally calls [`rasterio.vrt.WarpedVRT`](https://rasterio.readthedocs.io/en/stable/api/rasterio.vrt.html#rasterio.vrt.WarpedVRT)
2. [rioxarray's rio.reproject](https://corteva.github.io/rioxarray/html/rioxarray.html#rioxarray.raster_array.RasterArray.reproject): Internally calls [`rasterio.warp.reproject`](https://rasterio.readthedocs.io/en/stable/api/rasterio.warp.html#rasterio.warp.reproject)
3. [`rasterio.vrt.WarpedVRT`](https://rasterio.readthedocs.io/en/stable/api/rasterio.vrt.html#rasterio.vrt.WarpedVRT)
4. [`rasterio.warp.reproject`](https://rasterio.readthedocs.io/en/stable/api/rasterio.warp.html#rasterio.warp.reproject)
5. [`gdal.Warp`](https://gdal.org/en/latest/api/python/utilities.html#osgeo.gdal.Warp)
6. ...

Options 3-5 are complicated and I don't think we want to call them directly. Ideally, we should use option 1, so reprojection doesn't rely on an extra dependency `rioxarray`. However, as shown below, the result from `contextily.warp_tiles` doesn't work well with `rio.to_raster`.

Here are codes to compare the results with options 1 and 2:

```python
import xarray as xr
import rioxarray
import contextily as cx
import numpy as np

def image2raster(image, extent, crs="EPSG:3857"):
    """
    Convert an image to a xarray.DataArray raster.

    Copied from pygmt.datasets.tile_map.load_tile_map, except that crs is an input parameter.
    """
    # Turn RGBA img from channel-last to channel-first and get 3-band RGB only
    _image = image.transpose(2, 0, 1)  # Change image from (H, W, C) to (C, H, W)
    rgb_image = _image[0:3, :, :]  # Get just RGB by dropping RGBA's alpha channel
    
    # Georeference RGB image into an xarray.DataArray
    left, right, bottom, top = extent
    dataarray = xr.DataArray(
        data=rgb_image,
        coords={
            "band": np.array(object=[1, 2, 3], dtype=np.uint8),  # Red, Green, Blue
            "y": np.linspace(start=top, stop=bottom, num=rgb_image.shape[1]),
            "x": np.linspace(start=left, stop=right, num=rgb_image.shape[2]),
        },
        dims=("band", "y", "x"),
    )
    dataarray = dataarray.rio.write_crs(input_crs=crs)
    return dataarray

# Image and extent of the global tile
image, extent = cx.bounds2img(w=-180, e=180, s=-90, n=90, ll=True, zoom=0)

# Option 1: Use contextily.warp_tiles
image_warp, extent_warp = cx.warp_tiles(image, extent, "OGC:CRS84")
dataarray_warp = image2raster(image_warp, extent_warp, crs="OGC:CRS84")
print("Using contextily.warp_tiles:", dataarray_warp.rio.bounds())
dataarray_warp.rio.to_raster("result_warp.tiff")

# Option 2: Use rio.reproject
dataarray = image2raster(image, extent, crs="EPSG:3857")
dataarray_reproject = dataarray.rio.reproject("OGC:CRS84")
print("Using rio.reproject:", dataarray_reproject.rio.bounds())
dataarray_reproject.rio.to_raster("result_reproject.tiff")
```
The outputs are:
```
Using contextily.warp_tiles: (-180.55157789333614, -85.9082903958547, 180.18036434850873, 85.66487761291295)
Using rio.reproject: (-180.0, -84.8307625666717, 180.00000000000006, 85.11165075221737)
```
As you can see, the longitude range is (-180.55157789333614, 180.18036434850873) with `contextily.warp_tiles`, which will lead to GMT errors like:
```
grdimage [ERROR]: Map region exceeds 360 degrees
grdimage [ERROR]: General map projection error
```
Two more notes:

- There is an open request to use `rasterio.warp.reproject` instead of `rasterio.vrt.WarpedVRT` in `warp_tiles` (https://github.com/geopandas/contextily/issues/88). If it's done, then the result from `warp_tiles` will be likely the same as the one from `rio.reproject`
- Currently, we need to save raster into temporary files by calling `rio.to_raster` (https://github.com/GenericMappingTools/pygmt/blob/988746bba290d16826ab4040d973af7137c5d3be/pygmt/helpers/tempfile.py#L204). We don't have to do it after https://github.com/GenericMappingTools/pygmt/pull/3468. 

**Preview**: https://pygmt-dev--3554.org.readthedocs.build/en/3554/api/generated/pygmt.datasets.load_tile_map.html

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -156,28 +156,25 @@ def load_tile_map(
         )
         raise ImportError(msg)
 
-    contextily_kwargs = {}
-    if zoom_adjust is not None:
-        contextily_kwargs["zoom_adjust"] = zoom_adjust
-        if Version(contextily.__version__) < Version("1.5.0"):
-            msg = (
-                "The `zoom_adjust` parameter requires `contextily>=1.5.0` to work. "
-                "Please upgrade contextily, or manually set the `zoom` level instead."
-            )
-            raise TypeError(msg)
+    if zoom_adjust is not None and Version(contextily.__version__) < Version("1.5.0"):
+        msg = (
+            "The `zoom_adjust` parameter requires `contextily>=1.5.0` to work. "
+            "Please upgrade contextily, or manually set the `zoom` level instead."
+        )
+        raise ValueError(msg)
 
+    # Keyword arguments for contextily.bounds2img
+    contextily_kwargs = {
+        "zoom": zoom,
+        "source": source,
+        "ll": lonlat,
+        "wait": wait,
+        "max_retries": max_retries,
+        "zoom_adjust": zoom_adjust,
+    }
     west, east, south, north = region
     image, extent = contextily.bounds2img(
-        w=west,
-        s=south,
-        e=east,
-        n=north,
-        zoom=zoom,
-        source=source,
-        ll=lonlat,
-        wait=wait,
-        max_retries=max_retries,
-        **contextily_kwargs,
+        w=west, s=south, e=east, n=north, **contextily_kwargs
     )
 
     # Turn RGBA img from channel-last to channel-first and get 3-band RGB only
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -3,7 +3,6 @@ Function to load raster tile maps from XYZ tile providers, and load as
 :class:`xarray.DataArray`.
 """
 
-import contextlib
 from collections.abc import Sequence
 from typing import Literal
 
@@ -11,16 +10,21 @@ from packaging.version import Version
 
 try:
     import contextily
+    from rasterio.crs import CRS
     from xyzservices import TileProvider
 
     _HAS_CONTEXTILY = True
 except ImportError:
+    CRS = None
     TileProvider = None
     _HAS_CONTEXTILY = False
 
-with contextlib.suppress(ImportError):
-    # rioxarray is needed to register the rio accessor
+try:
     import rioxarray  # noqa: F401
+
+    _HAS_RIOXARRAY = True
+except ImportError:
+    _HAS_RIOXARRAY = False
 
 import numpy as np
 import xarray as xr
@@ -33,6 +37,7 @@ def load_tile_map(
     zoom: int | Literal["auto"] = "auto",
     source: TileProvider | str | None = None,
     lonlat: bool = True,
+    crs: str | CRS = "EPSG:3857",
     wait: int = 0,
     max_retries: int = 2,
     zoom_adjust: int | None = None,
@@ -42,7 +47,8 @@ def load_tile_map(
 
     The tiles that compose the map are merged and georeferenced into an
     :class:`xarray.DataArray` image with 3 bands (RGB). Note that the returned image is
-    in a Spherical Mercator (EPSG:3857) coordinate reference system.
+    in a Spherical Mercator (EPSG:3857) coordinate reference system (CRS) by default,
+    but can be customized using the ``crs`` parameter.
 
     Parameters
     ----------
@@ -80,6 +86,10 @@ def load_tile_map(
     lonlat
         If ``False``, coordinates in ``region`` are assumed to be Spherical Mercator as
         opposed to longitude/latitude.
+    crs
+        Coordinate reference system (CRS) of the returned :class:`xarray.DataArray`
+        image. Default is ``"EPSG:3857"`` (i.e., Spherical Mercator). The CRS can be in
+        either string or :class:`rasterio.crs.CRS` format.
     wait
         If the tile API is rate-limited, the number of seconds to wait between a failed
         request and the next try.
@@ -128,11 +138,21 @@ def load_tile_map(
     ...     raster.rio.crs.to_string()
     'EPSG:3857'
     """
+    _default_crs = "EPSG:3857"  # The default CRS returned from contextily
+
     if not _HAS_CONTEXTILY:
         msg = (
             "Package `contextily` is required to be installed to use this function. "
             "Please use `python -m pip install contextily` or "
             "`mamba install -c conda-forge contextily` to install the package."
+        )
+        raise ImportError(msg)
+
+    if crs != _default_crs and not _HAS_RIOXARRAY:
+        msg = (
+            f"Package `rioxarray` is required if CRS is not '{_default_crs}'. "
+            "Please use `python -m pip install rioxarray` or "
+            "`mamba install -c conda-forge rioxarray` to install the package."
         )
         raise ImportError(msg)
 
@@ -176,8 +196,12 @@ def load_tile_map(
         dims=("band", "y", "x"),
     )
 
-    # If rioxarray is installed, set the coordinate reference system
+    # If rioxarray is installed, set the coordinate reference system.
     if hasattr(dataarray, "rio"):
-        dataarray = dataarray.rio.write_crs(input_crs="EPSG:3857")
+        dataarray = dataarray.rio.write_crs(input_crs=_default_crs)
+
+    # Reproject raster image from EPSG:3857 to specified CRS, using rioxarray.
+    if crs != _default_crs:
+        dataarray = dataarray.rio.reproject(dst_crs=crs)
 
     return dataarray
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -138,9 +138,9 @@ def load_tile_map(
     ...     raster.rio.crs.to_string()
     'EPSG:3857'
     """
-    # The CRS of the returned image. Use the tile provider's CRS if available.
-    # Otherwise, default to EPSG:3857.
-    _default_crs = getattr(source, "crs", "EPSG:3857")
+    # The CRS of the source tile provider. If the source is a TileProvider object, use
+    # its crs attribute if available. Otherwise, default to EPSG:3857.
+    _source_crs = getattr(source, "crs", "EPSG:3857")
 
     if not _HAS_CONTEXTILY:
         msg = (
@@ -150,9 +150,9 @@ def load_tile_map(
         )
         raise ImportError(msg)
 
-    if crs != _default_crs and not _HAS_RIOXARRAY:
+    if crs != _source_crs and not _HAS_RIOXARRAY:
         msg = (
-            f"Package `rioxarray` is required if CRS is not '{_default_crs}'. "
+            f"Package `rioxarray` is required if CRS is not '{_source_crs}'. "
             "Please use `python -m pip install rioxarray` or "
             "`mamba install -c conda-forge rioxarray` to install the package."
         )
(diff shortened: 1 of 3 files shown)


pygmt/src/tilemap.py
```diff
@@ -111,38 +107,21 @@ def tilemap(
...
-    Raises
-    ------
-    ImportError
-        If ``rioxarray`` is not installed. Follow
-        :doc:`install instructions for rioxarray <rioxarray:installation>`, (e.g. via
-        ``python -m pip install rioxarray``) before using this function.
     """
     kwargs = self._preprocess(**kwargs)
 
-    if not _HAS_RIOXARRAY:
```
seisman, line 124: The `Figure.tilemap` method no longer requires `rioxarray` explicitly. Instead, `load_tile_map` will raise the error.

pygmt/src/tilemap.py
```diff
@@ -111,38 +107,21 @@ def tilemap(
...
-            "Please use `python -m pip install rioxarray` or "
-            "`mamba install -c conda-forge rioxarray` to install the package."
-        )
-
     raster = load_tile_map(
         region=region,
         zoom=zoom,
         source=source,
         lonlat=lonlat,
+        crs="OGC:CRS84" if lonlat is True else "EPSG:3857",
```
seisman, line 118: If `lonlat=True`, set the CRS to let `load_tile_map` do the raster reprojection for us.

pygmt/datasets/tile_map.py
```diff
@@ -3,25 +3,29 @@
...
 
-import contextlib
 from collections.abc import Sequence
 from typing import Literal
 
 from packaging.version import Version
 
 try:
     import contextily
+    from rasterio.crs import CRS
```
seisman, line 13: These coverage annotations are incorrect. As shown below, these lines are covered.

https://app.codecov.io/gh/GenericMappingTools/pygmt/blob/load_tile_map%2Fcrs/pygmt%2Fdatasets%2Ftile_map.py
https://app.codecov.io/gh/GenericMappingTools/pygmt/blob/load_tile_map%2Fcrs/pygmt%2Fsrc%2Ftilemap.py

pygmt/datasets/tile_map.py
```diff
@@ -176,8 +196,12 @@ def load_tile_map(
         dims=("band", "y", "x"),
     )
 
-    # If rioxarray is installed, set the coordinate reference system
+    # If rioxarray is installed, set the coordinate reference system.
```
seisman, line 199: These lines are also covered.

michaelgrund (approved):
Looks good to me!

pygmt/datasets/tile_map.py
```diff
@@ -128,6 +138,8 @@ def load_tile_map(
     ...     raster.rio.crs.to_string()
     'EPSG:3857'
     """
+    _default_crs = "EPSG:3857"  # The default CRS returned from contextily
```
weiji14, line 141: At https://contextily.readthedocs.io/en/latest/reference.html#contextily.bounds2img, there is this sentence:

> IMPORTANT: tiles are assumed to be in the Spherical Mercator projection (EPSG:3857), unless the crs keyword is specified.

While tilemaps are usually served in EPSG:3857, I know that there are some tilemaps that can be served in a different projection system by default. I'm hesitant to use this default of EPSG:3857 here, unless we can be confident that contextily only supports EPSG:3857 tiles.

pygmt/datasets/tile_map.py
```diff
@@ -3,25 +3,29 @@
...
 from packaging.version import Version
 
 try:
     import contextily
+    from rasterio.crs import CRS
     from xyzservices import TileProvider
 
     _HAS_CONTEXTILY = True
 except ImportError:
+    CRS = None
```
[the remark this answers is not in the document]
seisman: `rasterio` is a known dependency of `contextily` (xref: https://github.com/geopandas/contextily/blob/main/pyproject.toml), so it makes more sense to assume that they're installed together.

pygmt/datasets/tile_map.py
```diff
@@ -136,28 +148,33 @@ def load_tile_map(
...
+        raise ValueError(msg)
+
+    # Keyword arguments for contextily.bounds2img
+    contextily_kwargs = {
+        "zoom": zoom,
+        "source": source,
+        "ll": lonlat,
+        "wait": wait,
+        "max_retries": max_retries,
+        "zoom_adjust": zoom_adjust,
```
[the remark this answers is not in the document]
seisman: Fixed in 421c167cc

pygmt/datasets/tile_map.py
seisman: Looking at the contexily source codes, I think contexitly always assumes that CRS is EPSG:3857, at least when writing the image to raster files

https://github.com/geopandas/contextily/blob/eafeb3c993ab43356b65e7a29b9dd6df3ac76f00/contextily/tile.py#L358

pygmt/datasets/tile_map.py
```diff
@@ -176,8 +194,12 @@ def load_tile_map(
...
 
-    # If rioxarray is installed, set the coordinate reference system
+    # If rioxarray is installed, set the coordinate reference system.
     if hasattr(dataarray, "rio"):
-        dataarray = dataarray.rio.write_crs(input_crs="EPSG:3857")
+        dataarray = dataarray.rio.write_crs(input_crs=_default_crs)
+
+    # Reproject raster image from EPSG:3857 to specified CRS, using rioxarray.
+    if crs != _default_crs:
+        dataarray = dataarray.rio.reproject(dst_crs=crs)
```
weiji14, line 203: I realize there's an `if crs != _default_crs and not _HAS_RIOXARRAY:` statement at the top, so it's ok to keep this on the outer indent, but I still think it's safer to nest this under the `if hasattr(dataarray, "rio"):` statement.

```suggestion
        dataarray = dataarray.rio.write_crs(input_crs=_default_crs)

        # Reproject raster image from EPSG:3857 to specified CRS, using rioxarray.
        if crs != _default_crs:
            dataarray = dataarray.rio.reproject(dst_crs=crs)
```

pygmt/datasets/tile_map.py
weiji14: That is for the `warp_tiles` function. I don't see anything about EPSG:3857 in the [`bounds2img`](https://github.com/geopandas/contextily/blob/00ca0318033e69e29a164ce571d9c932a057ecc6/contextily/tile.py#L183-L293) function (though it's calling several functions and I haven't checked too closely).

Non-Web Mercator (EPSG:3857) XYZ tiles are not common, but they do exist, e.g. Openlayers allows configuring the projection system of the returned tiles:
- https://www.linz.govt.nz/guidance/data-service/linz-data-service-guide/map-tile-services/using-lds-xyz-services-openlayers#c-0-s-4
- https://openlayers.org/en/main/apidoc/module-ol_source_XYZ-XYZ.html
- https://stackoverflow.com/questions/65024840/xyz-in-non-web-mercator-projections

seisman: > Looking at the contexily source codes, I think contexitly always assumes that CRS is EPSG:3857, at least when writing the image to raster files

I meant this line:

https://github.com/geopandas/contextily/blob/eafeb3c993ab43356b65e7a29b9dd6df3ac76f00/contextily/tile.py#L175

In bounds2raster, it first calls `bounds2img` to get the `image` and `extent` (can be any CRS), but when writing to a raster image, `bounds2raster` assumes the `image` is `EPSG:3857`.

Looking at the xyzservices (https://github.com/geopandas/xyzservices/blob/c847f312d2e67a0a45ceae119fdcdf7cf60858b6/provider_sources/xyzservices-providers.json#L50), I can also see some providers that has the `crs` attribute, but it's unclear how the `crs` attribute is handled in contexitly.

weiji14: I tried using `xyzservices.providers.MapTiler.Basic4326` which comes in EPSG:4326:

```python
import contextily
import xyzservices

source = xyzservices.providers.MapTiler.Basic4326(
    key="BwTZdryxoypHc4BnCZ4"  # add an 'o' to the end
)
img, bounds = contextily.bounds2img(w=0, s=-80, e=150, n=80, ll=True, source=source)
# img.shape # (4096, 2048, 4)
plt.imshow(img[:, :, :3])
```

produces

temp

The extent doesn't seem to be correct (it should be plotting from Greenwich to Asia), possibly a bug in `contextily` hardcoding to EPSG:3857. But it does show that non-Web Mercator XYZ tiles are indeed possible.

weiji14: Found https://github.com/geopandas/contextily/issues/119, and had a closer look at the `bounds2img` code, specifically these lines - https://github.com/geopandas/contextily/blob/00ca0318033e69e29a164ce571d9c932a057ecc6/contextily/tile.py#L270-L272. It seems like `contextily` is using [`mercantile`](https://github.com/mapbox/mercantile) which only supports Spherical Mercator (EPSG:3857), see https://github.com/mapbox/mercantile/issues/109#issuecomment-636331797. Unless `contextily` switches to something like https://github.com/developmentseed/morecantile in the future, then it would only 'work' with EPSG:3857.

seisman: > The extent doesn't seem to be correct (it should be plotting from Greenwich to Asia), possibly a bug in `contextily` hardcoding to EPSG:3857.

https://github.com/geopandas/contextily/blob/eafeb3c993ab43356b65e7a29b9dd6df3ac76f00/contextily/tile.py#L256

When `ll=True`, `bounds2img` calls the private [`_sm2ll`](https://github.com/geopandas/contextily/blob/eafeb3c993ab43356b65e7a29b9dd6df3ac76f00/contextily/tile.py#L539) function to convert bounds in Spherical Mercator coordinates to lon/lat. I think the contextily package assumes "Spherical Mercator projection" in many places.

weiji14: How about checking `source.crs` (an attribute from an instance of the `xyzservices.TileProvider` class), and if it is present, use that as the default crs. If not, fallback to EPSG:3857.

seisman: Sounds good. Done in 6733e196a6ec5b845e235cbe9f44581b94d2c10d.

> `_default_crs = getattr(source, "crs", "EPSG:3857")`

Please note that this line of code works for `TileProvide`, `str` and `None`. `str` and `None` don't have `crs`, so `EPSG:3857` is returned.

weiji14 (approved):

Issue #3547, opened 2024-10-23
Consistent way to specify numpy dtype across the PyGMT codebase


seisman:
There are many different ways to specify the `dtype` argument in NumPy arrays, for example:

1. **Using NumPy types** (e.g., `dtype=np.int32`, `dtype=np.float64`)
2. **Using strings** (e.g., `dtype="int32"`, `dtype="float64"`)
3. **Using Python builtin types** (e.g, `dtype=int`, `dtype=float`)
4. **Using NumPy Array-protocol type strings** (e.g., `dtype='i4'`)
5. maybe more

Currently, we're using both 1 and 2 (i.e., `np.int32` and `"int32"`) across the PyGMT codebase. It would be better if we could consistently use only one way, which can improve code maintainability and also reduce confusion for new contributors/maintainers.

I think "NumPy dtypes" like `np.int32` is better:

1. Makes the code more readable by clearly indicating that the type is a NumPy-specific data type
2. NumPy official documentation uses NumPy types (e.g., `np.float64`) mostly (e.g., https://numpy.org/doc/stable/user/basics.creation.html)
3. Using NumPy types is also slightly faster than "strings"
```
In [1]: import numpy as np

In [2]: %timeit np.array([1, 2, 3], dtype=np.float64)
576 ns ± 1.29 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit np.array([1, 2, 3], dtype="float64")
763 ns ± 5.46 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Another related issue is to standardize the use of NumPy dtype names. Most NumPy dtypes have a canonical name and one or multiple alias names. For example, `np.single`/`np.float32` is the same numpy dtype (xref: https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.single). I think we should use dtypes like `np.float32`, which is more readable, and can ensure consistency across different platforms.

See https://numpy.org/doc/stable/reference/arrays.scalars.html for the upstream docuem

PR #3565, opened 2024-10-31
Replace the list of support numpy dtypes with DTYPES_NUMERIC in tests

seisman:
The long list of supported numpy types are used in many places. Currently, we usually use
```
"int8 int16 int32 int64 uint8 uint16 uint32 uint64 float32 float64".split()
```
it can be replaced by:
```
[dtype for dtype in DTYPES_NUMERIC if dtype != np.timedelta64]
```

This is also a small step towards #3547.

### DIFF
--- a/pygmt/tests/test_clib_virtualfiles.py
+++ b/pygmt/tests/test_clib_virtualfiles.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import numpy as np
 import pytest
 from pygmt import clib
+from pygmt.clib.session import DTYPES_NUMERIC
 from pygmt.exceptions import GMTCLibError, GMTInvalidInput
 from pygmt.helpers import GMTTempFile
 from pygmt.tests.test_clib import mock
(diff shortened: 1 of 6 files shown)


pygmt/tests/test_clib_virtualfiles.py
```diff
@@ -28,20 +28,7 @@ def fixture_dtypes():
     """
     List of supported numpy dtypes.
     """
-    return "int8 int16 int32 int64 uint8 uint16 uint32 uint64 float32 float64".split()
-
-
-@pytest.fixture(scope="module", name="dtypes_pandas")
```
seisman, line 34: This fixture is no longer used in this file.

pygmt/tests/test_grdimage_image.py
```diff
@@ -43,7 +45,7 @@ def test_grdimage_image_dataarray(xr_image):
 
 @pytest.mark.parametrize(
     "dtype",
-    ["int8", "uint16", "int16", "uint32", "int32", "float32", "float64"],
+    [dtype for dtype in DTYPES_NUMERIC if dtype not in {np.uint8, np.timedelta64}],
```
seisman, line 48: Currently, only `uint8` images are supported.

PR #3579, opened 2024-11-05
Figure.savefig: Clarify that 'transparent' also works for the PNG file associated with the KML format

seisman:
**Changes**

1. Clarify that `transparent` works for `.kml` format and add a test for it
2. Refactor the codes to make it more compact and readable.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -135,7 +135,7 @@ class Figure:
             wesn = lib.extract_region()
         return wesn
 
-    def savefig(  # noqa: PLR0912
+    def savefig(
         self,
         fname: str | PurePath,
         transparent: bool = False,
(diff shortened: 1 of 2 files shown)


pygmt/figure.py
```diff
@@ -226,14 +226,12 @@ def savefig(  # noqa: PLR0912
...
             case ext if ext not in fmts:
-                raise GMTInvalidInput(f"Unknown extension '.{ext}'.")
-
-        fmt = fmts[ext]
-        if transparent:
-            if fmt != "g":
-                msg = f"Transparency unavailable for '{ext}', only for png."
+                msg = f"Unknown extension '.{ext}'."
                 raise GMTInvalidInput(msg)
-            fmt = fmt.upper()
```
seisman, line 236: Changes to something like `"png": "G" if transparent is True else "g"` at line #210.

pygmt/figure.py
```diff
@@ -226,14 +226,12 @@ def savefig(  # noqa: PLR0912
...
+
+        if transparent and ext not in {"kml", "png"}:
```
seisman, line 233: Instead of checking `fmt != "g"`, here we explicitly check if `ext not in {"kml", "png"}`

michaelgrund (approved):

PR #3506, opened 2024-10-11
Add instructions for installing optional dependencies

seisman:
**Description of proposed changes**

**Preview**: https://pygmt-dev--3506.org.readthedocs.build/en/3506/install.html#dependencies

Closes https://github.com/GenericMappingTools/pygmt/issues/3371.

PR #3592, opened 2024-11-06
Add PyArrow as an optional dependency

seisman:
We're working towards supporting more PyArrow data types in #2800. I think it makes sense to explicitly add it as an optional dependency, just like what we're doing to contexily, geopandas, ipython and rioxarray. 

FYI, we already mentioned `pyarrow` as an optional dependency on [the PyGMT Ecosystem page](https://www.pygmt.org/dev/ecosystem.html) and recommended to install it in #3506.

This PR supersedes some changes in PR #2933.

### DIFF
--- a/.github/workflows/type_checks.yml
+++ b/.github/workflows/type_checks.yml
@@ -50,7 +50,7 @@ jobs:
           # 4. other packages that are used somewhere in PyGMT
           python -m pip install \
             numpy pandas xarray netcdf4 packaging \
-            contextily geopandas ipython rioxarray \
+            contextily geopandas ipython pyarrow rioxarray \
             mypy pandas-stubs \
             matplotlib pytest
           python -m pip list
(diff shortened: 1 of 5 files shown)


yvonnefroehlich (commented):
Probably we also need to update the installation guide.

*Edit*: Sorry, I read incorrectly. In the dev version it's already included in the installation guide https://github.com/GenericMappingTools/pygmt/blob/4498f71e29b7997033a481de72fb2b2d74aa5bac/doc/install.md?plain=1#L177

yvonnefroehlich (approved):

.github/workflows/type_checks.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           # 4. other packages that are used somewhere in PyGMT
           python -m pip install \
             numpy pandas xarray netcdf4 packaging \
-            contextily geopandas ipython rioxarray \
+            contextily geopandas ipython pyarrow rioxarray \
             mypy pandas-stubs \
```
weiji14, line 54: Add [pyarrow-stubs](https://github.com/zen-xu/pyarrow-stubs) as well?
```suggestion
            mypy pandas-stubs pyarrow-stubs \
```

weiji14 (approved):

Issue #3600, opened 2024-11-08
Supported data types by the private `_to_numpy` function


seisman:
Here is a central issue to track the supported data types for the private `_to_numpy` function.

### Numeric Dtypes

- Python built-in types: int, float, complex #3583
- NumPy: #3583
  - int8, int16, int32, int64, longlong
  - uint8, uint16, uint32, uint64, ulonglong
  - float16, float32, float64, longdouble
  - complex64, complex128, clongdouble

PR #3599, opened 2024-11-08
clib.conversion._to_numpy: Add tests for pyarrow.array with pyarrow numeric dtypes

seisman:
**Description of proposed changes**

This PR adds tests for `_to_numpy` to check if it works for PyArrow arrays with pyarrow numeric dtypes.

PyArrow provides 11 numeric dtypes (https://arrow.apache.org/docs/python/api/datatypes.html): int8, int16, int32, int64, uint8, uint16, uint32, uint64, float16, float32, and float64. All can be converted to numpy dtype without issues.

Here is a minimal example to play with pyarrow types.
```python
>>> import numpy as np
>>> import pyarrow as pa
>>> x = pa.array([1, 2, 3], type=pa.int16())
>>> x
<pyarrow.lib.Int16Array object at 0x7fc28a73fb80>
[
  1,
  2,
  3
]
>>> np.ascontiguousarray(x)
array([1, 2, 3], dtype=int16)

>>> x = pa.array([1, 2, None], type=pa.int16())
>>> x
<pyarrow.lib.Int16Array object at 0x7fc28a7ba620>
[
  1,
  2,
  null
]

>>> np.ascontiguousarray(x)
array([ 1.,  2., nan])
```

Please note that `pa.float16` needs special handling:
```python
>>> x = pa.array([1.0, 2.0, 3.0], type=pa.float16())
---------------------------------------------------------------------------
ArrowTypeError                            Traceback (most recent call last)
Cell In[9], line 1
----> 1 x = pa.array([1.0, 2.0, 3.0], type=pa.float16())

File ~/opt/miniforge/envs/pygmt/lib/python3.12/site-packages/pyarrow/array.pxi:370, in pyarrow.lib.array()

File ~/opt/miniforge/envs/pygmt/lib/python3.12/site-packages/pyarrow/array.pxi:42, in pyarrow.lib._sequence_to_array()

File ~/opt/miniforge/envs/pygmt/lib/python3.12/site-packages/pyarrow/error.pxi:155, in pyarrow.lib.pyarrow_internal_check_status()

File ~/opt/miniforge/envs/pygmt/lib/python3.12/site-packages/pyarrow/error.pxi:92, in pyarrow.lib.check_status()

ArrowTypeError: Expected np.float16 instance

>>> # The solution is from https://arrow.apache.org/docs/python/generated/pyarrow.float16.html#pyarrow.float16
>>> x = pa.array(np.array([1.0, 2.0, 3.0], dtype=np.float16), type=pa.float16())
[...]

PR #3584, opened 2024-11-05
clib.converison._to_numpy: Add tests for pandas.Series with pandas numeric dtypes

seisman:
This PR adds tests for pandas.Series with pandas/pyarrow numeric dtypes (in 7222db23bbdfbb298b211c02ab981e71ab433be6).

Tests pass with pandas 2.2 but fail with pandas 2.1. Check https://github.com/GenericMappingTools/pygmt/actions/runs/11714111582/job/32628172796?pr=3584 for details.

The failures are due to upstream pandas v2.2 changes. Previously, all pandas nullable dtypes are converted to np.object_ dtype. Since pandas v2.2, the new rules are (source: https://pandas.pydata.org/docs/whatsnew/v2.2.0.html#to-numpy-for-numpy-nullable-and-arrow-types-converts-to-suitable-numpy-dtype):

> - float dtypes are cast to NumPy floats
> - integer dtypes without missing values are cast to NumPy integer dtypes
> - integer dtypes with missing values are cast to NumPy float dtypes and NaN is used as missing value indicator
> - boolean dtypes without missing values are cast to NumPy bool dtype
> - boolean dtypes with missing values keep object dtype
> - datetime and timedelta types are cast to Numpy datetime64 and timedelta64 types respectively and NaT is used as missing value indicator

Following the first three points, we add the workaround for pandas<2.2 to explicitly specify how to map pandas dtypes into numpy dtypes (a4f15cd778b146cc787450ba81b36da81a3a2efd). It is an improved version of PR #3505.

Here is the minimal example to understand the behavior change:

**With pandas 2.1**
```python
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: x = pd.Series([1, 2, 3], dtype=pd.Int32Dtype())

In [4]: x.dtype
Out[4]: Int32Dtype()

In [5]: np.ascontiguousarray(x)
Out[5]: array([1, 2, 3], dtype=object)

In [6]: x = pd.Series([1, pd.NA, 3], dtype=pd.Int32Dtype())

In [7]: np.ascontiguousarray(x)
Out[7]: array([1, <NA>, 3], dtype=object)
```
**With pandas 2.2**
```python
In [1]: import numpy as np

In [2]: import pandas as pd

In [3]: x = pd.Series([1, 2, 3], dtype=pd.Int32Dtype())

In [4]: np.ascontiguousarray(x)
Out[4]: array([1, 2, 3], dtype=int32)

[...]

PR #3601, opened 2024-11-08
clib.conversion._to_numpy: Add tests for Python list of strings and NumPy array with string type

seisman:
Test Python's list of strings and NumPy's string arrays.

Related to #3600.

### DIFF
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -41,11 +41,12 @@ def _check_result(result, expected_dtype):
             np.complex128,
             id="complex",
         ),
+        pytest.param(["abc", "defh", "12345"], np.str_, id="string"),
     ],
 )
-def test_to_numpy_python_types_numeric(data, expected_dtype):
+def test_to_numpy_python_types(data, expected_dtype):
     """
-    Test the _to_numpy function with Python built-in numeric types.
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_clib_to_numpy.py
```diff
@@ -41,11 +41,12 @@ def _check_result(result, expected_dtype):
             np.complex128,
             id="complex",
         ),
+        pytest.param(["abc", "defh", "12345"], np.str_, id="string"),
```
seisman, line 44:
```suggestion
        pytest.param(["abc", "defg", "12345"], np.str_, id="string"),
```

seisman:
It's unclear what's happening with the Python 3.10 + macOS CI job. It started to fail since PR #3599, but changes in PR #3599 are totally unrelated since pyarrow is not installed in the Python 3.10 + macOS CI job and the failing tests are all about grdcontour and grdimage. What's more weird is that the same CI job then passes in https://github.com/GenericMappingTools/pygmt/pull/3584. 

In PR https://github.com/GenericMappingTools/pygmt/pull/3606, I've bumped our minimum supported Python version to 3.11 and the CI jobs pass, so I think we can ignore the Python 3.10 failures.

michaelgrund (approved):

weiji14 (approved):

PR #3603, opened 2024-11-08
Figure.plot/Figure.plot3d: Improve the check of the "style" parameter for "v" or "V"

seisman:
In GMT, when both symbol and size are read from files, the `-S` option can be specified without arguments. Thus, `style=""` or `style=True` should be valid.

However, currently we check `kwargs.get("S") is not None and kwargs["S"][0] in "vV"` which doesn't make sense if `style` is not a string or is an empty string. 

This PR fixes the checking of the `style` parameter.

Strictly speaking, it's a bug, but I don't want to add one more test for such a minor change. So mark it as "maintenance" and "skip-changelog".

### DIFF
--- a/pygmt/src/plot.py
+++ b/pygmt/src/plot.py
@@ -213,7 +213,8 @@ def plot(
     if kind == "empty":  # Add more columns for vectors input
         # Parameters for vector styles
         if (
-            kwargs.get("S") is not None
+            isinstance(kwargs.get("S"), str)
+            and len(kwargs["S"]) >= 1
             and kwargs["S"][0] in "vV"
             and is_nonstr_iter(direction)
         ):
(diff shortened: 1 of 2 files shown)


lattaai4:
Hello, I tried to fix the issue

This is what I did:

This change modifies the checking of the `style` parameter in both `plot` and `plot3d` functions to allow for empty strings and boolean values. It ensures that the `-S` option can be specified without arguments when both symbol and size are read from files.

I also created pull request: https://github.com/GenericMappingTools/pygmt/pull/3604


> [!CAUTION]
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!
>
> The fix provided by Latta AI might not be complete and it can serve as an inspiration.

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt fixing issues on your repository, you can block this account.

weiji14 (approved):
Probably should still mark it as a bug so it appears in the changelog. We can skip adding a regression test.

PR #3621, opened 2024-11-15
clib.conversion._to_numpy: Add tests for PyArrow's timestamp type

seisman:
[`pyarrow.timestamp`](https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html) is the datetime dtype in pyarrow. It has two parameters, `unit` and `tz`. 

This PR adds tests to ensure `_to_numpy` support `pyarrow.timestamp`. Please note that, `unit` can only be `s`/`ms`/`us`/`ns`, and ~~`tz` information is discarded when converting to np.datetime64.~~ Actually, when converting to np.datetime64, timezone is corrected to UTC.

```
In [1]: import pyarrow as pa

In [2]: import datetime

In [3]: import numpy as np

In [4]: x = pa.array([datetime.datetime(2024, 1, 1, 12, 34, 56)], type=pa.timestamp("s"))

In [5]: np.ascontiguousarray(x)
Out[5]: array(['2024-01-01T12:34:56'], dtype='datetime64[s]')

In [6]: x = pa.array([datetime.datetime(2024, 1, 1, 12, 34, 56)], type=pa.timestamp("ms"))

In [7]: np.ascontiguousarray(x)
Out[7]: array(['2024-01-01T12:34:56.000'], dtype='datetime64[ms]')

In [8]: x = pa.array([datetime.datetime(2024, 1, 1, 12, 34, 56)], type=pa.timestamp("ms", tz="UTZ"))

In [9]: np.ascontiguousarray(x)
Out[9]: array(['2024-01-01T12:34:56.000'], dtype='datetime64[ms]')
```

Related to #3600.

### DIFF
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -17,8 +17,10 @@ try:
     import pyarrow as pa
 
     _HAS_PYARROW = True
+    pa_timestamp = pa.timestamp
 except ImportError:
     _HAS_PYARROW = False
+    pa_timestamp = None
 
 
 def _check_result(result, expected_dtype):
@@ -235,6 +237,7 @@ def test_to_numpy_pandas_series_pyarrow_dtypes_date(dtype, expected_dtype):
 # - Date types:
 #   - date32[day]
 #   - date64[ms]
+# - Datetime types: timestamp
 #
 # In PyArrow, array types can be specified in two ways:
 #
@@ -357,3 +360,43 @@ def test_to_numpy_pyarrow_array_pyarrow_dtypes_date(dtype, expected_dtype):
         result,
         np.array(["2024-01-01", "2024-01-02", "2024-01-03"], dtype=expected_dtype),
     )
+
+
+@pytest.mark.skipif(not _HAS_PYARROW, reason="pyarrow is not installed")
+@pytest.mark.parametrize(
+    ("dtype", "expected_dtype"),
+    [
+        pytest.param(None, "datetime64[us]", id="None"),
+        pytest.param("timestamp[s]", "datetime64[s]", id="timestamp[s]"),
+        pytest.param("timestamp[ms]", "datetime64[ms]", id="timestamp[ms]"),
+        pytest.param("timestamp[us]", "datetime64[us]", id="timestamp[us]"),
+        pytest.param("timestamp[ns]", "datetime64[ns]", id="timestamp[ns]"),
+        pytest.param(
+            pa_timestamp("s", tz="UTC"), "datetime64[s]", id="timestamp[s, tz=UTC]"
+        ),  # pa.timestamp with tz has no string alias.
+        pytest.param(
+            pa_timestamp("s", tz="America/New_York"),
+            "datetime64[s]",
+            id="timestamp[s, tz=America/New_York]",
+        ),
+        pytest.param(
+            pa_timestamp("s", tz="+07:30"),
+            "datetime64[s]",
+            id="timestamp[s, tz=+07:30]",
+        ),
+    ],
+)
+def test_to_numpy_pyarrow_timestamp(dtype, expected_dtype):
+    """
+    Test the _to_numpy function with PyArrow arrays of PyArrow datetime types.
+
+    pyarrow.timestamp(unit, tz=None) can accept units "s", "ms", "us", and "ns".
+
+    Reference: https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html
+    """
+    data = [datetime(2024, 1, 2, 3, 4, 5), datetime(2024, 1, 2, 3, 4, 6)]
+    array = pa.array(data, type=dtype)
+    result = _to_numpy(array)
+    _check_result(result, np.datetime64)
+    assert result.dtype == expected_dtype
+    npt.assert_array_equal(result, array)
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -25,7 +25,7 @@ except ImportError:
         """
 
         @staticmethod
-        def timestamp(*args, **kwargs):
+        def timestamp(unit: str, tz: str | None = None):
             """
             A dummy function to mimic pyarrow.timestamp.
             """
@@ -407,7 +407,7 @@ def test_to_numpy_pyarrow_array_pyarrow_dtypes_date(dtype, expected_dtype):
 )
 def test_to_numpy_pyarrow_timestamp(dtype, expected_dtype):
     """
-    Test the _to_numpy function with PyArrow arrays of PyArrow datetime types.
+    Test the _to_numpy function with PyArrow arrays of PyArrow timestamp types.
 
     pyarrow.timestamp(unit, tz=None) can accept units "s", "ms", "us", and "ns".
 
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -418,4 +418,5 @@ def test_to_numpy_pyarrow_timestamp(dtype, expected_dtype):
     result = _to_numpy(array)
     _check_result(result, np.datetime64)
     assert result.dtype == expected_dtype
-    npt.assert_array_equal(result, array)
+    assert result[0] == np.datetime64("2024-01-02T03:04:05")
+    assert result[1] == np.datetime64("2024-01-02T03:04:06")
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -256,7 +256,7 @@ def test_to_numpy_pandas_series_pyarrow_dtypes_date(dtype, expected_dtype):
 # - Date types:
 #   - date32[day]
 #   - date64[ms]
-# - Datetime types: timestamp
+# - Timestamp types: timestamp[unit], timestamp[unit, tz]
 #
 # In PyArrow, array types can be specified in two ways:
 #
--- a/pygmt/tests/test_clib_to_numpy.py
+++ b/pygmt/tests/test_clib_to_numpy.py
@@ -17,10 +17,20 @@ try:
     import pyarrow as pa
 
     _HAS_PYARROW = True
-    pa_timestamp = pa.timestamp
 except ImportError:
+
+    class pa:  # noqa: N801
+        """
+        A dummy class to mimic pyarrow.
+        """
+
+        @staticmethod
+        def timestamp(*args, **kwargs):
+            """
+            A dummy function to mimic pyarrow.timestamp.
+            """
+
     _HAS_PYARROW = False
-    pa_timestamp = None
 
 
 def _check_result(result, expected_dtype):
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_clib_to_numpy.py
```diff
@@ -18,6 +18,18 @@
 
     _HAS_PYARROW = True
 except ImportError:
+
+    class pa:  # noqa: N801
```
seisman, line 22: `pa.timestamp()` with `tz` doesn't have string aliases, so we can't use strings like `timestamp[s, UTC]`. So we have to define a dummy function here.

pygmt/tests/test_clib_to_numpy.py
```diff
@@ -366,3 +379,43 @@ def test_to_numpy_pyarrow_array_pyarrow_dtypes_date(dtype, expected_dtype):
...
+def test_to_numpy_pyarrow_timestamp(dtype, expected_dtype):
+    """
+    Test the _to_numpy function with PyArrow arrays of PyArrow datetime types.
+
+    pyarrow.timestamp(unit, tz=None) can accept units "s", "ms", "us", and "ns".
+
+    Reference: https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html
+    """
+    data = [datetime(2024, 1, 2, 3, 4, 5), datetime(2024, 1, 2, 3, 4, 6)]
+    array = pa.array(data, type=dtype)
```
weiji14, line 417: Is the timezone part of the `pa.timestamp` type actually working?

```python
data = [datetime(2024, 1, 2, 3, 4, 5), datetime(2024, 1, 2, 3, 4, 6)]
array = pa.array(data, type=pa.timestamp("s", tz="America/New_York"))
print(array)
# [
#   2024-01-02 03:04:05Z,
#   2024-01-02 03:04:06Z
# ]
result = _to_numpy(array)
print(result)
# ['2024-01-02T03:04:05' '2024-01-02T03:04:06']
```

If using `pd.Timestamp` instead

```
import pandas as pd
import pyarrow as pa
from datetime import datetime, timezone
from zoneinfo import ZoneInfo
from pygmt.clib.conversion import _to_numpy

array = pa.array(
    [
        pd.Timestamp(2024, 1, 2, 3, 4, 5, tzinfo=ZoneInfo("America/New_York")),
        pd.Timestamp(2024, 1, 2, 3, 4, 6, tzinfo=ZoneInfo("America/New_York")),
    ]
)
print(array)
# [
#   2024-01-02 08:04:05.000000Z,
#   2024-01-02 08:04:06.000000Z
# ]
result = _to_numpy(array)
print(result)
# ['2024-01-02T08:04:05.000000' '2024-01-02T08:04:06.000000']
```

For the first one, the timezone offset for New York (UTC-5) doesn't seem to be applied? Whereas for the second one, it is converting from New York timezone (UTC-5) to UTC by adding 5 hours.

seisman: It seems the timezone information is applied but not shown when representing the array.

```python
In [1]: from datetime import datetime

In [2]: import pyarrow as pa

# For pa.scalar, it's shown with timezone offset
In [3]: pa.scalar(datetime(2024, 1, 2, 3, 4, 5), type=pa.timestamp("s", tz="America/New_York"))
Out[3]: <pyarrow.TimestampScalar: '2024-01-01T22:04:05-0500'>

In [4]: array = pa.array([datetime(2024, 1, 2, 3, 4, 5)], type=pa.timestamp("s", tz="America/New_York"))
# For pa.array, it's shown without timezone offset
In [5]: array
Out[5]:
<pyarrow.lib.TimestampArray object at 0x7f52feace740>
[
  2024-01-02 03:04:05Z
]

# When converted to pandas, it's shown with timezone offset
In [6]: array.to_pandas()
Out[6]:
0   2024-01-01 22:04:05-05:00
dtype: datetime64[s, America/New_York]

# When converted to numpy, timezone information is lost, since numpy doesn's have tz support.
In [7]: array.to_numpy()
Out[7]: array(['2024-01-02T03:04:05'], dtype='datetime64[s]')
```

seisman: Likely related to https://github.com/apache/arrow/issues/40122.

seisman: > Is the timezone part of the `pa.timestamp` type actually working?
> 
> ```python
> data = [datetime(2024, 1, 2, 3, 4, 5), datetime(2024, 1, 2, 3, 4, 6)]
> array = pa.array(data, type=pa.timestamp("s", tz="America/New_York"))
> print(array)
> # [
> #   2024-01-02 03:04:05Z,
> #   2024-01-02 03:04:06Z
> # ]
> result = _to_numpy(array)
> print(result)
> # ['2024-01-02T03:04:05' '2024-01-02T03:04:06']
> ```
> 
> If using `pd.Timestamp` instead
> 
> ```
> import pandas as pd
> import pyarrow as pa
> from datetime import datetime, timezone
> from zoneinfo import ZoneInfo
> from pygmt.clib.conversion import _to_numpy
> 
> array = pa.array(
>     [
>         pd.Timestamp(2024, 1, 2, 3, 4, 5, tzinfo=ZoneInfo("America/New_York")),
>         pd.Timestamp(2024, 1, 2, 3, 4, 6, tzinfo=ZoneInfo("America/New_York")),
>     ]
> )
> print(array)
> # [
> #   2024-01-02 08:04:05.000000Z,
> #   2024-01-02 08:04:06.000000Z
> # ]
> result = _to_numpy(array)
> print(result)
> # ['2024-01-02T08:04:05.000000' '2024-01-02T08:04:06.000000']
> ```
> 
> For the first one, the timezone offset for New York (UTC-5) doesn't seem to be applied? Whereas for the second one, it is converting from New York timezone (UTC-5) to UTC by adding 5 hours.


Also need to note the difference between `datetime` and `pd.Timestamp`. `datetime` is timezone-unware, while `pd.Timestamp` has timezone support.

So, in your first example:

- actual datetime: `2024-01-02T03:04:05+00:00`
- In pyarrow.array, the actual datetime is `2024-01-01T22:04:05-05:00`, but UTC time (`2024-01-02T03:04:05`) is shown 
- to_numpy: `2024-01-02T03:04:05` (UTC time)

In your 2nd example:

- the actual datetime: `2024-01-02T03:04:05-05:00`
- In pyarrow.array: the UTC time (`2024-01-02 08:04:05`) is shown
- to_numpy: `2024-01-02T08:04:05`

seisman: https://stackoverflow.com/a/73276431

> Arrow internally stores datetime as UTC + timezone info and will print it as such

weiji14: Ok, I see now that there is indeed consistency, thanks for clearing this up, the repr on pyarrow.array is indeed confusing :sweat_smile: . So in summary:

| class/type | has timezone support | Link |
|--|--|--|
| Python `datetime` | :white_check_mark: | https://docs.python.org/3/library/datetime.html#datetime.datetime.tzinfo |
| pandas.Timestamp | :white_check_mark: | https://pandas.pydata.org/pandas-docs/version/2.2/reference/api/pandas.Timestamp.html |
| pyarrow.timestamp (type) | :white_check_mark: | https://arrow.apache.org/docs/17.0/python/generated/pyarrow.timestamp.html|
| numpy.datetime64 | :x: | https://numpy.org/doc/2.1/reference/arrays.scalars.html#numpy.datetime64 |

Since we're converting to NumPy, the timezone information will always be lost (everything is converted to UTC).

seisman: As shown below, GMT itself doesn't have timezone support, similar to numpy.datetime64. The big difference is that GMT simply ignore the TZ infor, while the to_numpy conversion always converts to UTC.
```
gmt begin map
gmt basemap -R2024-01-01T00:00:00+08:00/2024-01-02T00:00:00+08:00/1/10 -JX10c/4c -Baf 
echo 2024-01-01T08:00:00+08:00 5 | gmt plot -Sc1c -Gblack
gmt end show
```
map

> Does it mean that users cannot plot data in a specific timezone, because it will always convert to UTC?

I think yes, unless we try hard to drop the TZ information before converting to numpy.datetime64. Below is an example, but if we want to support TZ, we need to find a more general way to do it.

```python
In [1]: import pyarrow as pa

In [2]: from datetime import datetime

In [3]: data = [datetime(2024, 1, 2, 3, 4, 5), datetime(2024, 1, 2, 3, 4, 6)]

In [4]: array = pa.array(data, type=pa.timestamp("s", tz="America/New_York"))

In [5]: array
Out[5]:
<pyarrow.lib.TimestampArray object at 0x7f78c07bafe0>
[
  2024-01-02 03:04:05Z,
  2024-01-02 03:04:06Z
]

In [6]: array.to_pandas().dt.tz_localize(None)
Out[6]:
0   2024-01-01 22:04:05
1   2024-01-01 22:04:06
dtype: datetime64[s]

In [7]: import numpy as np

In [8]: np.ascontiguousarray(array.to_pandas().dt.tz_localize(None))
Out[8]:
array(['2024-01-01T22:04:05', '2024-01-01T22:04:06'],
      dtype='datetime64[s]')
```

pygmt/tests/test_clib_to_numpy.py
seisman: We can, but I feel it will make the pytest parametrize more complicated.

pygmt/tests/test_clib_to_numpy.py
weiji14: Maybe we need to open an issue to discuss this - whether to:

1. Follow GMT (allow users to plot data at a non-UTC timezone, by ignoring the timezone offset)
2. Follow NumPy, whereby the data will from any non-UTC timezone will be converted to UTC always.

If going with 2, we should at least raise a warning if a non-UTC timezone is used, that a conversion is taking place. If going with 1, we would need to special-case datetime types, that might mean extra logic in the `_to_numpy` function, or having to keep `array_to_datetime` (so don't merge #3507 yet).

seisman: I'm also debating whether to following option 1 or 2, so a separate issue sounds good. 

Since the current behavior is option 2 (as done in `array_to_datetime`), I think we should go with option 2 first, then revisit the timezone support later, which will be a breaking change anyway.

weiji14: Ok, I'll open an issue to discuss. And sure, we can go with option 2 for now.

pygmt/tests/test_clib_to_numpy.py
```diff
@@ -18,6 +18,18 @@
...
     _HAS_PYARROW = True
 except ImportError:
+
+    class pa:  # noqa: N801
+        """
+        A dummy class to mimic pyarrow.
+        """
+
+        @staticmethod
+        def timestamp(*args, **kwargs):
```
weiji14, line 28: Follow the signature of `pyarrow.timestamp`.
```suggestion
        def timestamp(unit: str, tz: str | None = None):
```

pygmt/tests/test_clib_to_numpy.py
weiji14: Ok, let's keep this dummy class then.

pygmt/tests/test_clib_to_numpy.py
```diff
@@ -366,3 +379,44 @@ def test_to_numpy_pyarrow_array_pyarrow_dtypes_date(dtype, expected_dtype):
...
+        pytest.param(
+            pa.timestamp("s", tz="+07:30"),
+            "datetime64[s]",
+            id="timestamp[s, tz=+07:30]",
+        ),
+    ],
+)
+def test_to_numpy_pyarrow_timestamp(dtype, expected_dtype):
+    """
+    Test the _to_numpy function with PyArrow arrays of PyArrow datetime types.
```
weiji14, line 410:
```suggestion
    Test the _to_numpy function with PyArrow arrays of PyArrow timestamp types.
```

pygmt/tests/test_clib_to_numpy.py
```diff
@@ -244,6 +256,7 @@ def test_to_numpy_pandas_series_pyarrow_dtypes_date(dtype, expected_dtype):
 # - Date types:
 #   - date32[day]
 #   - date64[ms]
+# - Datetime types: timestamp
```
weiji14, line 259: Would this be strange?
```suggestion
# - Timestamp types:
#   - timestamp[s]
#   - timestamp[ms]
#   - timestamp[us]
#   - timestamp[ns]
```

seisman: I prefer to something like this:
```suggestion
# - Timestamp types: timestamp[unit] or timestamp[unit, tz]
```

PR #3639, opened 2024-11-21
WIP: Test lowest versions of all required and optional dependencies

weiji14:
**Description of proposed changes**

Test lowest versions of all direct dependencies of PyGMT listed in the pyproject.toml file (both required and optional dependencies).

This implements the idea at https://github.com/GenericMappingTools/pygmt/issues/3615#issuecomment-2472660976 on `.github/workflows/ci_tests_legacy.yaml` (instead of `ci_tests.yaml`) by using the `uv` package manager's `--resolution lowest-direct` strategy (https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy).

Related to optional dependency policy at https://github.com/GenericMappingTools/pygmt/issues/3456, and complements #3615 by testing lowest versions of transitive dependencies more thoroughly.

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3643, opened 2024-11-22
Remove the dependency of netCDF4

seisman:

Inspired by https://github.com/GenericMappingTools/pygmt/pull/3639#discussion_r1853413593, now it's possible to remove the netCDF4 package from PyGMT's dependency list.

**TODO**:

- [x] #3115
- [x] #3919 
- [x] #3922


pygmt/helpers/testing.py
```diff
@@ -152,8 +151,13 @@ def load_static_earth_relief():
...
         A grid of Earth relief for internal tests.
     """
-    fname = which("@static_earth_relief.nc", download="c")
-    return load_dataarray(fname)
+    with Session() as lib:
+        with lib.virtualfile_out(kind="grid") as voutgrd:
+            lib.call_module(
+                module="read", args=["@static_earth_relief.nc", voutgrd, "-Tg"]
+            )
+            grid = lib.virtualfile_to_raster(vfname=voutgrd, kind="grid")
```
weiji14, line 159: You know what? Maybe we could think of refactoring `pygmt.io.load_dataarray` to use this with block? It would be a breaking change, but that could mean we can use GMT's I/O directly.

seisman: Yes, actually, I'm thinking if we should provide a wrapper for the special `read` module, which can read a table, grid, or image, and return pd.DataFrame or xarray.DataArray. The function looks like:
```
def read(file, kind=None, **kwargs):
    if kind == None:
        # Try our best to decide the data kind. Or make 'kind' a required parameter
        pass 

    code = {"dataset": "d", "grid": "g", "image": "i"}[kind]
    func = 

    with Session() as lib:
        with lib.virtualfile_out(kind=kind) as voutfile:
            lib.call_module(
                module="read", args=[file, voutfile, f"-T{code}"]
            )
            if kind == "dataset":
                return lib.virtualfile_to_dataset(vfname=voutfile, **kwargs)
            elif kind in {"grid", "image"}
                return lib.virtualfile_to_raster(vfname=voutgrd, kind=kind)
```
For comparison, GMT.jl provides the [`gmtread`](https://www.generic-mapping-tools.org/GMTjl_doc/documentation/utilities/gmtread/index.html)/[`gmtwrite`](https://www.generic-mapping-tools.org/GMTjl_doc/documentation/utilities/gmtwrite/index.html#gmtwrite) functions, but As I understand it, the functions return the GMT_DATASET/GMT_GRID/GMT_IMAGE containers, rather than data structures in the Julia world.

weiji14:
Not sure if removing a required dependency counts as a breaking change (i.e. if we should should we put `**Breaking**` in the title), the main non-test-related user-facing change was to `_load_earth_relief_holes` in `pygmt/dataset/samples.py` in #3922 which is quite minor. I suppose we could just label this under 'deprecation' instead of 'maintenance' to make it a bit more prominent in the changelog?

weiji14 (commented):
There's a test failing on `ubuntu latest - Python 3.12`. Need to add a `@pytest.mark.skipif(condition=not _HAS_NETCDF4, reason="netCDF4 is not installed")` marker for this line:

https://github.com/GenericMappingTools/pygmt/blob/ae82fd646df42eb6a14d6124b44a803de73e2ba5/pygmt/tests/test_xarray_backend.py#L27

because `.to_netcdf()` still relies on the netCDF4 engine.

weiji14 (approved):
Nice, glad to have one less dependency to handle! Remember to remove `netCDF4` from the conda-forge feedstock for the next release :fire:

PR #3673, opened 2024-12-04
Add pygmt.read to read a dataset/grid/image into pandas.DataFrame/xarray.DataArray

seisman:
This PR adds the `pygmt.read` function to read any recognized data files (dataset, grid, or image) into a pandas.DataFrame/xarray.DataArray object.


seisman:
In the `_load_remote_dataset` function, we can't replace the following codes with the new `read` function, because in `read`, we call `which` to get the full path of the source grid, which doesn't work well for tiled grids.

https://github.com/GenericMappingTools/pygmt/blob/7768e93750524e8ed63266cd24372de21a5d9f42/pygmt/datasets/load_remote_dataset.py#L444-L459

seisman:
Now, the `load_dataarray` function is used in `pygmt/src/grdcut.py` only (related to https://github.com/GenericMappingTools/pygmt/pull/3115).

pygmt/src/read.py
```diff
@@ -0,0 +1,118 @@
...
+    kind
+        The kind of data to read. Valid values are ``"dataset"``, ``"grid"``, and
+        ``"image"``.
+    region
+        The region of interest. Only data within this region will be read.
+    column_names
+        A list of column names.
+    header
+        Row number containing column names. ``header=None`` means not to parse the
+        column names from table header. Ignored if the row number is larger than the
```
michaelgrund, line 47:
```suggestion
        column names from the table header. Ignored if the row number is larger than the
```

pygmt/src/read.py
```diff
@@ -0,0 +1,118 @@
...
+        number of headers in the table.
```
michaelgrund, line 48:
```suggestion
        number of header lines in the table.
```

michaelgrund (approved):

pygmt/src/read.py
```diff
@@ -0,0 +1,118 @@
...
+            case "dataset":
+                return lib.virtualfile_to_dataset(
+                    vfname=voutfile,
+                    column_names=column_names,
+                    header=header,
+                    dtype=dtype,
+                    index_col=index_col,
+                )
+            case "grid" | "image":
+                raster = lib.virtualfile_to_raster(vfname=voutfile, kind=kind)
```
weiji14, line 111: Debating on whether we should have a low-level clib `read` that reads into a GMT virtualfile, and a high-level `read` that converts the virtualfile to a `pandas.DataFrame` or `xarray.DataArray`.

doc/api/index.rst
```diff
@@ -172,6 +172,7 @@ Input/output
     :toctree: generated
 
     load_dataarray
+    read
```
weiji14, line 175: The `load_dataarray` function was put under the `pygmt.io` namespace. Should we consider putting `read` under `pygmt.io` too? (Thinking about whether we need a low-level `pygmt.clib.read` and high-level `pygmt.io.read` in my other comment).

pygmt/src/read.py
seisman: We can have a low-level `Session.read` method, but as far as I can see, currently it will be used only in the high-level `read` function, so it seems unecessary. We can refactor and add the low-level `Session.read` method when we need it in the future.

doc/api/index.rst
seisman: Yes, that sounds good. I have two questions:

1. Should we place the `read` source code in `pygmt/io.py`, or restructure `io.py` into a directory and put it in `pygmt/io/read.py` instead?  
2. Should we deprecate the `load_dataarray` function in favor of the new `read` function?

weiji14: > 1. Should we place the `read` source code in `pygmt/io.py`, or restructure `io.py` into a directory and put it in `pygmt/io/read.py` instead?

I think making the `io` directory sounds good, especially if you're planning on making a `write` function in the future.

> Should we deprecate the `load_dataarray` function in favor of the new `read` function?

No, let's keep `load_dataarray` for now. Something I'm contemplating is to make an [xarray BackendEntrypoint](https://docs.xarray.dev/en/v2025.03.1/generated/xarray.backends.BackendEntrypoint.html) that uses GMT `read`, so that users can then do `pygmt.io.load_dataarray(..., engine="gmtread")` or something like that. The `load_dataarray` function would use this new `gmtread` backend engine by default instead of `netcdf4`.

pygmt/src/read.py
weiji14: Yep, ok with just making a high-level `read` method for now.

pygmt/tests/test_io_gmtread.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.skipif(not _HAS_NETCDF4, reason="netCDF4 is not installed.")
+def test_io_gmtread_grid():
+    """
+    Test that reading a grid returns an xr.DataArray and the grid is the same as the one
+    loaded via xarray.load_dataarray.
+    """
+    grid = gmtread("@static_earth_relief.nc", kind="grid")
+    assert isinstance(grid, xr.DataArray)
+    expected_grid = xr.load_dataarray(which("@static_earth_relief.nc", download="a"))
+    assert np.allclose(grid, expected_grid)
```
weiji14, line 24: Also should have a similar test for `kind="image"`, comparing against `rioxarray.open_rasterio`?

pygmt/io/gmtread.py
```diff
@@ -0,0 +1,120 @@
...
+        A list of column names.
+    header
+        Row number containing column names. ``header=None`` means not to parse the
+        column names from table header. Ignored if the row number is larger than the
+        number of headers in the table.
+    dtype
+        Data type. Can be a single type for all columns or a dictionary mapping
+        column names to types.
+    index_col
+        Column to set as index.
```
weiji14, line 53: Should we indicate in the docstring that these params are only used for `kind="dataset"`?

pygmt/io/gmtread.py
```diff
@@ -0,0 +1,120 @@
...
+
+def gmtread(
+    file: str | PurePath,
+    kind: Literal["dataset", "grid", "image"],
+    region: Sequence[float] | str | None = None,
+    header: int | None = None,
+    column_names: pd.Index | None = None,
+    dtype: type | Mapping[Any, type] | None = None,
+    index_col: str | int | None = None,
+) -> pd.DataFrame | xr.DataArray:
```
weiji14, line 24: On second thought, I'm thinking if we should make `gmtread` a private function for internal use only for now, the fact that it can read either tabular or grid/image files seems like a lot of magic.

pygmt/io/gmtread.py
```diff
@@ -0,0 +1,120 @@
...
+import pandas as pd
+import xarray as xr
+from pygmt.clib import Session
+from pygmt.helpers import build_arg_list, is_nonstr_iter
+from pygmt.src.which import which
+
+
+def gmtread(
+    file: str | PurePath,
+    kind: Literal["dataset", "grid", "image"],
```
weiji14, line 18: Does GMT read also handle 'cube'?

seisman: Yes (xref: https://github.com/GenericMappingTools/gmt/blob/9a8769f905c2b55cf62ed57cd0c21e40c00b3560/src/gmtread.c#L75-L81), but need to wait for #3150, which may have upstream bugs.

pygmt/tests/test_io_gmtread.py
seisman: Done in a6c4ee70e87af2a9ecaeab5e8c352cbab819896c.

When I tried to add a test for reading datasets, I realized that the DataFrame returned by the `load_sample_data` is not ideal:
```
In [1]: from pygmt.datasets import load_sample_data

In [2]: data = load_sample_data("hotspots")

In [3]: data.dtypes
Out[3]: 
longitude      float64
latitude       float64
symbol_size    float64
place_name      object
dtype: object
```
The last column `place_name` should be string dtype, rather than `object`. We also have similar issues for other sample datasets. 

We have three options:

1. Do nothing and keep them unchanged
2. Fix and use appropriate dtypes
3. Use the new `gmtread` function instead of `pd.read_csv` in `_load_xxx` functions.

I'm inclined to option 3.

pygmt/io/gmtread.py
seisman: At line 31:

> For datasets, keyword arguments ``column_names``, ``header``, ``dtype``, and
> ``index_col`` are supported.

pygmt/io/gmtread.py
seisman: It seems the `gmtread` function is no longer needed if PR #3919 is implemented, right?

weiji14: Yes, not needed for grids/images, but we could still use `gmtread` for tabular datasets? Though let's think about https://github.com/GenericMappingTools/pygmt/pull/3673#discussion_r2048080261.

pygmt/tests/test_io_gmtread.py
weiji14: > 3\. Use the new `gmtread` function instead of `pd.read_csv` in `_load_xxx` functions.
> 
> I'm inclined to option 3.

Agree with this. We should also add dtype related checks for the tabular dataset tests in `pygmt/tests/test_datasets_samples.py`.

PR #3490, opened 2024-10-07
Add support for Python 3.13

weiji14:
**Description of proposed changes**

Python 3.13 has been released on 7 Oct 2024, changelog is at https://docs.python.org/3.13/whatsnew/3.13.html

Previous PR for Python 3.12 at https://github.com/GenericMappingTools/pygmt/pull/2711

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #2468, opened 2023-03-30
Temporarily pin funcy<2.0 for dvc fix

maxrjones:
**Description of proposed changes**

This PR applies the suggestion from https://github.com/GenericMappingTools/pygmt/pull/2464#issuecomment-1487613519 to pin funcy in order to fix the dvc issue in the test workflows.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3695, opened 2024-12-18
CI: Install dvc via uv in the "GMT Tests" workflow

seisman:
dvc is a complex package that has ~40 direct dependencies (https://github.com/iterative/dvc/blob/007041bfe953db08b832ddea2d3adeaeb74eec95/pyproject.toml#L33). The complicated dvc dependency tree sometimes causes failures or version conflicts  e.g., https://github.com/GenericMappingTools/pygmt/pull/2468, https://github.com/GenericMappingTools/pygmt/pull/2338, https://github.com/GenericMappingTools/pygmt/pull/1544.

In https://github.com/GenericMappingTools/pygmt/pull/3490, we're trying to add the Python 3.13 support to PyGMT. That PR is stale because some dvc dependencies don't have 3.13 support yet (https://github.com/GenericMappingTools/pygmt/pull/3490#discussion_r1889665320), and we may have to wait for a few more days or months.

Since dvc is a dev dependency, we don't have to wait for the conda-forge updates. This PR tries to install dvc in other ways. I've tried the following options:

1. Use https://github.com/iterative/setup-dvc
2. Use `pip`
3. Use `pip` on Linux/macOS and `choco`/`winget` on Windows
4. Use `uv`

It turns out that installing dvc on Linux/macOS only takes ~10 seconds, but on Windows, options 1-3 take 60 seconds, except for option 4, which takes ~10 seconds on all platforms.

The CI job runs (https://github.com/GenericMappingTools/pygmt/actions/runs/12389452771/job/34582514750?pr=3695) for commit 83bb2d73436447b03b92586fd8380a5da663a592 show that, with changes in this PR, we can continue the Python 3.13 support in PR #3490.

### DIFF
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -133,7 +133,6 @@ jobs:
             xarray${{ matrix.xarray-version }}
             netCDF4
             packaging
-            dvc
             make
             pip
             python-build
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -153,8 +153,13 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
-      - name: Setup data version control (DVC)
+      - name: Setup data version control (DVC) [Unix]
         run: python -m pip install dvc
+        if: runner.os != 'Windows'
+
+      - name: Setup data version control (DVC) [Windows]
+        run: choco install dvc
+        if: runner.os == 'Windows'
 
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -153,13 +153,8 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
 
-      - name: Setup data version control (DVC) [Unix]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yaml
```diff
@@ -154,9 +153,21 @@ jobs:
...
+          source .venv/bin/activate
+          uv pip install dvc
+          uv pip list
+
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
-        run: dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
+        run: |
+          source .venv/bin/activate
+          uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
```
seisman, line 170: Maybe we can combine the "Install dvc" and "Pull baseline image data from dvc remote" steps into one step "Setup DVC and pull baseline image data from dvc remote", so we don't have to `source .venv/bin/activate` twice.

weiji14: Let's just keep this as is, since I'm assuming this will be a temporary workaround until conda-forge's `dvc` supports Python 3.13. We might want to revert the changes here after a few months, so best to have two separate steps.

weiji14 (approved):
Wish that `setup-dvc` was faster on Windows, but ok to go with this here.

Issue #3476, opened 2024-10-01
Typesetting backticks ` and straight single quote '


seisman:
[Backtick](https://en.wikipedia.org/wiki/Backtick) (or grave accent) <code>\`</code> and [straight single quote](https://en.wikipedia.org/wiki/Quotation_mark) `'` are ASCII characters and can be typed using a normal keyboard.

If you compare the [Adobe ISOLatin1+ encoding](https://en.wikipedia.org/wiki/PostScript_Latin_1_Encoding), and the [ISO 8859-1 encoding](https://en.wikipedia.org/wiki/ISO/IEC_8859-1), you will see that, 

- code 027 is `'` in ISO 8859-1, and is `’` in Adobe ISOLatin1+
- code 060 is <code>\`</code> in ISO 8859-1, and is `‘` in Adobe ISOLatin1+

There must be some historical or typographic reasons behind the differences. Due to the above differences, typesetting <code>\`'</code> in PostScript results in `‘’`, which means users have to use octal codes if they need the <code>\`'</code> characters.

Here is a minimal example to reproduce the issue:

```python
import pygmt

label = "A ` ' \" “ ” ‘ ’ B"
fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 5], frame=True, projection="x1c")
fig.text(x=5, y=2, text=label, font="50p")
fig.show()
```
The script produces the following image:

![quotes](https://github.com/user-attachments/assets/fff6f74c-142b-4c75-b86e-ade3151691ca)

It's possible to patch the ISOLatin1+ encoding mapping table so that <code>\`'</code> are displayed like <code>\`'</code>, rather than `‘’`. But it's a breaking change. So, do we want to break it or not?

PR #3700, opened 2024-12-19
Adjust titles in polar projections overview

michaelgrund:

Adjust title for each plot so that the formatting looks like the string given to the projection parameter (double quotations are shown in the plot titles instead of single ones).

**Old:**

plot_old

***

**New:**

plot_new

**Reminders**

**Slash Commands**

### DIFF
--- a/examples/projections/nongeo/polar.py
+++ b/examples/projections/nongeo/polar.py
@@ -61,7 +61,7 @@ fig.basemap(
     # Set map width to 5 cm
     projection="P5c",
     # Set the frame and title; @^ allows for a line break within the title
-    frame=["xa45f", "+gbisque+tprojection='P5c' @^ region=[0, 360, 0, 1]"],
+    frame=["xa45f", '+gbisque+tprojection="P5c" @^ region=[0, 360, 0, 1]'],
 )
 
 fig.shift_origin(xshift="w+3c")
@@ -75,7 +75,7 @@ fig.basemap(
     # standard angle
     projection="P5c+a",
     # Set the frame and title; @^ allows for a line break within the title
-    frame=["xa45f", "+gbisque+tprojection='P5c+a' @^ region=[0, 360, 0, 1]"],
+    frame=["xa45f", '+gbisque+tprojection="P5c+a" @^ region=[0, 360, 0, 1]'],
 )
 
 fig.shift_origin(xshift="w+3c")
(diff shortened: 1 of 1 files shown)


seisman (approved):
Looks good.

examples/projections/nongeo/polar.py
```diff
@@ -61,7 +61,7 @@
     # Set map width to 5 cm
     projection="P5c",
     # Set the frame and title; @^ allows for a line break within the title
-    frame=["xa45f", "+gbisque+tprojection='P5c' @^ region=[0, 360, 0, 1]"],
+    frame=["xa45f", '+gbisque+tprojection="P5c" @^ region=[0, 360, 0, 1]'],
```
yvonnefroehlich, line 64: Hmm. I am a bit debating on what is more important, (i) using consistently double quotation marks for string arguments or (ii) making the title consistent with the argument. Currently, I am a bit more on the side of (i). A way to use both times double quotation marks is to add a leading `\` for the inner double quotation marks. But maybe this makes things only more complicated.

```suggestion
    frame=["xa45f", "+gbisque+tprojection=\"P5c\" @^ region=[0, 360, 0, 1]"],
```

seisman: Backslashes make the codes more difficult to read and should be avoided. 

Personally, I'm OK with displaying either single or double quotes in the plot title. Currently, single quotes are displayed like `’`, rather than `'`. See https://github.com/GenericMappingTools/pygmt/issues/3476 for the reason and #3105 for the fix.

michaelgrund: > Currently, single quotes are displayed like `’`, rather than `'`. See #3476 for the reason and #3105 for the fix.

That's the reason why I wanted to change it. The single quotes are looking like accute accents right now and that's something that seems to be weird for users when they see it in plots.

michaelgrund: I would still go with using double quotations here since the titles directly reference the `projection` parameter in the code. Using single quotations for a string in such cases is fine in my opinion. But I'm also fine if all others what to keep it as is.

seisman: I'm inclined to keep it as is.

yvonnefroehlich: Would also prefer to keep it as is, as I prefer to have (if possible) double quotation marks for the string inputs.

seisman:
I'm closing the PR.

PR #3711, opened 2024-12-24
Update the gallery example for plotting lines with LineString/MultiLineString geometry

michaelgrund:
**Description of proposed changes**

Fixes  #3320 

**Preview**: https://pygmt-dev--3711.org.readthedocs.build/en/3711/gallery/lines/linestrings.html#

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3750, opened 2025-01-07
ruff: Ignore the B018 rule (useless expressions) in examples

seisman:
Our examples are written as Python scripts but are rendered like Jupyter Notebooks. 

In Notebooks, the expression in the last notebook cell will be evaluated and printed as the cell's output, which is commonly used, but ruff complains that the expressions are unused ([B018](https://docs.astral.sh/ruff/rules/useless-expression/#useless-expression-b018)), thus we have to use `print` instead.

In this PR, the rule `B018` is ignored in examples.

**Previews**:

- https://pygmt-dev--3750.org.readthedocs.build/en/3750/gallery/images/rgb_image.html
- https://pygmt-dev--3750.org.readthedocs.build/en/3750/gallery/lines/linestrings.html
- https://pygmt-dev--3750.org.readthedocs.build/en/3750/gallery/maps/choropleth_map.html
- https://pygmt-dev--3750.org.readthedocs.build/en/3750/tutorials/basics/plot.html

### DIFF
--- a/examples/gallery/lines/linestrings.py
+++ b/examples/gallery/lines/linestrings.py
@@ -25,7 +25,7 @@ gdf = gpd.read_file(geodatasets.get_path("eea large_rivers"))
 
 # Convert object to EPSG 4326 coordinate system
 gdf = gdf.to_crs("EPSG:4326")
-print(gdf.head())
+gdf.head()
 
 # %%
 fig = pygmt.Figure()
--- a/examples/gallery/maps/choropleth_map.py
+++ b/examples/gallery/maps/choropleth_map.py
@@ -6,7 +6,7 @@ The :meth:`pygmt.Figure.plot` method allows us to plot geographical data such as
 polygons which are stored in a :class:`geopandas.GeoDataFrame` object. Use
 :func:`geopandas.read_file` to load data from any supported OGR format such as a
 shapefile (.shp), GeoJSON (.geojson), geopackage (.gpkg), etc. You can also use a full
-URL pointing to your desired data source. Then, pass the class:`geopandas.GeoDataFrame`
+URL pointing to your desired data source. Then, pass the :class:`geopandas.GeoDataFrame`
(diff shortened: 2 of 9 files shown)


examples/gallery/maps/choropleth_map.py
```diff
@@ -20,7 +20,7 @@
 
 # Read the example dataset provided by geodatasets.
 gdf = gpd.read_file(geodatasets.get_path("geoda airbnb"))
-print(gdf)
+gdf
```
seisman, line 23:
```suggestion
gdf.head()
```

examples/gallery/lines/linestrings.py
```diff
@@ -25,7 +25,7 @@
 
 # Convert object to EPSG 4326 coordinate system
 gdf = gdf.to_crs("EPSG:4326")
-print(gdf.head())
+gdf.head()
```
yvonnefroehlich, line 28: Hm. My background off using `print` was that if running the complete script as "normal" Python script no output is shown, only when running this specific section separately (see comment https://github.com/GenericMappingTools/pygmt/pull/3711#discussion_r1899736562).

examples/gallery/maps/choropleth_map.py
yvonnefroehlich: Maybe keep the `.head()` to show only the first rows, but use it with `print`. Then the output is adjusted and the wide table fits on the page.

michaelgrund (approved):

examples/gallery/lines/linestrings.py
seisman: Yes, output the data head information makes little sense when running the example as a normal Python script.

examples/gallery/maps/choropleth_map.py
seisman: Good suggestion. Done in 7e2cace77290ac59393d8f8c56142efcacf7a472.

Issue #3556, opened 2024-10-27
Consistent bullet point lists in docstrings 


yvonnefroehlich:
Let's take the documentation of `Figure.plot` (https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html#pygmt.Figure.plot) as an example. Currently, we use different ways to write bullet point lists in the docstrings:

- **modifier**: docstring  (e.g., `connection`)
- **modifier** - docstring  (e.g., `gap`)
- **modifier** docstring  (e.g., `header`)

I feel it would be good to decide for one way and then update all docstrings accordingly.

seisman:
Yes, I agree. Personally, I prefer using `:`.

yvonnefroehlich:
> Yes, I agree. Personally, I prefer using `:`.

@GenericMappingTools/pygmt-maintainers are there any other preferences? Otherwise we will go with `:`.

weiji14:
Yep, ok with using `:`.

PR #3752, opened 2025-01-07
DOC: Make bullet point lists consistent in docstrings (":" and lower-case) Part II

seisman:
Most changes are done automatically by the following `sed` command:
```
sed -i 's/- \*\*\(.\)\*\* - \(.*\)/- **\1**: \2/' **/*.py
```

Address https://github.com/GenericMappingTools/pygmt/issues/3556.

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -37,17 +37,17 @@ COMMON_DOCSTRINGS = {
             (using ``binary="o"``), where *ncols* is the number of data columns
             of *type*, which must be one of:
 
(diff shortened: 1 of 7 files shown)


seisman:
The option modifiers are one of the barriers towards a Pythonic interface, and the final goal is to avoid using any modifiers in the PyGMT arguments. In other words, most docstrings need rewritten to make PyGMT pythonic. Perhaps we should not make future changes to these docstrings now, but keep "colons" in mind when we rewrite the docstrings.

michaelgrund (approved):

yvonnefroehlich (approved):

PR #3756, opened 2025-01-08
DOC: Make usage of "lowercase" and "uppercase" consistent across all docs

michaelgrund:

In the docs we use lowercase/uppercase, lower-case/upper-case and lower case/upper case. This PR makes all appearances consistent to use lowercase/uppercase.

**Reminders**

**Slash Commands**


yvonnefroehlich (approved):
I personally do not really like the hyphen version and the internet says with white space is British and without American, so I agree with @michaelgrund to use the version without white space. Happy that this is now consistent.

Issue #1606, opened 2021-11-02
Sphinx LaTeX PDF documentation bundle


weiji14:
**Description of the desired feature**

I have a friend who will be working remotely in the field (Antarctica) without internet access for a few weeks, and he asked me about a PDF copy of the documentation. Upstream GMT has dropped PDF builds in GMT v6.0 as mentioned in https://forum.generic-mapping-tools.org/t/gmt-6-install-complete-documentation-fails/251/2 and https://github.com/GenericMappingTools/gmt/issues/805, but I think we could have it for PyGMT through sphinx's LaTeX PDF build, similar to what other Python packages do on readthedocs.

A few suggestions:

1. Add a `make latexpdf` option to https://github.com/GenericMappingTools/pygmt/blob/v0.5.0/doc/Makefile so that people can build the PDF locally for any version.
2. Build an official PDF only for each release, and have it as a downloadable release asset somewhere:
    - On the sidebar at https://www.pygmt.org
    - On https://github.com/GenericMappingTools/pygmt/releases
    - On Zenodo maybe?

Here's a PDF I managed to build locally for v0.5.0: [pygmt_v0.5.0_docs.pdf](https://github.com/GenericMappingTools/pygmt/files/7464524/pygmt_v0.5.0_docs.pdf). It contains 465 pages and is ~13MB, including all the gallery examples, tutorials and API docs. The commands I ran locally were:

```bash
cd ~/path/to/pygmt/doc
make all  # build the API docs
PYGMT_USE_EXTERNAL_DISPLAY="false" sphinx-build -b latex -d _build/doctrees -j auto . _build/latex
cd _build/latex/
make  # build the LaTeX PDF docs
```

There were a few errors related to missing LaTeX fonts (e.g. I had to do `sudo apt install latex-gyre`), unicode characters (all the emojis in the changelog :sweat_smile:) and SVG images (the DOI badges are SVG) which I had to skip through. It should be possible to work out those issues though with a bit of effort.

References:
- https://www.sphinx-doc.org/en/master/usage/quickstart.html#running-the-build
- https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.latex.LaTeXBuilder
- https://github.com/sphinx-doc/sphinx/blob/9e1b4a8f1678e26670d34765e74edf3a3be3c62c/sphinx/templates/quickstart/Makefile_t#L55-L60
[...]

PR #3761, opened 2025-01-09
Simplify doc/Makefile with sphinx-build make-mode

seisman:
[`sphinx-build`](https://www.sphinx-doc.org/en/master/man/sphinx-build.html) started to provide the so-called [make-mode](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-M) since v1.2.0, which has been [the default mode since v1.5.0](https://www.sphinx-doc.org/en/master/man/sphinx-quickstart.html#cmdoption-sphinx-quickstart-use-make-mode).

Here is a simple comparison for the make-mode and the "normal" mode:

Old normal mode:
```
$(SPHINXBUILD) -b html $(SPHINXOPTS) $(SOURCEDIR) $(BUILDDIR)/html
```
make mode:
```
$(SPHINXBUILD) -M html "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
```

These two modes are very similar, but in make-mode, we use `$(BUILDDIR)` not `$(BUILDDIR)/html`.

A good advantage is that we can use the same command for different Sphinx builders.

This PR updates `doc/Makefile` with the make-mode and also adds the `latex` target to generate LaTeX source files.

Address #1606.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -1,13 +1,11 @@
 # Makefile for Sphinx documentation
 
 # You can set these variables from the command line.
-SPHINXOPTS    = -j auto
-SPHINXBUILD   = sphinx-build
+SPHINXOPTS    ?= -j auto
+SPHINXBUILD   ?= sphinx-build
 SPHINXAUTOGEN = sphinx-autogen
+SOURCEDIR     = .
 BUILDDIR      = _build
-
-# Internal variables.
-ALLSPHINXOPTS = -d $(BUILDDIR)/doctrees  $(SPHINXOPTS) .
 
 .PHONY: help all api html server clean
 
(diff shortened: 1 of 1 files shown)


doc/Makefile
```diff
@@ -28,20 +26,20 @@ api:
...
 
-html: api
+html latex latexpdf: api
 	@echo
-	@echo "Building HTML files."
+	@echo "Running $(SPHINXBUILD) with the '$@' builder"
 	@echo
 	# Set PYGMT_USE_EXTERNAL_DISPLAY to "false" to disable external display
-	PYGMT_USE_EXTERNAL_DISPLAY="false" $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	PYGMT_USE_EXTERNAL_DISPLAY="false" $(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
```
seisman, line 34: `$@` is the target, i.e., `html` when running `make html`.

doc/Makefile
```diff
@@ -28,20 +26,20 @@ api:
 	@echo
 	$(SPHINXAUTOGEN) -i -t _templates -o api/generated api/*.rst
 
-html: api
+html latex latexpdf: api
```
seisman, line 29: `latex` and `latexpdf` are added here to support PDF documentation (xref #1606).

doc/Makefile
```diff
@@ -28,20 +26,20 @@ api:
 	@echo
 	$(SPHINXAUTOGEN) -i -t _templates -o api/generated api/*.rst
 
-html: api
+html latex: api
```
seisman, line 29: `latex` is added here to support PDF documentation (xref #1606).

michaelgrund (approved):

PR #3742, opened 2025-01-03
CI: Separate jobs for publishing to TestPyPI and PyPI

weiji14:
**Description of proposed changes**

Have a dedicated build distribution job, and split the publish to TestPyPI and PyPI jobs, to workaround attestation file issue. Xref https://github.com/pypa/gh-action-pypi-publish/issues/283

References:
- https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#separate-workflow-for-publishing-to-testpypi

Fixes #3736

**Preview**:

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3767, opened 2025-01-12
CI: Separate jobs to build and deploy documentation

seisman:
Similar to #3742, it makes more sense to build and deploy documentation in separate jobs.

This PR adds split the deployment part into a separate job.


seisman:
One disadvantage is that, the deployment job starts only after all the three matrix jobs in the `build-docs` job finish (there are no workarounds xref: https://github.com/orgs/community/discussions/42335). 

Considering that the Windows job takes much longer (~ 7 minutes) than the Linux/macOS jobs (~ 3 minutes), maybe having separate jobs is not a good idea?

seisman:
Let's keep it in a single job now. Will revisit this PR later.

Issue #3194, opened 2024-04-23
x2sys_cross tests fail on macOS-14


seisman:
https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/

`macos-latest` has migrated from macOS-12 to macOS-14 in a few hours ago. All tests are good except `x2sys_cross` tests.

- Latest run with macOS-12: https://github.com/GenericMappingTools/pygmt/actions/runs/8792433777/job/24128565756
- Latest run with macOS-14: https://github.com/GenericMappingTools/pygmt/actions/runs/8792433777/job/24153784055

The failures are:
```
=================================== FAILURES ===================================
_________________ test_x2sys_cross_input_file_output_dataframe _________________

    @pytest.mark.usefixtures("mock_x2sys_home")
    @pytest.mark.xfail(
        condition=Version(__gmt_version__) < Version("6.5.0"),
        reason="Upstream bug fixed in https://github.com/GenericMappingTools/gmt/pull/8188",
    )
    def test_x2sys_cross_input_file_output_dataframe():
        """
        Run x2sys_cross by passing in a filename, and output internal crossovers to a
        pandas.DataFrame.
        """
        with TemporaryDirectory(prefix="X2SYS", dir=Path.cwd()) as tmpdir:
            tag = Path(tmpdir).name
            x2sys_init(tag=tag, fmtfile="xyz", force=True)
            output = x2sys_cross(tracks=["@tut_ship.xyz"], tag=tag, coe="i")
    
            assert isinstance(output, pd.DataFrame)
>           assert output.shape == (14338, 12)
E           assert (14374, 12) == (14338, 12)
E             
E             At index 0 diff: 14374 != 14338
E             
E             Full diff:
E               (
E             -     14338,
E             ?        ^^
E             +     14374,
E             ?        ^^
E                   12,
E               )

../pygmt/tests/test_x2sys_cross.py:76: AssertionError
----------------------------- Captured stderr call -----------------------------
Warning: ss [WARNING]: No time column, use dummy times
_______________ test_x2sys_cross_region_interpolation_numpoints ________________

    @pytest.mark.usefixtures("mock_x2sys_home")
    @pytest.mark.xfail(
        condition=Version(__gmt_version__) < Version("6.5.0"),
        reason="Upstream bug fixed in https://github.com/GenericMappingTools/gmt/pull/8188",
    )
[...]

PR #3778, opened 2025-01-18
CI: Test on Linux arm64 runners

seisman:
xref:
- https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
- https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


weiji14:
Test failures on some `x2sys_cross` tests appear to be the same as https://github.com/GenericMappingTools/pygmt/issues/3194, we had some logic to handle the shape mismatches using an `if sys.platform == "darwin" check, e.g. here:

https://github.com/GenericMappingTools/pygmt/blob/8a09f75cae034ab0e01e1a29e46df65a116d9508/pygmt/tests/test_x2sys_cross.py#L273-277

Maybe we can use `if platform.machine() == "arm64"` to handle both Ubuntu and macOS arm64 variants, xref https://docs.python.org/3/library/platform.html#platform.machine?

seisman:
> Maybe we can use `if platform.machine() == "arm64"` to handle both Ubuntu and macOS arm64 variants, xref https://docs.python.org/3/library/platform.html#platform.machine?

`platform.machine()` returns `arm64` on macOS ARM and `aarch64` on Linux ARM. So, `if platform.machine() in {"aarch64", "arm64"}:` works, but there are still two failures related to `contour`.

seisman:
> > Maybe we can use `if platform.machine() == "arm64"` to handle both Ubuntu and macOS arm64 variants, xref https://docs.python.org/3/library/platform.html#platform.machine?
> 
> `platform.machine()` returns `arm64` on macOS ARM and `aarch64` on Linux ARM. So, `if platform.machine() in {"aarch64", "arm64"}:` works, but there are still two failures related to `contour`.

I've marked the two 'contour' tests as xfail on Linux ARM and now all tests pass.

weiji14 (approved):

Issue #3676, opened 2024-12-08
Add gallery example for plotting earthquakes on the profile view


maxrjones:
### Description of the desired feature

This was another request from the AGU workshop, which would make a great gallery example. The request was to plot the earthquakes within 100 km on either side of the line in the profile view.

### Are you willing to help implement and maintain this feature?

Maybe

PR #3526, opened 2024-10-16
Figure.meca: Add aliases for "-Fa", "-Fe", "-Fg", "-Ft", "-Fp"

yvonnefroehlich:
**Description of proposed changes**

This PR adds aliases for the GMT flags used for plotting the P and T axes for `Figure.meca`:

| GMT flag | PyGMT alias |
| --- | --- |
| **Fa** | ``pt_axes`` |
| **Fe** | ``t_axisfill`` |
| **Fg** | ``p_axisfill`` |
| **Ft** | ``t_axispen`` |
| **Fp** | ``p_axispen`` |

The docstrings are orientated on the upstream GMT documentation at https://docs.generic-mapping-tools.org/dev/supplements/seis/meca.html#f.

**Preview**:

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3817, opened 2025-02-21
Wrap coupe

### LINKED ISSUE #2019: pscoupe implementation

jhtong33:

Adding `coupe` module to plot the plot cross-sections of focal mechanisms, following https://github.com/GenericMappingTools/pygmt/issues/2019#issuecomment-1873612153 and #3551.
Test examples followed the existing scripts (description in test_coupe.py)

Some things I want to confirm:
- [ ] Making function `section_convention_code` into _FocalMechanismConvention
- [ ] PT axes Pythonic parameter #3526 
- [ ] The necessary for `plot_longitude` and `plot_latitude`

Fixes #2019  and #3676 

**Preview**:

**Reminders**

**Slash Commands**


pygmt/src/coupe.py
```diff
@@ -0,0 +1,379 @@
+"""
+coupe - Plot cross-sections of focal mechanisms.
+"""
+
+import numpy as np
+import pandas as pd
+from pygmt.clib import Session
+from pygmt.exceptions import GMTError, GMTInvalidInput
```
michaelgrund, line 8:
```suggestion
from pygmt.exceptions import GMTInvalidInput
```
Not used in your code.

michaelgrund (commented):
The whole code contains a lot of trailing white spaces. Please remove all of them.

pygmt/src/coupe.py
```diff
@@ -0,0 +1,383 @@
...
+    latitude=None,
+    depth=None,
+    section_format='lonlat_lonlat',
+    plot_longitude=None,
+    plot_latitude=None,
+    event_name=None,
+    **kwargs
+):
+    r"""
+    Plot focal mechanisms in a cross section.
```
seisman, line 64: Most of the docstrings are the same as the one in `Figure.meca`. I'd prefer to remove all the docstrings for the following reasons:

1. To avoid maintaining duplicated docstrings, maybe we can just add a link to `Figure.meca` and let users read `Figre.meca` docstrings first. If we decide to do this, we can remove these docstrings permanently.
2. If we still want the full docstrings, we can still remove these docstrings temporarily, so that we can focus on the codes, rather than the docstrings in the initial reviews.

michaelgrund: Would suggest to go with option 1. Another possibility would be to create a single docstring file and use it in both methods. That would avoid maintaining duplicated docstrings.

yvonnefroehlich: Agree with @michaelgrund! Maybe we can add these parameters to the decorators file?

yvonnefroehlich (commented):
A general aspect: Sofar, this PR is coming from @jhtong33's fork. 
As she is member of the PyGMT contributor team I believe she can create branches directly from the PyGMT repo. I feel this could make reviewing / testing this PR easier 🙂.

pygmt/src/coupe.py
jhtong33: > Agree with @michaelgrund! Maybe we can add these parameters to the decorators file?

Also agree. Do I need to help with this part, or can someone else add these parameters to the decorators file (open new PR)?

seisman: > Maybe we can add these parameters to the decorators file?

I don't think this is a good option, mainly because: (1) these are not real common docstrings. They're just shared by two or a few functions; (2) it will make the `COMMON_DOCSTRING` dictionary too big to maintain; (3) if we keep adding more "non-common" docstrings into `COMMON_DOCSTRINGS` in the future, we will likely have name conflicts and may make mistakes in using the placeholders.

michaelgrund: Agree with @seisman regarding the size of the `COMMON_DOCSTRING` dictionary. That's why I initially just copied one parameter docstring in the POC PR. Any other suggestions how we could handle this without writing duplicate code?

seisman:
Actually, I'm thinking if it's possible to wrap the `coupe` module in `Figure.meca`. I.e., we may have a parameter like `profile`(True or False). If `profile=True`, then call `coupe` to plot beachballs on profiles, otherwise call `meca` to plot beachballs on maps. In this way, many codes can be reused. The idea may work or not depending on how many common options/features the two modules share, and we need take time to investigate it.

jhtong33:
> Actually, I'm thinking if it's possible to wrap the `coupe` module in `Figure.meca`. I.e., we may have a parameter like `profile`(True or False). If `profile=True`, then call `coupe` to plot beachballs on profiles, otherwise call `meca` to plot beachballs on maps. In this way, many codes can be reused. The idea may work or not depending on how many common options/features the two modules share, and we need take time to investigate it.

The `coupe` module basically has two additional important parameters compared to `meca`. I’m wondering that users still need to specify the start and end longitude/latitude when profile=True in `meca`, similar to how it works in `project`.

pygmt/src/coupe.py
```diff
@@ -0,0 +1,383 @@
...
+    {transparency}
+    """
+    kwargs = self._preprocess(**kwargs)
+
+    ## The cross-sectional profile
+    if kwargs.get("A") is None:
+        raise GMTInvalidInput("The `section` parameter must be specified.")
+    kwargs["A"] = section_convention_code(section_format) + kwargs["A"]
+
+    if isinstance(spec, dict | pd.DataFrame):  # spec is a dict or pd.DataFrame
```
seisman, line 292: @jhtong33 Just want to let you know that, after #3831 and #3827, the long lines of codes can be simiplified to a few lines. Please check the latest `Figure.meca` source code for details.

jhtong33: OK! I will check these PRs later today.

jhtong33: One concern is that "A"(section) in coupe is differ from "A"(offset) in meca.  Therefore, I do not remove `section_convention_code` in coupe so far.

pygmt/src/coupe.py
```diff
@@ -0,0 +1,416 @@
...
+from pygmt.clib import Session
+from pygmt.exceptions import GMTInvalidInput
+from pygmt.helpers import (
+    build_arg_list,
+    data_kind,
+    fmt_docstring,
+    kwargs_to_strings,
+    use_alias,
+)
+from pygmt.src._common import _FocalMechanismConvention
```
seisman, line 16: You can directly import these functions and use them in `Figure.coupe`. Then you can remove lines from 32 to 120:
```suggestion
from pygmt.src.meca import _get_focal_convention, _preprocess_spec, _auto_offset
```

pygmt/src/coupe.py
```diff
@@ -0,0 +1,326 @@
...
+        "lonlat_lonlat": "a",
+        "lonlat_strlen": "b",
+        "xy_xy": "c",
+        "xy_strlen": "d"
+    }
+
+    if section_format in codes:
+        return codes[section_format]
+    else:
+        raise GMTInvalidInput(f"Invalid section format '{section_format}'.")
```
seisman, line 30:
```suggestion
    if section_format not in codes:
        msg = f"Invalid section format '{section_format}'."
        raise GMTInvalidInput(msg)
    return codes[section_format]
```

PR #3818, opened 2025-02-22
Refactor _validate_data_input

seisman:


pygmt/helpers/utils.py
```diff
@@ -115,34 +126,52 @@ def _validate_data_input(
...
+    # Check if too much data is provided.
+    if data is not None and any(v is not None for v in (x, y, z)):
+        msg = "Too much data. Use either data or x/y/z."
+        raise GMTInvalidInput(msg)
+
+    # Determine the data kind if not provided.
+    kind = kind or data_kind(data, required=required_data)
+
+    # Determine the required number of columns based on the required_z flag.
+    required_cols = 3 if required_z else 1
```
seisman, line 138: Most modules usually requires 2 columns but there are exceptions like `histogram` and `info` will works for 1 column. So the default value is 1 if `required_z` is False. 

Will refactor this part in PR #3369

Issue #3826, opened 2025-02-27
Wrong developer version up on v0.14.2


yvonnefroehlich:
There is an error with the versioning up on the release of v0.14.2. This is also wrong on the TestPyPi page at https://test.pypi.org/project/pygmt/#history.

| website | TestPyPi | 
| --- | --- |
| ![Image](https://github.com/user-attachments/assets/6b1bb1d7-36c9-4ad7-acc1-b516c11dbb89) | ![Image](https://github.com/user-attachments/assets/d8407c1f-8f44-43f3-8933-45da36bf22a3) |

Currently I can not install the developer version via

```
pip install --upgrade --pre --extra-index-url https://test.pypi.org/simple/ pygmt
```
and it always installs v0.14.2. To get the developer version I have to use 
```
pip install -i https://test.pypi.org/simple/ pygmt==0.14.1.dev69
```

seisman:
I also noticed the issue. It's very likely due to limitations in the [setuptools_scm](https://github.com/pypa/setuptools-scm/) packages

seisman:
The issue is that, v0.14.2 was created from the v0.14.x branch, rather than the main branch.

The setuptools_scm package uses the following git command (https://setuptools-scm.readthedocs.io/en/latest/config/#setuptools_scm.git.DEFAULT_DESCRIBE) to extract the version information:
```
$ git describe --dirty --tags --long --match '*[0-9]*'
v0.14.0-69-g81fb2e2c6
```
It means the latest commit on the main branch is 81fb2e2c6, and it's 69 commits since the latest tag on the main branch (v0.14.0 because v0.14.2 was not created on the main branch).

PR #3828 is the potential solution.

seisman:
> The key difference is that, after we release v0.14.0, what's the expected string for the next version? `v0.14.1.dev0` or `v0.15.0.dev0`.

Trying to read long [PEP440](https://peps.python.org/pep-0440/), but it's still unclear to me which one is recommended for dev versions.

seisman:
The issue should be fixed by PR #3828. 

The version history on TestPyPI looks correct. The dev documentation still shows v0.14.1.dev70. We need to wait for the next docs builds to verify that it works.

seisman:
> The dev documentation still shows v0.14.1.dev70. We need to wait for the next docs builds to verify that it works.

Now the dev version at https://www.pygmt.org/dev/ is correct.

yvonnefroehlich:
[...]

PR #3828, opened 2025-02-28
Use the 'release-branch-semver' version scheme for setuptools_scm

seisman:
See #3828 for the original issue report.

Upstream explainations at https://setuptools-scm.readthedocs.io/en/latest/extending/#setuptools_scmversion_scheme.

> Semantic versioning for projects with release branches. The same as guess-next-dev (incrementing the pre-release or micro segment) however when on a release branch: a branch whose name (ignoring namespace) parses as a version that matches the most recent tag up to the minor segment. Otherwise if on a non-release branch, increments the minor segment and sets the micro segment to zero, then appends .devN

With the `release-branch-semver` version scheme, the next dev version is `v0.15.0.dev0` rather than `v0.14.1.dev0` after we release `v0.14.0`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ include = ["pygmt*"]
 exclude = ["doc"]
 
 [tool.setuptools_scm]
+version_scheme = "release-branch-semver"
 local_scheme = "node-and-date"
 fallback_version = "999.999.999+unknown"
 


michaelgrund (approved):

seisman:
@weiji14 Hopefully you have time to review this PR and issue #3826, since you may have more experience with package versioning.

weiji14 (approved):
Sorry for the late reply, too much work+travel during the past two weeks. Using `release-branch-semver` seems ok, having `v0.15.0.devX` as the next dev version instead of  `v0.14.1.devX` makes sense, given that PyGMT doesn't do patch releases that often.

PR #3830, opened 2025-03-03
Update the option orders in the Sphinx configuration file

seisman:
The sphinx options in `doc/conf.py` are not well ordered. Sometimes I find it difficult to find if an option is set or not.

This PR updates the order of the sphinx options, following the order presented in https://www.sphinx-doc.org/en/master/usage/configuration.html.

Generally speaking, the options/variables in `doc/conf.py` can be divided into the following categories:

- Non-sphinx variables that can be reused somewhere
- project information
- general options for highlighting, source files, and more 
- Options for sphinx built-in extensions
- Options for third-party extensions
- Options for HTML output
- Options for LaTeX output


doc/conf.py
```diff
@@ -21,7 +23,28 @@
         "gmt": f">={required_gmt_version}",
     }
 )
+# Is a development version or not.
+isdev = "dev" in __version__ or __version__ == "unknown"
```
seisman, line 27: In the old version, we check if version is dev or not using 
```
len(__version__.split("+")) > 1 or __version__ == "unknown":
```
This new way also works and is easier to read.

doc/conf.py
```diff
@@ -21,7 +23,28 @@
...
+    commit_link = f'<a href="{repository_url}/commit/{__commit__}">{__commit__[:8]}</a>'
+else:
+    commit_link = (
+        f'<a href="{repository_url}/releases/tag/{__version__}">{__version__}</a>'
+    )
+
+# Projection information.
+project = "PyGMT"
+author = "The PyGMT Developers"
+copyright = f"2017-{datetime.date.today().year}, {author}"  # noqa: A001
```
seisman, line 42: We can use `f"2017-%Y, {author}"` after we requires sphinx>=8.1
(xref: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-copyright)

michaelgrund (approved):

PR #3832, opened 2025-03-04
Improve the docstrings of the validate_output_table_type function

seisman:
No code changes, just update the docstrings.


seisman:
Self-approving as it's just docstring changes in an internal function.

PR #3861, opened 2025-03-18
Add pre-commit hook for Zizmor

weiji14:
**Description of proposed changes**

A static analysis tool for GitHub Actions, designed to help identify security issues.

References:
- https://woodruffw.github.io/zizmor/usage/#use-with-pre-commit
- https://www.pyopensci.org/blog/python-packaging-security-publish-pypi.html#-3-add-zizmor-to-your-ci-workflows

Fixes #

**Preview**:

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #3866, opened 2025-03-24
Changelog entry for v0.15.0

seisman:
**Description of proposed changes**

This PR adds the changelog entry for v0.15.0. 

- **Release checklist**: #3851
- **Changelog PR of previous minor release**: #3730
- **Preview**: https://pygmt-dev--3866.org.readthedocs.build/en/3866/changes.html

**Checklist**

- [x] Add a new entry in `doc/_static/version_switch.js` for documentation switcher
- [x] Add the version to the compatibility table `doc/minversions.md`
- [ ] Update citation information `CITATION.cff` and BibTeX at https://github.com/GenericMappingTools/pygmt#citing-pygmt
  - [ ] Update authorship list if necessary
  - [x] Update DOI (and URL for BibTeX)
  - [x] Update version
  - [x] Update date released
  - [x] Update the BibTeX key
- [x] Copy the draft release note from https://github.com/GenericMappingTools/pygmt/releases to `doc/changes.md`
- [x] Edit the changelog to look nice ([see maintainers guide for details](https://www.pygmt.org/dev/maintenance.html#updating-the-changelog))
  - [x] Add the separator `---` between the old and new changelog sections
  - [x] Update the release date and the DOI badge in `doc/changes.md`
  - [x] Edit the list of people who contributed to the release, linking to their GitHub account. Sort their names by the number of commits made since the last release (e.g., use `git shortlog HEAD...v0.13.0 -sne`)
  - [x] Edit the change list to remove any trivial changes (updates to the README, typo fixes, CI configuration, etc)
  - [x] Update with any remaining PRs merged before release
  - [x] Add highlights

PR #3867, opened 2025-03-24
CI: Set persist-credentials to true to allow pushing to the gh-pages branch

seisman:
The "Docs" workflow failed to upload the built HTML files to the gh-pages branch since PR #3861.

The `git push -fq origin gh-pages` command failed with errors like below (https://github.com/GenericMappingTools/pygmt/actions/runs/13937126371/job/39007072378#step:14:4408):
```
fatal: could not read Username for 'https://github.com/': No such device or address
```
but the workflow step doesn't fail. That's why we didn't catch it in PR #3861.

The command fails because we set `persist-credentials` to `false` in that PR. This PR fixes the issue.

### DIFF
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -176,7 +176,7 @@ jobs:
           path: deploy
           # Download the entire history
           fetch-depth: 0
-          persist-credentials: false
+          persist-credentials: true
         if: (github.event_name == 'release' || github.event_name == 'push') && (matrix.os == 'ubuntu-latest')
 
       - name: Push the built HTML to gh-pages


weiji14 (approved):
> but the workflow step doesn't fail. That's why we didn't catch it in PR #3861.

Good catch, I knew there had to be at least one job that required extra permissions :laughing:

seisman:
Hmm, it still doesn't work, but now with different error (https://github.com/GenericMappingTools/pygmt/actions/runs/14044608036/job/39322435789#step:14:4472):
```
Pushing changes to gh-pages.
remote: Permission to GenericMappingTools/pygmt.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/GenericMappingTools/pygmt/': The requested URL returned error: 403
```

seisman:
It's likely due to the `permissions: {}` setting added in PR #3861, which disables all permissions for `GITHUB_TOKEN`.

seisman:
The slash command also no longer works (https://github.com/GenericMappingTools/pygmt/pull/3866#issuecomment-2749826736)

https://github.com/GenericMappingTools/pygmt/actions/runs/14049785898/job/39337881112?pr=3866

PR #3870, opened 2025-03-25
CI: Fix the condition check and permission in the Check Links workflow

seisman:
Breaking changes in lychee-actions v2.0.0 (https://github.com/lycheeverse/lychee-action/releases/tag/v2.0.0).

Need to check exit code in step output rather than env.

Also need to set the permissons so that it can create an issue.


seisman:
Triggered workflow run at https://github.com/GenericMappingTools/pygmt/actions/runs/14050259516/job/39339181791

michaelgrund (approved):

PR #3874, opened 2025-03-25
CI: Add issues/contents write permission to the slash commands

seisman:
The "Slash Command Dispatch" workflow needs the issue write permission to add reactions. 

The "format command" workflow needs the content write permission to push changes to PRs.

Not tested, but they should work based on my understanding.

### DIFF
--- a/.github/workflows/slash-command-dispatch.yml
+++ b/.github/workflows/slash-command-dispatch.yml
@@ -10,7 +10,8 @@ on:
     # Add "edited" type for test purposes. Where possible, avoid using to prevent processing unnecessary events.
     # types: [created, edited]
 
-permissions: {}
+permissions:
+  issues: write
 
 jobs:
   slashCommandDispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/slash-command-dispatch.yml
```diff
@@ -10,7 +10,8 @@ on:
     # Add "edited" type for test purposes. Where possible, avoid using to prevent processing unnecessary events.
     # types: [created, edited]
 
-permissions: {}
+permissions:
+  issues: write
```
weiji14, line 14: > The "Slash Command Dispatch" workflow needs the issue write permission to add reactions.
> 
> The "format command" workflow needs the content write permission to push changes to PRs.
> 
> Not tested, but they should work based on my understanding.

Not sure either, but see https://github.com/peter-evans/slash-command-dispatch/issues/376#issuecomment-2628908062? If this doesn't work, try `contents: write`.

weiji14 (approved):

PR #3879, opened 2025-03-27
DOC: Set needs_extensions to check sphinx extension versions

seisman:

Set `needs_extensions` to check the installed sphinx-gallery is >=0.19.0 before trying to build the docs. Related to #3805.

xref: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-needs_extensions

> If set, this value must be a dictionary specifying version requirements for extensions in [extensions](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions). The version strings should be in the 'major.minor' form. Requirements do not have to be specified for all extensions, only for those you want to check.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -62,6 +62,9 @@ extensions = [
     "sphinx_gallery.gen_gallery",
     "sphinxcontrib.cairosvgconverter",
 ]
+needs_extensions = {
+    "sphinx_gallery.gen_gallery": "0.19.0",
+}
 # Options for highlighting.
 pygments_style = "default"
 # Options for object signatures.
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -62,6 +62,9 @@
     "sphinx_gallery.gen_gallery",
     "sphinxcontrib.cairosvgconverter",
 ]
+needs_extensions = {
+    "sphinx_gallery.gen_gallery": "0.19.0",
```
weiji14, line 66: > The version strings should be in the 'major.minor' form.

Remove the patch version part?

```suggestion
    "sphinx_gallery.gen_gallery": "0.19",
```

weiji14 (approved):

Issue #3882, opened 2025-03-29
pygmt.grdclip: Clip the range of grid values


seisman:
*This is the central issue to track the implementation of the `pygmt.grdclip` method.*

- GMT doc: https://docs.generic-mapping-tools.org/6.5/grdclip.html
- PyGMT doc: https://www.pygmt.org/dev/api/generated/pygmt.grdclip.html
- GMT.jl doc: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/grdclip/

## GMT options

*Checked: Implemented; Unchecked: To be implemented/discussed; Strikethrough: Won't implement.*

- [x] `<ingrid>` #1261
- [x] `-G<outgrid>` #1261
- [x] `-R`: `region` #1261
- [x] `-Sa<high>/<above>[+e]`: `above` #1261
- [x] `-Sb<low>/<below>[+e]`: `below` #1261
- [x] `-Si<low>/<high>/<between>`: `between` #1261
- [x] `-Sr<old>/<new>`: `new` #1261
- [x] `-V`: `verbose` #1261
- [ ] ~~`--PAR=<value>`~~: Use `pygmt.config()`

seisman:
`grdclip` is one of the simple modules that have very few options. In PR #3883, I've tried to improve the `pygmt.grdclip` function, but before that, I'd like to ask your opinion about renaming the parameter `new` to `replace`.

> -Srold/new
> Set all data[i] == old to new. This is mostly useful when your data are known to be integer values. Repeat the option for as many replacements as are needed.

- PyGMT uses `new`
- GMT.jl supports `old` or `new` (GMT.jl allows multiple aliases for the same option)
- GMT recommends `replace` or `new` (https://github.com/GenericMappingTools/gmt/blob/23be98456397b75c44f5701d1c47936739f021f7/src/longopt/grdclip_inc.h#L30)

I feel `replace` is a better name than `new`.

yvonnefroehlich:
> I feel `replace` is a better name than `new`.

For me a renaming to `replace` makes sense, as this alias is more descriptive.

PR #3884, opened 2025-03-29
pygmt.grdclip: Deprecate parameter 'new' to 'replace' (remove in v0.19.0)

seisman:
Addressing https://github.com/GenericMappingTools/pygmt/issues/3882#issuecomment-2763311962.

Hopefully it can be approved and merged into v0.15.0

### DIFF
--- a/pygmt/tests/test_grdclip.py
+++ b/pygmt/tests/test_grdclip.py
@@ -82,3 +82,9 @@ def test_grdclip_replace():
     npt.assert_array_equal(np.unique(grid), [0, 1])  # Only have 0 and 1
     grid = grdclip(grid=grid, replace=[0, 2])  # Replace 0 with 2
     npt.assert_array_equal(np.unique(grid), [1, 2])
+
+    # Test for the deprecated 'new' parameter
+    # TODO(PyGMT>=0.19.0): Remove this test for the 'new' parameter
+    with pytest.warns(FutureWarning):
+        grid = grdclip(grid=grid, new=[1, 3])  # Replace 1 with 3
+        npt.assert_array_equal(np.unique(grid), [2, 3])
--- a/pygmt/tests/test_grdclip.py
+++ b/pygmt/tests/test_grdclip.py
@@ -4,9 +4,12 @@ Test pygmt.grdclip.
 
 from pathlib import Path
 
+import numpy as np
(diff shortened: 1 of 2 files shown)


yvonnefroehlich (commented):
Should we add tests for this parameter and the usage of the old parmeter name?

Similar to the `mode` parameter of `grdfill`: https://github.com/GenericMappingTools/pygmt/blob/a11fc1560f7ff692b055ce711ef80fd8dc17eb20/pygmt/tests/test_grdfill.py#L143-L150

Currently, it looks like only `above`, `below` and `between` are included in the tests (https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/tests/test_grdclip.py).

seisman:
> Should we add tests for this parameter and the usage of the old parmeter name?
> 
> Similar to the `mode` parameter of `grdfill`:
> 
> https://github.com/GenericMappingTools/pygmt/blob/a11fc1560f7ff692b055ce711ef80fd8dc17eb20/pygmt/tests/test_grdfill.py#L143-L150
> 
> Currently, it looks like only `above`, `below` and `between` are included in the tests (https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/tests/test_grdclip.py).

Thanks. Added.

pygmt/tests/test_grdclip.py
```diff
@@ -69,3 +72,19 @@ def test_grdclip_no_outgrid(grid, expected_grid):
...
+    """
+    Test the replace parameter for grdclip.
+    """
+    grid = load_earth_mask(region=[0, 10, 0, 10])
+    npt.assert_array_equal(np.unique(grid), [0, 1])  # Only have 0 and 1
+    grid = grdclip(grid=grid, replace=[0, 2])  # Replace 0 with 2
+    npt.assert_array_equal(np.unique(grid), [1, 2])
+
+    # Test for the deprecated 'new' parameter
+    # TODO(PyGMT>=0.19.0): Remove this test for the 'new' parameter
```
weiji14, line 87: Clarify that only the test below is to be removed, not the entire unit test.

```suggestion
    # TODO(PyGMT>=0.19.0): Remove this test below for the 'new' parameter
```

weiji14 (approved):

Issue #3746, opened 2025-01-05
Take co-authoring contributions into consideration


seisman:
Continue discussions in https://github.com/GenericMappingTools/pygmt/pull/3730#discussion_r1899195156.

Currently, we use the `git shortlog -sne` command to count author contributions. Below is a trimmed output on 2025/01/05:
```
  1054  Dongdong Tian <seisman.info@gmail.com>
   278  Yvonne Fröhlich <94163266+yvonnefroehlich@users.noreply.github.com>
   266  dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
   261  Wei Ji <23487320+weiji14@users.noreply.github.com>
   256  Leonardo Uieda <leouieda@gmail.com>
   192  Will Schlitzer <schlitzer90@gmail.com>
   130  Michael Grund <23025878+michaelgrund@users.noreply.github.com>
    68  Meghan Jones <meghanj@alum.mit.edu>
    33  Yao Jiayuan <coreman.seism@gmail.com>
    30  Meghan Jones <meghanj@hawaii.edu>
    16  Michael Grund <michael_grund@gmx.de>
    15  Liam Toney <38269494+liamtoney@users.noreply.github.com>
```
The GitHub contributors page (https://github.com/GenericMappingTools/pygmt/graphs/contributors) shows totally different numbers of commits, likely because GitHub also considers co-authorship. I feel these numbers (including the number of additions and deletions) make more sense.

To consider co-authorship, we can use the following command:
```
git shortlog -sne --group=author --group=trailer:co-authored-by
```
The output is:
```
  1396  Dongdong Tian <seisman.info@gmail.com>
   603  Wei Ji <23487320+weiji14@users.noreply.github.com>
   398  Yvonne Fröhlich <94163266+yvonnefroehlich@users.noreply.github.com>
   266  dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
   258  Leonardo Uieda <leouieda@gmail.com>
   256  Michael Grund <23025878+michaelgrund@users.noreply.github.com>
   237  Will Schlitzer <schlitzer90@gmail.com>
   129  Meghan Jones <meghanj@alum.mit.edu>
    53  Yao Jiayuan <coreman.seism@gmail.com>
    45  Meghan Jones <meghanj@hawaii.edu>
    22  Michael Grund <michael_grund@gmx.de>
    16  actions-bot <58130806+actions-bot@users.noreply.github.com>
[...]

Issue #1744, opened 2022-02-11
Add talks/posters/workshops section to AUTHORSHIP.md


meghanrjones:
**Description of the desired feature**

The authorship policy added in https://github.com/GenericMappingTools/pygmt/pull/726 is really helpful to have for releases and future papers. I found it a bit more confusing how authorship works for conference presentations. For the [AGU 2021 Abstract](https://github.com/GenericMappingTools/pygmt/issues/1395) and the [SciPy 2022 abstract](https://forum.generic-mapping-tools.org/t/pygmt-at-scipy-2022/2584/3) that I'll submit in a few hours, I shared the drafts and asked that the PyGMT authors who are interested in being involved sign on to the abstracts. This is the same approach that @weiji14 took for [EGU 2022](https://forum.generic-mapping-tools.org/t/egu-2022-short-course-proposal/1948). My question is whether it's worth adding recommendations for these types of presentations to [AUTHORSHIP.md](https://github.com/GenericMappingTools/pygmt/blob/main/AUTHORSHIP.md). I think the goals would be to help future developers who haven't presented on software projects before and to make it clear that we would welcome anyone teaching PyGMT without needing to involve all the [PyGMT authors](https://github.com/GenericMappingTools/pygmt/blob/main/AUTHORS.md).

seisman:
> I found it a bit more confusing how authorship works for conference presentations. For the [AGU 2021 Abstract](https://github.com/GenericMappingTools/pygmt/issues/1395) and the [SciPy 2022 abstract](https://forum.generic-mapping-tools.org/t/pygmt-at-scipy-2022/2584/3) that I'll submit in a few hours, I shared the drafts and asked that the PyGMT authors who are interested in being involved sign on to the abstracts. This is the same approach that @weiji14 took for [EGU 2022](https://forum.generic-mapping-tools.org/t/egu-2022-short-course-proposal/1948). 

In my opinion, this is a good way.

> My question is whether it's worth adding recommendations for these types of presentations to [AUTHORSHIP.md](https://github.com/GenericMappingTools/pygmt/blob/main/AUTHORSHIP.md). 

Yes, a section like "Authorship on talks/posters/workshops" sounds good.

> I think the goals would be to help future developers who haven't presented on software projects before and to make it clear that we would welcome anyone teaching PyGMT without needing to involve all the [PyGMT authors](https://github.com/GenericMappingTools/pygmt/blob/main/AUTHORS.md).

[...]

PR #3899, opened 2025-04-08
Reformat AUTHORSHIP.md to wrap lines at 88 characters

seisman:
No content changes, just rewrap short lines to 88 characters, so that we can better view the differences when addressing issues like #3746 and #1744.

### DIFF
--- a/AUTHORSHIP.md
+++ b/AUTHORSHIP.md
@@ -1,7 +1,7 @@
 # Authorship guidelines for academic papers and software archives
 
-First of all, we are deeply thankful to everyone who has helped make PyGMT what it is
-today. Our goal for this document is to establish guidelines for giving credit to
+First of all, we are deeply thankful to everyone who has helped to make PyGMT what it
+is today. Our goal for this document is to establish guidelines for giving credit to
 contributors for their work. To do so, we will attempt to define:
 
 - Fair and diverse ways of providing recognition for contributors' efforts.
--- a/AUTHORSHIP.md
+++ b/AUTHORSHIP.md
@@ -1,90 +1,83 @@
 # Authorship guidelines for academic papers and software archives
 
-First of all, we are deeply thankful to everyone who has helped make PyGMT
-what it is today. Our goal for this document is to establish guidelines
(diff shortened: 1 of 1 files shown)


AUTHORSHIP.md
```diff
@@ -1,90 +1,83 @@
 # Authorship guidelines for academic papers and software archives
 
-First of all, we are deeply thankful to everyone who has helped make PyGMT
-what it is today. Our goal for this document is to establish guidelines
-for giving credit to contributors for their work.
-To do so, we will attempt to define:
+First of all, we are deeply thankful to everyone who has helped make PyGMT what it is
+today. Our goal for this document is to establish guidelines for giving credit to
+contributors for their work. To do so, we will attempt to define:
```
yvonnefroehlich, line 5: For me, it sound better with a "to":

```suggestion
First of all, we are deeply thankful to everyone who has helped to make PyGMT what it
is today. Our goal for this document is to establish guidelines for giving credit to
contributors for their work. To do so, we will attempt to define:
```

seisman: The old one is also grammatically correct, right?

yvonnefroehlich: Hm. I am not sure 🙈.
What do the others @GenericMappingTools/pygmt-maintainers think?

seisman: ChatGPT:

> Both sentences are grammatically correct, but there's a subtle difference in formality and style:
> 
> - ✅ **"helped make"** is the more common and idiomatic phrasing in modern English.
> - ✅ **"helped to make"** is also correct but slightly more formal or traditional.
> 
> So:
> 
> > **"First of all, we are deeply thankful to everyone who has helped make PyGMT what it is today."**
> 
> …is preferred in most modern writing, especially in tech or open-source contexts like this. It's cleaner and more natural.
>

weiji14: Either should be fine, but I'll vote to add the extra "to" :slightly_smiling_face:

seisman: I'm OK with both and have applied the change.

PR #3854, opened 2025-03-15
POC: Extend GMTDataArrayAccessor to support grid operations

seisman:
Address https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2726421306.

This is a POC PR to show how GMTDataArrayAccessor can access GMT's grid-operation modules, i.e., `grid.gmt.fill(**kwargs)` instead of `pygmt.grdfill(grid, **kwargs)`:

```python
In [1]: import pygmt

In [2]: from pygmt.helpers.testing import load_static_earth_relief

In [3]: import numpy as np

In [4]: grid = load_static_earth_relief()

In [5]: grid[3:6, 3:5] = np.nan

In [6]: grid
Out[6]: 
<xarray.DataArray 'z' (lat: 14, lon: 8)> Size: 448B
array([[347.5, 344.5, 386. , 640.5, 617. , 579. , 646.5, 671. ],
       [383. , 284.5, 344.5, 394. , 491. , 556.5, 578.5, 618.5],
       [373. , 367.5, 349. , 352.5, 419.5, 428. , 570. , 667.5],
       [557. , 435. , 385.5,   nan,   nan, 496. , 519.5, 833.5],
       [561.5, 539. , 446.5,   nan,   nan, 553. , 726.5, 981. ],
       [310. , 521.5, 757. ,   nan,   nan, 524. , 686.5, 794. ],
       [521.5, 682.5, 796. , 886. , 571.5, 638.5, 739.5, 881.5],
       [308. , 595.5, 555.5, 556. , 580. , 770. , 927. , 920. ],
       [601. , 526.5, 535. , 299. , 398.5, 645. , 797.5, 964. ],
       [494.5, 488.5, 357. , 254.5, 286. , 484.5, 653.5, 930. ],
       [450.5, 395.5, 366. , 248. , 250. , 354.5, 550. , 797.5],
       [345.5, 320. , 335. , 292. , 207.5, 247. , 325. , 346.5],
       [349. , 313. , 325.5, 247. , 191. , 225. , 260. , 452.5],
       [347.5, 331.5, 309. , 282. , 190. , 208. , 299.5, 348. ]],
      dtype=float32)
Coordinates:
  * lon      (lon) float64 64B -54.5 -53.5 -52.5 -51.5 -50.5 -49.5 -48.5 -47.5
  * lat      (lat) float64 112B -23.5 -22.5 -21.5 -20.5 ... -12.5 -11.5 -10.5
Attributes:
    long_name:     elevation (m)
[...]

PR #3919, opened 2025-04-17
Implement gmt xarray BackendEntrypoint

weiji14:

Allow 'gmt' to be used as an engine in `xarray.open_dataarray` and `xarray.open_dataset` for decoding raster NetCDF (grid) or GeoTIFF (image) files!

TODO:
- [x] Initial implementation supporting `kind="grid"`
- [x] Let `pygmt.io.load_dataarray()` use `engine="gmt"` by default
- [x] Support `raster_kind="image"`
- [x] Documentation improvements

References:
- https://docs.xarray.dev/en/v2025.03.1/internals/how-to-add-new-backend.html
- https://github.com/weiji14/cog3pio/pull/14

Implements idea in https://github.com/GenericMappingTools/pygmt/pull/3673#discussion_r2047833894

**Preview**: https://pygmt-dev--3919.org.readthedocs.build/en/3919/api/generated/pygmt.GMTBackendEntrypoint.html

**Reminders**

**Slash Commands**


weiji14:
Probably best to split this into two PRs so that we have two changelog entries:
1. (feature) Implement gmtread xarray BackendEntrypoint
2. \*\*Breaking\*\* Set `load_dataarray`'s default engine to `engine="gmtread"`, and require `decode_kind` parameter to be set (e.g. to `decode_kind='grid'`)

But please take a look at the current implementation, and suggest comments first.

weiji14:
> 1. Why not just use `"gmt"` as the engine name?

Yes, we could do that actually, and `engine="gmt"` will make more sense if we want to implement writing to a NetCDF driver later via [`xr.DataArray.to_netcdf`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.to_netcdf.html#xarray.DataArray.to_netcdf).

> 2. The `load_dataarray` function is almost the same as the [`xr.load_dataarray`](https://docs.xarray.dev/en/stable/generated/xarray.load_dataarray.html) function, except that `load_dataarray` also get the GMT accessor information. With the "gmtread" engine added, `xr.load_dataarray("xxx.grd", engine="gmtread", decode_kind="grid")` works exactly the same as `load_dataarray("xxx.grd", decode_kind="grid")`. So maybe we can deprecate `load_dataarray`?

True, the GMT accessor loading is already in the BackendEntrypoint, so we might as well deprecate `load_dataarray`. Ok to do it in 2 releases (remove in v0.18.0) or 4 releases (remove in v0.20.0)?

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+from pygmt.src.which import which
+from xarray.backends import BackendEntrypoint
+
+
+class GMTBackendEntrypoint(BackendEntrypoint):
+    """
+    Xarray backend to read raster grid/image files using 'gmt' engine.
+
+    Relies on the libgdal-netcdf driver used by GMT C for NetCDF files, and libgdal for
+    GeoTIFF files.
```
seisman, line 22: This is not 100% dorrect. GMT uses the netCDF libray for reading/writing .nc files, not the libgdal-netcdf driver. I'm even unsure if libgdal-netcdf is needed.

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    description = "Open raster (.grd, .nc or .tif) files in Xarray via GMT read."
```
seisman, line 25:
```suggestion
    description = "Open raster (.grd, .nc or .tif) files in Xarray via GMT."
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+    open_dataset_parameters = ("filename_or_obj", "kind")
```
seisman, line 26:
```suggestion
    open_dataset_parameters = ("filename_or_obj", "decode_kind")
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def guess_can_open(self, filename_or_obj) -> bool:
+        """
+        Backend open_dataset method used by Xarray in :py:func:`~xarray.open_dataset`.
+        """
+        try:
+            ext = Path(filename_or_obj).suffix
+        except TypeError:
+            return False
+        return ext in {".grd", ".nc", ".tif"}
```
seisman, line 72: What about `.tiff`?

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,57 @@
...
+from pygmt.enums import GridRegistration, GridType
+from pygmt.exceptions import GMTInvalidInput
+
+
+def test_xarray_backend_gmt_read_grid():
+    """
+    Ensure that passing engine='gmt' to xarray.open_dataarray works for reading
+    NetCDF grids.
+    """
+    with xr.open_dataarray(
```
seisman, line 18: Maybe also add one test for `xr.load_dataarray`?

pygmt/xarray_backend.py
weiji14: Hmm, you're right. I was getting some error when I had a GMT dev `6.6.0.dev5+b3580a7` version installed without `libgdal-netcdf`, but can't seem to reproduce it anymore. So to update the note here, we say that GMT is calling the NetCDF C library directly somehow?

pygmt/xarray_backend.py
weiji14:
```suggestion
        return ext in {".grd", ".nc", ".tif", ".tiff"}
```

weiji14: Ok, added `.tiff`. Any other known extensions to add?

pygmt/xarray_backend.py
seisman: > we say that GMT is calling the NetCDF C library directly somehow?

Yes.

pygmt/io.py
```diff
@@ -2,10 +2,14 @@
 PyGMT input/output (I/O) utilities.
 """
 
+from typing import Literal
```
seisman, line 5: Since we're going to deprecate `load_dataarray`(xref: https://github.com/GenericMappingTools/pygmt/pull/3919#issuecomment-2812018012), we should revert the changes in this file.

pygmt/helpers/testing.py
```diff
@@ -154,7 +154,7 @@ def load_static_earth_relief():
         A grid of Earth relief for internal tests.
     """
     fname = which("@static_earth_relief.nc", download="c")
-    return load_dataarray(fname)
+    return load_dataarray(fname, decode_kind="grid")
```
seisman, line 157: Again, with `load_dataarray` deprecated, we should use `xr.load_dataarray(fname, engine="gmt", decode_kind="grid")` instead, but it can be done in a separate PR.

doc/api/index.rst
```diff
@@ -173,6 +173,7 @@ Input/output
 .. autosummary::
     :toctree: generated
 
+    GMTBackendEntrypoint
```
seisman, line 176: I'm wondering if it makes more sense to have a separate category like "Xarray Integration" and put both `GMTBackendEntrypoint` and `GMTDataArrayAccessor` in the category. Also, if #3854 is implemented, `GMTDataArrayAccessor` no longer makes sense in the "Metadata" category.

pygmt/xarray_backend.py
weiji14: Done in c5ebf9f27b8cecbbdb998ffe833df838f3cdd200

pygmt/tests/test_xarray_backend.py
weiji14: Added in commit c0ccd4b0cdbf1708ad7b6adca0750b0907a7005f

doc/api/index.rst
weiji14: Yeah, I was wondering about this too. I can put `GMTBackendEntrypoint` in a new "Xarray Integration" section as you suggested.

pygmt/helpers/testing.py
weiji14: Started draft PR at #3922, will cherrypick things across.

pygmt/io.py
weiji14: Do we want to change the default engine in load_dataarray to `engine="gmt"` or no?

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,76 @@
...
+        filename_or_obj="@earth_relief_20m_holes.grd", engine="gmt", decode_kind="grid"
+    ) as da:
+        assert isinstance(
+            da.data,
+            np.ndarray,  # ensure data is in memory and not a dask array
+        )
+        assert da.sizes == {"lat": 31, "lon": 31}
+        assert da.dtype == "float32"
+        assert da.gmt.registration == GridRegistration.GRIDLINE
+        assert da.gmt.gtype == GridType.GEOGRAPHIC
```
seisman, line 57: No need to call `xr.load_dataarray` as a context manager, right?

pygmt/io.py
seisman: Changing the default engine to `engine="gmt"` means that users have to set `decode_kind`, so it's a breaking change. It makes little sense to introduce a breaking change to a deprecated function, so my answer is no.

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+                )
+                # Add "source" encoding
+                source = which(fname=filename_or_obj)
+                raster.encoding["source"] = (
+                    source[0] if isinstance(source, list) else source
+                )
+                _ = raster.gmt  # Load GMTDataArray accessor information
+                return raster.to_dataset()
+
+    def guess_can_open(self, filename_or_obj) -> bool:
```
seisman, line 64: Maybe we should add tests to ensure the `guess_can_open` function works?

doc/api/index.rst
seisman: Move `GMTDataArrayAccessor` to the "Xarray Integration" section in this PR?

doc/api/index.rst
```diff
@@ -197,6 +197,14 @@ Getting metadata from tabular or grid data:
     info
     grdinfo
 
+Xarray integration
```
seisman, line 200:
```suggestion
Xarray Integration
```

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def test_xarray_backend_gmt_open_nc_grid():
+    """
+    Ensure that passing engine='gmt' to xarray.open_dataarray works for opening
+    NetCDF grids.
+    """
+    with xr.open_dataarray(
+        filename_or_obj="@static_earth_relief.nc", engine="gmt", decode_kind="grid"
+    ) as da:
```
seisman, line 21:
```suggestion
    with xr.open_dataarray(
       "@static_earth_relief.nc", engine="gmt", decode_kind="grid"
    ) as da:
```

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def test_xarray_backend_gmt_load_grd_grid():
+    """
+    Ensure that passing engine='gmt' to xarray.open_dataarray works for loading
+    GRD grids.
+    """
+    with xr.load_dataarray(
+        filename_or_obj="@earth_relief_20m_holes.grd", engine="gmt", decode_kind="grid"
+    ) as da:
```
seisman, line 49:
```suggestion
    with xr.load_dataarray(
        "@earth_relief_20m_holes.grd", engine="gmt", decode_kind="grid"
    ) as da:
```

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,76 @@
...
+        assert isinstance(
+            da.data,
+            np.ndarray,  # ensure data is in memory and not a dask array
+        )
```
seisman, line 53:
```suggestion
        # ensure data is in memory and not a dask array
        assert isinstance(da.data, np.ndarray)
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+from pygmt.src.which import which
+from xarray.backends import BackendEntrypoint
+
+
+class GMTBackendEntrypoint(BackendEntrypoint):
+    """
+    Xarray backend to read raster grid/image files using 'gmt' engine.
+
+    Internally, GMT uses the NetCDF C library to read NetCDF files, and GDAL for GeoTIFF
+    files.
```
seisman, line 22: Need to add a few examples in the docstring.

pygmt/xarray_backend.py
seisman: > guess_can_open is used to identify the proper engine to open your data file automatically in case the engine is not specified explicitly.

I was expecting that the following code should work:
```
import pygmt
import xarray as xr 

xr.load_dataarray("earth_relief_01d_g.grd", decode_kind="grid")
```
but I got the following error:
```
NetCDF4BackendEntrypoint.open_dataset() got an unexpected keyword argument 'decode_kind'
```
It's likely that the built-in NetCDF4 backend has a higher priority than other 3rd-party engines, so the `guess_can_open` function doesn't work.

weiji14: > It's likely that the built-in NetCDF4 backend has a higher priority than other 3rd-party engines, so the `guess_can_open` function doesn't work.

Yes, according to the docs at https://github.com/pydata/xarray/blob/ee862fe12163b728a638216195217aa935238dc3/xarray/backends/api.py#L750-L752 `engine="netcdf4"` will be preferred (if installed) over other engines. Either use `xr.load_dataarray("earth_relief_01d_g.grd", engine="gmt", decode_kind="grid")`, or uninstall `netcdf4`.

pygmt/tests/test_xarray_backend.py
weiji14: Not sure. The context manager is supposed to ensure the file handle gets closed after use (see note under https://docs.xarray.dev/en/v2025.03.1/user-guide/io.html#reading-multi-file-datasets), at least for `xr.open_dataset`. For `load_dataarray`, I'm not sure if the file handle gets closed automatically on load.

To be honest, I'm unsure how the 'gmt' engine handles file locks compared to the 'netcdf4' engine when dealing with NetCDF4 files. E.g. if it respects environment variables like `HDF5_USE_FILE_LOCKING`.

doc/api/index.rst
weiji14: > Move `GMTDataArrayAccessor` to the "Xarray Integration" section in this PR?

I think do that in #3854. Hopefully this PR can be done before that.

pygmt/io.py
weiji14: Ok, I've reverted changes to `io.py` in commit f722b479903458f4349a3e0afb75d6c6fcae461f. So keeping default engine behaviour as before for the `load_dataarray` function that will be deprecated in #3922.

pygmt/tests/test_xarray_backend.py
seisman: > Not sure. The context manager is supposed to ensure the file handle gets closed after use (see note under https://docs.xarray.dev/en/v2025.03.1/user-guide/io.html#reading-multi-file-datasets), at least for `xr.open_dataset` with `engine="netcdf4"`. For `load_dataarray`, I'm not sure if the file handle gets closed automatically on load.

Based on my understanding of the `load_dataarray` source code https://github.com/pydata/xarray/blob/main/xarray/backends/api.py#L313-L336, I think the file handle is always closed. 

> To be honest, I'm unsure how the 'gmt' engine handles file locks compared to the 'netcdf4' engine when dealing with NetCDF4 files. E.g. if it respects environment variables like `HDF5_USE_FILE_LOCKING`.

Based on the GMT source code (https://github.com/GenericMappingTools/gmt/blob/9a8769f905c2b55cf62ed57cd0c21e40c00b3560/src/gmt_nc.c#L1545), GMT always closes the nc file after reading.

pygmt/xarray_backend.py
seisman: I guess it means we can remove the `guess_can_open` function.

pygmt/xarray_backend.py
weiji14: Added a couple of examples in commit 0c6b8a3f225058dbf401daf365b815307760d9ab

pygmt/xarray_backend.py
weiji14: I think it makes sense to keep `guess_can_open` (even if it is optional), if we remove the NetCDF4 dependency (https://github.com/GenericMappingTools/pygmt/pull/3643). Or we can simply do `def guess_can_open: return False` following [this](https://github.com/axiom-data-science/extract_model/blob/b9aaeddaa98d51609d169a0a1a9d4a097545385d/extract_model/xr/triangular_mesh_netcdf.py#L66-L68) so that `engine='gmt'` always has to be declared explicitly.

seisman: > I think it makes sense to keep `guess_can_open` (even if it is optional), if we remove the NetCDF4 dependency (#3643).

Without netCDF4 installed, users are required to specify the `decode_kind` parameter, e.g., `xr.open_dataarray("xxx.nc", decode_kind="grid")`. However, this call fails when netCDF4 is installed, making `xr.open_dataarray("xxx.nc", decode_kind="grid")` non-portable. In other words, if people want to use the gmt engine, it's always recommended to specify `engine="gmt"`.

https://docs.xarray.dev/en/latest/internals/how-to-add-new-backend.html#guess-can-open

> If you are not interested in supporting this feature, you can skip this step since [BackendEntrypoint](https://docs.xarray.dev/en/latest/generated/xarray.backends.BackendEntrypoint.html#xarray.backends.BackendEntrypoint) already provides a default [guess_can_open()](https://docs.xarray.dev/en/latest/generated/xarray.backends.BackendEntrypoint.guess_can_open.html#xarray.backends.BackendEntrypoint.guess_can_open) that always returns False.

As mentioned above, `def guess_can_open: return False` is not necessary.

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
```
seisman, line 22: Many other raster formats can be read via GDAL.
```suggestion
    Internally, GMT uses the NetCDF C library to read NetCDF files, and GDAL for GeoTIFF
    and other raster formats.
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+
+    When using :py:func:`xarray.open_dataarray` or :py:func:`xarray.load_dataarray` with
+    ``engine='gmt'``, pass the ``decode_kind`` parameter that can be either:
```
seisman, line 25:
```suggestion
    ``engine="gmt"``, pass the ``decode_kind`` parameter that can be either:
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+
+    - ``grid`` - for reading single-band raster grids
+    - ``image`` - for reading multi-band raster images
```
seisman, line 28:
```suggestion
    - ``"grid"`` - for reading single-band raster grids
    - ``"image"`` - for reading multi-band raster images
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+    Examples
+    --------
+    Read a single-band NetCDF file using ``decode_kind="grid"``
+
+    >>> import pygmt
+    >>> import xarray as xr
+    >>>
+    >>> da_grid: xr.Dataset = xr.open_dataarray(
+    ...     "@static_earth_relief.nc", engine="gmt", decode_kind="grid"
+    ... )
```
seisman, line 39: This type hint is not necessary:
```suggestion
    >>> da_grid = xr.open_dataarray(
    ...     "@static_earth_relief.nc", engine="gmt", decode_kind="grid"
    ... )
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+        Conventions:   CF-1.7
+        title:         Produced by grdcut
+        history:       grdcut @earth_relief_01d_p -R-55/-47/-24/-10 -Gstatic_eart...
+        description:   Reduced by Gaussian Cartesian filtering (111.2 km fullwidt...
+        actual_range:  [190. 981.]
+        long_name:     elevation (m)
+
+    Read a multi-band GeoTIFF file using ``decode_kind="image"``
+
+    >>> da_image: xr.Dataset = xr.open_dataarray(
```
seisman, line 56:
```suggestion
    >>> da_image = xr.open_dataarray(
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+        long_name:  z
+    """
+
+    description = "Open raster (.grd, .nc or .tif) files in Xarray via GMT."
+    open_dataset_parameters = ("filename_or_obj", "decode_kind")
+    url = "https://github.com/GenericMappingTools/pygmt"
+
+    def open_dataset(  # type: ignore[override]
+        self,
+        filename_or_obj: str | os.PathLike,
```
seisman, line 76:
```suggestion
        filename_or_obj: PathLike,
```

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,120 @@
...
+        *,
+        drop_variables=None,  # noqa: ARG002
+        decode_kind: Literal["grid", "image"],
```
seisman, line 79: Why choose the name `decode_kind`? The original [`xr.open_dataarray`](https://docs.xarray.dev/en/stable/generated/xarray.open_dataarray.html#xarray-open-dataarray) method has parameters like `decode_cf`/`decode_times`/`decode_timedelta`, but here `decode` implies reading metadata from netCDF and converting to meaningful Python objects (e.g., datetime64). Maybe `raster_kind` is a better name?

pygmt/tests/test_xarray_backend.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    with xr.load_dataarray(
+        "@earth_relief_20m_holes.grd", engine="gmt", decode_kind="grid"
+    ) as da:
+        # ensure data is in memory and not a dask array
+        assert isinstance(da.data, np.ndarray)
+        assert da.sizes == {"lat": 31, "lon": 31}
+        assert da.dtype == "float32"
+        assert da.gmt.registration == GridRegistration.GRIDLINE
+        assert da.gmt.gtype == GridType.GEOGRAPHIC
```
seisman, line 53: Remove the `with` statement; no need to mention dask array; check the min value.
```suggestion
    da = xr.load_dataarray(
        "@earth_relief_20m_holes.grd", engine="gmt", decode_kind="grid"
    )
    # Ensure data is in memory.
    assert isinstance(da.data, np.ndarray)
    assert da.data.min() == -4929.5
    assert da.sizes == {"lat": 31, "lon": 31}
    assert da.dtype == "float32"
    assert da.gmt.registration == GridRegistration.GRIDLINE
    assert da.gmt.gtype == GridType.GEOGRAPHIC
```

pygmt/xarray_backend.py
weiji14: What about if we remove `.nc` from the `guess_can_open` list, and just keep `.grd`, `.tif`, etc?

pygmt/xarray_backend.py
weiji14: Yeah, it does sound like `decode_` is more decoding CF metadata variables, especially after reading https://github.com/pydata/xarray/issues/8548 where they're talking about `encode` and `decode` methods. Lemme rename `decode_kind` to `raster_kind` then.

pygmt/xarray_backend.py
seisman: > In that case, I suppose we should just remove the `guess_can_open` method from GMTBackendEntrypoint so that `engine="gmt"` always needs to be passed in explicitly.

Yes, I feel it's always good to explicitly specify the engine.

weiji14: The `guess_can_open` method has been removed in commit c97bd82448225fd8feea57d9aab24e10fb844710.

(I had an idea that we can try detecting if the file starts with `@`, since only GMT supports it as far as I know, but let's not go there :laughing:)

seisman:
This PR almost looks good, except that the doctest doesn't look good (what's `ivar`?):
Screenshot from 2025-04-23 10-25-32

weiji14:
> This PR almost looks good, except that the doctest doesn't look good (what's `ivar`?):

Appears to be a sphinx thing. I found mention of a [`napoleon_use_ivar`](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#confval-napoleon_use_ivar) setting that converts the `.. attribute::` directive, and because the docstring output has `Attributes:`, sphinx converts it to ivar :sweat_smile:

seisman:
> > This PR almost looks good, except that the doctest doesn't look good (what's `ivar`?):
> 
> Appears to be a sphinx thing. I found mention of a [`napoleon_use_ivar`](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#confval-napoleon_use_ivar) setting set [here](https://github.com/GenericMappingTools/pygmt/blame/4a2d45d9e2b0239ba20b38ca2e9ffa3510d7ed1f/doc/conf.py#L126) (added in #383) that converts the `.. attribute::` directive, and because the docstring output has `Attributes:`, sphinx converts it to ivar 😅

The xarray documentation also uses napoleon, but the xarray repr looks correct (e.g., https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html#xarray.DataArray), but I don't find any special settings in the `conf.py` file (https://github.com/pydata/xarray/blob/main/doc/conf.py).

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,106 @@
...
+        self,
+        filename_or_obj: PathLike,
+        *,
+        drop_variables=None,  # noqa: ARG002
+        raster_kind: Literal["grid", "image"],
+        # other backend specific keyword arguments
+        # `chunks` and `cache` DO NOT go here, they are handled by xarray
+    ) -> xr.Dataset:
+        """
+        Backend open_dataset method used by Xarray in :py:func:`~xarray.open_dataset`.
```
seisman, line 83: With the fix in #3927, sphinx-autogen will generate stub file for this method. It's not ideal, and we need add a "Parameters" section at least.

pygmt/xarray_backend.py
weiji14, line 1: Considering moving this file to `pygmt/xarray/backend.py`, and then potentially in #3854, we can move `pygmt/accessors.py` to `pygmt/xarray/accessors.py`? And if we do pandas accessors (xref https://github.com/GenericMappingTools/pygmt/pull/3854#discussion_r2010845385) in the future, we could put it under `pygmt/pandas/accessor.py`?

seisman: Sounds good

pygmt/xarray_backend.py
```diff
@@ -0,0 +1,106 @@
...
+    Coordinates:
+      * lat      (lat) float64... -23.5 -22.5 -21.5 -20.5 ... -12.5 -11.5 -10.5
+      * lon      (lon) float64... -54.5 -53.5 -52.5 -51.5 -50.5 -49.5 -48.5 -47.5
+    Attributes:
+        Conventions:   CF-1.7
+        title:         Produced by grdcut
+        history:       grdcut @earth_relief_01d_p -R-55/-47/-24/-10 -Gstatic_eart...
+        description:   Reduced by Gaussian Cartesian filtering (111.2 km fullwidt...
+        actual_range:  [190. 981.]
+        long_name:     elevation (m)
```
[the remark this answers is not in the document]
weiji14: I put the ellipisis after the colon in 'Attributes:`, and it seemed to remove the `:ivar` from showing up (see commit f2446fc9c3fed7a7b051bebad1818e905a291927):

image

Only downside is the ellipsis shows as `Attributes: ...`, but should be acceptable.

pygmt/xarray/backend.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Internally, GMT uses the netCDF C library to read netCDF files, and GDAL for GeoTIFF
+    and other raster formats. See also
+    :gmt-docs:`reference/features.html#grid-file-format`.
+
+    When using :py:func:`xarray.open_dataarray` or :py:func:`xarray.load_dataarray` with
+    ``engine="gmt"``, pass the ``raster_kind`` parameter that can be either:
+
+    - ``"grid"`` - for reading single-band raster grids
+    - ``"image"`` - for reading multi-band raster images
```
seisman, line 28: xref: https://github.com/GenericMappingTools/pygmt/issues/3556
```suggestion
    - ``"grid"``: for reading single-band raster grids
    - ``"image"``: for reading multi-band raster images
```

pygmt/xarray/backend.py
```diff
@@ -0,0 +1,116 @@
...
+class GMTBackendEntrypoint(BackendEntrypoint):
+    """
+    Xarray backend to read raster grid/image files using 'gmt' engine.
+
+    Internally, GMT uses the netCDF C library to read netCDF files, and GDAL for GeoTIFF
+    and other raster formats. See also
+    :gmt-docs:`reference/features.html#grid-file-format`.
+
+    When using :py:func:`xarray.open_dataarray` or :py:func:`xarray.load_dataarray` with
+    ``engine="gmt"``, pass the ``raster_kind`` parameter that can be either:
```
seisman, line 25:
```suggestion
    ``engine="gmt"``, the ``raster_kind`` parameter is required and can be either:
```

pygmt/xarray/backend.py
```diff
@@ -0,0 +1,116 @@
...
+from xarray.backends import BackendEntrypoint
+
+
+class GMTBackendEntrypoint(BackendEntrypoint):
+    """
+    Xarray backend to read raster grid/image files using 'gmt' engine.
+
+    Internally, GMT uses the netCDF C library to read netCDF files, and GDAL for GeoTIFF
+    and other raster formats. See also
+    :gmt-docs:`reference/features.html#grid-file-format`.
```
seisman, line 22: Maybe add a few sentences explaining why users may want to read netCDF/GeoTIFF files via GMT, rather than netcdf4/rasterio, something like:

Compared to the "netcdf4"/"rasterio" engines, the "gmt" engine can read GMT remote files (file names starting with `@`) directly and provides the GMTDataArray accessor `.gmt` for easy access to GMT-specific features and metadata.

weiji14: Ok, updated the description in commit acab8b2cfbb2260960fccef6201dcc215980cad9, please review. I didn't mention the `netcdf4`/`rasterio` engines directly, but just mentioned that the GMT engine works better with GMT remote files and features in general.

seisman:
Looks pretty good to me.

weiji14:
Will leave this up for review until end of the week-ish.

seisman (approved):

PR #3920, opened 2025-04-17
TYP: Add PathLike and TableLike types for type hints and remove the table-like placeholder

seisman:
This PR adds two types for type hints:

- `PathLike`: `str | os.PathLike`, representing a path to an ASCII file
- `TableLike`: ` dict | np.ndarray | pd.DataFrame | xr.Dataset | gpd.GeoDataFrame` . 2-D table-like object that contains tabular data. Similar to the `table-like` placeholder https://github.com/GenericMappingTools/pygmt/blob/99a63402ccdbca4b35619f004b833406916f562c/pygmt/helpers/decorators.py#L458-L460

and does the following replacements:

- `data` => `data: PathLike | TableLike`
- `data=None` => `data: PathLike | TableLike | None = None`
- `outgrid: str | None = None` => `outgrid: PathLike | None = None`
- `grid` => `grid: PathLike | xr.DataArray`
- `outgrid: str | None = None` => `outgrid: PathLike | None = None`

**Preview**:

- https://pygmt-dev--3920.org.readthedocs.build/en/3920/api/generated/pygmt.grdclip.html#pygmt.grdclip
- https://pygmt-dev--3920.org.readthedocs.build/en/3920/api/generated/pygmt.xyz2grd.html#pygmt.xyz2grd

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -582,18 +582,18 @@ def build_arg_list(  # noqa: PLR0912
 
     if infile:  # infile can be a single file or a list of files
         if isinstance(infile, str | os.PathLike):
-            gmt_args = [str(infile), *gmt_args]
+            gmt_args = [os.fspath(infile), *gmt_args]
         else:
-            gmt_args = [str(_file) for _file in infile] + gmt_args
+            gmt_args = [os.fspath(_file) for _file in infile] + gmt_args
     if outfile is not None:
         if (
             not isinstance(outfile, str | os.PathLike)
-            or str(outfile) in {"", ".", ".."}
-            or str(outfile).endswith(("/", "\\"))
+            or os.fspath(outfile) in {"", ".", ".."}
+            or os.fspath(outfile).endswith(("/", "\\"))
         ):
             msg = f"Invalid output file name '{outfile}'."
             raise GMTInvalidInput(msg)
-        gmt_args.append(f"->{outfile}")
+        gmt_args.append(f"->{os.fspath(outfile)}")
     return gmt_args
 
 
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -4,7 +4,6 @@ Utilities and common tasks for wrapping the GMT modules.
 
 import io
 import os
-import pathlib
 import shutil
 import string
 import subprocess
@@ -17,6 +16,7 @@ from pathlib import Path
 from typing import Any, Literal
 
 import xarray as xr
+from pygmt._typing import PathLike
 from pygmt.encodings import charset
 from pygmt.exceptions import GMTInvalidInput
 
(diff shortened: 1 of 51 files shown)


pygmt/helpers/utils.py
```diff
@@ -387,10 +387,10 @@ def data_kind(
     match data:
         case None if required:  # No data provided and required=True.
             kind = "empty"
-        case str() | pathlib.PurePath():  # One file.
+        case str() | os.PathLike():  # One file.
```
weiji14, line 390: A little unsure about this, [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) is an abstract base class that only needs `__fspath__` to be implemented as far as I can tell, and that can return either `str` or `bytes`. GMT can accept `str`, but unsure about raw `bytes`.

weiji14 (commented):
Couple of (optional) suggestions on output typehints for info/grdinfo, and one other comment/question to verify.

pygmt/helpers/utils.py
seisman: The GMT API does not support raw `bytes` as file paths. However, according to the Python documentation for `__fspath__`:

> The method should only return a str or bytest object, with a preference for str.

To my knowledge, no built-in Python classes return `bytes` from `__fspath__()`, so this case should be rare in practice. In earlier versions, such `bytes`-returning path-like objects are recognized as vectors; now, they are treated as file inputs instead.

pygmt/helpers/utils.py
```diff
@@ -581,13 +581,13 @@ def build_arg_list(  # noqa: PLR0912
         gmt_args.extend(f"--{key}={value}" for key, value in confdict.items())
 
     if infile:  # infile can be a single file or a list of files
-        if isinstance(infile, str | pathlib.PurePath):
+        if isinstance(infile, str | os.PathLike):
             gmt_args = [str(infile), *gmt_args]
```
weiji14, line 585: > The GMT API does not support raw `bytes` as file paths. However, according to the Python documentation for `__fspath__`:
> 
> > The method should only return a str or bytest object, with a preference for str.
> 
> To my knowledge, no built-in Python classes return `bytes` from `__fspath__()`, so this case should be rare in practice. In earlier versions, such `bytes`-returning path-like objects are recognized as vectors; now, they are treated as file inputs instead.

Ok, let's assume the `bytes` case is rare enough to ignore. We're still doing `str(infile)` here, but maybe we'll need to use something like `str(infile.__fspath__())` instead if using `os.PathLike`?

seisman: Changed to [`os.fspath`](https://docs.python.org/3/library/os.html#os.fspath) in commit 7c0a109fc1f874ab23afbe90fd63381ee9cf621f.
```
In [1]: from pathlib import Path

In [2]: import os

In [3]: os.fspath(Path("abc/def"))
Out[3]: 'abc/def'

In [4]: os.fspath("path/to/file")
Out[4]: 'path/to/file'
```

weiji14: Nice, I should have scrolled up the page a bit to see that `os.fspath` function! Also found [PEP 519](https://peps.python.org/pep-0519) to be a good reference that does seem to suggest `os.fspath`.

weiji14 (approved):

Issue #3843, opened 2025-03-11
Add documentation for common options/parameter in the "Technical Reference" section


seisman:
GMT modules have common options (e.g., -J/-R/-V), which are aliased to standardized parameters in PyGMT (e.g., projection, region, verbose). Currently, the docstrings for these parameters are lengthy and take up a significant portion of the documentation pages, making it harder to focus on module-specific options.

To improve readability, I propose moving the detailed explanations of these common parameters to a dedicated page in the "Technical Reference" section. In the wrapper documentation, their descriptions can be condensed, with a link to the full explanation. 

For example, currently, the `verbose` has docstrings like below

![Image](https://github.com/user-attachments/assets/b60907bf-1d0b-42c1-9e6f-377fb0f89950)

and it can be simplified to:
```
verbose  
    Select verbosity level. [See detailed explanation.]
```

This approach reduces the documentation length of wrappers, and makes wrapper-specific parameters more prominent.

yvonnefroehlich:
This approach seems to be similar to what is done in the GMT docs:
![Image](https://github.com/user-attachments/assets/a46f99fe-936f-438f-bbce-e4f5eec95f5a)

If I understand it corrently this would replace the enteries in the decorators file. And we can consider this to avoid writing dupliate code in `Figure.meca` and `Figure.cope`.

seisman:
> If I understand it correctly, this would replace the entries in the decorators file. 

No, it won't replace the entries but can simplify them.

> And we can consider this to avoid writing duplicate code in `Figure.meca` and `Figure.cope`.

This "Common parameters" page should only be used for common parameters shared by all GMT modules, so it won't solve the issue for `Figure.meca`/`Figure.coupe`.

PR #3923, opened 2025-04-22
Improve the description for the 'cores' parameter

seisman:
Upstream documentation at: https://docs.generic-mapping-tools.org/6.5/gmt.html#core-full

Address https://github.com/GenericMappingTools/pygmt/issues/3843. The docstring of this option is not that long so we don't have to move it to https://www.pygmt.org/dev/techref/common_parameters.html

**Preview**: https://pygmt-dev--3923.org.readthedocs.build/en/3923/api/generated/pygmt.grdsample.html#pygmt.grdsample

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -71,14 +71,11 @@ COMMON_DOCSTRINGS = {
             :gmt-docs:`gmt.html#f-full`.""",
     "cores": r"""
         cores : bool or int
-            [[**-**]\ *n*].
-            Limit the number of cores to be used in any OpenMP-enabled
-            multi-threaded algorithms. By default we try to use all available
-            cores. Set a number *n* to only use n cores (if too large it will
-            be truncated to the maximum cores available). Finally, give a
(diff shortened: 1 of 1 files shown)


pygmt/helpers/decorators.py
```diff
@@ -71,14 +71,11 @@
...
-            multi-threaded algorithms. By default we try to use all available
-            cores. Set a number *n* to only use n cores (if too large it will
-            be truncated to the maximum cores available). Finally, give a
-            negative number *-n* to select (all - *n*) cores (or at least 1 if
-            *n* equals or exceeds all).
-            """,
+            Specify the number of active cores to be used in any OpenMP-enabled
+            multi-threaded algorithms. By default, all available cores are used. Set a
+            number *n* to use *n* cores (if too large it will be truncated to the
+            maximum cores available); or set a negative number *-n* to select
```
yvonnefroehlich, line 77: Even though this is only a detail, we should update the upstream GMT documentation (https://docs.generic-mapping-tools.org/dev/gmt.html#core-full) to use `negative number -n` (with the minus sign).

yvonnefroehlich (approved):

PR #3930, opened 2025-04-26
DOC: Improve the docstrings of the 'panel' parameter

seisman:
**Preview**: https://pygmt-dev--3930.org.readthedocs.build/en/3930/api/generated/pygmt.Figure.basemap.html#pygmt.Figure.basemap

| Old | New |
|---|---|
| image|image|


pygmt/helpers/decorators.py
```diff
@@ -262,14 +262,15 @@
...
-            [*row,col*\|\ *index*].
-            Select a specific subplot panel. Only allowed when in subplot
-            mode. Use ``panel=True`` to advance to the next panel in the
-            selected order. Instead of *row,col* you may also give a scalar
-            value *index* which depends on the order you set via ``autolabel``
-            when the subplot was defined. **Note**: *row*, *col*, and *index*
-            all start at 0.
-         """,
+            Select a specific subplot panel. Only allowed when used in
+            :meth:`Figure.subplot`.
```
yvonnefroehlich, line 266: Hm. With the current formulation, it sounds a bit like that `panel` is a parameter of `Figure.subplot`. Maybe just keep the "mode"? Unfortunately, I don't have a good formulation at the moment.

```suggestion
            Select a specific subplot panel. Only allowed when used in
            :meth:`Figure.subplot` mode.
```
or 
```suggestion
            Select a specific subplot panel. Only allowed within the code
            block related to :meth:`Figure.subplot` used as a context
            manager.
```

Issue #3836, opened 2025-03-06
Ideas on refactoring the Session.virtualfile_in method


seisman:
The `Session.virtualfile_in` method is one of the most commonly used low-level API functions when wrapping GMT modules. Currently, its definition signature is as below:

https://github.com/GenericMappingTools/pygmt/blob/1791ca29547a41bef791be2df78c44620e327423/pygmt/clib/session.py#L1768-L1778

Here are some ideas to refactor the method signature:

1. Remove the `extra_arrays` parameter (Refer to 
https://github.com/GenericMappingTools/pygmt/pull/3823#issue-2880404911 for more detailed reasoning). When a module takes more than 3 columns, we can build a dict of columns instead. This is currently WIP in #3823.
2. Remove `required_z` and add `required_ncols` (or `ncols` for a shorter name). In this way, we can check if a table input contains enough number of columns. `required_z=True` is equivalent to `required_ncols=3`. This is currently WIP in #3369 
3. Rename `required_data` to `required_z`. After addressing points 1 and 2, the function definition will be:
```
 def virtualfile_in( 
     self, 
     check_kind=None, 
     data=None, 
     x=None, 
     y=None, 
     z=None, 
     ncols=2, 
     required_data=True, 
 ): 
```
I think it makes more sense to rename `required_data` to `required`. 
4. In some wrappers (currently, `plot`/`plot3d`/`text`/`legend`/`x2sys_cross`), we need to call `data_kind` to decide the data kind. In this way, we already know the data kind before calling `Session.virtualfile_in`, so it's unnecessary to call `data_kind` and check if the kind is valid in `Session.virtualfile_in`. So, what about renaming `check_kind` to `kind`? Currently, `check_kind` can take two values, `vector` and `raster`). In this way, the new `kind` parameter can accept the following values, in addition to `"vector"` and `"raster"` https://github.com/GenericMappingTools/pygmt/blob/1791ca29547a41bef791be2df78c44620e327423/pygmt/helpers/utils.py#L267
In this way, `"vector"` and `"raster"` can be thought of as two special/general kinds. And in the Sesssion.virtualfile_in method, the related codes can be rewritten to something like:
```
if kind in {"vector", "raster"}:  # Two general kinds
    valid_kinds = ("file", "arg") if required_data is False else ("file",)
    if kind == "raster":
        valid_kinds += ("grid", "image")
    elif kind == "vector":
        valid_kinds += ("empty", "matrix", "vectors", "geojson")
[...]

PR #3931, opened 2025-04-28
Session.virtualfile_in: Deprecate parameter 'required_data' to 'required' (will be removed in v0.20.0)

seisman:
Address https://github.com/GenericMappingTools/pygmt/issues/3836.

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1785,9 +1785,6 @@ class Session:
             data input.
         x/y/z : 1-D arrays or None
             x, y, and z columns as numpy arrays.
-        mincols
-            Number of minimum required columns. Default is 2 (i.e. require x and y
-            columns).
         required : bool
             Set to True when 'data' or ('x' and 'y') is required. Set to False when
             dealing with optional virtual files. Default is True.
@@ -1795,6 +1792,9 @@ class Session:
             .. versionchanged:: v0.16.0
                The parameter 'required_data' is renamed to 'required'. The parameter
                'required_data' is deprecated in v0.16.0 and will be removed in v0.20.0.
+        mincols
+            Number of minimum required columns. Default is 2 (i.e. require x and y
+            columns).
         required_z : bool
             State whether the 'z' column is required.
 
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -1789,8 +1789,8 @@ class Session:
             Number of minimum required columns. Default is 2 (i.e. require x and y
             columns).
         required : bool
-            Set to True when 'data' is required, or False when dealing with optional
-            virtual files. Default is True.
+            Set to True when 'data' or (`x` and `y`) is required . Set to False when
+            dealing with optional virtual files. Default is True.
 
             .. versionchanged:: v0.16.0
                The parameter 'required_data' is renamed to 'required'. The parameter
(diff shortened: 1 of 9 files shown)


pygmt/clib/session.py
```diff
@@ -1838,19 +1847,19 @@ def virtualfile_in(  # noqa: PLR0912
...
             z=z,
             mincols=mincols,
-            required_data=required_data,
+            required=required,
             kind=kind,
         )
 
         if check_kind:
-            valid_kinds = ("file", "arg") if required_data is False else ("file",)
+            valid_kinds = ("file", "arg") if required is False else ("file",)
```
weiji14, line 1862: I feel that `required` by itself isn't a very descriptive parmeter name. This `required_data` parameter was added in #2493 to better handle optional virtual files (via the 'arg' kind). Would a name like `accept_vfile` (default to False) or something along those lines work? Then the check would be:

```python
            valid_kinds = ("file", "arg") if accept_vfile else ("file",)
```

Or we could also just keep the `required_data` name as is? Thoughts?

pygmt/src/text.py
```diff
@@ -261,9 +261,7 @@ def text_(  # noqa: PLR0912
 
     with Session() as lib:
         with lib.virtualfile_in(
-            check_kind="vector",
-            data=textfiles or data,
-            required_data=required_data,
+            check_kind="vector", data=textfiles or data, required=required_data
```
weiji14, line 264: At L193 and L194 above. The `required_data` variable wasn't changed to `required`?

pygmt/clib/session.py
seisman: > Would a name like `optional_vfile` (default to True) or something along those lines work?

You meant `optional_vfile` default to False?

weiji14: Ah yes, I meant `optional_vfile=False` by default, but that's probably not a great name either. On second thought, let's go with the `required` parameter name then. Using just `required` in `data_kind` makes sense because we're only passing in an argument to `data`, and I guess we do validate whether `data`/`x`/`y` are not None in `virtualfile_in` so there shouldn't be any ambiguity, but need to update the docstring a bit.

pygmt/clib/session.py
```diff
@@ -1783,9 +1788,13 @@ def virtualfile_in(  # noqa: PLR0912
         mincols
             Number of minimum required columns. Default is 2 (i.e. require x and y
             columns).
-        required_data : bool
-            Set to True when 'data' is required, or False when dealing with
-            optional virtual files. [Default is True].
+        required : bool
+            Set to True when 'data' is required, or False when dealing with optional
+            virtual files. Default is True.
```
weiji14, line 1793: Based on logic in `_validate_data_input`. Feel free to reword as needed.

```suggestion
            Set to True when 'data' or (`x` and `y`) is required . Set to False when
            dealing with optional virtual files. Default is True.
```

pygmt/src/text.py
seisman: Yes, `required_data` at L193-L194 is a local variable, and `required` is not a good name for it. Perhaps rename it to `data_is_required`?

pygmt/clib/session.py
```diff
@@ -1783,9 +1788,13 @@ def virtualfile_in(  # noqa: PLR0912
         mincols
             Number of minimum required columns. Default is 2 (i.e. require x and y
             columns).
```
weiji14, line 1790: Also move this docstring to below `required`.

weiji14 (approved):
Just one other suggestion. Tests failing appear unrelated, was there an update to the `earth_relief` grids again?

PR #3932, opened 2025-04-29
Allow passing region to GMTBackendEntrypoint.open_dataset

weiji14:

Support passing in a region as a Sequence [xmin, xmax, ymin, ymax] or ISO country code to `xarray.open_dataset` when using `engine="gmt"`.

Usage:

```python
import numpy.testing as npt
import xarray as xr

da = xr.open_dataarray("@static_earth_relief.nc", engine="gmt", raster_kind="grid", region=[-52, -48, -18, -12])
assert da.sizes == {"lat": 6, "lon": 4}
npt.assert_allclose(da.lat, [-17.5, -16.5, -15.5, -14.5, -13.5, -12.5])
npt.assert_allclose(da.lon, [-51.5, -50.5, -49.5, -48.5])
```

This PR also refactors the internals of `_load_remote_dataset` to use `xr.load_dataarray(engine="gmt", ...)` instead of low-level calls to clib functions.

Extends #3919, adapted from #3673

**Preview**:

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -7,10 +7,8 @@ from collections.abc import Sequence
 from typing import Any, Literal, NamedTuple
 
 import xarray as xr
-from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import build_arg_list, kwargs_to_strings
-from pygmt.src import which
+from pygmt.helpers import kwargs_to_strings
 
 with contextlib.suppress(ImportError):
     # rioxarray is needed to register the rio accessor
@@ -581,22 +579,9 @@ def _load_remote_dataset(
         raise GMTInvalidInput(msg)
 
     fname = f"@{prefix}_{resolution}_{reg}"
-    kwdict = {"R": region, "T": {"grid": "g", "image": "i"}[dataset.kind]}
-    with Session() as lib:
-        with lib.virtualfile_out(kind=dataset.kind) as voutgrd:
-            lib.call_module(
-                module="read",
-                args=[fname, voutgrd, *build_arg_list(kwdict)],
-            )
-            grid = lib.virtualfile_to_raster(
-                kind=dataset.kind, outgrid=None, vfname=voutgrd
-            )
-
-    # Full path to the grid if not tiled grids.
-    source = which(fname, download="a") if not resinfo.tiled else None
-    # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
-    if source:
-        grid.encoding["source"] = source
+    grid = xr.load_dataarray(
+        fname, engine="gmt", raster_kind=dataset.kind, region=region
+    )
 
     # Add some metadata to the grid
     grid.attrs["description"] = dataset.description
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -500,7 +500,6 @@ datasets = {
 }
 
 
-@kwargs_to_strings(region="sequence")
 def _load_remote_dataset(
     name: str,
     prefix: str,
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -8,7 +8,6 @@ from typing import Any, Literal, NamedTuple
 
 import xarray as xr
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import kwargs_to_strings
 
 with contextlib.suppress(ImportError):
     # rioxarray is needed to register the rio accessor
--- a/pygmt/tests/test_xarray_accessor.py
+++ b/pygmt/tests/test_xarray_accessor.py
@@ -144,7 +144,7 @@ def test_xarray_accessor_grid_source_file_not_exist():
     # Registration and gtype are correct.
     assert grid.gmt.registration == GridRegistration.PIXEL
     assert grid.gmt.gtype == GridType.GEOGRAPHIC
-    # The source grid file is undefined.
+    # The source grid file is undefined for tiled grids.
     assert grid.encoding.get("source") is None
 
     # For a sliced grid, fallback to default registration and gtype, because the source
--- a/pygmt/tests/test_xarray_accessor.py
+++ b/pygmt/tests/test_xarray_accessor.py
@@ -2,6 +2,7 @@
 Test the behaviour of the GMTDataArrayAccessor class.
 """
 
+import importlib
 import sys
 from pathlib import Path
 
@@ -13,6 +14,8 @@ from pygmt.clib import __gmt_version__
 from pygmt.datasets import load_earth_relief
 from pygmt.enums import GridRegistration, GridType
 from pygmt.exceptions import GMTInvalidInput
+
+_HAS_NETCDF4 = bool(importlib.util.find_spec("netCDF4"))
 
 
 def test_xarray_accessor_gridline_cartesian():
@@ -102,6 +105,7 @@ def test_xarray_accessor_set_invalid_registration_and_gtype():
 
 
 # TODO(GMT>=6.5.0): Remove the xfail marker for GMT>=6.5.0.
+@pytest.mark.skipif(condition=not _HAS_NETCDF4, reason="netCDF4 is not installed")
 @pytest.mark.xfail(
     condition=sys.platform == "win32" and Version(__gmt_version__) < Version("6.5.0"),
     reason="Upstream bug fixed in https://github.com/GenericMappingTools/gmt/pull/7573",
@@ -119,7 +123,7 @@ def test_xarray_accessor_sliced_datacube():
             "https://github.com/pydata/xarray-data/raw/master/eraint_uvz.nc",
             download="u",
         )
-        with xr.open_dataset(fname) as dataset:
+        with xr.open_dataset(fname, engine="netcdf4") as dataset:
             grid = dataset.sel(level=500, month=1, drop=True).z
 
         assert grid.gmt.registration == GridRegistration.GRIDLINE
--- a/pygmt/tests/test_xarray_accessor.py
+++ b/pygmt/tests/test_xarray_accessor.py
@@ -24,8 +24,8 @@ def test_xarray_accessor_gridline_cartesian():
     Cartesian, gridline-registered grid.
     """
     grid = xr.load_dataarray("@test.dat.nc", engine="gmt", raster_kind="grid")
-    assert grid.gmt.registration == GridRegistration.GRIDLINE
-    assert grid.gmt.gtype == GridType.CARTESIAN
+    assert grid.gmt.registration is GridRegistration.GRIDLINE
+    assert grid.gmt.gtype is GridType.CARTESIAN
 
 
 def test_xarray_accessor_pixel_geographic():
@@ -34,8 +34,8 @@ def test_xarray_accessor_pixel_geographic():
     geographic, pixel-registered grid.
     """
     grid = xr.load_dataarray("@earth_relief_01d_p", engine="gmt", raster_kind="grid")
-    assert grid.gmt.registration == GridRegistration.PIXEL
-    assert grid.gmt.gtype == GridType.GEOGRAPHIC
+    assert grid.gmt.registration is GridRegistration.PIXEL
+    assert grid.gmt.gtype is GridType.GEOGRAPHIC
 
 
 def test_xarray_accessor_set_registration():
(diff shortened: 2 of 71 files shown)


pygmt/tests/test_xarray_accessor.py
```diff
@@ -144,7 +144,7 @@ def test_xarray_accessor_grid_source_file_not_exist():
     # Registration and gtype are correct.
     assert grid.gmt.registration == GridRegistration.PIXEL
     assert grid.gmt.gtype == GridType.GEOGRAPHIC
-    # The source grid file is undefined.
+    # The source grid file is undefined for tiled grids.
     assert grid.encoding.get("source") is None
```
weiji14, line 148: Should we keep `grid.encoding["source"]` as undefined/None for tiled grids (xref https://github.com/GenericMappingTools/pygmt/pull/3673#issuecomment-2527475167)? Or select the first tile (e.g. `S90E000.earth_relief_05m_p.nc`)? May need to update this test depending on what we decide.

pygmt/xarray/backend.py
```diff
@@ -111,7 +117,7 @@ def open_dataset(  # type: ignore[override]
                     vfname=voutfile, kind=raster_kind
                 )
                 # Add "source" encoding
-                source = which(fname=filename_or_obj)
+                source: str | list = which(fname=filename_or_obj)
                 raster.encoding["source"] = (
                     source[0] if isinstance(source, list) else source
                 )
```
weiji14, line 123: Do we actually need the `_ = raster.gmt` line at L124 to load GMTDataArray accessor info, since `lib.virtualfile_to_raster` already calls `self.read_virtualfile(vfname, kind=kind).contents.to_xarray()` which sets the registration and gtype based on the header?

https://github.com/GenericMappingTools/pygmt/blob/74de7d81e2acd50778de1a711d0f1088faa5214e/pygmt/datatypes/grid.py#L197-L201

seisman: that's a good point. please try remove it and see if everything works fine.

weiji14: There is one extra test failure at https://github.com/GenericMappingTools/pygmt/actions/runs/14743641047/job/41386753150#step:10:1049:

```
_____________ [doctest] pygmt.xarray.accessor.GMTDataArrayAccessor _____________
026     Examples
027     --------
028     For GMT's built-in remote datasets, these GMT-specific properties are automatically
029     determined and you can access them as follows:
030 
031     >>> from pygmt.datasets import load_earth_relief
032     >>> # Use the global Earth relief grid with 1 degree spacing
033     >>> grid = load_earth_relief(resolution="01d", registration="pixel")
034     >>> # See if grid uses Gridline or Pixel registration
035     >>> grid.gmt.registration
Expected:
    <GridRegistration.PIXEL: 1>
Got:
    1
```

I think this is because the _GMT_GRID_HEADER.registration property returns an `int` instead of an enum:

https://github.com/GenericMappingTools/pygmt/blob/d29303bb04a43303e27847247b8eb9db6164c1e4/pygmt/datatypes/header.py#L81-L82

and we overrode `grid.gmt.registration` with `1` instead of `<GridRegistration.PIXEL: 1>`. Should be a quick fix we can do in a separate PR.

seisman: Making the following changes should fix the issue:
https://github.com/GenericMappingTools/pygmt/blob/33fadc32693109e5f67308e7241f69cd38313106/pygmt/xarray/accessor.py#L139-L142
```python
        if (_source := self._obj.encoding.get("source")) and Path(_source).exists():
            with contextlib.suppress(ValueError):                                      
                registration, gtype = map(  # type: ignore[assignment]
                    int, grdinfo(_source, per_column="n").split()[-2:]
                )                                                                     
            self._registration = GridRegistration(registration)           
            self._gtype = GridType(gtype)
```

pygmt/tests/test_xarray_accessor.py
seisman: > Or select the first tile (e.g. `S90E000.earth_relief_05m_p.nc`)?

Sounds good.

pygmt/xarray/backend.py
```diff
@@ -71,15 +72,17 @@ class GMTBackendEntrypoint(BackendEntrypoint):
     """
 
     description = "Open raster (.grd, .nc or .tif) files in Xarray via GMT."
-    open_dataset_parameters = ("filename_or_obj", "raster_kind")
+    open_dataset_parameters = ("filename_or_obj", "raster_kind", "region")
     url = "https://pygmt.org/dev/api/generated/pygmt.GMTBackendEntrypoint.html"
 
+    @kwargs_to_strings(region="sequence")
```
seisman, line 78: I've been think if we should avoid using the `@kwargs_to_strings` decorator in new functions/methods, and instead write a new function like `seqjoin` which does exactly the same thing.

pygmt/xarray/backend.py
```diff
@@ -111,9 +117,8 @@ def open_dataset(  # type: ignore[override]
                     vfname=voutfile, kind=raster_kind
                 )
                 # Add "source" encoding
-                source = which(fname=filename_or_obj)
+                source: str | list = which(fname=filename_or_obj)
                 raster.encoding["source"] = (
                     source[0] if isinstance(source, list) else source
                 )
-                _ = raster.gmt  # Load GMTDataArray accessor information
                 return raster.to_dataset()
```
seisman, line 124: Actually, it's likely that the accessor information will be lost when converting via `to_dataset`.

pygmt/datasets/load_remote_dataset.py
```diff
@@ -581,22 +579,9 @@ def _load_remote_dataset(
         raise GMTInvalidInput(msg)
 
     fname = f"@{prefix}_{resolution}_{reg}"
```
seisman, line 581: I see a lot of error messages like:
```
Error: h [ERROR]: Tile @S90W180.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90W150.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90W120.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90W090.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90W060.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90W030.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90E000.earth_age_01m_g.nc not found!
Error: h [ERROR]: Tile @S90E030.earth_age_01m_g.nc not found!
```
This is because, in the GMT backend, we use something like `which("@earth_age_01m_g")` to get the file path, which doesn't work well for tiled grids.

weiji14: Yeah, we used to do this:

```python
    # Full path to the grid if not tiled grids.
    source = which(fname, download="a") if not resinfo.tiled else None
    # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
    if source:
        grid.encoding["source"] = source
```

i.e. only add the source for non-tiled grids, so that the accessor's `which` call doesn't report this error. I'm thinking if it's possible to either 1) silence the `which` call (does `verbose="q"` work?), or 2) add some heuristic/logic to determine whether the source is a tiled grid before calling `which` in `GMTBackendEntrypoint`

pygmt/xarray/backend.py
weiji14: It seems like enum comparison should be done using `is` instead of `==` according to https://docs.python.org/3/howto/enum.html#comparisons (see also https://stackoverflow.com/questions/25858497/should-enum-instances-be-compared-by-identity-or-equality).

```python
assert grid.gmt.registration is enums.GridRegistration.PIXEL # OK

assert 1 is enums.GridRegistration.PIXEL # AssertionError
```

Wish there was a ruff lint for this (xref https://github.com/astral-sh/ruff/discussions/11617), but until then, will need to fix it manually.

pygmt/datasets/load_remote_dataset.py
seisman: > I'm thinking if it's possible to either 1) silence the `which` call (does `verbose="q"` work?), or 2) add some heuristic/logic to determine whether the source is a tiled grid before calling `which` in `GMTBackendEntrypoint`

I think either works. Perhaps `verbose="q"` is easier?

weiji14: Done in commit 5557b3343247dd1a5312b0ff921daf6f06d56d9d

pygmt/datasets/load_remote_dataset.py
```diff
@@ -7,10 +7,8 @@
 from typing import Any, Literal, NamedTuple
 
 import xarray as xr
-from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import build_arg_list, kwargs_to_strings
-from pygmt.src import which
+from pygmt.helpers import kwargs_to_strings
```
seisman, line 11: At line 503, `@kwargs_to_strings(region="sequence")` can be removed, since the `region` parameter, either a sequence or a string, can be directly passed to `xr.load_dataarray`.

pygmt/tests/test_xarray_accessor.py
```diff
@@ -144,17 +147,25 @@ def test_xarray_accessor_grid_source_file_not_exist():
...
     sliced_grid = grid[1:3, 1:3]
-    assert sliced_grid.gmt.registration is GridRegistration.GRIDLINE
-    assert sliced_grid.gmt.gtype is GridType.CARTESIAN
-
-    # Still possible to manually set registration and gtype.
-    sliced_grid.gmt.registration = GridRegistration.PIXEL
-    sliced_grid.gmt.gtype = GridType.GEOGRAPHIC
+    assert sliced_grid.encoding["source"].endswith("S90E000.earth_relief_05m_p.nc")
     assert sliced_grid.gmt.registration is GridRegistration.PIXEL
     assert sliced_grid.gmt.gtype is GridType.GEOGRAPHIC
```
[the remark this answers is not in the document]
seisman: > Do we preserve the source encoding now even after slicing?!!

I've tried both xarray v2025.03 and v0.15 (in 2020). It seems the source encoding is always kept.

```python
In [1]: import xarray as xr

In [2]: grid = xr.load_dataarray("earth_relief_01d_g.grd")

In [3]: grid.encoding
Out[3]: 
{'zlib': True,
 'szip': False,
 'zstd': False,
 'bzip2': False,
 'blosc': False,
 'shuffle': True,
 'complevel': 9,
 'fletcher32': False,
 'contiguous': False,
 'chunksizes': (181, 181),
 'source': '/home/seisman/.gmt/server/earth/earth_relief/earth_relief_01d_g.grd',
 'original_shape': (181, 361),
 'dtype': dtype('int16'),
 '_FillValue': -32768,
 'scale_factor': 0.5}

In [4]: grid2 = grid[0:10, 0:10]

In [5]: grid2.encoding
Out[5]: 
{'zlib': True,
 'szip': False,
 'zstd': False,
 'bzip2': False,
 'blosc': False,
 'shuffle': True,
 'complevel': 9,
 'fletcher32': False,
 'contiguous': False,
 'chunksizes': (181, 181),
 'source': '/home/seisman/.gmt/server/earth/earth_relief/earth_relief_01d_g.grd',
 'original_shape': (181, 361),
 'dtype': dtype('int16'),
 '_FillValue': -32768,
 'scale_factor': 0.5}
```

> I.e. is #524 fixed? 

I think #524 is a different issue. The main point in #524 is that, for slice operations, the accessor information is lost, so we need to call `grdinfo` on the source grid to determine gtype and registration. In previous versions (before GMT_GRID was implemented), we use temporary files a lot, so the source file may not exists. I guess we need to revisit #524 and see if it can be closed.

seisman:
> Should we keep `grid.encoding["source"]` as undefined/None for tiled grids (xref [#3673 (comment)](https://github.com/GenericMappingTools/pygmt/pull/3673#issuecomment-2527475167))? Or select the first tile (e.g. `S90E000.earth_relief_05m_p.nc`)? May need to update this test depending on what we decide.

@weiji14 Perhaps we should cherry-pick commits related to these changes into a separate PR, with a title like "Store the first tile as source encoding for tiled grids" and linking to issue #524.

pygmt/xarray/backend.py
weiji14: Probably best to open a separate issue/PR for this.

pygmt/xarray/backend.py
```diff
@@ -111,9 +145,8 @@ def open_dataset(  # type: ignore[override]
                     vfname=voutfile, kind=raster_kind
                 )
                 # Add "source" encoding
-                source = which(fname=filename_or_obj)
+                source: str | list = which(fname=filename_or_obj, verbose="q")
                 raster.encoding["source"] = (
                     source[0] if isinstance(source, list) else source
```
seisman, line 150: Need to ensure tiles are sorted.
```suggestion
                    sorted(source)[0] if isinstance(source, list) else source
```

seisman (approved):
Looks good to me.

Issue #3881, opened 2025-03-29
DOC: Remove the list of aliases and 'full option link' if all functionalities are implemented?


seisman:
Taking [`pygmt.grdfill`](https://www.pygmt.org/dev/api/generated/pygmt.grdfill.html) as an example. Currently, all the options and functionalities are implemented (except `-Lp` but see #3768 for the reason). In this case, perhaps we can remove the "full option" link and the list of aliases, i.e., removing the following part:

![Image](https://github.com/user-attachments/assets/7be7a17e-ff74-42ca-a3c7-bc40db0205b8)

yvonnefroehlich:
Regarding the link to the GMT documentation: I personally used it, but not sure how often users have a look at the upstream docs. I am wondering if we can rename it to something like "GMT documentation at ..."?

seisman:
> I think the alias list can be still useful for users coming from GMT.

The alias will be incomplete and make little sense once we implement the new alias system (PR #3238) which maps multiple PyGMT parameters into one single GMT option. We can revisit the issue when we finish the new alias system.

> Regarding the link to the GMT documentation: I personally used it, but not sure how often users have a look at the upstream docs. I am wondering if we can rename it to something like "GMT documentation at ..."?

What about changing it to "This function wraps the GMT module `grdfill` and the GMT documentation is at ..."?

yvonnefroehlich:
> > Regarding the link to the GMT documentation: I personally used it, but not sure how often users have a look at the upstream docs. I am wondering if we can rename it to something like "GMT documentation at ..."?
> 
> What about changing it to "This function wraps the GMT module `grdfill` and the GMT documentation is at ..."?

Sounds good!

ezevazquez:
I personally think the alias list still adds value, particularly for GMT users who are looking for a one-to-one mapping. Even if the new alias system eventually replaces it, having both in the short term might help ease the transition.

I’d vote for keeping the alias list for now, and maybe just rewording the link to say something like > "For the complete GMT documentation, see..."

seisman:
Closed by #3944 and #3965.

PR #2798, opened 2023-11-05
POC: Add Figure.choropleth to plot choropleth maps

seisman:
**Description of proposed changes**

Add `Figure.choropleth` to plot choropleth maps, which can simplify #2796

```python
import geopandas as gpd
import pygmt

gdf = gpd.read_file("https://geodacenter.github.io/data-and-lab/data/airbnb.zip")

fig = pygmt.Figure()
pygmt.makecpt(
    cmap="acton",
    series=[gdf["population"].min(), gdf["population"].max(), 10],
    continuous=True,
    reverse=True,
)
fig.choropleth(gdf, fillcol="population", pen="0.3p,gray10")
fig.colorbar(frame=True)
fig.show()
```

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #3944, opened 2025-05-10
DOC: Improve link to the GMT documentation

seisman:
Address #3881.

### DIFF
--- a/pygmt/src/grdfill.py
+++ b/pygmt/src/grdfill.py
@@ -125,13 +125,16 @@ def grdfill(
     r"""
     Interpolate across holes in a grid.
 
+    .. note::
+
+       Wraps the GMT module ``grdfill``.
+       The GMT documentation is at :gmt-docs:`grdfill.html`.
+
     Read a grid that presumably has unfilled holes that the user wants to fill in some
     fashion. Holes are identified by NaN values but this criteria can be changed via the
     ``hole`` parameter. There are several different algorithms that can be used to
     replace the hole values. If no holes are found the original unchanged grid is
     returned.
-
-    Full option list at :gmt-docs:`grdfill.html`.
 
     {aliases}
 


seisman:
I'll apply the same changes to other wrappers if approved.

yvonnefroehlich (approved):
For me this change looks good in the docs.

seisman:
Ping @michaelgrund and @weiji14 for comments before I make changes.

Actually, I'd still like to rephrase the sentence so the links appear on one line, but I haven’t found a good solution yet.

pygmt/src/grdfill.py
```diff
@@ -125,14 +125,17 @@ def grdfill(
     r"""
     Interpolate across holes in a grid.
 
+    .. note::
+
+       Wraps the GMT module ``grdfill``.
+       The GMT documentation is at :gmt-docs:`grdfill.html`.
+
```
yvonnefroehlich, line 132: Hm. If it should fit into one line, maye we can put the link to the docs within the ``grdfill``.  
Not sure, if the link works in this way or the full URL is required. 

```suggestion
    .. note::

       Wraps the GMT module [``grdfill``](:gmt-docs:`grdfill.html`).

```

pygmt/src/grdfill.py
```diff
@@ -125,14 +125,16 @@ def grdfill(
...
+       Wraps the GMT module :gmt-docs:`grdfill <grdfill.html>`.
+
     Read a grid that presumably has unfilled holes that the user wants to fill in some
     fashion. Holes are identified by NaN values but this criteria can be changed via the
     ``hole`` parameter. There are several different algorithms that can be used to
     replace the hole values. If no holes are found the original unchanged grid is
     returned.
 
-    Full option list at :gmt-docs:`grdfill.html`.
```
weiji14, line 134: The `note::` admonition stands out a bit too much. Perhaps just reword `Full option list at ...` to `Full GMT docs at ...`, similar to @ezevazquez's suggestion at https://github.com/GenericMappingTools/pygmt/issues/3881#issuecomment-2843105418?

seisman: > The `note::` admonition stands out a bit too much.

I agree. 

> Perhaps just reword `Full option list at ...` to `Full GMT docs at ...`

I prefer to explicitly mention the name of the GMT module, since the function/method names may differ from the module names. What about the one below?

image

weiji14: > since the function/method names may differ from the module names

I thought we usually use the same method name as upstream GMT? Different ones I can find are:
- https://www.pygmt.org/v0.15.0/api/generated/pygmt.Figure.set_panel.html (upstream uses `gmt subplot set`), and that page is missing the upstream doc link btw.
- https://www.pygmt.org/dev/api/generated/pygmt.which.html#pygmt.which (upstream uses `gmtwhich`)

The names seem to follow closely enough that we don't need to repeat the name really.

seisman: We have methods like `hlines`/`vlines` and will have methods like `scatter` (#scatter), `choropleth` (#2798), which wraps `plot`. I guess we still need to link to `plot` in these methods.

weiji14: > We have methods like `hlines`/`vlines` and will have methods like `scatter` (#3602 ) and `choropleth` (#2798), which wrap `plot`. I guess we still need to link to `plot` in these methods.

For those convenience methods, then yes, it makes sense to use "Wraps GMT module ``scatter``, with full docs at :gmt-docs:`plot.html`.". But for most other ones, probably ok to use just "Full GMT docs at ..."?

seisman: OK. I've changed it to "Full GMT docs at ...".

PR #3948, opened 2025-05-12
Figure: Deprecated the private _preprocess method (will be removed in v0.20.0)

seisman:
The `Figure._preprocess` method was originally introduced in commit 4b414ce125df5ee8aeeb78a2487980603d02b531 and later extended to wrap `Figure._activate_figure` in commit 5e692855cde37f8bd09d76ac146cf70509ad4b48.

The method was intended to take `kwargs` as input, modify them, and return the result. However, in practice, the input `kwargs` are never changed. As a result, `Figure._preprocess` serves only as a thin wrapper around `Figure._activate_figure` without adding any additional functionality.

This PR replaces all calls to `Figure._preprocess` with direct calls to `Figure._activate_figure` and deprecates the `_preprocess` method, which is scheduled for removal in four releases.


pygmt/figure.py
```diff
@@ -119,12 +120,24 @@ def _activate_figure(self) -> None:
         with Session() as lib:
             lib.call_module(module="figure", args=[self._name, fmt])
 
+    # TODO(PyGMT>=v0.20.0):  Remove the _preprocess method.
```
weiji14, line 123: Technically this `_preprocess` method is private so we could just remove it :fire: But if you think people are using it, we could do with 2 release cycles instead of 4.

seisman: Yes, people are using it. 2 release cycles sounds good to me.

PR #3950, opened 2025-05-14
Store first tile as source encoding for tiled grids

weiji14:

Tiled [remote dataset](https://www.generic-mapping-tools.org/remote-datasets/) grids used to have an undefined `grid.encoding["source"]` because there was a list of grids returned from `pygmt.which`, and we did not select any one. The change here means that we set `grid.encoding["source"]` to the first grid in the list (sorted alphabetically).

This means that slicing `xarray.DataArray` tiled grids should return proper GMT Accessor `GridType` and `GridRegistration` values, because the source file can be queried using `grdinfo` to get the values.

Addresses https://github.com/GenericMappingTools/pygmt/pull/3932#issuecomment-2868773168 and #524

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -593,7 +593,7 @@ def _load_remote_dataset(
             )
 
     # Full path to the grid
-    source: str | list = which(fname, download="a", verbose="q")
+    source: str | list = which(fname, verbose="q")
     if resinfo.tiled:
         source = sorted(source)[0]  # get first grid for tiled grids
     # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -595,7 +595,7 @@ def _load_remote_dataset(
     # Full path to the grid
     source: str | list = which(fname, download="a")
     if resinfo.tiled:
-        source = source[0]  # get first grid for tiled grids
+        source = sorted(source)[0]  # get first grid for tiled grids
     # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
     if source:
         grid.encoding["source"] = source
--- a/pygmt/datasets/load_remote_dataset.py
+++ b/pygmt/datasets/load_remote_dataset.py
@@ -597,8 +597,7 @@ def _load_remote_dataset(
     if resinfo.tiled:
         source = sorted(source)[0]  # get first grid for tiled grids
     # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
-    if source:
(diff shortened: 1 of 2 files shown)


pygmt/datasets/load_remote_dataset.py
```diff
@@ -592,8 +592,10 @@ def _load_remote_dataset(
...
 
-    # Full path to the grid if not tiled grids.
-    source = which(fname, download="a") if not resinfo.tiled else None
+    # Full path to the grid
+    source: str | list = which(fname, download="a")
+    if resinfo.tiled:
+        source = sorted(source)[0]  # get first grid for tiled grids
     # Manually add source to xarray.DataArray encoding to make the GMT accessors work.
     if source:
         grid.encoding["source"] = source
```
weiji14, line 601: Could add `verbose="q"` to the `which` call here, and probably remove the `if source:` line, but doesn't really matter if it's all gonna go away after the refactor to use `xr.load_dataarray` in #3932.

seisman: I prefer to have verbose="d" added and if source removed, so that this PR is fully functional on its own.

pygmt/datasets/load_remote_dataset.py
```diff
@@ -592,11 +592,12 @@ def _load_remote_dataset(
                 kind=dataset.kind, outgrid=None, vfname=voutgrd
             )
 
-    # Full path to the grid if not tiled grids.
-    source = which(fname, download="a") if not resinfo.tiled else None
+    # Full path to the grid
+    source: str | list = which(fname, download="a", verbose="q")
```
seisman, line 596: Hmmm. Actually, this won't work, since `which("@earth_relief_03s", download="a", verbose="q")` will download all tiles, which take forever.

pygmt/tests/test_xarray_accessor.py
```diff
@@ -144,17 +147,25 @@ def test_xarray_accessor_grid_source_file_not_exist():
...
-    assert sliced_grid.gmt.gtype is GridType.CARTESIAN
-
-    # Still possible to manually set registration and gtype.
-    sliced_grid.gmt.registration = GridRegistration.PIXEL
-    sliced_grid.gmt.gtype = GridType.GEOGRAPHIC
+    assert sliced_grid.encoding["source"].endswith("S90E000.earth_relief_05m_p.nc")
     assert sliced_grid.gmt.registration is GridRegistration.PIXEL
     assert sliced_grid.gmt.gtype is GridType.GEOGRAPHIC
+
+    # For a grid that underwent mathematical operations, fallback to  default
```
seisman, line 160:
```suggestion
    # For a grid that underwent mathematical operations, fallback to default
```

seisman (approved):
Looks good except one minor edit.

PR #3961, opened 2025-05-27
Add a helper function sequence_join to join a sequence by a separator

seisman:
This PR extends the private `_parse_sequence` function in `grdclip` to a more general function `sequence_join`, which can join a 1-D or 2-D sequence by a separator.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -827,4 +827,4 @@ def sequence_join(
     if size is not None and any(len(i) not in size for i in value):
         msg = f"{errmsg['name']}Expected a sequence of {errmsg['sizes']} values."
         raise GMTInvalidInput(msg)
-    return [separator.join(str(j) for j in value[i]) for i in range(len(value))]
+    return [separator.join(str(j) for j in sub) for sub in value]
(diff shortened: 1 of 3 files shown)


pygmt/helpers/utils.py
```diff
@@ -711,3 +711,120 @@ def args_in_kwargs(args: Sequence[str], kwargs: dict[str, Any]) -> bool:
     return any(
         kwargs.get(arg) is not None and kwargs.get(arg) is not False for arg in args
     )
+
+
+def sequence_join(
```
weiji14, line 716: Keep this as a private function?

```suggestion
def _sequence_join(
```

weiji14 (approved):

pygmt/helpers/utils.py
seisman: This function is used in `grdclip.py` and will likely be used in many other functions. So, I'd prefer to make it public.

weiji14: Ok, not strong opinions, just worried we might need to refactor it again in the future (which requires more caution for public-facing functions).

Issue #3239, opened 2024-05-09
New alias system towards a more Pythonic interface


seisman:
GMT's single-letter option flags (e.g. `-B`) are difficult to read/understand, so they are not recommended for use in PyGMT. Instead, PyGMT uses long-form parameters and the PyGMT alias system is responsible for translating PyGMT long-form parameters into the corresponding short-form GMT options. The alias system was originally implemented by @leouieda in aad12e0879bc71a7dbb396298ed87d68c7d53c80 (seven years ago!), and hasn't changed much since then. The alias system has some limitations and flaws that prevent us from achieving the project goal: ["Build a Pythonic API for GMT"](https://www.pygmt.org/dev/#project-goals). Now it's time to design a new alias system. This issue reviews the current alias system and proposes a new alias system. The new alias system is implemented in #3238.

## The current alias system

Currently, the alias system looks like this:
```python
@fmt_docstrings
@use_alias(
    R="region",
    B="frame",
    J="projection",
)
@kwargs_to_strings(R="sequence")
def func(self, **kwargs):
    with Session() as lib:
        lib.call_module("basemap", args=build_arg_list(kwargs))
```
The `kwargs_to_string` decorator is responsible for converting an argument to a string. It also takes care of joining a sequence by a specific separator (e.g., converting `region=[10, 20, 30, 40]` to `region="10/20/30/40"`). The `use_alias` decorator is responsible for mapping long-form PyGMT parameters (e.g, `region`) to short-form GMT options (e.g., `R`). The short-form options are then stored in `kwargs` and `build_arg_list` (previously `build_arg_string`) convert the keyword dictionary `kwargs` to a list/string that GMT API can take.

The current alias system has some known limitations and flaws:

1. The argument is still not readable/writable.

   Since each GMT option usually has many modifiers, the argument is usually long and no tab autocompletion is possible.

   Here is an example from #1082:
   ```
   fig.logo(position="jTR+o0.3c/0.6c+w3c", box="+p1p+glightblue")
   ```
   The parameter names `position` and `box` are good, but their arguments are difficult to write/read. In #1082, some candidate solutions (dict, class or function) were proposed. Please refer to #1082 for detailed discussions.

2. Abuse of the `kwargs` parameter.

   Short-form GMT options are stored in the keyword argument `kwargs`, so it must be last parameter for all wrappers that use the alias system.

3. Can't access the *original* argument by the long-form parameter name inside the wrappers

[...]

PR #3965, opened 2025-06-03
Improve decorators @use_alias/@fmt_docstrings to list non-aliased aliases

seisman:
This is another approach to addressing the 2nd issue raised in https://github.com/GenericMappingTools/pygmt/issues/3881.

This PR improves the `@use_alias` decorator to support non-aliased aliases. Here, non-aliased aliases are marked by a trailing dash, e.g., `D="resolution"` vs `D="resolution-"`.

Non-aliased aliases mean that the alias will be shown in the alias list, but the long-form parameters won't be converted to short-form flags, so that we can still use `resolution` rather than `kwargs["D"]` in the codes.

Please note that this is just a temporary fix, and we need to find a different way to list aliases when we start to remove `@use_alias`.

Supersedes #3945.

**Preview**: https://pygmt-dev--3965.org.readthedocs.build/en/3965/api/generated/pygmt.grdfill.html

### DIFF
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -555,7 +555,15 @@ def use_alias(**aliases):
             for short_param, long_alias in aliases.items():
                 if long_alias.endswith("-"):
                     # Trailing dash means it's not aliased but should be listed.
+                    if short_param in kwargs:
+                        msg = (
+                            f"Short-form parameter ({short_param}) is not recognized. "
+                            f"Use long-form parameter(s) '{long_alias.strip('-')}' "
+                            "instead."
+                        )
+                        raise GMTInvalidInput(msg)
                     continue
+
                 if long_alias in kwargs and short_param in kwargs:
                     msg = (
                         f"Parameters in short-form ({short_param}) and "
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -452,7 +452,7 @@ def fmt_docstring(module_func):
             alias = module_func.aliases[arg]
             # Trailing dash means it's not aliased but should be listed.
             # Remove the trailing dash if it exists.
-            aliases.append(f"   - {arg} = {alias.rstrip('-')}")
+            aliases.append(f"   - {arg} = *{alias.rstrip('-')}*")
         filler_text["aliases"] = "\n".join(aliases)
 
     filler_text["table-classes"] = (
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -554,15 +554,16 @@ def use_alias(**aliases):
             """
             for short_param, long_alias in aliases.items():
                 if long_alias.endswith("-"):
+                    _long_alias = long_alias.rstrip("-")
                     # Trailing dash means it's not aliased but should be listed.
-                    _alias_list = long_alias.rstrip("-").split("/")
+                    _alias_list = _long_alias.split("/")
                     if (
                         any(_alias in kwargs for _alias in _alias_list)
                         and short_param in kwargs
                     ):  # Both long- and short- forms are given.
                         msg = (
                             f"Parameters in short-form ({short_param}) and "
-                            f"long-form ({long_alias.strip('-')}) can't coexist."
+                            f"long-form ({_long_alias}) can't coexist."
                         )
                         raise GMTInvalidInput(msg)
                     if short_param in kwargs:  # Only short-alias is given
@@ -570,11 +571,11 @@ def use_alias(**aliases):
                             msg = (
                                 f"Short-form parameter ({short_param}) is not "
                                 f"recognized. Use long-form parameter(s) "
-                                f"'{long_alias.strip('-')}' instead."
+                                f"'{_long_alias}' instead."
                             )
                             raise GMTInvalidInput(msg)
                         # If there is only one long-form parameter, use it.
-                        kwargs[long_alias] = kwargs.pop(short_param)
+                        kwargs[_long_alias] = kwargs.pop(short_param)
                     continue
 
                 if long_alias in kwargs and short_param in kwargs:
--- a/pygmt/helpers/decorators.py
+++ b/pygmt/helpers/decorators.py
@@ -555,13 +555,26 @@ def use_alias(**aliases):
             for short_param, long_alias in aliases.items():
                 if long_alias.endswith("-"):
                     # Trailing dash means it's not aliased but should be listed.
-                    if short_param in kwargs:
+                    _alias_list = long_alias.rstrip("-").split("/")
+                    if (
+                        any(_alias in kwargs for _alias in _alias_list)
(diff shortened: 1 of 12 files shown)


pygmt/helpers/decorators.py
```diff
@@ -450,7 +450,9 @@ def fmt_docstring(module_func):
         aliases.append("   :columns: 3\n")
         for arg in sorted(module_func.aliases):
             alias = module_func.aliases[arg]
-            aliases.append(f"   - {arg} = {alias}")
+            # Trailing dash means it's not aliased but should be listed.
+            # Remove the trailing dash if it exists.
+            aliases.append(f"   - {arg} = {alias.rstrip('-')}")
```
weiji14, line 455: I'm thinking if we should have some sort of hint that there is a long alias already, maybe by making the italicizing the alias? (haven't tested if the below code works).

```suggestion
            # Trailing dash means it's not aliased but should be listed.
            # Remove the trailing dash if it exists.
            aliases.append(f"   - {arg} = *{alias.rstrip('-')}*")
```

weiji14: Oh wait no, this italicizes all the aliases :sweat_smile: Maybe just leave it to keep things simple.

pygmt/helpers/decorators.py
```diff
@@ -450,7 +450,9 @@ def fmt_docstring(module_func):
         aliases.append("   :columns: 3\n")
         for arg in sorted(module_func.aliases):
             alias = module_func.aliases[arg]
-            aliases.append(f"   - {arg} = {alias}")
+            # Trailing dash means it's not aliased but should be listed.
+            # Remove the trailing dash if it exists.
+            aliases.append(f"   - {arg} = *{alias.rstrip('-')}*")
```
seisman:
```suggestion
            aliases.append(f"   - {arg} = {alias.rstrip('-')}")
```

pygmt/helpers/decorators.py
```diff
@@ -548,6 +553,9 @@ def new_module(*args, **kwargs):
             New module that parses and replaces the registered aliases.
             """
             for short_param, long_alias in aliases.items():
+                if long_alias.endswith("-"):
+                    # Trailing dash means it's not aliased but should be listed.
+                    continue
                 if long_alias in kwargs and short_param in kwargs:
                     msg = (
                         f"Parameters in short-form ({short_param}) and "
```
weiji14: Should there be some extra logic in case someone uses a short alias with a long alias that has '-' at the end? E.g. this doesn't error:

```python
from pygmt import Figure

fig = Figure()
fig.coast(
    region=[-180, 180, -80, 80],
    projection="M15c",
    frame="af",
    land="#aaaaaa",
    resolution="crude",
    D="high",  # duplicate resolution/D
    water="white",
)
```

In this case, `resolution="crude"` will be used instead of `D="high"`, but maybe best if an error is raised instead of passing silently.

seisman: In a4f87793df08e6f44bc314d671900c1881deccb9 I have added codes to raise an erorr if a short-form parameter is used, but it also means single-letter alias is no longer supported for these parameters.

```python
In [1]: from pygmt import Figure
   ...: 
   ...: fig = Figure()
   ...: fig.coast(
   ...:     region=[-180, 180, -80, 80],
   ...:     projection="M15c",
   ...:     frame="af",
   ...:     land="#aaaaaa",
   ...:     resolution="crude",
   ...:     D="high",  # duplicate resolution/D
   ...:     water="white",
   ...: )
Cell In[1], line 4
      1 from pygmt import Figure
      3 fig = Figure()
----> 4 fig.coast(
      5     region=[-180, 180, -80, 80],
      6     projection="M15c",
      7     frame="af",
      8     land="#aaaaaa",
      9     resolution="crude",
     10     D="high",  # duplicate resolution/D
     11     water="white",
     12 )

File ~/OSS/gmt/pygmt/pygmt/helpers/decorators.py:564, in use_alias.<locals>.alias_decorator.<locals>.new_module(*args, **kwargs)
    558     if short_param in kwargs:
    559         msg = (
    560             f"Short-form parameter ({short_param}) is not recognized. "
    561             f"Use long-form parameter(s) '{long_alias.strip('-')}' "
    562             "instead."
    563         )
--> 564         raise GMTInvalidInput(msg)
    565     continue
    567 if long_alias in kwargs and short_param in kwargs:

GMTInvalidInput: Short-form parameter (D) is not recognized. Use long-form parameter(s) 'resolution' instead.

In [2]: from pygmt import Figure
   ...: 
   ...: fig = Figure()
   ...: fig.coast(
   ...:     region=[-180, 180, -80, 80],
   ...:     projection="M15c",
   ...:     frame="af",
   ...:     land="#aaaaaa",
   ...:     #resolution="crude",
   ...:     D="high",  # duplicate resolution/D
   ...:     water="white",
   ...: )
Cell In[2], line 4
      1 from pygmt import Figure
      3 fig = Figure()
----> 4 fig.coast(
      5     region=[-180, 180, -80, 80],
      6     projection="M15c",
      7     frame="af",
      8     land="#aaaaaa",
      9     #resolution="crude",
     10     D="high",  # duplicate resolution/D
     11     water="white",
     12 )

File ~/OSS/gmt/pygmt/pygmt/helpers/decorators.py:564, in use_alias.<locals>.alias_decorator.<locals>.new_module(*args, **kwargs)
    558     if short_param in kwargs:
    559         msg = (
    560             f"Short-form parameter ({short_param}) is not recognized. "
    561             f"Use long-form parameter(s) '{long_alias.strip('-')}' "
    562             "instead."
    563         )
--> 564         raise GMTInvalidInput(msg)
    565     continue
    567 if long_alias in kwargs and short_param in kwargs:

GMTInvalidInput: Short-form parameter (D) is not recognized. Use long-form parameter(s) 'resolution' instead.
```

weiji14: > but it also means single-letter alias is no longer supported for these parameters.

Hmm, that's not ideal. Not sure if we want to slowly implement https://github.com/GenericMappingTools/pygmt/issues/262, given that most of the comments seem to be on keeping the single-letter options. Is there a way to refactor the logic to still allow single-letter parameters?

seisman: After 4dafc576697003fb54327ae3b3e5f2ae1b6fb4ab, the current behavior is:

- Raise an error if both long and short parameters are given
- Raise an error if short parameter is given but it's aliased to multiple parameter. For example, in `grdfill`, `-A` is aliased to `constantfill/gridfill/neighborfill/splinefill/mode`. In the current situation, `kwargs["A"]` is always set with 
```
    kwargs["A"] = (
        _parse_fill_mode(constantfill, gridfill, neighborfill, splinefill)
        if mode is None
        else mode
    )
```
we may have a better solution after #3239.
- If short parameter is given and it's aliased to one long-form parameter, e.g., `-D` is aliased to `resolution` in `coast`, the value of the short parameter will be used

Issue #3984, opened 2025-06-20
Standardized error messages for invalid arguments/parameters


seisman:
For invalid arguments, we usually raise error messages like:
https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/figure.py#L497-L500

https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/datasets/samples.py#L349

https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/datasets/earth_relief.py#L157-L160

There are also error messages about required parameters or invalid combinations of parameters, e.g., 
https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/src/coast.py#L208-L213
https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/src/dimfilter.py#L140-L145
https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/src/filter1d.py#L114-L116

https://github.com/GenericMappingTools/pygmt/blob/c156826133b98d71880bdb28a39cf02dec32856e/pygmt/src/grdclip.py#L109-L114

So, it's better to centralize and standardize the error messasges. As far as I can see, there are some options:

### Option 1: Multiple helper functions
We may define many helper functions like:
```
def errmsg_invalid_arguments(name, value, valid_values=None):
    msg = f"Invalid value for parameter {name}: '{value}'." 
    if valid_values:
        msg += f" Valid values are {valid_values}'.".  
    return msg
```
### Option 2: Single helper functions
```
def errmsg(type, *args, **kwargs):
    match type:
        case "invalid_values":
            msg = f"Invalid value for parameter {name}: '{value}'." 
            if kwargs.get("valid_values"):
                msg += f" Valid values are {kwargs.get('valid_values')}'.".  
        case "invalid_parameters":
             ...
    return msg
```

### Option 3: Custom exception with custom error messages
```
class InvalidValueError(ValueError):
    def __init__(self, arg, valid_values=None):
        message = f"Invalid value: {arg!r}"
        if valid_values is not None:
            message += f". Valid values are: {valid_values!r}"
        super().__init__(message)
```

[...]

Issue #3707, opened 2024-12-23
Add new exceptions for invalid inputs and deprecated GMTInvalidInput exception


seisman:
Currently, we have several exceptions defined in https://github.com/GenericMappingTools/pygmt/blob/main/pygmt/exceptions.py.

Below is a simple command to count how often each exception is raised in the codebase (The `tests` directory and the `exceptions.py` file are excluded).
```bash
for e in \
    GMTError \
    GMTOSError \
    GMTCLibError \
    GMTCLibNotFoundError \
    GMTCLibNoSessionError \
    GMTVersionError \
    GMTInvalidInput \
    GMTImageComparisonFailure; 
do
    n=$(grep -R --exclude-dir=tests --exclude="exceptions.py" "raise $e" pygmt | wc -l)
    echo "$e: $n" 
done
```
the output is:
```
GMTError: 1
GMTOSError: 1
GMTCLibError: 16
GMTCLibNotFoundError: 1
GMTCLibNoSessionError: 1
GMTVersionError: 1
GMTInvalidInput: 87
GMTImageComparisonFailure: 1
```
Most exceptions are about the GMT C library, and is only used once or a few times. But the `GMTInvalidInput` exception is used 87 times and will definitely increase as we wrap more modules.

The `GMTInvalidInput` exception is general and doesn't tell exactly what's happening. I haven't checked the source codes carefully, but I feel the exceptions can be grouped into a few categories:

1. A required parameter is not given
2. Given parameters have conflicts 
3. A parameter is given with an incorrect type
4. A parameter is given with invalid values
5. Maybe more

I think we should have more accurate exceptions, e.g., 

- `GMTParameterError`: for cases 1 and 2, maybe having more fine-grain exceptions like `GMTRequiredParameterError`/`GMTConflictParameterError`
- `GMTTypeError`: for case 3, similar to the built-in `TypeError`
- `GMTValueError`: for case 4, similar to the built-in `ValueError`.

The exception may even start with `PYGMT`, not `GMT`, since these are not directly related to GMT.

weiji14:
Does GMT C have different Error types or enum variants that we should try to match? I found https://github.com/GenericMappingTools/gmt/blob/1ed5e1bf4be46714e4124f9c1935e5bb14a6fb99/src/gmt_error_codes.h#L32, but unsure if there are some other ones. 

[...]

PR #3696, opened 2024-12-18
GMTDataArrayAccessor: Support passing values using enums GridRegistration and GridType for grid registration and type

seisman:
**Description of proposed changes**

This PR improves the `GMTDataArrayAccessor` by allowing setting grid registration and gtype using more readable enums, rather than numeric values.

Refer to https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2271046956 for the initial thoughts but implemented using `enums` as proposed in https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2272010810.

This PR does a few things:

1. 8c7f8f676093546e7ad4b41c84899bd1c4891dfd: Add enums `GridReg`/`GridType`.
2. 94086f70a27fa1faba04f184631b4e6ca621b154: Update `GMTDataArrayAccessor` to support passing enum values (e.g., `grid.gmt.registration=GridReg.PIXEL`, not `grid.gmt.registration=1`).
3. 8dd3d5f34eee1c32ac3d6e83c039cd25a621e040: Update the existing tests to check grid registration and gtype using enums (e.g., `assert grid.gmt.registration==GridReg.PIXEL`, not `assert grid.gmt.registration==1`
4. fd099b084ba10178b2eb815ee5b7543c0f70065c, f0b30ab63c7bbf92e91fa8ffa2ebc1c581e5933d: Update the tests in `test_accessor.py`

~~The plan is to cherry-pick the commit 8c7f8f676093546e7ad4b41c84899bd1c4891dfd into a separate PR and keep the other commits in this PR, so that we can have two separate entries in the release changelog.~~ [Done in #3693]

Please note that:

1. Changes in this PR are backward-compatible.
2. String-type values (e.g., `grid.gmt.registration = "gridline"`) initially proposed in https://github.com/GenericMappingTools/pygmt/issues/499#issuecomment-2271046956 are not supported in this PR. It's simple to support string-type values, but I feel it's not necessary (see the reverted changes in https://github.com/GenericMappingTools/pygmt/pull/3696/commits/259771abe92166262532d290a2bf26c14fb7b7b4). 

**TODO**:

- [x] Add enums `GridReg`/`GridType` to API docs [in #3693]

**Preview**: https://pygmt-dev--3696.org.readthedocs.build/en/3696/api/generated/pygmt.GMTDataArrayAccessor.html

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
[...]

PR #3985, opened 2025-06-20
BREAKING: Raise GMTValueError exception for invalid values. Previously raise GMTInvalidInput

seisman:
Add new exception GMTValueError. The error message will be like

> Invalid value: 'xxx'. Expected one of: "a", "b", 0, 1. Detailed reasons explaining why it's invalid.

Address #3984 and #3707.

### DIFF
--- a/pygmt/exceptions.py
+++ b/pygmt/exceptions.py
@@ -4,6 +4,9 @@ Custom exception types used throughout the library.
 All exceptions derive from GMTError.
 """
 
+from collections.abc import Iterable
+from typing import Any
+
 
 class GMTError(Exception):
     """
@@ -51,3 +54,57 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
     """
     Raised when a comparison between two images fails.
     """
+
+
+class GMTValueError(GMTError):
+    """
+    Raised when an invalid value is passed to a function/method.
+
+    Parameters
+    ----------
+    value
+        The invalid value.
+    description
+        The description of the value.
+    choices
+        The valid choices for the value.
+    reason
+        The detailed reason why the value is invalid.
+
+    Examples
+    --------
+    >>> raise GMTValueError("invalid")
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid value: 'invalid'.
+    >>> raise GMTValueError("invalid", description="constant name")
+    Traceback (most recent call last):
+    ...
+    pygmt...GMTValueError: Invalid constant name: 'invalid'.
+    >>> raise GMTValueError("invalid", choices=["a", "b", 1, 2])
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid value: 'invalid'. Expected one of: 'a', 'b', 1, 2.
+    >>> raise GMTValueError("invalid", choices=["a", 0, True, False, None])
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid .... Expected one of: 'a', 0, True, False, None.
+    >>> raise GMTValueError("invalid", reason="Explain why it's invalid.")
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid .... Explain why it's invalid.
+    """
+
+    def __init__(
+        self,
+        value: Any,
+        description: str = "value",
+        choices: Iterable[Any] | None = None,
+        reason: str | None = None,
+    ):
+        msg = f"Invalid {description}: {value!r}."
+        if choices:
+            msg += f" Expected one of: {', '.join(repr(c) for c in choices)}."
+        if reason:
+            msg += f" {reason}"
+        super().__init__(msg)
--- a/pygmt/exceptions.py
+++ b/pygmt/exceptions.py
@@ -89,11 +89,17 @@ class GMTValueError(GMTError):
     Traceback (most recent call last):
         ...
     pygmt...GMTValueError: Invalid .... Expected one of: 'a', 0, True, False, None.
+
+    >>> from pygmt.enums import GridType
+    >>> raise GMTValueError("invalid", choices=GridType)
+    Traceback (most recent call last):
+        ...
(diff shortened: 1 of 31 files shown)


pygmt/exceptions.py
```diff
@@ -51,3 +54,57 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
     """
     Raised when a comparison between two images fails.
     """
+
+
+class GMTValueError(GMTError):
```
weiji14: Should this be subclassed from both `GMTError` and `ValueError`?

pygmt/exceptions.py
```diff
@@ -51,3 +54,57 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
...
+    >>> raise GMTValueError("invalid", reason="Explain why it's invalid.")
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid .... Explain why it's invalid.
+    """
+
+    def __init__(
+        self,
+        value: Any,
+        description: str = "value",
```
weiji14: From your changes, it seems like the `value` parameter is used in a positional-only way. Do you want to enforce it with https://peps.python.org/pep-0570/#syntax-and-semantics, or still let `value` be a 'positional_or_keyword_parameters'?

```suggestion
        value: Any,
        /,
        description: str = "value",
```

pygmt/exceptions.py
```diff
@@ -51,3 +54,57 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
...
+    pygmt...GMTValueError: Invalid value: 'invalid'. Expected one of: 'a', 'b', 1, 2.
+    >>> raise GMTValueError("invalid", choices=["a", 0, True, False, None])
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid .... Expected one of: 'a', 0, True, False, None.
+    >>> raise GMTValueError("invalid", reason="Explain why it's invalid.")
+    Traceback (most recent call last):
+        ...
+    pygmt...GMTValueError: Invalid .... Explain why it's invalid.
+    """
```
weiji14: Include another test where the input to `choices` is an enum like `GridRegistration` or `GridType`?

pygmt/xarray/accessor.py
```diff
@@ -180,11 +180,9 @@ def registration(self) -> GridRegistration:
     def registration(self, value: GridRegistration | int):
         # TODO(Python>=3.12): Simplify to `if value not in GridRegistration`.
         if value not in GridRegistration.__members__.values():
-            msg = (
-                f"Invalid grid registration: '{value}'. Should be either "
-                "GridRegistration.GRIDLINE (0) or GridRegistration.PIXEL (1)."
+            raise GMTValueError(
+                value, description="grid registration", choices=GridRegistration
```
weiji14, line 184: Highlighting this for the above comment, to check that enums like `GridRegistration` are formatted properly. I.e. the error should still print "GridRegistration.GRIDLINE (0) or GridRegistration.PIXEL (1)" per https://github.com/GenericMappingTools/pygmt/pull/3696#discussion_r1948299336

weiji14 (commented):
Thinking if this should be marked as a breaking change, because the exception type changes from `GMTInvalidInput` to `GMTValueError`, and code like `try: ... except GMTInvalidInput: ...` would break. But unsure if users even have code like that which matches on GMTInvalidInput.

Other than that, some minor suggestions/questions below.

pygmt/exceptions.py
```diff
@@ -51,3 +54,64 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
...
+        /,
+        description: str = "value",
+        choices: Iterable[Any] | None = None,
+        reason: str | None = None,
+    ):
+        msg = f"Invalid {description}: {value!r}."
+        if choices:
+            msg += f" Expected one of: {', '.join(repr(c) for c in choices)}."
+        if reason:
+            msg += f" {reason}"
```
weiji14, line 116: These lines would probably be a good use of [t-strings](https://realpython.com/python-t-strings/) (Template strings, [PEP 750](https://peps.python.org/pep-0750/)), but it is for Python 3.14+

pygmt/exceptions.py
```diff
@@ -51,3 +54,64 @@ class GMTImageComparisonFailure(AssertionError):  # noqa: N818
...
+    >>> raise GMTValueError("invalid", choices=["a", 0, True, False, None])
+    Traceback (most recent call last):
+        ...
+    pygmt.exceptions.GMTValueError: Invalid value: 'invalid'. Expected one of: 'a', 0, True, False, None.
+
+    >>> from pygmt.enums import GridType
+    >>> raise GMTValueError("invalid", choices=GridType)
+    Traceback (most recent call last):
+        ...
+    pygmt.exceptions.GMTValueError: Invalid value: 'invalid'. Expected one of: <GridType.CARTESIAN: 0>, <GridType.GEOGRAPHIC: 1>.
```
weiji14, line 97: Error message is slightly different from the previous formatting `GridRegistration.GRIDLINE (0) or GridRegistration.PIXEL (1)`, but hopefully ok.

pygmt/exceptions.py
seisman: Done in 16c15c977328882af4e66bca5ce51abe84619b2a.

pygmt/exceptions.py
seisman: `<GridType.CARTESIAN: 0>` is from the `repr()` output:
```
>>> from pygmt.enums import GridType

>>> repr(GridType.CARTESIAN)
<GridType.CARTESIAN: 0>
```
So I think it's better than previously custom string format.

seisman:
> Thinking if this should be marked as a breaking change, because the exception type changes from `GMTInvalidInput` to `GMTValueError`, and code like `try: ... except GMTInvalidInput: ...` would break. But unsure if users even have code like that which matches on GMTInvalidInput.

I've updated the PR title to indicate this is a breaking change.

weiji14 (approved):
> I've updated the PR title to indicate this is a breaking change.

Breaking, not breadking :laughing:

PR #3986, opened 2025-06-21
New Alias System: Add the private _to_string function

seisman:


pygmt/alias.py
```diff
@@ -0,0 +1,134 @@
...
+                pass
+            case True:
+                value = value[0]
+            case Mapping():
+                if value not in mapping and value not in mapping.values():
+                    raise GMTValueError(
+                        value,
+                        description="value for parameter {name!r}" if name else "value",
+                        choices=mapping.keys(),
+                    )
```
seisman: `mapping=True` means using the first letter of a long-form argument as its short-form argument. For example, `resolution="high"` will be mapped to `resolution="h"`. If we don't support `mapping=True`, lines 112-123 can be simplified to a few lines:
```python
                if value not in mapping and value not in mapping.values():
                    raise GMTValueError(
                        value,
                        description="value for parameter {name!r}" if name else "value",
                        choices=mapping.keys(),
                    )
```
but without `mapping=True`, we need to write something like
```
mapping={"full": "f", "high": "h", "intermediate": "i", "low": "l", "crude": "c"}
```
when calling the `_to_string` function.

weiji14: I think better to be explicit (use the full mapping {"key": "value"} dict), and just have one way to parse things. We can use the first letter for `resolution`, but for others like `interpolation`/[`-n`](https://docs.generic-mapping-tools.org/6.5/gmt.html#n-full), where the mapping is `{"b-spline": "b", "bicubic": "c", "bilinear": "l", ...}`, it doesn't work.

pygmt/alias.py
```diff
@@ -0,0 +1,124 @@
+"""
+The PyGMT alias system to convert PyGMT long-form arguments to GMT's short-form.
```
weiji14:
```suggestion
The PyGMT alias system to convert PyGMT's long-form arguments to GMT's short-form.
```

pygmt/alias.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _to_string(
+    value: Any,
+    prefix: str = "",  # Default to an empty string to simplify the code logic.
+    mapping: Mapping | None = None,
+    separator: Literal["/", ","] | None = None,
+    size: int | Sequence[int] | None = None,
+    ndim: int = 1,
+    name: str | None = None,
```
weiji14, line 19: Missing docstring for these three parameters?

pygmt/alias.py
seisman: I've removed `mapping=True` in commit c499aab2c1274b1c9b7251439548aa0028c68495.

pygmt/alias.py
seisman: Added in 232ff52e291a03426be72b86d42ab3cd56b19e35.

seisman:
> This looks ok to me, was there anything else you wanted to add?

Need to wait for #3985

PR #3993, opened 2025-07-10
New Alias System: Add the Alias class

seisman:
This PR adds the `Alias` class, which wraps the _to_string function and is responsible for converting any value into a string or a list of strings.

It's the basis of the new alias system  (#4000) and class-like parameters (see #3995). So, please review PRs #3995 and #4000 first.

### DIFF
--- a/pygmt/alias.py
+++ b/pygmt/alias.py
@@ -181,13 +181,8 @@ class Alias:
         size: int | Sequence[int] | None = None,
         ndim: int = 1,
     ):
-        self.value = value
         self.name = name
         self.prefix = prefix
-        self.mapping = mapping
-        self.separator = separator
-        self.size = size
-        self.ndim = ndim
         self._value = _to_string(
             value=value,
             name=name,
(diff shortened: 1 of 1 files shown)


pygmt/alias.py
```diff
@@ -131,3 +132,66 @@ def _to_string(
     # "prefix" and "mapping" are ignored. We can enable them when needed.
     _value = sequence_join(value, separator=separator, size=size, ndim=ndim, name=name)
     return f"{prefix}{_value}"
+
+
+@dataclasses.dataclass
```
seisman: Since it's just a simple class with one property, I'm wondering if we should avoid using `dataclass` and instead write it like:
```
class Alias:
    def __init__(value, name=None, prefix="", ....):
        self.value = value
        self.name = name
        self.prefix = prefix
        ...
        self._value = _to_string(
            value=self.value,
            name=self.name,
            prefix=self.prefix,
            mapping=self.mapping,
            separator=self.separator,
            size=self.size,
            ndim=self.ndim,
        )
```

seisman: The plain class is about 30% faster than dataclass. Considering that the Alias class is used extensively in the project, I've rewritten it in 4cdef8b62bf8d51aed25e5dead727ab7e5975771.

Issue #2896, opened 2023-12-19
Refactor the function definitions to better support type hints and auto-completion


seisman:
## Current style of function definitions

Currently, most of our functions/methods are defined like this:
```python
@use_alias(                                                                     
    C="per_column",                                                             
    I="spacing",                                                                
    T="nearest_multiple",                                                       
    V="verbose",                                                                
    a="aspatial",                                                               
    f="coltypes",                                                               
    i="incols",                                                                 
    r="registration",                                                           
)                                                                               
def info(data, **kwargs):  
    ...
```
in which `**kwargs` can be any parameter name, and PyGMT will convert the recognized long-form aliases (e.g., `verbose`) to the short-form flags (e.g., `-V`). Using `**kwargs` makes the function definition short and easy to maintain.

## Auto-completion issues

Using `**kwargs` in function definitions makes it difficult to support tab auto-completion in text editors/IDEs/Notebooks, as reported in #1203.

In issue #1203, I initially proposed to refactor the function definition to something like below so that PyGMT can provide tab auto-completion in any text editors/IDEs/Notebooks.
```python
@use_alias(
    C="per_column",
    I="spacing",
    T="nearest_multiple",
    V="verbose",
    a="aspatial",
    f="coltypes",
    r="registration",
)
def info(
    table,
    per_column=False,
    spacing=None,
    nearest_multiple=None,
    verbose=None,
    aspatial=None,
    coltypes=None,
    registration=None,
    **kwargs
):
```
The above proposal was rejected (https://github.com/GenericMappingTools/pygmt/issues/1203#issuecomment-817257594) because it means we have to duplicate the aliases in three places (the `use_alias` decorator, the function definition, and the docstrings). Then, in PR #1282, @maxrjones improved the `use_alias` decorator to automatically insert the long-form aliases to the function's `__signature__` (https://github.com/GenericMappingTools/pygmt/issues/1203#issuecomment-817378035), which makes it possible to support auto-completion in Jupyter Notebooks and IPython. 

However, the implementation in PR #1282 has some known limitations:

1. Auto-completion doesn't work in VSCode (https://github.com/GenericMappingTools/pygmt/pull/1282#issuecomment-846493716) [I don't know the details, but I guess it's because, since `__signature__` is dynamically updated, so a Python language server is needed. Jupyter Notebooks has a Python "kernel" but VSCode or other text editors are just an editor).
[...]

Issue #2831, opened 2023-11-26
Higher-level plotting methods for scale bar, map direction rose and magnetic rose?


seisman:
The [basemap](https://docs.generic-mapping-tools.org/6.4/basemap.html) and [coast](https://docs.generic-mapping-tools.org/6.4/coast.html) modules provide the exactly same options for  adding scale bars (`-L`), direction rose (`-Td`) and magnetic rose (`-Tm`) to maps.

I think it's more intuitive if we can provide three different higher-level plotting methods, e.g., `Figure.scale_bar`, `Figure.direction_rose`, `Figure.magnatic_rose`?

I'll take the `Figure.scale_bar` as an example. The `-L` full syntax is:
```
-L[g|j|J|n|x]refpoint+wlength[e|f|k|M|n|u][+aalign][+c[[slon/]slat]][+f][+jjustify][+l[label]][+odx[/dy]][+u][+v]
```
and the function definition will be like:
```
Figure.scale_bar(
   location=None,  # [g|j|J|n|x]refpoint
   length=None,  # +w
   label_alignment=None,  # +a
   scale_latitude=None, # +c
   scale_longitude=None,  # +c
   fancy=False,  # +f or style="plain" or "fancy"
   justification=None, # +j
   label=None,  # +l
   offset=None, # +o 
   unit=None, # +u
   vertical=None, # +v 
)
```
Of course, we also need to decide what to do with the `box` parameter.

Thoughts?

seisman:
We have decided to implement the three methods separately. The progress will be tracked in #4303, #4302, #4268.

PR #4000, opened 2025-07-16
New Alias System: Add the AliasSystem class

seisman:
This PR implements the `AliasSystem` class for the new alias system proposed in #3239.

As mentioned in #3239, the new alias system has the following pros and cons:

**pros**:

- The new and old alias systems can coexist, so we don't have to migrate all wrappers in a single PR.
- Allow building a GMT option from multiple PyGMT parameters (More Pythonic)
- No abuse of `kwargs` anymore
- Access the original argument by parameter name, not by dict lookup like kwargs.get("S") (Maybe faster)
- Autocompletion of all function parameters after #2896.
- Autocompletion for long-form arguments if we add type hints.
- Backward compatible with single-letter parameters #262

Cons:

- Big refactors may introduce new bugs. [We can always fix them if any.]
- The placeholder `{aliases}` in docstrings is not supported in the new alias system, since we can't access the local variables in a decorator. So we need to manually edit the `{aliases}` in docstrings.

### DIFF
--- a/pygmt/alias.py
+++ b/pygmt/alias.py
@@ -280,23 +280,38 @@ class AliasSystem:
         for short_param, value in kwargs.items():
             # Long-form parameters exist.
             if aliases := self.aliasdict.get(short_param):
-                long_params = ", ".join(
-                    [repr(alias.name) for alias in aliases]
-                    if isinstance(aliases, Sequence)
-                    else [repr(aliases.name)]
-                )
+                if not isinstance(aliases, Sequence):  # Single Alias object.
+                    _params = repr(aliases.name)
+                    _msg_long_params = f"Use long-form parameter {_params} instead."
+                else:  # Sequence of Alias objects.
+                    _params = ", ".join(
+                        [repr(alias.name) for alias in aliases if not alias.prefix]
+                    )
+                    _modifiers = ", ".join(
+                        [
+                            repr(alias.name) + f" ({alias.prefix})"
+                            for alias in aliases
+                            if alias.prefix
+                        ]
+                    )
+                    _msg_long_params = (
+                        f"Use long-form parameters {_params}, "
+                        f"with optional parameters {_modifiers} instead."
+                    )
+
                 # Long-form parameters are already specified.
                 if short_param in self.kwdict:
                     msg = (
-                        f"Parameter in short-form {short_param!r} conflicts with "
-                        f"long-form parameter {long_params}."
+                        f"Short-form parameter {short_param!r} conflicts with "
+                        "long-form parameters and is not recommended. "
+                        f"{_msg_long_params}"
                     )
                     raise GMTInvalidInput(msg)
 
                 # Long-form parameters are not specified.
                 msg = (
                     f"Short-form parameter {short_param!r} is not recommended. "
-                    f"Use long-form parameter {long_params} instead."
+                    f"{_msg_long_params}"
                 )
                 warnings.warn(msg, category=SyntaxWarning, stacklevel=2)
 
--- a/pygmt/alias.py
+++ b/pygmt/alias.py
@@ -3,6 +3,7 @@ The PyGMT alias system to convert PyGMT's long-form arguments to GMT's short-for
 """
 
 import dataclasses
+from collections import defaultdict
 from collections.abc import Mapping, Sequence
 from typing import Any, Literal
 
@@ -195,3 +196,70 @@ class Alias:
             size=self.size,
             ndim=self.ndim,
         )
+
+
+class AliasSystem:
+    """
+    Alias system for converting long-form PyGMT parameters to GMT short-form options.
+
+    This class is initialized with keyword arguments, where each key is a GMT option
+    flag, and the corresponding value is an ``Alias`` object or a list of ``Alias``
+    objects.
+
+    The class provides the ``kwdict`` attribute, which is a dictionary mapping each GMT
+    option flag to its current value. The value can be a string or a list of strings.
+    This keyword dictionary can then be passed to the ``build_arg_list`` function.
+
+    Examples
+    --------
+    >>> from pygmt.alias import Alias, AliasSystem
+    >>> from pygmt.helpers import build_arg_list
+    >>>
+    >>> def func(
+    ...     par0, par1=None, par2=None, frame=False, repeat=None, panel=None, **kwargs
+    ... ):
+    ...     alias = AliasSystem(
+    ...         A=[
+    ...             Alias(par1),
+    ...             Alias(par2, prefix="+o", separator="/"),
+    ...         ],
+    ...         B=Alias(frame),
+    ...         D=Alias(repeat),
+    ...         c=Alias(panel, separator=","),
+    ...     )
+    ...     return build_arg_list(alias.kwdict | kwargs)
+    >>> func(
+    ...     "infile",
+    ...     par1="mytext",
+    ...     par2=(12, 12),
+    ...     frame=True,
+    ...     repeat=[1, 2, 3],
+    ...     panel=(1, 2),
+    ...     J="X10c/10c",
+    ... )
+    ['-Amytext+o12/12', '-B', '-D1', '-D2', '-D3', '-JX10c/10c', '-c1,2']
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Initialize the alias system and create the keyword dictionary that stores the
+        current parameter values.
+        """
+        # Keyword dictionary with an empty string as default value.
+        self.kwdict = defaultdict(str)
+
+        for option, aliases in kwargs.items():
+            if not is_nonstr_iter(aliases):  # A single alias.
+                self.kwdict[option] = aliases._value
+                continue
+
+            for alias in aliases:  # List of aliases.
+                match alias._value:
+                    case None:
+                        continue
+                    case str():
+                        self.kwdict[option] += alias._value
+                    case list():
+                        # A repeatable option should have only one alias, so break.
+                        self.kwdict[option] = alias._value
+                        break
--- a/pygmt/alias.py
+++ b/pygmt/alias.py
@@ -3,11 +3,12 @@ The PyGMT alias system to convert PyGMT's long-form arguments to GMT's short-for
 """
 
 import dataclasses
+import warnings
 from collections import defaultdict
 from collections.abc import Mapping, Sequence
 from typing import Any, Literal
 
-from pygmt.exceptions import GMTValueError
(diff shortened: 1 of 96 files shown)


pygmt/src/basemap.py
```diff
@@ -2,17 +2,15 @@
 basemap - Plot base maps and frames.
 """
 
+from pygmt.alias import Alias, AliasSystem
```
seisman: Changes in `basemap.py` is only meant for proof of concept. I plan to revert the changes in `basemap.py` and open separate PRs for it.

seisman:
There is still one failing test `test_plot_datetime`, in which 
```
region=[                                                                        
    np.array("2010-01-01T00:00:00", dtype=np.datetime64),                       
    pd.Timestamp("2020-01-01"),                                                 
    0,                                                                        
    10,                                                                         
],
```
is converted to `"-R2010-01-01T00:00:00/2020-01-01 00:00:00/0/10"`. For such cases, we need to add a workaround like
https://github.com/GenericMappingTools/pygmt/blob/25febab8a7f2ee6f46a4e7ae709feb34314545bf/pygmt/helpers/decorators.py#L767-L776
However, since arguments `label=["label A", "label B", "label C"]` are also valid and are expected to be converted to a string like `-Llabel A/label B/label C` (e.g., in `Figure.ternary`), the workaround above won't work in such cases. 

Since, it's likely only the `region` parameter can have datetime-like arguments, we can add a parameter like `allow_datetime` (default to `False`) to the `_to_string` function and the `Alias` class. Anyway, these are just rare cases and can be fixed later. So let's focus on the new alias system here.

pygmt/tests/test_alias_system.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_alias_system_multiple_aliases_short_form():
+    """
+    Test that the alias system works with multiple aliases when short-form parameters
+    are used.
+    """
+    # Long-form exists but is not given, and short-form is given.
+    with pytest.warns(
+        SyntaxWarning,
+        match="Short-form parameter 'U' is not recommended. Use long-form parameter 'label', 'text' instead.",
```
weiji14: This error message might be confusing. So `label` is an alias for `-U`, and `text` is an alias for the `+t` modifier. Could we improve the message to something like:

```
Short-form parameter 'U' is not recommended. Use long-form parameter 'label' instead, with optional 'text' (+t) modifier.
```

Or something that makes it clearer that `U=label` and `+t=text`.

weiji14 (commented):
Thanks @seisman, I like that this allows us to migrate to the new Alias system gradually!

> However, since arguments label=["label A", "label B", "label C"] are also valid and are expected to be converted to a string like -Llabel A/label B/label C (e.g., in Figure.ternary), the workaround above won't work in such cases.
> 
> Since, it's likely only the region parameter can have datetime-like arguments, we can add a parameter like allow_datetime (default to False) to the _to_string function and the Alias class. Anyway, these are just rare cases and can be fixed later. So let's focus on the new alias system here.

Agree that we can handle these rare (but important) cases later. It might be that we need to pull out some of the logic in the `kwargs_to_strings` decorator function to share it with this AliasSystem class.

pygmt/tests/test_alias_system.py
seisman: I've improved the error message in 76f7c6ee125aad2ec9ea5254efc97e86c1a94286.

Now the messages are:

> Short-form parameter 'U' is not recommended. Use long-form parameters 'label', with optional parameters 'text' (+t) instead.

and
> Short-form parameter 'U' conflicts with long-form parameters and is not recommended. Use long-form parameters 'label', with optional parameters 'text' (+t) instead.

seisman:
> * The placeholder `{aliases}` in docstrings is not supported in the new alias system, since we can't access the local variables in a decorator. So we need to manually edit the `{aliases}` in docstrings.

Since we need to manually maintain the aliases in docstrings, we may want to discuss how the docstrings look like, especially for option flags that are aliased to multiple parameters. Let me take `-U` as an example. The GMT CLI syntax is `-U[label][+jjustify][+odx[/dy]][+ttext]`, then the docstrings can be:

**U**=label, justify (**+j**), offset (**+o**), text (**+t**)

or 

**U**=label, **+j**: justify, **+o**: offset, **+t**: text

Maybe we should open a separate issue to discuss the details and maybe also add a documentation explaining how the new alias system works and how to read and understand the aliases in docstrings.

weiji14:
> Maybe we should open a separate issue to discuss the details and maybe also add a documentation explaining how the new alias system works and how to read and understand the aliases in docstrings.

Yeah, please open a separate issue. I also want to know how we might want to handle conflicting modifiers, e.g. the `+o` offset modifier is used by `-L`, `-T`, `-U` in [basemap](https://docs.generic-mapping-tools.org/6.5/basemap.html)...

seisman:
> I also want to know how we might want to handle conflicting modifiers, e.g. the `+o` offset modifier is used by `-L`, `-T`, `-U` in [basemap](https://docs.generic-mapping-tools.org/6.5/basemap.html)...

Actually, the idea is to have standalone plotting methods for `-L`, `-T` (i.e., #2831).

pygmt/src/basemap.py
```diff
@@ -83,5 +81,12 @@ def basemap(self, **kwargs):
     {transparency}
     """
     self._activate_figure()
+
+    alias = AliasSystem(
+        R=Alias(region, name="region", separator="/", size=[4, 6]),
+        J=Alias(projection, name="projection"),
+        B=Alias(frame, name="frame"),
+    ).update(kwargs)
```
seisman: Actually, I'm still debating if we should change these lines to:
```suggestion
    kwdict = AliasSystem(
        R=Alias(region, name="region", separator="/", size=[4, 6]),
        J=Alias(projection, name="projection"),
        B=Alias(frame, name="frame"),
    ).update(kwargs).kwdict
```
since only `alias.kwdict` is used below.

weiji14: I did wonder if we could turn the `AliasSystem` class into a subclass of [`collections.UserDict`](https://docs.python.org/3/library/collections.html#userdict-objects) or [`collections.OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict) (or maybe [`collections.ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap)?), since it is essentially just holding a dictonary with a custom [`update`](https://docs.python.org/3/library/stdtypes.html#dict.update) method. But then you'll need to reconcile `self.aliasdict` with `self.kwdict`.

seisman: > I did wonder if we could turn the `AliasSystem` class into a subclass of [`collections.UserDict`](https://docs.python.org/3/library/collections.html#userdict-objects) or [`collections.OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict) (or maybe [`collections.ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap)?), since it is essentially just holding a dictonary with a custom [`update`](https://docs.python.org/3/library/stdtypes.html#dict.update) method.

Then we will have codes like
```
alias = AliasSystem(A=Alias(...), B=Alias(...), ...)
build_arg_list(alias)
```
or rename `alias` to `kwdict`, but both may be confusing.

weiji14: > or rename `alias` to `kwdict`, but both may be confusing.

how about `aliasdict = AliasSystem(...)`?

seisman: `aliasdict` looks good.

I just tried `UserDict` and it looks good. Here is a minimal example:
```python
from collections import UserDict
from collections.abc import Sequence
from pygmt.alias import Alias

class AliasSystem(UserDict):

    def __init__(self, **kwargs):
        self.aliasdict = kwargs

        kwdict = {}
        for option, aliases in kwargs.items():
            if isinstance(aliases, Sequence):
                values = [alias._value for alias in aliases if alias._value is not None]
                if values:
                    kwdict[option] = "".join(values)
            elif aliases._value is not None:
                kwdict[option] = aliases._value
        super().__init__(kwdict)

    def update(self, kwargs):
        print("Updating dict")
        # Add more checks later.
        for short_param, value in kwargs.items():
            self[short_param] = value
        return self

aliasdict = AliasSystem(
    A=Alias("label"),
    B=Alias((0, 10), separator="/"),
    C=[Alias("text"), Alias("TL", prefix="+j")]
).update({"C": "abc"})
print(aliasdict)
```
The script output is:
```
Updating dict
Updating dict
{'A': 'label', 'B': '0/10', 'C': 'abc'}
```
As you can see, it prints `Updating dict` twice. One is by the `super().__init__` call, another by the `AliasSystem().update` call. Since `update` is a built-in method of dict/UserDict, overriding it is not a good idea. I guess we need to change the method name from `update()` to something like `merge()`/`check()`?

seisman: I've updated the `AliasSystem` class using `UserDict` in 99845c2537d95de2796b95fb1a2528534c391a6f.

pygmt/alias.py
```diff
@@ -195,3 +197,116 @@ def _value(self) -> str | list[str] | None:
...
+    ... )
+    ['-Amytext+o12/12', '-B', '-D1', '-D2', '-D3', '-JX10c/10c', '-c1,2']
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Initialize the alias system as a dictionary with current parameter values.
+        """
+        # Store the aliases in a dictionary, to be used in the merge() method.
+        self.aliasdict = kwargs
```
weiji14, line 246: Should't this be overriding the UserDict's `data` attribute? https://docs.python.org/3/library/collections.html#collections.UserDict.data

```suggestion
        self.data = kwargs
```

Then I suppose you can use `update` as method name? Unless it's still not advised to override that.

seisman: `data` is the real dictionary that stores the contents of the UserDict class, so changing the UserDict object also affects `data`.

```python
from collections import UserDict

alias = UserDict(A="label", B="text")
print(alias.data)
print(alias)

alias.data = {"C": True}
print(alias.data)
print(alias)

alias.update({"D": "good"})
print(alias.data)
print(alias)
```
The output is:
```
{'A': 'label', 'B': 'text'}
{'A': 'label', 'B': 'text'}
{'C': True}
{'C': True}
{'C': True, 'D': 'good'}
{'C': True, 'D': 'good'}
```

weiji14 (approved):
Don't want to hold this up since I'll be busy over the next few weeks (conference + workshops), so I think ok to merge this so the work on the new Alias system can advance. Thanks again @seisman!

PR #3999, opened 2025-07-15
BREAKING: Raise GMTTypeError exception for invalid types. Previously raise GMTInvalidInput

seisman:
Following PR #3985. Addressing #3707 and #3984.

This PR adds a new exception `GMTTypeError`. The error message is like:

> Unrecognized data type: xxxx. Explain the reason.

PR #4003, opened 2025-07-21
Add new exception GMTParameterError for invalid parameters

seisman:
Need to wait for #3999 first.


pygmt/exceptions.py
```diff
@@ -130,3 +130,48 @@ def __init__(self, dtype: object, /, reason: str | None = None):
...
+        require_any: Set[str] | None = None,
+        exclusive: Set[str] | None = None,
+        reason: str | None = None,
+    ):
+        msg = ""
+        if required:
+            msg = (
+                "Required parameter(s) are missing: "
+                f"{', '.join(repr(par) for par in required)}."
+            )
```
weiji14, line 164: For this case where the parameter(s) is required/compulsory, I'm wondering if we can move towards having the error being thrown natively by Python, instead of writing code like:

```python
if kwargs.get("F") is None:
    raise GMTParameterError(required={"filter_type"})
```

Should we revisit #2896 which mentions [PEP692](https://peps.python.org/pep-0692/), and also think about using [PEP655](https://peps.python.org/pep-0655/)'s `typing.Required` qualifier (https://typing.python.org/en/latest/spec/callables.html#required-and-non-required-keys)? Maybe a step towards #262 is to disallow single character arguments for required parameters, so that we can have native Python errors?

seisman: The main issue is that, with the new alias system #4000 implemented, `**kwargs` will no longer be needed. We will still keep `**kwargs` to support single-letter option flags, but other parameters won't be passed via `kwargs` anymore.

Chuan1937:
@seisman
GMTParameterError is a single error type; do we need to add parameter and conflict exceptions? Like I did in #4370.

PR #4004, opened 2025-07-22
is_nonstr_iter: 0-D array is now recognized as non-iterable correctly

### LINKED ISSUE #4001: is_nonstr_iter recognize 0-D numpy array as iterable
As shown below, a 0-D numpy array is recognized as iterable by the `is_nonstr_iter` function, but actually we can't loop over it.
```python
In [1]: import numpy as np

In [2]: x = np.array(40)

In [3]: x.ndim
Out[3]: 0

In [4]: from pygmt.helpers import is_nonstr_iter

In [5]: is_nonstr_iter(x)
Out[5]: True

In [6]: from collections.abc import Iterable, Iterator

In [7]: isinstance(x, Iterable)
Out[7]: True

In [8]: [i for i in x]
TypeError                                 Traceback (most recent call last)
Cell In[9], line 1
----> 1 [i for i in x]

TypeError: iteration over a 0-d array

In [9]: "/".join(x)
TypeError                                 Traceback (most recent call last)
Cell In[10], line 1
----> 1 "/".join(x)

TypeError: can only join an iterable
```

So, a 0-D numpy array is iterable (has the `__iter__` method) but is not an iterator (doens't have the `__next__` method). 
```
In [8]: isinstance(x, Iterator)
Out[8]: False
```
~~Because we usually need to loop over an iterable or join an iterable, perhaps we should use `isinstance(x, Iterator)` rather than `isinstance(x, Iterable)`?~~

Edit: `isinstance([2, 3, 4], Iterator)` returns `False`, so it doens't work.

seisman:
Fixes #4001.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -604,7 +604,8 @@ def build_arg_list(  # noqa: PLR0912
 
 def is_nonstr_iter(value: Any) -> bool:
     """
-    Check if the value is iterable (e.g., list, tuple, array) but not a string.
+    Check if the value is iterable (e.g., list, tuple, array) but not a string or a 0-D
+    array.
 
     Parameters
     ----------
@@ -633,8 +634,14 @@ def is_nonstr_iter(value: Any) -> bool:
     True
     >>> is_nonstr_iter(np.array(["abc", "def", "ghi"]))
     True
+    >>> is_nonstr_iter(np.array(42))
+    False
     """
-    return isinstance(value, Iterable) and not isinstance(value, str)
+    return (
+        isinstance(value, Iterable)
+        and not isinstance(value, str)
+        and not (hasattr(value, "ndim") and value.ndim == 0)
+    )
 
 
 def launch_external_viewer(fname: PathLike, waiting: float = 0) -> None:


pygmt/helpers/utils.py
```diff
@@ -633,8 +635,14 @@ def is_nonstr_iter(value: Any) -> bool:
...
     >>> is_nonstr_iter(np.array(["abc", "def", "ghi"]))
     True
+    >>> is_nonstr_iter(np.array(42))
+    False
     """
-    return isinstance(value, Iterable) and not isinstance(value, str)
+    return (
+        isinstance(value, Iterable)
+        and not isinstance(value, str)
+        and not (isinstance(value, np.ndarray) and value.ndim == 0)
```
weiji14: Should we just check for the `ndim` attribute, since it part of the Python array API standard? Xref https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.array.ndim.html#ndim

```suggestion
        and not (hasattr(value, "ndim") and value.ndim == 0)
```

pygmt/helpers/utils.py
```diff
@@ -15,6 +15,7 @@
 from pathlib import Path
 from typing import Any, Literal
 
+import numpy as np
```
weiji14:
```suggestion
```

pygmt/helpers/utils.py
weiji14: Oh, I see you did that first, and then copilot suggested otherwise :sweat_smile:

seisman: This is exactly what I wrote in the intial version, then copilot made the suggestion at https://github.com/GenericMappingTools/pygmt/pull/4004#discussion_r2221038723.

weiji14: I think ok to use `hasattr` to avoid an import, and also allow for duck typing.

seisman: Reverted in 001cbb985afbda1823f13615f4e0f90656c26b5e.

PR #4008, opened 2025-07-23
Regression test to ensure gmt accessor info is kept after file deletion

### LINKED ISSUE #4005: deprecated `pygmt.load_dataarray()` inconsistent with `xr.load_datarray`

I'm not sure if this is a bug or a change to PyGMT which I missed. Since the `pygmt.load_dataarray` function is deprecated, I'm using the suggested alternative, `xr.load_datarray(..., engine="gmt", raster_kind='grid')`, however, this gives an incorrect grid registration. 

As you can see below, a grid with a registration of 0 (gridline), when converted to pixle registered with `grdedit` and then opened again, gives the correct pixel registration (1) when using `pygmt.load_datarray`, but an incorrect gridline registration (0) when opened with `xr.load_dataarray`.

```python
import pygmt
import xarray as xr

grid = pygmt.datasets.load_earth_relief(
    resolution="30s", region=[144.5, 145.5, 13, 14.5], registration="gridline"
)
print(grid.gmt.registration)
---> 0

with pygmt.clib.Session() as ses:
    with ses.virtualfile_from_grid(grid) as f_in:
        with pygmt.helpers.GMTTempFile(suffix=".nc") as tmpfile:
            args = f"{f_in} -T -G{tmpfile.name}"
            ses.call_module("grdedit", args)
            f_out = pygmt.load_dataarray(tmpfile.name)
print(f_out.gmt.registration)
---> 1 # this is correct

with pygmt.clib.Session() as ses:
    with ses.virtualfile_from_grid(grid) as f_in:
        with pygmt.helpers.GMTTempFile(suffix=".nc") as tmpfile:
            args = f"{f_in} -T -G{tmpfile.name}"
            ses.call_module("grdedit", args)
            f_out = xr.load_dataarray(tmpfile.name, engine='gmt', raster_kind='grid')
print(f_out.gmt.registration)
---> 0  # this is incorrect
```

### Full error message

```python-traceback

```

### System information

```bash
PyGMT information:
  version: v0.16.0
System information:
  python: 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0]
  executable: /home/sungw937/miniforge3/envs/polartoolkit/bin/python
  machine: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.39
Dependency information:
  numpy: 2.2.6
  pandas: 2.3.1
  xarray: 2025.7.1
  packaging: 25.0
  contextily: None
  geopandas: 1.1.1
  IPython: 9.4.0
  pyarrow: 20.0.0
  rioxarray: 0.19.0
  gdal: None
  ghostscript: 10.04.0
GMT library information:
  version: 6.5.0
  padding: 2
  share dir: /home/sungw937/miniforge3/envs/polartoolkit/share/gmt
  plugin dir: /home/sungw937/miniforge3/envs/polartoolkit/lib/gmt/plugins
  library path: /home/sungw937/miniforge3/envs/polartoolkit/lib/libgmt.so
  cores: 8
  grid layout: rows
  image layout: 
  binary version: 6.5.0
```

weiji14:

Ensure that using `load_dataarray/open_dataarray(..., engine="gmt", ...)` caches the `GMTDataArrayAccessor` info, so that the correct registration/gtype is returned even if the original source NetCDF file is deleted.

TODO:
- [x] Add regression test
- [ ] Fix bug by caching the accessor values properly

Fixes #4005

**Reminders**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_xarray_backend.py
+++ b/pygmt/tests/test_xarray_backend.py
@@ -9,6 +9,7 @@ import numpy as np
 import numpy.testing as npt
 import pytest
 import xarray as xr
+from pygmt.clib import Session
 from pygmt.enums import GridRegistration, GridType
 from pygmt.exceptions import GMTValueError
 from pygmt.helpers import GMTTempFile
@@ -36,6 +37,28 @@ def test_xarray_backend_load_dataarray():
         assert dataarray.gmt.registration is GridRegistration.PIXEL
         # ensure data array can be saved back to a NetCDF file
         dataarray.to_netcdf(tmpfile.name)
+
+
+def test_xarray_backend_load_dataarray_temp_nc_grid():
+    """
+    Check that xarray.load_dataarray works to read a temporary netCDF grid, and ensure
+    that GMTDataArrayAccessor information is retained after original file is deleted.
+
+    This is a regression test for
+    https://github.com/GenericMappingTools/pygmt/issues/4005
+    """
+
+    with Session() as lib:
+        with GMTTempFile(suffix=".nc") as tmpfile:
+            args = f"@earth_relief_01d_g -T -G{tmpfile.name}"  # change from gridline to pixel registration
+            lib.call_module(module="grdedit", args=args)
+            dataarray = xr.load_dataarray(
+                tmpfile.name, engine="gmt", raster_kind="grid"
+            )
+
+        # Ensure GMTDataArrayAccessor info is preserved after tempfile is deleted
+        assert dataarray.gmt.registration is GridRegistration.PIXEL
+        assert dataarray.gmt.gtype is GridType.CARTESIAN
 
 
 def test_xarray_backend_gmt_open_nc_grid():


pygmt/xarray/backend.py
```diff
@@ -150,4 +150,6 @@ def open_dataset(  # type: ignore[override]
                 raster.encoding["source"] = (
                     sorted(source)[0] if isinstance(source, list) else source
                 )
-                return raster.to_dataset()
+                raster_dataset: xr.Dataset = raster.to_dataset()
+                _ = raster_dataset.gmt  # Load GMTDataArray accessor information
+                return raster_dataset
```
weiji14, line 155: _Originally posted by @seisman in https://github.com/GenericMappingTools/pygmt/pull/3932#discussion_r2073658948_

> Actually, it's likely that the accessor information will be lost when converting via `to_dataset`.

So we can set the accessor info at the `xr.Dataset` level, but not sure if it's a good idea because then `xr.Dataset` would have GMT-functions enabled...

seisman: The test is still failing. I guess that the accessor information is lost in both dataarray->dataset and dataset->dataarray conversions. 

When calling `xr.open_dataarray` with the "gmt" engine, it actually calls the GMTBackendEntry's `open_dataset` method first, so the conversion from dataset to dataarray cannot be avoided.

seisman:
```python
>>> import pygmt
>>> import xarray as xr

>>> grid = xr.open_dataarray("@static_earth_relief.nc", engine="gmt", raster_kind="grid")

>>> grid.encoding["source"]
np.str_('/home/seisman/.gmt/cache/static_earth_relief.nc')

>>> grid.gmt.registration, grid.gmt.gtype
(<GridRegistration.PIXEL: 1>, <GridType.GEOGRAPHIC: 1>)

>>> # Set source encoding to a fake path, similar to a temporary file being deleted
>>> grid.encoding["source"] = "/path/to/nothing"  
>>> grid.gmt.registration, grid.gmt.gtype
(<GridRegistration.PIXEL: 1>, <GridType.GEOGRAPHIC: 1>)

>>> # Convert a dataarray to a dataset. The source encoding is not kept.
>>> da = grid.to_dataset()
>>> da.encoding
{}

>>> # Convert a dataset to a dataarray.
>>> # Xref: https://github.com/pydata/xarray/blob/fb49a3b0a83272467aa01ac189b95e9e7ebe01aa/xarray/backends/api.py#L950
>>> (ds, ) = da.data_vars.values()
>>> ds.encoding
{'source': '/path/to/nothing'}

>>> ds.gmt.registration, ds.gmt.gtype
(<GridRegistration.GRIDLINE: 0>, <GridType.CARTESIAN: 0>)
```

seisman: As shown above, the accessor information is lost during the dataset->dataarray conversion, and we rely on the source encoding to decide the registration and gtype, which default to 0 for temporary files being deleted.

I don't think we have any solutions here, and need to document the limitation and recommend:

1. Avoiding using temporary files if possible 
2. If temporary files cannot be avoided, use `_ = grid.gmt` before deleting the temporary file

weiji14: Yeah, I found this upstream issue - https://github.com/pydata/xarray/issues/3268 that accessors are not meant to cache a state. I do wish though that the `@xr.register_dataarray_accessor` decorator could work by us returning an `xr.DataArray` directly (instead of having to return an `xr.Dataset` that then gets converted to an `xr.DataArray`).

Agree that we should document that temporary files should be avoided and/or trigger caching using `_ = grid.gmt`, then close #4005 as wontfix (with a workaround).

weiji14: > I do wish though that the `@xr.register_dataarray_accessor` decorator could work by us returning an `xr.DataArray` directly (instead of having to return an `xr.Dataset` that then gets converted to an `xr.DataArray`).

Opened an upstream issue on xarray for this -> https://github.com/pydata/xarray/issues/10562

seisman:
Perhaps we need to bump this PR to v0.18.0?

Issue #4011, opened 2025-07-24
Better way to document the aliases in docstrings


seisman:
In the initial version of the alias system, each GMT option flag is aliased to a single parameter PyGMT, so documenting the aliases in docstrings is simple and can be fully automatic. The aliases docstrings are like:
```
**Aliases:**

- J=projection
- B=frame
- R=region
```
Then we realized that sometimes an option flag is too complicated and should be aliased to multiple parameters, and they're documented like below since #3965:
```
**Aliases:**

- J=projection
- B=frame
- R=region
- A=constantfill/gridfill/neighborfill/splinefill
```

With the new alias system introduced in #4000, we have to manually maintain the aliases in docstrings.

We need to discuss what the docstrings look like, especially for option flags that are aliased to multiple parameters. Let me take `-U` as an example. The GMT CLI syntax is `-U[label][+jjustify][+odx[/dy]][+ttext]`, then the docstrings can be:

**U**=label, justify (**+j**), offset (**+o**), text (**+t**)

or 

**U**=label, **+j**: justify, **+o**: offset, **+t**: text

We also need to add a documentation explaining how the new alias system works and how to read and understand the aliases in docstrings.

_Originally posted by @seisman in https://github.com/GenericMappingTools/pygmt/issues/4000#issuecomment-3111800039_

seisman:
Ping @GenericMappingTools/pygmt-maintainers Any comments on the docstring styles for aliases?

yvonnefroehlich:
> **U**=label, justify (**+j**), offset (**+o**), text (**+t**)
>
> or 
>
> **U**=label, **+j**: justify, **+o**: offset, **+t**: text

I prefer the second version for the syntax, as the GMT flag or modifier is consistently followed by the alias.

---
Hm, maybe this was only an example with focus on modifiers which have or will have their own alias. Or do we now also want to mention the flags which are implemented as standalone methods? **-U** (as well as **-X** and **-Y**) are not anymore supported but we have `Figure.timestamp` and `Figure.shift_origin`.

seisman:
> do we now also want to mention the flags which are implemented as standalone methods? **-U** (as well as **-X** and **-Y**) are not anymore supported but we have `Figure.timestamp` and `Figure.shift_origin`.

I prefer not to list these flags in the docstrings.

yvonnefroehlich:
[...]

PR #4024, opened 2025-07-29
Migrate the 'projection' parameter to the new alias system

seisman:
This PR will migrate the `projection` parameter in all wrappers to the new alias system.

Currently, only one file is updated. I'll apply the same change to all wrappers if approved.

### DIFF
--- a/pygmt/src/basemap.py
+++ b/pygmt/src/basemap.py
@@ -2,6 +2,7 @@
 basemap - Plot base maps and frames.
 """
 
+from pygmt.alias import Alias, AliasSystem
 from pygmt.clib import Session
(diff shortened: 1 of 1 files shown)


pygmt/src/basemap.py
```diff
@@ -39,6 +39,7 @@ def basemap(self, **kwargs):
     Full GMT docs at :gmt-docs:`basemap.html`.
 
     {aliases}
+       - J=projection
```
seisman, line 42: As mentioned in https://github.com/GenericMappingTools/pygmt/issues/4011, with the new alias system, we have to manually maintain the list of aliases. Currently, the `J` alias is added at the end of the alias list. I guess we should sort the aliases alphabetically when we migrate more aliases into the new alias system.

michaelgrund (approved):

PR #4030, opened 2025-07-30
Migrate the panel parameter to the new alias system

seisman:
As the PR title says.

### DIFF
--- a/pygmt/src/x2sys_cross.py
+++ b/pygmt/src/x2sys_cross.py
@@ -250,9 +250,12 @@ def x2sys_cross(
                     unit = time_unit.upper() if time_unit in "wd" else time_unit
                     scale = 1.0
 
-            columns = result.columns[2:4]
-            result[columns] *= scale
-            result[columns] = result[columns].apply(pd.to_timedelta, unit=unit)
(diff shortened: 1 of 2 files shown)


pygmt/src/basemap.py
```diff
@@ -86,6 +86,7 @@ def basemap(self, projection=None, **kwargs):
     self._activate_figure()
     aliasdict = AliasSystem(
         J=Alias(projection, name="projection"),
+        c=Alias(panel, name="panel", sep=",", size=2),
```
seisman: I'm considering if we should define a function like:
```
def common_alias(value, name):
    match name:
        case "projection":
            return Alias(value, name="projection")
        case "panel":
            return Alias(value, name="panel", sep=",", size=2)
```
then in wrappers, we can use it like:
```
from pygmt.alias import common_alias
aliasdict = AliasSystem(
    J=common_alias(projection, name="projection"),
    c=common_alias(panel, name="panel"),
)
```
so that we don't have to repeat `alias(panel, name="panel", sep=",", size=2),` in all wrappers.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final reviews. I plan to merge it in 24 hours.

PR #4089, opened 2025-09-15
Figure.image: Add the alias 'invert' for option -I

seisman:
Alias `-I` option to `invert`.

Upstream docs: https://docs.generic-mapping-tools.org/6.5/image.html#i

**Preview**:

### DIFF
--- a/pygmt/src/image.py
+++ b/pygmt/src/image.py
@@ -19,6 +19,7 @@ def image(
     imagefile: PathLike,
     projection=None,
     box: Box | bool = False,
+    invert: bool = False,
     verbose: Literal["quiet", "error", "warning", "timing", "info", "compat", "debug"]
     | bool = False,
     panel: int | tuple[int, int] | bool = False,
@@ -35,6 +36,7 @@ def image(
 
     {aliases}
        - F = box
+       - I = invert
        - J = projection
        - V = verbose
        - c = panel
@@ -68,6 +70,9 @@ def image(
         to make those pixels transparent. Can be repeated with different
         settings. Alternatively, for color images you can select a single
         *color* that should be made transparent instead (**+t**).
+    invert
+        Invert 1-bit image before plotting, i.e., black pixels (on) becomes white (off)
+        and vice versa. Ignored if used with color images.
     monochrome : bool
         Convert color image to monochrome grayshades using the (television)
         YIQ-transformation.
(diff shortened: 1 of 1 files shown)


pygmt/src/image.py
```diff
@@ -68,6 +70,9 @@ def image(
         to make those pixels transparent. Can be repeated with different
         settings. Alternatively, for color images you can select a single
         *color* that should be made transparent instead (**+t**).
+    invert
+        Invert 1-bit image before plotting, i.e., black pixels (on) becomes white (off)
+        and vice versa. Ignored if used with color images.
```
yvonnefroehlich, line 90: Hm, just trying this and maybe I am missing something here. I thought the remote raster image "vader1.png" is such a 1-bit image (see the last example at https://docs.generic-mapping-tools.org/dev/image.html#examples), but using ``I=True`` does not flip the `black` and `white` pixels as expected:

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[-5, 5] * 2, projection="X3c/1c", frame=0)

fig.image(imagefile="@vader1.png", position="jLM")
fig.image(imagefile="@vader1.png", position="jMC", bitcolor=["gray+f", "blue+b"])
fig.image(imagefile="@vader1.png", position="jRM", I=True)  # Does not do something
# fig.image(imagefile="@vader1.png", position="jRM", bitcolor=["black+f", "white+b"])

fig.show()
```
image_invert

Same for 
```bash
gmt image @vader1.png -I -png image_I_gmt
````
image_I_gmt

Comment (seisman) [reply pygmt/src/image.py:90]: Hmm, it looks like an upstream bug.

Comment (yvonnefroehlich) [reply pygmt/src/image.py:90]: I tested it with the gmt dev version, and still no inverted colors. Thus, I wrote an issue report at https://github.com/GenericMappingTools/gmt/issues/8795.

Comment (seisman) [reply pygmt/src/image.py:90]: The upstream bug has been fixed.

I've added a note in the docstrings.

File: pygmt/src/image.py [hunk 2]
```diff
@@ -85,6 +85,15 @@ def image(
     monochrome
         Convert color image to monochrome grayshades using the (television)
         YIQ-transformation.
+    invert
+        Invert 1-bit image before plotting, i.e., black pixels (on) becomes white (off)
```
michaelgrund:
```suggestion
        Invert 1-bit image before plotting, i.e., black pixels (on) become white (off)
```

PR #2885, opened 2023-12-16
Refactor the kwargs_to_strings decorator to avoid try-except in loops

seisman:
**Description of proposed changes**

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #4097, opened 2025-09-17
sequence_join: Correctly join datetime-like values in pd.DataFrame/xr.DataArray objects

seisman:
In the `sequence_join` function, we use `str(v)` to convert any value into a string that can be passed to the GMT API. However, `str(v)` can't handle some datetime-like objects. For example, `datetime.datetime(2010, 2, 1)` is converted to `2010-02-01 00:00:00` with a whitespace. Below is an example showing the issue:
```python
>>> from pygmt.helpers import sequence_join
>>> import pandas as pd
>>> import numpy as np
>>> import xarray as xr
>>> import datetime
>>> sequence_join(
...     [
...         datetime.date(2010, 1, 1),
...         np.datetime64("2010-01-01T16:00:00"),
...         np.array("2010-03-01T00:00:00", dtype=np.datetime64),
...         datetime.datetime(2010, 3, 1),
...         pd.Timestamp("2015-01-01T12:00:00.123456789"),
...         xr.DataArray(data=np.datetime64("2005-01-01T08:00:00", "ns")),
...     ],
...     sep="/",
... )
"2010-01-01/2010-01-01T16:00:00/2010-03-01T00:00:00/2010-03-01 00:00:00/2015-01-01 12:00:00.123456789/<xarray.DataArray ()> Size: 8B\narray('2005-01-01T08:00:00.000000000', dtype='datetime64[ns]')"
```

This PR fixes the issue by adapting the existing codes in the `kwargs_to_strings` decorator:
https://github.com/GenericMappingTools/pygmt/blob/63d4561e2c349dd2dd82719fd525d3a94364d2b8/pygmt/helpers/decorators.py#L737-L746

With the changes, the above code produces the following output:
```
'2010-01-01/2010-01-01T16:00:00/2010-03-01T00:00:00/2010-03-01T00:00:00.000000/2015-01-01T12:00:00.123456/2005-01-01T08:00:00.000000000
```

The changes are necessary to support arguments like 
```
region=[datetime.date(2010, 1, 1), pd.Timestamp("2015-01-01T12:00:00.123456789"), 0, 10]
```
in the new alias system.

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -883,7 +883,7 @@ def sequence_join(
             else s
             for item in value
         ]
-        return sep.join(_values)
+        return sep.join(_values)  # type: ignore[arg-type]
 
     # Now it must be a 2-D sequence.
     if ndim == 1:


pygmt/helpers/utils.py
```diff
@@ -848,7 +873,17 @@ def sequence_join(
...
+        # objects (e.g., datetime.datetime, pandas.Timestamp), which contains a space.
+        # If so, use np.datetime_as_string to convert it to ISO 8601 string format
+        # YYYY-MM-DDThh:mm:ss.ffffff.
+        _value = [
+            np.datetime_as_string(np.asarray(item, dtype=np.datetime64))
+            if " " in str(item)
+            else item
+            for item in value
+        ]
+        return sep.join(str(v) for v in _value)
```
weiji14: We have some similar code from #2885/#562 here:

https://github.com/GenericMappingTools/pygmt/blob/bb6e084bc946beb6870e300097aa8d03c6c7ee8d/pygmt/helpers/decorators.py#L722-L732

should there be a common function maybe to avoid duplication?

seisman: Yes, the new codes are modified from the one you posted. I prefer to keep the `kwargs_to_strings` decorator unchanged, since we're working towards removing this decorator.

weiji14: Hmm ok. I'm just trying to think of how to remove the double `str()` call somehow. Something like this, but not exactly perfect:

```suggestion
        # 'str(v)' produces a string like '2024-01-01 00:00:00' for some datetime-like
        # objects (e.g., datetime.datetime, pandas.Timestamp), which contains a space.
        # If so, use np.datetime_as_string to convert it to ISO 8601 string format
        # YYYY-MM-DDThh:mm:ss.ffffff.
        _values = [
            np.datetime_as_string(np.asarray(item, dtype=np.datetime64))
            if " " in str(item)
            else str(item)
            for item in value
        ]
        return sep.join(_values)
```

seisman: I've changed it to:
```python
        _values = [
            np.datetime_as_string(np.asarray(item, dtype="datetime64"))
            if " " in (s := str(item))
            else s
            for item in value
        ]
        return sep.join(_values)  # type: ignore[arg-type]
```

PR #4099, opened 2025-09-17
AliasSystem: Migrate the region parameter to the new alias system

seisman:
- **sequence_join: Correctly join datetime-like values in pd.DataFrame/xr.DataArray objects**
- **Add region to the common parameter**
- **Update basemap**
- **Update other wrappers**

### DIFF
--- a/pygmt/alias.py
+++ b/pygmt/alias.py
@@ -300,6 +300,10 @@ class AliasSystem(UserDict):
         """
         for key, value in kwargs.items():
             match key:
+                case "J":
+                    alias = Alias(value, name="projection")
+                case "R":
+                    alias = Alias(value, name="region", sep="/", size=(4, 6))
                 case "V":
                     alias = Alias(
                         value,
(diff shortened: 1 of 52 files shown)


pygmt/alias.py
```diff
@@ -300,6 +300,10 @@ def add_common(self, **kwargs):
         """
         for key, value in kwargs.items():
             match key:
+                case "J":
+                    alias = Alias(value, name="projection")
+                case "R":
+                    alias = Alias(value, name="region", sep="/", size=(4, 6))
```
seisman, line 306: The size is usually 4, but can be 6 for 3-D plots.

pygmt/src/grdlandmask.py
```diff
@@ -115,7 +111,7 @@ def grdlandmask(
     >>> # latitude range of 30° N to 35° N, and a grid spacing of 1 arc-degree
     >>> landmask = pygmt.grdlandmask(spacing=1, region=[125, 130, 30, 35])
     """
-    if kwargs.get("I") is None or kwargs.get("R") is None:
+    if kwargs.get("I") is None or kwargs.get("R", region) is None:
```
seisman, line 111: Check if `region`/`R` is not specified.

PR #4103, opened 2025-09-18
Migrate the 'no_clip' parameter to the new alias system

seisman:
Some wrappers have the `no_clip` parameter, which can be either False or True. This PR migrates the parameter to the new alias system.

`no_clip` in `Figure.plot` and `Figure.plot3d` are not updated in this PR, because in these two methods, `no_clip` can be False, True, `"c"` or `"r"` and we need to discuss the long-form arguments for "c"/"r". So leave it to a separate PR instead.

### DIFF
--- a/pygmt/src/contour.py
+++ b/pygmt/src/contour.py
@@ -5,7 +5,7 @@ contour - Contour table data by direct triangulation.
 from typing import Literal
 
 from pygmt._typing import PathLike, TableLike
-from pygmt.alias import AliasSystem
+from pygmt.alias import Alias, AliasSystem
 from pygmt.clib import Session
 from pygmt.helpers import (
     build_arg_list,
@@ -23,7 +23,6 @@ from pygmt.helpers import (
     C="levels",
     G="label_placement",
     L="triangular_mesh_pen",
-    N="no_clip",
     R="region",
     S="skip",
     W="pen",
@@ -43,6 +42,7 @@ def contour(
     x=None,
     y=None,
     z=None,
+    no_clip: bool = False,
     projection=None,
     verbose: Literal["quiet", "error", "warning", "timing", "info", "compat", "debug"]
     | bool = False,
@@ -62,6 +62,7 @@ def contour(
 
     {aliases}
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
        - t = transparency
@@ -109,9 +110,9 @@ def contour(
         Color the triangles using CPT.
     triangular_mesh_pen : str
         Pen to draw the underlying triangulation [Default is ``None``].
-    no_clip : bool
-        Do **not** clip contours or image at the frame boundaries
-        [Default is ``False`` to fit inside ``region``].
+    no_clip
+        Do **not** clip contours or image at the frame boundaries [Default is ``False``
+        to fit inside ``region``].
     Q : float or str
         [*cut*][**+z**].
         Do not draw contours with less than *cut* number of points.
@@ -159,7 +160,9 @@ def contour(
             else:  # Multiple levels
                 kwargs[arg] = ",".join(f"{item}" for item in kwargs[arg])
 
-    aliasdict = AliasSystem().add_common(
+    aliasdict = AliasSystem(
+        N=Alias(no_clip, name="no_clip"),
+    ).add_common(
         J=projection,
         V=verbose,
         c=panel,
--- a/pygmt/src/contour.py
+++ b/pygmt/src/contour.py
@@ -107,7 +107,7 @@ def contour(
         five controlling algorithms. See :gmt-docs:`contour.html#g` for
         details.
     I : bool
-        Color the triangles using CPT.
+        Color the triangles using the CPT if given via ``levels``.
     triangular_mesh_pen : str
         Pen to draw the underlying triangulation [Default is ``None``].
     no_clip
--- a/pygmt/src/contour.py
+++ b/pygmt/src/contour.py
@@ -111,8 +111,8 @@ def contour(
     triangular_mesh_pen : str
         Pen to draw the underlying triangulation [Default is ``None``].
     no_clip
-        Do **not** clip contours or image at the frame boundaries [Default is ``False``
-        to fit inside ``region``].
+        Do **not** clip contours or colored triangles at the frame boundaries [Default
+        is ``False`` to fit inside ``region``].
     Q : float or str
         [*cut*][**+z**].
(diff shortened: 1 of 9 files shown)


pygmt/src/contour.py
```diff
@@ -109,9 +110,9 @@ def contour(
         Color the triangles using CPT.
     triangular_mesh_pen : str
         Pen to draw the underlying triangulation [Default is ``None``].
-    no_clip : bool
-        Do **not** clip contours or image at the frame boundaries
-        [Default is ``False`` to fit inside ``region``].
+    no_clip
+        Do **not** clip contours or image at the frame boundaries [Default is ``False``
+        to fit inside ``region``].
```
yvonnefroehlich: I am wondering to what "image" is refering here.

seisman: It comes from the upstream documentation at https://docs.generic-mapping-tools.org/6.5/contour.html#n.

Here "image" is used because `contour` can color the triangles using a CPT. See the example below:
```
gmt begin map
    gmt makecpt -Cseis -T600/1000/100
    gmt basemap -R0.5/6/0.5/6 -JX10c/10c -Baf
    gmt plot @Table_5_11.txt -Sc0.2c
    gmt contour @Table_5_11.txt -Wthin -C -I -N -t50
gmt end show
```

map

pygmt/src/grdimage.py
```diff
@@ -76,6 +76,7 @@ def grdimage(
 
     {aliases}
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
        - t = transparency
```
yvonnefroehlich, line 83:
```suggestion
      - J = projection
      - N = no_clip
      - V = verbose
      - c = panel
      - t = transparency
```

pygmt/src/inset.py
```diff
@@ -37,6 +38,7 @@ def inset(
     {aliases}
        - F = box
        - J = projection
+       - N = no_clip
        - V = verbose
```
yvonnefroehlich, line 42:
```suggestion
      - F = box
      - J = projection
      - N = no_clip
      - V = verbose
```

pygmt/src/meca.py
```diff
@@ -203,6 +203,7 @@ def meca(  # noqa: PLR0913
 
     {aliases}
        - J = projection
+       - N = no_clip
        - S = scale/convention/component
        - V = verbose
        - c = panel
```
yvonnefroehlich, line 209:
```suggestion
      - J = projection
      - N = no_clip
      - S = scale/convention/component
      - V = verbose
      - c = panel
```

pygmt/src/text.py
```diff
@@ -76,6 +76,7 @@ def text_(  # noqa: PLR0912, PLR0913, PLR0915
     {aliases}
        - F = **+a**: angle, **+c**: position, **+j**: justify, **+f**: font
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
        - t = transparency
```
yvonnefroehlich, line 82:
```suggestion
      - F = **+a**: angle, **+c**: position, **+j**: justify, **+f**: font
      - J = projection
      - N = no_clip
      - V = verbose
      - c = panel
      - t = transparency
```

pygmt/src/tilemap.py
```diff
@@ -59,6 +59,7 @@ def tilemap(  # noqa: PLR0913
 
     {aliases}
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
        - t = transparency
```
yvonnefroehlich, line 65:
```suggestion
      - J = projection
      - N = no_clip
      - V = verbose
      - c = panel
```

pygmt/src/velo.py
```diff
@@ -67,6 +62,7 @@ def velo(
 
     {aliases}
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
```
yvonnefroehlich, line 67:
```suggestion
      - J = projection
      - N = no_clip
      - V = verbose
      - c = panel
```

pygmt/src/contour.py
```diff
@@ -62,6 +62,7 @@ def contour(
 
     {aliases}
        - J = projection
+       - N = no_clip
        - V = verbose
        - c = panel
```
yvonnefroehlich, line 67:
```suggestion
      - J = projection
      - N = no_clip
      - V = verbose
      - c = panel
```

pygmt/src/grdimage.py
yvonnefroehlich: Hm, not sure what GitHub is doing here, but actually I only removed one whitespace before the newly added bullet point for **-N**.

seisman: Why do you want to remove one whitespace before `-N`? I believe the current one is correct. See https://pygmt-dev--4103.org.readthedocs.build/en/4103/api/generated/pygmt.Figure.grdimage.html#pygmt.Figure.grdimage

yvonnefroehlich: Actually, I am unsure. Directly on GitHub it looks like all bullet points have the same number of leading whitespaces, but when I look at it in the suggestion view, it seems like there is one whitespace more for **-N**:

| github | suggestion view |
|  --- | --- |
| list_alias_whitspace | list_alias_whitspace_review |

seisman: Same for me with the new GitHub UX. It's more likely a GitHub bug.

yvonnefroehlich: Yeah, I experienced some unexpected behavior in the last days... Then we just leave this as is and in case one of the lists is not correctly aligned, we can still make a PR to fix this.

pygmt/src/contour.py
yvonnefroehlich: Ah, thanks for the example 🙂!
Not really happy with this formulation, but do not have a good alternative.

pygmt/src/contour.py
```diff
@@ -109,9 +110,9 @@ def contour(
         Color the triangles using CPT.
```
yvonnefroehlich:
```suggestion
        Color the triangles using the CPT if given via ``levels``.
```

pygmt/src/contour.py
yvonnefroehlich:
```suggestion
    no_clip
        Do **not** clip contours or colored triangles at the frame boundaries [Default
        is ``False`` to fit inside ``region``].
```

PR #4105, opened 2025-09-20
mailmap: Add comments to explain the use of the mailmap file

seisman:
`.mailmap` was initially added in https://github.com/GenericMappingTools/pygmt/pull/3905. This PR add some comments so that we don't forget why this file exists.


.mailmap
```diff
@@ -1,3 +1,5 @@
+# Map author/committer names and/or E-Mail addresses for Git.
```
yvonnefroehlich: Is there a reason for using upper case letters here?

```suggestion
# Map author/committer names and/or email addresses for git.
```

PR #4106, opened 2025-09-20
Figure.timestamp: Simplify the docstrings for parameter 'justify'

seisman:
No code changes.

**Preview**: https://pygmt-dev--4106.org.readthedocs.build/en/4106/api/generated/pygmt.Figure.timestamp.html

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,3 +1,5 @@
+# Map author/committer names and/or email addresses for git.
+# Reference: https://git-scm.com/docs/gitmailmap
 Dongdong Tian <seisman.info@gmail.com>
 Dongdong Tian <seisman.info@gmail.com> <seisman@users.noreply.github.com>
 Jiayuan Yao <coreman.seism@gmail.com>
--- a/pygmt/src/grdclip.py
+++ b/pygmt/src/grdclip.py
@@ -97,16 +97,16 @@ def grdclip(
     Example
     -------
     >>> import pygmt
-    >>> # Load a grid of @earth_relief_30m data, with a longitude range of
-    >>> # 10° E to 30° E, and a latitude range of 15° N to 25° N
+    >>> # Load the 30 arc-minutes Earth relief grid, with a longitude range of 10° E to
+    >>> 30° E, and a latitude range of 15° N to 25° N
     >>> grid = pygmt.datasets.load_earth_relief(
     ...     resolution="30m", region=[10, 30, 15, 25]
     ... )
     >>> # Report the minimum and maximum data values
     >>> [grid.data.min(), grid.data.max()]
     [183.5, 1807.0]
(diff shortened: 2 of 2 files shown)


pygmt/src/timestamp.py
```diff
@@ -41,12 +41,9 @@ def timestamp(
...
     justify
-        Justification of the timestamp box relative to the plot's bottom-left corner
-        (i.e., the plot origin). Give a two-character code that is a combination of a
-        horizontal (**L**\ (eft), **C**\ (enter), or **R**\ (ight)) and a vertical
-        (**T**\ (op), **M**\ (iddle), or **B**\ (ottom)) code. For example,
-        ``justify="TL"`` means choosing the **T**\ op **L**\ eft point of the timestamp
-        as the anchor point.
+        Specify a :doc:`2-character justification code </techref/justification_codes>`
+        of the timestamp box relative to the plot's bottom-left corner (i.e., the plot
+        origin). E.g., e.g., the default is ``"BL"`` for Bottom Left.
```
seisman:
```suggestion
        origin). E.g., the default is ``"BL"`` for Bottom Left.
```

pygmt/src/timestamp.py
```diff
@@ -41,12 +41,9 @@ def timestamp(
...
     justify
-        Justification of the timestamp box relative to the plot's bottom-left corner
-        (i.e., the plot origin). Give a two-character code that is a combination of a
-        horizontal (**L**\ (eft), **C**\ (enter), or **R**\ (ight)) and a vertical
-        (**T**\ (op), **M**\ (iddle), or **B**\ (ottom)) code. For example,
-        ``justify="TL"`` means choosing the **T**\ op **L**\ eft point of the timestamp
-        as the anchor point.
+        Specify a :doc:`2-character justification code </techref/justification_codes>`
+        of the timestamp box relative to the plot's bottom-left corner (i.e., the plot
+        origin). E.g., the default is ``"BL"`` for Bottom Left.
```
yvonnefroehlich: If the plotting origin was changed before plotting the timestamp the new plotting origin is considered (it is not reset before plotting the timestamp to Left Bottom). 

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[20, 30, -10, 10], projection="X6c/3c", frame=True)
fig.timestamp()
fig.shift_origin(xshift="6c", yshift="3c")
fig.timestamp(font="red")

fig.show()
```

timestamp_referencepoint

Thus, I think we should use a more general formulation here (if changed, we need also to update the text at the beginning).

```suggestion
        Specify a :doc:`2-character justification code </techref/justification_codes>`
        for the timestamp box relative to the current plot origin. The default is the
        Bottom Left (``"BL"``) corner.
```

PR #4108, opened 2025-09-20
pygmt.grdfill: Migrate the hole parameter to the new alias system and add a test

seisman:
Migrate the 'hole' parameter and add a test to ensure that both `hole` and `no_data` (deprecated) work as expected.

### DIFF
--- a/pygmt/tests/test_grdfill.py
+++ b/pygmt/tests/test_grdfill.py
@@ -114,6 +114,31 @@ def test_grdfill_gridfill_dataarray(grid):
     npt.assert_array_equal(result[3:6, 3:5], bggrid[3:6, 3:5])
 
 
+def test_grdfill_hole(grid, expected_grid):
+    """
+    Test grdfill with different hole values.
+    """
+    # Prepare for a grid with -99999 as holes.
+    grid_no_nan = grdfill(grid=grid, constantfill=-99999)
+    assert not np.isnan(grid_no_nan).any()
+    assert -99999 in grid_no_nan
+    # Now that we have a grid with -99999 as holes, fill it with a constant value of 20.
+    result = grdfill(grid=grid_no_nan, constantfill=20, hole=-99999)
+
+    # check information of the output grid
+    assert isinstance(result, xr.DataArray)
+    assert result.gmt.gtype is GridType.GEOGRAPHIC
+    assert result.gmt.registration is GridRegistration.PIXEL
+    # check information of the output grid
+    xr.testing.assert_allclose(a=result, b=expected_grid)
+
+    # Test the deprecated 'no_data' parameter.
+    # TODO(PyGMT>=0.19.0): Remove the following lines.
+    with pytest.warns(FutureWarning):
+        result2 = grdfill(grid=grid_no_nan, constantfill=20, no_data=-99999)
+    xr.testing.assert_allclose(a=result2, b=expected_grid)
+
+
 def test_grdfill_inquire(grid):
     """
     Test grdfill with inquire mode.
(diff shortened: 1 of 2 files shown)


pygmt/tests/test_grdfill.py
```diff
@@ -114,6 +114,31 @@ def test_grdfill_gridfill_dataarray(grid):
     npt.assert_array_equal(result[3:6, 3:5], bggrid[3:6, 3:5])
 
 
+def test_grdfill_hole(grid, expected_grid):
+    """
+    Test grdfill with different hole values.
+    """
+    # Prepare for a grid with -99999 as holes.
```
yvonnefroehlich: Maybe this sounds nicer.

```suggestion
    # Prepare for a grid with a node value of -99999 for holes.
```

pygmt/tests/test_grdfill.py
```diff
@@ -114,6 +114,31 @@ def test_grdfill_gridfill_dataarray(grid):
...
+    grid_no_nan = grdfill(grid=grid, constantfill=-99999)
+    assert not np.isnan(grid_no_nan).any()
+    assert -99999 in grid_no_nan
+    # Now that we have a grid with -99999 as holes, fill it with a constant value of 20.
```
yvonnefroehlich: Strickely speaking it is plural.

```suggestion
    # Now that we have a grid with -99999 as holes, fill them with a constant value of 20.
```

pygmt/tests/test_grdfill.py
```diff
@@ -114,6 +114,31 @@ def test_grdfill_gridfill_dataarray(grid):
...
+    result = grdfill(grid=grid_no_nan, constantfill=20, hole=-99999)
+
+    # check information of the output grid
+    assert isinstance(result, xr.DataArray)
+    assert result.gmt.gtype is GridType.GEOGRAPHIC
+    assert result.gmt.registration is GridRegistration.PIXEL
+    # check information of the output grid
+    xr.testing.assert_allclose(a=result, b=expected_grid)
```
yvonnefroehlich: Maybe it is enough to have this comment once.

```suggestion
    # Check information of the output grid
    assert isinstance(result, xr.DataArray)
    assert result.gmt.gtype is GridType.GEOGRAPHIC
    assert result.gmt.registration is GridRegistration.PIXEL
    xr.testing.assert_allclose(a=result, b=expected_grid)
```

pygmt/tests/test_grdfill.py
```diff
@@ -114,6 +114,31 @@ def test_grdfill_gridfill_dataarray(grid):
     npt.assert_array_equal(result[3:6, 3:5], bggrid[3:6, 3:5])
 
 
+def test_grdfill_hole(grid, expected_grid):
+    """
+    Test grdfill with different hole values.
```
yvonnefroehlich: Hm. I am not sure if I understand what you mean with "different" here. As far as I know, there could be only one value used to define the nodes considered as holes at a time. Maybe "different" refers to the fact that the value for nodes considered as holes is set to a new value in this test?

pygmt/tests/test_grdfill.py
```diff
@@ -114,6 +114,30 @@ def test_grdfill_gridfill_dataarray(grid):
...
```
seisman:
```suggestion
    Test grdfill with a custom value (not NaN) as holes.
```

pygmt/tests/test_grdfill.py
seisman: I've rephrased it to "Test grdfill with a custom value (not NaN) as holes."

PR #4125, opened 2025-09-29
Bump to Ghostscript 10.06.0

seisman:
Check if GMT 6.5 works well with gs 10.05 and 10.06.


seisman:
There are ~10 failures with gs 10.05.1, but these failures are all caused by tiny differences (See the artifacts in https://github.com/GenericMappingTools/pygmt/actions/runs/18094291533?pr=4125). So GMT 6.5 works well with gs 10.05.1!

Edit: Same for gs 10.06.0

yvonnefroehlich (approved):

Issue #4017, opened 2025-07-27
Drop GMT 6.4.0 support and add GMT 6.6.0 support


seisman:
GMT 6.6.0 was released a few hours ago, and GMT 6.4 was released in June, 2022.

Following [our support policy](https://www.pygmt.org/dev/minversions.html),

> Drop support for GMT versions 3 years after their initial release, while ensuring at least two latest minor versions remain supported.

I think it's time to drop the GMT 6.4.0 support.

seisman:
**To-Do for bumping the GMT version in CI**:

- [x] Bump the GMT version in CI (1 PR) #4018
  - [x] Update `environment.yml`
  - [x] Update `ci/requirements/docs.yml`
  - [x] Update `.github/workflows/benchmarks.yml`
  - [x] Update `.github/workflows/cache_data.yaml`
  - [x] Update `.github/workflows/ci_doctests.yaml`
  - [x] Update `.github/workflows/ci_docs.yml`
  - [x] Update `.github/workflows/ci_tests.yaml`
  - [x] Add the legacy GMT version to `.github/workflows/ci_tests_legacy.yaml`
  - [x] Update GMT 6.x minor version for `extlinks` urls in `doc/conf.py`
  - [ ] ~~Update GMT 6.x minor version in `.github/workflows/check-links.yml`~~
- [x] Fix failing tests (1 or more PRs)
- [x] Fix failing doctests reported in the ["Doctest" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_doctests.yaml)
- [x] Fix [xfail](https://docs.pytest.org/en/stable/skipping.html#xfail-mark-test-functions-as-expected-to-fail) pytest markers on tests that are now xpass

**To-Do for bumping the minimum required GMT version**:

- [x] Bump the minimum required GMT version (1 PR) #4127
  - [x] Update `required_gmt_version` in `pygmt/clib/__init__.py`
  - [ ] ~~Update `test_get_default` in `pygmt/tests/test_clib.py`~~
  - [ ] ~~Update minimum required versions in `doc/minversions.md`~~
  - [x] Remove unsupported GMT version from `.github/workflows/ci_tests_legacy.yaml`
- [x] Remove [xfail](https://docs.pytest.org/en/stable/skipping.html#xfail-mark-test-functions-as-expected-to-fail) pytest markers on tests that are now xpass
[...]

Issue #4100, opened 2025-09-18
Release PyGMT v0.17.0


seisman:
**Release**: [v0.x.x](https://github.com/GenericMappingTools/pygmt/milestone/25)
**Scheduled Date**: 2025/MM/DD
**Pull request due date**: 20YY/MM/DD
**DOI**: `10.5281/zenodo.XXXXXXX`
**Announcement draft**: https://hackmd.io/@pygmt/xxxxxxxx

**Priority PRs/issues to complete prior to release**

- [ ] Wrap X ()
- [ ] Wrap Y ()

**Before release**:

- [ ] Check [SPEC 0](https://scientific-python.org/specs/spec-0000/) to see if we need to bump the minimum supported versions of GMT, Python and core package dependencies (NumPy, pandas, Xarray)
- [ ] Review the ["PyGMT Team" page](https://www.pygmt.org/dev/team.html)
- [ ] README looks good on TestPyPI. Visit [TestPyPI](https://test.pypi.org/project/pygmt/#history), click the latest pre-release, and check the homepage.
- [ ] Check to ensure that:
  - [ ] Deprecated workarounds/codes/tests are removed. Run `grep "# TODO" **/*.py` to find all potential TODOs.
  - [ ] All tests pass in the ["GMT Legacy Tests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_tests_legacy.yaml)
  - [ ] All tests pass in the ["GMT Dev Tests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_tests_dev.yaml)
  - [ ] All tests pass in the ["Doctests" workflow](https://github.com/GenericMappingTools/pygmt/actions/workflows/ci_doctests.yaml)
- [ ] Update warnings in `pygmt/_show_versions.py` as well as notes in [Not working transparency](https://www.pygmt.org/dev/install.html#not-working-transparency) regarding GMT-Ghostscript incompatibility
- [ ] Reserve a DOI on [Zenodo](https://zenodo.org) by clicking on "New Version"
- [ ] Finish up the "Changelog entry for v0.x.x" Pull Request (Use the previous changelog PR as a reference)
- [ ] Run `make codespell` to check common misspellings. If there are any, either fix them or add them to `ignore-words-list` in `pyproject.toml`
- [ ] Draft the announcement on https://hackmd.io/@pygmt

**Release**:

[...]

PR #4127, opened 2025-09-30
Drop GMT 6.4 support 

seisman:
Following our [support policy](https://www.pygmt.org/v0.16.0/minversions.html), we should drop support for GMT versions older than three years.

Address https://github.com/GenericMappingTools/pygmt/issues/4017.

### DIFF
--- a/.github/workflows/ci_tests_legacy.yaml
+++ b/.github/workflows/ci_tests_legacy.yaml
@@ -84,11 +84,7 @@ jobs:
           # Download cached files to ~/.gmt directory and list them
           gh run download --name gmt-cache --dir ~/.gmt/
           # Change modification times of the two files, so GMT won't refresh it.
-          # The two files are in the `~/.gmt/server` directory for GMT<=6.4, and in the
-          # `~/.gmt` directory for GMT>=6.5.
-          mkdir -p ~/.gmt/server/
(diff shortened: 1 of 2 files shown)


seisman:
Triggered "GMT Legacy Tests" runs at https://github.com/GenericMappingTools/pygmt/actions/runs/18114613055.

.github/workflows/ci_tests_legacy.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
           create-args: >-
             python=3.11
             gmt=${{ matrix.gmt_version }}
-            ghostscript<10
+            ghostscript
```
weiji14, line 63: Since ghostscript>=10 can be used now, should the note here be updated?

https://github.com/GenericMappingTools/pygmt/blob/06e75f5b982734cb3d7f08bb7873b5044e79d12a/doc/install.md?plain=1#L326-L327

Or do it in #4128?

seisman: Yes, or do it in a separate PR as mentioned in https://github.com/GenericMappingTools/pygmt/issues/4100#issuecomment-3347945572.

weiji14 (approved):
:rocket:

PR #4129, opened 2025-09-30
Update the GMT/Ghostscript compatibility notes

seisman:
We know that gs 10.03-10.06 works well with GMT.


yvonnefroehlich (commented):
That's great that there are no transparency issues with GMT and these gs versions.

PR #2402, opened 2023-03-07
Update the documentation template to show class attributes

seisman:
**Description of proposed changes**

**Preview** at https://pygmt-dev--2402.org.readthedocs.build/en/2402/api/generated/pygmt.GMTDataArrayAccessor.html#pygmt.GMTDataArrayAccessor.gtype

Python class properties shown for GMTDataArrayAccessor

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.
- [ ] Use underscores (not hyphens) in names of Python files and directories.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #4131, opened 2025-09-30
Make the doc pages for the Box and Pattern classes consistent

yvonnefroehlich:

Make the doc pages for the `Box` abd `Pattern` classes consistent, to use "Parameters" as section.

If possible it would be nice to remove the "Attribute" section, as there is not entery.
But keep it for the `pygm.Figure` class.

**Preview**:
* `Box`
   * OLD: https://www.pygmt.org/dev/api/generated/pygmt.params.Box.html
   * NEW: https://pygmt-dev--4131.org.readthedocs.build/en/4131/api/generated/pygmt.params.Box.html
* `Pattern`
   * OLD: https://www.pygmt.org/dev/api/generated/pygmt.params.Pattern.html
   * NEW: https://pygmt-dev--4131.org.readthedocs.build/en/4131/api/generated/pygmt.params.Pattern.html
* `Figure`  (should still have the "Attribute" section)
   * OLD: https://www.pygmt.org/dev/api/generated/pygmt.Figure.html
   * NEW: 

**Reminders**

**Slash Commands**

### DIFF
--- a/doc/_templates/autosummary/class.rst
+++ b/doc/_templates/autosummary/class.rst
@@ -4,7 +4,7 @@
 
 .. autoclass:: {{ objname }}
 
-{% if attributes %}
+{% if '_aliases' not in attributes %}
 .. rubric:: Attributes
 
 {% for item in attributes %}
--- a/doc/_templates/autosummary/class.rst
+++ b/doc/_templates/autosummary/class.rst
@@ -4,7 +4,7 @@
 
 .. autoclass:: {{ objname }}
 
-{% if '_aliases' not in attributes %}
+{% if attributes %}
 .. rubric:: Attributes
 
 {% for item in attributes %}
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -151,7 +151,7 @@ jobs:
           GH_TOKEN: ${{ github.token }}
 
       - name: Install uv
-        uses: astral-sh/setup-uv@b75a909f75acd358c2196fb9a5f1299a9a8868a4 # v6.7.0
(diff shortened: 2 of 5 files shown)


weiji14:
> If possible it would be nice to remove the "Attribute" section, as there is not entery.

The Attributes section was added in #2402, and it seems to count `@property` methods that start with an underscore. Will need to update the code here:

https://github.com/GenericMappingTools/pygmt/blob/35f06da7a32c3aaf1559c62b4e95dbfe9ab60152/doc/_templates/autosummary/class.rst?plain=1#L7-L14

If you change that L7 from `{% if attributes %} ` to `{% if '_aliases' not in attributes %}`, the Attributes section should be hidden for these Param classes.

doc/_templates/autosummary/class.rst
```diff
@@ -4,7 +4,7 @@
 
 .. autoclass:: {{ objname }}
 
-{% if attributes %}
+{% if '_aliases' not in attributes %}
```
weiji14: Actually, we might need a better solution, because other class pages now show an empty 'Attributes' section :sweat_smile: E.g. in these places:
- https://pygmt-dev--4131.org.readthedocs.build/en/4131/api/generated/pygmt.config.html
- https://pygmt-dev--4131.org.readthedocs.build/en/4131/api/generated/pygmt.GMTBackendEntrypoint.html#pygmt.GMTBackendEntrypoint

Maybe revert this change, and we'll fix it in a separate PR?

yvonnefroehlich: Ah, OK.
Yeah, we can handel this later in a separate PR.

weiji14 (approved):

PR #4150, opened 2025-10-10
Add support for Python 3.14

seisman:
Python 3.14 has been released on 7 Oct 2025, changelog is at https://docs.python.org/3.14/whatsnew/3.14.html

Previous PR for Python 3.13 at #3490.

- [ ] Update the branch protection rules after merging this PR.

### DIFF
--- a/.github/workflows/type_checks.yml
+++ b/.github/workflows/type_checks.yml
@@ -44,16 +44,13 @@ jobs:
         uses: actions/setup-python@v6.1.0
         with:
           python-version: '3.14'
-          # Need to install four groups of packages:
+          # Need to install following packages:
           # 1. required packages
-          # 2. optional packages
-          # 3. type checker and stub packages
-          # 4. other packages that are used somewhere in PyGMT
+          # 2. type checker and stub packages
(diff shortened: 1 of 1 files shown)


weiji14:
If I understand correctly, most packages like `pyogrio` (used by `geopandas`) and `rasterio` (used by `rioxarray`) already have Python 3.14 packages on conda-forge, but there is a conflict on some Python 3.14 packages being built on `libgdal-core=3.11` only, while rasterio is still stuck on `libgdal-core=3.10` (xref https://github.com/conda-forge/rasterio-feedstock/pull/332). So might still need to wait for https://github.com/rasterio/rasterio/issues/3406

seisman:
All workflows now pass with Python 3.14, except the "Static Type Checks" workflow, which fails because rasterio 1.4.3 doesn't provide a Python 3.14 binary wheel on PyPI.

But, it's likely we don't have to install optional dependencies when running mypy.

Edit: I've removed all optional dependencies in c73aac8 and the "Static Type Checks" workflow now passed. Don't remember why I added them in PR #2808.

PR #4172, opened 2025-10-23
pygmt.project: Migrate the 'center'/'endpoint'/'pole'/'width'/'length' parameters to the new alias system

seisman:
Parameters `center`/`endpoint`/`pole`/`width` all accept a tuple of two values. This PR migrates them to the new alias system.

The `length` parameter is slightly complicated in that it can accept a tuple or a string `"w"` (with the long name `"limit"` from https://github.com/GenericMappingTools/gmt/blob/88b836ea9f6429e4d94fbadd508ef97497671eb1/src/longopt/project_inc.h#L37).

### DIFF
--- a/pygmt/src/project.py
+++ b/pygmt/src/project.py
@@ -220,8 +220,8 @@ def project(  # noqa: PLR0913
         - :class:`pandas.DataFrame` or :class:`numpy.ndarray` if ``outfile`` is not set
           (depends on ``output_type``)
     """
-    if kwargs.get("C") is None:
-        msg = "The 'center' parameter must be specified."
+    if kwargs.get("C", center) is None:
+        msg = "Parameter 'center' must be specified."
         raise GMTInvalidInput(msg)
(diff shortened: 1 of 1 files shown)


pygmt/src/project.py
```diff
@@ -241,7 +236,13 @@ def project(
     if output_type == "pandas" and kwargs.get("G") is not None:
         column_names = list("rsp")
 
-    aliasdict = AliasSystem().add_common(
+    aliasdict = AliasSystem(
+        C=Alias(center, name="center", sep="/", size=2),
+        E=Alias(endpoint, name="endpoint", sep="/", size=2),
+        L=Alias(length, name="length", sep="/", size=2, mapping={"limit": "w"}),
```
seisman, line 242: I've added doctests in PR https://github.com/GenericMappingTools/pygmt/pull/4173 to ensure that parameters `sep` and `mapping` can coexist.

michaelgrund (approved):

PR #4173, opened 2025-10-23
_to_string: Add doctest to ensure it works for both Sequence and Mappable objects

seisman:
Motivated by #4172.


yvonnefroehlich:
Locally I do not have a download issue with the river data and do not ge this error, which leads to the failing doc builds
> requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: https://www.eea.europa.eu/data-and-maps/data/wise-large-rivers-and-large-lakes/zipped-shapefile-with-wise-large-rivers-vector-line/zipped-shapefile-with-wise-large-rivers-vector-line/at_download/file

PR #4174, opened 2025-10-23
Figure.subplot: Migrate the 'margins' parameter to the new alias system

seisman:


pygmt/src/subplot.py
```diff
@@ -108,21 +109,20 @@ def subplot(
...
         scales, bars, text, etc.
     {projection}
-    margins : str or list
-        This is margin space that is added between neighboring subplots (i.e.,
-        the interior margins) in addition to the automatic space added for tick
-        marks, annotations, and labels. The margins can be specified as either:
+    margins
+        This is margin space that is added between neighboring subplots (i.e., the
+        interior margins) in addition to the automatic space added for tick marks,
+        annotations, and labels. The margins can be specified as either:
```
yvonnefroehlich:
```suggestion
        Margin space that is added between neighboring subplots (i.e., the
        interior margins) in addition to the automatic space added for tick marks,
        annotations, and labels. The margins can be specified as either:
```

pygmt/src/subplot.py
```diff
@@ -108,21 +109,20 @@ def subplot(
...
 
         - a single value (for same margin on all sides). E.g. ``"5c"``.
-        - a pair of values (for setting separate horizontal and vertical
-          margins). E.g. ``["5c", "3c"]``.
-        - a set of four values (for setting separate left, right, bottom, and
-          top margins). E.g. ``["1c", "2c", "3c", "4c"]``.
+        - a pair of values (for setting separate horizontal and vertical margins). E.g.,
+          ``("5c", "3c")``.
+        - a set of four values (for setting separate left, right, bottom, and top
+          margins). E.g., ``("1c", "2c", "3c", "4c")``.
```
yvonnefroehlich: I think we can skipe "setting" here.
```suggestion
        - a pair of values (for separate horizontal and vertical margins). E.g.,
          ``("5c", "3c")``.
        - a set of four values (for separate left, right, bottom, and top margins). E.g.,
          ``("1c", "2c", "3c", "4c")``.
```

PR #4175, opened 2025-10-23
Figure.set_panel: Migrate the 'panel' parameter to the new alias system

seisman:


pygmt/src/subplot.py
```diff
@@ -266,6 +264,11 @@ def set_panel(
 
     with Session() as lib:
         lib.call_module(
-            module="subplot", args=["set", str(panel), *build_arg_list(aliasdict)]
+            module="subplot",
+            args=[
+                "set",
+                _to_string(panel, sep=",", size=2),
```
seisman: We can't directly use the Alias/AliasSystem classes like in other wrappers, because `panel` is an argument of a GMT option. The GMT CLI syntax is:
```
gmt subplot set <panel>
```
so we have to use the private `_to_string` method here.

seisman: Not ideal. Converting to draft PR.

seisman: > so we have to use the private `_to_string` method here.

Now using `Alias._value` instead.

PR #4182, opened 2025-11-01
AliasSystem: Migrate the 'registration' parameter to the new alias system and support descriptive arguments

seisman:
GMT documentation: https://docs.generic-mapping-tools.org/6.6/gmt.html#nodereg-full

The GMT CLI syntax is `-r[g|p]`. Here, we use the descriptive values "gridline" and "pixel".

I'm still thinking if we should support the `GridRegistration` enums (https://www.pygmt.org/dev/api/generated/pygmt.enums.GridRegistration.html#pygmt.enums.GridRegistration), i.e.,
```
import pygmt
from pygmt.enums import GridRegistration

pygmt.xyz2grd(..., registration=GridRegistration.PIXEL)
```
@GenericMappingTools/pygmt-maintainers What do you think?

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -40,7 +40,7 @@ jobs:
 
     - name: Link Checker
       id: lychee
-      uses: lycheeverse/lychee-action@885c65f3dc543b57c898c8099f4e08c8afd178a2 # v2.6.1
+      uses: lycheeverse/lychee-action@a8c4c7cb88f0c7386610c35eb25108e448569cb0 # v2.7.0
       with:
         fail: false  # Don't fail action on broken links
         output: /tmp/lychee-out.md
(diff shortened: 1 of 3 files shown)


pygmt/src/binstats.py
```diff
@@ -49,6 +48,7 @@ def binstats(
     ] = "number",
     quantile_value: float = 50,
     region: Sequence[float | str] | str | None = None,
+    registration: Literal["gridline", "pixel"] | bool = False,
```
weiji14, line 51: Docstring for registration should be updated to 1) add bool type hint, and 2) indicate that `registration=True` means pixel registration.

https://github.com/GenericMappingTools/pygmt/blob/25f265647de8097758ba7f42a141a50706590650/pygmt/helpers/decorators.py#L279-L283

seisman: > add bool type hint,

Type hints are now added via function signatures, not in docstrings.

Issue #4147, opened 2025-10-09
Running "dvc pull" now requires authentication


seisman:
The Contributors' guides (https://www.pygmt.org/dev/contributing.html#using-data-version-control-dvc-to-manage-test-images) mention that:

> To pull or sync files from the dvc remote to your local repository, use the commands below.
> ```
> dvc status  # should report any files 'not_in_cache'
> dvc pull  # pull down files from DVC remote cache (fetch + checkout)
> ```

> To push or sync changes from your local repository up to the dvc remote at DAGsHub, you will first need to set up authentication using the commands below.
> ```
> dvc remote modify upstream --local auth basic
> dvc remote modify upstream --local user "$DAGSHUB_USER"
> dvc remote modify upstream --local password "$DAGSHUB_PASS"
> ```

So, previously, authentication is not required when running `dvc pull`, but is required for `dvc push`. It makes sense for a public DVC repository like pygmt or gmt.

However, it appears that `dvc pull` requires authentication recently, which causes issues such as https://github.com/GenericMappingTools/pygmt/pull/4078 and creates additional barriers for contributors (e.g., 
https://github.com/GenericMappingTools/gmt/issues/8807 and https://github.com/GenericMappingTools/pygmt/pull/4135#pullrequestreview-3292651421).

It's unclear if the authentication changes are intentional or not.

joa-quim:
So, this is why the GMT tests are all failing now.

seisman:
> So, this is why the GMT tests are all failing now.

The GMT CI already work as expected after my fix in https://github.com/GenericMappingTools/gmt/pull/8799.

joa-quim:
I'll have to confirm it though, because I think I saw total failures (related to DVC) more recent than that fix.

joa-quim:
Yep, [#8804](https://github.com/GenericMappingTools/gmt/pull/8804) was where I saw:

<img width="733" height="182" alt="Image" src="https://github.com/user-attachments/assets/c5e5efdd-e95c-483d-a335-02f1261e6bc0" />

seisman:
In CI, we use a DagsHub token to pull baseline images from DagsHub, but it seems the token can't be used in PRs from forks.

joa-quim:
Ah, not ideal but OK.

weiji14:
> it seems the token can't be used in PRs from forks.

[...]

PR #4184, opened 2025-11-02
CI: Add a new workflow to cache dvc cache

seisman:
For new contributors, it's frustrating to see that so mamy workflows fail due to the DVC issue reported in https://github.com/GenericMappingTools/pygmt/issues/4147.

This PR adds a workflow for this issue. It adds a new workflow `cache_dvc.yaml` which pulls the baseline images from DagsHub and uploads the `.dvc/cache` file as GitHub artifacts.

For PRs from forks, the `dvc pull` command fails, so we can download the dvc cache from the GitHub artifact.

Below are workflows that use `dvc pull` and may fail:

- [x] `ci_tests_dev.yaml`
- [x] `ci_tests_dev.yaml`
- [ ] ~~`dvc-diff.yaml`~~: The cache won't work in this workflow, since the dvc cache reflect baseline images in the main branch, but this workflow needs baseline images in both the main and the current branch. But it's unlikely that PRs from new contributors will trigger this workflow, since they can't do 'dvc push' either. So no need to apply the workaround
- [ ] ~~`release-baseline-images.yml`~~: Doesn't run in PRs, so no need to apply the workaround

Tests

- The workflow run succeeded and the `.dvc/cache` directory was uploaded https://github.com/GenericMappingTools/pygmt/actions/runs/19008233397?pr=4184
- The "Download DVC cache as artifacts from GitHub" step is skipped when `dvc pull` command succeeds (https://github.com/GenericMappingTools/pygmt/actions/runs/19008233364/job/54285305945?pr=4184)
- In commit 6d9be5734 (with a fix in bd4b3483b8b6735deb587415ce88daacb1cc99b0), I made changes to make the "DVC pull" step fail with an exit code of 1. As you can see, the workflows downloads the cache from GitHub artifact and then the workflows pass.

### DIFF
--- a/doc/maintenance.md
+++ b/doc/maintenance.md
@@ -95,9 +95,9 @@ controlled by workflow files located in `.github/workflows`. Here we briefly
 summarize the functions of the workflows. Please refer to the comments in the
 workflow files for more details.
 
-- `benchmarks.yml`: Benchmarks the execution speed of tests to track performance of PyGMT functions
-- `cache_data.yaml`: Cache GMT remote data files and uploads as artifacts
-- `cache_dvc.yaml`: Cache baseline images tracked by DVC and upload as artifacts.
+- `benchmarks.yml`: Benchmark the execution speed of tests to track performance of PyGMT functions
+- `cache_data.yaml`: Cache GMT remote data files and upload as artifacts
+- `cache_dvc.yaml`: Cache baseline images tracked by DVC and upload as artifacts
 - `check-links.yml`: Check links in the repository and documentation
 - `ci_docs.yml`: Build documentation on Linux/macOS/Windows and deploy to GitHub
 - `ci_doctest.yaml`: Run all doctests on Linux/macOS/Windows


.github/workflows/cache_dvc.yaml
```diff
@@ -0,0 +1,59 @@
...
+# any changes to the baseline images.
+#
+name: Cache DVC
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'pygmt/tests/baseline/*.png.dvc'
+  pull_request:  # For testing only.
```
seisman:
```suggestion
  # pull_request:  # For testing only.
```

.github/workflows/ci_tests.yaml
```diff
@@ -163,12 +163,20 @@ jobs:
...
+        id: dvc-pull
+        continue-on-error: true
         run: |
           uv run dvc remote modify upstream url https://${DAGSHUB_TOKEN}@dagshub.com/GenericMappingTools/pygmt.dvc --local
           uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
         env:
           DAGSHUB_TOKEN: ${{ secrets.DAGSHUB_TOKEN }}
 
+      - name: Download DVC cache as artifacts from GitHub
+        if: steps.dvc-pull.outcome == 'failure'
```
seisman, line 175: Reference: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context

.github/workflows/ci_tests.yaml
```diff
@@ -163,12 +163,23 @@ jobs:
 
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
+        id: dvc-pull
+        continue-on-error: true
         run: |
           uv run dvc remote modify upstream url https://${DAGSHUB_TOKEN}@dagshub.com/GenericMappingTools/pygmt.dvc --local
-          uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
+          # uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
+          exit 1
```
seisman:
```suggestion
          # uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
          exit 1
```
```suggestion
          uv run dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
```

.github/workflows/ci_tests_dev.yaml
```diff
@@ -161,12 +161,23 @@ jobs:
 
       # Pull baseline image data from dvc remote (DAGsHub)
       - name: Pull baseline image data from dvc remote
+        id: dvc-pull
+        continue-on-error: true
         run: |
           dvc remote modify upstream url https://${DAGSHUB_TOKEN}@dagshub.com/GenericMappingTools/pygmt.dvc --local
-          dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
+          # dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
+          exit 1
```
seisman:
```suggestion
          dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
```

.github/workflows/cache_dvc.yaml
```diff
@@ -0,0 +1,59 @@
...
+  workflow_dispatch:
+  # Schedule runs on 12 noon every Sunday
+  schedule:
+    - cron: '0 12 * * 0'
+
+permissions: {}
+
+jobs:
+  dvc_cache:
```
weiji14, line 29: Could probably this as a separate job under `.github/workflows/cache_data.yaml` if want to reduce the number of workflow files.

doc/maintenance.md
```diff
@@ -97,6 +97,7 @@ workflow files for more details.
 
 - `benchmarks.yml`: Benchmarks the execution speed of tests to track performance of PyGMT functions
 - `cache_data.yaml`: Cache GMT remote data files and uplodas as artifacts
```
weiji14: Typo
```suggestion
- `cache_data.yaml`: Cache GMT remote data files and uploads as artifacts
```

weiji14 (approved):
Thank you! Just a couple of suggestions.

.github/workflows/cache_dvc.yaml
seisman: These two jobs have different trigger conditions. The `cache_data` workflow is usually scheduled to run weekly but the `cache_dvc` workflow needs to run when `.dvc` files change.

weiji14: Ah yes, ok to keep them separate then.

PR #4185, opened 2025-11-02
pygmt.grdsample: Migrate the 'translate' parameter to the new alias system

seisman:


pygmt/src/grdsample.py
```diff
@@ -61,9 +56,10 @@ def grdsample(
     {outgrid}
     {spacing}
     {region}
-    translate : bool
-        Translate between grid and pixel registration; if the input is
-        grid-registered, the output will be pixel-registered and vice-versa.
+    translate
+        Translate between grid and pixel registration; if the input is grid-registered,
+        the output will be pixel-registered and vice-versa. This is a destructive grid
+        change.
```
weiji14: Follow upstream docs at https://docs.generic-mapping-tools.org/6.5/grdsample.html#t

```suggestion
        the output will be pixel-registered and vice-versa. This is a *destructive* grid
        change; see :gmt-docs:`reference/options.html#switch-registrations`.
```

seisman:
In addition to `grdsample`, there are other modules (grdedit, grdfilter, dimfilter) that also have the same option. 

Currently, it's aliased to `translate` in `grdsample`, but is aliased to `toggle` in `grdfilter`. GMT recommends `toggle`/`toggle_registration` 

https://github.com/GenericMappingTools/gmt/blob/21a81c7354a76e17584870f8ef0cf73ad5d23c1c/src/longopt/grdsample_inc.h#L30

Maybe we should use `toggle` consistently?

weiji14:
Sure, `toggle` sounds good.

PR #4195, opened 2025-11-10
pygmt.grdproject: Improve the checking of required parameter 'projection'/'J'

### LINKED ISSUE #4189: Consistent parameter validation in wrappers
In PyGMT wrappers, we typically need to validate parameters, for example, to ensure that required parameters are provided and that mutually exclusive parameters are not used together. 

Currently, parameter validation is done inconsistently across wrappers. This is further complicated because most parameters are aliases of GMT’s single-letter options, especially after migrating to the new alias system.

1. Check short-form parameters in `kwargs` directly. e.g.,

https://github.com/GenericMappingTools/pygmt/blob/4c7e3569f69a00d85c1a4b81635b1e3b73b7347e/pygmt/src/filter1d.py#L114-L116

- **Pros**: Early validation and simple.
- **Cons**: Doesn’t work with the new alias system.

2. Check long-form parameters directly, e.g., 

https://github.com/GenericMappingTools/pygmt/blob/4c7e3569f69a00d85c1a4b81635b1e3b73b7347e/pygmt/src/grdproject.py#L118-L120

- **Pros**: Simple and intuitive.
- **Cons**: Doesn’t catch users passing short-form options like R="...".

3. Check both long-form and short-form parameters in `kwargs`,e.g.,

https://github.com/GenericMappingTools/pygmt/blob/4c7e3569f69a00d85c1a4b81635b1e3b73b7347e/pygmt/src/grdlandmask.py#L111-L113

Here `kwargs.get("R", region)` checks both `region` and `R`.

**Pros**: 
- Covers both long- and short- forms
- Works before initializing `AliasSystem` so early validation
**Cons**:
- Slightly hard to read and sometimes we may forget to check both

4. Check the `aliasdict` after initializing `AliasSystem`, e.g., `aliasdict.get("r")` in PR #4183.

https://github.com/GenericMappingTools/pygmt/blob/c614d6f23224761fc4416bb0094194e265e0e663/pygmt/src/grdsample.py#L110-L112

**Pros**: 
- Handles both long and short forms
- Readale and no need to check the long-form explicitely
**Cons**
- Must be done after initializing `AliasSystem`
- Not all parameters are used in the `Aliasystem`, which means sometimes we still need to use option 2

I guess option 3 is better?

seisman:

See PR #4189 for context. 

I've checked all wrappers and it turns out that we only need to fix `grdproject`.

Closes #4189.

### DIFF
--- a/pygmt/src/grdproject.py
+++ b/pygmt/src/grdproject.py
@@ -116,8 +116,8 @@ def grdproject(
     >>> # Project the geographic gridded data onto a rectangular grid
     >>> new_grid = pygmt.grdproject(grid=grid, projection="M10c", region=region)
     """
-    if projection is None:
-        msg = "The projection must be specified."
+    if kwargs.get("J", projection) is None:
+        msg = "Parameter 'projection' must be specified."
         raise GMTInvalidInput(msg)
 
     aliasdict = AliasSystem().add_common(


yvonnefroehlich (approved):
But did not double check the other wappers.

PR #4199, opened 2025-11-11
pygmt.grdproject: Migrate the 'center' parameter to the alias system

seisman:


pygmt/src/grdproject.py
```diff
@@ -71,13 +72,12 @@ def grdproject(
...
-        is relative to lower left corner]. Optionally, add offsets in the
-        projected units to be added (or subtracted when ``inverse`` is set) to
-        (from) the projected coordinates, such as false eastings and
-        northings for particular projection zones [Default is ``[0, 0]``].
+    center
+        If ``True``, let projected coordinates be relative to projection center [Default
+        is relative to lower left corner]. Optionally, set offsets (*dx*, *dy*) in the
+        projected units to be added (or subtracted when ``inverse`` is set) to (from)
+        the projected coordinates, such as false eastings and northings for particular
+        projection zones [Default is ``(0, 0)``].
```
yvonnefroehlich:
```suggestion
    center
        If ``True``, let the projected coordinates be relative to the projection center [Default
        is relative to the lower left corner]. Optionally, set offsets (*dx*, *dy*) in the
        projected units to be added (or subtracted when ``inverse`` is set) to (from)
        the projected coordinates, such as false eastings and northings for particular
        projection zones [Default is ``(0, 0)``].
```

Issue #4168, opened 2025-10-16
Add the Position class for GMT embellishment placement


seisman:
Currently, GMT provides 9 different embellishments

- Map scale by `basemap -L` or `coast -L`
- Directional rose by `basemap -Td` or `coast -Td`
- Magnetic rose by `basemap -Tm` or `coast -Tm`
- Colorbar by `colorbar`
- Legend by `legend`
- Raster images by `image`
- GMT logo by `logo`
- Map inset by `inset`
- Vertical scale of wiggles in `wiggle`

Refer to https://docs.generic-mapping-tools.org/6.5/reference/features.html#plot-embellishments for details.

These embellishments have different properties thus different option flags/modifiers, but they all share the same syntax for positioning, as shown below:

- [ ] Map scale: `-L[g|j|J|n|x]refpoint+wlength[e|f|k|M|n|u][+aalign][+c[[slon/]slat]][+f][+jjustify][+l[label]][+odx[/dy]][+u][+v]` #4015
- [ ] Directional rose: `-Td[g|j|J|n|x]refpoint[+f|F[level]][+jjustify][+l[w,e,s,n]][+odx[/dy]][+wwidth]` #4025
- [ ] Magnetic rose: `-Tm[g|j|J|n|x]refpoint[+ddec[/dlabel]]][+ipen][+jjustify][+l[w,e,s,n]][+ppen][+tints][+odx[/dy]][+wwidth]` #4051
- [ ] Colorbar: `-D[g|j|J|n|x]refpoint[+wlength[/width]][+e[b|f][length]][+h|v][+jjustify][+m[a|c|l|u]][+n[txt]][+odx[/dy]][+r]` #4048
- [ ] Legend: `-D[g|j|J|n|x]refpoint+wwidth[/height][+jjustify][+lspacing][+odx[/dy]]` #4046
- [ ] Raster image: `-D[g|j|J|n|x]refpoint+rdpi[c]+w[-]width[/height][+jjustify][+nnx[/ny] ][+odx[/dy]]` #4045
- [ ] GMT logo: `-D[g|j|J|n|x]refpoint[+hheight|+wwidth][+jjustify][+odx[/dy]]` #4014
- [ ] Map inset: `-D[g|j|J|n|x]refpoint+wwidth[/height][+jjustify][+odx[/dy]]` #4047
[...]

PR #4212, opened 2025-11-16
Add the Position class for GMT embellishment placement

seisman:
See https://github.com/GenericMappingTools/pygmt/issues/4168#issuecomment-3537572950 for context.

This PR implements the `Position` class.

Usage:
```python
>>> import pygmt
>>> from pygmt.params import Position
>>> fig = pygmt.Figure()
>>> fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=True)
>>> fig.logo(
...:     position=Position((3, 3), cstype="mapcoords", anchor="ML", offset=(0.2, 0.2)),
...:     box=True,
...: )
>>> fig.show()
```

**GMT Documentation**: https://docs.generic-mapping-tools.org/dev/reference/features.html#reference-and-anchor-point-specification

**Preview**: https://pygmt-dev--4212.org.readthedocs.build/en/4212/api/generated/pygmt.params.Position.html

### DIFF
--- a/examples/gallery/embellishments/solar.py
+++ b/examples/gallery/embellishments/solar.py
@@ -32,7 +32,7 @@ for terminator in ["day_night", "civil", "nautical", "astronomical"]:
         terminator_datetime=reference_time,
         # Set the fill for the night area to navy blue with 85 % transparency
         fill="navyblue@85",
-        pen="0.5p",  # Set the outline to be 0.5 points thick
+        pen="0.5p",  # Set the outline to be 0.5-point thick
     )
 
 fig.show()
--- a/examples/gallery/images/cross_section.py
+++ b/examples/gallery/images/cross_section.py
@@ -46,7 +46,7 @@ fig.colorbar(
     # directions, respectively; move the x-label above the horizontal colorbar ("+ml")
     position="jBR+o0.7c/0.8c+h+w5c/0.3c+ml",
     # Add a box around the colobar, filled in white and a 30 % transparency, with a
-    # 0.8-points thick, black, outline.
+    # 0.8-point thick, black, outline.
     box=Box(pen="0.8p,black", fill="white@30"),
     # Add x- and y-labels ("+l")
     frame=["x+lElevation", "y+lm"],
@@ -103,7 +103,7 @@ fig.plot(
     x=[0, 15],
     y=[0, 0],
     fill="lightblue",  # Fill the polygon in "lightblue"
-    pen="0.25p,black,solid",  # Draw a 0.25-points thick, black, solid outline
+    pen="0.25p,black,solid",  # Draw a 0.25-point thick, black, solid outline
     close="+y-8000",  # Force closed polygon
 )
 
--- a/examples/gallery/maps/shorelines.py
+++ b/examples/gallery/maps/shorelines.py
@@ -11,6 +11,6 @@ import pygmt
 fig = pygmt.Figure()
 # Make a global Mollweide map with automatic ticks
 fig.basemap(region="g", projection="W15c", frame=True)
-# Display the shorelines as black lines with 0.5 point thickness
+# Display the shorelines as black lines with 0.5-point thickness
 fig.coast(shorelines="0.5p,black")
 fig.show()
--- a/examples/gallery/seismology/meca.py
+++ b/examples/gallery/seismology/meca.py
@@ -45,7 +45,7 @@ fig.meca(
     # Fill extensive quadrants with color "cornsilk"
     # [Default is "white"]
     extensionfill="cornsilk",
-    # Draw a 0.5 points thick dark gray ("gray30") solid outline via
+    # Draw a 0.5-point thick dark gray ("gray30") solid outline via
     # the pen parameter [Default is "0.25p,black,solid"]
     pen="0.5p,gray30,solid",
 )
--- a/examples/tutorials/advanced/3d_perspective_image.py
+++ b/examples/tutorials/advanced/3d_perspective_image.py
@@ -91,7 +91,7 @@ fig.grdview(
     surftype="s",
     cmap="geo",
     plane="1000+ggrey",
-    # Set the contour pen thickness to "0.1p"
+    # Set the contour pen thickness to 0.1 points
     contour_pen="0.1p",
 )
 fig.colorbar(perspective=True, frame=["a500", "x+lElevation", "y+lm"])
--- a/examples/tutorials/advanced/cartesian_histograms.py
+++ b/examples/tutorials/advanced/cartesian_histograms.py
@@ -120,7 +120,7 @@ fig.histogram(
     frame=["WSne", "xaf10", "ya1f1+lCounts"],
     data=data02,
     series=10,
-    # Draw a 1.5-points thick, dashed outline in "orange"
+    # Draw a 1.5-point thick, dashed outline in "orange"
     pen="1.5p,orange,dashed",
     histtype=0,
     stairs=True,
--- a/examples/tutorials/advanced/legends.py
+++ b/examples/tutorials/advanced/legends.py
@@ -89,7 +89,7 @@ fig.plot(x=0, y=0, style="c0.25c", fill="orange", label="orange circle")
 fig.plot(x=1, y=0, style="t0.3c", fill="pink", pen="black", label="pink triangle")
 fig.plot(x=[-3, 3], y=[-2, -2], pen="darkred", label="darkred line")
 
-# Add a box with a 2-points thick blue, solid outline and a white fill with a
+# Add a box with a 2-point thick blue, solid outline and a white fill with a
 # transparency of 30 percent ("@30").
 fig.legend(position="jTL+o0.3c/0.2c", box=Box(pen="2p,blue", fill="white@30"))
 
--- a/examples/tutorials/advanced/working_with_panel.py
+++ b/examples/tutorials/advanced/working_with_panel.py
@@ -50,7 +50,7 @@ fig.coast(
     frame="g30",  # Add frame and gridlines in steps of 30 degrees on top
     land="gray",  # Color land masses in "gray"
     water="lightblue",  # Color water masses in "lightblue"
-    # Add coastlines with a 0.25 points thick pen in "gray50"
+    # Add coastlines with a 0.25-point thick pen in "gray50"
     shorelines="1/0.25p,gray50",
 )
 fig.show()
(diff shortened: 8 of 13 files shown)


pygmt/params/position.py
```diff
@@ -0,0 +1,87 @@
...
+    >>> fig.logo(
+    ...     position=Position((3, 3), type="mapcoords", anchor="ML", offset=(0.2, 0.2)),
+    ...     box=True,
+    ... )
+    >>> fig.show()
+    """
+
+    #: Specify the reference point on the plot. The method of defining the reference
+    #: point is controlled by ``type``, and the exact location is set by ``position``.
+    location: Sequence[float | str] | AnchorCode
```
seisman: `position` or `location`?

yvonnefroehlich: I think `location` is better as the parameter is already called `position` and we overall will have the `Position` class.

weiji14: The docstring above calls this 'reference point', so how about something like `ref` or `refpt`?

seisman: Maybe `refpoint`?

In fact, I'd prefer to make it a positional-only parameter, so it will be used like `Position("TL", type="inside")`, `Position((1, 2))`. 

In this case, the specific parameter name doesn't really matter, though `dataclasses` doesn't support positional-only attributes.

weiji14: Yeah, I did think of positional-only, but a little tricky with dataclasses as you said. Ok to got with `refpoint`, but let's see what @yvonnefroehlich thinks.

pygmt/params/position.py
```diff
@@ -0,0 +1,209 @@
...
+from pygmt._typing import AnchorCode
+from pygmt.alias import Alias
+from pygmt.exceptions import GMTValueError
+from pygmt.params.base import BaseParam
+
+
+@dataclasses.dataclass(repr=False)
+class Position(BaseParam):
+    """
+    Class for positioning embellishments on a plot.
```
seisman, line 19: The docstrings are modified from the GMT documentation at https://docs.generic-mapping-tools.org/dev/reference/features.html#reference-and-anchor-point-specification.

pygmt/params/position.py
```diff
@@ -0,0 +1,209 @@
...
+        ``location`` as (*x*, *y*) with units (e.g., inches, centimeters, points).
+        Useful for precise layout control.
+
+        **Example:** ``location=("2c", "2.5c"), type="plotcoords"``
+
+    ``type="boxcoords"`` Normalized Coordinates
+        Use normalized coordinates where (0, 0) is the lower-left corner and (1, 1) is
+        the upper-right corner of the bounding box of the current plot. Specify
+        ``location`` as (*nx*, *ny*). Useful for positioning relative to plot dimensions
+        without units.
```
seisman: It should be noted that, although the GMT documentation says that nx/ny are values in 0-1, but actually other values are also supported. For example:
```bash
gmt begin map
gmt basemap -R0/10/0/5 -JX10c/5c -Baf
gmt logo -Dn-1/-1
gmt end show
```

Image

pygmt/params/position.py
```diff
@@ -0,0 +1,209 @@
...
+
+    #: Types of the reference point. Valid values are:
+    #:
+    #: - ``"mapcoords"``: Map/Data coordinates
+    #: - ``"plotcoords"``: Plot coordinates
+    #: - ``"boxcoords"``: Normalized coordinates
+    #: - ``"inside"`` or ``"outside"``: Justification codes
+    #:
+    #: If not specified, defaults to ``"inside"`` if ``location`` is a justification
+    #: code; otherwise defaults to ``"plotcoords"``.
```
seisman: In GMT, the default is "plotcoords", i.e., the code **x**. Thus, `-D1/1` is equivalent to `-Dx1/1`. 

When `location` is an anchor code, we know that "plotcoords" makes no sense (e.g., `-DxTL`). `type` should be either "inside" or "outside". Here, I choose "inside" as the default.

pygmt/params/position.py
```diff
@@ -0,0 +1,209 @@
...
+    #: If a single value is given, the offset is applied to both x and y directions.
+    offset: Sequence[float | str] | None = None
+
+    def _validate(self):
+        """
+        Validate the parameters.
+        """
+        _valid_anchors = {f"{h}{v}" for v in "TMB" for h in "LCR"} | {
+            f"{v}{h}" for v in "TMB" for h in "LCR"
+        }
```
seisman, line 165: We probably can reuse these codes to validate anchor codes in other wrappers, e.g., the `justify` parameter in the `text` module. But I feel we can do it in the future.

pygmt/params/position.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Default to "inside" if type is not specified and location is an anchor code.
+        if self.type is None:
+            self.type = "inside" if isinstance(self.location, str) else "plotcoords"
+
+        # Validate the location based on type.
+        match self.type:
+            case "mapcoords" | "plotcoords" | "boxcoords":
+                if not isinstance(self.location, Sequence) or len(self.location) != 2:
```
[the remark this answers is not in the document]
seisman: Fixed in 10a0dfb4acb013492170b7a2acdbc6928a03a684.

pygmt/tests/test_params_position.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_params_position_anchor_offset():
+    """
+    Test the Position class with anchor and offset parameters.
+    """
+    assert str(Position((10, 20), type="mapcoords", anchor="TL")) == "g10/20+jTL"
+    assert str(Position((10, 20), type="mapcoords", offset=(1, 2))) == "g10/20+o1/2"
+    pos = Position("TL", type="inside", anchor="MC", offset=("1c", "2c"))
+    assert str(pos) == "jTL+jMC+o1c/2c"
+
```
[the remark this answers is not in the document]
seisman: Done in d82f4ba53a66d3a83527e9300c307f7e46d8dc97.

seisman:
Ping @GenericMappingTools/pygmt-maintainers for final reviews. I plan to merge it in 48 hours if no further comments.

pygmt/params/position.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+@dataclasses.dataclass(repr=False)
+class Position(BaseParam):
+    """
+    Class for positioning embellishments on a plot.
+
+    .. figure:: https://docs.generic-mapping-tools.org/dev/_images/GMT_anchor.png
+       :width: 600 px
+       :align: center
```
yvonnefroehlich, line 23: I feel we should make the images here and the one in the Technical References at https://pygmt-dev--4212.org.readthedocs.build/en/4212/techref/justification_codes.html consistent regarding the used colors.

pygmt/params/position.py
```diff
@@ -0,0 +1,211 @@
...
+    Think of it like dropping an anchor from a boat:
+
+    1. The boat navigates to the *reference point* (a location on the plot)
+    2. The *anchor point* (a specific point on the embellishment) is aligned with the
+       *reference point*
+    3. The embellishment is "dropped" at that position
+
+    **Reference Point**
+
+    The *reference point* can be specified in five different ways using the ``type`` and
```
yvonnefroehlich: I am unsure about ``type`` as attribute name, as ``type`` or ``type()`` seems to be a reserved keyword or function by Python itself:

position_class_attribute_type

We often use the name of a parameter or attribute for the name of the variable passed to the parameter or attribute:

```python
import pygmt

v = 5

# %% 
type(v)
# works and outputs int

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=True)
fig.logo(position=Position("TL", type="inside", offset="0.2c"), box=True)
fig.show()

# %%
type = "inside"
type(v)
# fails as function was overwritten

fig = pygmt.Figure()
fig.basemap(region=[0, 10, 0, 10], projection="X10c", frame=True)
fig.logo(position=Position("TL", type=type, offset="0.2c"), box=True)
fig.show()
```

pygmt/params/position.py
seisman: Here we use the image from the GMT documentation, mainly because it's technically difficult to generate an image dynamically in docstrings.

pygmt/params/position.py
seisman: Yes, `type` is not an ideal name. Several options are:

- `coord_type`: accurate but a little too long
- `kind`: a general term, similar to `type`
- `cs_type`: cs is short for "coordinate system", not that bad.

pygmt/params/position.py
```diff
@@ -0,0 +1,211 @@
...
+
+        **Example:** ``refpoint=(0.2, 0.1), type="boxcoords"``
+
+    ``type="inside"`` Inside Plot
+        Select one of the nine :doc:`justification codes </techref/justification_codes>`
+        as the *reference point*. The *anchor point* defaults to be the same as the
+        *reference point*, so the embellishment is placed inside the plot.
+
+        **Example:** ``refpoint="TL", type="inside"`` [anchor point defaults to
+        ``"TL"``]
```
yvonnefroehlich: Here, it's not directly clear that the anchor point depends on the reference point (this is explained later in the part for the anchor point). So, I think it's better to remove the default for the anchor point to avoid that people think the default for the anchor point is always `` TL``.

```suggestion
        **Example:** ``refpoint="TL", type="inside"``
```

pygmt/params/position.py
```diff
@@ -0,0 +1,211 @@
...
+
+    ``type="outside"`` Outside Plot
+        Similar to ``type="inside"``, but the *anchor point* defaults to the mirror
+        opposite of the *reference point*. Useful for placing embellishments outside
+        the plot boundaries (e.g., color bars).
+
+        **Example:** ``refpoint="TL", type="outside"`` [anchor point defaults to
+        ``"BR"``]
```
yvonnefroehlich:
```suggestion
        **Example:** ``refpoint="TL", type="outside"``
```

yvonnefroehlich:
```suggestion
        **Example:** ``refpoint="TL", type="outside"``
```

pygmt/params/position.py
yvonnefroehlich: I am fine with `refpoint`. Unfortunately, I have only little experience with positional-only parameters.

pygmt/params/position.py
yvonnefroehlich: I am fine with `coord_type` or `cs_type`. `coord_type` is more descriptive, but I have a slight preference for `cs_type`, as I feel for the arguments `inside` and `outside` "coordinate system" makes more sense. As geopandas uses `crs` for "coordinate reference systems", `cs` for "coordinate system" should be fine.

seisman: Maybe `cstype`, since `cs` is already a shorthand for `coordinate system`?

PR #4218, opened 2025-11-20
pygmt.grdgradient: Migrate the 'azimuth'/'radiance' parameters to the new alias system

seisman:
Just regular migration. No other changes.

### DIFF
--- a/pygmt/src/grdgradient.py
+++ b/pygmt/src/grdgradient.py
@@ -10,7 +10,7 @@ from pygmt._typing import PathLike
 from pygmt.alias import Alias, AliasSystem
 from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import args_in_kwargs, build_arg_list, fmt_docstring, use_alias
+from pygmt.helpers import build_arg_list, fmt_docstring, use_alias
 
 __doctest_skip__ = ["grdgradient"]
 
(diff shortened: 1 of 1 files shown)


pygmt/src/grdgradient.py
```diff
@@ -50,15 +43,17 @@ def grdgradient(
...
     ----------
     {grid}
     {outgrid}
-    azimuth : float, str, or list
-        *azim*\ [/*azim2*].
+    azimuth
+        *azimuth* or (*azimuth*, *azimuth2*).
         Azimuthal direction for a directional derivative; *azim* is the
         angle in the x,y plane measured in degrees positive clockwise from
         north (the positive y-direction) toward east (the positive x-direction). The
```
yvonnefroehlich, line 59: If we change "azim" to "azimuth" we need to update the following docstrings.

```suggestion
        Azimuthal direction for a directional derivative; *azimuth* is the
        angle in the x,y plane measured in degrees positive clockwise from
        north (the positive y-direction) toward east (the positive x-direction). The
```

Unfortuantely I can only suggest changes for the first few lines.

```
        negative of the directional derivative,
        :math:`-(\frac{{dz}}{{dx}}\sin(\mbox{{azimuth}}) + \
        \frac{{dz}}{{dy}}\cos(\mbox{{azimuth}}))`, is found; negation yields
        positive values when the slope of :math:`z(x,y)` is downhill in the
        *azimuth* direction, the correct sense for shading the illumination of an
        image by a light source above the x,y plane shining from the *azimuth*
        direction. Optionally, supply two azimuths, *azimuth*/*azimuth2*, in which
        case the gradients in each of these directions are calculated and the
        one larger in magnitude is retained; this is useful for illuminating
        data with two directions of lineated structures, e.g., *0*/*270*
        illuminates from the north (top) and west (left).  Finally, if *azimuth*
        is a file it must be a grid of the same domain, spacing and
        registration as *grid* that will update the azimuth at each output
        node when computing the directional derivatives.
```

seisman: I've changed it to *azim*. Will work on the docstrings in future PRs instead.

yvonnefroehlich (approved):

PR #4221, opened 2025-11-24
pygmt.info: Migrate the 'spacing' parameter to the new alias system

seisman:
Just normal migrations. Docstrings are not updated.

### DIFF
--- a/pygmt/src/info.py
+++ b/pygmt/src/info.py
@@ -59,7 +59,7 @@ def info(
         {table-classes}.
     per_column : bool
         Report the min/max values per column in separate columns.
-    spacing : str
+    spacing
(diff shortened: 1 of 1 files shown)


pygmt/src/info.py
```diff
@@ -88,7 +86,9 @@ def info(
         - :class:`numpy.ndarray` if either of the above parameters are used.
         - str if none of the above parameters are used.
     """
-    aliasdict = AliasSystem().add_common(
+    aliasdict = AliasSystem(
+        I=Alias(spacing, name="spacing", sep="/"),
```
seisman, line 90: Need to note that `spacing` can be a sequence of two or more values, so we don't check its size here.

yvonnefroehlich (approved):

PR #4228, opened 2025-11-27
Example for geopandas polygon geometry: Use the Natural Earth dataset, change to Africa

yvonnefroehlich:
**Description of proposed changes**

Use the Use the Natural Earth dataset in the example for geopandas polygon geometry. For details see https://github.com/GenericMappingTools/pygmt/issues/4223#issue-3658907831.
Let's hope the request / download is more stable and this will make our docs build less often fail :rocket:.

Have to choose a region (and adjust the positionn of the colorbar):
- Africa (we have this already in the paper)
- South America
- Europe (we have this already for the rivers)
- or any other suggestion people like

| Chicago (currently) | Africa | South America | Europe |
| --- | --- | --- | --- |
| Image | Image | Image | Image |

Related to
* #4229
* #2786
* #2796
* #1374

Fixes #4223

**Preview**: https://pygmt-dev--4228.org.readthedocs.build/en/4228/gallery/maps/choropleth_map.html

**Guidelines**

- [General Guidelines for Pull Request](https://www.pygmt.org/dev/contributing.html#general-guidelines-for-making-a-pull-request-pr)
- [Guidelines for Contributing Documentation](https://www.pygmt.org/dev/contributing.html#contributing-documentation)
- [Guidelines for Contributing Code](https://www.pygmt.org/dev/contributing.html#contributing-code)

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

Issue #2786, opened 2023-10-31
Which data to use for geopandas.GeoDataFrame choropleth example?


michaelgrund:
I wanted to continue on the choropleth example mentioned in [#1374](https://github.com/GenericMappingTools/pygmt/issues/1374#issuecomment-1512650427), however, in the meanwhile 
`world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))` raises a deprection warning saying the datasets will be removed in geopandas v1.0 (see discussion in [#2751](https://github.com/geopandas/geopandas/issues/2751)).

Thus, we need to use another data source. Any ideas what can be used without to much effort and without adding further dependecies to PyGMT?

seisman:
Maybe read a URL directly?
```
world = gpd.read_file("https://naciscdn.org/naturalearth/110m/physical/ne_110m_land.zip")
```

michaelgrund:
> Maybe read a URL directly?
> 
> ```
> world = gpd.read_file("https://naciscdn.org/naturalearth/110m/physical/ne_110m_land.zip")
> ```

Yes that works but unfortunately the fetched data does neither contain population values nor it's possible to filter e.g. for a specific continent/region like "Europe".

yvonnefroehlich:
I faced this issue when starting to work on possible figures for the PyGMT paper.

So far I understood that there is now a separate repro from the geopandas organization for datasets: `geodatasets` https://github.com/geopandas/geodatasets and https://geodatasets.readthedocs.io/en/latest/.

These datasets can be accessed:
```python
import geopandas
import geodatasets

airbnb = geopandas.read_file(geodatasets.get_path("geoda.airbnb"))
```

It is possible to get more information about these datasets (please see https://geodatasets.readthedocs.io/en/latest/#how-to-use):
```python
import geodatasets
geodatasets.data
```
Gives:
```
{'geoda': {'airbnb': {'url': 'https://geodacenter.github.io/data-and-lab//data/airbnb.zip',
   'license': 'NA',
   'attribution': 'Center for Spatial Data Science, University of Chicago',
   'name': 'geoda.airbnb',
   'description': 'Airbnb rentals, socioeconomics, and crime in Chicago',
   'geometry_type': 'Polygon',
   'nrows': 77,
   'ncols': 21,
   'details': 'https://geodacenter.github.io/data-and-lab//airbnb/',
[...]

Issue #1374, opened 2021-07-06
Add gallery example for plotting geopandas.GeoDataFrame objects


weiji14:
**Description of the desired feature**

It would be great to have some **gallery examples** on how users can plot geopandas.GeoDataFrame objects in PyGMT. This will be a follow up to #1000. You will need to have some familiarity with `geopandas` and/or `PyGMT`, but if you're still keen to learn, let us know and we'll guide you along!

Note that if you are interested in writing a **full tutorial** for geopandas inputs, see #1268.

## Examples of what you can do

These are just suggestions of things that you can plot. Feel free to come up with other ideas, especially if you know of a cool dataset! See also https://geopandas.org/gallery/index.html for inspiration

1. Points: Capital cities in a region, Weather stations in a country, etc
2. Lines: Road centrelines, etc
3. Polygons: Geological map of a place, vegetation type/land cover map, etc

## How to make the change :monocle_face:

1. First, read up our contributing guidelines at https://www.pygmt.org/v0.4.0/contributing.html#editing-the-documentation
2. Leave a comment below, stating that you will tackle this issue. We will assign you to this issue and you can then start working on it.
3. Go to e.g. https://github.com/GenericMappingTools/pygmt/tree/master/examples/gallery/maps, and click on 'Add File'. Or you may want to do this in your code editor locally.
4. Submit a pull request, and be sure to fill it up with as much detail as possible.

**Are you willing to help implement and maintain this feature?** Happy to mentor someone to handle this task!

meghanrjones:
Assigning this issue to @Location-Artistry who's requested to work on this issue for the SciPy2021 sprint 😄

michaelgrund:
I played a little bit around with a dataset containing the road net of the US in shapefile format (TIGER/Line shapefiles).  

However, since the dataset is huge for the whole US I selected a subset containing only the [road net of Hawai'i](https://catalog.data.gov/dataset/tiger-line-shapefile-2015-state-hawaii-primary-and-secondary-roads-state-based-shapefile) which is only around 2.5 MB (I made a zoom on O'ahu). 

I think this would be enough to generate a nice example which shows how we could use shapefile data loaded via geopandas in PyGMT. My only question is: how and where do we want to store data for generating the geopandas examples?

![TEST_map](https://user-images.githubusercontent.com/23025878/131250854-3c0311cf-9698-4f3c-a571-7e13d94e95b5.png)

weiji14:
[...]

PR #4229, opened 2025-11-27
Example for geopandas line geometry: Use Natural Earth dataset, change to Asia

### LINKED ISSUE #4223: Use the Natural Earth datasets rather than geodataset datasets in geopandas gallery examples
Currently, our [LineStrings](https://www.pygmt.org/dev/gallery/lines/linestrings.html) example uses the "eea large_rivers" sample data provided by `geodatasets` package. The EEA dataset is relatively large (339 KB) and its download link (https://www.eea.europa.eu/data-and-maps/data/wise-large-rivers-and-large-lakes/zipped-shapefile-with-wise-large-rivers-vector-line/zipped-shapefile-with-wise-large-rivers-vector-line/at_download/file) occasionally times out during documentation builds, causing unnecessary failures.

I find that the Natural Earth datasets (https://www.naturalearthdata.com) are better and usually have smaller file sizes. And the mirror URL https://naciscdn.org/naturalearth/ is even more stable.

We can improve both the [LineString](https://www.pygmt.org/dev/gallery/lines/linestrings.html) and [choropleth map](https://www.pygmt.org/dev/gallery/maps/choropleth_map.py) examples using the Natural Earth data.

yvonnefroehlich:

Use the Use the Natural Earth dataset in the example for geopandas line geometry. For details see https://github.com/GenericMappingTools/pygmt/issues/4223#issue-3658907831.
Let's hope the request / download is more stable and this will make our docs build less often fail :rocket:.

For the rivers there is only one river plotted for Europe. When changing to 50 m (instead of 110 m) I get the error
```
File C:\ProgramData\Anaconda3\envs\pygmt_env_dev_650\Lib\site-packages\pygmt\helpers\tempfile.py:148 in tempfile_from_geojson
    for col in geojson.columns:

  File C:\ProgramData\Anaconda3\envs\pygmt_env_dev_650\Lib\site-packages\pandas\core\generic.py:6321 in __getattr__
    return object.__getattribute__(self, name)

AttributeError: 'GeoSeries' object has no attribute 'columns'

RuntimeError: GDAL Error: Features without geometry not supported by GMT writer.. Failed to write record: <fiona.model.Feature object at 0x00000206E466CFB0>. 
```
When using the subset `rivers[rivers["scalerank"] != 5]` it works. Interestingly the subset `rivers_five = rivers[rivers["scalerank"] == 5]` is empty. For 10 m, I do not face such an error.

| Europe (currently) | Europe 110 m | Europe 50 m | Europe 10 m |
| --- | --- | --- | --- |
| Image | image | image | image |

Related to 
* #4228
* #3711
* #2786
* #1374

Fixes #4223

**Preview**: https://pygmt-dev--4229.org.readthedocs.build/en/4229/gallery/lines/linestrings.html

**Guidelines**

**Slash Commands**

### DIFF
--- a/examples/gallery/lines/linestrings.py
+++ b/examples/gallery/lines/linestrings.py
@@ -29,29 +29,3 @@ fig.coast(land="gray95", shorelines="1/0.3p,gray50")
 fig.plot(data=rivers_asia, pen="1p,steelblue")
 
 fig.show()
-
-
-# %%
-rivers_australia = rivers.cx[111:155, -40:-9].copy()
-
-fig = pygmt.Figure()
-fig.basemap(region=[111, 155, -40, -9], projection="M10c", frame=True)
-fig.coast(land="gray95", shorelines="1/0.3p,gray50")
-
-# Add rivers to map
-fig.plot(data=rivers_australia, pen="1p,steelblue")
-
-fig.show()
-
-
-# %%
-rivers_sa = rivers.cx[-84.5:-33, -56.5:13].copy()
-
-fig = pygmt.Figure()
-fig.basemap(region=[-84.5, -33, -56.5, 13], projection="M10c", frame=True)
-fig.coast(land="gray95", shorelines="1/0.3p,gray50")
-
-# Add rivers to map
-fig.plot(data=rivers_sa, pen="1p,steelblue")
-
-fig.show()


michaelgrund:
Third option, just show line objects for a geopandas line geometry example, nothing else. So, no cities should appear 😉 .

yvonnefroehlich:
> Third option, just show line objects for a geopandas line geometry example, nothing else. So, no cities should appear 😉 .

Yep, here for the line example we will stay with the rivers only.
The image with the cities was from the related issue, as I was wondering if we maybe want to add a new third example for points.

seisman:
> When changing to 50 m (instead of 110 m) I get the error
> 
> ```
> File C:\ProgramData\Anaconda3\envs\pygmt_env_dev_650\Lib\site-packages\pygmt\helpers\tempfile.py:148 in tempfile_from_geojson
>     for col in geojson.columns:
> 
>   File C:\ProgramData\Anaconda3\envs\pygmt_env_dev_650\Lib\site-packages\pandas\core\generic.py:6321 in __getattr__
>     return object.__getattribute__(self, name)
> 
> AttributeError: 'GeoSeries' object has no attribute 'columns'
> 
> RuntimeError: GDAL Error: Features without geometry not supported by GMT writer.. Failed to write record: <fiona.model.Feature object at 0x00000206E466CFB0>. 
> ```

It's likely there are one or a few rivers that doesn't have geometry information.

This can be avoided if we focus on the rivers in Europe only. This can be done by:
```
rivers = rivers.cx[-10:30, 35:57]
```

examples/gallery/lines/linestrings.py
```diff
@@ -14,33 +14,19 @@
...
-    shorelines="1/0.1p,gray50",
-    borders="1/0.1,gray30",
-    frame=True,
-)
+fig.basemap(region=[-10, 30, 35, 57], projection="M10c", frame=True)
+fig.coast(land="gray95", shorelines="1/0.3p,gray50", borders="1/0.1p,black")
 
 # Add rivers to map
-fig.plot(data=gdf, pen="1p,steelblue")
+fig.plot(data=rivers["geometry"], pen="1p,steelblue")
```
seisman:
```suggestion
fig.plot(data=rivers, pen="1p,steelblue")
```

yvonnefroehlich:
I am wondering if it would be welcome to change the region from Europe to South America for the line geometry example. Then we can use Europe for an example on the point geometry. The plotting of the point geometry is straightforward, but I think it's good to show that users can pass the columns of a gdf directly to the `Figure.text` method to label features, e.g., cities, which is much easier than having to set up a GMT-specific file. I prefer to have Europe for the points example because there are no overlapping labels.

| rivers, lines, South America, 50 m (this PR #4229) | cities, points, Europe, 50 m (new PR #4231) |
|  --- | --- |
| image | image |

examples/gallery/lines/linestrings.py
```diff
@@ -4,43 +4,40 @@
...
 
-# Read a sample dataset provided by the geodatasets package.
-# The dataset contains large rivers in Europe, stored as LineString/MultiLineString
-# geometry types.
-gdf = gpd.read_file(geodatasets.get_path("eea large_rivers"))
+# Read a sample dataset provided by Natural Earth. The dataset contains rivers stored
+# as LineString/MultiLineString geometry types. Here will focus on South America.
+provider = "https://naciscdn.org/naturalearth"
+rivers = gpd.read_file(f"{provider}/50m/physical/ne_50m_rivers_lake_centerlines.zip")
+rivers_sa = rivers.cx[-84.5:-33, -56.5:13]
```
seisman:
```suggestion
rivers_sa = rivers.cx[-84.5:-33, -56.5:13].copy()
```

yvonnefroehlich: Yep, also saw the warning. Fixed in commit https://github.com/GenericMappingTools/pygmt/pull/4229/commits/39918b64943c60aa7ba4e06dc93441f0f0159b05.

Esteban82:
A suggestion regarding figures in tutorials: I think that horizontally-oriented (landscape) figures are often more effective than vertical (portrait) ones. Because they don't fill the entire screen, landscape figures enable users to see the visual aid and the relevant text side-by-side, which creates a smoother and more efficient reading flow.

This approach is implemented effectively in these PyGMT tutorials:

- https://www.pygmt.org/latest/tutorials/basics/plot.html#sphx-glr-tutorials-basics-plot-py
- https://www.pygmt.org/latest/tutorials/basics/coastlines.html#sphx-glr-tutorials-basics-coastlines-py

yvonnefroehlich:
> A suggestion regarding figures in tutorials: I think that horizontally-oriented (landscape) figures are often more effective than vertical (portrait) ones. Because they don't fill the entire screen, landscape figures enable users to see the visual aid and the relevant text side-by-side, which creates a smoother and more efficient reading flow.

Hey @Esteban82. Thanks for looking at this example and your feedback. Yeah, your are right and I agree the South American map gets quite large in the gallery, which is not optimal. To get a more horizontally-orientated figure, we could use Australia or Asia as region. For Australia, there are only a few rivers for the 50 m dataset. Thus I would go with Asia.

| South America | Australia | Asia |
| --- |  --- | --- |
| image | image | image image |

Esteban82:
Hi Yvonne! Yes, I agree with Asia.

seisman:
Sorry, but I prefer not to use the Asia map. China and India have [unresolved boundary disputes](https://en.wikipedia.org/wiki/Sino-Indian_border_dispute), and from China’s perspective, the boundary lines in this dataset—and even in GMT’s GSHHG dataset—are **incorrect**. Since I am from China and leading this project, I prefer not to include any national boundary lines of China to avoid potential political sensitivities.

yvonnefroehlich:
> Sorry, but I prefer not to use the Asia map. China and India have [unresolved boundary disputes](https://en.wikipedia.org/wiki/Sino-Indian_border_dispute), and from China’s perspective, the boundary lines in this dataset—and even in GMT’s GSHHG dataset—are **incorrect**. Since I am from China and leading this project, I prefer not to include any national boundary lines of China to avoid potential political sensitivities.

No worries! This is totally fair and an important aspect.
Would not plotting the national boundaries help? Otherwise we need another region; maybe Australia even there are only a few rivers for the 50 m dataset.

seisman:
> Would not plotting the national boundaries help?

Yes.

Issue #4208, opened 2025-11-13
Figure.grdview: Create 3-D perspective image or surface mesh from a grid


seisman:
*This issue serves as the central place for discussing and tracking the implementation of the `grdview` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/grdview.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/grdview
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.Figure.grdview.html

## GMT Option Flags and Modifieres

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [x] `ingrid`: `grid`
- [x] `-R`: `region` 
- [x] `-J`: `projection`
- [x] `-Jz`: `zscale`/`zsize`
- [x] `-B`: `frame`
- [x] `-C`: `cmap` #4443
- [x] `-G`: `drape_grid` #4442
- [ ] `-I`
- [x] `-N`: `plane`/`facade_fill` #4235 @seisman
- [x] `-Q`: `surftype`/`dpi`/`meshfill`/`nan_transparent`/`monochrome` #4234 @seisman
- [x] `-S`: `smooth` #4444
- [ ] `-T` #3355
- [ ] ~~`-U`~~: Use `Figure.timestamp` instead.
- [x] `-V`: verbose
- [x] `-W`:
  - [x] `-Wc`: `contour_pen`
  - [x] `-Wf`: `facade_pen` #4235 @seisman 
  - [x] `-Wm`: `mesh_pen` #4234
- [ ] ~~`-X`/`-Y`~~: Use `Figure.shift_origin` instead.
- [ ] `-f`
- [ ] `-b`
- [x] `-p`
- [x] `-t`
- [ ] ~~`--PAR=value`~~: Use `pygmt.config` instead.

## Linked Pull Requests

- [x] Initial feature implementation #330 @weiji14 
- [ ] Add a tutorial or gallery example

## Related Issues and Discussions

*Add links to related wrapper discussions, API design threads, or upstream GMT changes here.*

seisman:
[...]

PR #4234, opened 2025-12-02
Figure.grdview: Add parameters surftype/dpi/mesh_fill/nan_transparent/monochrome to control surface types

seisman:
This PR improves `grdview`'s `-Q` option in a more Pythonic way by splitting it into multiple parameters:

- `surftype`
- `dpi`
- `nan_transparent`
- `monochrome`
- `mesh_fill`

Address https://github.com/GenericMappingTools/pygmt/issues/4208#issuecomment-3528640135.

**Preview** at https://pygmt-dev--4234.org.readthedocs.build/en/4234/api/generated/pygmt.Figure.grdview.html

**Example**
```python
import pygmt

grd_relief = pygmt.datasets.load_earth_relief(resolution="15m", region=[-25, -13, 63.2, 66.7])

args = dict(
    region=[-25, -13, 63.2, 66.7], 
    projection="M?", 
    perspective=(-150, 25),
    cmap="SCM/oleron",
    zsize="1.5c",
    plane="+ggray",
    panel=True,
)

fig = pygmt.Figure()
with fig.subplot(nrows=5, ncols=3, subsize=(5, 5), margins=(1, 0.2)):
    for surftype in ("surface", "mesh", "surface+mesh", "image", "waterfall_x", "waterfall_y"):
        fig.grdview(
            grid=grd_relief,
            frame=["xaf", "yaf", f"WSenZ+t{surftype}"],
            surftype=surftype,
            **args,
        )
    for surftype in ("surface", "mesh", "surface+mesh", "image", "waterfall_x", "waterfall_y"):
        fig.grdview(
            grid=grd_relief,
            frame=["xaf", "yaf", f"WSenZ+t{surftype}"],
            surftype=surftype,
            monochrome=True,
            **args,
        )
    for surtype in ("mesh", "waterfall_x", "waterfall_y"):
        fig.grdview(
            grid=grd_relief,
            frame=["xaf", "yaf", f"WSenZ+t{surftype}"],
            surftype=surftype,
            meshfill="red@50",
            **args,
        )
fig.show()
```

map

### DIFF
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -163,9 +163,11 @@ To increase the chances of getting your pull request accepted quickly, try to:
   - Write tests for the code you wrote/modified if needed.
     Please refer to [Testing your code](contributing.md#testing-your-code) or
     [Testing plots](contributing.md#testing-plots).
-  - Include an example of new features in the gallery or tutorials.
-    Please refer to [Gallery plots](contributing.md#contributing-gallery-plots)
-    or [Tutorials](contributing.md#contributing-tutorials).
+  - Include an example of new features in the gallery or tutorials. Please refer to
+    [Gallery plots](contributing.md#contributing-gallery-plots) or
+    [Tutorials](contributing.md#contributing-tutorials). If adding a new
+    method/function/class, the gallery example or tutorial should be submitted in a
+    separate pull request.
 * Have a good coding style
   - Use readable code, as it is better than clever code (even with comments).
   - Follow the [PEP8](https://pep8.org) style guide for code and the
@@ -466,14 +468,15 @@ function/class/module/method.
 
 ### PyGMT Code Overview
 
-The source code for PyGMT is located in the `pygmt/` directory. When contributing
-code, be sure to follow the general guidelines in the
-[pull request workflow](contributing.md#pull-request-workflow) section.
+The source code for PyGMT is located in the `pygmt/` directory. When contributing code,
+please open an issue first to discuss the feature and its implementation and be sure to
+follow the general guidelines in the [pull request workflow](#pull-request-workflow)
+section.
 
 ### Code Style
 
 We use the [ruff](https://docs.astral.sh/ruff) tool to format the code, so we
-don't have to think about it. It loosely follow the [PEP8](https://pep8.org) guide
+don't have to think about it. It loosely follows the [PEP8](https://pep8.org) guide
 but with a few differences. Regardless, you won't have to worry about formatting
 the code yourself. Before committing, run it to automatically format your code:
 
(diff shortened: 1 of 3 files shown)


pygmt/src/grdview.py
```diff
@@ -159,9 +174,45 @@ def grdview(  # noqa: PLR0913
...
         JZ=Alias(zsize, name="zsize"),
+        Q=[
+            Alias(
+                surftype,
+                name="surftype",
+                mapping=_surtype_mapping if not _old_surftype_syntax else None,
+            ),
+            Alias(dpi, name="dpi"),
+            Alias(meshfill, name="meshfill"),
+            Alias(monochrome, name="monochrome", prefix="+m"),
```
weiji14: Shouldn't `+m` be a suffix?

pygmt/src/grdview.py
```diff
@@ -159,9 +174,45 @@ def grdview(  # noqa: PLR0913
...
+        "surface": "s",
+        "mesh": "m",
+        "surface+mesh": "sm",
+        "image": "c" if nan_transparent is True else "i",
+        "waterfallx": "mx",
+        "waterfally": "my",
+    }
+
+    # surftype was aliased to Q previously.
+    _old_surftype_syntax = surftype not in _surtype_mapping and surftype is not None
```
weiji14: Can you add a unit test to ensure that backwards compatibility is preserved when a user passes `surftype="c"`, but did not set `nan_transparent=True` (which is added in this PR). The output grdview plot should still have NaN transparency.

pygmt/src/grdview.py
seisman: The full syntax is `-Qc|i|m[x|y]|s[m][color][+m]`. It's a suffix for the `-Q` option, but for `monochrome`, it's a prefix. Since `monochrome` is a boolean, `monochrome` would be `+m`, so here it makes no difference if `+m` is a prefix or a suffix.

weiji14: Ah right, got it now.

pygmt/src/grdview.py
seisman: Added in aa9eb737c310dbf16be042878828ac62cfcb5618.

pygmt/src/grdview.py
weiji14: Just a note that after the `suffix` feature is implemented in #4259, this `+m` does need to be a `prefix`, otherwise a list of values will be returned, xref https://github.com/GenericMappingTools/pygmt/pull/4259#discussion_r2600879217.

pygmt/tests/test_grdview.py
```diff
@@ -98,44 +226,6 @@ def test_grdview_with_perspective_and_zsize(xrgrid):
...
 
 
-@pytest.mark.mpl_image_compare
-def test_grdview_with_cmap_for_image_plot(xrgrid):
-    """
-    Run grdview by passing in a grid and setting a colormap for producing an image plot.
-    """
-    fig = Figure()
-    fig.grdview(grid=xrgrid, cmap="SCM/oleron", surftype="i")
-    return fig
```
seisman, line 108: Superseded by `test_grdview_surftype`

pygmt/tests/test_grdview.py
```diff
@@ -98,44 +226,6 @@ def test_grdview_with_perspective_and_zsize(xrgrid):
...
-
-
-@pytest.mark.mpl_image_compare
-def test_grdview_with_cmap_for_surface_monochrome_plot(xrgrid):
```
seisman, line 112: Superseded by `test_grdview_monochrome`

pygmt/tests/test_grdview.py
```diff
@@ -98,44 +226,6 @@ def test_grdview_with_perspective_and_zsize(xrgrid):
...
-    """
-    Run grdview by passing in a grid and setting a colormap for producing a surface
-    monochrome plot.
-    """
-    fig = Figure()
-    fig.grdview(grid=xrgrid, cmap="SCM/oleron", surftype="s+m")
-    return fig
-
-
-@pytest.mark.mpl_image_compare
```
seisman, line 122: Superseded by `test_grdview_surftype`

pygmt/src/grdview.py
```diff
@@ -97,24 +143,30 @@ def grdview(  # noqa: PLR0913
...
-
-        For any of these choices, you may force a monochrome image by appending the
-        modifier **+m**.
+    surftype
+        Specify surface type for the grid. Valid values are:
+
+        - ``"mesh"``: mesh plot [Default].
+        - ``"surface"``: surface plot.
+        - ``"surface+mesh"``: surface plot with mesh lines drawn on top of the surface.
+        - ``"image"``: image plot.
```
weiji14, line 185: I've used `surftype="sim"` before [here](https://github.com/weiji14/deepbedmap/blob/4d437dd04f1e12bde0b53f59c1505914a1d09bb6/deepbedmap.py#L290). It's been a few years so I can't remember, but maybe check if `-Qsim` is equal to `-Qsm`, or if we need to add a `surface+image+mesh` option :joy:

seisman: `surftype="sim"` is identical to `surftype="s"`:
```
import pygmt

grd_relief = pygmt.datasets.load_earth_relief(
    resolution="15m", region=[-25, -13, 63.2, 66.7]
)

args = dict(
    region=[-25, -13, 63.2, 66.7],
    projection="M10c",
    perspective=(-150, 25),
    cmap="SCM/oleron",
    frame=True,
    zsize="1.5c",
    plane="+ggray",
)

fig = pygmt.Figure()
fig.grdview(grid=grd_relief, surftype="sim", **args)
fig.shift_origin(xshift="w+2c")
fig.grdview(grid=grd_relief, surftype="surface", **args)
fig.show()
```
map

`-Qsim` is taken as `-Qs` based on the GMT source code https://github.com/GenericMappingTools/gmt/blob/a091f7542cfded0ab0cc7f7fe1f899790b5eed56/src/grdview.c#L773-L777.

weiji14: Nice, glad to know we don't need to add more combinations!

pygmt/src/grdview.py
```diff
@@ -97,24 +143,30 @@ def grdview(  # noqa: PLR0913
...
-          top of the surface.
-        - **i**: image plot.
-        - **c**: Same as **i** but will make nodes with z = NaN transparent.
-
-        For any of these choices, you may force a monochrome image by appending the
-        modifier **+m**.
+    surftype
+        Specify surface type for the grid. Valid values are:
+
+        - ``"mesh"``: mesh plot [Default].
```
weiji14, line 182: Default is technically `None` if no argument is passed to `surftype`. Should we remove the `[Default]` then, or allow for `surftype=True` to mean `surftype="m"`?

seisman: I feel `surftype=True` is not intuitive, mainly because:

- `surftype=False` is meaningless in this context (`False` usually means "disable")
- `surftype=True` is mapped to `-Q` but `-Q` without any arguments is invalid in GMT.

weiji14: Agree, so let's just remove the `[Default]` to only allow string literals

```suggestion
        - ``"mesh"``: mesh plot.
```

and probably need to update the docs to remove `[Default]` too at https://docs.generic-mapping-tools.org/6.6/grdview.html#q if `-Q` by itself is invalid. Unless I'm missing some context?

seisman: "Default" actually means `-Q` is not required, and if not given, `-Qm` is used.

weiji14: Ah ok, so `surftype=None` essentially maps to `-Qm` (even if `-Q` is not passed to GMT grdview). Then ok to keep the [Default] (though it is slightly confusing).

Issue #2014, opened 2022-07-19
Usage of underscores in aliases


yvonnefroehlich:
In the Contributors Guide in the section [Code Style](https://www.pygmt.org/dev/contributing.html#code-style):

> When wrapping a new alias, use an underscore to separate words bridged by vowels (aeiou), such as no_skip and z_only. Do not use an underscore to separate words bridged only by consonants, such as distcalc, and crossprofile. This convention is not applied by the code checking tools, but the PyGMT maintainers will comment on any pull requests as needed.

However, some aliases seem not to follow this convention. For example:
- [`pygmt.Figure.plot()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.plot.html):  `straight_line` (`A`) , `error_bar` (`E`), `nodata` (`d`)
 Following the convention, they should be `straigthline`, `errorbar` , `no_data`
- [`pygmt.Figure.rose()`](https://www.pygmt.org/dev/api/generated/pygmt.Figure.rose.html): `vector_params` (`M`)
 Should be `vectorparams`.

_Please note_: This is primarily meant as a general question for understanding, rather than a request for renaming these aliases. :wink:

maxrjones:
I think that the aliases that do not follow this convention were added to PyGMT before the convention was established. For context, the idea of a convention to ensure consistency for alias names was first suggested in https://github.com/GenericMappingTools/pygmt/pull/1190#issuecomment-814451969 and added in https://github.com/GenericMappingTools/pygmt/pull/1256.

yvonnefroehlich:
Thanks @maxrjones for the explanation and the background information!

seisman:
I don't think we will rename these aliases. So closing the issue.

seisman:
Here is a Python script written by ChatGPT to print all parameters of PyGMT functions and Figure's plotting methods:
```python
import inspect
import pygmt

def list_pygmt_parameters():
    results = {}

    # A helper to extract parameters from a callable
    def extract(name, obj):
        try:
            sig = inspect.signature(obj)
            params = list(sig.parameters.keys())
            results[name] = params
        except (TypeError, ValueError):
            pass

    # 1. Inspect top-level PyGMT module
    for name, obj in inspect.getmembers(pygmt):
        if inspect.isfunction(obj):
            extract(f"pygmt.{name}", obj)

    # 2. Inspect Figure class methods
    for name, obj in inspect.getmembers(pygmt.Figure):
[...]

PR #1256, opened 2021-05-03
Add alias name convention to CONTRIBUTING.md

meghanrjones:
**Description of proposed changes**

This PR adds a convention for when to separate words in aliases by an underscore to CONTRIBUTING.md. Originally suggested by @weiji14 in https://github.com/GenericMappingTools/pygmt/pull/1190#issuecomment-814451969.

Fixes #

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #1190, opened 2021-04-06
Complete documentation for grdtrack

weiji14:
**Description of proposed changes**

Finish up aliasing parameters for `grdtrack`, improving the documentattion to be better than what we currently have at https://www.pygmt.org/v0.3.1/api/generated/pygmt.grdtrack.html! For reference, upstream GMT docs is at https://docs.generic-mapping-tools.org/6.1/grdtrack.html and https://github.com/GenericMappingTools/gmt/blame/master/doc/rst/source/grdtrack.rst.

**Preview** at https://pygmt-git-complete-docs-grdtrack-gmt.vercel.app/api/generated/pygmt.grdtrack.html

TODO:

Arguments/Aliases to document (see also https://www.generic-mapping-tools.org/GMT.jl/dev/#GMT.grdtrack). New aliases are highlighted in **bold**:

- [x] A - **resample**
- [x] C - **crossprofile**
- [x] D - **dfile**?
- [x] [E](https://docs.generic-mapping-tools.org/6.1/grdtrack.html#e) - **profile**
- [x] F - **critical** ?
- [x] G - outfile (but not really an alias...)
- [x] N - **no_skip**
- [x] R - region
- [x] [S](https://docs.generic-mapping-tools.org/6.1/grdtrack.html#s) - **stack**
- [x] T - **radius**
- [x] V - verbose
- [x] Z - **z_only**
- [x] f - coltypes
- [x] i - **incols**
- [x] j - **distcalc**
- [x] n - interpolation

Finalizes #307. **Note**: Merge this PR after #1189.

**Reminders**

- [x] Run `make format` and `make check` to make sure the code follows the style guide.
- [x] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [x] Write detailed docstrings for all functions/methods.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash commands are:

- `/format`: automatically format and lint the code
- `/test-gmt-dev`: run full tests on the latest GMT development version

PR #4235, opened 2025-12-02
Figure.grdview: Improve parameters plane/facade_fill/facade_pen to set the plane and facade

seisman:
This PR splits the `-N` option into two parameters, `plane` for setting the z-level and `facade_fill` for setting the fill for facade. 

**Preview**: https://pygmt-dev--4235.org.readthedocs.build/en/4235/api/generated/pygmt.Figure.grdview.html#pygmt.Figure.grdview

Related to #4208

### DIFF
--- a/pygmt/src/grdview.py
+++ b/pygmt/src/grdview.py
@@ -169,7 +169,7 @@ def grdview(  # noqa: PLR0913
     """
     self._activate_figure()
 
-    # Enable 'plane' if 'facade_fill' or 'facade_pen' is set
+    # Enable 'plane' if 'facade_fill' or 'facade_pen' are set
     if plane is False and (facade_fill is not None or facade_pen is not None):
         plane = True
 


pygmt/src/grdview.py
```diff
@@ -18,18 +18,19 @@
...
     Wm="meshpen",
-    Wf="facadepen",
     I="shading",
     f="coltypes",
     n="interpolation",
 )
 def grdview(  # noqa: PLR0913
     self,
     grid: PathLike | xr.DataArray,
+    plane: float | bool = False,
```
seisman, line 34: Following our naming convention, the variables `facadepen` and `facadefill` should be `facade_pen`/`facade_fill` instead, but we also have `contourpen` and `meshpen`, so maybe we should keep `facadepen`/`facadefill` here.

seisman:
Changes in this PR will affect our [Figure 3](https://github.com/GenericMappingTools/pygmt-paper-figures), so ping @yvonnefroehlich for a review.

pygmt/src/grdview.py
yvonnefroehlich: Hm, this naming convention... Is this a widely used and strictly applied rule? In principle, using an underscore between two words always helps to increase the readability. 

We already have `compressionfill` and `extensionfill` from the `Figure.meca` method. And I am unsure if it is directly clear to users why these parameter names are without an underscore but `facade_fill` has an underscore. On the other side, having a convention that we do not really follow is a bit meaningless. Maybe we should revise this convention?

Looking at the docs from matplotlib: in https://matplotlib.org/stable/api/contour_api.html#matplotlib.contour.ContourLabeler.clabel they use underscores in `inline_spacing`, `use_clabeltext`, and `rightside_up`, but in https://matplotlib.org/devdocs/api/colorbar_api.html#matplotlib.colorbar.Colorbar it's `drawedeges`.

seisman: Related discussions at https://github.com/GenericMappingTools/pygmt/issues/2014, https://github.com/GenericMappingTools/pygmt/pull/1190#issuecomment-814451969, and https://github.com/GenericMappingTools/pygmt/pull/1256

seisman: I've reopened https://github.com/GenericMappingTools/pygmt/issues/2014 for more discussions.

pygmt/src/grdview.py
```diff
@@ -110,12 +107,19 @@ def grdview(  # noqa: PLR0913
...
-        Set the pen attributes used for the facade. You must also select ``plane`` for
-        the facade outline to be drawn.
     mesh_pen
         Set the pen attributes used for the mesh. You must also select ``surftype`` of
         **m** or **sm** for meshlines to be drawn.
+    plane
+        Draw a plane at the specified z-level. If ``True``, default to the minimum value
+        in the grid. However, if ``region`` was used to set zmin/zmax then that value is
+        used if it is less than the grid minimum value. Use ``facade_pen`` and
+        ``facade_fill`` to control the appearance of the plane.
```
yvonnefroehlich, line 117:
```suggestion
    plane
        Draw a plane at the specified z-level. If ``True``, defaults to the minimum value
        in the grid. However, if ``region`` was used to set *zmin/zmax* then *zmin* is
        used if it is less than the grid minimum value. Use ``facade_pen`` and
        ``facade_fill`` to control the appearance of the plane.
```

pygmt/src/grdview.py
```diff
@@ -110,12 +107,19 @@ def grdview(  # noqa: PLR0913
...
+    facade_fill
+        Fill for the frontal facade between the plane specified by ``plane`` and the
+        data perimeter.
+    facade_pen
+        Set the pen attributes used for the facade.
     shading : str
```
yvonnefroehlich:
```suggestion
    shading : str or float
```

examples/tutorials/advanced/3d_perspective_image.py
```diff
@@ -67,8 +67,8 @@
     zsize="1.5c",
     surftype="s",
     cmap="geo",
-    # Set the plane elevation to 1,000 meters and make the fill "gray"
-    plane="1000+ggray",
+    plane=1000,  # Set the plane elevation to 1,000 meters
+    facade_fill="gray",  # Fill the facade with "gray"
```
yvonnefroehlich:
```suggestion
    facade_fill="gray",  # Color the facade in "gray"
```

pygmt/src/grdview.py
```diff
@@ -165,9 +169,17 @@ def grdview(  # noqa: PLR0913
     """
     self._activate_figure()
 
+    # Enable 'plane' if 'facade_fill' or 'facade_pen' is set
```
michaelgrund:
```suggestion
    # Enable 'plane' if 'facade_fill' or 'facade_pen' are set
```

michaelgrund (approved):

Issue #4243, opened 2025-12-05
Figure.colorbar: Plot gray scale or color scale bar


seisman:
*This issue serves as the central place for discussing and tracking the implementation of the `colorbar` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/colorbar.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/colorbar
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.Figure.colorbar.html

## GMT Option Flags and Modifiers

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [ ] `-R`: `region`
- [ ] `-J`: `projection`
- [ ] `-B`: `frame`
- [ ] `-C`: `cmap`
- [ ] `-D`
- [ ] `-F`
- [ ] `-G`
- [ ] `-I`
- [ ] `-L`
- [ ] `-M`
- [ ] `-N`
- [ ] `-Q`
- [ ] `-S`
- [ ] ~~`-U`~~: Use `Figure.timestamp` instead.
- [ ] `-V`: verbose
- [ ] `-W`
- [ ] - [ ] ~~`-X`/`-Y`~~: Use `Figure.shift_origin` instead.
- [ ] `-Z`
- [ ] `-c`: `panel`
- [ ] `-p`: `perspective`
- [ ] `-t`: `transparency`
- [ ] ~~`--PAR=value`~~: Use `pygmt.config` instead.

## Linked Pull Requests

- [ ] Initial feature implementation
- [ ] Add a tutorial or gallery example

## Related Issues and Discussions

*Add links to related wrapper discussions, API design threads, or upstream GMT changes here.*

seisman:
For a colorbar, the annotations and labels can be set by either the [`-B`](https://docs.generic-mapping-tools.org/dev/gmt.html#b-full) option (can't be used with `-L`) or the `-S` option (a simplified version of the `-B` option). 

The `-B` option is one of the most complicated GMT options, but most `-B` settings do not affect the colorbar. For example, `-BWSEN+tTITLE` doesn't work. As far as I can see, the `-B` syntax that has an effect for a colorbar is

[...]

PR #4244, opened 2025-12-05
Figure.colorbar: Migrate the log/scale parameters to the new alias system

seisman:
Just regular migration.

Related to #4243.

### DIFF
--- a/pygmt/src/colorbar.py
+++ b/pygmt/src/colorbar.py
@@ -130,7 +130,7 @@ def colorbar(  # noqa: PLR0913
     log
         Select logarithmic scale and power of ten annotations. All z-values in the CPT
         will be converted to :math:`p = \log_{10}(z)` and only integer p values will be
-        annotated using the 10^p format [Default is linear scale].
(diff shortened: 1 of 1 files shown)


pygmt/src/colorbar.py
```diff
@@ -123,10 +127,10 @@ def colorbar(  # noqa: PLR0913
...
         ``shading`` is used each rectangle will have its constant color
         modified by the specified intensity.
-    log : bool
-        Select logarithmic scale and power of ten annotations. All z-values
-        in the CPT will be converted to p = log10(z) and only integer p-values
-        will be annotated using the 10^p format [Default is linear scale].
+    log
+        Select logarithmic scale and power of ten annotations. All z-values in the CPT
+        will be converted to :math:`p = \log_{10}(z)` and only integer p values will be
+        annotated using the 10^p format [Default is linear scale].
```
yvonnefroehlich:
```suggestion
        annotated using the :math:`10^{p}` format [Default is linear scale].
```

PR #4255, opened 2025-12-07
Fix typos TM->MC in two legend tests

seisman:
`TM` is not a valid anchor code. GMT defaults to "MC" when an invalid anchor code is given.
This PR fixes `TM` to `MC` so that we don't need to update baseline images.

xref: https://www.pygmt.org/dev/techref/justification_codes.html

### DIFF
--- a/pygmt/tests/test_legend.py
+++ b/pygmt/tests/test_legend.py
@@ -112,7 +112,7 @@ def test_legend_stringio(legend_spec):
     spec = io.StringIO(legend_spec)
     fig = Figure()
     fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
-    fig.legend(spec, position="JMC+jCM+w5i")
+    fig.legend(spec, position="JMC+jMC+w5i")
     return fig
 
 


pygmt/tests/test_legend.py
```diff
@@ -100,7 +100,7 @@ def test_legend_specfile(legend_spec):
         Path(specfile.name).write_text(legend_spec, encoding="utf-8")
         fig = Figure()
         fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
-        fig.legend(specfile.name, position="JTM+jCM+w5i")
+        fig.legend(specfile.name, position="JMC+jCM+w5i")
```
yvonnefroehlich: To follow our convention to set first the vertical and then the horizontal code.
Our do we want to test, that the order is in principle not important?

```suggestion
        fig.legend(specfile.name, position="JMC+jMC+w5i")
```

pygmt/tests/test_legend.py
```diff
@@ -112,7 +112,7 @@ def test_legend_stringio(legend_spec):
     spec = io.StringIO(legend_spec)
     fig = Figure()
     fig.basemap(projection="x6i", region=[0, 1, 0, 1], frame=True)
-    fig.legend(spec, position="JTM+jCM+w5i")
+    fig.legend(spec, position="JMC+jCM+w5i")
```
yvonnefroehlich:
```suggestion
    fig.legend(spec, position="JMC+jMC+w5i")
```

yvonnefroehlich (approved):

PR #4259, opened 2025-12-08
AliasSystem: Support adding a suffix to a value and simplify Figure.wiggle

seisman:
This PR adds `suffix` support to the `Alias` class, which is necessary to support some options, e.g., `wiggle`'s `-G` option has a syntax `-Gfill[+n][+p]`. Currently, we have to write a custom function to parse the argument. With the `suffix` support, it can be greatly simplified.

### DIFF
--- a/pygmt/src/wiggle.py
+++ b/pygmt/src/wiggle.py
@@ -113,9 +113,9 @@ def wiggle(  # noqa: PLR0913
         or **p** to indicate the distance unit (centimeters, inches, or
         points); if no unit is given we use the default unit that is
         controlled by :gmt-term:`PROJ_LENGTH_UNIT`.
-    positive_fill : str
+    positive_fill
         Set color or pattern for filling positive wiggles [Default is no fill].
-    negative_fill : str
+    negative_fill
         Set color or pattern for filling negative wiggles [Default is no fill].
     track : str
         Draw track [Default is no track]. Append pen attributes to use


pygmt/alias.py
```diff
@@ -283,8 +302,12 @@ def __init__(self, **kwargs):
...
         # The value of each key in kwargs is an Alias object or a sequence of Alias
         # objects. If it is a single Alias object, we will use its _value property. If
-        # it is a sequence of Alias objects, we will concatenate their _value properties
-        # into a single string.
+        # it is a sequence of Alias objects, we will check if any have suffix:
+        #
+        # - If any Alias has a suffix, return a list of values, for repeated GMT options
+        #   like -Cblue+l -Cred+r
+        # - Otherwise, concatenate into a single string for combined modifiers like
+        #   -BWSen+ttitle+gblue.
```
seisman:
```
        G=[
            Alias(fillpositive, name="fillpositive", suffix="+p"),
            Alias(fillnegative, name="fillnegative", suffix="+n"),
        ],
```
Taking `wiggle`'s `-G` option as an example. In the previous version, the values will be concatenated into a single string, so `fillpositive="blue", fillnegative="red"` will be `-Gblue+pred+n`, which is invalid.

I've updated the logic of AliasSystem, so that it will return a list of values, rather than concatenating them, when any Alias has a suffix.

pygmt/alias.py
```diff
@@ -135,9 +145,9 @@ def _to_string(
     # None and False are converted to None.
     if value is None or value is False:
         return None
-    # True is converted to an empty string with the optional prefix.
+    # True is converted to an empty string with the optional prefix and suffix.
     if value is True:
-        return f"{prefix}"
+        return f"{prefix}{suffix}"
```
weiji14, line 148: Would it be confusing if we allowed both `prefix` and `suffix` when value is True? E.g. at https://github.com/GenericMappingTools/pygmt/pull/4234#discussion_r2600852001, we used `prefix`, but it could also be `suffix`.

Or I guess it doesn't matter, and we should just be consistent that every if-branch in this `_to_string` function handles prefix and suffix.

pygmt/src/wiggle.py
```diff
@@ -188,7 +158,10 @@ def wiggle(  # noqa: PLR0913
             ),
             Alias(label, name="label", prefix="+l"),
         ],
-        G=Alias(_fills, name="positive_fill/negative_fill"),
+        G=[
+            Alias(positive_fill, name="positive_fill", suffix="+p"),
+            Alias(negative_fill, name="negative_fill", suffix="+n"),
```
weiji14, line 163: Maybe take this opportunity to update the type-hint of `postive_fill` and `negative_fill` at L47-48 above?

pygmt/alias.py
weiji14: Maybe this a short note about this should be added to the docstring. I'm afraid that this will be missed, e.g. in https://github.com/GenericMappingTools/pygmt/pull/4234#discussion_r2718540463 if we forget that `suffix` has this special treatment.

pygmt/alias.py
seisman: > Would it be confusing if we allowed both `prefix` and `suffix` when value is True? E.g. at [#4234 (comment)](https://github.com/GenericMappingTools/pygmt/pull/4234#discussion_r2600852001), we used `prefix`, but it could also be `suffix`.

`prefix` and `suffix` cannot coexist. This function does do the check to avoid too much function overhead. It's our responsibility to pass the correct `prefix`/`suffix` values.

pygmt/alias.py
seisman: What about moving the whole note lines 303-310 to docstrings?

weiji14: Sure, that works too.

pygmt/src/wiggle.py
seisman: Done in 12e451c1221f2887c4ccaa8c2f21d2dad058b7e4

pygmt/alias.py
seisman: Done in 7ceea09c73baab47ecc90e64d2f63cc67132fbc6

pygmt/src/wiggle.py
weiji14: Should remove `: str` from L116 and L118 too?

PR #4279, opened 2025-12-12
Figure.histogram: Deprecate parameter barwidth to bar_width (Will be removed in v0.20.0) 

yvonnefroehlich:

Use underscores between words in parameter names; related to #2014.

**Preview**: https://pygmt-dev--4279.org.readthedocs.build/en/4279/api/generated/pygmt.Figure.histogram.html

**Guidelines**

**Slash Commands**


pygmt/src/histogram.py
```diff
@@ -8,15 +8,22 @@
...
 
 
 @fmt_docstring
+@deprecate_parameter("barwidth", "bar_width", "v0.18.0", remove_version="v0.20.0")
 @use_alias(
     A="horizontal",
     C="cmap",
     D="annotate",
-    E="barwidth",
+    E="bar_width",
```
seisman, line 26: What about split this option into two parameter `bar_width` and `bar_offset`?

yvonnefroehlich: This would make the code more Pythonic and readable.

seisman: Are you interested in working on this? It's a good time to get familiar with the new alias system.

Better to do it in a separate PR, so that we can have two entries in the changelog, one for the deprecation, one for the enhancement.

yvonnefroehlich: Sounds good. Will give it a try 🙂.

PR #4283, opened 2025-12-12
pygmt.select: Deprecate parameter gridmask to mask_grid (Will be removed in v0.20.0) 

yvonnefroehlich:

Use underscores between words in parameter names; related to #2014.

**Preview**: https://pygmt-dev--4283.org.readthedocs.build/en/4283/api/generated/pygmt.select.html

**Guidelines**

**Slash Commands**

### DIFF
--- a/pygmt/src/select.py
+++ b/pygmt/src/select.py
@@ -23,12 +23,12 @@ __doctest_skip__ = ["select"]
 
 
 @fmt_docstring
-@deprecate_parameter("gridmask", "grid_mask", "v0.18.0", remove_version="v0.20.0")
+@deprecate_parameter("gridmask", "mask_grid", "v0.18.0", remove_version="v0.20.0")
 @use_alias(
     A="area_thresh",
     C="dist2pt",
     F="polygon",
-    G="grid_mask",
+    G="mask_grid",
     I="reverse",
     L="dist2line",
     N="mask",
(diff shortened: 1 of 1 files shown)


seisman:
I'm thinking if mask_grid makes more sense than grid_mask

yvonnefroehlich:
> I'm thinking if mask_grid makes more sense than grid_mask

Hm. I also though a bit about changing the order here and I am unsure what is better. I think having "grid" as first word should indicate that the input to this parameter is a grid. However, for me, `mask_grid` sounds better.

> gridmask ([str](https://docs.python.org/3/library/stdtypes.html#str)) – Pass all locations that are inside the valid data area of the grid gridmask. Nodes that are outside are either NaN or zero.

---

Another aspect is the alias of **-N**:
* for `pygmt.grdlandmask` it is `maks_values`
* for `pygmt.select` it is `mask`  (the [GMT docs](https://docs.generic-mapping-tools.org/dev/gmtselect.html#n) uses `mask_values`)

seisman:
> Another aspect is the alias of **-N**:
> 
> * for `pygmt.grdlandmask` it is `maks_values`
> * for `pygmt.select` it is `mask`  (the [GMT docs](https://docs.generic-mapping-tools.org/dev/gmtselect.html#n) uses `mask_values`)

Yes, this should also be consistent.

PR #4287, opened 2025-12-15
Figure.grdview: Fix the default z-plane to the grid z minimum when plane is set to True or only facede_fill/facede_pen is set

seisman:
Related to an pstream GMT bug originally found in https://github.com/GenericMappingTools/pygmt-paper-figures/pull/17#discussion_r2617288646 and fixed in https://github.com/GenericMappingTools/gmt/pull/8838.

This PR adds a workaround for it and fix the baseline image for the test.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -84,7 +84,7 @@ jobs:
 
       # Run the benchmark tests
       - name: Run benchmarks
-        uses: CodSpeedHQ/action@6a8e2b874c338bf81cc5e8be715ada75908d3871 # v4.3.4
+        uses: CodSpeedHQ/action@346a2d8a8d9d38909abd0bc3d23f773110f076ad # v4.4.1
         with:
           mode: "instrumentation"
           # 'bash -el -c' is needed to use the custom shell.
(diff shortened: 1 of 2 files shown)


pygmt/src/grdview.py
```diff
@@ -174,6 +178,18 @@ def grdview(  # noqa: PLR0913
     if plane is False and (facade_fill is not None or facade_pen is not None):
         plane = True
 
+    # Workaround for GMT bug https://github.com/GenericMappingTools/gmt/pull/8838
+    # Fix the plane value to be the grid minimum if plane=True.
+    # Notes:
+    # 1. It's the minimum of the grid, not a subset of the grid defined by `region'.
```
yvonnefroehlich:
```suggestion
    # 1. It's the minimum of the grid, not a subset of the grid defined by 'region'.
```

pygmt/src/grdview.py
```diff
@@ -174,6 +178,18 @@ def grdview(  # noqa: PLR0913
...
+    # 2. The GMT docs says "if -R was used to set zmin/zmax then we use that value if
+    #    it is less than the grid minimum value.". We can't add a workaround for this
+    #    case since we can't parse zmin/zmax from `region' if `region' was set in
+    #    previous plotting commands.
```
yvonnefroehlich:
```suggestion
    #    case since we can't parse zmin/zmax from 'region' if 'region' was set in
    #    previous plotting commands.
```

PR #4296, opened 2025-12-17
Figure.inset: Deprecate parameter margin to clearance (Will be removed in v0.20.0)

yvonnefroehlich:

To make the parameter name consistent with `Figure.suplot` and to consider that a list can be passed with different margins for the different sides.
**_Note_**: Since GMT 6.4.0 **-M** is changed to **-C** (see https://github.com/GenericMappingTools/pygmt/pull/4296#discussion_r2625629349), so it should be not `margins` but `clearance` (see https://github.com/GenericMappingTools/pygmt/pull/4296#issuecomment-3663727351). 

**Preview**:

**Guidelines**

**Slash Commands**

### DIFF
--- a/pygmt/src/inset.py
+++ b/pygmt/src/inset.py
@@ -116,7 +116,7 @@ def inset(
     Examples
     --------
     >>> import pygmt
-    >>> from pygmt.params import Box
+    >>> from pygmt.params import Box, Position
     >>>
     >>> # Create the larger figure
     >>> fig = pygmt.Figure()
@@ -133,9 +133,8 @@ def inset(
     ...         dcw="MG+gred",
     ...     )
     ...
-    >>> # Map elements outside the "with" statement are plotted in the main
-    >>> # figure
-    >>> fig.logo(position="jBR+o0.2c+w3c")
+    >>> # Map elements outside the "with" statement are plotted in the main figure
+    >>> fig.logo(position=Position("BR", offset=0.2), width="3c")
     >>> fig.show()
     """
     self._activate_figure()
--- a/.github/ISSUE_TEMPLATE/5-bump_gmt_checklist.md
+++ b/.github/ISSUE_TEMPLATE/5-bump_gmt_checklist.md
@@ -31,6 +31,7 @@ mamba install -c conda-forge gmt=X.Y.Z
 - [ ] Run `grep -r "# TODO(GMT.*)" --include="*.py" .` to find TODO items related to GMT
   - [ ] Remove related workarounds/patches
   - [ ] Fix tests with pytest `xfail`/`skipif` markers that now pass
+- [ ] Update URLs to manually inserted images of the GMT documentation
 - [ ] Update this checklist if necessary
 
 **To-Do for bumping the minimum required GMT version**:
--- a/examples/gallery/embellishments/gmt_logo.py
+++ b/examples/gallery/embellishments/gmt_logo.py
@@ -7,12 +7,12 @@ The :meth:`pygmt.Figure.logo` method allows to place the GMT logo on a figure.
 
 # %%
 import pygmt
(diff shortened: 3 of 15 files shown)


pygmt/src/inset.py
```diff
@@ -8,15 +8,22 @@
...
 from pygmt.params import Box
 
 __doctest_skip__ = ["inset"]
 
 
 @fmt_docstring
+@deprecate_parameter("margin", "margins", "v0.18.0", remove_version="v0.20.0")
 @contextlib.contextmanager
-@use_alias(D="position", M="margin")
+@use_alias(D="position", M="margins")
```
seisman: It was -M in GMT<=6.3.0 but has changed to `-C` since GMT 6.4.0.
```suggestion
@use_alias(D="position", C="margins")
```

pygmt/src/inset.py
```diff
@@ -90,7 +97,7 @@ def inset(
         box is drawn using :gmt-term:`MAP_FRAME_PEN`. To customize the box appearance,
         pass a :class:`pygmt.params.Box` object to control style, fill, pen, and other
         box properties.
-    margin : float, str, or list
+    margins : float, str, or list
         This is clearance that is added around the inside of the inset.
```
seisman, line 101: Also need to update the docstrings.

seisman:
It's a good time to decide the implemention of the `margins` parameter. The GMT documentation says:

> Reserve a space of dimension clearance between the actual inset plot area and the given inset box on the specified side, using side values from w, e, s, or n, or x for both w and e or y for both s and n. No side means all sides. The option is repeatable to set aside space on more than one side. Alternatively, if all sides are to be set you can also give a pair of values separated by slashes (for setting separate horizontal and vertical margins), or the full set of four separate margins. Such space will be left untouched by the inset map plotting. Append units as desired [Default is set by [PROJ_LENGTH_UNIT](https://docs.generic-mapping-tools.org/6.6/gmt.conf.html#term-PROJ_LENGTH_UNIT)].

So, the GMT CLI usage can be:

- `-C1`: for all sides
- `-C1/2`: for x and y sides, respectively
- `-C1/2/3/4`: for w, e, s, and n sides, respectively

Or, it's possible to set margins with repeated options, e.g., 

- `-Cx1`
- `-Cx1 -Cy2`
- `-Cw1 -Cs3`
- `-Cw1 -Ce2 -Cs3 -Cn4`

I feel it's technically possible but difficult to support both. To keep it simple, I would prefer to only support the first usage, i.e.,:

- `margins=1`
- `margins=(1, 2)`
- `margins=(1, 2, 3, 4)`

To set margins for specific sides only, users can set it to 0. For example,

- `-Cx1` is equivalent to `margins=(1, 0)`
- `-Cw1 -Cs3` is equivalent to `margins=(1, 0, 3, 0)`

seisman:
Actually, it should be `clearance` instead. 

In subplot, `-C` is aliased to `clearance` and `-M` is aliased to `margins`.

pygmt/src/inset.py
yvonnefroehlich: Good to know - thanks for potting out!

yvonnefroehlich:
> Actually, it should be `clearance` instead.
> 
> In subplot, `-C` is aliased to `clearance` and `-M` is aliased to `margins`.

With the change from **-M** to **-C** we should also change the parameter name in PyGMT to have this consistent. Updatde to `clearance` in commit https://github.com/GenericMappingTools/pygmt/pull/4296/commits/77faddeafbb19c34f7d16553f4b572b5cd120a55.

seisman (approved):

seisman:
I feel this PR is good to merge. We can update the docstrings of `Figure.inset` and `Figure.subplot` in a separate PR.

PR #4297, opened 2025-12-17
pygmt.project: Deprecate parameter unit to units (Will be removed in v0.2X.0)

yvonnefroehlich:

Use the plural `units` if the units of multiple quanities can be adjusted; related to #2014.

**Preview**:

**Guidelines**

**Slash Commands**

### DIFF
--- a/pygmt/src/grdproject.py
+++ b/pygmt/src/grdproject.py
@@ -10,14 +10,13 @@ from pygmt._typing import PathLike
 from pygmt.alias import Alias, AliasSystem
 from pygmt.clib import Session
 from pygmt.exceptions import GMTInvalidInput
-from pygmt.helpers import build_arg_list, deprecate_parameter, fmt_docstring, use_alias
+from pygmt.helpers import build_arg_list, fmt_docstring, use_alias
 
 __doctest_skip__ = ["grdproject"]
 
 
 @fmt_docstring
-@deprecate_parameter("unit", "units", "v0.18.0", remove_version="v0.20.0")
-@use_alias(E="dpi", F="scaling", I="inverse", M="units", n="interpolation")
+@use_alias(E="dpi", F="scaling", I="inverse", M="unit", n="interpolation")
 def grdproject(
     grid: PathLike | xr.DataArray,
     outgrid: PathLike | None = None,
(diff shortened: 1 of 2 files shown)


pygmt/src/grdproject.py
```diff
@@ -10,13 +10,14 @@
...
-from pygmt.helpers import build_arg_list, fmt_docstring, use_alias
+from pygmt.helpers import build_arg_list, deprecate_parameter, fmt_docstring, use_alias
 
 __doctest_skip__ = ["grdproject"]
 
 
 @fmt_docstring
-@use_alias(E="dpi", F="scaling", I="inverse", M="unit", n="interpolation")
+@deprecate_parameter("unit", "units", "v0.18.0", remove_version="v0.20.0")
+@use_alias(E="dpi", F="scaling", I="inverse", M="units", n="interpolation")
```
seisman: I think it should be `unit`, not `units`, since it takes one single unit in this module.

yvonnefroehlich: I was wondering it it makes sense to have this consistent across function / methods and always use plural.

seisman: I think we should use `unit` and `units` in different cases.

yvonnefroehlich: So, we only want to change `pygmt.project`?

seisman:
> Specify that x, y, r, s are in degrees while p, q, dist, lmin, lmax, wmin, wmax are in km. If -Q is not set, then all these are assumed to be in the same units.

Actually, I feel aliasing `-Q` to `unit` is a mistake. GMT.jl and GMT use `km` instead, but I  feel `km` is also not a good name.

yvonnefroehlich:
> > Specify that x, y, r, s are in degrees while p, q, dist, lmin, lmax, wmin, wmax are in km. If -Q is not set, then all these are assumed to be in the same units.
> 
> Actually, I feel aliasing `-Q` to `unit` is a mistake. GMT.jl and GMT use `km` instead, but I feel `km` is also not a good name.

I have to look at this in more detail, as I do not use `pygmt.project` often. But agree `km` is not optimal as a name. Converted the PR to draft.

Issue #4252, opened 2025-12-05
pygmt.grdfilter: Filter a grid in the space (or time) domain


seisman:
*This issue serves as the central place for discussing and tracking the implementation of the `grdfilter` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/grdfilter.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/grdfilter
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.grdfilter.html

## GMT Option Flags and Modifiers

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [ ] `<ingrid>`
- [ ] `-D`
- [ ] `-F`
- [ ] `-G`: `outgrid`
- [ ] `-I`
- [ ] `-N`
- [ ] `-R`: `region`
- [ ] `-T`
- [ ] `-V`: `verbose`
- [ ] `-f`
- [ ] `-r`
- [ ] `-x`
- [ ] ~~`--PAR=value`~~: Use `pygmt.config` instead.

## Linked Pull Requests

- [ ] Initial feature implementation
- [ ] Add a tutorial or gallery example

## Related Issues and Discussions

*Add links to related wrapper discussions, API design threads, or upstream GMT changes here.*

PR #4310, opened 2025-12-20
pygmt.grdfilter: Let the parameter 'nans' support descriptive arguments

seisman:
This PR migrates the `nans` parameter to the new alias system and also make it support descrptive arguments `ignore`, `replace` and `preserve`.

**Preview**: 

- GMT documentation: https://docs.generic-mapping-tools.org/6.6/grdfilter.html#n
- GMT.jl documentation: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/grdfilter.html

Currently, GMT recommends `ignore`/`coregnan`/`anynan`, but I feel they're not good. An upstream PR at https://github.com/GenericMappingTools/gmt/pull/8842.

Related to #4252.

### DIFF
--- a/pygmt/src/grdfilter.py
+++ b/pygmt/src/grdfilter.py
@@ -125,8 +125,8 @@ def grdfilter(
     --------
     >>> from pathlib import Path
     >>> import pygmt
-    >>> # Apply a filter of 600 km (full width) to the @earth_relief_30m_g file and
-    >>> # return a filtered field (saved as netCDF)
+    >>> # Apply a median filter of 600 km (full width) to the @earth_relief_30m_g grid
+    >>> # and return a filtered grid (saved as netCDF file).
     >>> pygmt.grdfilter(
     ...     grid="@earth_relief_30m_g",
     ...     filter="m600",


pygmt/src/grdfilter.py
```diff
@@ -89,21 +91,22 @@ def grdfilter(
...
-        Determine how NaN-values in the input grid affect the filtered output.
-        Use **i** to ignore all NaNs in the calculation of the filtered value
-        [Default]. **r** is same as **i** except if the input node was NaN then
-        the output node will be set to NaN (only applies if both grids are
-        co-registered). **p** will force the filtered value to be NaN if any
-        grid nodes with NaN-values are found inside the filter circle.
-    $region
+    nans
+        Determine how NaN-values in the input grid affect the filtered output. Choose
+        one of:
```
yvonnefroehlich:
```suggestion
        Determine how NaN-values in the input grid affect the filtered output grid.
        Choose one of:
```

pygmt/src/grdfilter.py
```diff
@@ -89,21 +91,22 @@ def grdfilter(
           calculation.
         - ``"5"``: grid (x,y) in Mercator ``projection="m1"`` img units,
           *width* in km, Spherical distance calculation.
-
     $spacing
-    nans : str or float
```
yvonnefroehlich, line 94: Seems like the input datatype "float" was actually not correct / a typo.

pygmt/src/grdfilter.py
```diff
@@ -133,13 +136,16 @@ def grdfilter(
     ...     outgrid="filtered_pacific.nc",
     ... )
     >>> Path("filtered_pacific.nc").unlink()  # Cleanup file
-    >>> # Apply a Gaussian smoothing filter of 600 km to the input DataArray
-    >>> # and return a filtered DataArray with the smoothed field
+    >>> # Apply a Gaussian smoothing filter of 600 km to the input DataArray and return
+    >>> # a filtered DataArray with the smoothed field
```
yvonnefroehlich:
```suggestion
    >>> # Apply a Gaussian smoothing filter of 600 km to the input DataArray and return
    >>> # a filtered DataArray with the smoothed grid.
```

pygmt/src/grdfilter.py
```diff
@@ -89,21 +91,22 @@ def grdfilter(
...
-        grid nodes with NaN-values are found inside the filter circle.
-    $region
+    nans
+        Determine how NaN-values in the input grid affect the filtered output. Choose
+        one of:
+
+        - ``"ignore"``: Ignore all NaNs in the calculation of filtered value [Default].
+        - ``"replace"``: Similar to ``"ignore"`` except if the input node was NaN then
+          the output node will be set to NaN (only applies if both grids are
+          co-registered).
```
yvonnefroehlich, line 102:
```suggestion
        - ``"replace"``: Similar to ``"ignore"`` except if the input node was NaN then
          the output node will be set to NaN (only applied if both grids are
          co-registered).
```

pygmt/src/grdfilter.py
```diff
@@ -122,8 +125,8 @@ def grdfilter(
     --------
     >>> from pathlib import Path
     >>> import pygmt
-    >>> # Apply a filter of 600 km (full width) to the @earth_relief_30m_g file
-    >>> # and return a filtered field (saved as netCDF)
+    >>> # Apply a filter of 600 km (full width) to the @earth_relief_30m_g file and
+    >>> # return a filtered field (saved as netCDF)
```
yvonnefroehlich:
```suggestion
    >>> # Apply a median filter of 600 km (full width) to the @earth_relief_30m_g grid
    >>> # and return a filtered grid (saved as netCDF file).
```

Not sure what "full width" means. The [GMT docs for **D** of `grdfilter`](https://docs.generic-mapping-tools.org/dev/grdfilter.html#d) say for `D=4` (`distance=4`):

> flag = 4: grid (x, y) in degrees, width in km, Spherical distance calculation.

PR #4353, opened 2026-01-14
CI: Combine the steps for building GMT source code in the 'GMT Dev Tests' workflow

seisman:
With changes in #4329, now it's possible to have a single step for building GMT source code on all platforms.

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -104,8 +104,7 @@ jobs:
       # Build GMT from source on Linux/macOS, script is adapted from
       # https://github.com/GenericMappingTools/gmt/blob/6.5.0/ci/build-gmt.sh
       - name: Build GMT from source
-        # Use non-login shell on Windows and login shell on Linux/macOS
-        shell: ${{ runner.os == 'Windows' && 'bash' || 'bash -l {0}' }}
+        shell: bash
         run: |
           cd gmt/
           mkdir build


.github/workflows/ci_tests_dev.yaml
```diff
@@ -103,44 +103,30 @@ jobs:
...
-          cd ..
-          rm -rf gmt/
-        env:
-          GMT_INSTALL_DIR: ${{ runner.temp }}/gmt-install-dir
-        if: runner.os != 'Windows'
-
-      - name: Build GMT on Windows
-        shell: bash
+      - name: Build GMT from source
+        shell: bash -l {0}
```
weiji14: Can probably remove this since already set in defaults above. Should have done this in https://github.com/GenericMappingTools/pygmt/pull/4329#discussion_r2662684385

```suggestion
```

.github/workflows/ci_tests_dev.yaml
```diff
@@ -103,44 +103,30 @@ jobs:
...
             -DGMT_ENABLE_OPENMP=TRUE \
             -DGMT_USE_THREADS=TRUE
           cmake --build .
           cmake --build . --target install
-          cd ..
-          rm -rf gmt/
+          cd ../../
+          rm -rf gmt
         env:
           GMT_INSTALL_DIR: ${{ runner.temp }}/gmt-install-dir
```
weiji14, line 129: See if this would work, based on https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#example

```suggestion
          CMAKE_PREFIX_PATH: ${{ startsWith(matrix.os, 'windows') && '' || '-DCMAKE_PREFIX_PATH=$MAMBA_ROOT_PREFIX/envs/pygmt/Library' }}
          GMT_INSTALL_DIR: ${{ runner.temp }}/gmt-install-dir
```

.github/workflows/ci_tests_dev.yaml
```diff
@@ -103,44 +103,30 @@ jobs:
...
         run: |
           cd gmt/
           mkdir build
           cd build
-          cmd.exe /c "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          if [[ "$RUNNER_OS" == "Windows" ]]; then
+            cmd.exe /c "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+            CMAKE_PREFIX_PATH="-DCMAKE_PREFIX_PATH=$MAMBA_ROOT_PREFIX/envs/pygmt/Library"
+          else
+            CMAKE_PREFIX_PATH=""
```
weiji14:
```suggestion
```

seisman:
Still not lucky. It seems that we need `shell: bash -l ${0}` (the default) for macOS/Linux but `shell: bash` for Windows.

seisman:
> Still not lucky. It seems that we need `shell: bash -l ${0}` (the default) for macOS/Linux but `shell: bash` for Windows.

I still don’t fully understand the underlying cause, but the observation appears reliable.

On Windows, the login shell seems to interfere with the Visual Studio activation script (`"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"`), so a **non-login** shell is required and `CMAKE_PREFIX_PATH` must point to `"$MAMBA_ROOT_PREFIX/envs/pygmt/Library"`.

On Linux/macOS, a **login** shell automatically activates the environment, and CMake locates the libraries without help; with a **non-login** shell, we need to set `CMAKE_PREFIX_PATH` to `"$MAMBA_ROOT_PREFIX/envs/pygmt"`.

Now building source codes succeeds on all platforms. The failing test is unrelated.

CI Runs: https://github.com/GenericMappingTools/pygmt/actions/runs/20982948742

.github/workflows/ci_tests_dev.yaml
seisman: I didn't try this, because now we need to set `CMAKE_PREFIX_PATH` to different values based on the OS.

weiji14 (approved):
Looks good. Just curious though, you know what's up with the skipped job at https://github.com/GenericMappingTools/pygmt/actions/runs/20982938310/job/60311514437?

PR #4358, opened 2026-01-17
Support passing np.timedelta64 to parameters that expect a sequence (e.g., 'region')

seisman:
Related to https://github.com/GenericMappingTools/pygmt/issues/2848.

```python
import pygmt
import numpy as np

fig = pygmt.Figure()
fig.basemap(
    projection="X8c/5c",
    region=[np.timedelta64(0, "D"), np.timedelta64(8, "D"), 0, 10],
    frame=["WSne", "xaf+lForecast Days", "yaf+lRMSE"],
)
fig.plot(
    x=np.arange(np.timedelta64(0, "D"), np.timedelta64(8, "D")),
    y=np.geomspace(start=0.1, stop=9, num=8),
    style="c0.2c",
    pen="1p",
)
fig.show()
```
timedelta64

### DIFF
--- a/pygmt/helpers/utils.py
+++ b/pygmt/helpers/utils.py
@@ -884,14 +884,8 @@ def sequence_join(
         # or integer number for timedelta64.
         _values = []
         for item in value:
-            if isinstance(item, np.timedelta64):
-                # Convert timedelta64 to numeric value. For generic units, preserve the
-                # raw integer value; for specific units, convert in that unit.
-                unit = np.datetime_data(item.dtype)[0]
-                if unit == "generic":
-                    _values.append(str(int(item)))
-                else:
-                    _values.append(str(item.astype(f"timedelta64[{unit}]").astype(int)))
+            if isinstance(item, np.timedelta64):  # Convert timedelta64 to numeric value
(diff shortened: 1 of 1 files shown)


pygmt/helpers/utils.py
```diff
@@ -854,16 +876,28 @@ def sequence_join(
...
+        _values = []
+        for item in value:
+            if isinstance(item, np.timedelta64):
+                # Convert timedelta64 to numeric value. For generic units, preserve the
+                # raw integer value; for specific units, convert in that unit.
+                unit = np.datetime_data(item.dtype)[0]
+                if unit == "generic":
+                    _values.append(str(int(item)))
+                else:
+                    _values.append(str(item.astype(f"timedelta64[{unit}]").astype(int)))
```
weiji14: Why is item converted into a timedelta64 value again here?

seisman: You're right, conversion is not necessary. Simplified in 92280b90b963c6b6f3ae66d7b62d0ffef4da96a7.

weiji14 (approved):

PR #4361, opened 2026-01-18
pygmt.info: Migrate the 'per_column' parameter to the new alias system

seisman:


pygmt/src/info.py
```diff
@@ -57,8 +54,9 @@ def info(
     data
         Pass in either a file name to an ASCII data table, a 1-D/2-D
         $table_classes.
-    per_column : bool
-        Report the min/max values per column in separate columns.
+    per_column
+        Report the min/max values per column in separate columns [Default uses <min/max>
+        format].
```
yvonnefroehlich:
```suggestion
    per_column
        Report the min/max values per column in separate columns [Default is the format
        <min/max>].
```

PR #4367, opened 2026-01-22
Fix the failing tests for the updated gravimetric grids

weiji14:

Specifically the `earth_faa`, `earth_faaerror`, `earth_ndefl`, `earth_edefl` and `earth_vgg` grids. Xref https://github.com/GenericMappingTools/gmtserver-admin/issues/291

Fixes #

**Preview**:

**Guidelines**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_datasets_earth_deflection.py
+++ b/pygmt/tests/test_datasets_earth_deflection.py
@@ -23,8 +23,8 @@ def test_earth_edefl_01d():
     assert data.gmt.registration is GridRegistration.GRIDLINE
     npt.assert_allclose(data.lat, np.arange(-90, 91, 1))
     npt.assert_allclose(data.lon, np.arange(-180, 181, 1))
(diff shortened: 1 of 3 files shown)


Esteban82:
Out of curiosity: what is the main goal of these tests?
Are they meant to verify that the provided grids are correct, to check that the command runs as expected, or something else?

weiji14:
Yes, we're just checking that the function runs ok, and the grids returns the expected values (no NaNs, etc). It helps when we do major refactoring, to ensure nothing goes wrong.

Esteban82:
It sounds to me that you could use the static server instead. 
https://static.generic-mapping-tools.org/

This is its readme file: 
```
# This directory is meant to be static, i.e., we only place unchanged data
# sets that our test scripts need to use so that we do not get massive test
# failures when the topography is updated a bit.
```

Right now it doesn't have any gravimetric data, but I think we could add it.

PR #4370, opened 2026-01-26
BREAKING: Raise GMTRequiredParameterError/GMTConflictParameterError exception for invalid values. Previously raise #3707

Chuan1937:
Addressing #3707
This PR adds a new parameter exception class, GMTParameterError, and its subclasses for reporting parameter-related errors. 

The error message format is: 
`Missing required parameter: {parameters}. Explain the reason. `
 `Conflicting parameters: {parameters}. Explain the reason.`

### DIFF
--- a/pygmt/tests/test_alias_system.py
+++ b/pygmt/tests/test_alias_system.py
@@ -4,7 +4,7 @@ Tests for the alias system.
 
 import pytest
 from pygmt.alias import Alias, AliasSystem
-from pygmt.exceptions import GMTInvalidInput
+from pygmt.exceptions import GMTConflictParameterError
 from pygmt.helpers import build_arg_list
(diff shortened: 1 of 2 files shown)


seisman:
@Chuan1937 Thanks for your work on this! There's actually a similar PR at #4003 that takes a different approach. It would be great to compare the two implementations, decide on the best path forward, and then combine efforts accordingly.

Chuan1937:
> @Chuan1937 Thanks for your work on this! There's actually a similar PR at #4003 that takes a different approach. It would be great to compare the two implementations, decide on the best path forward, and then combine efforts accordingly.

ok,I'll do it.

seisman:
Duplicate of PR #4003.

PR #4401, opened 2026-02-09
pygmt.grdfilter: Add parameters filter/width/highpass to set the filter [Part 1]

seisman:
This PR is a subset of PR #4378, aiming to implement a Pythonic interface for the `-F` option of the `grdfilter` module.

The full syntax of the `-F` option is (https://docs.generic-mapping-tools.org/dev/grdfilter.html#f):

**-Fx***width*[/*width2*][**+c**\|**+h**\|**+l**\|**+q***quantile*\|**+u**]

GMT supports 12 different types of filters, which can be specified by **x**. For most filters, the syntax is very simple (e.g., `-Fg600`). Some filters can accept more modifiers. For example, `histogram` filter accepts `+c`/`+l`/`+u`; `median` filter accepts `+q`; `Maximum likelihood probability` filter accepts `+l`/`+u`; and all filters accept `+h` (for highpass). For custom and operator filters, *width* is actually a weight file. As a result, this option is complicated to implement. 

PR #4378 tries to alias the `-F` option to multiple parameters `filter_type`/`filter_width`/`highpass`/`hist_bin_width`/`median_quantile`/`hist_center_bins`/`mode_extreme`. As you can see, that PR is already very big and there are still many edge cases to be covered. I feel that PR will take a very long time to finish and will definitely delay the submission of the paper. Instead, I'm opening this PR to implement a subset of the `-F` option, which should cover most (80%, I guess) use cases of the `-F`option. A more feature-complete version can be added after the paper submission.

Previously, the `-F` option is aliased to a single parameter `filter`. In this PR, three new parameters are added:

- `filter_type`: specify the filter type
- `filter_width`: specify the filter width
- `highpass`: for the `+h` modifier

Then, in our paper, `filter="g600"` would be `filter_type="gaussian", filter_width=600`.

We still need to discuss the parameter names:

- `filter` is a built-in function in Python, so the old alias `filter` is not a good name.
- `filter_type`/`filter_width` are good, but the prefix `filter_` is not necessary as the method name is `grdfilter` (and `da.gmt.filter` in the GMT accessor). But `type` is also a built-in function in Python.

I feel we have some choices:

1. `filter_type`/`filter_width`
2. `filter_type`/`width`
3. `filter`/`width`
4. `type`/`width` 

I'm a little inclined to `filter`/`width`. What do you think @GenericMappingTools/pygmt-maintainers?

**Edit**: I've renamed the parameters to `filter`/`width`.

**Preview**: https://pygmt-dev--4401.org.readthedocs.build/en/4401/api/generated/pygmt.grdfilter.html

### DIFF
--- a/pygmt/src/magnetic_rose.py
+++ b/pygmt/src/magnetic_rose.py
@@ -110,11 +110,15 @@ def magnetic_rose(  # noqa: PLR0913
 
     position = _parse_position(position, default=Position("BL", cstype="inside"))
 
-    if declination_label is not None and declination is None:
-        raise GMTParameterError(
-            required="declination",
-            reason="Required when 'declination_label' is set.",
-        )
+    if declination_label is not None:
+        if declination is None:
+            raise GMTParameterError(
+                required="declination",
+                reason="Required when 'declination_label' is set.",
+            )
+        # Upstream issue that declination label with spaces is not properly handled.
+        if " " in declination_label:
+            declination_label = declination_label.replace(" ", "\\040")
 
     aliasdict = AliasSystem(
         F=Alias(box, name="box"),
--- a/pygmt/tests/baseline/test_magnetic_rose_complete.png.dvc
+++ b/pygmt/tests/baseline/test_magnetic_rose_complete.png.dvc
@@ -1,5 +1,5 @@
 outs:
-- md5: 78c53322b2514fbcd56483b793d47ebe
-  size: 28831
+- md5: c39fb0511ea0a4b92d8d0202252d72ad
+  size: 28825
   hash: md5
   path: test_magnetic_rose_complete.png
--- a/pygmt/tests/test_magnetic_rose.py
+++ b/pygmt/tests/test_magnetic_rose.py
@@ -34,7 +34,7 @@ def test_magnetic_rose_complete():
         outer_pen="1p,red",
         inner_pen="1p,blue",
         declination=11.5,
-        declination_label="11.5°E",
+        declination_label="11.5 °E",
     )
     return fig
 


pygmt/src/grdfilter.py
```diff
@@ -9,16 +9,76 @@
...
+            filter_type,
+            name="filter_type",
+            mapping={
+                "boxcar": "b",
+                "cosine_arch": "c",
+                "gaussian": "g",
+                "minall": "l",
+                "minpos": "L",
+                "maxall": "u",
+                "maxneg": "U",
```
seisman: These long names come from https://github.com/GenericMappingTools/gmt/blob/a1d6cdfca838326e324b809a8e5ace6a61e7f8c6/src/longopt/grdfilter_inc.h#L29-L32.

pygmt/src/grdfilter.py
```diff
@@ -58,19 +117,35 @@ def grdfilter(
...
+
+      **Note**: There are still a few other filter types available in GMT (e.g.,
+      histogram and mode filters), but they are not implemented in PyGMT yet. As a
+      workaround, pass the raw GMT command string to this parameter to use these other
+      filter types. Refer to :gmt-docs:`grdfilter.html#f` for the full syntax of this
+      parameter.
+    width
+        The full diameter width of the filter. It can be a single value for an isotropic
+        filter, or a pair of values for a rectangular filter (width in x- and
+        y-directions, requiring ``distance`` be either ``"p"`` or ``0``).
```
seisman: Update after #4405
```suggestion
        y-directions, requiring ``distance`` be either ``"pixel"`` or ``"cartesian"``).
```

PR #4406, opened 2026-02-10
Add the Frame/Axis class for setting frame and axes

seisman:
This PR implements the `Frame`/`Axis` class for setting frame and axes attributes. It's a subset of PR #4016 because the `-B` option is complicated, and it will take a long time to finish #4016, which again will delay the paper submission.

Currently, only a small subset of the `-B` option is implemented in this PR, but it should already cover most use cases.

Here is a comparison table between the GMT CLI and the PyGMT versions:

| GMT | PyGMT |
|---|---|
| `-B` | `frame=True` |
| `-BWSEN` | `frame=Frame("WSEN")` or `frame=Frame(axes="WSEN")` |
| `-BWSEN+tMY TITLE`| `frame=Frame("WSEN", title="MY TITLE")` |
| `-Ba` | `frame=Axis(annnot=True)` |
| `-Bafg`| `frame=Axis(annot=True, tick=True, grid=True)` |
| `-Ba30f15g5`| `frame=Axis(annot=30, tick=15, grid=5)` |
| `-Ba30f15g5+lLABEL`| `frame=Axis(annot=30, tick=15, grid=5, label="LABEL")` |
| `-BWSEN+tTITLE -Bxa30f15g5+lXlabel -Bya4f2g1+lYlabel` | `frame=Frame("WSEN", title="TITLE", xaxis=Axis(annot=30, tick=15, grid=5, label="Ylabel"), yaxis=Axis(annot=4, tick=4, grid=1, label="Ylabel"))` |

This PR doens't break backward-compatibility, so `frame="WSEN"`, `frame="afg"`, `frame="a30f15g5"`, `frame=["WSEN+tTITLE", "xafg", "yafg"]` are still supported.

**Preview**:

- https://pygmt-dev--4406.org.readthedocs.build/en/4406/api/generated/pygmt.params.Frame.html
- https://pygmt-dev--4406.org.readthedocs.build/en/4406/api/generated/pygmt.params.Axis.html

PR #4407, opened 2026-02-10
Figure.colorbar: Add parameters label/unit/annot/tick/grid and more to set colorbar annotations/labels

seisman:
See https://github.com/GenericMappingTools/pygmt/issues/4243#issuecomment-3878550996 for context. After #4406.

**Preview**: https://pygmt-dev--4407.org.readthedocs.build/en/4407/api/generated/pygmt.Figure.colorbar.html#pygmt.Figure.colorbar

### DIFF
--- a/.github/workflows/cache_data.yaml
+++ b/.github/workflows/cache_data.yaml
@@ -74,7 +74,7 @@ jobs:
 
       # Upload the downloaded files as artifacts to GitHub
       - name: Upload artifacts to GitHub
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         with:
           name: gmt-cache
           include-hidden-files: true
--- a/.github/workflows/cache_dvc.yaml
+++ b/.github/workflows/cache_dvc.yaml
@@ -52,7 +52,7 @@ jobs:
           dvc pull --no-run-cache --verbose && ls -lhR pygmt/tests/baseline/
 
       - name: Upload DVC cache as artifacts to GitHub
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         with:
           name: dvc-cache
           include-hidden-files: true
--- a/.github/workflows/ci_docs.yml
+++ b/.github/workflows/ci_docs.yml
@@ -150,7 +150,7 @@ jobs:
           cd ../..
 
       - name: Upload PDF as artifact for previewing on pull requests
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         if: github.event_name == 'pull_request' && matrix.os == 'ubuntu-latest'
         with:
           name: artifact-pygmt-docs-pdf
--- a/.github/workflows/ci_tests.yaml
+++ b/.github/workflows/ci_tests.yaml
@@ -190,7 +190,7 @@ jobs:
 
       # Upload diff images on test failure
       - name: Upload diff images if any test fails
-        uses: actions/upload-artifact@v6.0.0
+        uses: actions/upload-artifact@v7.0.0
         if: failure()
         with:
           name: artifact-${{ matrix.os }}-${{ matrix.python-version }}
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -187,7 +187,7 @@ jobs:
 
       # Upload diff images on test failure
       - name: Upload diff images if any test fails
-        uses: actions/upload-artifact@v6.0.0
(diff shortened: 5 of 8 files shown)


seisman:
Ping @GenericMappingTools/pygmt-maintainers for final reviews.

pygmt/src/colorbar.py
```diff
@@ -244,6 +317,24 @@ def colorbar(  # noqa: PLR0913
...
+    tick
+    grid
+        Intervals for annotations, ticks, and gridlines. Refer to
+        :class:`pygmt.params.Axis` for more details on how these parameters work.
+    annot_prefix
+    annot_unit
+    annot_angle
+        The prefix, unit and angle for the annotations. The prefix is placed before the
+        annotation text; the unit is placed after the annotation text; and the angle is
+        the angle of the annotation text.
```
yvonnefroehlich, line 334: I think this is coming from to be consistent with GMT.
I personally find it confusing to call the y-label "unit" in case the frame is related to a colorbar. Thus we have here "annot_unit" for the unit opitionally added after the annotations, but "unit" in case the frame is related to a basemap.

seisman: I'm also debating which one to choose:

- If we follow the convention used for the `frame` parameter, the names should be  `xlabel`/`ylabel`/`annot_prefix`/`annot_unit`/`annot_angle` (x- and y-label mean labels of long and short dimensions of the colorbar)
- If we follow the convention used by the GMT colorbar docs, the names should be `label`/`unit`/`annot_prefix`/`annot_unit`/`annot_angle` (`label` and `unit` are text strings displayed along the long and short dimensions of a colorbar; `unit` doesn't have to be a "unit"; `unit` and `annnot_unit` may be more confusing)

Or maybe we should call them `xlabel`/`ylabel`/`annot_prefix`/`annot_suffix`/`annot_angle`?

yvonnefroehlich: > I'm also debating which one to choose:
> 
> * If we follow the convention used for the `frame` parameter, the names should be  `xlabel`/`ylabel`/`annot_prefix`/`annot_unit`/`annot_angle` (x- and y-label mean labels of long and short dimensions of the colorbar)

Hm. Maybe I am already confused by myself 🙃. But do you maybe mean `xlabel`/`ylabel`/`prefix`/`unit`/`angle` here?

seisman: No, I meant `annot_prefix`/`annot_unit`/`annot_angle`.

While `prefix`/`unit`/`angle` work well in the `Axis` class for the `frame` parameter, they're too ambiguous for colorbar annotations. In a colorbar context, it's unclear what `prefix` mean? I think adding the `annot_` prefix can make the parameter's purpose clear to understand.

seisman: Looking at the changes in https://github.com/GenericMappingTools/pygmt-paper-figures/pull/39/changes, only `label`/`unit`/`annot`/`tick`/`grid` are used in the examples, and `annot_prefix`/`annot_unit`/`annot_angle` are not used.

So we just need to decide if we want `xlabel`/`ylable` or `label`/`unit`. We can discuss/implement `annot_prefix`/`annot_unit`/`annot_angle` in a separate PR instead.

yvonnefroehlich: I prefer `xlabel` / `ylabel`. @weiji14, @michaelgrund what do you think / prefer?

seisman: I also prefer `xlabel`/`ylabel`, so we have at least two votes in favor. We'll proceed with `xlabel`/`ylabel` unless any of @weiji14, @michaelgrund, @willschlitzer   has strong objections.

weiji14: I somewhat prefer label/unit because it'll get confusing when you have a vertical colorbar (see elevation example in https://www.pygmt.org/v0.18.0/gallery/embellishments/colorbar.html), in which case xlabel (label) is on the 'y-axis' and ylabel (unit) is on the x-axis. But no strong opinions if we decide to go with xlabel/ylabel.

seisman: That's a good point.

yvonnefroehlich: Another aspect is, that we have for the `move_text` parameter already `"label"` and `"unit"` (and `"annotations"`) as arguments. If we use now `xlabel` and `ylabel` as parameter names we probably have to adjust the arguments for `move_text`.

seisman: So let's follow the GMT convention and use label/unit. Users need to read the documentation anyway

yvonnefroehlich: Agree, with these two arguments, keeping `label` and `unit` appears better to me.

Issue #4416, opened 2026-02-14
pygmt.grdsample: Resample a grid onto a new lattice


seisman:
*This issue serves as the central place for discussing and tracking the implementation of the `grdsample` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/grdsample.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/grdsample
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.grdsample.html

## GMT Option Flags and Modifiers

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [X] `ingrid`
- [X] `-G`: `outgrid`
- [X] `-I`: `spacing`
- [X] `-R`: `region`
- [X] `-T`: `translate`
- [X] `-V`: `verbose`
- [X] `-f`: `coltypes`
- [X] `-n`: `interpolation`
- [X] `-r`: `registration`
- [X] `-x`: `cores`
- [ ] ~~`--PAR`: Use `pygmt.config` instead~~

## Linked Pull Requests

- [X] Initial feature implementation #1380
- [ ] Add a tutorial or gallery example

PR #4417, opened 2026-02-14
pygmt.grdsample: Improve docstrings and add missing parameter to alias list

seisman:
Related to https://github.com/GenericMappingTools/pygmt/issues/4416.

### DIFF
--- a/pygmt/src/grdsample.py
+++ b/pygmt/src/grdsample.py
@@ -55,7 +55,7 @@ def grdsample(
     with ``interpolation``. Note that using ``region`` only is equivalent to
     :func:`pygmt.grdcut` or ``grdedit -S``. ``grdsample`` safely creates a fine mesh
     from a coarse one; the converse may suffer aliasing unless the data are filtered
-    :func:`pygmt.grdfilter` or ``grdfft``.
+    using :func:`pygmt.grdfilter` or ``grdfft``.
 
     Full GMT docs at :gmt-docs:`grdsample.html`.
 


pygmt/src/grdsample.py
```diff
@@ -30,34 +30,40 @@ def grdsample(
...
-    When ``region`` is omitted, the output grid will cover the same region as
-    the input grid. When ``spacing`` is omitted, the grid spacing of the
-    output grid will be the same as the input grid. Either ``registration`` or
-    ``toggle`` can be used to change the grid registration. When omitted,
-    the output grid will have the same registration as the input grid.
+    A bicubic [Default], bilinear, B-spline, or nearest-neighbor interpolation is set
+    with ``interpolation``. Note that using ``region`` only is equivalent to
+    :func:`pygmt.grdcut` or ``grdedit -S``. ``grdsample`` safely creates a fine mesh
+    from a coarse one; the converse may suffer aliasing unless the data are filtered
+    :func:`pygmt.grdfilter` or ``grdfft``.
```
yvonnefroehlich:
```suggestion
    A bicubic [Default], bilinear, B-spline, or nearest-neighbor interpolation is set
    with ``interpolation``. Note that using ``region`` only is equivalent to
    :func:`pygmt.grdcut` or ``grdedit -S``. ``grdsample`` safely creates a fine mesh
    from a coarse one; the converse may suffer aliasing unless the data are filtered
    using :func:`pygmt.grdfilter` or ``grdfft``.
```

PR #4428, opened 2026-02-25
pygmt.grdmask: Create mask grid from polygons or point coverage(required arguments)

Chuan1937:
pygmt.grdmask: Create mask grid from polygons or point coverage(required arguments)

**Preview**: https://pygmt-dev--4428.org.readthedocs.build/en/4428/api/generated/pygmt.grdmask.html#pygmt.grdmask


doc/api/index.rst
```diff
@@ -155,6 +155,7 @@ Operations on raster data
     grdhisteq.equalize_grid
     grdhisteq.compute_bins
     grdlandmask
+    grdmask
```
seisman: Although the module name starts with `grd`, it actually operates on tabular data and generates raster data. I feel it should be put in the "Operations on tabular data" category.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,110 @@
...
+    data,
+    outgrid: PathLike | None = None,
+    spacing: Sequence[float | str] | None = None,
+    region: Sequence[float | str] | str | None = None,
+    mask_values: Sequence[float] | None = None,
+    verbose: Literal["quiet", "error", "warning", "timing", "info", "compat", "debug"]
+    | bool = False,
+    **kwargs,
+) -> xr.DataArray | None:
+    r"""
```
seisman: `r` is not requred here.
```suggestion
    """
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,110 @@
...
+    ----------
+    data
+        Pass in either a file name, :class:`pandas.DataFrame`, :class:`numpy.ndarray`,
+        or a list of file names containing the polygon(s) or data points. Input can be:
+
+        - **Polygon mode**: One or more files containing closed polygon coordinates
+        - **Point coverage mode**: Data points (used with ``search_radius`` parameter)
+    $outgrid
+    $spacing
+    mask_values : list of float, optional
```
seisman:
```suggestion
    mask_values
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,110 @@
...
+          ``outgrid``)
+
+    Example
+    -------
+    >>> import pygmt
+    >>> import numpy as np
+    >>> # Create a simple polygon as a triangle
+    >>> polygon = np.array([[125, 30], [130, 30], [130, 35], [125, 30]])
+    >>> # Create a mask grid with 1 arc-degree spacing
+    >>> mask = pygmt.grdmask(data=polygon, spacing=1, region=[125, 130, 30, 35])
```
seisman, line 159: It will be useful to show the value of `mask`.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,110 @@
...
+    ----------
+    data
+        Pass in either a file name, :class:`pandas.DataFrame`, :class:`numpy.ndarray`,
+        or a list of file names containing the polygon(s) or data points. Input can be:
+
+        - **Polygon mode**: One or more files containing closed polygon coordinates
+        - **Point coverage mode**: Data points (used with ``search_radius`` parameter)
+    $outgrid
+    $spacing
+    mask_values
```
seisman: Both GMT and GMT.jl use `out_edge_in`, but I feel `mask_values` is readable and more consistent with modules like `grdlandmask` and `select`.

What's your plan about syntax like `-Nz`/`-NZ`/`-Np`/`-NP`?

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,110 @@
...
+    edge, or outside the polygons (or within the search radius from data points) are
+    assigned values based on ``mask_values``.
+
+    The mask grid can be used to mask out specific regions in other grids using
+    :func:`pygmt.grdmath` or similar tools. For masking based on coastline features,
+    consider using :func:`pygmt.grdlandmask` instead.
+
+    Full GMT docs at :gmt-docs:`grdmask.html`.
+
+    $aliases
```
seisman:
```suggestion
    **Aliases**
    
    .. hlist::
       :columns: 3
       
```

pygmt/src/grdmask.py
Chuan1937: ok

pygmt/src/grdmask.py
Chuan1937: The plan is to make the mask_values parameter support both lists and strings:
List format: mask_values=[0, 0, 1] (current usage),
String format: mask_values="Nz" / “NZ” / ‘Np’ / “NP” (corresponding to GMT's -Nz/-NZ/-Np/-NP),
What are your thoughts?

seisman: Looking at the GMT documentation, it seems the syntax is more complicated than that. The following are also valid: `-Nz/1`, `-Np5`, but syntax like `mask_values="z/1"`, or `mask_values="p5"` are still not Pythonic.

Chuan1937: Three new parameters—id_mode, id_start, and out_value—can be added to correspond to -Nz/Z/p/P and their variants, while being mutually exclusive with mask_values.

seisman: I'm thinking about `outside`, `edge`, `inside` rather than `mask_values`.

seisman: I see that you've already implemented the `outside`/`edge`/`inside` options. Could you please elaborate on the other related parameters? Just describe what you're planning to do, and there's no need to implement it yet. That way, we can review whether the proposed approach makes sense before moving forward.

seisman: > use -Nz to set polygon insides to the z-value obtained from the data (either segment header -Zzval, -Lheader or via -aZ=name); use -NZ to consider the polygon boundary as part of the inside. Alternatively, use -Np to use a running number as polygon ID; optionally append start of the sequence [0].

Actually, I don't fully understand the GMT documentation. As I understand it, `-Nz` and `-Np` control the value for polygon `inside`. `-Nz` sets the `inside` to be the z-value from the segment header, and `-Np` sets it to a polygon ID (starting from 0). Then `-NZ` and `-NP` have the same meaning but also force `edge` to be the same as `inside`. So, maybe we should allow syntax like `inside="z"` and `inside="id"`.

- `outside=1, edge=2, inside=3` -> `1/2/3`
- `inside="z"` -> `-Nz`
- `inside="id"` -> `-Np`
- `outside=1, inside="z"` -> `-Nz/1` [Can we set `edge` in this case?]
- `outside=1, inside="id"` -> `-Np/1` [Can we set `edge` in this case?]
- `outside=1, inside="z", edge="z"` -> `-NZ/1`
- `outside=1, inside="id", edge="id"` -> `-NP/1` 
- `outside=1, inside="z", edge="id"`: invalid

Chuan1937: `search_radius` (-S):

`search_radius`="50k" → -S50k (point coverage mode, 50km search radius)
`search_radius`="z" → -Sz (read radius from data column 3)
`search_radius`="1c" → -S1c (1 grid cell)

`clobber` (-C):

`clobber`="first" → -Cf (use the first overlapping polygon)
`clobber`="last" → -Co (use the last overlapping polygon)
`clobber`="low" → -Cl (use the polygon with the lowest z-value)
`clobber`="upper" → -Cu (use the polygon with the highest z-value)

`straight_line` (-A):

`straight_line`=True → -A (Use straight lines instead of great circles for geographic coordinates)
`straight_line`="m" → -Am (Meridians first, parallels second)
`straight_line`="p" → -Ap (Parallels first, meridians second)

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,172 @@
...
+           [0., 0., 0., 0., 1., 0.],
+           [0., 0., 1., 1., 1., 0.],
+           [0., 0., 1., 1., 1., 0.],
+           [0., 0., 1., 1., 1., 0.],
+           [0., 0., 0., 0., 0., 0.]])
+    """
+    if spacing is None or region is None:
+        raise GMTParameterError(required=["region", "spacing"])
+
+    # Build the -N parameter string
```
seisman: Please put the related codes into a private function (e.g., ~~`_build_N_option`~~ `_alias_option_N`). See `colorbar.py` for an example.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,213 @@
...
+    Returns
+    -------
+    Alias
+        An Alias object representing the -N option value.
+
+    Raises
+    ------
+    GMTParameterError
+        If inside and edge are both special modes but different values.
+
```
seisman: These parameters are already documented in the main function. No need to repeat them in the private funciton.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,213 @@
...
+    >>> _alias_option_N(inside="z", outside=1)._value
+    'z/1'
+    >>> _alias_option_N(inside="z", edge="z")._value
+    'Z'
+    >>> _alias_option_N(inside="id")._value
+    'p'
+    >>> _alias_option_N(inside="id", edge="id")._value
+    'P'
+    """
+    special_modes = {"z", "id"}
```
seisman: I feel the codes can be simplified if using a dict:
```suggestion
    special_modes = {"z": "z", "id": "p"}
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,213 @@
...
+            reason=msg + "When both are special modes, they must be the same."
+        )
+
+    # Build -N argument
+    if inside_is_special:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        if edge == inside:
+            mode_char = "Z" if inside == "z" else "P"
+        else:
+            mode_char = "z" if inside == "z" else "p"
```
seisman:
```suggestion
        mode_char = special_modes[inside]
        if edge == inside:
            mode_char = mode_char.upper()
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,193 @@
...
+    Builds the -N parameter string for grdmask based on the inside, edge, and
+    outside values. Handles special modes "z" (use z-value from polygon data)
+    and "id" (use running polygon ID).
+
+    Examples
+    --------
+    >>> _alias_option_N()._value
+    '0/0/1'
+    >>> _alias_option_N(outside=1, edge=1, inside=1)._value
+    '1/1/1'
```
seisman: So that we know the order:
```suggestion
    >>> _alias_option_N(outside=1, edge=2, inside=3)._value
    '1/2/3'
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,193 @@
...
+from pygmt.exceptions import GMTParameterError
+from pygmt.helpers import build_arg_list, fmt_docstring
+
+__doctest_skip__ = ["grdmask"]
+
+
+def _alias_option_N(  # noqa: N802
+    outside: float | Literal["z", "id"] = 0,
+    edge: float | Literal["z", "id"] = 0,
+    inside: float | Literal["z", "id"] = 1,
```
seisman: Although they default to 0,0,1, we usually set the default to `None`. 
```suggestion
    outside: float | Literal["z", "id"] | None = None,
    edge: float | Literal["z", "id"] | None = None,
    inside: float | Literal["z", "id"] | None = None,
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,193 @@
...
+    >>> _alias_option_N(inside="z", outside=1)._value
+    'z/1'
+    >>> _alias_option_N(inside="z", edge="z")._value
+    'Z'
+    >>> _alias_option_N(inside="id")._value
+    'p'
+    >>> _alias_option_N(inside="id", edge="id")._value
+    'P'
+    """
+    special_modes = {"z": "z", "id": "p"}
```
seisman: I feel the code logic can be further simplified, but I don't have time checking it yet.

Chuan1937: I simplified the code logic.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+            reason=msg + "When both are special modes, they must be the same."
+        )
+
+    # Build -N argument
+    if inside in {"z", "id"}:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode_char = "z" if inside == "z" else "p"
+        if edge == inside:
+            mode_char = mode_char.upper()
+        n_value = mode_char if outside == 0 else f"{mode_char}/{outside}"
```
seisman:
```suggestion
        mask_values = mode_char if outside is None else [mode_char, outside]
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+        return Alias(n_value, name="mask_values")
```
seisman:
```suggestion
        return Alias(mask_values, name="mask_values", sep="/", size=2)
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+) -> Alias:
+    """
+    Return an Alias object for the -N option.
+
+    Builds the -N parameter string for grdmask based on the inside, edge, and
+    outside values. Handles special modes "z" (use z-value from polygon data)
+    and "id" (use running polygon ID).
+
+    Examples
+    --------
```
seisman, line 31: Please check the output of `_alias_option_N()`

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+
+    aliasdict = AliasSystem(
+        I=Alias(spacing, name="spacing", sep="/", size=2),
+        N=_alias_option_N(outside=outside, edge=edge, inside=inside),
+    )
+
+    aliasdict = aliasdict.add_common(
+        R=region,
+        V=verbose,
+    )
```
seisman:
```suggestion
    ).add_common(
        R=region,
        V=verbose,
    )
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+        invalid and will raise an error.
+    inside
+        Set the value assigned to nodes inside the polygons. Default is 1.
+        Can be any number, or one of ``None``, ``"NaN"``, and ``np.nan`` for NaN.
+
+        Special values:
+
+        - ``"z"``: Use the z-value from polygon data (segment header ``-Zzval``,
+          ``-Lheader``, or via ``-aZ=name``). Corresponds to GMT ``-Nz``.
+        - ``"id"``: Use a running polygon ID number. Corresponds to GMT ``-Np``.
```
seisman:
```suggestion
    outside
    edge
    inside
        Set the value assigned to nodes outside, on the edge, or inside the polygons.
        Can be any number, or one of ``None``, ``"NaN"``, and ``np.nan`` for NaN.
        
        ``inside`` can also be set to one of the following values: 

        - ``"z"``: Use the z-value from polygon data (segment header ``-Zzval``,
          ``-Lheader``, or via ``-aZ=name``).
        - ``"id"``: Use a running polygon ID number.
        
        To treats edges as inside, using the same value as ``inside``. 
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,189 @@
...
+    # Build -N argument
+    if inside in {"z", "id"}:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode_char = "z" if inside == "z" else "p"
+        if edge == inside:
+            mode_char = mode_char.upper()
+        mask_values = mode_char if outside in {None, 0} else [mode_char, outside]
+        return Alias(mask_values, name="mask_values", sep="/", size=2)
+    # Standard mode: outside/edge/inside
+    return Alias([outside, edge, inside], name="mask_values", sep="/", size=3)
```
seisman:
```suggestion
    else:
        mask_values = [outside, edge, inside]
    return Alias([outside, edge, inside], name="mask_values", sep="/", size={2, 3})
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,177 @@
...
+            reason=msg + "When both are special modes, they must be the same."
+        )
+
+    # Build -N argument
+    if inside in {"z", "id"}:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode_char = "z" if inside == "z" else "p"
+        if edge == inside:
+            mode_char = mode_char.upper()
+        mask_values = mode_char if outside in {None, 0} else [mode_char, outside]
```
seisman: `outside` defaults to `None`, so if users explicitly set `outside=0`, we should pass it to GMT.
```suggestion
        mask_values = mode_char if outside is None else [mode_char, outside]
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,177 @@
...
+    'p'
+    >>> _alias_option_N(outside=0, edge="id", inside="id")._value
+    'P'
+    """
+    # Validate combinations
+    if inside in {"z", "id"} and edge in {"z", "id"} and inside != edge:
+        msg = f"Invalid combination: inside={inside!r} and edge={edge!r}. "
+        raise GMTParameterError(
+            reason=msg + "When both are special modes, they must be the same."
+        )
```
seisman:
```suggestion
    _inside_modes = {"z": "z", "id": "p"}
    # Validate combinations
    if inside in _inside_modes and edge in _inside_modes and inside != edge:
        msg = f"Invalid combination: inside={inside!r} and edge={edge!r}. "
        raise GMTParameterError(
            reason=msg + "When both are special modes, they must be the same."
        )
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # Build -N argument
+    if inside in {"z", "id"}:
```
seisman:
```suggestion
    if inside in _inside_modes:
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,177 @@
...
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode_char = "z" if inside == "z" else "p"
```
seisman:
```suggestion
        mode_char = _inside_modes[inside]
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+    Return an Alias object for the -N option.
+
+    Builds the -N parameter string for grdmask based on the inside, edge, and
+    outside values. Handles special modes "z" (use z-value from polygon data)
+    and "id" (use running polygon ID).
+
+    Examples
+    --------
+    >>> _alias_option_N(outside=0, edge=0, inside=1)._value
+    '0/0/1'
```
seisman: `_alias_option_N()` should return `None`.
```suggestion
    >>> _alias_option_N()._value
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+    >>> _alias_option_N(outside=0, edge=0, inside="z")._value
+    'z'
+    >>> _alias_option_N(outside=1, edge=0, inside="z")._value
+    'z/1'
+    >>> _alias_option_N(outside=0, edge="z", inside="z")._value
+    'Z'
+    >>> _alias_option_N(outside=0, edge=0, inside="id")._value
+    'p'
+    >>> _alias_option_N(outside=0, edge="id", inside="id")._value
+    'P'
```
seisman:
```suggestion
    >>> _alias_option_N(inside="z")._value
    'z'
    >>> _alias_option_N(outside=1, inside="z")._value
    'z/1'
    >>> _alias_option_N(edge="z", inside="z")._value
    'Z'
    >>> _alias_option_N(inside="id")._value
    'p'
    >>> _alias_option_N(edge="id", inside="id")._value
    'P'
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+            reason=msg + "When both are special modes, they must be the same."
+        )
+
+    # Build -N argument
+    if inside in _inside_modes:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode_char = _inside_modes[inside]
+        if edge == inside:
+            mode_char = mode_char.upper()
+        mask_values = mode_char if outside is None else [mode_char, outside]
```
seisman:
```suggestion
        # Mode: -Nz, -NZ, -Np, or -NP
        mode = _inside_modes[inside]
        if edge == inside:
            mode = mode.upper()
        mask_values = mode if outside is None else [mode, outside]
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+    edge: float | Literal["z", "id"] = 0,
+    inside: float | Literal["z", "id"] = 1,
+    verbose: Literal["quiet", "error", "warning", "timing", "info", "compat", "debug"]
+    | bool = False,
+    **kwargs,
+) -> xr.DataArray | None:
+    """
+    Create mask grid from polygons or point coverage.
+
+    Reads one or more files (or standard input) containing polygon or data point
```
seisman: There is no standard in in Python.
```suggestion
    Reads one or more files containing polygon or data point
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+    coordinates, and creates a binary grid file where nodes that fall inside, on the
```
seisman:
```suggestion
    coordinates, and creates a grid where nodes that fall inside, on the
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,178 @@
...
+       - I = spacing
+       - N = outside/edge/inside
+       - R = region
+       - V = verbose
+
+    Parameters
+    ----------
+    data
+        Pass in either a file name, :class:`pandas.DataFrame`, :class:`numpy.ndarray`,
+        or a list of file names containing the polygon(s) or data points. Input can be:
```
seisman:
```suggestion
        Pass in either a file name to an ASCII data table, a 2-D $table_classes 
        containg the polygon(s) or data points. Input can be:
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,182 @@
...
+from pygmt.alias import Alias, AliasSystem
+from pygmt.clib import Session
+from pygmt.exceptions import GMTParameterError
+from pygmt.helpers import build_arg_list, fmt_docstring
+
+__doctest_skip__ = ["grdmask"]
+
+
+def _alias_option_N(  # noqa: N802
+    outside: float | Literal["z", "id"] | None = None,
```
seisman: `outside` can't be `z` or `id`:
```suggestion
    outside: float None = None,
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,182 @@
...
+
+@fmt_docstring
+def grdmask(
+    data,
+    outgrid: PathLike | None = None,
+    spacing: Sequence[float | str] | None = None,
+    region: Sequence[float | str] | str | None = None,
+    outside: float | Literal["z", "id"] = 0,
+    edge: float | Literal["z", "id"] = 0,
+    inside: float | Literal["z", "id"] = 1,
```
seisman:
```suggestion
    outside: float | None = None, 
    edge: float | Literal["z", "id"] | None = None,
    inside: float | Literal["z", "id"] | None = None,
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    Return an Alias object for the -N option.
+
+    Builds the -N parameter string for grdmask based on the inside, edge, and
+    outside values. Handles special modes "z" (use z-value from polygon data)
+    and "id" (use running polygon ID).
+
+    Examples
+    --------
+    >>> _alias_option_N()._value
```
seisman, line 32: What if users pass `outside=1` only?

Chuan1937: `outside=1` alone would generate -N1/None/None which passes invalid None strings to GMT. Fixed by returning None when inside is not set, so GMT uses its default values.

seisman: I think your changes don't make sense. `outside=3` should return `-N3/0/1` rather than `None`.

Chuan1937: I've added a more comprehensive test

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,193 @@
...
+    | bool = False,
+    **kwargs,
+) -> xr.DataArray | None:
+    """
+    Create mask grid from polygons or point coverage.
+
+    Reads one or more files containing polygon or data point
+    coordinates, and creates a grid where nodes that fall inside, on the
+    edge, or outside the polygons (or within the search radius from data points) are
+    assigned values based on ``outside``, ``edge``, and ``inside`` parameters.
```
seisman: Please rewrap these lines to 88 characters.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,193 @@
...
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode = _inside_modes[inside]  # type: ignore[index]
+        if edge == inside:
+            mode = mode.upper()
+        mask_values: str | list[str | float] | None = (
+            mode if outside is None else [mode, outside]  # type: ignore[assignment]
+        )
+    elif inside is None and outside is None and edge is None:
+        # All three are None, return None (GMT uses default 0/0/1)
+        mask_values = None
```
seisman: Change it to something like
```
if all(v is None for v in (outside, inside, edge)):
    return Alias(None, name="mask_values")
```
and move it to the start of the function.

Chuan1937:
image

`make format` will delete `# noqa: PERF102` locally.But CI test will fail.

seisman:
> image
> `make format` will delete `# noqa: PERF102` locally.But CI test will fail.

You need to update your `ruff` format.

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,194 @@
...
+
+    # Build -N argument
+    if inside in _inside_modes:
+        # Mode: -Nz, -NZ, -Np, or -NP
+        mode = _inside_modes[inside]  # type: ignore[index]
+        if edge == inside:
+            mode = mode.upper()
+        mask_values: str | list[str | float] | None = (
+            mode if outside is None else [mode, outside]  # type: ignore[assignment]
+        )
```
seisman: Now it can be simplified.
```suggestion
        mask_values = mode if outside is None else [mode, outside]
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,194 @@
...
+    >>> mask = pygmt.grdmask(data=polygon, spacing=1, region=[125, 130, 30, 35])
+    >>> mask.values
+    array([[0., 0., 0., 0., 0., 0.],
+            [0., 0., 1., 1., 1., 0.],
+            [0., 0., 0., 1., 1., 0.],
+            [0., 0., 0., 0., 1., 0.],
+            [0., 0., 0., 0., 0., 0.],
+           [0., 0., 0., 0., 0., 0.]])
+    """
+    if spacing is None or region is None:
```
seisman:
```suggestion
    if kwargs.get("I", spacing) is None or kwargs.get("R", region) is None:
```

pygmt/src/grdmask.py
```diff
@@ -0,0 +1,192 @@
...
+    # Validate combinations
+    if edge in _inside_modes and inside != edge:
+        msg = f"Invalid combination: edge={edge!r} requires inside={edge!r}."
+        raise GMTParameterError(reason=msg)
+
+    if inside in _inside_modes and edge in _inside_modes and inside != edge:
+        msg = f"Invalid combination: inside={inside!r} and edge={edge!r}. "
+        raise GMTParameterError(
+            reason=msg + "When both are special modes, they must be the same."
+        )
```
seisman: I think it makes more sense to raise GMTValueError.

Chuan1937: ok

seisman:
@Chuan1937 This PR was closed since you deleted your own fork. Is it intentional?

Chuan1937:
> @Chuan1937 This PR was closed since you deleted your own fork. Is it intentional?

No. This should be the automated behavior of the PR.

PR #4464, opened 2026-03-21
Add a helper function to check if a parameter is given or not

### LINKED ISSUE #4355: Add a helper function to check if a parameter is given or not
## Motivation

In PyGMT, most parameters default to `False` (for boolean-only parameters) or `None` (for non-boolean parameters), which means the parameters are not given and should not be used in building the CLI option string. We usually need to check if a parameter is given, like below
```python
if value is not False:  # For boolean-only parameters
   ...

if value is not None:  # For non-boolean parameters
   ...

if value is not None and value is not False:  # A more general check
   ...
```

This repetitive pattern:

- Reduces code readability
- Makes maintenance harder [Sometimes we may incorrectly use `if value is not None` when `value` is a boolean type]
- Makes the intent less clear to new contributors

## Solution

Introduced a new `is_given()` helper function to simplify parameter validation across PyGMT and refactored existing code to use it consistently.
```python
def is_given(value: Any) -> bool:
    return value is not None and value is not False
```
Then we should 

- replace `x is not None`/`x is not False`/`x is not None and x is not False` with `is_given(x)`.
- replace `x is None or x is False` to `not is_given(x)`

This is just an internal change and doesn't break any existing behavior.

Comments?

Chuan1937:
Replace #4439
close #4355


seisman:
@Chuan1937 I wonder if you would like to continue with this PR?

Chuan1937:
> @Chuan1937 I wonder if you would like to continue with this PR?

of course

seisman (approved):
Looks good to me.

PR #4582, opened 2026-04-13
Migrate frame settings in pygmt/tests/test_config.py

seisman:
## Summary
- migrate list-style frame settings in pygmt/tests/test_config.py to Frame/Axis
- preserve the mixed-case axis coverage and primary/secondary time-axis settings in the config tests


weiji14 (approved):
Looks ok, but...

> * python -m py_compile pygmt/tests/test_config.py

Why [`py_compile`](https://docs.python.org/3/library/py_compile.html)?

seisman:
This PR was created by codex. Sometimes it may write codes or descriptions that make no sense.

PR #4585, opened 2026-04-14
Migrate the frame setting in the first figure tutorial

seisman:
## Summary
- migrate examples/intro/01_first_figure.py to use Axis and Frame for frame examples
- update the tutorial text to describe the structured frame API instead of legacy frame strings and lists

**Preview**: https://pygmt-dev--4585.org.readthedocs.build/en/4585/intro/01_first_figure.html

### DIFF
--- a/examples/intro/01_first_figure.py
+++ b/examples/intro/01_first_figure.py
@@ -16,7 +16,7 @@ address other PyGMT methods.
 # are accessible from the :mod:`pygmt` top level package.
 
 import pygmt
-from pygmt.params import Axis, Frame
+from pygmt.params import Frame
 
 # %%
 # Creating a figure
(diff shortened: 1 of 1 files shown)


examples/intro/01_first_figure.py
```diff
@@ -128,7 +129,7 @@
     land="lightgreen",
     water="lightblue",
     projection="M10c",
-    frame="a",
+    frame=True,
```
seisman, line 132: I feel using `frame=True` in the first example makes more sense than the much complicated `frame=Axis(annot=True)`.

examples/intro/01_first_figure.py
```diff
@@ -152,7 +148,7 @@
     land="lightgreen",
     water="lightblue",
     projection="M10c",
-    frame=["a", "+tMaine"],
+    frame=Frame(title="Maine"),
```
seisman, line 151: Same here, make the `frame` setting as simple as possible.

yvonnefroehlich:
I am sorry for being too late with reviewing this OR. But it looks good!

PR #4587, opened 2026-04-14
DOC/ gallery example "Vector heads and tails": Add note on GMT 4 vector syntax

yvonnefroehlich:

The definition of vectors has changed between GMT 4 and 5. The GMT 4 syntax is still supported. Having an entirely filled vector can be useful. I think (at least) adding a note to the gallery example "Vector heads and tails" that this can be achieved with the GMT 4 syntax can be helpful for users. 

**Preview**:

**Guidelines**

**Slash Commands**

### DIFF
--- a/examples/gallery/lines/vector_heads_tails.py
+++ b/examples/gallery/lines/vector_heads_tails.py
@@ -2,29 +2,28 @@ r"""
 Vector heads and tails
 ======================
 
-Many methods in PyGMT allow plotting vectors with individual
-heads and tails. For this purpose, several modifiers may be appended to
-the corresponding vector-producing parameters for specifying the placement
-of vector heads and tails, their shapes, and the justification of the vector.
+Many methods in PyGMT allow plotting vectors with individual heads and tails.
+For this purpose, several modifiers may be appended to the corresponding
+vector-producing parameters for specifying the placement of vector heads and
+tails, their shapes, and the justification of the vector.
 
-To place a vector head at the beginning of the vector path
-simply append **+b** to the vector-producing option (use **+e** to place
-one at the end). Optionally, append **t** for a terminal line, **c** for a
-circle, **a** for arrow (default), **i** for tail, **A** for plain open
-arrow, and **I** for plain open tail. Further append **l** or **r** (e.g.
-``"+bar"``) to only draw the left or right half-sides of the selected head/tail
-(default is both sides) or use **+l** or **+r** to apply simultaneously to both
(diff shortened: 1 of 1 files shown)


examples/gallery/lines/vector_heads_tails.py
```diff
@@ -27,6 +27,11 @@
 the vector type, here ``"v"`` for Cartesian vector, see also the
 :doc:`Vector types example </gallery/lines/vector_styles>`)
 by passing the corresponding shortcuts to the ``style`` parameter.
+
+Note: Vectors were completely redesigned for GMT 5 which separated the vector
+head (a polygon) from the vector stem (a line). In GMT 4, the entire vector was
+a polygon. Thus, to fill the entire vector with a color users have to follow the
+GMT 4 synthax explained at :gmt-docs:`plot.html#id13`.
```
seisman: The `id13` anchor my change in the future, so it's not ideal to reference it here. We already have this sentence in this example:

> For further modifiers see https://docs.generic-mapping-tools.org/6.6/plot.html#vector-attributes.

I think we can use that link instead.

yvonnefroehlich: I reformulated the text in commit https://github.com/GenericMappingTools/pygmt/pull/4587/commits/7e709f897a0b9240a5420e9df7759e22f9d1a2bd to only use the link to the vector attribut section.

examples/gallery/lines/vector_heads_tails.py
```diff
@@ -2,31 +2,34 @@
...
-:doc:`Vector types example </gallery/lines/vector_styles>`)
-by passing the corresponding shortcuts to the ``style`` parameter.
+:doc:`Vector types example </gallery/lines/vector_styles>`) by passing the
+corresponding shortcuts to the ``style`` parameter.
+
+For further modifiers see :gmt-docs:`plot.html#vector-attributes`. Please note
+that vectors were completely redesigned for GMT 5 which separated the vector
+head (a polygon) from the vector stem (a line). In GMT 4, the entire vector was
+a polygon. Thus, to fill the entire vector with a color users have to follow the
+GMT 4 synthax explained at :gmt-docs:`plot.html#id13`.
```
seisman:
```suggestion
GMT 4 synthax explained in the above link.
```

yvonnefroehlich: Oh, no I am sorry, I forgot to save the file before committing /:
Now, the link should be removed.

Issue #4588, opened 2026-04-15
Frame/Axis classes for frame settings to wrap GMT's -B option


seisman:
## Summary

This issue tracks the implementation of `Frame` and `Axis` classes in PyGMT as a structured wrapper around GMT’s `-B` option. The goal is to provide a Pythonic API for frame settings while preserving compatibility with existing `frame` usage.

## Documentation

- GMT `-B`: https://docs.generic-mapping-tools.org/dev/gmt.html#b-full
- PyGMT `Frame`: https://www.pygmt.org/dev/api/generated/pygmt.params.Frame.html
- PyGMT `Axis`: https://www.pygmt.org/dev/api/generated/pygmt.params.Axis.html

## Motivation

GMT’s `-B` syntax is powerful but hard to read and discover in Python code. For example:

```python
frame=["WStr+tTitle", "xa2f1g2+lxlabel", "ya5f1g5+lylabel"]
```

A structured API would improve readability, discoverability, validation, testing, and documentation. We decide to implement two classes, Frame and Axis, to control the frame and axis settings. An example usage is like below:
```python
from pygmt.params import Axis, Frame

frame = Frame(
    axes="WStr",
    title="Title",
    xaxis=Axis(annot=2, tick=1, grid=2, label="xlabel"),
    yaxis=Axis(annot=5, tick=1, grid=5, label="ylabel"),
)
```

### `Axis`

The GMT CLI syntax is 
```
-B[p|s][x|y|z]intervals[+aangle|n|p][+e[l|u]][+f][+l|Llabel][+pprefix][+s|Sseclabel][+uunit]
```
in which *intervals* is in the form of `[a|f|g][stride][phase][unit]`.

- [x] `a`: `annot`, set annotation interval #4406
- [x] `f`: `tick`, set tick interval #4406
- [x] `g`: `grid`, set gridline interval #4406
- [x] `+p`: `prefix`, set annotation prefix text #4406
- [x] `+u`: `unit`, append annotation unit text #4406
- [x] `+a`: `angle`, set annotation angle #4406
  - [x] `+aangle` #4406 
  - [ ] `+an|p`
- [ ] `+e`
- [ ] `+f` 
- [x] `+l`: `label`, set axis label text #4406
- [x] `+L`: `hlabel` set horizontal label for y-axis #4589
- [x] `+s`: `alt_label`: set alternative label for upper or right axis #4589
- [x] `+S`: `alt_label`: set horizontal label for right axis #4589
[...]

PR #4591, opened 2026-04-16
params.Frame: Add subtitle support

seisman:
Address https://github.com/GenericMappingTools/pygmt/issues/4588

Example
```python
import pygmt

fig = pygmt.Figure()
fig.basemap(
    region=[0, 20, 0, 3], 
    projection="M10c", 
    frame=Frame(
        axes="WSEN", 
        title="Title", 
        subtitle="SubTitle", 
        axis=Axis(annot=True)
    )
)
fig.show()
```

### DIFF
--- a/pygmt/params/frame.py
+++ b/pygmt/params/frame.py
@@ -222,6 +222,11 @@ class Frame(BaseParam):
         """
         Validate the parameters of the Frame class.
         """
+        if self.subtitle is not None and self.title is None:
+            raise GMTParameterError(
+                required="title",
+                reason="The 'subtitle' parameter requires 'title' to be set.",
+            )
         if self.axis is not None and any(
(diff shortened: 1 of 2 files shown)


pygmt/params/frame.py
```diff
@@ -187,6 +189,10 @@ class Frame(BaseParam):
     #: The title string centered above the plot frame [Default is no title].
     title: str | None = None
 
+    #: The subtitle string centered below the plot title [Requires ``title``. Default is
+    #: no subtitle].
```
yvonnefroehlich:
```suggestion
    #: The subtitle string centered below the plot title. Requires ``title`` to be set.
    #: [Default is no subtitle].
```

yvonnefroehlich (approved):

PR #4594, opened 2026-04-19
params.Frame: Add parameters for 3-D plots

seisman:
This PR adds parameters for -B modifiers related to 3-D plots:

- `+b`: `box`
- `+w`: `wall_pen`
- `+x`: `yzfill`
- `+y`: `xzfill`
- `+z`: `xyfill`

GMT uses `pen` for `+b` (xref: https://github.com/GenericMappingTools/gmt/blob/ac580b28da697232ae21cfa2ec102610572af2d8/src/gmt_common_longoptions.h#L72), but I feel `pen` is too general, and it's also a little misleading because it only sets pen for x-z and y-z planes, not the frame pen for the x-y plane.

Example
```
import pygmt

fig = pygmt.Figure()
fig.basemap(
    region=(0, 1, 0, 2, 0, 3),
    projection="X10c/10c",
    zsize="5c",
    perspective=(120, 30),
    frame=Frame(
        axes="WSENZ",
        axis=True,
        zaxis=True,
        box=True,
        wall_pen="1p,red",
        xzfill="lightgreen",
        xyfill="lightblue",
        yzfill="lightyellow",
    )
)
fig.show()
```
map

Related to #4588

### DIFF
--- a/pygmt/tests/test_params_frame.py
+++ b/pygmt/tests/test_params_frame.py
@@ -58,7 +58,7 @@ def test_params_frame_only():
         xzfill="green",
         xyfill="blue",
     )
-    assert str(frame) == "WSrtZ+b+ggray+tMy Title+w0.5p,black+xred+ygreen+zblue"
+    assert str(frame) == "WSrtZ+ggray+tMy Title+b+w0.5p,black+xred+ygreen+zblue"
 
 
 def test_params_frame_axis():
--- a/pygmt/tests/test_params_frame.py
+++ b/pygmt/tests/test_params_frame.py
@@ -52,17 +52,13 @@ def test_params_frame_only():
         axes="WSrtZ",
         box=True,
         title="My Title",
-        subtitle="My Subtitle",
         fill="gray",
         wall_pen="0.5p,black",
         yzfill="red",
         xzfill="green",
         xyfill="blue",
     )
-    assert (
-        str(frame)
-        == "WSrtZ+b+ggray+tMy Title+sMy Subtitle+w0.5p,black+xred+ygreen+zblue"
-    )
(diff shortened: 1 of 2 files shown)


michaelgrund (approved):

pygmt/tests/test_params_frame.py
```diff
@@ -42,6 +42,24 @@ def test_params_frame_only():
...
+
+    frame = Frame(
+        axes="WSrtZ",
+        box=True,
+        title="My Title",
+        fill="gray",
+        wall_pen="0.5p,black",
+        yzfill="red",
+        xzfill="green",
+        xyfill="blue",
```
yvonnefroehlich, line 59: I believe if all three of `yzfill`, `xzfill` and `xyfill` are used, `fill` will not have an effect.

seisman: Yes, `fill` sets the fill of the three back-faces but can be overridden by `yzfill`/`xzfill`/`xyfill`. We need to improve the docstring to clarify it.

Actually, as a non-native English speaker, I don't know the standard terminology to describe the elements of a cube. Currently, we describe them as "frontground lines", "outlines of x-z and y-z planes", "yz-plane"/"xz-plane"/"xy-plane", but it seems in standard terminology, they should be called faces (front face? back face?), and edges.

yvonnefroehlich:
> GMT uses pen for +b (xref: https://github.com/GenericMappingTools/gmt/blob/ac580b28da697232ae21cfa2ec102610572af2d8/src/gmt_common_longoptions.h#L72), but I feel pen is too general, and it's also a little misleading because it only sets pen for x-z and y-z planes, not the frame pen for the x-y plane.

I like `box` for **+b**.

Issue #3981, opened 2025-06-18
Combine bool and string in a list


Esteban82:
### Description of the problem

This does not work
`fig.basemap(frame=[True, "+tThe Caribbean Sea"])`

From the forum: [link]( https://forum.generic-mapping-tools.org/t/combine-bool-and-string-in-a-list-for-pymgt/6005/1)

### Minimal Complete Verifiable Example

```python
import pygmt
fig = pygmt.Figure()
fig.coast(region="IHO27",projection="M15c",water="lightblue")
fig.basemap(frame=[True, "+tThe Caribbean Sea"])
fig.savefig("1_FirstMap.png")
```

### Full error message

```python-traceback
basemap [ERROR]: Bad interval in -B option (x-component, a-info): True gave interval = 0
basemap [ERROR]: Bad interval in -B option (y-component, a-info): True gave interval = 0
basemap [ERROR]: Option -B parsing failure. Correct syntax:

-B Specify both (1) basemap frame settings and (2) axes parameters.
Frame settings are modified via an optional single invocation of -B[<axes>][+b][+g<fill>][+i[<val>]][+n][+o<lon>/<lat>][+s<subtitle>][+t<title>][+w[<pen>]][+x<fill>]⏎
       …[+y<fill>][+z<fill>]
Axes parameters are specified via one or more invocations of -B[p|s][x|y|z]<intervals>[+a<angle>|n|p][+e[l|u]][+f][+l|L<label>][+p<prefix>][+s|S<secondary_label>]⏎
       …[+u<unit>
<intervals> is composed of concatenated [<type>]<stride>[l|p] sub-strings. See basemap documentation for more details and examples of all settings.
basemap [ERROR]: Offending option -BTrue
Traceback (most recent call last):
  File "/home/esteban82/Github/Esteban82/Privado/4_EarthScope_GMT-ShortCourse/1_Modules/6_GMT_Advance/1_Scripts/3_PyGMT/1_FirstMap.py", line 32, in <module>
    fig.basemap(frame=[True, "+tThe Caribbean Sea"])
  File "/home/esteban82/miniforge3/lib/python3.10/site-packages/pygmt/helpers/decorators.py", line 609, in new_module
    return module_func(*args, **kwargs)
  File "/home/esteban82/miniforge3/lib/python3.10/site-packages/pygmt/helpers/decorators.py", line 773, in new_module
    return module_func(*bound.args, **bound.kwargs)
[...]

PR #4595, opened 2026-04-19
DOC/Test "test_basemap": Migrate frame settings to the new Frame/Axis syntax 

Chuan1937:
Update the test pygmt/tests/test_basemap.py to use the new implemented Frame and Axis parameter classes.

Fixes partly https://github.com/GenericMappingTools/pygmt/issues/4494

### DIFF
--- a/pygmt/tests/test_basemap.py
+++ b/pygmt/tests/test_basemap.py
@@ -180,7 +180,11 @@ def test_basemap_frame_sequence_true():
     Test for https://github.com/GenericMappingTools/pygmt/issues/3981.
     """
     fig = Figure()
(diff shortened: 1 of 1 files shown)


pygmt/tests/test_basemap.py
```diff
@@ -180,7 +180,11 @@ def test_basemap_frame_sequence_true():
     Test for https://github.com/GenericMappingTools/pygmt/issues/3981.
```
seisman, line 180: Perhaps we should keep this test unchanged, because it's a regression test for a specific usage reported in #3981.

Chuan1937: ok

PR #4601, opened 2026-04-26
Update the frame type hints to support Axis and Frame

seisman:
Old type hint: `frame: str | Sequence[str] | Literal["none"] | bool = False,`

New type hint: `frame: Frame | Axis | Literal["none"] | bool = False`

`ste | Sequence[str]` is still supported, but not type hinted.

### DIFF
--- a/pygmt/src/basemap.py
+++ b/pygmt/src/basemap.py
@@ -20,7 +20,7 @@ def basemap(  # noqa: PLR0913
     zscale: float | str | None = None,
     zsize: float | str | None = None,
     region: Sequence[float | str] | str | None = None,
-    frame: Frame | Axis | Literal["none"] | bool = False,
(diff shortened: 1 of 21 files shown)


weiji14:
> `str | Sequence[str]` is still supported, but not type hinted.

This might break static type checkers (e.g. mypy/ty/pyright) for some downstream users? Maybe keep for now?

seisman:
> > `str | Sequence[str]` is still supported, but not type hinted.
> 
> This might break static type checkers (e.g. mypy/ty/pyright) for some downstream users? Maybe keep for now?

Done in https://github.com/GenericMappingTools/pygmt/pull/4601/changes/cf06ec5dd156a30a5253e6a070640a9d0cf6e13a

PR #3849, opened 2025-03-13
Figure.pygmtlogo: Initial version for the circular, colored PyGMT logo

yvonnefroehlich:
**Description of proposed changes**

Related to https://github.com/GenericMappingTools/pygmt/issues/1404#issuecomment-2718559910, create the PyGMT logo in Python with PyGMT:

- [x] Python script
- [x] Separate method `Figure.pygmtlogo` or `Figure.logo_pygmt` as part of the `Figure` class _or_
       addition to the method `Figure.logo` (which plots the GMT logo)
- [ ] Documentation
- [x] Tests -> only circular, color, light and dark themes
- [ ] ~Gallery example~ -> separate PR

**Related to**

- #1404
- #3848
- #3845

**Preview**:
https://pygmt-dev--3849.org.readthedocs.build/en/3849/api/generated/pygmt.Figure.pygmtlogo.html

|  |  |  |  |
| --- | --- | --- | --- |
| image | image | image | image |
| image | image | image | image |
| image | image | image | image |

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #4612, opened 2026-04-29
Figure: Refactor how plotting methods are attached to the Figure class

seisman:
## Motivation

In PR #3849, we're seeing many static type check errors (https://github.com/GenericMappingTools/pygmt/actions/runs/25118862756/job/73613589767?pr=3849), mainly because in `pygmtlogo`, we're trying to re-import the `Figure` class, which causes some circular imports. 

Codex recommends a different way to attach the plotting wrappers to the `Figure` class, which can solve the static type error found in PR #3849. Changes in this PR are made by Codex.

## Summary
This PR refactors how plotting methods are attached to `pygmt.Figure` and clarifies the role of `pygmt.src`.

## Why
Previously, `Figure` methods were attached through a class-scoped `from pygmt.src import (...)` block in `pygmt/figure.py`. That worked at runtime, but it had a few drawbacks:

- It depended on package-level re-exports from `pygmt/src/__init__.py` just to populate `Figure`.
- Static analysis could misinterpret some imported names as modules instead of callables. (see the mypy errors in found https://github.com/GenericMappingTools/pygmt/actions/runs/25118862756/job/73613589767?pr=3849)
- The class-scoped import block was harder to read and maintain than explicit assignments.
- It blurred two separate concepts:
  - standalone functions that are part of the functional API
  - plotting functions that are intended to be used as methods of `Figure`

### `pygmt/figure.py`
- Import each plotting callable directly from its concrete `pygmt.src.*` module.
- Attach these callables explicitly on the `Figure` class body, e.g. `plot = _plot`.
- Add a short comment above the attachment block to clarify the pattern.

### `pygmt/src/__init__.py`
- Keep only standalone functions that are intended to be called directly.
- Stop re-exporting `Figure`-bound plotting methods from `pygmt.src`.
- Add a short comment explaining the distinction between standalone functions and `Figure` methods.

## Why this is better
This makes the API structure clearer:

- `pygmt.src` becomes a cleaner namespace for standalone operations.
- `Figure` is the clear home for plotting methods.

It also improves maintainability:

- Each attached `Figure` method points to one explicit source module.
- The attachment logic is easier to scan and modify.
- `Figure` no longer depends on package-level re-export behavior for method attachment.
- The structure is friendlier to static analysis and code review.

## Behavioral intent
This is intended as an internal API-structure cleanup, not a plotting behavior change. The plotting implementations themselves are unchanged; only the way they are attached/exported has been reorganized.

### DIFF
--- a/pygmt/figure.py
+++ b/pygmt/figure.py
@@ -136,42 +136,6 @@ class Figure:
         self._preview_dir = TemporaryDirectory(prefix=f"{self._name}-preview-")
         self._activate_figure()
 
-        # Attach plotting functions implemented in pygmt/src as bound instance methods.
-        self.basemap = _basemap
-        self.choropleth = _choropleth
-        self.coast = _coast
-        self.colorbar = _colorbar
-        self.contour = _contour
-        self.directional_rose = _directional_rose
-        self.grdcontour = _grdcontour
-        self.grdimage = _grdimage
-        self.grdview = _grdview
-        self.histogram = _histogram
-        self.hlines = _hlines
-        self.image = _image
-        self.inset = _inset
-        self.legend = _legend
-        self.logo = _logo
-        self.magnetic_rose = _magnetic_rose
-        self.meca = _meca
-        self.paragraph = _paragraph
-        self.plot = _plot
-        self.plot3d = _plot3d
-        self.psconvert = _psconvert
-        self.rose = _rose
-        self.scalebar = _scalebar
-        self.set_panel = _set_panel
-        self.shift_origin = _shift_origin
-        self.solar = _solar
-        self.subplot = _subplot
-        self.ternary = _ternary
-        self.text = _text
-        self.tilemap = _tilemap
-        self.timestamp = _timestamp
-        self.velo = _velo
-        self.vlines = _vlines
-        self.wiggle = _wiggle
-
     def __del__(self) -> None:
         """
         Clean up the temporary directory that stores the previews.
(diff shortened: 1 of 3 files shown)


seisman:
This PR is backward-compatible, but it's a big low-level refactor. Ping @GenericMappingTools/pygmt-maintainers for reviews.

Edit: Previously, it was possible to do `from pygmt.src import plot`, but it makes no sense to call `plot` as as standalone function. After this PR, `from pygmt.src import plot` no longer works.

seisman:
This PR also renames the `text_` method to `text` in commit d74da28a0ced61c910891d953ca864ba871eedaf. The name `text_` was initially introduced to suppress a warning mentioned in https://github.com/GenericMappingTools/pygmt/pull/808#issuecomment-769418112.

pygmt/src/__init__.py
```diff
@@ -2,71 +2,40 @@
 Source code for PyGMT methods.
 """
 
-from pygmt.src.basemap import basemap
+# Re-export standalone functions that can be used directly.
+# Figure plotting methods are attached in pygmt.figure and are not exported here.
```
yvonnefroehlich:
```suggestion
# Figure plotting methods are attached in pygmt/figure.py and are not exported here.
```

pygmt/figure.py
```diff
@@ -408,42 +442,41 @@ def _repr_html_(self) -> str:
...
-        ternary,
-        text,
-        tilemap,
-        timestamp,
-        velo,
-        vlines,
-        wiggle,
-    )
+    # Attach plotting functions implemented in pygmt.src as Figure methods.
+    basemap = _basemap
```
weiji14, line 446: Couldn't we just directly do

```suggestion
    from pygmt.src.basemap import basemap
```

So that the `_basemap` is not needed?

weiji14: Ah ok, if we did this, we would still need to handle:

```
error: Unsupported class scoped import  [misc]
PLC0415 `import` should be at the top-level of a file
```

weiji14: Can these go into the `__init__` method? Might need to rewrite as `self.basemap = _basemap`?

seisman: Done in c5035c9

weiji14: Oo, doesn't actually work. Getting `TypeError("coast() missing 1 required positional argument: 'self'")` :sweat_smile:

weiji14: Maybe revert that commit. We actually do want the figure methods as class variables, not instance variables (https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables).

weiji14 (approved):
Have some ideas using `importlib` or Mixins (following ideas in https://stackoverflow.com/questions/47561840/how-can-i-separate-the-functions-of-a-class-into-multiple-files), but the diff might be even bigger, so will just settle with this for now. Thanks @seisman.

pygmt/figure.py
```diff
@@ -408,42 +442,41 @@ def _repr_html_(self) -> str:
...
-        subplot,
-        ternary,
-        text,
-        tilemap,
-        timestamp,
-        velo,
-        vlines,
-        wiggle,
-    )
+    # Attach plotting functions implemented in pygmt.src as Figure methods.
```
seisman:
```suggestion
    # Attach plotting functions implemented in pygmt/src as Figure methods.
```

PR #4616, opened 2026-05-02
Implement Figure.pygmtlogo for the PyGMT logo

seisman:
This PR tracks the implementation of the PyGMT logos in different versions.

## Logo implementation status

| Variant | Colored + Light | Colored + Dark | Grayscale + Light | Grayscale + Dark |
|--------|------------------|-------------------|----------------|------------------|
| Circular, no wordmark | ✅ | ✅  | ✅ | ✅ |
| Circular, horizontal wordmark |✅ |✅ |✅ |✅ |
| Circular, vertical wordmark |✅ |✅ |✅ |✅ |
| Hexagon, no wordmark |✅ |✅ |✅ |✅ |
| Hexagon, horizontal wordmark |✅ | ✅| ✅| ✅|
| Hexagon, vertical wordmark |✅ |✅ |✅ | ✅|

## Related PRs and issues

- overall issue: #1404
- #3849
- #4623
- #4625
- #4617
- #4627 
- #4637
- #4631
- #3845

## Hexagon shape

**Pro**
 * Seems to be quite popular (even for Python)
 
**Contra** 
- lost of context / connection to Earth and compass
- difficult of implement based on fractions rather than on empircal factors
- difficult to make to logo conistent across both shapes
- 12 variants already for circle shape (color/bw, no/hor./vert. wordmark, light/dark), multiple tests with baseline images
- longer code, difficult to maintain

## Some important comments from #3849

- hexagon shape: support or drop
  https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r3106827807
  * [x] outer endings of main compass lines -> coverage by other shape
  * [ ] arrow length of letter T
  * [ ] same figure size across shapes
  * [x] position of wordmark, distance to visual
- examples: no `projection` and `region` parameters
  * [x] inline: https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r3143072063
  * [ ] gallery: separate PR
- wordmark: consistency with GMT
  * [x] font type: https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r2001963066
  * [ ] URL: https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r2321473375
  * [x] default: https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r3124963322
  * [ ] color choice: https://github.com/GenericMappingTools/pygmt/pull/4627#issuecomment-4416440150
- radii: fractions
[...]

PR #4617, opened 2026-05-02
Figure.pygmtlogo: Force the default logo size to 2-cm and rename test_pygmtlogo to test_pygmtlogo_circle_no_wordmark

seisman:
As shown in the table in PR https://github.com/GenericMappingTools/pygmt/pull/4616, we will likely have 24 PyGMT logo variants. Maintaining 24 separate baseline images is impractical, so we plan to group multiple variants into a single test. A reasonable approach is to define six tests based on logo shape and wordmark configuration. Each test would include four variants: colored and black-and-white logos in both light and dark themes.

This PR renames the existing test from test_pygmtlogo (which is too general) to test_pygmtlogo_circle_no_wordmark. It also adds a basemap to make the default logo size easier to inspect.

It appears that the default width/height of the logos is slightly smaller than 2 cm, although I am not sure why this value is used.

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -30,6 +30,7 @@ def _create_logo(  # noqa: PLR0915
     from pygmt.figure import Figure  # noqa: PLC0415
 
     # Helpful definitions
+    # See https://github.com/GenericMappingTools/pygmt/pull/4617#issuecomment-4366140123
     size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
     region = [-size, size] * 2
     proj = "x1c"
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -30,7 +30,7 @@ def _create_logo(  # noqa: PLR0915
     from pygmt.figure import Figure  # noqa: PLC0415
 
     # Helpful definitions
-    size = 4
+    size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
     region = [-size, size] * 2
     proj = "x1c"
     # Rotation around z-axis by 30 degrees counter-clockwise placed in the center.
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -30,7 +30,7 @@ def _create_logo(  # noqa: PLR0915
     from pygmt.figure import Figure  # noqa: PLC0415
 
     # Helpful definitions
-    size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
+    size = 4
     region = [-size, size] * 2
     proj = "x1c"
     # Rotation around z-axis by 30 degrees counter-clockwise placed in the center.
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -30,7 +30,6 @@ def _create_logo(  # noqa: PLR0915
     from pygmt.figure import Figure  # noqa: PLC0415
 
     # Helpful definitions
-    # See https://github.com/GenericMappingTools/pygmt/pull/4617#issuecomment-4366140123
     size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
     region = [-size, size] * 2
     proj = "x1c"
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -10,6 +10,7 @@ from typing import Literal
 
 import numpy as np
 from pygmt._typing import AnchorCode, PathLike
+from pygmt.exceptions import GMTValueError
 from pygmt.helpers import GMTTempFile, fmt_docstring
 from pygmt.params import Box, Position
 
@@ -304,7 +305,8 @@ def pygmtlogo(  # noqa: PLR0913
     width
     height
         Width or height of the PyGMT logo. Since the aspect ratio is fixed, only one of
-        the two can be specified.
+        the two can be specified. If not specified, the default size of the visual logo
+        is set to 2 cm.
     box
         Draw a background box behind the logo. If set to ``True``, a simple rectangular
         box is drawn using :gmt-term:`MAP_FRAME_PEN`. To customize the box appearance,
@@ -333,6 +335,20 @@ def pygmtlogo(  # noqa: PLR0913
     >>> fig.pygmtlogo(wordmark="horizontal", position="BR", height="1c")
     >>> fig.show()
     """
+    # Set the default size of the visual logo to 2 cm.
+    if width is None and height is None:
+        match wordmark:
+            case "none" | "vertical":
+                width = width or "2c"
+            case "horizontal":
+                height = height or "2c"
+            case _:
+                raise GMTValueError(
+                    wordmark,
+                    description="value for wordmark",
+                    choices={"none", "horizontal", "vertical"},
+                )
+
     with GMTTempFile(suffix=".eps") as logofile:
         # Create logo file
         _create_logo(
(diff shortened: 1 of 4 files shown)


pygmt/tests/test_pygmtlogo.py
```diff
@@ -4,16 +4,20 @@
 
 import pytest
 from pygmt import Figure
+from pygmt.params import Axis, Position
 
 
-@pytest.mark.benchmark
 @pytest.mark.mpl_image_compare
-def test_pygmtlogo():
+def test_pygmtlogo_circle_no_wordmark():
```
seisman, line 11: Other tests will be named like

- `test_pygmtlogo_circle_horizontal_wordmark`
- `test_pygmtlogo_circle_vertical_wordmark`
- `test_pygmtlogo_hexagon_no_wordmark`
- `test_pygmtlogo_hexagon_horizontal_wordmark`
- `test_pygmtlogo_hexagon_vertical_wordmark`

pygmt/tests/test_pygmtlogo.py
```diff
@@ -4,16 +4,20 @@
...
-    fig.pygmtlogo()
-    fig.shift_origin(xshift="+w")
-    fig.pygmtlogo(theme="dark")
+    fig.basemap(region=[-0.5, 5.5, -0.5, 2.5], projection="x1c", frame=Axis(grid=0.5))
+    fig.pygmtlogo(
+        position=Position((1, 1), anchor="CM", cstype="mapcoords"), theme="light"
+    )
+    fig.pygmtlogo(
+        position=Position((4, 1), anchor="CM", cstype="mapcoords"), theme="dark"
+    )
```
seisman, line 25: Plan to add the two black-and-white logos below the colored versions in later PRs.

seisman:
> It appears that the default width/height of the logos is slightly smaller than 2 cm, although I am not sure why this value is used.

One question, should we use the default logo size? I have no idea how the value is determined. 

The default GMT logo is 2.54 cm high, and the "no label" version is close to 2 cm. Maybe we should force the PyGMT logo height to 2-cm for the no-wordmark and horizontal-wordmark version, and width to 2-cm for the vertical-wordmark version?

```python
import pygmt

fig = pygmt.Figure()
fig.basemap(region=[0, 8, 0, 2], projection="x1c", frame="g1")
fig.logo(style="no_label")
fig.pygmtlogo(position=(6, 0))
fig.show()
```
logos

yvonnefroehlich:
> > It appears that the default width/height of the logos is slightly smaller than 2 cm, although I am not sure why this value is used.
> 
> One question, should we use the default logo size? I have no idea how the value is determined.
> 
> The default GMT logo is 2.54 cm high, and the "no label" version is close to 2 cm. Maybe we should force the PyGMT logo height to 2-cm for the no-wordmark and horizontal-wordmark version, and width to 2-cm for the vertical-wordmark version?

Hmm. This is an interesting question! I have to think about this tomorrow in more detail to understand the behaviour.

Currently we have in `_create_logo`:
```python
size = 4
region = [-size, size] * 2   # -> 8 plot-units
proj = "x1c"   # -> 1 cm per plot-units
```

I thougth the PyGMT logo saved to a file will have a size (diameter) of 8 cm for the circle version without wordmark.
For the PNG format (right), the default size is as expected when adding the image via `Figure.image` (code is at the end of the comment). But for the EPS format (left), the default size seems to be different:
image

---
Generating the PNG and EPS files with
```python
size = 1
region = [-size, size] * 2
proj = "x1c"
```
Leads to:
image

---
Setting `position="jMC+jMC+w2c` gives:
image

---
_Code_:
```python
import pygmt

fig = pygmt.Figure()

fig.basemap(region=[-4, 4] * 2, projection="x1c", frame=1)
fig.image(imagefile="pygmt_logo_saved_external.eps")#, position="jMC+jMC+w2c")
fig.basemap(frame="g1")

fig.shift_origin(xshift="+w+1c")

fig.basemap(region=[-4, 4] * 2, projection="x1c", frame=1)
fig.image(imagefile="pygmt_logo_saved_external.png")#, position="jMC+jMC+w2c")
fig.basemap(frame="g1")

fig.show()
```

seisman:
From the `image` source code, it seems the default size is controlled by the `GMT_GRAPHICS_DPU` setting, which defaults to 300 dots per inch.

yvonnefroehlich:
> From the `image` source code, it seems the default size is controlled by the `GMT_GRAPHICS_DPU` setting, which defaults to 300 dots per inch.

Thanks for potting out this parameter, I did not know about it and that it affects the `image` module.

Currently we have 
```python
size = 4
region = [-size, size] * 2   # -> 8 plot-units
proj = "x1c"   # -> 1 cm per plot-units
```
with `GMT_GRAPHICS_DPU = 300` (default).

By only adding the logo image in EPS format (without a basemap) and using **-V** we can manually find out the width and hight after applying `Figure.image`. For the circle version (without the wordmark) the width (or hight) is the diameter.
```python
import pygmt

fig = pygmt.Figure()
fig.image(imagefile="pygmt_logo_saved_external.eps")  # Logo is circle; use no basemap 
fig.show(V=True)
```
> psconvert [INFORMATION]: Figure dimensions: Width: 54.4932 points [1.9224 cm]  Height: 54.4932 points [1.9224 cm]

So we could either adjust `size` or `GMT_GRAPHICS_DPU` to make the default size of the logo (added via `Figure.image` using the EPS format) match 2 cm:

* size = (4 / 1.9224) * 2 = 4.161464835622139
* GMT_GRAPHICS_DPU = 300 - ((300/1.9224) * 2 - 300) = 287.8901373283396

> Figure dimensions: Width: 56.7848 points [2.00324 cm]  Height: 56.7848 points [2.00324 cm]

Not sure if this is an ideal solution.

seisman:
I've rounded `size` to `4.1615` so the default logo width is ~2-cm.

pygmt/src/pygmtlogo.py
```diff
@@ -30,7 +30,7 @@ def _create_logo(  # noqa: PLR0915
     from pygmt.figure import Figure  # noqa: PLC0415
 
     # Helpful definitions
-    size = 4
+    size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
```
yvonnefroehlich: Maybe we can add the link to the "calculation". I think it's difficult to understand how the value for `size` is determined because this behavior is unexpected and the parameter `GMT_GRAPHICS_DPU` is not so widely used.

```suggestion
    # See https://github.com/GenericMappingTools/pygmt/pull/4617#issuecomment-4366140123
    size = 4.1615  # size to make the circular logo 2-cm wide when plotted with 300 dpi.
```

seisman:
> I've rounded `size` to `4.1615` so the default logo width is ~2-cm.

Actually I feel this is not a good idea, since the actual size still depends on the GMT_GRAPHICS_DPU setting. Users would be surprised to see that GMT_GRAPHICS_DPU can affect the size of the default PyGMT logo size. I think the best solution is still to set the default logo size to 2-cm.

yvonnefroehlich:
> > I've rounded `size` to `4.1615` so the default logo width is ~2-cm.
> 
> Actually I feel this is not a good idea, since the actual size still depends on the GMT_GRAPHICS_DPU setting. Users would be surprised to see that GMT_GRAPHICS_DPU can affect the size of the default PyGMT logo size. I think the best solution is still to set the default logo size to 2-cm.

I am fine with both ways.
My estimation is definitely not an elegant way to solve this issue 🙂. Besides the effect by `GMT_GRAPHICS_DPU` am I am especially surprised, that the default size depends on the file format.

seisman:
Then I guess this PR can be approved and merged, right?

yvonnefroehlich (approved):

yvonnefroehlich:
> Then I guess this PR can be approved and merged, right?

Yes, sure. Sorry, did not see that you already changed code and docs.

PR #4623, opened 2026-05-06
Figure.pygmtlogo: Add a test to check the design details of the circular logo

seisman:
The existing test `test_pygmtlogo_circle_no_wordmark` works well for testing different variants of the PyGMT logos. However, the baseline image is relatively small (2-cm) and low resolution (maybe 150 dpi?), making minor logo changes difficult to detect. 

This PR adds a new test to verify the correctness of the circular logo, including all debugging gridlines and circles. The logo is 8 cm wide, and the baseline image is saved at 600 dpi, making the test more sensitive to subtle changes.

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -17,12 +17,12 @@ from pygmt.params import Box, Position
 __doctest_skip__ = ["pygmtlogo"]
 
 
-def _create_logo(  # noqa: PLR0915
+def _create_logo(  # noqa: PLR0915, PLR0912
     shape: Literal["circle", "hexagon"] = "circle",
     theme: Literal["light", "dark"] = "light",
     wordmark: Literal["none", "horizontal", "vertical"] = "none",
     color: bool = True,
-    figname: PathLike = "pygmt_logo.eps",
+    figname: PathLike | None = None,
     debug: bool = False,
 ):
     """
@@ -236,8 +236,8 @@ def _create_logo(  # noqa: PLR0915
         from pygmt import config  # noqa: PLC0415
 
         # Gridlines
-        with config(MAP_FRAME_TYPE="inside", MAP_GRID_PEN="0.1p,gray30"):
-            fig.basemap(frame="g1")
+        with config(MAP_GRID_PEN="0.1p,gray30"):
+            fig.basemap(frame="00g1")
         # Circles for the different radii
         for r in [r0, r1, r2, r3, r4, r5]:
             fig.plot(x=0, y=0, style=f"c{2 * r}c", pen="0.3p,gray30")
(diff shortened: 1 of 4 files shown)


pygmt/src/pygmtlogo.py
```diff
@@ -247,7 +247,10 @@ def _vline_coords():
         fig.hlines(y=[r4, r5], xmin=-3, pen=pen, perspective=True)
         fig.vlines(x=[r4, (thick_gap + r4) / 2], ymax=3, pen=pen, perspective=True)
 
-    fig.savefig(fname=figname)
+    if figname:
```
seisman, line 250: `figname` default to `None`. If `None`, return the `Figure` instance rather than saving it.

pygmt/src/pygmtlogo.py
```diff
@@ -236,8 +236,8 @@ def _vline_coords():
         from pygmt import config  # noqa: PLC0415
 
         # Gridlines
-        with config(MAP_FRAME_TYPE="inside", MAP_GRID_PEN="0.1p,gray30"):
```
seisman, line 239: `MAP_FRAME_TYPE=inside` is not needed because no annotations are shown

pygmt/src/pygmtlogo.py
```diff
@@ -236,8 +236,8 @@ def _vline_coords():
...
-            fig.basemap(frame="g1")
+        with config(MAP_GRID_PEN="0.1p,gray30"):
+            fig.basemap(frame="00g1")
```
seisman, line 240: `00` in `00g1` means no frame, only gridlines. The syntax is not Pythonic but I think it's fine to use it here.

pygmt/tests/test_pygmtlogo.py
```diff
@@ -5,6 +5,20 @@
...
+@pytest.mark.mpl_image_compare(savefig_kwargs={"dpi": 600})
+def test_pygmtlogo_circle_design():
+    """
+    Test the design details of the PyGMT circular logo.
+
+    This is a regression test to ensure that the design of the logo does not change
+    unintentionally. The debugging lines (gridlines and circles) are helpful for
+    implementing the logo, but they are not included in the final logo design.
+    """
+    fig = _create_logo(debug=True)
```
seisman, line 20: Use the private `_create_logo` function because it's the underlying function that produces the logo.

pygmt/src/pygmtlogo.py
yvonnefroehlich: Oh yes, I think I removed `a1` or used it only locally.

pygmt/src/pygmtlogo.py
yvonnefroehlich: Agree, we can use here the GMT synthax. No need to make the code longer using the newly implemented Parameter classes.

yvonnefroehlich (approved):

PR #4625, opened 2026-05-07
Figure.pygmtlogo: Add tests for the circular, black/white PyGMT logo

seisman:
Finalize the circular, black/white PyGMT logo for both light/dark themes.

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -221,7 +221,7 @@ def _create_logo(  # noqa: PLR0915, PLR0912
             x=0,
             y=0,
             style=f"{symbol}{size_shape + thick_shape}c",
-            pen=f"1p,{color_dark}",
+            pen=f"{thick_comp / 2.0}c,{color_bg}",
             perspective=True,
             no_clip=True,
         )
--- a/pygmt/tests/baseline/test_pygmtlogo_circle_no_wordmark.png.dvc
+++ b/pygmt/tests/baseline/test_pygmtlogo_circle_no_wordmark.png.dvc
@@ -1,5 +1,5 @@
 outs:
-- md5: b1dee02932b292335cf5f8b95676a258
-  size: 19161
+- md5: dcb54b2c28e02c4fc83d38278b2700c6
+  size: 31454
   hash: md5
   path: test_pygmtlogo_circle_no_wordmark.png
--- a/pygmt/tests/test_pygmtlogo.py
+++ b/pygmt/tests/test_pygmtlogo.py
@@ -37,4 +37,14 @@ def test_pygmtlogo_circle_no_wordmark():
         position=Position((3.5, 3.5), anchor="CM", cstype="mapcoords"),
         theme="dark",
     )
+    fig.pygmtlogo(
+        position=Position((1, 1), anchor="CM", cstype="mapcoords"),
+        theme="light",
+        color=False,
+    )
+    fig.pygmtlogo(
+        position=Position((3.5, 1), anchor="CM", cstype="mapcoords"),
+        theme="dark",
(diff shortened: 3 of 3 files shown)


pygmt/src/pygmtlogo.py
```diff
@@ -221,7 +221,7 @@ def _vline_coords():
             x=0,
             y=0,
             style=f"{symbol}{size_shape + thick_shape}c",
-            pen=f"1p,{color_dark}",
+            pen=f"{thick_comp / 2.0}c,{color_bg}",
```
seisman, line 224: `color_bg` is the same as `color_dark` in this case. I prefer not to use `color_dark` because it's not used after line 70.

The outline thickness `1p` is not ideal, because it doesn't scale well with `size`. Here I set the thickness to "half of the compass line".

Below is the high-resolution version of the black/white dark logo.

Image

yvonnefroehlich (commented):
Maybe change the title to: "Figure.pygmtlogo: Test the ...".

seisman:
> Maybe change the title to: "Figure.pygmtlogo: Test the ...".

Done.

pygmt/src/pygmtlogo.py
yvonnefroehlich: Looks good!

yvonnefroehlich: But wait maybe better to use `thick_gap` here for consistency.

yvonnefroehlich, line 224:
```suggestion
            pen=f"{thick_gap}c,{color_bg}",
```

seisman: `thick_gap` controls the spacing between letter M and G. It is noticeably thicker than the gaps surrounding the vertical red bar at the top.

Currently, `thick_comp = thick_shape/3` and `thick_gap=thick_shape/4`. Actually, I'm thinking about if we should set `thick_gap` to `thick_comp / 2.0`.

| thick_gap=thick_shape/3 | thick_gap = thick_comp/2.0| 
|---|---|
|logo1|logo2|

yvonnefroehlich: > `thick_gap` controls the spacing between letter M and G. It is noticeably thicker than the gaps surrounding the vertical red bar at the top.

I am sorry, the `/ 2.0` is still needed. 
What I like is that the thickness of this outline is identical to the gap between the red line and red arrowhead and the blue shape.

> Currently, `thick_comp = thick_shape/3` and `thick_gap=thick_shape/4`. Actually, I'm thinking about if we should set `thick_gap` to `thick_comp / 2.0`.

I _believe_ (did not search in the initial PR) we had `thick_gap = thick_comp / 2.0` at the beginning, and I increased the value to make sure the gap between the letters G and M is still clearly visible for a small size of the logo.

seisman: > What I like is that the thickness of this outline is identical to the gap between the red line and red arrowhead and the blue shape.

This gap between the red line and the blue shape is exactly `thick_comp / 2.0`, not `thick_gap`.

seisman: > I increased the value to make sure the gap between the letters G and M is still clearly visible for a small size of the logo.

It makes sense.

yvonnefroehlich: > > What I like is that the thickness of this outline is identical to the gap between the red line and red arrowhead and the blue shape.
> 
> This gap between the red line and the blue shape is exactly `thick_comp / 2.0`, not `thick_gap`.

OK. Then it should be fine 🙂. Seems like I confused myself with this radii.

yvonnefroehlich (approved):

PR #4626, opened 2026-05-07
Fix two static type errors reported by mypy 2.0.0

seisman:
The new mypy v2.0.0 reports two new errors:

```
mypy pygmt
pygmt/clib/session.py:113: error: Need type annotation for "GMT_CONSTANTS" (hint: "GMT_CONSTANTS: dict[<type>, <type>] = ...")  [var-annotated]
pygmt/figure.py:96: error: Need type annotation for "SHOWED_FIGURES" (hint: "SHOWED_FIGURES: list[<type>] = ...")  [var-annotated]
Found 2 errors in 2 files (checked 128 source files)
make: *** [Makefile:74: typecheck] Error 1
```

xref: https://github.com/GenericMappingTools/pygmt/actions/runs/25474333778/job/74744492600

### DIFF
--- a/pygmt/clib/session.py
+++ b/pygmt/clib/session.py
@@ -110,7 +110,7 @@ DTYPES_TEXT = {
 DTYPES = DTYPES_NUMERIC | DTYPES_TEXT
 
 # Dictionary for storing the values of GMT constants.
-GMT_CONSTANTS = {}
+GMT_CONSTANTS: dict[str, int] = {}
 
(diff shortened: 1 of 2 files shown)


weiji14 (approved):
Was gonna recommend using [parallel type checking](https://mypy.readthedocs.io/en/stable/changelog.html#parallel-type-checking), but then remembered this runs on ubuntu-slim with just 1 CPU 😆

PR #4628, opened 2026-05-08
Figure.set_panel: Fix the bug when panel is not set

seisman:
In the GMT CLI, using `gmt subplot set` without specifying the panel activates the next panel. For example:
```
gmt begin map
	gmt subplot begin 1x2 -Fs6c/3c
	gmt subplot set
	gmt basemap -R0/3/0/3 -B
	gmt subplot set 
	gmt basemap -R0/3/0/3 -B
	gmt subplot end
gmt end show

```

 The equivalent PyGMT script is:
```python
import pygmt

fig = pygmt.Figure()
with fig.subplot(nrows=1, ncols=2, figsize=("6c", "3c")):
     with fig.set_panel():
     with fig.set_panel():
fig.show()
```
But it doesn't work as expected:
```
subplot [ERROR]: No subplot information file!
...
```
The bug is because we always pass `panel` to `gmt subplot set` even if it's `None`. The bug can be traced back to the initial implementation in PR #822.

This PR fixes the long-term bug and add a test for it.

### DIFF
--- a/pygmt/src/subplot.py
+++ b/pygmt/src/subplot.py
@@ -419,13 +419,11 @@ def set_panel(
     aliasdict = AliasSystem(A=Alias(tag, name="tag")).add_common(V=verbose)
     aliasdict.merge(kwargs)
 
+    args = ["set"]
+    if panel is not None:
+        args.append(Alias(panel, name="panel", sep=",", size=2)._value)  # type: ignore[arg-type]
+    args.extend(build_arg_list(aliasdict))
+
     with Session() as lib:
-        lib.call_module(
-            module="subplot",
-            args=[
-                "set",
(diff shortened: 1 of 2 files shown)


weiji14 (approved):
Somewhat prefer having the panel be explicit, but our typehint does say `panel: int | Sequence[int] | None = None`, so I guess the default of `panel=None` should work somehow like in GMT.

seisman:
Sometimes, it's easier to activate the next panel without specifying its index by using `panel=True` in plotting methods. However, this method can't override the subplot tag. In such cases, we can use `Figure.set_panel(tag="XXX")`. This is how I discovered the bug.

PR #4631, opened 2026-05-09
Figure.pygmtlogo: Improve the design of the hexagonal logo

seisman:
Different designs for the hexagonal logo:
| V1 | V2| V3| V4| V5|
|---|---|---|---|---|
| logo-v0 | logo-v1 | logo-v2  | logo-v3 | logo-v3 |
| Current version. Angle of diagonal line = 45 | Angle of diagonal line = 30 | Angle of diagonal line = 60 | Remove diagonal lines | Remove all lines |

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -22,6 +22,8 @@ exclude-labels:
   - 'skip-changelog'
 exclude-contributors:
   - 'actions-bot'
+  - 'Copilot'
+  - 'copilot-swe-agent[bot]'
   - 'dependabot'
   - 'dependabot[bot]'
 category-template: '### $TITLE'
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -48,7 +48,7 @@ replacers:
 template: |
   ## Release v$RESOLVED_VERSION (20YY/MM/DD)
 
-  [![Digital Object Identifier for PyGMT v$RESOLVED_VERSION](https://zenodo.org/badge/DOI/10.5281/zenodo.3781524.svg)](https://doi.org/10.5281/zenodo.3781524)
+  [![Digital Object Identifier for PyGMT v$RESOLVED_VERSION](https://img.shields.io/badge/DOI-10.5281/zenodo.3781524-blue)](https://doi.org/10.5281/zenodo.3781524)
 
   ### Highlights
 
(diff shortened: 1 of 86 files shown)


yvonnefroehlich:
> Different designs for the hexagonal logo:

Hmm, not sure about this. I think we should keep the compass lines consistent with the circle version. If the horizontal/vertical lines as well as diagonal lines are not orthogonal to each other anymore they do only make little sense in the context of a compass.

seisman:
We once had a discussion (https://github.com/GenericMappingTools/pygmt/pull/3849#discussion_r3116095201) about the size of the hexagon outline, and it turns out there are many possible choices. 
image

Clearly, the version where the r1 circle serves as the inscribed circle (the r0 - 0.3 one) is the simplest, since it requires no additional adjustments, except the length of the yellow compass lines and the offsets (maybe also font size) of the wordmark. I prefer this option because of its simplicity. 

@yvonnefroehlich mentioned that 

> I would like to have exactly the same height and width (or at least the same aspect ratio) of the logo when the trademark is included, independent of the chosen shape.

I think that in practice, people are unlikely to display the circular and hexagonal logos side by side, so the difference in overall size does not matter much.

In 3d92d329e9deeb74e831e27d30298c86261bea13, I've changed the shape size from `(r0 + 0.34) * 2` to the more accuarate size `(r0 + r1) / np.cos(np.deg2rad(30))`.

The top post shows different versions of the hexagon logo with different compass-line designs. Here are my thoughts:

* V5: The interior feels too empty without compass lines.
* V4: Not bad. The yellow line can represent the equator. Or the red line and letter T represents N-S direction, while the yellow line represent the E-W direction.
* V3: I don't like it. The yellow lines make the whole logo look like a lightning torch.
* V2: The yellow lines are evenly distributed and symmetric. However, the two yellow lines at the bottom intersect with the red letter “T”, which is not ideal. As pointed out by @yvonnefroehlich "If the horizontal/vertical lines as well as diagonal lines are not orthogonal to each other anymore they do only make little sense in the context of a compass.". So this one is also not good.
* V1: The current version. It is compatible with the circular logo, but the diagonal compass lines are not perpendicular to the edges of the hexagon outline, which I don’t like.

Personablly I like V4 for its simplicity, although I agree with @yvonnefroehlich that "we should keep the compass lines consistent with the circle version" if possible. But as I mentioned above "in practice, people are unlikely to display the circular and hexagonal logos side by side", and the main elements are unchanged between the circular and hexagonal logos.

PR #4634, opened 2026-05-11
Figure.pygmtlogo: Set the region based on wordmark for easier design

seisman:
Currently, the `region` is set to `[-4, 4, -4, 4]`. The issues are:

1. We need to use `no_clip=True` when adding the wordmark
2. We can't adding debugging lines (gridlines) for wordmarks, so we can't control the size of the wordmark accurately.

This PR does three things:

- sets the `region` based on `wordmark`, i.e., larger x-range for horizontal wordmark and larger y-range for vertical wordmark
- slightly modifies the code for adding debugging lines to ensure the debugging lines are plotted inside the visual logo only
- adds two baseline images for the logo design with horizontal/vertical wordmark. The orginal baseline image for a visual logo without a wordmark is no longer needed, because  it's already covered by the two new baseline images.

~~Ideally, we should be able to remove the `no_clip=True` argument from the `Figure.text` call. But as you can see, the wordmark in the vertical version is slightly wider than the visual logo.~~

Edit: I've slightly decreased the font size for the vertical wordmark, and then remove the `no_clip` parameter from `Figure.text`.

| Old version | New version |
|---|---|
|image | image|

Please note that the design of wordmarks is still ongoing in PR #4627. So, this PR focuses on adding tests to make the wordmark design easier.

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -246,7 +246,7 @@ def _create_logo(  # noqa: PLR0915
         # Lines for letter M
         fig.hlines(y=[r4, r5], xmin=-size, xmax=size, pen=pen, perspective=True)
         m_mid = (thick_gap + r4) / 2
-        fig.vlines(x=[r4, m_mid], ymin=size, ymax=size, pen=pen, perspective=True)
+        fig.vlines(x=[r4, m_mid], ymin=-size, ymax=size, pen=pen, perspective=True)
 
     if figname:
         fig.savefig(fname=figname)
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -32,8 +32,13 @@ def _create_logo(  # noqa: PLR0915
 
     # Helpful definitions
     size = 4
-    region = [-size, size] * 2
     proj = "x1c"
+    region = {
+        "horizontal": [-size, size * 8.0, -size, size],
+        "vertical": [-size, size, -size * 1.75, size],
+        "none": [-size, size, -size, size],
+    }[wordmark]
+
     # Rotation around z-axis by 30 degrees counter-clockwise placed in the center.
     perspective = "30+w0/0"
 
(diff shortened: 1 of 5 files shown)


pygmt/src/pygmtlogo.py
```diff
@@ -240,8 +245,9 @@ def _vline_coords():
         pen = "0.3p,gray30,2_2"
         fig.plot(x=0, y=0, style=f"c{2 * (r2 + (r3 - r4))}c", pen=pen)
         # Lines for letter M
-        fig.hlines(y=[r4, r5], xmin=-3, pen=pen, perspective=True)
-        fig.vlines(x=[r4, (thick_gap + r4) / 2], ymax=3, pen=pen, perspective=True)
+        fig.hlines(y=[r4, r5], xmin=-size, xmax=size, pen=pen, perspective=True)
+        m_mid = (thick_gap + r4) / 2
+        fig.vlines(x=[r4, m_mid], ymin=size, ymax=size, pen=pen, perspective=True)
```
seisman: Limits the debugging lines to `[-size, size, -size, size]`.

pygmt/src/pygmtlogo.py
```diff
@@ -240,8 +244,9 @@ def _compass_lines():
...
```
yvonnefroehlich:
```suggestion
        fig.vlines(x=[r4, m_mid], ymin=-size, ymax=size, pen=pen, perspective=True)
```

pygmt/src/pygmtlogo.py
```diff
@@ -240,8 +244,9 @@ def _compass_lines():
         pen = "0.3p,gray30,2_2"
         fig.plot(x=0, y=0, style=f"c{2 * (r2 + (r3 - r4))}c", pen=pen)
         # Lines for letter M
-        fig.hlines(y=[r4, r5], xmin=-3, pen=pen, perspective=True)
-        fig.vlines(x=[r4, (thick_gap + r4) / 2], ymax=3, pen=pen, perspective=True)
+        fig.hlines(y=[r4, r5], xmin=-size, xmax=size, pen=pen, perspective=True)
+        m_mid = (thick_gap + r4) / 2
+        fig.vlines(x=[r4, m_mid], ymin=-size, ymax=size, pen=pen, perspective=True)
```
yvonnefroehlich: Would make these lines sligthly shorter:

```suggestion
        size_s = size - size * 0.1
        fig.hlines(y=[r4, r5], xmin=-size_s, xmax=size_s, pen=pen, perspective=True)
        m_mid = (thick_gap + r4) / 2
        fig.vlines(x=[r4, m_mid], ymin=-size_s, ymax=size_s, pen=pen, perspective=True)
```

|  |  |
| --- | --- |
| Image | Image |

seisman: maybe 0.9*size

yvonnefroehlich (approved):

PR #4635, opened 2026-05-11
Figure.pygmtlogo: Increase the gap around the vertical red line and arrow

seisman:
Currently, the gap between the vertical red line and the outline shape is set to `thick_comp / 2`, i.e., half the thickness of the compass line. This value was chosen arbitrarily.

This PR changes the gap from `thick_comp / 2` to `thick_gap`, making it consistent with the gap between the letters G and M. The new gap is slightly larger, making it more visible when logos are small.

After this change, the variable `thick_comp` is used only for the compass lines.

| Old | New |
|---|---|
| image | image|

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -139,14 +139,21 @@ def _create_logo(  # noqa: PLR0915
         mask = np.abs(t_x) <= (thick_gap + r4) / 2
         return {"x": t_x[mask], "y": t_y[mask]}
 
+    def _vline_coords():
+        """
+        Coordinates for the vertical line at the top.
+        """
+        x0 = thick_gt / 2
+        return {"x": [-x0, -x0, x0, x0], "y": [r0, r3, r3, r0]}
+
     def _bg_arrow_coords():
         """Coordinates for the background arrow."""
         # x0, y0 is the same as in _letter_t_coords().
         x0 = thick_gt / 2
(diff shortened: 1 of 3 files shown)


pygmt/src/pygmtlogo.py
```diff
@@ -139,14 +139,21 @@ def _letter_t_coords():
         mask = np.abs(t_x) <= (thick_gap + r4) / 2
         return {"x": t_x[mask], "y": t_y[mask]}
 
+    def _vline_coords():
```
seisman, line 142: I moved `_vline_coords` after `_letter_t_coords`, because the red vertical line at the top is more related to the letter T.

pygmt/src/pygmtlogo.py
```diff
@@ -139,14 +139,21 @@ def _letter_t_coords():
...
+        """
+        x0 = thick_gt / 2
+        return {"x": [-x0, -x0, x0, x0], "y": [r0, r3, r3, r0]}
+
     def _bg_arrow_coords():
         """Coordinates for the background arrow."""
         # x0, y0 is the same as in _letter_t_coords().
         x0 = thick_gt / 2
         y0 = 1.8 * x0 * np.sqrt(3)
-        # The background arrow is thick_comp wider than the letter T.
```
seisman, line 147: Replace `thick_comp` with `thick_gap`.

yvonnefroehlich (approved):

PR #4638, opened 2026-05-13
Figure.pygmtlogo: Remove the additional black outline for the black-and-white dark-theme logo

seisman:
Currently, we add an additional black outline to the black-and-white logo in dark theme. But is it really necessary? Dark mode usually implies a dark (black) background, and the white shape outline is already visually distinct from the background (see below).
```python
import pygmt
from pygmt.params import Frame

fig = pygmt.Figure()
fig.basemap(region=[-2, 2, -2, 2], projection="X4c", frame=Frame(fill="darkgray"))
fig.pygmtlogo(color=False, theme="dark", width="3c")
fig.show()
```
| With black outline | Without black outline |
|---|---|
| logo-dark | logo-dark|

### DIFF
--- a/pygmt/src/pygmtlogo.py
+++ b/pygmt/src/pygmtlogo.py
@@ -216,17 +216,6 @@ def _create_logo(  # noqa: PLR0915
     # Upper vertical line
     fig.plot(data=_vline_coords(), fill=red, perspective=True)
 
-    # Outline around the shape for black and white color with dark theme
-    if not color and theme == "dark":
-        fig.plot(
-            x=0,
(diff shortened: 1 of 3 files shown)


yvonnefroehlich:
The outline is nice from a visual perspective, but from a consistency perspective, I prefer to remove this outline. I never really liked it, because it affects the size of the logo or the thickness of the surrounding shape.

seisman:
I've remove the black outline, and also updated the baseline image with a gray background.

yvonnefroehlich (approved):

Issue #4599, opened 2026-04-24
DOC: Support dynamic  filtering by tags for gallery examples with Sphinx-Gallery v0.21.0


seisman:
[Sphinx-Gallery v0.21.0](https://github.com/sphinx-gallery/sphinx-gallery/releases/tag/v0.21.0) introduces tag support for gallery examples, which allows examples to be tagged and filtered more effectively. See https://sphinx-gallery.github.io/stable/configuration.html#tagging-examples-and-filtering for the usage and https://sphinx-gallery.github.io/stable/auto_examples/index.html?#general-examples for a live demo.

This feature would be very beneficial for the PyGMT gallery, which has grown substantially and can be difficult to navigate, especially for new users. Tag support would allow:

* Better discoverability of examples
* More flexible navigation
* Easier onboarding for new users
* Improved documentation structure without major reorganization

## Potential checklist

- [x] Decide if we want to support tags 
- [ ] Decide on the list of allowed tags for PyGMT
- [ ] Document the list of allowed tags
- [x] Upgrade to Sphinx-Gallery ≥ 0.21.0
- [ ] Add tags to gallery examples using the new tagging mechanism [definitely in multiple PRs]
- [ ] See if we should reorganize the structure of the gallery
- [ ] Update contributing guides

Chuan1937:
Sure, we should add tags. Such as:
| Tag | Description |                                                                                                                                           
  |-----|-------------|                                                                                                                                           
  | `maps` | Maps and map elements |                                                                                                                              
  | `lines` | Lines and curves |                                                                                                                                  
  | `symbols` | Symbols and markers |                                                                                                                             
  | `images` | Images and grids |                                                                                                                                 
  | `3d` | 3D plots |                                             
  | `seismology` | Seismology |
  | `basemaps` | Basemaps (dual y-axis, ternary, etc.) |
  | `histograms` | Histograms and rose diagrams |
  | `embellishments` | Embellishments (colorbars, legends, etc.) |
  | `projections` | Map projections |                                                                                                                             
  | `beginner` | Beginner-friendly |
  | `advanced` | Advanced |

yvonnefroehlich:
I agree that adding tags will make it easier to navigate through the examples.

[...]

PR #4643, opened 2026-05-19
Add filtering tags for gallery examples of projections

seisman:
Changes in this PR:

- [x] Bump to sphinx-gallery >= 0.21.0
- [X] Add tags for projection gallery examples [Based on the figure and table in https://docs.generic-mapping-tools.org/dev/reference/options.html#coordinate-transformations-and-map-projections-the-j-option]

**Preview**: https://pygmt-dev--4643.org.readthedocs.build/en/4643/projections/index.html

The first PR to address #4599

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -162,6 +162,18 @@ html.writer-html5 .rst-content dl.field-list {
     max-width: 80%!important;
 }
 
+.sphx-glr-tag-label {
+    font-weight: bold;
+    background-color: #f2f3f5;
+    padding: 0.1rem 0.35rem;
+}
+
+.sphx-glr-tag {
+    background-color: #e6eef8;
+    margin: 0.1rem 0.2rem 0.1rem 0;
+    padding: 0.1rem 0.35rem;
+}
+
 /*
  * Styles for aligning table cells.
  * https://myst-parser.readthedocs.io/en/latest/syntax/tables.html#markdown-syntax
--- a/environment.yml
+++ b/environment.yml
@@ -41,7 +41,7 @@ dependencies:
     - sphinx-autodoc-typehints
     - sphinx-copybutton
     - sphinx-design
-    - sphinx-gallery>=0.19.0
+    - sphinx-gallery>=0.21.0
     - sphinx_rtd_theme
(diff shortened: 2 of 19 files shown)


yvonnefroehlich:
Looking at the example by Sphinx at https://sphinx-gallery.github.io/stable/auto_examples/index.html?#general-examples, there should be a nice tag symbol. However, in the preview of this PR, it looks a bit strange. Besides this, I feel it could improve the structure if "Tags" were written in bold font, but I think we cannot change this.

| Preview of this PR | Example by Sphinx |
| --- | --- |
| tag_symbol_pygmt | tag_symbol_example |

seisman:
> there should be a nice tag symbol. However, in the preview of this PR, it looks a bit strange.

I can see the tag symbol in the preview, so it's likely a browser issue.


> I feel it could improve the structure if "Tags" were written in bold font, but I think we cannot change this.

Actually we can, I've adjusted the style in commit fe40f73a9b6535e53dabac28a21a65961e6fe1b3

image

yvonnefroehlich:
> > there should be a nice tag symbol. However, in the preview of this PR, it looks a bit strange.
> 
> I can see the tag symbol in the preview, so it's likely a browser issue.

Your are probably right, on my smarphone it looks good.

> > I feel it could improve the structure if "Tags" were written in bold font, but I think we cannot change this.
> 
> Actually we can, I've adjusted the style in commit [fe40f73](https://github.com/GenericMappingTools/pygmt/commit/fe40f73a9b6535e53dabac28a21a65961e6fe1b3)

That's good and with the changes it looks much better.

weiji14 (commented):
I think only [Mollweide](https://en.wikipedia.org/wiki/Mollweide_projection) (equal-area) is missing a tag. Otherwise should be ok.

seisman:
> I think only [Mollweide](https://en.wikipedia.org/wiki/Mollweide_projection) (equal-area) is missing a tag. Otherwise should be ok.

Nice catch. Fixed in f71ee743d7b0ac18af6124c0b92108c3dd0bb019.

weiji14 (approved):

PR #4652, opened 2026-05-22
Figure.text: Allow passing a sequence of offset to the 'offset' parameter

seisman:
Passing a sequence of offsets to `Figure.text` is frequently used but not supported yet.

This PR supports it and also migrates `offset` to the new alias system.

### DIFF
--- a/examples/gallery/images/cross_section.py
+++ b/examples/gallery/images/cross_section.py
@@ -61,7 +61,7 @@ fig.text(
     x=[lonA, lonB],
     y=[latA, latB],
     text=["A", "B"],
-    offset=(0, 0.3),  # Move text 0.2 centimeters up (y-direction)
+    offset=(0, 0.3),  # Move text 0.3 centimeters up (y-direction)
     font="15p,red",  # Use a red font with a size of 15 points
 )
 
--- a/examples/gallery/images/cross_section.py
+++ b/examples/gallery/images/cross_section.py
@@ -61,7 +61,7 @@ fig.text(
     x=[lonA, lonB],
     y=[latA, latB],
     text=["A", "B"],
-    offset="0c/0.3c",  # Move text 0.2 centimeters up (y-direction)
+    offset=(0, 0.3),  # Move text 0.2 centimeters up (y-direction)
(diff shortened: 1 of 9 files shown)


examples/gallery/images/cross_section.py
```diff
@@ -61,7 +61,7 @@
     x=[lonA, lonB],
     y=[latA, latB],
     text=["A", "B"],
-    offset="0c/0.3c",  # Move text 0.2 centimeters up (y-direction)
+    offset=(0, 0.3),  # Move text 0.2 centimeters up (y-direction)
```
seisman:
```suggestion
    offset=(0, 0.3),  # Move text 0.3 centimeters up (y-direction)
```

examples/gallery/images/cross_section.py
```diff
@@ -61,7 +61,7 @@
     x=[lonA, lonB],
     y=[latA, latB],
     text=["A", "B"],
-    offset="0c/0.3c",  # Move text 0.2 centimeters up (y-direction)
+    offset=(0, 0.3),  # Move text 0.3 centimeters up (y-direction)
```
weiji14, line 64: Should we keep the "c" unit specifier? The [gmt.conf docs](https://docs.generic-mapping-tools.org/6.6/gmt.conf.html#description) say:

> It is recommended that users specify the units for distances and lengths by appending c (cm), i (inch), or p (points) when changing parameters using any of these methods. By default, when no unit is specified the value will be assumed to be cm for parameters not related to fonts or pen thicknesses and will be assumed to be points for parameters related to fonts or pen thicknesses.

So recommended is to append the unit, but no unit means centimeter in some but not all cases (I guess `offset` counts as non-font/pen thickness?). Maybe keep it as `("0c", "0.3c")` to be clearer?

seisman: Yes, appending the unit specifier is recommended in the GMT CLI. However, in PyGMT, I feel it is more natural to use plain numbers for quantities such as offsets, widths, and sizes. They are much shorter to type and can also be used directly in calculations.

Issue #4677, opened 2026-06-14
Add an AGENTS.md file


seisman:
As AI coding agents such as GitHub Copilot, Claude Code, Codex, and Cursor become more widely used, it may be helpful for PyGMT to provide an AGENTS.md file with repository-specific instructions.

The file could briefly describe:

- The purpose and structure of PyGMT
- How to set up a development environment
- Testing requirements
- Coding and documentation conventions
- Useful links to contributor and developer documentation
- More

Many projects are starting to adopt AGENTS.md files to help AI agents understand project workflows and produce contributions that better follow project standards.

References:
 
- https://agents.md/

weiji14:
Is the intention to just provide an AGENTS.md for developing/contributing/maintaining the PyGMT repository?

> * The purpose and structure of PyGMT
> * Coding and documentation conventions

These should be general enough to go in the AGENTS.md file, since they apply to every situation.

I'm still a little hesitant in adding this now, because some agent harnesses might use different file-naming conventions (e.g. CLAUDE.md, SOUL.md, etc), and you need to symlink the file to the correct name to have it work by default... But also maybe things will standardize to AGENTS.md soon?

> * How to set up a development environment
> * Testing requirements
> * Useful links to contributor and developer documentation

These feel more like [SKILLS](https://agentskills.io/home), that only apply in specific situations (e.g. tests don't need to run all the time). We might need to separate these out?

Side note: if we ever think about distributing more user-facing skills, we should think about packaging them properly - https://pavel.pink/blog/pixi-skills

seisman:
I don't have much experience with AGENTS.md yet. When leveraging AI tools for repetitive tasks, I’ve occasionally seen the AI lack project context and use incorrect ways to test. As an example, in https://github.com/GenericMappingTools/pygmt/pull/4582#pullrequestreview-4110944262, Codex validated the changes with `py_compile`, though I expected `pytest`. That led me to wonder how we can teach the AI to adhere to our specific instructions and read a little more about AGENTS.md.

> I'm still a little hesitant in adding this now, because some agent harnesses might use different file-naming conventions (e.g. CLAUDE.md, SOUL.md, etc), and you need to symlink the file to the correct name to have it work by default... But also maybe things will standardize to AGENTS.md soon?

[...]

PR #4680, opened 2026-06-15
WIP + POC: Add AGENTS.md

seisman:
Initial draft done by codex.

Address #4677


AGENTS.md
```diff
@@ -0,0 +1,234 @@
...
+Use Sphinx roles consistently:
+
+- `:func:`, `:class:`, `:meth:`, `:mod:` for Python API references.
+- `:doc:` for documentation pages.
+- `:gmt-docs:` for GMT documentation pages.
+- `:gmt-term:` for GMT configuration parameters.
+
+## Maintainer and Release Context
+
+Most agents should not perform maintainer-only operations unless explicitly asked.
```
willschlitzer, line 210: Should this explicitly list the maintainer-only operations that are forbidden? I'm assuming the intent is to keep agents from submitting PRs without human oversight?

seisman: First of all, the current version of the AGENTS.md file was generated entirely by Codex, so some wording may not be ideal.

I actually think it's fine to let agents make whatever changes they consider appropriate. The responsibility is then on us, as maintainers, to carefully review the resulting PRs and merge them manually.

willschlitzer: My concern is that agents will begin sending in excessive PRs that aren't necessarily constructive; I obviously don't think it will be as bad as on really popular packages like curl, but I think that agent-submitted PRs could lead to extra work for volunteer maintainers.

In my own experience with using AI to suggest code improvements, I find that it makes a lot of style suggestions that don't actually fix bugs, add features, or meaningfully improve documentation without more specific direction. My concern is that there would be PRs that would simply suggest renaming variables/functions or other similar changes that would take away from reviewers' time.

Also, as someone who came into open-source development by finding the low-hanging fruit on a PyGMT, my concern is that agents will consume all of the fixes that used to be tagged as "good first issue" and limit the chances for newcomers to the project to meaningfully contribute without extensive experience.

seisman: Those are good points. I agree that a flood of low-value PRs would create unnecessary work for maintainers, and I also wouldn't want AI to crowd out “good first issues” that help new contributors get involved.

That said, I think these are project governance questions rather than something AGENTS.md itself can solve. AGENTS.md tells agents how to contribute, while the AI-use policies probably belong in contributing guides or the project's review guidelines.

AGENTS.md
weiji14, line 1: In general, I feel that this file is too long (200+ lines) and might fill up a good portion of an agent's context window. We should distinguish between general guidelines that should apply for every prompt/message (e.g. the style guide), vs specialized [SKILLS](https://agentskills.io/home) that only apply for specific situations (e.g. wrapping a GMT module).

willschlitzer: I would agree with @weiji14 on this one that it is going to eat up tokens to have this AGENTS.md for every AI interactions. Would we be able to have a shorter AGENTS.md in the main directory for overarching guidance and then have nested AGENTS.md in src/, tests/, docs/, etc. for GenAI/LLM instructions relative to those specific directories?

PR #4682, opened 2026-06-16
Fix package names ruff->Ruff and ghostscript->Ghostscript

seisman:
As per the official documentation (https://docs.astral.sh/ruff/), the official package name is `Ruff`, not `ruff`.


yvonnefroehlich (approved):
We should also fix this in Acknowledgments of the manuscript either during the second of review or when checking the proofs.

doc/changes.md
```diff
@@ -576,8 +576,8 @@
 * Adopt SPEC 0 policy and drop NEP 29 policy ([#3037](https://github.com/GenericMappingTools/pygmt/pull/3037))
 * Document the support policy for minimum required GMT versions ([#3070](https://github.com/GenericMappingTools/pygmt/pull/3070))
 * Bump to ghostscript 10.03.0 ([#3112](https://github.com/GenericMappingTools/pygmt/pull/3112))
-* Bump to ruff 0.3.0 ([#3081](https://github.com/GenericMappingTools/pygmt/pull/3081))
-* Enable ruff's PTH (flake8-use-pathlib) rules and fix violations ([#3129](https://github.com/GenericMappingTools/pygmt/pull/3129))
+* Bump to Ruff 0.3.0 ([#3081](https://github.com/GenericMappingTools/pygmt/pull/3081))
+* Enable Ruff's PTH (flake8-use-pathlib) rules and fix violations ([#3129](https://github.com/GenericMappingTools/pygmt/pull/3129))
```
weiji14: Probably don't need to capitalize if it's written like \<package> \<version> (so use ruff 0.3.0 here)? E.g. we don't capitalize ghostscript on the line above before its version number?

seisman: > E.g. we don't capitalize ghostscript on the line above before its version number?

Actually, we use `Ghostscript` on other lines.

seisman: I've fixed ghostscript to Ghostscript in 439e6cc69054a5aa5573fabd000557444a5f38e2.

weiji14 (approved):

PR #4687, opened 2026-06-21
Figure.pygmtlogo: Extend the region to get rid of the 'no_clip' parameter

seisman:
Extending the `region` parameter by 20% so that we no longer need the `no_clip` parameter when plotting hexagonal logos and wordmarks.

Changes this PR don't affect the logo appearance, but affect the debugging basemap in the tests for logo designs.


yvonnefroehlich (commented):
Looks good to me!

PR #4702, opened 2026-06-29
Figure.directional_rose/Figure.magnetic_rose: Set the default position to the top right

seisman:

For GMT<=6.6.0, the position argument is required when plotting directional or magnetic roses. When implementing `Figure.directional_rose` and `Figure.magnetic_rose` in PyGMT, we decided to set the default position to the "bottom left" corner.

In a recent PR https://github.com/GenericMappingTools/gmt/pull/9074, the default position was changed to "Top Right." As a result, the default positions will be inconsistent between GMT and PyGMT.

Fortunately, `Figure.directional_rose` and `Figure.magnetic_rose` were added recently and have not yet been released. This gives us the opportunity to change the default behavior without breaking backward compatibility.

This PR updates the default position.

### DIFF
--- a/examples/gallery/embellishments/magnetic_rose.py
+++ b/examples/gallery/embellishments/magnetic_rose.py
@@ -13,7 +13,7 @@ fig = pygmt.Figure()
 fig.basemap(region=[-8, 8, -7, 7], projection="M10c", frame=True)
 
 # Add a magnetic rose
-# By default, it's placed in the lower left corner
+# By default, it's placed in the Top Right corner
 fig.magnetic_rose()
 
 # Add a magnetic rose with several adjustments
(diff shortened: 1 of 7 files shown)


examples/gallery/embellishments/magnetic_rose.py
```diff
@@ -13,7 +13,7 @@
 fig.basemap(region=[-8, 8, -7, 7], projection="M10c", frame=True)
 
 # Add a magnetic rose
-# By default, it's placed in the lower left corner
+# By default, it's placed in the Top Right corner
```
yvonnefroehlich, line 16: We also need to updated the intro text the gallery example "Directional map rose" at https://www.pygmt.org/dev/gallery/embellishments/directional_roses.html.

yvonnefroehlich (commented):
I agree the default position should be consistent across GMT and PyGMT.

yvonnefroehlich:
Seems like we have to do similar changes to `Figure.inset` (xref https://github.com/GenericMappingTools/gmt/pull/9077), if we want consistent default positions for the plotting embellishments.

PR #4704, opened 2026-06-29
Deploy documentation after Linux docs build

seisman:
## Summary
- split docs CI into a Linux docs build that produces deployable artifacts and a macOS/Windows docs check job
- deploy documentation from a separate job that depends only on the Linux docs build
- extract shared docs build steps into a local composite action while preserving the current GMT data server setting and Windows make-error workaround

## Verification
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci_docs.yml"); YAML.load_file(".github/workflows/actions/build_docs/action.yml"); puts "YAML OK"'
- git diff --check -- .github/workflows/ci_docs.yml .github/workflows/actions/build_docs/action.yml

Note: actionlint is not installed in this environment.


seisman:
Closing this because the separate deploy-job design duplicates too much of the docs build workflow. The lower-duplication design is to keep deployment as Ubuntu-only steps in the existing docs matrix job: the Ubuntu docs build completes, then its deploy steps run, without waiting for macOS/Windows matrix children. A separate downstream deploy job cannot depend only on the Ubuntu matrix child in GitHub Actions without splitting/duplicating the build path.

PR #4729, opened 2026-07-17
Figure.directional_rose/Figure.magnetic_rose: Fix compatibility with upcoming GMT 6.7.0

seisman:
As shown in the GMT Dev Tests (https://github.com/GenericMappingTools/pygmt/actions/runs/29382671247/job/87249463921), calling `Figure.directional_rose()` and `Figure.magnetic_rose()` without any arguments don't work for GMT 6.7.0.

This is because for GMT 6.7.0, we set the default position to `None`, but in PyGMT, it means the `-Td` option is not specified, but actually we need `-Td`. We can either set `default=True` or simply always set the `position` parameter. This PR takes the 2nd option.

Patches https://github.com/GenericMappingTools/pygmt/pull/4702.


yvonnefroehlich:
Hmm, as this is related to the upcoming GMT 6.7.0 version, is this a bug or a breaking change for PyGMT?

seisman:
It's a bug

seisman:
It has been three weeks since v0.19.0, and this seems to be the only bug for the upcoming GMT 6.7.0. Should we release v0.19.1 soon or go with v0.20.0?

Issue #4732, opened 2026-07-17
pygmt.config() does not apply FONT_ANNOT_PRIMARY when called before Figure() on PyGMT 0.18.0 / GMT 6.6.0


jpatiani:
## Description

I found an unexpected behavior of `pygmt.config()` when changing GMT font settings.

The configuration works if `pygmt.config()` is called **after** creating a `Figure`, but it is ignored if called **before** `Figure()`.

This behavior differs from what I expected and from GMT command-line behavior.

---

## Environment

```text
PyGMT v0.18.0
GMT 6.6.0

Python 3.14.3
macOS 26.5.1 (Apple Silicon)

PyGMT information:
  version: v0.18.0

GMT library information:
  version: 6.6.0
```

---

### Case 1 (configuration ignored)

```python
import pygmt

with pygmt.config(FONT_ANNOT_PRIMARY="32p,Courier,blue"):
    fig = pygmt.Figure()

    fig.basemap(
        region=[0,1,0,1],
        projection="X10c",
        frame="af",
    )

    fig.show()
```

Result:

- annotation color remains black
- annotation font remains default

---

### Case 2 (works)

```python
import pygmt

fig = pygmt.Figure()

with pygmt.config(FONT_ANNOT_PRIMARY="32p,Courier,blue"):

    fig.basemap(
        region=[0,1,0,1],
        projection="X10c",
        frame="af",
    )

    fig.show()
```

Result:

- annotation becomes blue
- font changes correctly

---

## Verification

Inside the first example,

```python
with pygmt.config(FONT_ANNOT_PRIMARY="32p,Courier,blue"):
    import subprocess
    subprocess.run(["gmt", "get", "FONT_ANNOT_PRIMARY"])
```

prints

```
32p,Courier,blue
```

so the configuration appears to be set.

---

## GMT command line

The equivalent GMT command works correctly.

```bash
gmt begin test pdf
gmt set FONT_ANNOT_PRIMARY 32p,Courier,blue
gmt basemap -R0/1/0/1 -JX10c -Baf
gmt end show
```

The annotations are blue as expected.

---

## Expected behavior

I expected both examples to produce identical figures because
`pygmt.config()` should configure the GMT session regardless of whether
`Figure()` has already been created.

---

## Actual behavior

The configuration only takes effect if `Figure()` is created before entering the `pygmt.config()` context.

---

## Question

Is this intended behavior due to how GMT modern-mode sessions are initialized, or is this a PyGMT bug?

[...]

Issue #3041, opened 2024-02-06
Colorbar annotation size does not change


bjmarfito:
### Description of the problem

I set the FONT_ANNOT_PRIMARY and FONT_LABEL to 20 but the annotation on the colorbar does not change

### Minimal Complete Verifiable Example

```python
import pygmt

pygmt.config(FONT_LABEL = 20)
pygmt.config(FONT_ANNOT_PRIMARY = 20)

fig = pygmt.Figure()
fig.colorbar(position="x+3.68i/0.38i+w1.9i/0.090i+ef+mul+h", frame="a100f50", cmap=cptFileSlip)
fig.show()
```

### System information

```bash
PyGMT information:
  version: v0.10.0
System information:
  python: 3.11.7 (main, Dec 10 2023, 14:13:07) [Clang 15.0.0 (clang-1500.0.40.1)]
  executable: /opt/local/bin/python
  machine: macOS-14.3-arm64-arm-64bit
Dependency information:
  numpy: 1.26.2
  pandas: 2.1.1
  xarray: 2022.12.0
  netCDF4: 1.6.5
  packaging: 23.2
  contextily: None
  geopandas: None
  IPython: 8.14.0
  rioxarray: None
  ghostscript: 10.02.1
GMT library information:
  binary version: 6.5.0
  cores: 10
  grid layout: rows
  image layout: 
  library path: /opt/local/lib/gmt6/lib/libgmt.dylib
  padding: 2
  plugin dir: /opt/local/lib/gmt6/lib/gmt/plugins
  share dir: /opt/local/lib/gmt6/share/gmt
  version: 6.5.0
```

weiji14:
Hmm, not very well documented at https://docs.generic-mapping-tools.org/6.5/gmt.conf.html#term-FONT_LABEL, but could you try adding `p` (for the unit 'point') and see if it works? E.g.:

```python
with pygmt.config(FONT_LABEL="20p", FONT_ANNOT_PRIMARY="20p"):
    fig = pygmt.Figure()
    fig.colorbar(position="x+3.68i/0.38i+w1.9i/0.090i+ef+mul+h", frame="a100f50", cmap=cptFileSlip)
    fig.show()
```

bjmarfito:
Hi @weiji14, your proposed fix does not work on my side.

yvonnefroehlich:
Can you please try to see if it works when `config` is called after  initializing the `Figure` instance:

```python
import pygmt

size = 5

[...]

PR #4735, opened 2026-07-19
POC: DOC: A note recommending to call pygmt.config after pygmt.Figure()

yvonnefroehlich:

Related to several issues, e.g., #733, #4732.

Need to read about [**-H** of colorbar](https://docs.generic-mapping-tools.org/dev/colorbar.html#h) available for GMT >=6.7.0 (as mentioned in https://forum.generic-mapping-tools.org/t/psscale-colorbar-and-its-label-font-size/6472/4?u=yvonnefroehlich) to see if this can be helpful for PyGMT also and whether we should at an aliasa for **-H**; related to #3041 -> PR #4783.

**Preview**: https://pygmt-dev--4735.org.readthedocs.build/en/4735/api/generated/pygmt.config.html

**Guidelines**

**Slash Commands**


pygmt/src/config.py
```diff
@@ -24,6 +24,13 @@ class config:  # noqa: N801
 
     Full GMT defaults list at :gmt-docs:`gmt.conf.html`.
 
+    .. note::
+
+        Currently, :class:`pygmt.config` needs to be called **after** setting
```
seisman: Actually, the note only make sense when multiple Figure instances are created in one session.

pygmt/src/config.py
```diff
@@ -24,6 +24,14 @@ class config:  # noqa: N801
 
     Full GMT defaults list at :gmt-docs:`gmt.conf.html`.
 
+    .. note::
+
+        For GMT defaults to adjust the figure appearance, e.g., the font size,
+        :class:`pygmt.config` needs to be called **after** setting up the
```
seisman, line 30: Maybe just say "it's recommended to call `pygmt.config` after ..."

seisman: I feel the wording "global" and "local" are a little misleading. Perhaps we should call them "permanent" and "temporary" instead, similar to the wording we use for `Figure.shift_origin`.

We also need to document https://docs.generic-mapping-tools.org/dev/reference/features.html#gmt-modern-mode-hierarchical-levels, but it could be done in a separate PR.

pygmt/src/config.py
yvonnefroehlich, line 23: > I feel the wording "global" and "local" are a little misleading. Perhaps we should call them "permanent" and "temporary" instead

Meaning we need to update the docs (and likely some examples) make the formulations consistent?

seisman: Yes, the word "global" gives users the wrong impression that it has a global effect, but actually it only affects the configurations of the current figure, current subplot, or current inset (xref: https://docs.generic-mapping-tools.org/dev/reference/features.html#gmt-modern-mode-hierarchical-levels), depending on where `pygmt.config()` is called.

yvonnefroehlich: I agree, we need to document the behaviour more detailed and clearly.

Issue #4738, opened 2026-07-20
Migrate baseline images from DVC to Git LFS


seisman:
The baseline images in `pygmt/tests/baseline/*.png` have been tracked using DVC since PR #1036 (five years ago; the original idea came from https://github.com/GenericMappingTools/pygmt/issues/963#issuecomment-789277464). This setup has worked well, but it also has several drawbacks. I think it's time to consider whether we have a better solution.

## Drawbacks of DVC

- Requires installing `dvc`, which is unfamiliar to most contributors
- The workflow is relatively complicated: `dvc add`, `dvc push`, `git add`, and `git push` (xref: https://www.pygmt.org/v0.19.0/contributing.html#using-data-version-control-dvc-to-manage-test-images)
- Every maintainer needs to configure DVC locally
- Contributors must be added as collaborators on DagsHub before they can push DVC-tracked files
- Pulling DVC-tracked files now requires authentication (#4147)
- Several DVC-specific GitHub workflows are required to support the current setup

## Git LFS

[Git LFS](https://git-lfs.com/) is likely a better option for our use case. For files tracked by Git LFS, only small text pointer files are stored in the Git repository (similar to the current `*.dvc` files), while the actual file contents are stored separately. So updating baseline images won't increase the Git repository size.

Currently, PyGMT has ~240 baseline images with a total size of ~10 MB, which is well within GitHub LFS's free quota of 10 GB of storage and 10 GB of monthly bandwidth (xref: https://docs.github.com/en/billing/concepts/product-billing/git-lfs#free-use-of-git-lfs).

The setup is straightforward and only requires a few commands (see https://git-lfs.com/). Once configured, baseline images can be managed just like regular Git-tracked files using `git add`, `git pull`, and `git push`. This makes the workflow much more transparent and familiar for both maintainers and contributors.

Git LFS also integrates well with GitHub and pull requests. In my own fork, I have already migrated the baseline images to Git LFS on the main branch (https://github.com/seisman/pygmt). For example, if you visit https://github.com/seisman/pygmt/blob/main/pygmt/tests/baseline/test_basemap.png, GitHub displays the image directly, with a "Stored with Git LFS" note in the upper-left corner. When a baseline image is updated in a pull request, GitHub can also display image diffs directly (see https://github.com/seisman/pygmt/pull/4/changes for an example).

Overall, I think Git LFS satisfies our requirements while providing a simpler and more contributor-friendly workflow than DVC.

## TODO

[...]

PR #4743, opened 2026-07-22
DOC: Update the Contributing Guides after migration from DVC to Git LFS

seisman:
Address #4738

**Preview**: https://pygmt-dev--4743.org.readthedocs.build/en/4743/contributing.html#using-git-large-file-storage-lfs-to-manage-test-images

### DIFF
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -683,15 +683,16 @@ Git LFS, as explained in the next section.
 #### Using [Git Large File Storage (LFS)](https://git-lfs.com) to Manage Test Images
 
 As the baseline images are quite large blob files that can change often (e.g., with new
-GMT versions), they are managed with Git LFS rather than regular Git objects. Install
-Git LFS and enable it once on your machine:
+GMT versions), they are managed with Git LFS rather than regular Git objects.
+
+You need to run the following command once to set up Git LFS on your machine:
 
     git lfs install
 
-Once Git LFS is installed, `git clone` and `git pull` download the baseline images
-automatically. If LFS downloads were deliberately skipped, run `git lfs pull` to
-download the missing images. The images in `pygmt/tests/baseline` can then be used by
-the image-comparison tests as usual.
+After that, `git clone` and `git pull` download the baseline images automatically. If
+Git LFS downloads were deliberately skipped, run `git lfs pull` to download the missing
+images. The images in `pygmt/tests/baseline` can then be used by the image-comparison
+tests as usual.
 
 The entire workflow for generating or modifying baseline test images can be summarized
 as follows:
--- a/doc/contributing.md
+++ b/doc/contributing.md
@@ -198,7 +198,7 @@ In particular, these are some of the key development dependencies you will need
 to install to build the documentation and run the unit tests locally:
 
 - git (for cloning the repo and tracking changes in code)
-- dvc (for downloading baseline images used in tests)
+- [Git LFS](https://git-lfs.com) (for downloading baseline images used in tests)
 - pytest-mpl (for checking that generated plots match the baseline)
 - sphinx-gallery (for building the gallery example page)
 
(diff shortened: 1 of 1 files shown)


michaelgrund (approved):
Looks good!

doc/contributing.md
```diff
@@ -677,75 +677,37 @@ If it's correct, copy it (and only it) to `pygmt/tests/baseline`.
...
-before proceeding with the next steps.
-
-The entire workflow for generating or modifying baseline test images can be
-summarized as follows:
-
-    # Sync with both git and dvc remotes
+Don't forget to commit the baseline image as well! Baseline images are tracked with
+Git LFS, as explained in the next section.
+
+#### Using [Git LFS](https://git-lfs.com) to Manage Test Images
```
yvonnefroehlich: Maybe once explain the shorcut.

```suggestion
#### Using [Git Large File Storage (LFS)](https://git-lfs.com) to Manage Test Images
```

doc/contributing.md
```diff
@@ -677,75 +677,37 @@ If it's correct, copy it (and only it) to `pygmt/tests/baseline`.
...
-
-    # Sync with both git and dvc remotes
+Don't forget to commit the baseline image as well! Baseline images are tracked with
+Git LFS, as explained in the next section.
+
+#### Using [Git Large File Storage (LFS)](https://git-lfs.com) to Manage Test Images
+
+As the baseline images are quite large blob files that can change often (e.g., with new
+GMT versions), they are managed with Git LFS rather than regular Git objects. Install
+Git LFS and enable it once on your machine:
```
yvonnefroehlich: Would add a link to an installation guide to make clear that the command below does not install LFS.

```suggestion
GMT versions), they are managed with Git LFS rather than regular Git objects. [Install
Git LFS](https://github.com/git-lfs/git-lfs?utm_source=gitlfs_site&utm_medium=installation_link&utm_campaign=gitlfs#installing) and then use the command below to enable it once on your machine:
```

seisman: `git-lfs` is added to `environment.yml` (#4754) and will be available after setting up the environment, so I feel we can remove "Install Git LFS"

doc/contributing.md
```diff
@@ -677,75 +677,37 @@ If it's correct, copy it (and only it) to `pygmt/tests/baseline`.
...
+
+    git lfs install
```
seisman, line 690: The command `git lfs install` actually installs Git LFS configuration, so the instruction is a little misleading.

seisman: Done in beccd66993bf602c1fd8a9ac106831dfbe3ba8bb.

doc/contributing.md
yvonnefroehlich: Now I am confused 🙁.

Following PR #4754, I did run

> pip install git-lfs

within my PyGMT dev environment (Windows-Subsystem for Linux (WSL)). This should have been successful.

But when running 

> git lfs install

I still get the error

> git: 'lfs' is not a git command. See 'git --help'.

----

From reading the websites https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage and https://git-lfs.com/ I believe users need to externally install LFS (xref: https://stackoverflow.com/questions/48734119/git-lfs-is-not-a-git-command-unclear).

Beside this looking at the description on PyPI https://pypi.org/project/git-lfs/:

> A lightweight Git Large File Storage fetcher written in python.
>
> This module cannot fully replace the official git-lfs client, it only knows how to download the files, cache them (the same way the official client does), and place them in a checkout directory. Uploading files is not implemented at all.

Based on what I understand so far, uploading files is required in case we make changes to the baseline images. So, the functionality provided by `git-lfs` is not sufficient for our needs.

In case the others think everything looks correct, just go ahead with merging this PR. I will read through the links in more detail in the next days, trying to understand it better.

seisman: I believe the `git-lfs` on PyPI (https://pypi.org/project/git-lfs/, v1.6, maintaind by an individual) is not exactly the one on conda-forge (https://anaconda.org/channels/conda-forge/packages/git-lfs, v3.7.1, with source code at https://github.com/git-lfs/git-lfs ).

yvonnefroehlich: Oh no! You are right. Thanks! After installing the conda-forge one it works as expected 🙂.

yvonnefroehlich (approved):
I think now it looks good.

PR #4747, opened 2026-07-22
DOC: Improve the docstring for the region parameter

seisman:
`[xmin, xmax, ymin, ymax]` format is more commonly used.


yvonnefroehlich (approved):
We could similarly update the docstrings for `Figure.rose`, `Figure.ternary`, and `pygmt.dimfilter`.

PR #4750, opened 2026-07-22
Figure.scalebar: Remove the note that height only works for fancy scalebar

seisman:
`MAP_SCALE_HEIGHT` works for both plain and fancy scalebars, which can be verified below:
```
gmt begin map
gmt basemap -R0/20/0/10 -JM10c -Baf
gmt basemap -Lx1/1+w300k
gmt basemap -Lx3/1+w300k+f
gmt basemap -Lx5/1+w300k --MAP_SCALE_HEIGHT=1c
gmt basemap -Lx7/1+w300k+f --MAP_SCALE_HEIGHT=1c
gmt end show
```

map


yvonnefroehlich (approved):
Seems like this!

PR #4753, opened 2026-07-23
Figure.legend: Fix incorrect docstrings

seisman:


pygmt/src/legend.py
```diff
@@ -93,9 +93,11 @@ def legend(
         width. If height is given as percentage then height is recomputed as that
         fraction of the legend width (not plot height).
 
-        **Note:** Currently, the automatic height calculation only works when legend
```
seisman, line 96: Here is what the GMT documentation says (https://docs.generic-mapping-tools.org/dev/legend.html):

> If +w is not given then we compute the width within the Postscript code. Currently, this is only possible if just legend codes D, H, L, S, or V are used and that the number of symbol columns (N) is 1. If height is zero or not given then we estimate height based the expected vertical extent of the items to be placed.

examples/tutorials/advanced/legends.py
```diff
@@ -21,12 +21,11 @@
 # :meth:`pygmt.Figure.vlines` as well as :meth:`pygmt.Figure.histogram`. Therefore the
 # ``label`` parameter has to be specified to state the desired text for the legend entry
 # (white spaces are supported). Here, we use :meth:`pygmt.Figure.plot`, exemplary. By
-# default, the legend is placed in the Upper Right corner with an offset of 0.1
```
seisman, line 24: The default offset is `0.2 c` not `0.1 cm`. I feel it's better not to mention such details in a tutorial.

pygmt/src/legend.py
```diff
@@ -93,9 +93,11 @@ def legend(
...
+        **Note:** Currently, the automatic width calculation only works when legend
         codes **D**, **H**, **L**, **S**, or **V** are used and that the number of
-        symbol columns (**N**) is 1.
+        symbol columns (**N**) is 1. If height is zero or is not given, GMT estimates it
+        based the expected vertical extent of the items to be placed. Legends containing
+        paragraph text may require an explicit height.
```
yvonnefroehlich:
```suggestion
        symbol columns (**N**) is 1. If height is zero or not given, GMT estimates it
        based on the expected vertical extent of the items to be placed. A legend
        containing paragraph text may require an explicit height.
```

yvonnefroehlich (approved):

examples/tutorials/advanced/legends.py
yvonnefroehlich: Yes, as the defaults may change over time, and we often miss to update all occurrences, especially in the examples.

PR #4756, opened 2026-07-23
Figure: Defer figure activation until plotting arguments are prepared

seisman:
Move `self._activate_figure()` later in plotting wrappers, immediately before the first GMT session call.

This lets PyGMT perform pure-Python operations and validations before activating a figure or changing GMT state. Invalid input now fails earlier without creating or modifying the figure state.

`Figure.hlines` and `Figure.vlines` retain earlier activation because they may need to retrieve the current plot region before delegating to `Figure.plot`.


weiji14 (approved):
Looks reasonable, just need to be careful with new wrappers going forward.

PR #4772, opened 2026-07-28
Ruff: Enable RUF106 (rule-codes-in-suppression-comments)

seisman:
xref: https://docs.astral.sh/ruff/rules/rule-codes-in-suppression-comments/

Rule names are more readable than rule codes.


yvonnefroehlich (approved):
> Rule names are more readable than rule codes.
Agree.

PR #4777, opened 2026-07-29
Adds better support for plotting text in 3D at different heights

### LINKED ISSUE #1393: Better support for plotting text in 3D at different heights

For 3D plots, it should be possible to plot text at different heights based on a 'z' column, see https://docs.generic-mapping-tools.org/6.2/text.html#z. The current implementation of `text` in PyGMT (see https://www.pygmt.org/v0.4.0/api/generated/pygmt.Figure.text.html) does support this, but users have to provide the xyz input into the `textfiles` parameter, and it is also not very discoverable for new users.

It would be nice to have a `z` parameter (in addition to `x` and `y`) to be clear that `text` does support multi-height plotting. Here's a checklist for anyone wanting to work on this 'enhancement'.

TODO:
1. Alias the [`-Z`](https://docs.generic-mapping-tools.org/6.2/text.html#z), call it `zvalue` (as in `plot3d`)?
2. Add the `z` parameter here:
    https://github.com/GenericMappingTools/pygmt/blob/945bbc57ffc89a7c29384df48a653295b3f1a435/pygmt/src/text.py#L48-L49
3. Modify the code here to mention that x/y/z triples are also possible (need to wait for #1121 though):
    https://github.com/GenericMappingTools/pygmt/blob/945bbc57ffc89a7c29384df48a653295b3f1a435/pygmt/src/text.py#L160-L170
4. Update the `text` API docstring to document the new behaviour.
5. (Optional) Update the tutorial at https://www.pygmt.org/v0.4.0/tutorials/text.html to mention that z-value inputs are supported too.

**Are you willing to help implement and maintain this feature?** Please wait for #1121 before starting work on this.

landryza:

Adds a `z` parameter to `Figure.text` to explicitly expose support for 3D text placement in perspective plots. Previously, users could not provide z coordinates through the PyGMT API even though the underlying GMT module supports 3D text input.
 
This PR:
 
- Adds the `z` parameter to `text.py`
- Aliases the GMT `-Z` option via `zvalue`
- Updates the docstring and parameter descriptions to document z-coordinate support
- Updates parameter validation and error messages to include `z`
Fixes #1393 

### Reviewer feedback requested
 
I would especially appreciate guidance on tests for this feature. I ran the existing tests to make sure nothing was broken with this implementation, but I'm struggling to create tests that ensure this was implemented correctly.

**Preview**:

N/A

### DIFF
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -251,6 +251,8 @@ def text(  # ruff: ignore[too-many-branches, too-many-statements]
     data = None
     if kind == "empty":
         data = {"x": x, "y": y}
+        if z is not None:
+            data["z"] = z
 
         for arg, flag, name in array_args:
             if is_nonstr_iter(arg):
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -249,7 +249,10 @@ def text(  # ruff: ignore[too-many-branches, too-many-statements]
     confdict = {}
     data = None
     if kind == "empty":
-        data = {"x": x, "y": y, "z": z}
+        data = {"x": x, "y": y}
+
+        if z is not None:
+            data["z"] = z
 
         for arg, flag, name in array_args:
             if is_nonstr_iter(arg):
--- a/pygmt/src/text.py
+++ b/pygmt/src/text.py
@@ -100,6 +100,7 @@ def text(  # ruff: ignore[too-many-branches, too-many-statements]
 
         * *x*: X coordinate or longitude
         * *y*: Y coordinate or latitude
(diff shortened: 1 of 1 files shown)


pygmt/src/text.py
```diff
@@ -248,6 +251,9 @@ def text(  # ruff: ignore[too-many-branches, too-many-statements]
     if kind == "empty":
         data = {"x": x, "y": y}
 
+        if z is not None:
+            data["z"] = z
```
seisman, line 255: Could you please ada a test to ensure it works?

PR #4783, opened 2026-07-30
WIP: Figure.colorbar: Add alias "no_font_scaling" for "-H"

yvonnefroehlich:

This PR aims to add an alias for the newly added [**-H**](https://docs.generic-mapping-tools.org/dev/colorbar.html#h) flag of `colorbar` of GMT 6.7.0. Currently it's `no_font_scaling`. Feel free to make suggestion for a suitable alias.

**TODO**
- [ ] Find alias name: `no_font_scaling` default `False`
- [ ] Add alias 
- [ ] Update note

**Related**: #4243

**Preview**: https://pygmt-dev--4783.org.readthedocs.build/en/4783/api/generated/pygmt.Figure.colorbar.html

**Example**

```python
import pygmt

size = 5
args_bmap = {
    "region": [-size, size] * 2,
    "projection": f"X{size * 2}c",
    "frame": True,
}
args_cb = {
    "cmap": "batlow",
    "position": "jMC+h+w1c",
    "frame": "x+ltitle for a very short cb",
}


fig = pygmt.Figure()

fig.basemap(**args_bmap)
fig.colorbar(**args_cb)

fig.shift_origin(xshift=f"+{size*2+1}c")

fig.basemap(**args_bmap)
fig.colorbar(no_font_scaling=True, **args_cb)   # H = True

fig.show()
```

**Guidelines**

**Slash Commands**


seisman:
maybe font_scaling with default  
to True

yvonnefroehlich:
> maybe font_scaling with default to True

Hmmm. Seems we will have fun with double negatives here 😅. My understanding and thoughts so far are:
In GMT, not using **-H** means the auto-scaling applies. Setting **-H** turns off the auto-scaling. Setting a flag in GMT without any argument translates in PyGMT to ``True``. So ``H=True`` leads to "no auto-scaling is applied". Based on this, the alias for ``H`` should contain the "no", with the default ``False`` as then _no_ no auto-scaling (i.e., auto-scaling) is applied 😆.

seisman:
we can use `not font_scaling` internally

seisman:
> > maybe font_scaling with default to True
> 
> Hmmm. Seems we will have fun with double negatives here 😅. My understanding and thoughts so far are: In GMT, not using **-H** means the auto-scaling applies. Setting **-H** turns off the auto-scaling. Setting a flag in GMT without any argument translates in PyGMT to `True`. So `H=True` leads to "no auto-scaling is applied". Based on this, the alias for `H` should contain the "no", with the default `False` as then _no_ no auto-scaling (i.e., auto-scaling) is applied 😆.

After thinking it twice, I agree that `font_scaling=True` as the default is confusing. So let's use `no_font_scaling`.

pygmt/src/colorbar.py
```diff
@@ -463,6 +467,10 @@ def colorbar(
 
         If not specified, GMT uses its default encoding behavior, and the default dpi
         is 600 if the colorbar is drawn as image.
+    no_font_scaling
+        For GMT >=6.5.0, the fontsizes used for the colorbar are scaled automaticly
+        based on the length of the colorbar. For GMT >= 6.7.0, use this parameter to
+        turn of the this automatic fontsize scaling.
```
seisman: Since PyGMT requires GMT>=6.5.0, I feel we don't need to mention "GMT>=6.5.0" here.
```suggestion
    no_font_scaling
        By default, the font sizes used for the colorbar are scaled automatically
        based on the length of the colorbar. Set this parameter to ``True`` to 
        turn off the automatic scaling [Requires GMT>=6.7.0].
```

pygmt/src/colorbar.py
```diff
@@ -463,6 +467,10 @@ def colorbar(
...
     $projection
     $region
     $verbose
```
seisman, line 476: The `-H` option only exists in GMT>=6.7.0, so need to check the current GMT version, and raise an exception if GMT 6.5/6.6 is used and `no_font_scaling=True`.

pygmt/src/colorbar.py
```diff
@@ -296,20 +297,23 @@ def colorbar(
     .. note::
        For GMT >=6.5.0, the fontsizes of the colorbar x-label, x-annotations,
        and y-label are scaled based on the width of the colorbar following
```
seisman:
```suggestion
      By default, the fontsizes of the colorbar annotations, label, and unit
      are scaled based on the width of the colorbar following
```

pygmt/src/colorbar.py
```diff
@@ -296,20 +297,23 @@ def colorbar(
...
-       :math:`\sqrt{colorbar\_width / 15}`. To set a desired fontsize via the
+       :math:`\sqrt{colorbar\_width / 15}`. To set a desired fontsize use the
        GMT default parameters :gmt-term:`FONT_ANNOT_PRIMARY`,
        :gmt-term:`FONT_ANNOT_SECONDARY`, and :gmt-term:`FONT_LABEL` (or jointly
-       :gmt-term:`FONT`) users have to divide the desired fontsize by the value
-       calculated with the formula given above before passing it to the default
-       parameters. To only affect fontsizes related to the colorbar, the
-       defaults can be changed locally only using ``with pygmt.config(...):``.
+       :gmt-term:`FONT`)
+       For GMT >=6.7.0, users should use ``no_font_scaling = True`` to turn of
```
seisman:
```suggestion
       For GMT >=6.7.0, users should use ``no_font_scaling = True`` to turn off
```

seisman:
> > > maybe font_scaling with default to True
> > 
> > 
> > Hmmm. Seems we will have fun with double negatives here 😅. My understanding and thoughts so far are: In GMT, not using **-H** means the auto-scaling applies. Setting **-H** turns off the auto-scaling. Setting a flag in GMT without any argument translates in PyGMT to `True`. So `H=True` leads to "no auto-scaling is applied". Based on this, the alias for `H` should contain the "no", with the default `False` as then _no_ no auto-scaling (i.e., auto-scaling) is applied 😆.
> 
> After thinking it twice, I agree that `font_scaling=True` as the default is confusing. So let's use `no_font_scaling`.

Actually, I wonder if we should have the default to be no font scaling. From what I hear, the font scaling feature has confused users for a few years, and people may prefer no font scaling as the default. So, what about having 

- `font_scaling=False`: The default. Translated to `-H`.
- `font_scaling=True`: If users want such a feature. No `-H`.

I know it's inconsistent with the default GMT behavior, but I wonder if no font scaling is more frequently used.

seisman:
> > > > maybe font_scaling with default to True
> > > 
> > > 
> > > Hmmm. Seems we will have fun with double negatives here 😅. My understanding and thoughts so far are: In GMT, not using **-H** means the auto-scaling applies. Setting **-H** turns off the auto-scaling. Setting a flag in GMT without any argument translates in PyGMT to `True`. So `H=True` leads to "no auto-scaling is applied". Based on this, the alias for `H` should contain the "no", with the default `False` as then _no_ no auto-scaling (i.e., auto-scaling) is applied 😆.
> > 
> > 
> > After thinking it twice, I agree that `font_scaling=True` as the default is confusing. So let's use `no_font_scaling`.
> 
> Actually, I wonder if we should have the default to be no font scaling. From what I hear, the font scaling feature has confused users for a few years, and people may prefer no font scaling as the default. So, what about having
> 
> * `font_scaling=False`: The default. Translated to `-H`.
> * `font_scaling=True`: If users want such a feature. No `-H`.
> 
> I know it's inconsistent with the default GMT behavior, but I wonder if no font scaling is more frequently used.

Likely a bad idea, because `-H` requires GMT>=6.7.0, thus people with different GMT versions will see different default behavior. Let's use `no_font_scaling=False` as default.

pygmt/src/colorbar.py
```diff
@@ -294,22 +295,25 @@ def colorbar(
...
-       :gmt-term:`FONT`) users have to divide the desired fontsize by the value
-       calculated with the formula given above before passing it to the default
-       parameters. To only affect fontsizes related to the colorbar, the
-       defaults can be changed locally only using ``with pygmt.config(...):``.
+       :gmt-term:`FONT`)
+       For GMT >=6.7.0, users should use ``no_font_scaling = True`` to turn off
+       the automatic fontsize scaling.
+       For GMT 6.5.0 - 6.6.0, users have to divide the desired fontsize by the
+       value calculated with the formula given above before passing it to the
+       default parameters.
```
seisman, line 308: I feel this note can be moved to the docstrings for `no_font_scaling`.

PR #4787, opened 2026-07-31
Change 'import importlib' to 'import importlib.util' in test files.

willschlitzer:
I ran the ty typechecker and it raised `warning[possibly-missing-submodule]: Submodule `util` might not have been imported` for `import importlib`. This adds util to the import statement in several test files.

I was surprised since there wasn't an error when running the tests, and I figured util is a sub-module for importlib. Digging into this, it sounds like util needs its own import since Python 3.10, and this may have gone unnoticed because it was imported elsewhere when tests were running (I don't understand why that would have worked). Python's [documentation](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) includes an example of explicitly importing importlib.util

**Preview**:

**Guidelines**

**Slash Commands**

### DIFF
--- a/pygmt/tests/test_clib_vectors_to_arrays.py
+++ b/pygmt/tests/test_clib_vectors_to_arrays.py
@@ -3,7 +3,7 @@ Test the vectors_to_arrays function in the clib.conversion module.
 """
 
 import datetime
-import importlib
+import importlib.util
 
 import numpy as np
 import numpy.testing as npt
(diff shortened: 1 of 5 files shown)


seisman (approved):

seisman:
Looks good. Great to know the [ty](https://github.com/astral-sh/ty) typechecker. Will see if we want to adopt it in the future.

PR #4793, opened 2026-08-03
README: Reorganize the Contributing section

seisman:
Our "Imposter syndrome disclaimer" was adapted from the MetPy project. After reviewing the MetPy README (https://github.com/Unidata/MetPy#code-of-conduct), I feel that their structure is more appropriate.

1. Previously, "Code of Conduct" was a subsection under "Contributing". However, the Code of Conduct applies not only to contributing but also to other project-related activities, such as asking questions on forums. Therefore, this PR promotes "Code of Conduct" from a level-3 heading to a level-2 heading.

2. Currently, we have two separate subsections: "Contributing Guidelines" and "Imposter syndrome disclaimer". Following the MetPy style, this PR combines these two subsections into a single section.

Our "Imposter syndrome disclaimer" is now different from the MetPy one, I wonder if we should just update ours to match the MetPy one. If yes, then it should be done in a separate PR.

**Preview**: https://github.com/GenericMappingTools/pygmt/tree/readme/coc


yvonnefroehlich (approved):
Looks good!

PR #4807, opened 2026-08-09
DOC: List all dependencies in the installation guides

seisman:
Similar to the "Dependencies" section in https://github.com/GenericMappingTools/pygmt#dependencies.

**Preview**: https://pygmt-dev--4807.org.readthedocs.build/en/4807/install.html#dependencies

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -33,8 +33,9 @@ import pygmt
 pygmt.show_versions()
 ```
 
-You are now ready to make your first figure! Start by looking at our [Intro](intro/index.rst),
-[Tutorials](tutorials/index.rst), and [Gallery](gallery/index.rst). Good luck!
+You are now ready to make your first figure! Start by looking at our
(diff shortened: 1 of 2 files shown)


doc/install.md
```diff
@@ -61,26 +62,25 @@ Compiled conda packages of GMT for Linux, macOS and Windows are provided through
 [build GMT from source](https://github.com/GenericMappingTools/gmt/blob/master/BUILDING.md)
 instead.
 
-We recommend following the instructions further on to install GMT 6.
+We recommend following the instructions further on to install GMT.
 
 ## Dependencies
```
yvonnefroehlich, line 67: Maybe move the "Dpendencies" section to the top, similar as in the README? To let users know which packages will be installed together with PyGMT.

seisman: I prefer to keep the "Quickstart" section as the first section, so that users can install PyGMT without knowing any details.

README is mainly for describing the project, so I think in that file, it makes more sense to have "Dependencies" before "QuickStart".

PR #4805, opened 2026-08-08
DOC/website: Use card layout for presentations

yvonnefroehlich:
**Description of proposed changes**

I am wondering whether the newly added presentation page of the website can also benefit from using cards. Before coverting the remaining presentations it would be great to hear your impressions and adjustments.

| [Currently](https://www.pygmt.org/dev/presentations.html) | [This PR](https://pygmt-dev--4805.org.readthedocs.build/en/4805/presentations.html) |
| --- | --- |
| image |  image |

Maybe we could also use buttons for "Slides", "Poster", "Recording", etc. Therefore it would be great to have buttons placed next to each other. Maybe via nested grids. Or [badges](https://sphinx-design.readthedocs.io/en/rtd-theme/badges_buttons.html), which seem to be easier to arrange, but appear to be smaller than buttons. Not happy so far; but too late for today, will continue the next day(s).

|  |  |  |  |
| --- | --- | ---  | --- |
| image | image | image | image |

--- 

**Preview**: https://pygmt-dev--4805.org.readthedocs.build/en/4805/presentations.html

**Guidelines**

- [General Guidelines for Pull Request](https://www.pygmt.org/dev/contributing.html#general-guidelines-for-making-a-pull-request-pr)
- [Guidelines for Contributing Documentation](https://www.pygmt.org/dev/contributing.html#contributing-documentation)
- [Guidelines for Contributing Code](https://www.pygmt.org/dev/contributing.html#contributing-code)

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #4816, opened 2026-08-11
Figure.tilemap/pygmt.datasets.tile_map: Set User-Agent to avoid blocked access to OSM tile server

### LINKED ISSUE #4815: Access blocked - App is not following OpenStreetMap tile usage policy

Hi everyone,
I just rerun a previously working pygmt code and now I cannot plot OpenStreetMap anymore with tilemap. The tiles are now showing an access blocked form OpenStreetMap (see added figure). I checked it with both PyGMT version 0.17.0 and the newest version 0.19.0. Also I tried to rerun the tile maps example (https://www.pygmt.org/v0.19.0/gallery/maps/tilemaps.html). And I still get the same message as for my own code. If I use alternatively a Cartopy script to plot OpenStreetMap it works, so it seems to be an PyGMT issue. 
Based on the linked webpage, PyGMT encounters a General Block: https://wiki.openstreetmap.org/wiki/Blocked_tiles
I would be happy if you could check if this is also happening for you. Thank you!
Greetings
Sarah

Image

. 

```python
"""
Tile maps
=========

The :meth:`pygmt.Figure.tilemap` method allows to plot
tiles from a tile server or local file as a basemap or overlay.
"""

# %%
import contextily
import pygmt
from pygmt.params import Axis

fig = pygmt.Figure()
fig.tilemap(
    region=[-157.84, -157.8, 21.255, 21.285],
    projection="M12c",
    # Set level of details (0-22)
    # Higher levels mean a zoom level closer to the Earth's
    # surface with more tiles covering a smaller
    # geographic area and thus more details and vice versa
    # Please note, not all zoom levels are always available
    zoom=14,
    # Use tiles from OpenStreetMap tile server
    source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
    frame=Axis(annot=True, tick=True, grid=True),
)

fig.show()
```

### Full error message

```python-traceback
Access blocked 403
App is not following the tile usage policy of OpenStreetMap's volunteer-run servers: osm.wiki/Blocked
```

### System information

```bash
PyGMT information:
  version: v0.19.0
System information:
  python: 3.14.6 | packaged by conda-forge | (main, Jul 24 2026, 16:09:24) [GCC 14.3.0]
  executable: ~/.conda/envs/pygmt/bin/python
  machine: Linux-6.12.0-160000.36-default-x86_64-with-glibc2.40
Dependency information:
  numpy: 2.5.2
  pandas: 3.0.5
  xarray: 2026.7.0
  packaging: 26.3
  contextily: 1.7.1
  geopandas: None
  IPython: None
  pyarrow: None
  rioxarray: 0.23.0
  gdal: 3.13.2
  ghostscript: 10.07.1
GMT library information:
  version: 6.7.0
  padding: 2
  share dir: ~/.conda/envs/pygmt/share/gmt
  plugin dir: ~/.conda/envs/pygmt/lib/gmt/plugins
  library path: ~/.conda/envs/pygmt/lib/libgmt.so
  cores: 16
  grid layout: rows
  image layout: 

Or:
PyGMT information:
  version: v0.17.0
System information:
  python: 3.13.9 | packaged by conda-forge | (main, Oct 16 2025, 10:31:39) [GCC 14.3.0]
  executable: ~/conda/envs/seispy_env/bin/python
  machine: Linux-6.12.0-160000.36-default-x86_64-with-glibc2.40
Dependency information:
  numpy: 2.3.5
  pandas: 2.2.3
  xarray: 2025.11.0
  packaging: 25.0
  contextily: 1.7.0
  geopandas: 1.1.1
  IPython: 9.7.0
  pyarrow: 21.0.0
  rioxarray: 0.20.0
  gdal: 3.10.2
  ghostscript: 10.06.0
GMT library information:
  version: 6.5.0
  padding: 2
  share dir: ~/conda/envs/seispy_env/share/gmt
  plugin dir: ~/conda/envs/seispy_env/lib/gmt/plugins
  library path: ~/conda/envs/seispy_env/lib/libgmt.so
  cores: 16
  grid layout: rows
  image layout: 
  binary version: 6.5.0
```

seisman:
Not sure why it worked before, but it turns out that now we need to set "User-Agent" when requesting tiles from the OSM tile server.

This PR sets the default User-Agent to something like `PyGMT/v0.19.0 (+https://www.pygmt.org)` and allows custom headers. 

Reference:

- https://wiki.openstreetmap.org/wiki/Blocked_tiles
- https://operations.osmfoundation.org/policies/tiles/

Closes https://github.com/GenericMappingTools/pygmt/issues/4815.

**Preview**: 

- Current main: https://www.pygmt.org/dev/gallery/maps/tilemaps.html
- This PR: https://pygmt-dev--4816.org.readthedocs.build/en/4816/gallery/maps/tilemaps.html
- https://pygmt-dev--4816.org.readthedocs.build/en/4816/api/generated/pygmt.Figure.tilemap.html
- https://pygmt-dev--4816.org.readthedocs.build/en/4816/api/generated/pygmt.datasets.load_tile_map.html

### DIFF
--- a/examples/gallery/maps/tilemaps.py
+++ b/examples/gallery/maps/tilemaps.py
@@ -2,8 +2,8 @@
 Tile maps
 =========
 
-The :meth:`pygmt.Figure.tilemap` method allows to plot
-tiles from a tile server or local file as a basemap or overlay.
+The :meth:`pygmt.Figure.tilemap` method allows to plot tiles from a tile server or
+local file as a basemap or overlay.
 """
 
 # %%
@@ -15,14 +15,13 @@ fig = pygmt.Figure()
 fig.tilemap(
     region=[-157.84, -157.8, 21.255, 21.285],
     projection="M12c",
-    # Set level of details (0-22)
-    # Higher levels mean a zoom level closer to the Earth's
-    # surface with more tiles covering a smaller
-    # geographic area and thus more details and vice versa
-    # Please note, not all zoom levels are always available
+    # Set level of details (0-22).
+    # Higher levels mean a zoom level closer to the Earth's surface with more tiles
+    # covering a smaller geographic area and thus more details and vice versa.
+    # Please note, not all zoom levels are always available.
     zoom=14,
     # Use tiles from OpenStreetMap tile server
-    source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
+    source="https://tile.openstreetmap.org/{z}/{x}/{y}.png",
     frame=Axis(annot=True, tick=True, grid=True),
 )
 
@@ -30,9 +29,9 @@ fig.show()
 
 # %%
 # It's also possible to use tiles provided via the
-# `contextily <https://github.com/geopandas/contextily>`__
-# library. See :doc:`Contextily providers <contextily:providers_deepdive>`
-# for a list of possible tilemap options.
+# `contextily <https://github.com/geopandas/contextily>`__ library. See
+# :doc:`Contextily providers <contextily:providers_deepdive>` for a list of possible
+# tilemap options.
 
 fig = pygmt.Figure()
 fig.tilemap(
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -107,7 +107,7 @@ def load_tile_map(
         HTTP headers to include with requests to the tile server. This can be useful
         for authentication or to set a custom User-Agent. When supported by
         ``contextily`` (>=1.7.0), PyGMT sets a default ``User-Agent`` header like
-        ``PyGMT/v0.19.0 (+https://www.pygmt.org)``.
+        ``PyGMT/vX.Y.Z (+https://www.pygmt.org)``.
 
         .. note::
            Requires ``contextily>=1.7.0``.
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -6,7 +6,9 @@ Function to load raster tile maps from XYZ tile providers, and load as
 from collections.abc import Sequence
 from typing import Literal
 
+from packaging.version import Version
 from pygmt._show_versions import __version__ as _pygmt_version
+from pygmt.exceptions import GMTParameterError
 
 try:
     import contextily
@@ -161,10 +163,6 @@ def load_tile_map(
         )
         raise ImportError(msg)
 
-    # Set default HTTP headers.
-    if headers is None:
-        headers = {"User-Agent": f"PyGMT/{_pygmt_version} (+https://www.pygmt.org)"}
-
     # Keyword arguments for contextily.bounds2img
     contextily_kwargs = {
         "zoom": zoom,
@@ -173,8 +171,20 @@ def load_tile_map(
         "wait": wait,
         "max_retries": max_retries,
         "zoom_adjust": zoom_adjust,
-        "headers": headers,
     }
+
+    # TODO(contextily>=1.7.0): Remove once contextily>=1.7.0 is required.
+    # The 'headers' parameter was added in contextily v1.7.0
+    if Version(contextily.__version__) < Version("1.7.0"):
+        if headers is not None:
+            raise GMTParameterError(
+                reason="The 'headers' parameter requires contextily>=1.7.0."
+            )
+    else:
+        # Set default HTTP headers.
+        if headers is None:
+            headers = {"User-Agent": f"PyGMT/{_pygmt_version} (+https://www.pygmt.org)"}
+        contextily_kwargs["headers"] = headers
 
     west, east, south, north = region
     image, extent = contextily.bounds2img(
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -108,6 +108,9 @@ def load_tile_map(
         for authentication or to set a custom User-Agent. By default, only passing
         PyGMT (with the version string and URL) as the User-Agent.
 
+        .. note::
+           The ``headers`` parameter requires ``contextily>=1.7.0``.
+
     Returns
     -------
     raster
--- a/pygmt/datasets/tile_map.py
+++ b/pygmt/datasets/tile_map.py
@@ -6,6 +6,8 @@ Function to load raster tile maps from XYZ tile providers, and load as
 from collections.abc import Sequence
 from typing import Literal
 
+from pygmt._show_versions import __version__ as _pygmt_version
+
 try:
     import contextily
     from rasterio.crs import CRS
@@ -40,6 +42,7 @@ def load_tile_map(
     wait: int = 0,
     max_retries: int = 2,
     zoom_adjust: int | None = None,
+    headers: dict[str, str] | None = None,
 ) -> xr.DataArray:
     """
     Load a georeferenced raster tile map from XYZ tile providers.
@@ -75,7 +78,7 @@ def load_tile_map(
           OpenStreetMap Humanitarian web tiles.
         - A web tile provider in the form of a URL. The placeholders for the XYZ in the
           URL need to be {x}, {y}, {z}, respectively. E.g.
-          ``https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png``.
+          ``https://tile.openstreetmap.org/{z}/{x}/{y}.png``.
         - A local file path. The file is read with :doc:`rasterio <rasterio:index>` and
           all bands are loaded into the basemap. See
           :doc:`contextily:working_with_local_files`.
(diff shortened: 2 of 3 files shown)


seisman:
Ping @GenericMappingTools/pygmt-maintainers for reviewing this PR.

**Related but slightly off-topic:**
Considering that the `Figure.tilemap` method is used in the PyGMT paper and it doesn't work now, I think we should make a new release, maybe v0.19.1?

yvonnefroehlich:
> Not sure why it worked before, but it turns out that now we need to set "User-Agent" when requesting tiles from the OSM tile server.

This seems to be a quite new issue: In PR #4809 merged on Sunday the tile maps gallery example looks good (https://pygmt-dev--4809.org.readthedocs.build/en/4809/gallery/maps/tilemaps.html), but in #4805 merged on Monday it fails and shows the reported issue (https://pygmt-dev--4805.org.readthedocs.build/en/4805/gallery/maps/tilemaps.html) .

```
$ git log
commit 1c6346bf2a55fd3db519d3ca2b066644dd6e750c (HEAD -> main, origin/main, origin/HEAD)
Author: Dongdong Tian <seisman.info@gmail.com>
Date:   Tue Aug 11 16:38:51 2026 +0800

    Update the URL links to the GMT documentation (#4814)

commit a0fd47979b1fd0c5dcea9e66f206b4a219ef3679 (gmt/url)
Author: Yvonne Fröhlich <94163266+yvonnefroehlich@users.noreply.github.com>
Date:   Mon Aug 10 11:29:55 2026 +0200

    DOC/website: Use card layout for presentations (#4805)

    Co-authored-by: Dongdong Tian <seisman.info@gmail.com>

commit 9505baa5de1dfa487dc53bc118f1a08770735fad
Author: Dongdong Tian <seisman.info@gmail.com>
Date:   Sun Aug 9 18:52:45 2026 +0800

    README: Use absolute URLs for PyGMT logos (#4809)
```

Looking at the wikipedia pages for openstreetmap there are quite recent updates, https://wiki.openstreetmap.org/w/index.php?title=Blocked_tiles&action=history and https://wiki.openstreetmap.org/w/index.php?title=Referer&action=history.

----

> This PR sets the default User-Agent to something like PyGMT/v0.19.0 (+https://www.pygmt.org) and allows custom headers.

With the changes in this PR the tile maps gallery example seems to work again, including when building the documentation locally.

I am wondering if our weekly Link-Checks on Sundays can catch this. Maybe we should add a note to the release checklist to manually check this gallery example shortly before making a new release.

---- 

> **Related but slightly off-topic:**
> Considering that the Figure.tilemap method is used in the PyGMT paper and it doesn't work now, I think we should make a new release, maybe v0.19.1?

I am afraid yes. This issue affects all PyGMT versions, i.e. using a lower version does not work as an (unelegant) workaround. Thus, the `Figure.tilemap` method is currently not working at all, and it's likely multiple users will face this issue (so far I see no reports on the forum). It's almost one and a half months until the end of September, which is the considered time range for the release of v0.20.0.

pygmt/datasets/tile_map.py
```diff
@@ -162,6 +173,7 @@ def load_tile_map(
         "wait": wait,
         "max_retries": max_retries,
         "zoom_adjust": zoom_adjust,
+        "headers": headers,
```
weiji14: Need to bump to `contextily>=1.7` in pyproject.toml, ref https://github.com/geopandas/contextily/pull/266 and https://github.com/geopandas/contextily/releases/tag/v1.7.0

weiji14 (commented):
Hmm, I've scanned through contextily's code/issues and found https://github.com/geopandas/contextily/issues/75, and it seems they should have set a user-agent header of some sort already since 2019? Maybe the `contextily-` header is getting rate-limited 😅 so ok to set a custom PyGMT one, but I hope this doesn't get rate-limited too (if so, we'll know PyGMT's fig.tilemap is getting used a lot 😆).

> > **Related but slightly off-topic:**
> > Considering that the Figure.tilemap method is used in the PyGMT paper and it doesn't work now, I think we should make a new release, maybe v0.19.1?
> 
> I am afraid yes. This issue affects all PyGMT versions, i.e. using a lower version does not work as an (unelegant) workaround. Thus, the `Figure.tilemap` method is currently not working at all, and it's likely multiple users will face this issue (so far I see no reports on the forum). It's almost one and a half months until the end of September, which is the considered time range for the release of v0.20.0.

It's only not working for OpenStreetMap tiles, but should be ok for other tilemap providers. That said, yeah a v0.19.1 would probably be good in this case.

pygmt/datasets/tile_map.py
```diff
@@ -154,6 +161,10 @@ def load_tile_map(
         )
         raise ImportError(msg)
 
+    # Set default HTTP headers.
+    if headers is None:
+        headers = {"User-Agent": f"PyGMT/{_pygmt_version} (+https://www.pygmt.org)"}
```
weiji14: Actually, should we append a unique uuid also similar to what `contextily` did in https://github.com/geopandas/contextily/pull/81/changes/f58cb550ce59ef96ed9b264cec192fe39fb5ab38?

pygmt/datasets/tile_map.py
seisman: contextily 1.7 was released in November 2025, less than a year ago. It would be good if we could still support old versions.

It seems that we only need to pass the `User-Agent` header when using OSM (need to verify it). Perhaps we can continue supporting older versions of contextily in general, and require `contextily>=1.7` only when OSM is used (i.e., checking if  `source` is openstreetmap)?

Edit: Done in 0473d497039ca0467c61f44b764a3aaec30a8715.

pygmt/datasets/tile_map.py
seisman: I feel it doens't follow the common format for User-Agent: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent

seisman:
> This issue affects all PyGMT versions, i.e. using a lower version does not work as an (unelegant) workaround.

Actually, a simple workaround is to use a different OSM tile server, e.g., the German one: `https://tile.openstreetmap.de/{z}/{x}/{y}.png`. In other words, perhaps we don't have to release v0.19.1?

yvonnefroehlich:
> > This issue affects all PyGMT versions, i.e. using a lower version does not work as an (unelegant) workaround.
> 
> Actually, a simple workaround is to use a different OSM tile server, e.g., the German one: `https://tile.openstreetmap.de/{z}/{x}/{y}.png`.

That's interesting. Did not check this yesterday night; only looked at tiles beside OSM and these seem to work.

examples/gallery/maps/tilemaps.py
```diff
@@ -15,24 +15,23 @@
...
-    # surface with more tiles covering a smaller
-    # geographic area and thus more details and vice versa
-    # Please note, not all zoom levels are always available
+    # Set level of details (0-22).
+    # Higher levels mean a zoom level closer to the Earth's surface with more tiles
+    # covering a smaller geographic area and thus more details and vice versa.
+    # Please note, not all zoom levels are always available.
     zoom=14,
     # Use tiles from OpenStreetMap tile server
-    source="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
```
seisman, line 25: No `{s}` according to https://operations.osmfoundation.org/policies/tiles/.

pygmt/datasets/tile_map.py
weiji14: I think it's good to set the `User-Agent` header even if using a non-openstreetmap source, the changes look good to me.

pygmt/datasets/tile_map.py
```diff
@@ -98,6 +103,14 @@ def load_tile_map(
     zoom_adjust
         The amount to adjust a chosen zoom level if it is chosen automatically. Values
         outside of -1 to 1 are not recommended as they can lead to slow execution.
+    headers
+        HTTP headers to include with requests to the tile server. This can be useful
+        for authentication or to set a custom User-Agent. When supported by
+        ``contextily`` (>=1.7.0), PyGMT sets a default ``User-Agent`` header like
+        ``PyGMT/v0.19.0 (+https://www.pygmt.org)``.
```
weiji14: Do we want to state `v0.19.0` here, or can we just use `PyGMT/vX.Y.Z (+https://...)`?

seisman: Fixed in e575bf2c2437e2ae8e01685d7727c4fc0a19afdc

weiji14 (approved):
Thanks @seisman, looks reasonable to me, maybe get @yvonnefroehlich to have another quick look too before merging.

yvonnefroehlich:
> In other words, perhaps we don't have to release v0.19.1?

This workaround is a nice trick!
As I am away from my laptop this morning, I did not check yet if this works for every region and if the tiles are exactly identical. Maybe partly there are some differences (font size, position, language) regarding the text labels for cities, etc. 
Waiting until September would allow us to account for similar issues that may come up in the next days? Not sure if the other OSM providers or other providers will have such blocking in the future also, but maybe setting a User-Agent works for them also to avoid blocking.

PR #4808, opened 2026-08-09
docs: update project description to reflect geospatial visualization and analysis focus

erensh27:
Updates the project description from "A Python interface for the Generic Mapping Tools" to "A Python interface to the Generic Mapping Tools for geospatial visualization and analysis" in:

- `pyproject.toml` (project metadata shown on PyPI/GitHub)
- `doc/index.md` (homepage tagline)

As proposed in the issue, this helps users unfamiliar with GMT understand PyGMT's purpose. The PyGMT paper title is "PyGMT: Bridging Python and the Generic Mapping Tools for Geospatial Visualization and Analysis"; the new description stays close to it without being identical.

Closes #4744

**Preview**:

- https://pygmt-dev--4808.org.readthedocs.build/en/4808/

PR #4771, opened 2026-07-27
Add blockmedian and surface gallery example

Aayush7788:
**Description of proposed changes**

Adds a gallery example showing how to reduce irregular bathymetry observations
with `pygmt.blockmedian` before interpolating them with `pygmt.surface`.

The example uses the built-in bathymetry sample dataset and compares the reduced
observations with the resulting grid.

Fixes #1352

**Preview**: https://pygmt-dev--4771.org.readthedocs.build/en/4771/gallery/images/gridding_with_surface.html

**Guidelines**

- [General Guidelines for Pull Request](https://www.pygmt.org/dev/contributing.html#general-guidelines-for-making-a-pull-request-pr)
- [Guidelines for Contributing Documentation](https://www.pygmt.org/dev/contributing.html#contributing-documentation)
- [Guidelines for Contributing Code](https://www.pygmt.org/dev/contributing.html#contributing-code)

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #4820, opened 2026-08-12
CI: Fix the 'Greet First-Time Contributors' workflow

seisman:
The "Greet First-Time Contributors" workflows works well for issues (https://github.com/GenericMappingTools/pygmt/issues/4815#issuecomment-5251260288), but not for pull requests (e.g., https://github.com/GenericMappingTools/pygmt/pull/4808, https://github.com/GenericMappingTools/pygmt/actions/runs/31300559514/job/93216830178?pr=4808).

This PR fixes the issue by changing the triggering event from `pull_request` to `pull_request_target`.

I didn't test it. Will know if it works or not when merging #4771.

### DIFF
--- a/.github/workflows/welcome.yml
+++ b/.github/workflows/welcome.yml
@@ -5,7 +5,7 @@
 name: Greet First-Time Contributors
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
   issues:
@@ -21,7 +21,7 @@ permissions:
 jobs:
   greeting:
     name: Greet First-Time Contributors
-    if: github.event_name == 'issues' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository)
+    if: github.event_name == 'issues' || (github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository)
     runs-on: ubuntu-slim
 
     steps:


.github/workflows/welcome.yml
```diff
@@ -5,7 +5,7 @@
 name: Greet First-Time Contributors
 
 on:
-  pull_request:
+  pull_request_target:
```
seisman, line 8: zizmor marks `pull_request_target` as a dangerous trigger:

```
    --> .github/workflows/welcome.yml:7:1
     |
   7 | / on:
   8 | |   pull_request_target:
   9 | |     types:
  10 | |       - opened
  11 | |   issues:
  12 | |     types:
  13 | |       - opened
     | |______________^ pull_request_target is almost always used insecurely
     |
     = note: audit confidence → Medium
     = help: audit documentation → https://docs.zizmor.sh/audits/#dangerous-triggers
```
but maybe it's OK since the workflow only contains one action?

weiji14: Let's not risk it, especially for PRs. Too hard to secure against random actors nowadays.

I think first time contributors opening a PR get a pop up on the GitHub UI that points to contributing.md so we could just put this info there?

seisman: Do you mean adding a new section to `CONTRIBUTING.md`?

weiji14: This is how it looks like if I visit a repo with a CONTRIBUTING.md that I haven't made a PR for before:

image

So yeah, we could just transfer what's under the `pr_message: |` line to the CONTRIBUTING.md file. I think most of the content overlaps already, so maybe just need a few extra sentences.

seisman:
OK. Closing the PR.

Issue #4853, opened 2026-08-24
Figure.sac: Plot seismograms in SAC format


Chuan1937:
*This issue serves as the central place for discussing and tracking the implementation of the `sac` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/supplements/seis/sac.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/sac
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.Figure.sac.html

## GMT Option Flags and Modifiers

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [x] `-B`: `frame`
- [x] `-J`: `projection`
- [x] `-R`: `region`
- [ ] `-C`: `time_window`
- [ ] `-D`: `offset`
- [ ] `-E`: `profile`
- [ ] `-F`: `preprocess`
- [ ] `-G`: `fill`
- [ ] `-M`: `amplitude_scale`
- [ ] `-Q`: `vertical`
- [ ] `-S`: `time_scale`
- [ ] `-T`: `reduction_velocity`, `time_shift`, `time_reference`
- [x] `-W`: `pen`
- [x] `-V`: `verbose`
- ~~`-X`/`-Y`~~: Use `Figure.shift_origin` instead.
- ~~`--PAR=value`~~: Use `pygmt.config` instead.

## Related GMT configurations

*List any related GMT configurations that may affect the behavior.*

## Notes on Input Formats

The GMT `sac` module accepts SAC waveform files as input. Since ObsPy is widely used for seismic waveform processing in Python, support for ObsPy `Trace`/`Stream` objects could also be considered.

For example:

```python
from obspy import read

trace = read("example.sac")[0]

fig.sac(spec=trace)
```
## Linked Pull Requests

- [x] Initial feature implementation(#4854)
- [ ] Add a tutorial or gallery example

## Related Issues and Discussions

*Add links to related wrapper discussions, API design threads, or upstream GMT changes here.*

seisman:
There were some discussions in https://github.com/GenericMappingTools/pygmt/issues/967.

seisman:
[...]

PR #4854, opened 2026-08-24
Figure.sac: Plot seismograms in SAC format

Chuan1937:
This pull request wraps the module `sac`.

**Preview** at https://pygmt-dev--4854.org.readthedocs.build/en/4854/api/generated/pygmt.Figure.sac.html

Related to https://github.com/GenericMappingTools/pygmt/issues/4853.


pygmt/src/sac.py
```diff
@@ -0,0 +1,81 @@
...
+from pygmt.alias import Alias, AliasSystem
+from pygmt.clib import Session
+from pygmt.helpers import build_arg_list, fmt_docstring
+from pygmt.params import Axis, Frame
+
+
+@fmt_docstring
+def sac(
+    self,
+    spec: PathLike | str | Sequence[PathLike | str],
```
seisman: Maybe just use `data`?
```suggestion
    data: PathLike | str | Sequence[PathLike | str],
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,81 @@
...
+    region: Sequence[float | str] | str | None = None,
+    frame: Frame | Axis | Literal["none"] | str | Sequence[str] | bool = False,
+    verbose: Literal["quiet", "error", "warning", "timing", "info", "compat", "debug"]
+    | bool = False,
+    panel: int | Sequence[int] | bool = False,
+    perspective: float | Sequence[float] | str | bool = False,
+    transparency: float | None = None,
+    **kwargs,
+):
+    r"""
```
seisman:
```suggestion
    """
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,81 @@
...
+       - V = verbose
+       - W = pen
+       - c = panel
+       - p = perspective
+       - t = transparency
+
+    Parameters
+    ----------
+    spec : str, pathlib.Path, or a sequence of them
+        The SAC waveform file(s) to plot.
```
seisman:
```suggestion
    data
        The SAC waveform file(s) to plot.
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,81 @@
...
+    pen : str
```
seisman:
```suggestion
    pen
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,208 @@
...
+        multiple times, and the order controls the processing order, e.g.,
+        ``"rii"`` converts acceleration to displacement.
+    fill
+        Paint the positive or negative portion of the traces. Use ``p``/``n`` to
+        paint the positive/negative portion [Default paints the positive
+        portion], ``+g`` *fill* to set the fill color [Default is ``"black"``],
+        ``+z`` *zero* to define the zero line, and ``+t`` *t0*/*t1* to paint a
+        time window only. Can be repeated to paint the positive and negative
+        portions separately, e.g., ``["p+gblack", "n+gred"]``. Set to ``True``
+        to paint the positive portion with the default fill.
```
seisman: This parameter is not Pythonic yet.

pygmt/src/sac.py
```diff
@@ -0,0 +1,208 @@
...
+    offset
+        Offset the seismogram positions by the given amounts *dx*[/ *dy*]
+        [Default is no offset]. If *dy* is not given, it is set equal to *dx*.
+    profile
+        Choose the profile type, i.e., the type of the y axis. Use ``a`` for
+        azimuth, ``b`` for back-azimuth, ``k`` for epicentral distance in km,
+        ``d`` for epicentral distance in degrees, ``n`` for trace number (the
+        first trace is numbered *n*, e.g., ``n0``), and ``u`` for user-defined
+        profile (the y positions are determined by the SAC header variable
+        ``usern``, e.g., ``u0``).
```
seisman: Maybe support more readable arguments like: `azimuth`, `back_azimuth`, `distance_in_km`, `distance_in_degree`, `number`, `user0`, ..., `user9`?

yvonnefroehlich: Maybe we can check which names ObsPy is using and use the same names here.

seisman: These single letters correspond to the following SAC header variables:

- `a`: `az`
- `b`: `baz`
- `k`: `dist` distance in km
- `d`: `gcarc` great-circle distance in degrees
- `n`: just trace number
- `u0`: `user0`
- `u9`: `user9`

pygmt/src/sac.py
```diff
@@ -0,0 +1,281 @@
...
+    data
+        The SAC waveform file(s) to plot.
+    pen
+        Set pen attributes for all traces [Default is ``"0.25p,black,solid"``].
+    time_window
+        Read and plot seismograms in the time window between *t0* and *t1* only,
+        where *t0* and *t1* are relative to the reference time (see
+        ``time_reference``). If no reference time is set, the reference time in
+        the SAC header is used. Set to ``True`` to determine *t0*/*t1* from the
+        *xmin*/*xmax* of the ``region`` parameter.
```
seisman, line 122:
```suggestion
    time_window
        (*t0*, *t1*).
        Read and plot seismograms in the time window between *t0* and *t1* only,
        where *t0* and *t1* are relative to the reference time (see
        ``time_reference``). If no reference time is set, the reference time in
        the SAC header is used. Set to ``True`` to determine *t0* and *t1* from the
        *xmin*/*xmax* of the ``region`` parameter.
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,282 @@
...
+        "user3",
+        "user4",
+        "user5",
+        "user6",
+        "user7",
+        "user8",
+        "user9",
+    ]
+    | None = None,
+    trace_number: int | None = None,
```
seisman: Maybe let `trace_number` default to `0`?
```suggestion
    trace_number: int = 0,
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,282 @@
...
+            description="parameters 'profile' and 'trace_number'",
+            reason="Only one can be specified.",
+        )
+
+    profile_mapping = {
+        "azimuth": "a",
+        "back_azimuth": "b",
+        "distance_in_km": "k",
+        "distance_in_degree": "d",
+        "trace_number": "n",
```
seisman: `trace_number` default to `0`:
```suggestion
        "trace_number": f"n{trace_number}",
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,282 @@
...
+        **{f"user{number}": f"u{number}" for number in range(10)},
+    }
+    profile_alias = (
+        Alias(trace_number, name="trace_number", prefix="n")
+        if trace_number is not None
+        else Alias(profile, name="profile", mapping=profile_mapping)
+    )
```
seisman: Can be removed after `trace_number` default to `0`, right?
```suggestion
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,282 @@
...
+    $panel
+    $perspective
+    $transparency
+    """
+    if profile is not None and trace_number is not None:
+        raise GMTValueError(
+            [profile, trace_number],
+            description="parameters 'profile' and 'trace_number'",
+            reason="Only one can be specified.",
+        )
```
seisman: I don't think this check makes sense. Users can use `profile="trace_number", trace_number=5`, right?

pygmt/src/sac.py
```diff
@@ -0,0 +1,282 @@
...
+        raise GMTValueError(
+            [fill_zero, fill_time_window],
+            description="parameters 'fill_zero' and 'fill_time_window'",
+            reason="At least one of 'positive_fill' or 'negative_fill' must be specified.",
+        )
+
+    aliasdict = AliasSystem(
+        C=Alias(time_window, name="time_window", sep="/", size=2),
+        D=Alias(offset, name="offset", sep="/", size=(1, 2)),
+        E=profile_alias,
```
seisman:
```suggestion
        E=Alias(profile, name="profile", mapping=profile_mapping,
```

pygmt/src/sac.py
Chuan1937: ok

pygmt/src/sac.py
Chuan1937: yes

pygmt/src/sac.py
```diff
@@ -0,0 +1,277 @@
...
+    $panel
+    $perspective
+    $transparency
+    """
+    if profile is not None and trace_number is not None:
+        raise GMTValueError(
+            [profile, trace_number],
+            description="parameters 'profile' and 'trace_number'",
+            reason="Only one can be specified.",
+        )
```
seisman:
```suggestion
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,277 @@
...
+
+    profile_mapping = {
+        "azimuth": "a",
+        "back_azimuth": "b",
+        "distance_in_km": "k",
+        "distance_in_degree": "d",
+        "trace_number": f"n{trace_number}",
+        **{f"user{number}": f"u{number}" for number in range(10)},
+    }
```
seisman: Since `profile_mapping` is used only once, maybe move this mapping definition to line 232.

pygmt/src/sac.py
```diff
@@ -36,8 +36,7 @@ def sac(
         "user7",
         "user8",
         "user9",
-    ]
-    | None = None,
+    ] = "trace_number",
```
seisman: The change is incorrect. By default, seismograms are not plotted in profile mode, i.e., the y axis is the amplitude, not the profile. So we can't let `profile` default to anything.

Chuan1937: OK，I'll restore profile=None and update the test to explicitly use `profile=trace_number` with `trace_number=1`.

pygmt/src/sac.py
```diff
@@ -0,0 +1,270 @@
...
+    $transparency
+    """
+    profile_mapping = {
+        "azimuth": "a",
+        "back_azimuth": "b",
+        "distance_in_km": "k",
+        "distance_in_degree": "d",
+        "trace_number": f"n{trace_number}",
+        **{f"user{number}": f"u{number}" for number in range(10)},
+    }
```
seisman:
```suggestion
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,270 @@
...
+        raise GMTValueError(
+            [fill_zero, fill_time_window],
+            description="parameters 'fill_zero' and 'fill_time_window'",
+            reason="At least one of 'positive_fill' or 'negative_fill' must be specified.",
+        )
+
+    aliasdict = AliasSystem(
+        C=Alias(time_window, name="time_window", sep="/", size=2),
+        D=Alias(offset, name="offset", sep="/", size=(1, 2)),
+        E=Alias(profile, name="profile", mapping=profile_mapping),
```
seisman:
```suggestion
        E=Alias(profile, name="profile", mapping={
            "azimuth": "a",
            "back_azimuth": "b",
            "distance_in_km": "k",
            "distance_in_degree": "d",
            "trace_number": f"n{trace_number}",
            **{f"user{number}": f"u{number}" for number in range(10)},
        }),
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,270 @@
...
+        "user3",
+        "user4",
+        "user5",
+        "user6",
+        "user7",
+        "user8",
+        "user9",
+    ]
+    | None = None,
+    trace_number: int = 0,
```
seisman: Maybe rename it to `trace_number_start`, which is more readable.
```suggestion
    trace_number_start: int = 0,
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,273 @@
...
+    time_window
+        (*t0*, *t1*).
+        Read and plot seismograms in the time window between *t0* and *t1* only,
+        where *t0* and *t1* are relative to the reference time (see
+        ``time_reference``). If no reference time is set, the reference time in
+        the SAC header is used. Set to ``True`` to determine *t0* and *t1* from the
+        *xmin*/*xmax* of the ``region`` parameter.
+    offset
+        Offset the seismogram positions by the given amounts *dx*[/ *dy*]
+        [Default is no offset]. If *dy* is not given, it is set equal to *dx*.
```
seisman:
```suggestion
    offset
        *offset* or (*offset_x*, *offset_y*).
        Offset the seismogram positions by *offset_x* and *offset_y* [Default is no offset].
        If a single value *offset* is given, *offset_x* = *offset_y* = *offset*. 
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,273 @@
...
+    if fill_modifiers and not any(fill_options):
+        raise GMTValueError(
+            [fill_zero, fill_time_window],
+            description="parameters 'fill_zero' and 'fill_time_window'",
+            reason="At least one of 'positive_fill' or 'negative_fill' must be specified.",
+        )
+
+    aliasdict = AliasSystem(
+        C=Alias(time_window, name="time_window", sep="/", size=2),
+        D=Alias(offset, name="offset", sep="/", size=(1, 2)),
```
seisman: size=1 is internally supported, so no need to specify `size=1`
```suggestion
        D=Alias(offset, name="offset", sep="/", size=2),
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,273 @@
...
+        ``"user9"``. User-defined profiles use the corresponding SAC header
+        variable, e.g., ``"user0"`` uses ``user0``.
+    trace_number_start
+        Set the number of the first trace when ``profile="trace_number"``.
+        [Default is 0].
+    preprocess
+        Preprocess the data before plotting. Use ``i`` for integral, ``q`` for
+        square, and ``r`` for removing the mean value. The letters can repeat
+        multiple times, and the order controls the processing order, e.g.,
+        ``"rii"`` converts acceleration to displacement.
```
seisman: I wonder if it's more readable to have syntax like:
```
preprocess = ["integral", "square", "demean"]
```

Chuan1937: I agree.I will map `["integral", "square", "demean"]` to `iqr`.

pygmt/src/sac.py
```diff
@@ -0,0 +1,274 @@
...
+                    "t0": 0,
+                    "t1": 1,
+                    "t2": 2,
+                    "t3": 3,
+                    "t4": 4,
+                    "t5": 5,
+                    "t6": 6,
+                    "t7": 7,
+                    "t8": 8,
+                    "t9": 9,
```
seisman:
```suggestion
                    **{f"t{i}": str(i) for i in range(10)},
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,274 @@
...
+            Alias(time_shift, name="time_shift", prefix="+s"),
+            Alias(
+                time_reference,
+                name="time_reference",
+                prefix="+t",
+                mapping={
+                    "b": -5,
+                    "e": -4,
+                    "o": -3,
+                    "a": -2,
```
seisman:
```suggestion
                    "b": "-5",
                    "e": "-4",
                    "o": "-3",
                    "a": "-2",
```

pygmt/src/sac.py
```diff
@@ -0,0 +1,280 @@
...
+        for operation in preprocess:
+            value = Alias(
+                operation,
+                name="preprocess",
+                mapping={"integral": "i", "square": "q", "demean": "r"},
+            )._value
+            if not isinstance(value, str):
+                raise GMTValueError(value, description="preprocess operation")
+            preprocess_values.append(value)
+        preprocess_value = "".join(preprocess_values)
```
seisman, line 201: Please move this code block into a private function `_alias_option_F` and add some doctests for it. You can use `subplot.py` as a reference.

PR #4856, opened 2026-08-25
Figure.basemap: Improve the docstring for coltypes

seisman:
Usually, the `-f` option sets the data types of input/output columns. But in `basemap`, it only sets the coordinate types specified in the `-R` option (xref: https://docs.generic-mapping-tools.org/dev/basemap.html). So the common description for `coltypes` makes no sense in `Figure.basemap`.

This PR fixes the docstring for `coltypes` in `Figure.basemap` to clarify it.

**Preview**: https://pygmt-dev--4856.org.readthedocs.build/en/4856/api/generated/pygmt.Figure.basemap.html#pygmt.Figure.basemap


yvonnefroehlich (approved):
Interesting - new used this until now.

PR #4860, opened 2026-08-25
WIP: CI: Fix the Visual Studio path due to the upgrading of Windows runner image

seisman:
The `windows-latest`/`windows-2025` runner image upgraded its Visual Studio to 2026 in June, 2026 (https://github.blog/changelog/2026-05-14-github-actions-upcoming-image-migrations/#windows-2025-visual-studio-2026-image-migration).

This PR updates the Visual Studio path based on https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-VS2026-Readme.md#visual-studio-enterprise-2026. This update is necessary, otherwise the CI can't find Visual Studio and use MinGW GCC to compile the source code.

It's weird. The CI still can't find VS correctly (xref: https://github.com/GenericMappingTools/pygmt/actions/runs/32818816716/job/97712485125?pr=4860), but a similar fix in the GMT repository seems to be working (https://github.com/GenericMappingTools/gmt/actions/runs/32817632151/job/97709027710?pr=9144)

PR #3911, opened 2025-04-14
CI: Test on Windows arm64 runners

weiji14:
**Description of proposed changes**

Because why not have more failing Windows tests :laughing:

References:
- https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/

Similar to #3778 but for Windows

**Preview**:

**Reminders**

- [ ] Run `make format` and `make check` to make sure the code follows the style guide.
- [ ] Add tests for new features or tests that would have caught the bug that you're fixing.
- [ ] Add new public functions/methods/classes to `doc/api/index.rst`.
- [ ] Write detailed docstrings for all functions/methods.
- [ ] If wrapping a new module, open a 'Wrap new GMT module' issue and submit reasonably-sized PRs.
- [ ] If adding new functionality, add an example to docstrings or tutorials.

**Slash Commands**

You can write slash commands (`/command`) in the first line of a comment to perform
specific operations. Supported slash command is:

- `/format`: automatically format and lint the code

PR #4861, opened 2026-08-25
CI: Use setup-msvc-dev action instead of hardcoding the VS install path on Windows

seisman:
The Windows runner image upgrade moved Visual Studio from "...\2022\Enterprise\..." to "...\18\Enterprise\...", breaking the hardcoded vcvars64.bat path (see PR #4860). 

This PR uses the [`TheMrMilchmann/setup-msvc-dev`](https://github.com/TheMrMilchmann/setup-msvc-dev) action, which locates the VS installation dynamically via `vswhere` and exports the environment to later steps, so future runner image upgrades don't require another hardcoded-path fix.

Supersedes #4860.

### DIFF
--- a/.github/workflows/ci_tests_dev.yaml
+++ b/.github/workflows/ci_tests_dev.yaml
@@ -105,8 +105,14 @@ jobs:
           path: 'gmt'
           persist-credentials: false
 
-      # Build GMT from source on Linux/macOS, script is adapted from
-      # https://github.com/GenericMappingTools/gmt/blob/6.5.0/ci/build-gmt.sh
+      - name: Set up MSVC environment (Windows)
+        if: runner.os == 'Windows'
+        uses: TheMrMilchmann/setup-msvc-dev@368ef7d1ee4d1171b31d4a7f67f4d954f903f5a9 # v4.1.0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests_dev.yaml
```diff
@@ -105,16 +105,21 @@ jobs:
           path: 'gmt'
           persist-credentials: false
 
-      # Build GMT from source on Linux/macOS, script is adapted from
-      # https://github.com/GenericMappingTools/gmt/blob/6.5.0/ci/build-gmt.sh
+      - name: Set up MSVC environment (Windows)
+        if: runner.os == 'Windows'
+        uses: TheMrMilchmann/setup-msvc-dev@368ef7d1ee4d1171b31d4a7f67f4d954f903f5a9 # v4.1.0
+        with:
+          arch: x64
```
weiji14, line 112: Not an issue now, but we might need to set an `or` statement to the arch to  `x86_arm64` or `amd64_arm64` when https://github.com/GenericMappingTools/pygmt/pull/3911 lands.

weiji14 (approved):

PR #4877, opened 2026-09-04
CI: Add a composite action for caching Git LFS objects and saving LFS bandwidth

seisman:
Address the LFS bandwidth quota issue reported in https://github.com/GenericMappingTools/pygmt/issues/4738#issuecomment-5467473657.

This PR adds a composite action to cache Git LFS objects (i.e., baseline images) as GitHub cache, so that `git lfs pull` only pulls LFS objects when missing. It should help reduce our Git LFS bandwidth.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,9 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/actions/*"
     schedule:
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -3,7 +3,9 @@
 version: 2
 updates:
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
```
seisman, line 6: xref: https://github.com/dependabot/dependabot-core/issues/6704#issuecomment-3232918902.

PR #4881, opened 2026-09-04
Figure.histogram: Let parameter 'cumulative' support Pythonic arguments

seisman:
Migrate the parameter `cumulative` to the new alias system. 

Previously, `cumulative` can be either `True` or `"r"`. This PR lets it support more Pythonic argument, i.e., `cumulative="reverse"`.

For comparison, here is matplotlib's `hist` method (https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html):

> cumulativebool or -1, default: False
If True, then a histogram is computed where each bin gives the counts in that bin plus all bins for smaller values. The last bin gives the total number of datapoints.
> If density is also True then the histogram is normalized such that the last bin equals 1.
> If cumulative is a number less than 0 (e.g., -1), the direction of accumulation is reversed. In this case, if density is also True, then the histogram is normalized such that the first bin equals 1.

So, matplotlib uses `cumulative=-1` for reversed cumulative histogram, but I feel it's not as readable as `cumulative="reverse"`.

**Preview**: https://pygmt-dev--4881.org.readthedocs.build/en/4881/api/generated/pygmt.Figure.histogram.html#pygmt.Figure.histogram

### DIFF
--- a/pygmt/src/histogram.py
+++ b/pygmt/src/histogram.py
@@ -114,8 +114,8 @@ def histogram(
         * 1 = median and L1 scale (1.4826 \* median absolute deviation; MAD);
         * 2 = LMS (least median of squares) mode and scale.
     cumulative
-        Draw a cumulative histogram. Setting it to ``"reversed"`` to compute the
-        reversed cumulative histogram instead.
+        Draw a cumulative histogram. Set it to ``"reverse"`` to compute the reverse
+        cumulative histogram instead.
     extreme : str
         **l**\|\ **h**\|\ **b**.
         The modifiers specify the handling of extreme values that fall outside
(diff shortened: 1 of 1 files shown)


pygmt/src/histogram.py
```diff
@@ -112,10 +113,9 @@ def histogram(
         * 0 = mean and standard deviation [Default];
         * 1 = median and L1 scale (1.4826 \* median absolute deviation; MAD);
         * 2 = LMS (least median of squares) mode and scale.
-    cumulative : bool or str
-        [**r**].
-        Draw a cumulative histogram by passing ``True``. Use **r** to display
-        a reverse cumulative histogram.
+    cumulative
+        Draw a cumulative histogram. Set it to ``"reverse"`` to draw the reverse
```
yvonnefroehlich:
```suggestion 
        Draw a cumulative histogram by passing ``True``. Set it to ``"reverse"`` to draw the reverse
```

yvonnefroehlich:
> So, matplotlib uses cumulative=-1 for reversed cumulative histogram, but I feel it's not as readable as cumulative="reverse".

Agree. ``cumulative=-1`` is somehow smart from a developer perspective but not really intuitive for (new or learning) users.

yvonnefroehlich (commented):
Looks good to me.

Issue #4251, opened 2025-12-05
Figure.histogram: Calculate and plot histograms


seisman:
*This issue serves as the central place for discussing and tracking the implementation of the `histogram` method in PyGMT. The issue will be closed when the initial implementation is complete. Progress is tracked at https://github.com/orgs/GenericMappingTools/projects/3.*

## Documentation

- GMT: https://docs.generic-mapping-tools.org/dev/histogram.html
- GMT.jl: https://www.generic-mapping-tools.org/GMTjl_doc/documentation/modules/histogram
- PyGMT: https://www.pygmt.org/dev/api/generated/pygmt.Figure.histogram.html

## GMT Option Flags and Modifiers

☑️: *Implemented*; ⬜: *To be implemented/discussed*; ~~Strikethrough~~: *Won't implement*.

- [ ] `table`:
- [ ] `-A`
- [ ] `-B`: `frame`
- [ ] `-C`
- [ ] `-D`
- [ ] `-E`
- [ ] `-F`
- [x] `-G`: `fill`
- [ ] `-I`
- [ ] `-Jx|X`
- [ ] `-L`
- [ ] `-N`
- [ ] `-Q`
- [x] `-R`: `region`
- [ ] `-S`
- [ ] `-T`
- [ ] ~~`-U`~~: Use `Figure.timestamp` instead.
- [x] `-V`: `verbose`
- [x] `-W`: `pen`
- [ ] ~~`-X`/`-Y`~~: Use `Figure.shift_origin` instead.
- [ ] `-Z`
- [x] `-c`: `panel`
- [ ] `-bi`
- [ ] `-di`
- [ ] `-e`
- [ ] `-f`
- [ ] `-h`
- [x] `-i`: `incols`
- [ ] `-o`
- [x] `-p`: `perspective`
- [ ] `-qi`
- [ ] `-s`
- [x] `-t`: `transparency`
- [ ] `-w`
- [ ] ~~`--PAR=value`~~: Use `pygmt.config` instead.

## Notes on Input Formats

*Add any notes on the input formats, especially the meaning of columns.*

## Linked Pull Requests

- [ ] Initial feature implementation
- [ ] Add a tutorial or gallery example

## Related Issues and Discussions

*Add links to related wrapper discussions, API design threads, or upstream GMT changes here.*

PR #4886, opened 2026-09-09
Figure.histogram: Improved syntax for filling bars by position/value

seisman:
In `histogram`, bars can be filled by a constant color/pattern, or by the bin position or value  (count, frequency). This is achieved by the `-C` option, with a syntax like `-C<cmap>+b`.

This PR implements a Pythonic syntax for such features by extending the existing parameter `fill` (alias of `-G`). In addition to color or pattern, now `fill` can accept two special values: `"position"` and `"value"`.

## GMT CLI and PyGMT comparison

Here is a comparison:

| GMT CLI | PyGMT | Notes |
|---|---|---|
| `-Gred` | `fill="red"` | Fill by a constant color | 
| `-Gp5` | `fill=Pattern(5)` | Fill by a pattern |
| `-C` | `fill="position"` | Fill based on the current CPT and the mid-coordinate of each bar |
| `-Ctest.cpt` | `fill="position", cmap="test.cpt"` | Same as above but use `test.cpt` |
| `-C+b` | `fill="value"` | Fill based on the current CPT and the bin value (count or frequency, i.e., the bin height) |
| `-Ctest.cpt+b` | `fill="value", cmap="test.cpt"` | Same as above but use `test.cpt` |

## Examples
```
import numpy as np
import pygmt
from pygmt.params import Frame, Pattern

rng = np.random.default_rng(seed=100)
mean = 100  # mean of distribution
stddev = 25  # standard deviation of distribution
data = rng.normal(loc=mean, scale=stddev, size=521)

fig = pygmt.Figure()
fig.histogram(data=data, frame=Frame(axes="WSen", title="Fill by color"), series=10, fill="green")

fig.shift_origin(xshift="w+1")
fig.histogram(data=data, frame=Frame(axes="WSen", title="Fill by pattern"), series=10, fill=Pattern(10, dpi=50))

fig.shift_origin(xshift="w+1")
pygmt.makecpt(cmap="turbo", series=(0, 200, 25))
fig.histogram(data=data, frame=Frame(axes="WSen", title="Fill by position"), series=10, pen="1p", fill="position")
fig.colorbar()

fig.shift_origin(xshift="w+4")
pygmt.makecpt(cmap="turbo", series=(0, 100, 25))
fig.histogram(data=data, frame=Frame(axes="WSen", title="Fill by value"), series=10, pen="1p", fill="value")
fig.colorbar()

fig.show()
```
histogram

## Alternative implementation

An alternative implementation is to keep `fill` unchanged, but alias `-C` to two separate parameters, `cmap` and `fill_by_value` (or any other names). The cons are:

- User may be confused by the two similar parameters `fill` and `fill_by_value`
- `-G` can't be used with `-C`, so we need to deal with conflicts between `fill` and `cmap`/`fill_by_value` carefully

Address #4251

### DIFF
--- a/pygmt/src/histogram.py
+++ b/pygmt/src/histogram.py
@@ -87,15 +87,16 @@ def histogram(
         [Default is no outline].
     fill
         Set color or pattern for filling bars [Default is no fill]. Set it to one of the
-        two special values to fill bars by looking up the color from ``cmap`` instead of
-        a constant color:
+        two special values to fill bars by looking up the color from a CPT instead of a
+        constant color:
 
         - ``"position"``: look up the color using the mid-coordinate of the bin. This is
           the default when ``cmap`` is set.
         - ``"value"``: look up the color using the bin value, i.e., the bar count or
           frequency.
 
-        The special values require ``cmap``, and can't be used with ``fill``.
+        The special values require a CPT (either the current CPT or explicitely set by
+        ``cmap``), and can't be used with ``fill``.
     $cmap
     annotate : bool or str
         [**+b**][**+f**\ *font*][**+o**\ *off*][**+r**].
--- a/pygmt/src/histogram.py
+++ b/pygmt/src/histogram.py
@@ -12,6 +12,7 @@ from pygmt.exceptions import GMTParameterError
 from pygmt.helpers import (
     build_arg_list,
     fmt_docstring,
+    is_given,
     kwargs_to_strings,
     use_alias,
 )
@@ -43,7 +44,7 @@ def histogram(
     bar_offset: float | str | None = None,
     cmap: str | bool = False,
     pen: str | None = None,
-    fill: str | None = None,
+    fill: str | Literal["position", "value"] | None = None,
     horizontal: bool = False,
     projection: str | None = None,
     region: Sequence[float | str] | str | None = None,
@@ -64,7 +65,7 @@ def histogram(
     $aliases
        - A = horizontal
        - B = frame
-       - C = cmap
+       - C = cmap, **+b**: fill
        - E = bar_width, **+o**: bar_offset
        - G = fill
        - J = projection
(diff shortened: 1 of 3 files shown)


pygmt/src/histogram.py
```diff
@@ -87,12 +88,22 @@ def histogram(
...
+        Set color or pattern for filling bars [Default is no fill]. Set it to one of the
+        two special values to fill bars by looking up the color from a CPT instead of a
+        constant color:
+
+        - ``"position"``: look up the color using the mid-coordinate of the bin. This is
+          the default when ``cmap`` is set.
+        - ``"value"``: look up the color using the bin value, i.e., the bar count or
+          frequency.
+
+        The special values require a CPT (either the current CPT or explicitely set by
```
yvonnefroehlich:
```suggestion
        The special values require a CPT (either the current CPT or explicitly set by
```

pygmt/src/histogram.py
```diff
@@ -87,12 +88,22 @@ def histogram(
...
         Draw bar outline (or stair-case curve) using the specified pen thickness
         [Default is no outline].
     fill
-         Set color or pattern for filling bars [Default is no fill].
+        Set color or pattern for filling bars [Default is no fill]. Set it to one of the
+        two special values to fill bars by looking up the color from a CPT instead of a
+        constant color:
+
+        - ``"position"``: look up the color using the mid-coordinate of the bin. This is
+          the default when ``cmap`` is set.
```
yvonnefroehlich, line 100: Here we write "bin", but the GMT says "bar", please see https://docs.generic-mapping-tools.org/dev/histogram.html#c:

> Give a CPT. The mid-coordinate for each bar is used to look up the bar color.

I agree that "bin" makes more sense. The x-position of the bars can be changed, e.g., via ``bar_offset``.

seisman: I think "bar" represents the visual element, while "bin" represents the actual data range of a bar.

PR #4890, opened 2026-09-15
Figure.histogram: Add the parameter 'weights' for weighting data

seisman:
In `histogram`, `-Z+w` means using the second data column as weights. This PR provides the `weights` parameter which is more Pythonic. It can cover three use cases:

1. `data` is a file: `weights=True` means using the 2nd column as weights 
2. `data` is a 2-D array: `weights=True` means using the 2nd column as weights
3. `data` is a 1-D array: `weights` must be a 1-D array as weights

For comparison, [`matplotlib.pyplot.hist`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html) also provides the `weights` parameter.

Here is an example showing that PyGMT and matplotlib produce the same plot:
```python
data = [1, 1, 1, 1, 1, 1, 2, 2, 2, 3, 4, 5, 6, 7, 8, 8, 8, 8, 8, 8]
weights = [0.1, 0.1, 0.2, 0.2, 0.3, 0.3, 0.4, 0.4, 0.5, 0.5, 0.6, 0.6, 0.7, 0.7, 0.8, 0.8, 0.9, 0.9, 1.0, 1.0]

import numpy as np

plt.hist(data, weights=weights, bins=np.arange(0, 10, 1))
plt.show()


import pygmt
fig = pygmt.Figure()
fig.histogram(data=data, weights=weights, series=1, fill="lightblue", frame=True, region=[0, 10, 0, 0])
fig.show()
```


pygmt/tests/test_histogram.py
```diff
@@ -45,9 +76,74 @@ def test_histogram_baroffset(data):
         fig.histogram(
             data=data,
             projection="X10c/10c",
-            region=[0, 9, 0, 6],
+            region=[0, 10, 0, 6],
```
seisman:
```suggestion
            region=[0, 9, 0, 6],
```

PR #4910, opened 2026-09-22
Bump to setuptools>=80 and setuptools-scm>=9.2

seisman:
Bump setuptools and setuptools-scm following the official recommendations of setuptools-scm (https://setuptools-scm.readthedocs.io/latest/usage/#simplified-activation-new).


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=77", "setuptools_scm[toml]>=6.2"]
```
seisman, line 2: `setuptools-scm[toml]` is changed to `setuptools-scm[simple]`. 

`toml` is no longer needed because setuptools-scm started to use the built-in `tomllib` library since 2022 (https://github.com/pypa/setuptools-scm/issues/740)

pyproject.toml
```diff
@@ -1,5 +1,5 @@
...
+requires = ["setuptools>=80", "setuptools-scm[simple]>=9.2"]
```
weiji14: Can we bump to `setuptools-scm>=10.0.4` to enable `pre_parse = "fetch_on_shallow"` mentioned in #4909?

seisman: Probably, still need to test if `pre_parse = "fetch_on_shallow"` works as expected.

Another concern is that setuptools-scm 10.0.4 was released in March, 2026, which is quite new.

pyproject.toml
weiji14: > `setuptools-scm[toml]` is changed to `setuptools-scm[simple]`.

I don't think we're supposed use `simple` since we're using a custom version scheme in `[tool.setuptools_scm]`? At least that's what I'm inferring from the docs at https://setuptools-scm.readthedocs.io/v9.2.2/usage/#simplified-activation-new:

image

So maybe remove the `[simple]`?

pyproject.toml
weiji14: Ok, let's just do a simple bump to `setuptools-scm>=9.2` here, and then try `>=10.0.4` in #4909. 6 months isn't too bad, at least this isn't a user-facing dependency so we don't need to wait for a year or something.

weiji14:
Btw, I've manually triggered CI at https://github.com/GenericMappingTools/pygmt/actions/runs/35690233375 to check that things work ok.

pyproject.toml
seisman: I've removed `[simple]` and it still works.

From what I understand, `[simple]` is meant for use cases with minimal configuration (e.g.,, no `[tool.setuptools_scm]` section).

See the "Legacy simplified activation (removed in 9.2.0)" box at https://setuptools-scm.readthedocs.io/latest/usage/#using-setuppy-to-pass-code

weiji14 (approved):

